PR #665, opened 2013-09-27
Qt Embedded 4.8.5 ( and FastCGIqt to test )

Ximi1970:
Adding toolchains with Gcc 4.7.3, Gcc 4.8.1 and Glibc 2.18.
Adding libqt4-embedded that will require the new toolchains.
These extensions have been tested on cedarview and 88f6281, but should also work for bromolow and x86.

Part 3 of 3: Adding libqt4-embedded (and Helloworld-FastCGIqt package for testing)


Ximi1970:
You can  use my copy ( it is not a branch ) of spksrc. You can find it here: https://github.com/Ximi1970/spksrc-ximi

Diaoul:
Not enough knowledge on my side to be able to maintain this. Moreover, it seemed like an edge case so better in a fork until we really need it.

Also I don't really get it, Qt for headless?

HakShak:
Unfortunately, Qt has a lot of constructs that get used outside of the GUI domain. For example, I was looking into this to get a quassel core compiled.

PR #795, opened 2014-01-08
Etherpad SPK

Dr-Bean:


ndouglas:
This still works with 5.1, but SPK_DEPENDS is set to "node", and I suppose the official Node.js package has some different identifier.  Removing the line (with the official package installed, of course) allows installation and everything seems functional.

PR #960, opened 2014-05-01
Add OctoPrint package

Decipher:
OctoPrint is a python web based 3D printer host that allows you to control your 3D printer.

The only outstanding issue is that PyYAML can't be packaged as a wheel, as per #764.


Decipher:
Hmm, rebase not the best approach :|

Dr-Bean:
Btw, not sure if you noticed, but PyYAML is now a Python module (add an SPK_DEPENDS on Python>=2.7.8-11). No need to build it separately anymore.

Decipher:
Yup, was jumping back onto the issue in the hope of getting it wrapped up, but it's been a while since I've done anything spksrc.

PR #991, opened 2014-05-13
Vim: src/*.sh scripts fix + vimdiff symlink

kmarty:
src/dsm-control.sh:
	- Add simple check for "status" + appropriate return codes
	- Fix "log" return code to non-error state

src/installer.sh:
	- Add "vimdiff" symlink into /usr/local/bin

### DIFF
--- a/spk/vim/src/dsm-control.sh
+++ b/spk/vim/src/dsm-control.sh
@@ -13,10 +13,14 @@ case $1 in
         exit 0
         ;;
     status)
-        exit 1
+	if [ -e /usr/local/bin/vim ]; then
+		exit 0
+	else
+		exit 150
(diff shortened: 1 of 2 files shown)


spk/vim/src/dsm-control.sh
```diff
@@ -13,10 +13,14 @@ case $1 in
         exit 0
         ;;
     status)
-        exit 1
+	if [ -e /usr/local/bin/vim ]; then
```
Dr-Bean: Realistically, the package could be considered broken only if `/usr/local/vim/bin/vim-utf8` is not available. The symlink is just to make life easier, but it's hardly critical if it's not there.

PR #1024, opened 2014-05-27
pyLoad: New package

kmarty:
+ +cross/tesseract-ocr

- No Pyload control panel - just basic preinstall questions + htaccess.example for possible access to pyload through virtualhost reverseproxy.

Added patches:
- "hack" to allow listen on IPv4&IPv6 "::" address
- replace '/' to '_' in downloaded filename (it has really happen)
- added simple checking for extended filename content disposition.


Dr-Bean:
My apologies for the delay (time and priorities and all that jazz). Overall, looks ok, although I'll probably come up with some comments here and there ;)
The installer and start-stop script are quite large and I'll be taking some time to look at what's going on there.

That said, I noticed the patches that you added (IPv6 support, renaming and extended filenames) Have you considered entering those as patches over on the pyload repository, or are those perhaps already merged?

spk/pyload/Makefile
```diff
@@ -0,0 +1,59 @@
...
+	install -m 644 src/default_syn.conf $(STAGING_DIR)/share/pyload/module/config/default_syn.conf
+	install -m 644 src/pyload_init.sql $(STAGING_DIR)/share/pyload/module/config/pyload_init.sql
+	install -m 644 src/htaccess.example $(STAGING_DIR)/share/pyload/module/config/htaccess.example
+	install -m 755 -d $(STAGING_DIR)/app/images
+	for size in 16 24 32 48 64 72 256; do \
+		convert $(SPK_ICON) -thumbnail $${size}x$${size} \
+		        $(STAGING_DIR)/app/images/$(SPK_NAME)-$${size}.png ; \
+	done
+
+$(WORK_DIR)/.tessdata_install_done:
```
Diaoul: What is that? You already depend on tesseract-ocr in `DEPENDS`

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+# Package
+PACKAGE="pyload"
+DNAME="pyLoad"
+
+PIDFILE="/var/run/pyload.pid"
+PYLOADCORE="/usr/local/python/bin/python /usr/local/pyload/share/pyload/pyLoadCore.py --pidfile=${PIDFILE}"
+export PATH="/usr/local/pyload/bin:$PATH"
```
Diaoul: You don't need export here

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,146 @@
...
+        mkdir -p "${SYNOPKG_PKGINST_TEMP_DIR}/etc"
+        touch "${SYNOPKG_PKGINST_TEMP_DIR}/etc/pyload-pkg.conf"
+
+        # Check whether ${wizard_download_dir} is in existing shared folder.
+        VOLUME=$(echo "${wizard_download_dir}" | cut -d/ -f2)
+        WIZ_SHARE=$(echo "${wizard_download_dir}" | cut -d/ -f3)
+        SHARE_EXISTS=$(synoshare --get "${WIZ_SHARE}" | sed -ne 's/^[[:space:]]*Path[[:space:]\.]*\[\(.*\)\].*$/\1/p')  # FIXME: Better variable name
+
+        if [ -z "${SHARE_EXISTS}" ]; then
+            synoshare --add "${WIZ_SHARE}" "" "/${VOLUME}/${WIZ_SHARE}" "" "" "" 1 0 > ${SYNOPKG_TEMP_LOGFILE} || exit 1
```
Diaoul: I'm not a fan of creating shares in packages, just test if the folder the user input exists and if not just exit gracefully with an error message.

Dr-Bean:
Heh...worried I wasn't going to comment on the same items, @Diaoul? :P

kmarty:
Dr-Bean: IPv6 support is "hack" to allow "::" instead of IPv4-only "0.0.0.0" (just pyload config parser issue) and I'm not sure if it'll be accepted as is directly in pyload repo. Patch with Extended Content-Disposition is now part of 0.4.10 (not released yet and 0.4.9 doesn't have support to it).

spk/pyload/Makefile
kmarty: "tesseract-ocr" is binary only. Any tessdata package is optional. I thought it is the best keep tesseract-ocr without any tessdata and if any needed (pyload in this case), it downloads them itself. Actually I thought it is not good idea put all available tessdata packages into "tesseract-ocr" package.
Or do you mean something different?

spk/pyload/src/dsm-control.sh
kmarty: I'm not sure. But I recheck it, of course.

spk/pyload/src/installer.sh
kmarty: OK, I'll change it.

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,146 @@
...
+    rm -f ${INSTALL_DIR}
+
+    exit 0
+}
+
+preupgrade ()
+{
+    # Keep existing cfg while upgrading
+    if [ -d "${SYNOPKG_PKGINST_TEMP_DIR}" -a -d "${INSTALL_DIR}/etc" ]; then
+        cp -a "${INSTALL_DIR}/etc" "${SYNOPKG_PKGINST_TEMP_DIR}"
```
Dr-Bean: Wouldn't this leave stuff behind? Otherwise, do a cleanup in `postupgrade()`

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,146 @@
...
+
+    exit 0
+}
+
+preuninst ()
+{
+    if [ "${SYNOPKG_PKG_STATUS}" == "UNINSTALL" ]; then
+	test -f "${INSTALL_DIR}/etc/pyload-pkg.conf" && . "${INSTALL_DIR}/etc/pyload-pkg.conf"
+        if [ "${WIZ_DWNLD_DIR}" ]; then
+            :   # IDEA: If download folder was created by postinst(), remove it. But do I know whether user didn't start using it for anything else after install?
```
Dr-Bean: To answer the question: We never remove dirs. Once created (regardless of how that happened), they are not yours to control anymore.

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,146 @@
...
+        SALT=$((RANDOM%99999+10000))
+        SALTED_PW_HASH=${SALT}$(echo -n "${SALT}${wizard_password}" | openssl dgst -sha1 2>/dev/null | cut -d" " -f2)
+
+        # init DB & add 'admin' user
+        echo -n "4" > "${INSTALL_DIR}/etc/files.version"
+        sqlite3 "${INSTALL_DIR}/etc/files.db" < "${INSTALL_DIR}/share/pyload/module/config/pyload_init.sql" > ${SYNOPKG_TEMP_LOGFILE} || exit 1
+        sqlite3 "${INSTALL_DIR}/etc/files.db" "INSERT INTO users (name, password) VALUES ('admin', '${SALTED_PW_HASH}')" > ${SYNOPKG_TEMP_LOGFILE} || exit 1
+
+        # Additional (optional) config/examples
+        cp -f "${INSTALL_DIR}/share/pyload/module/config/htaccess.example" "${INSTALL_DIR}/etc/"
```
Dr-Bean: See my earlier comment: put these files in `$(INSTALL_DIR/var` during build, then you don't need to move it around afterwards.

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/sh
+
+# Package
+PACKAGE="pyload"
+DNAME="pyLoad"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PATH="${INSTALL_DIR}/bin:/usr/local/bin:/bin:/usr/bin:/usr/syno/bin:/usr/syno/sbin"
```
Dr-Bean: Use ${PATH} here, see other packages.

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,42 @@
...
+    start)
+        if [ -f /usr/syno/etc/ssl/ssl.crt/server.crt -a \
+             -f /usr/syno/etc/ssl/ssl.key/server.key ]; then
+            ln -s /usr/syno/etc/ssl/ssl.crt/server.crt /usr/local/pyload/etc/ssl.crt
+            ln -s /usr/syno/etc/ssl/ssl.key/server.key /usr/local/pyload/etc/ssl.key
+            sed -i -e 's/bool https : "Use HTTPS" = \(True|False\)/bool https : "Use HTTPS" = True/' /usr/local/pyload/etc/pyload.conf
+        else
+            sed -i -e 's/bool https : "Use HTTPS" = \(True|False\)/bool https : "Use HTTPS" = False/' /usr/local/pyload/etc/pyload.conf
+        fi
+        ${PYLOADCORE} --daemon
```
Dr-Bean: Looks like the package runs as root: that's a security risk. It'd be better to use a specific user for the package (use busybox' adduser, for instance)

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,42 @@
...
+PACKAGE="pyload"
+DNAME="pyLoad"
+
+PIDFILE="/var/run/pyload.pid"
+PYLOADCORE="/usr/local/python/bin/python /usr/local/pyload/share/pyload/pyLoadCore.py --pidfile=${PIDFILE}"
+export PATH="/usr/local/pyload/bin:$PATH"
+
+case $1 in
+    start)
+        if [ -f /usr/syno/etc/ssl/ssl.crt/server.crt -a \
```
Dr-Bean: Seeing as this is a one-time thing, it'd be best to move this to the installer code.
You could add an `enable SSL` step in the wizard if it's optional.

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Package
+PACKAGE="pyload"
+DNAME="pyLoad"
+
+PIDFILE="/var/run/pyload.pid"
+PYLOADCORE="/usr/local/python/bin/python /usr/local/pyload/share/pyload/pyLoadCore.py --pidfile=${PIDFILE}"
+export PATH="/usr/local/pyload/bin:$PATH"
+
+case $1 in
```
Dr-Bean: I'd prefer the same structure as we use in other packages: split off the various sections into their own functions, and call those functions.

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+# Package
+PACKAGE="pyload"
+DNAME="pyLoad"
+
+PIDFILE="/var/run/pyload.pid"
+PYLOADCORE="/usr/local/python/bin/python /usr/local/pyload/share/pyload/pyLoadCore.py --pidfile=${PIDFILE}"
```
Dr-Bean: Change this to `PYLOADCORE = "${INSTALL_DIR}/share/pyload/pyLoadCore.py`, and call something like this in the start function:
`${PYTHON} ${PYLOAD} --pidfile ${PIDFILE}`

Along with that, you should use a Python virtual environment instead of the system Python (to be created during install). Other Python-based packages could serve as an example.

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+# Package
+PACKAGE="pyload"
+DNAME="pyLoad"
+
+PIDFILE="/var/run/pyload.pid"
```
Dr-Bean: Move the pidfile to `${INSTALL_DIR}/var/pyload.pid`, we want packages to be self-contained as much as possible.

spk/pyload/Makefile
```diff
@@ -0,0 +1,59 @@
...
+.PHONY: pyload_extra_install
+pyload_extra_install: $(WORK_DIR)/.tessdata_install_done
+	install -m 755 -d $(STAGING_DIR)/app
+	install -m 644 src/app/config $(STAGING_DIR)/app/config
+	install -m 755 src/app/pyload.cgi.sh $(STAGING_DIR)/app/pyload.cgi
+	install -m 644 src/default_syn.conf $(STAGING_DIR)/share/pyload/module/config/default_syn.conf
+	install -m 644 src/pyload_init.sql $(STAGING_DIR)/share/pyload/module/config/pyload_init.sql
+	install -m 644 src/htaccess.example $(STAGING_DIR)/share/pyload/module/config/htaccess.example
+	install -m 755 -d $(STAGING_DIR)/app/images
+	for size in 16 24 32 48 64 72 256; do \
```
Dr-Bean: We don't use 256, best to keep in line with the other packages :)

spk/pyload/Makefile
```diff
@@ -0,0 +1,59 @@
...
+TESSDATA_DIST_NAME = tesseract-ocr-3.02.eng.tar.gz
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: pyload_extra_install
+pyload_extra_install: $(WORK_DIR)/.tessdata_install_done
+	install -m 755 -d $(STAGING_DIR)/app
+	install -m 644 src/app/config $(STAGING_DIR)/app/config
+	install -m 755 src/app/pyload.cgi.sh $(STAGING_DIR)/app/pyload.cgi
+	install -m 644 src/default_syn.conf $(STAGING_DIR)/share/pyload/module/config/default_syn.conf
```
Dr-Bean: Looks like you could just rename this file to pyload.conf, and put it in ${INSTALL_DIR}/var, right? Same goes for the other files you're adding in the next lines: put them in the same location.

Dr-Bean:
I've added some comments, take a look :)

Thanks for the answer on the patches. Maybe it's worth seeing if the devs know about the IPv6 issue: even though your solution might not be optimal, adding support in some way would make sense imo.

spk/pyload/src/app/config
```diff
@@ -0,0 +1,12 @@
+{
+    ".url": {
+        "com.synocommunity.packages.pyload": {
+            "title": "pyLoad",
+            "desc": "pyLoad",
+            "icon": "images/pyload-{0}.png",
+            "type": "url",
+            "url": "3rdparty/pyload/pyload.cgi",
```
Dr-Bean: I'm not sure about this approach. Seeing as the interface is fixed to port 5080, wouldn't it make sense to just use that as the url?

spk/pyload/src/app/config
```diff
@@ -0,0 +1,12 @@
...
+            "allUsers": true
+        }
```
Dr-Bean: Maybe you'd want to add a `"grantPrivilege": local` to keep the same behaviour as we use for other packages.

spk/pyload/Makefile
Dr-Bean: Maybe add it in the cross/tesseract-ocr package? Looks like this one is more or less the only mandatory package, right?

Diaoul: Create a cross/tessdata like cross/sabnzbd then. Otherwise it will download the tarball for each architecture.

Diaoul:
Well, @Dr-Bean did all the job so I have no more comments to add.

spk/pyload/src/installer.sh
kmarty: "/var", not "/etc"?
I don't really think that is a good idea to use non-standard locations. The "/etc" is for configurations and it is pyload's default. Keeping them in their default locations makes things easier with pyload itself.

Diaoul: `/var` and `/etc` are both eligible for backup during package update but it's best if you can centralize everything in one directory, e.g. `/var`.

For the `salt-master` SPK it made thing easier too to use `/etc`, take a look at the installer if you want.

spk/pyload/src/app/config
kmarty: Actually, it isn't. TCP port is freely configurable through Pyload WebUI, port 5080 is just a predefined (although different than pyloads default port 8001). That reminds me complication with "pyload.sc" file.

spk/pyload/src/app/config
kmarty: Added.

spk/pyload/src/dsm-control.sh
kmarty: Maybe I'm wrong, but SSL certificates are created by DSM only when necessary. If you turn off SSL in DSM, they're deleted (I have to recheck it, but I think it is so) => Pyload doesn't start. If you turn SSL on, then there is high probability you want it also in pyload.

Dr-Bean: That might be, but the symlinks would only work if you run the package as root, and that should change anyway. 
An alternative is to copy the certificate files to your /var directory, correct the permissions, and use those for SSL access.

spk/pyload/src/app/config
Dr-Bean: Well, how about removing the complication :) If nothing else, predetermining a port removes ambiguity for less experienced users, and helps us with package support, because everyone basically starts off with the same configuration.

So, I'd propose to set the port to a fixed (free) port. Check the wiki pages to be sure which ports are available. You can use the directions at the top to determine which port to use, probably somewhere in the 81xx range or so. Then, use that port for `ADMIN_PORT`, the service configure file and the config/shortcut.

kmarty: OK, I can set everything static. Except Pyload itself, user can change listening port anytime through Pyload WebUI.

Diaoul: That's the case for SABnzbd/SickBeard/CouchPotato too and the port is fixed anyway. If one changes the port number he should know the implications of that so that's OK.

This is how we do to avoid useless overhead in the code.

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,128 @@
...
+            s|@LNG@|$(lng2iso ${SYNOPKG_DSM_LANGUAGE})| \
+               " \
+                  < "${INSTALL_DIR}/share/pyload/module/config/default_syn.conf" \
+                  > "${INSTALL_DIR}/etc/pyload.conf"
+
+        # hash password
+        SALT=$((RANDOM%99999+10000))
+        SALTED_PW_HASH=${SALT}$(echo -n "${SALT}${wizard_password}" | openssl dgst -sha1 2>/dev/null | cut -d" " -f2)
+
+        # init DB & add 'admin' user
```
Dr-Bean: According to the documentation, the more correct way to init the database and add the user, is to use `python pyLoadCore.py -s`, where you can insert the username, password and email of the admin account (db init is part of Pyload itself)
Seems to me that's a better way than the current method, so that might be worth investigating.

kmarty: Yes, standard way is `python pyLoadCore.py -s`. But it is interactive setup only, unusable for SPK package. It is the only reason why I'm doing it this way.

Diaoul: What about using expect? Or patch the pyLoadCore such that it has a non-interactive mode? Upstream could then merge this

Dr-Bean: Was just testing it. About 20 entries, of which you'd need to change 6, and simply <enter> 14 times. I'd agree that it's not ideal at all...

kmarty: Expect is not available in DSM.
Patching pyload because of this is out of my ability.

Diaoul: We can include expect in the SPK. If I have enough time I'll try to take a look at patching pyLoad to make a non-interactive mode.

spk/pyload/src/app/config
kmarty: SABnzbd and SickBeard uses the same way to detect http/https and port using CGI script.

Dr-Bean: Just bad examples ;) I should get to rebasing/merging the PR I had for that eventually. Sickbeard-custom or CP-custom are better examples.

kmarty: OK, I set it to static on http. Everyone who want to use it with SSL has to set everything manually.

spk/pyload/src/dsm-control.sh
kmarty: I know. I have to look on it in detail because there were struggles about changing ownership and permission of downloaded files.

spk/pyload/Makefile
kmarty: Done

spk/pyload/src/dsm-control.sh
kmarty: Done

spk/pyload/src/installer.sh
kmarty: Done

spk/pyload/src/installer.sh
kmarty: Rewritten

spk/pyload/src/installer.sh
kmarty: Deleted, folder left as is.

spk/pyload/src/installer.sh
kmarty: Done

spk/pyload/src/installer.sh
kmarty: Done

spk/pyload/src/dsm-control.sh
kmarty: Done. Now it runs as user `pyload`

spk/pyload/src/dsm-control.sh
kmarty: Done. Anything about SSL removed, standard installation is HTTP only.

spk/pyload/src/dsm-control.sh
kmarty: Done

spk/pyload/src/dsm-control.sh
kmarty: Done

spk/pyload/src/dsm-control.sh
kmarty: Done

spk/pyload/Makefile
kmarty: Removed

spk/pyload/Makefile
kmarty: Done. Files are in `etc` at build.

spk/pyload/src/app/config
kmarty: Done

spk/pyload/src/installer.sh
kmarty: I don't want to arguing, but ... :-)
What's wrong with that? Whole setup is about content of `etc/pyload.conf` (package has own, prefilled, setup is not required) and database. Database is standard sqlite database, its structure is known and only one insert is required. I thought it is the easiest way how to do it.

spk/pyload/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	install -m 644 src/pyload_init.sql $(STAGING_DIR)/etc/pyload_init.sql
+	install -m 644 src/htaccess.example $(STAGING_DIR)/etc/htaccess.example
+	install -m 755 -d $(STAGING_DIR)/var
+	install -m 755 -d $(STAGING_DIR)/app/images
+	for size in 16 24 32 48 64 72; do \
+		convert $(SPK_ICON) -thumbnail $${size}x$${size} \
+		        $(STAGING_DIR)/app/images/$(SPK_NAME)-$${size}.png ; \
+	done
+
+$(STAGING_DIR)/share/requirements.pybundle:
```
Dr-Bean: If configobj is required, could you change it to how I did it in this commit: https://github.com/SynoCommunity/spksrc/commit/4e0e61c544b267e05b87e019c0b2af1ba3fb372c?
Pybundle is deprecated, and it's in the works to replace it for all packages (otherwise we'll never be able to update Pip anymore)

spk/pyload/src/installer.sh
Dr-Bean: Yes, it's easy, but it's harder on maintenance, e.g. handling changes to the database model would be more difficult.
I assume you installed the package manually, and got the db model from a working installation, right? To be sure we wouldn't miss a change, we'd need to do that on every update.

kmarty: Changes database model could be but:
- on new release of pyLoad, not SPK
- pyLoad has own upgrade mechanism when DB model is changed (has older model respectivelly), so this creation is required on new installation only. Even if new installation had an old model (that's what `files.version` file is for), it'll be upgraded to current model on first start of pyLoad.

DB creation is on `pyload/share/pyload/module/database/DatabaseBackend.py` but of course, it was easier to me to dump using `sqlite3 files.db .schema`.

spk/pyload/Makefile
kmarty: Actually I know nothing about configobj. It was blindly copied from another pkg.

cross/tessdata-eng/Makefile
```diff
@@ -0,0 +1,25 @@
+TESSDATA_DIST_SITE = http://tesseract-ocr.googlecode.com/files
```
Dr-Bean: Leftovers on line 1 and 2 :)

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,76 @@
...
+PYTHON="${INSTALL_DIR}/env/bin/python"
+PYLOAD="${INSTALL_DIR}/share/pyload/pyLoadCore.py"
+LOG_FILE="${INSTALL_DIR}/etc/Logs/log.txt"
+PID_FILE="${INSTALL_DIR}/var/pyload.pid"
+
+
+
+start_daemon ()
+{
+    su - ${USER} -c "PATH=${PATH} ${PYTHON} ${PYLOAD} --pidfile=${PID_FILE} --daemon"
```
Dr-Bean: Use `su ${USER} -c "PATH=${PATH} ${PYTHON} ${PYLOAD} --pidfile=${PID_FILE} --daemon"` instead (removed the minus sign after 'su')

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,76 @@
...
+}
+
+stop_daemon ()
+{
+    test "${SYNOPKG_TEMP_LOGFILE}" && exec > "${SYNOPKG_TEMP_LOGFILE}"
```
Dr-Bean: No need for this, right?

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,76 @@
...
+    ${PYTHON} ${PYLOAD} --pidfile=${PID_FILE} --quit
```
Dr-Bean: Wrap this in `su` too

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,76 @@
...
+}
+
+daemon_status ()
+{
+    ${PYTHON} ${PYLOAD} --pidfile=${PID_FILE} --status > /dev/null
```
Dr-Bean: ...and this one too

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,157 @@
...
+
+        if [ -z "${SHARE_EXISTS}" ]; then
+               echo "ERROR: Share '${WIZ_SHARE}' doesn't exist."
+               exit 1
+        fi
+
+        if [ ! -d ${wizard_download_dir} ]; then
+            echo "ERROR: Download directory ${wizard_download_dir} doesn't exist."
+            exit 1
+        fi
```
Dr-Bean: This is shorter (taken from sabnzbd installer): it doesn't matter if it's a shared folder, just that the dir exists.
```
 # Check directory
    if [ ! -d ${wizard_download_dir:=/volume1/downloads} ]; then
        echo "Download directory ${wizard_download_dir} does not exist."
        exit 1
    fi
fi

File: spk/pyload/src/installer.sh [hunk 25]
```diff
@@ -0,0 +1,157 @@
...
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # Create a Python virtualenv
+    ${VIRTUALENV} --system-site-packages ${INSTALL_DIR}/env > /dev/null
+
+    # Create user
+    adduser -h ${INSTALL_DIR}/etc -g "${DNAME} User" -G ${GROUP} -s /bin/sh -S -D ${USER}
+
+    # Set cfg location
+    echo "${INSTALL_DIR}/etc" > "${INSTALL_DIR}/share/pyload/module/config/configdir"
```
Dr-Bean: If `/usr/local/pyload/etc` is the default, do you need to set it explicitly?

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,157 @@
...
+
+    if [ "${SYNOPKG_PKG_STATUS}" == "INSTALL" ]; then
+        test "${wizard_download_dir}" || wizard_download_dir="/volume1/downloads"
```
Dr-Bean: This can be removed, see comment below

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,157 @@
...
+
+        # Set group and permissions on download dir for DSM5
+        if [ `/bin/get_key_value /etc.defaults/VERSION buildnumber` -ge "4418" ]; then
+            chgrp users "${wizard_download_dir}"
```
Dr-Bean: If `wizard_download_dir` is blank (even though you prevent that in the wizard), this'll prefill it with the default value:
```
chgrp users ${wizard_download_dir:=/volume1/downloads}
chmod g+rw ${wizard_download_dir:=/volume1/downloads}
```

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,157 @@
...
+            chmod g+rwx "${wizard_download_dir}"
+        fi
+
+        # Edit the configuration according to the wizard
+        sed -i -e " \
```
Dr-Bean: Can you check the indents here? Check the [syncserver installer](https://github.com/SynoCommunity/spksrc/blob/develop/spk/syncserver/src/installer.sh#L63) for example.

spk/pyload/src/wizard/install_uifile
```diff
@@ -0,0 +1,32 @@
...
+                    "errorText": "Path should begin with /volume?/ with ? the number of the volume"
+                }
+            }
+        }]
+    }]
+}, {
+    "step_title": "Basic configuration",
+    "items": [{
+        "type": "password",
+        "desc": "pyLoad web interface needs own 'admin' account. Please enter a password for the pyLoad 'admin' user",
```
Dr-Bean: Maybe change this to: `"An 'admin' account will be created for the pyLoad web interface. Please enter a password for the pyLoad 'admin' user."`

Dr-Bean:
Second review. Nice work so far :)

spk/pyload/src/dsm-control.sh
kmarty: omg. Yep, you are right :-)

spk/pyload/src/installer.sh
kmarty: "default" as "system default" = /etc
The real pyLoads "default" is `.pyload` in $HOME.

Dr-Bean: Right, let's keep that then ;) Unless, of course, we get the expect/non-interactive mode in :)

cross/tessdata-eng/Makefile
kmarty: Guilty :-)

spk/pyload/src/installer.sh
kmarty: I saw it in sanbnzd, but I thought that checking for "share" will be usefull. OK, I'll change it.

Dr-Bean: Here's why: synoshare fails to recognize a 'normal' unix dir, because of the missing properties (ACLs and whatever else Synology might add to shared folders)
On the other hand, a normal dir check works on both.

spk/pyload/src/installer.sh
kmarty: I'm sorry, my english is not good. "Keep" as "keep that line in script" or "keep that real default location"?

spk/pyload/src/installer.sh
```diff
@@ -0,0 +1,143 @@
...
+    if [ "${SYNOPKG_PKG_STATUS}" == "INSTALL" ]; then
+        # Set group and permissions on download dir for DSM5
+        if [ `/bin/get_key_value /etc.defaults/VERSION buildnumber` -ge "4418" ]; then
+            chgrp users "${wizard_download_dir:=/volume1/downloads}"
+            chmod g+rwx "${wizard_download_dir:=/volume1/downloads}"
+        fi
+
+        # Edit the configuration according to the wizard
+        sed -i -e "s|@W_DOWNLOAD_DIR@|${wizard_download_dir:=/volume1/downloads}|" \
+               -e "s|@LNG@|$(lng2iso ${SYNOPKG_DSM_LANGUAGE})|" \
```
Dr-Bean: I'm wondering if it's worth changing the interface language based on the device language, instead of just defaulting to 'en'? I assume you can change the language via the interface, right?

kmarty: It is, ehm, convenience(?) :-)

spk/pyload/Makefile
```diff
@@ -0,0 +1,45 @@
...
+POST_STRIP_TARGET = pyload_extra_install
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: pyload_extra_install
+pyload_extra_install:
+	install -m 755 -d $(STAGING_DIR)/app
+	install -m 644 src/app/config $(STAGING_DIR)/app/config
+	install -m 755 -d $(STAGING_DIR)/etc
+	install -m 644 src/default_syn.conf $(STAGING_DIR)/etc/pyload.conf
```
Dr-Bean: Forgot this one: Probably rename this file to `pyload.conf`, instead of `default_syn.conf`

spk/pyload/src/installer.sh
Dr-Bean: I wouldn't mind too much if it stays, just thinking out loud. Personally, I don't like language changes based on my locale ;)

spk/pyload/src/installer.sh
Dr-Bean: I'm sorry, I see that it could be misunderstood. I mean: keep that line exactly as it is, no changes :)

spk/pyload/src/dsm-control.sh
```diff
@@ -0,0 +1,75 @@
...
+}
+
+stop_daemon ()
+{
+    su ${USER} -c "PATH=${PATH} ${PYTHON} ${PYLOAD} --pidfile=${PID_FILE} --quit"
+}
+
+daemon_status ()
+{
+    su ${USER} -c "PATH=${PATH} ${PYTHON} ${PYLOAD} --pidfile=${PID_FILE} --status > /dev/null"
```
Dr-Bean: I think you want this:
```
su ${USER} -c "PATH=${PATH} ${PYTHON} ${PYLOAD} --pidfile=${PID_FILE} --status" > /dev/null
```
instead of this:
```
su ${USER} -c "PATH=${PATH} ${PYTHON} ${PYLOAD} --pidfile=${PID_FILE} --status > /dev/null"
```

kmarty: `/dev/null` is accessible for both, for user pyload as well as for user root.

Dr-Bean: Yep, its more aesthetics and keeping in line with the other commands.

kmarty: OK :-)

spk/pyload/src/installer.sh
Dr-Bean: I've been looking at cross-compiling `expect`, and it turns out it's a total pain in the neck. I've got it working, although I'll need to handle one last issue (the compiled library is put in `lib/subdir`, and the main executable doesn't look further than `lib`...so, probably needs another patch on top of the rest).
I'm leaning towards ditching the whole thing, and maybe see if [pexpect](http://pexpect.readthedocs.org/en/latest/) would make for a more sustainable solution. Python is already needed, so that fits nicely.

kmarty:
Hi,
Is it meant as "canceled at all" or "merged into spksrc branch"?

Diaoul:
This was closed by mistake automatically by GitHub when I renamed develop branch to master. I'm waiting for an answer from the support. I still want this to be merged.

Diaoul:
Maybe you can reopen against master branch?

PR #1170, opened 2014-08-05
Add Symlinks in the path for several packages + Permissions fix

tcavallari:
This pull request adds symlinks into /usr/local/bin as a post installation step for the following packages:
- git
- node
- ruby
- python

Additionally, it also fixes the permissions and the ownership for the installed files. 
Before the fix the files were assigned to the same uid of the user building the package on the build system which could be a problem since a user on the synology might have had the same uid and thus might have been able to tamper with the installed files. Now all the files are owned by root.

Other packages might be affected by the same problem.

### DIFF
--- a/spk/debian-chroot/src/installer.sh
+++ b/spk/debian-chroot/src/installer.sh
@@ -23,6 +23,10 @@ postinst ()
     # Link
     ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
 
+    # Set the permissions
+    chown -hR root:root ${SYNOPKG_PKGDEST}
+    chmod -R go-w ${SYNOPKG_PKGDEST}
+
     # Create a Python virtualenv
     ${VIRTUALENV} --system-site-packages ${INSTALL_DIR}/env > /dev/null
 
--- a/spk/git/src/installer.sh
+++ b/spk/git/src/installer.sh
@@ -21,6 +21,10 @@ postinst ()
     # Install busybox stuff
     ${INSTALL_DIR}/bin/busybox --install ${INSTALL_DIR}/bin
 
+    # Set the permissions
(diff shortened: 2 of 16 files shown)


Dr-Bean:
Since the permissions issues have been resolved, this would need a rebase. @Diaoul I assume you don't object to the general approach of having the symlinks added?

spk/debian-chroot/src/installer.sh
```diff
@@ -23,6 +23,10 @@ postinst ()
     # Link
     ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
 
+    # Set the permissions
+    chown -hR root:root ${SYNOPKG_PKGDEST}
+    chmod -R go-w ${SYNOPKG_PKGDEST}
```
Diaoul: Please use `${INSTALL_DIR}` here

spk/git/src/installer.sh
```diff
@@ -21,6 +21,18 @@ postinst ()
     # Install busybox stuff
     ${INSTALL_DIR}/bin/busybox --install ${INSTALL_DIR}/bin
 
+    # Set the permissions
+    chown -hR root:root ${SYNOPKG_PKGDEST}
+    chmod -R go-w ${SYNOPKG_PKGDEST}
```
Diaoul: Same here

spk/debian-chroot/src/installer.sh
Dr-Bean: I thought #1226 adds the necessary code to be able to remove this part of this PR, doesn't it?

Diaoul:
I don't like symlinks because it can conflicts with other packages during both install/remove steps and runtime. I think it's better to let the user edit it's PATH variable in host own .profile so he knows exactly where the binaries are from.

PR #1212, opened 2014-09-02
update sslh to v1.16

ffung:


cross/sslh/Makefile
```diff
@@ -1,42 +1,26 @@
 PKG_NAME = sslh
-PKG_VERS = 1.13
-SRC_VERSION = 1.13b
+PKG_VERS = v1.16
+SRC_VERSION = v1.16
```
Dr-Bean: Seems this is no longer needed, right? Otherwise, using `${PKG_VERS}` would be just as good.

cross/sslh/Makefile
```diff
@@ -1,42 +1,26 @@
...
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(SRC_VERSION).$(PKG_EXT)
 PKG_DIST_SITE = http://www.rutschle.net/tech
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 SRC_DIR= $(WORK_DIR)/$(PKG_NAME)-$(PKG_VERS)/
-VERSION="v1.13b"
+VERSION="v1.16"
```
Dr-Bean: Seems this is no longer needed, right? Otherwise, using `${PKG_VERS}` would be just as good.

spk/sslh/src/start-stop-status
```diff
@@ -46,17 +37,12 @@ stop_daemon ()
...
-    if [ -h $SYNO3APP/sslh ]
-    then
-		rm -f $SYNO3APP/sslh
-    fi
 }
 
 daemon_status ()
 {
-	[ -h $SYNO3APP/sslh ]
+    PID=`ps | grep $SSLH | grep -v grep |awk '{print $1}'`
```
Dr-Bean: Couldn't you assign a PID file with startup option `-P`, and get the status from that?

Dr-Bean:
Added some minor comments. Nice cleanup :)

PR #1221, opened 2014-09-03
add tmux

mbaraban:
Based on vim spk.  Tested on evansport.

Signed-off-by: Michael Barabanov <michael.barabanov@gmail.com>


mezzoatic:
This works great. I've compiled and installed tmux on my 415+ and it runs perfectly.

cross/tmux/Makefile
```diff
@@ -0,0 +1,23 @@
+PKG_NAME = tmux
+PKG_VERS = 1.9a
+PKG_VERS_CUT = 1.9a
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = http://cznic.dl.sourceforge.net/project/tmux/tmux/tmux-1.9
```
Dr-Bean: I think it'd make sense to use the direct download link, so something like `http://downloads.sourceforge.net/project/$(PKG_NAME)/$(PKG_NAME)/$(PKG_NAME)-1.9/$(PKG_DIST_NAME` might work best here.
That prevents issues with downloading when the mirror is down.

spk/tmux/Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+DEPENDS = cross/$(SPK_NAME)
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = tmux is a terminal multiplexer, it enables a number of terminals or windows to be accessed and controlled from a single terminal. tmux is intended to be a simple, modern, BSD-licensed alternative to programs such as GNU screen.
+RELOAD_UI = no
+DISPLAY_NAME = Tmux
+STARTABLE = no
+BETA = 1
+CHANGELOG = "1. Support UTF8 by default"
```
Dr-Bean: Just mention `Initial release` here, or remove the entry completely imo :)

Dr-Bean:
Sorry about the delay in reviewing. Added a couple comments, but other then that, looks good. If you could, please squash (and maybe rebase while you're at it), that'd be great.

mbaraban:
I've rebased and updated the patch to address your comments.

Dr-Bean:
Merged with bbdfe5194e70b7a94a5fed6d33389bf2de58092e, thanks for the addition :)

Packages will be available once I've compiled them all, so probably later tonight.

Dr-Bean:
...and done: the packages are published.

PR #1262, opened 2014-10-02
Update Flexget to 1.2.197

Dr-Bean:


avogabo:
its possible update flexget to last version please

PR #1274, opened 2014-10-11
Submission of ncdc

sirjeannot:
compilation works for arch 88f6281, working fine and tested on my ds213air.
doesn't compile for x86. didn't try compilation on other archs.
for now, compilation restricted to 88f6281.


Dr-Bean:
@sirjeannot Could you split off the `screen` SPK into it's own PR, e.g. use a separate branch for that. Makes it easier to review and merge. Thanks.

PR #1362, opened 2014-11-27
initial release of myreadings package

phil0u:
Hi,

This is my first attempt to bring MyReadings package to SynoCommunity/spksrc

MyReadings is a web interface for Calibre software, which allows you to read ebooks from tablets, PCs, Macs, smartphones and other media devices.

It's similar to the COPS package already present in spksrc. These files are mainly modifications from the original spksrc COPS files.

Thanks,

phil0u


Dr-Bean:
No need to open a separate issue :) Neither me or Diaoul have found the time to look at it earlier.

First thing I noticed (just a quick check): this PR does not include the wizard, so settings in config.php cannot be set.
Apart from that, I had some issues accessing my library via this package, so I need to check if it's the package, or my library causing that. Did you run into any problems with the compatibility tests in the admin interface?

phil0u:
Oh, sorry for that, i thought i had to, as i didn't have any answer , positive or negative ( i know you're quite busy with the Avoton architecture, Toolchains, etc.. :) )

Right now, i'm not familiar enough with the wizard to be able to automatically adjust the PHP open_basedir settings. You have to manually add the path for now. I think it could be easily automated.
Then, you have to manually configure MyReadings settings through the Admin Interface

No problem on my side with the compatibility tests in the admin interface , besides the fact that CBR can't be read because the PHP unrar extension can't be installed, at least right now. You have to be careful not to forget the final "/" in the Calibre Libray Path.
Other than that, it works fine

Thanks,

Phil0u

Dr-Bean:
If you copy the wizard from COPS, you should be almost there. It's basically the same thing, right?
The settings in there, along with the installer (https://github.com/SynoCommunity/spksrc/pull/1362/files#diff-587836e426071358a70a3e24e9d2a5e9R29) should be sufficient to correct the open_basedir stuff.
As for the PHP Rar extension, that's not included by Synology, so that's something to consider asking about. (I have no idea if we can add it ourselves, any clue?)

phil0u:
Yes, I'll work it out so that the PHP open_basedir is changed during the wizard.

As for the PHP RAR extension, i tried a few weeks ago, but didn't managed to get it to work, as they are many dependancies IIRC. But i can definitely ask Synology about it, and i will.

Thanks for your time.

Phil0u

Dr-Bean:
@phil0u I noticed the wizard wasn't added yet: just checking if you're still working on it?

phil0u:
@Dr-Bean Yes, i'm still working on it.  Unfortunately, i didn't have much time on my hands lately, but i do plan to sort this out. Thanks much.

PR #1366, opened 2014-11-30
Rules 'n' things

Dr-Bean:
Follow-up of #1276. Some fixes, some new functionality.

The `kernel_required` rule passes on information from a sub-make to the top level. A check file is created on each level if REQ_KERNEL is set, and is passed on from level to level, until it ends up at the top level.
The downside is the rules don't look great, but I'm not sure there's a way around that.

x64-related: the evansport arch is not included in Synology's x64 arch, probably because it's 32-bit (?).

### DIFF
--- a/mk/spksrc.cross-cc.mk
+++ b/mk/spksrc.cross-cc.mk
@@ -20,9 +20,24 @@ endif
 
 ifneq ($(REQ_KERNEL),)
   ifeq ($(ARCH),x64)
-    $(error x64 arch cannot be used when REQ_KERNEL is set )
+    @$(error x64 arch cannot be used when REQ_KERNEL is set )
   endif
 endif
+
+# Check if package supports ARCH
+ifneq ($(UNSUPPORTED_ARCH),)
+  ifeq ($(UNSUPPORTED_ARCH),$(ARCH))
+    @$(error Stop: ARCH $(ARCH) is not a supported architecture )
+  endif
+endif
+
+# Check minimum DSM requirements of package
+ifneq ($(REQUIRED_DSM),)
+  ifneq ($(REQUIRED_DSM),$(firstword $(sort $(TCVERSION) $(REQUIRED_DSM))))
+    @$(error Stop: Toolchain $(TCVERSION) is lower than required version in Makefile $(REQUIRED_DSM) )
+  endif
+endif
+
 
 #####
 
@@ -71,16 +86,7 @@ smart-clean:
 clean:
 	rm -fr work work-*
 
-# Compare optional Makefile REQUIRED_DSM to provided TCVERSION. If REQ_DSM is lower than TCVERSION, exit
-checkversion:
-ifneq ($(REQUIRED_DSM),)
-  ifneq ($(REQUIRED_DSM),$(firstword $(sort $(TCVERSION) $(REQUIRED_DSM))))
-	@$(MSG) "Stop: Toolchain $(TCVERSION) is lower than required version in Makefile $(REQUIRED_DSM) "
-	@exit 1
-  endif
-endif
-
-all: checkversion install
+all: install
 
 
 SUPPORTED_TCS = $(notdir $(wildcard ../../toolchains/syno-*))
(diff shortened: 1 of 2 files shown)


Dr-Bean:
Rebased PR#1366. Three things: 
- At the moment, this is only a rebase to see what remains
- I don't like the hardcoded arches in `spk.mk`, but I don't know that there's a way around it.
- You'd probably lose the parallel make support again

Todo, continuing from the other discussion:
- Define what the default rule should look like. We probably would use the new default for the repository too, so in effect, the following consequences goes for the repo too:
  - only arches with DSM5.1 toolchains
  - we drop 4.3 support (FAQ entry somewhere on that subject would be great)
  - drop support for 88f5281, powerpc, ppc853 and ppc824 (right?)
  - add x64 (unless REQ_KERNEL, then x64 is replaced by bromolow, cedarview, avoton)
- Check which rules we can/should remove/rename. Haven't thought about it.

Diaoul:
By drop 4.3 I mean just don't include the build in the default rule, keep the possibility in the repo and spksrc to provide them (if asked for example).

Dr-Bean:
Yeah, got that. Keep the toolchains, and e.g. the default `make arch-88f6281` should still work, as well as `make all-archs` should do what it says on the tin.

Maybe a `make all-archs-default` rule, with the above restrictions, or something?

Diaoul:
What about all-supported and all-legacy (= all-archs - all-supported)?

Dr-Bean:
See my edit, instead of `-default`, we use your `-supported`: I'm good with that. For SynoCommunity, I assume we use both those rules then? (until otherwise decided)

Diaoul:
Yes, that's for SynoCommunity and for clarity. I'd like developers to spend time patching some stuff to get everything working on all-supported but spending time to fix problems on legacy archs... not important. The idea is to keep the focus on *currently* supported NASes.

Dr-Bean:
Yep, I can agree with that. Same as what we did with Node, for instance.

Can we get reports on DSM versions using the repository though? It would be interesting to know if older versions have dropped significantly since last time (20% on 4.x, or what was it?). If at all possible, split out per model, because that would give an insight into what we're actually supporting right now.

Diaoul:
@piwi82: do you think you can provide that?

The server needs some love, currently we're pretty bad at providing statistics. We would need someone dedicated on that topic with devops knowledge (CM, Docker) and monitoring (Kibana).

I've started and I have a clear picture where I want to be but it's really time consuming. And of course the server is running fine so that's pretty low on my TODO list :)

piwi82:
I don't know any of the tools to do the stats (CM ?). It will take much more time to me to have it working than you or anyone with better skills than me.

You will have to summarize every statistic needed and check if the required data is available and easily accessible.
I doubt the tools you mentioned will cover all cases (logstash, kibana).

Diaoul:
CM stands for Configuration Management such as SaltStack, Puppet, Chef, and so on. I know it won't cover all cases but that seems to me it's the bare minimum for proper monitoring.

Dr-Bean:
I'd agree that server is running great lately, nothing wrong for it being low on the list :)

To be fair, I'd say that server provisioning, server monitoring vs. repository usage statistics can be considered separate subjects, and we can treat them as such. The same monitoring tooling might be involved at some point, imo, that depends on how/where usage statistics can be logged/accessed.

All my knowledge on sysops is outdated, so I'd have to start from scratch, and unfortunately, I can't dedicate enough time/get things going within a reasonable timeframe.

Dr-Bean:
Btw, while I'm thinking of that, @Diaoul, I assume it would be possible to store the model/serial/whatnot in the database when it hits the database for a particular package right? Or would it make more sense to get it from the web server logs, because that happens periodically (That was the previous method, right?)

Diaoul:
Yes the data is stored in a pgsql database: https://github.com/Diaoul/spkrepo/blob/master/spkrepo/views/nas.py#L147
Although it's not used anywhere to create statistics about repository usage. That should be possible with a simple SQL request. I'll give it a shot maybe tomorrow

PR #1369, opened 2014-12-02
Add libev, greenlet and gevent

alucryd:
I first intended to propose an Odoo spk as well since there is no sign of Synology upgrading OpenERP, but I can't get python to build because of sphinxbase and thus couldn't test it. On Arch Linux Python 2.x is python2, and Python 3.x is python, which I'm assuming has a lot to do with it failing. The sphinxbase cross Makefile invokes cython (cython2 here), and even when that is corrected, it fails later on for an obscure reason I haven't had time to investigate. I may request another pull at a later point for Odoo if I can get sphinxbase to build.

However, i can confirm greenlet and gevent do work, my manual Odoo installation on the NAS is happy with them, I now have proper multithreading.

FYI, the spksrc.python-wheel.mk doesn't play well on Arch Linux, our "rename" doesn't accept -f as an argument.


Diaoul:
@Dr-Bean: can we merge this?

Dr-Bean:
Better not, tbh. This PR builds the wheels as part of the Python SPK...I believe you're not a fan of that approach ;)
The reason I left it open is because it serves as a reminder for me to get going with the fxasync package, the exact same wheels are part of that, without affecting the Python SPK.

PR #1381, opened 2014-12-08
FIX: Debian chroot fails to start up slow scripts

HakShak:
This script starts all Debian chroot services defined via the DSM Debian chroot web interface. If an init.d script didn't return immediately, it would wait on the script, except it doesn't seem this was ever tested as the time package is never imported. When this script would wait on a slow init.d script, it would throw an exception and error out. I discovered this while trying to run ElasticSearch on startup.


mitom:
Is there a reason why this was simply closed? I believe I'm facing the same issue running flexget.

Diaoul:
Yes, automatic closing by GitHub due to branch renaming (develop to master).
Please reopen against master branch.

Issue #1284, opened 2014-10-21
ZNC missing the compile module


alexrengel:
A module in the ZNC app was not included that has the option of sending everyone that is connected to one ZNC account the same private message, it is called privmsg. It cannot be compiled because znc-buildmod, the compiler module, is not included either. Can privmsg be included in the next ZNC update?

Thank you in advance!

Dr-Bean:
ZNC 1.4-10 has just been published on the repository, and includes the requested module.

alexrengel:
Hey guys. I keep getting a "Port 8250 configured for this package is either used by another service or reserved for system use. Please disable or modify the conflicting service, or contact developer to modify the package configuration". I disabled all apps but still no luck. The port is not reserved on the router either.

Dr-Bean:
Check `/usr/syno/etc/services.d` for `znc.sc`. It's possible you ran into the bug with Synology's servicetool that it didn't remove the sc-file on uninstall of the previous package.

alexrengel:
Dr-Bean, I have the following files in /usr/syno/etc/services.d
netbkp_server.sc
synoha.sc
syno_syslog.sc
syno_dhcp.sc
syno_ntp.sc
versionbkp_server.sc

No znc.sc

Am I supposed to uninstall znc and install the newest version? I've been trying to updated the package. The old znc package was stopped before updating.

Dr-Bean:
Wait, wrong location. I meant to say `/usr/local/etc/services.d`

alexrengel:
oh yes... I see it.

[znc]
title="ZNC"
desc="ZNC"
port_forward="yes"
dst.ports="8250/tcp"

Should I delete the file, or just change the port?

Dr-Bean:
Hm. You say the previous ZNC package is still installed? Then the new package should automatically overwrite the existing file...not sure why it wouldn't...
Anyway, you can remove it.

alexrengel:
Nothing. I renamed it to znc-bu.sc and then tried updating once again, but I still get the same error message.

Dr-Bean:
That's not very surprising if you left the file in the same place: it'll still be read ;)
So change the port, or remove the file, either should work.

alexrengel:
Dr-Bean, thanks! I was able to update it.

PR #1397, opened 2014-12-17
Add 'privmsg' and 'palaver' modules

adamroe:
Added [privmsg](https://github.com/kylef/znc-contrib/blob/master/privmsg.cpp) and [palaver](https://github.com/cocodelabs/znc-palaver/blob/master/palaver.cpp). This will close bug #1284.


Dr-Bean:
Other than a changelog entry in `spk/Makefile`, should be good to go. Can you add that in?

adamroe:
Sorry about that, Dr-Bean -- added a changelog!

Dr-Bean:
Thanks :)

PR #1409, opened 2014-12-26
NzbDrone to Sonarr + dependencies updated

maxrogers:


spk/sonarr/Makefile
```diff
@@ -0,0 +1,60 @@
+SPK_NAME = nzbdrone
```
kayone: makes sense,
thanks, wasn't sure if we needed to maintain backwards compatibility.

Dr-Bean:
Perfect :)  Guess I'll be compiling some packages then.

Issue #1428, opened 2015-01-03
[REQUEST] mediainfo


r0g3rd:
Hi,

It would be nice to see mediainfo as previously managed by MissileHugger.com adopted as a separate package.

Cheers

Dr-Bean:
MediaInfo has just been published on the repository, thanks to @cytec.

PR #1443, opened 2015-01-09
added MediaInfo spk

cytec:
as requested in #1428


spk/mediainfo/Makefile
```diff
@@ -0,0 +1,35 @@
...
+SPK_VERS = 0.7.72
+SPK_REV = 1
+SPK_ICON = src/mediainfo.png
+
+DEPENDS  = cross/$(SPK_NAME)
+
+RELOAD_UI = yes
+STARTABLE = no
+
+MAINTAINER = cytec
```
Dr-Bean: @Diaoul Do we have a general approach on the maintainer content? `SynoCommunity/maintainer`, `SynoCommunity`, or is `maintainer` accepted too?
Keep in mind that Package Center probably adds a linebreak without hyphenation in the middle of `maintainer` with the first option.

spk/mediainfo/src/installer.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Package
+PACKAGE="mediainfo"
+DNAME="MediaInfo"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+SSS="/var/packages/${PACKAGE}/scripts/start-stop-status"
+PATH="${INSTALL_DIR}/bin:${PATH}"
+GROUP="users"
```
Dr-Bean: Leftover? :)

Dr-Bean:
LGTM. 
One issue remains, I think: Considering that there's already a package (Missilehugger) doing the same thing, you're probably using the same directories. If so, during installation of this package, you're overwriting those dirs, without uninstalling the MissileHugger package. That should be solved before this package can be published.

Diaoul:
I'm not sure we should do anything for conflicting repositories. It's up to the user to handle this.

Dr-Bean:
In this particular case the effects might not disastrous, but since we are aware that a conflicting package exists, and a potential user might not, imo, we should inform them/take care to not break things. E.g. check if the dir already exists and exit (which wouldn't be a bad thing to do regardless) or provide a warning during installation.

As an fyi, there's another repo, [synologyitalia](http://spk.synologyitalia.com/), which started hosting all the MH packages, although I do not know if they also maintain/update the packages.

spk/mediainfo/Makefile
Diaoul: For now we use SynoCommunity to avoid user confusion. There is a maintainer branch on which we put SynoCommunity/maintainer but it's too long so we'd need something else.

Let's put SynoCommunity there for now and we'll put the maintainer once I've implemented a developer page on spkrepo.

spk/mediainfo/src/installer.sh
cytec: oops... only things needed are PACKAGE, DNAME, INSTALL_DIR ... right?

spk/mediainfo/Makefile
cytec: ok i'll fix it then :)

cytec:
I don't know which path the MH package uses... but you may be right and those could conflict.
But i'm with @Diaoul on this one. There are several repos out there which also host `python` older/newer versions of `sabnzbd` etc. which might conflict with SynoCommunity.

i think the User is responsible for his own DS and what he installs from which repo.

PS: also i don't know if thats still a "problem" in 5.X in 4.3 the only thing that was checked was the Package name and version. So basically you would overwrite the whole spk if names are equal and versions differ.

example:
MissileHugger:

	Package Name: MediaInfo
	Version: 0.7.71

SynoCommunity:

	Package Name: MediaInfo
	Version: 0.7.72

in that case the MissileHugger Package would be completely overwritten by the SynoCommunity one because the version is higher.

Dr-Bean:
That's a good point there, if the SPK name is the same, the package is overwritten completely. Well then, guess it's not an issue after all ;)
@cytec If you can squash the commits, I'll merge it when I can.

cytec:
@Dr-Bean squashed :)

Dr-Bean:
Done and done. I'd publish the packages, but I can't yet :/

Dr-Bean:
@cytec By the way, did you compile all archs successfully? 88f5181-3.1 segfaults on libmediainfo here.
We should probably set a minimum firmware version of 4.x, or add 88f5281 as an incompatible arch in the cross package.

cytec:
@Dr-Bean i was able to compile all arch's but i only tested compileing with 4.x toolchains

Dr-Bean:
Ah right, I use `all-archs-dsms`/`all-archs-latest` to compile. 88f5181 isn't published, so it doesn't matter from that point of view, just curious if it was something on my side.

Pulper:
By the way, a new version is available 0.7.73, 2015-04-09: https://mediaarea.net/MediaInfo/ChangeLog

Aimhere2k:
Hi,

I installed MediaInfo from the Synology Package Manager, but I found installing it this way did not add it to my shell path when I'm connected via SSH/telnet. I had to edit my shell profile and manually append the location of MediaInfo.

Shouldn't the package installer take care of this? Or doesn't it work like that? (I'm coming at this as a Windows user, my knowledge of Linux is fairly limited.)

cytec:
@Aimhere2k nope the installer doesn't do anything with you PATH... just installs the binary to `/usr/local/mediainfo/bin` so you'll have to add that to your PATH manually ;)

PR #1444, opened 2015-01-09
added testdisk spk

### LINKED ISSUE #1407: [Request] TestDisk
seems like there is already a compiled binary for ds 211:
http://www.cgsecurity.org/wiki/TestDisk_Download

cytec:
fixes #1407


spk/testdisk/Makefile
```diff
@@ -0,0 +1,36 @@
...
+SPK_REV = 1
+SPK_ICON = src/testdisk.png
+
+DEPENDS  = cross/$(SPK_NAME)
+
+RELOAD_UI = yes
+STARTABLE = no
+BETA = 1
+
+MAINTAINER = cytec
```
Dr-Bean: SynoCommunity here :)

Dr-Bean:
Just one comment ;) Other than that, with a squash we should be good imo.

cytec:
@Dr-Bean fixed :)

Dr-Bean:
Noticed that the digests were incorrect. Probably the developer updated the file?

PR #1475, opened 2015-02-07
Update node to v0.12.0

wind-rider:
Patches were not needed anymore for compilation (at least on armadaxp)
Updated SPK_REV to 3

Please comment :)


Dr-Bean:
This has merge conflicts, could you please do a rebase? :)
Apart from that, I'll compile a couple arches, see if I can confirm the patches are no longer needed.

Dr-Bean:
I did a (very) quick check by compiling 88f6281, and it seems we can do two things:

1) Keep support for DSM4.3 TC's. That would mean keeping the patches (at least `fix-pthread.patch` with a rework, the other one might also be needed, didn't check specifically), and probably add a new patch or maybe a `CONFIGURE_ARGS` to successfully compile `v8` on 88f6281.

2) Remove support for DSM4.3 TC's, and only compile with DSM5.x TC's. Looks like no patches are needed for 88f6281. We can use a `REQUIRED_DSM = 5.0` flag in the Makefile to enforce that.

I'm leaning towards the second option right now. @Diaoul?

Diaoul:
Yep, let's drop 4.3.

wind-rider:
Thank you for your feedback, have a look at my new commit.

Dr-Bean:
Just a heads up: Even though node compiles successfully, I get a core dump when using `node` or `npm` after installation on an 88f6281 model.
As a result, this can't be merged, until that's cleared up. It might be caused by the removal of one of the patches, might be something else. Unfortunately, I'm not getting any debug/trace, so I can't tell.

wind-rider:
The package runs fine on my nas (armadaxp). I compiled with `make arch-armadaxp-5.1`

Since I don't have a model with that architecture, I can't test it myself.

@Dr-Bean, could you help me with testing?

Dr-Bean:
Well, did a bit of digging. As indicated [here]( https://code.google.com/p/v8/issues/detail?id=3285), v8 `3.18` is the last version with official support for ARMv5 (which is what 88f6281 is based on). 
Sidenote: NodeJS 10.33 was based on v8 `3.28.73`, and it still worked with that. Apparently, they hadn't removed all the supporting code at that time, I can't think of another reason.

Anyway, I did attempt to force an ARMv5 build, but v8 doesn't want to play along. I'm not going to attempt getting the code back in, so I guess we'll be dropping support for 88f6281.

Dr-Bean:
@wind-rider, could you please do another squash and rebase of your commits (then force-push that commit to your fork on GIthub)? If not, I'll probably end up redoing your commits myself ;)

If you want, you can put `REQUIRED_DSM = 5.0` in the cross Makefile, and remove `88f6281` from [this line](https://github.com/SynoCommunity/spksrc/blob/develop/cross/node/Makefile#L18), and add in the missing `x64` arch.

wind-rider:
@Dr-Bean Thank you for sorting out that crash!

I tried to do squash and rebase my commits, commit https://github.com/wind-rider/spksrc/commit/a238f0a19b77669e6e6e3a447a644d1a54842452 is the one you need.

I don't understand exactly why the last one was appended :S

wind-rider:
@Dr-Bean @Diaoul could you have a look at this?

Diaoul:
This cannot be merged as-is due to merge conflicts. Please rebase your changes. You can save your changes and start over a new fork with a new PR.

wind-rider:
@Diaoul I already rebased my changes after @Dr-Bean's comment.

Github says "This pull request can be automatically merged by project collaborators." - so I don't understand what is conflicted?

Diaoul:
The PR modifies files outside node, see the diff. That's problematic.

wind-rider:
@Diaoul @Dr-Bean I replaced this pull request by https://github.com/SynoCommunity/spksrc/pull/1529

PR #1480, opened 2015-02-08
Add OctoPrint

Decipher:
As per #960, which was inadvertently closed due to a bad rebase and other such things, this PR adds the OctoPrint package.


spk/octoprint/src/installer.sh
```diff
@@ -0,0 +1,91 @@
...
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # Create a Python virtualenv
+    ${VIRTUALENV} --system-site-packages ${INSTALL_DIR}/env > /dev/null
+
+    # Install the wheels/requirements
+    ${INSTALL_DIR}/env/bin/pip install --use-wheel --no-deps --no-index -U --force-reinstall -f ${INSTALL_DIR}/share/wheelhouse -r ${INSTALL_DIR}/share/wheelhouse/requirements.txt > /dev/null 2>&1
+    ${INSTALL_DIR}/env/bin/pip install -I -r ${INSTALL_DIR}/share/OctoPrint/requirements.txt > /dev/null 2>&1
```
Dr-Bean: Best to remove this line, the previous line should cover all the requirements, right?

Dr-Bean:
Added one comment, and I think that's all I can add ;) @Diaoul, I believe you're familiar with Octoprint?

Decipher:
My bad, likely leftover from past attempt, will remedy ASAP.

Diaoul:
Looks good to me. The config is stored in user home that is backed up at each upgrade so that should be OK for upgrading the SPK.

Dr-Bean:
One last thing: port 8088 is used as `ADMIN_PORT`, might want to add a service configure file then. The general approach is available in https://github.com/SynoCommunity/spksrc/pull/1000.

Decipher:
Something in the requested changes is preventing the service to being run, trying to narrow it down.

Decipher:
Looks like the First line of the wheel/requirements was the one to remove not the Second?
- Removed second line, wouldn't run package.
- Put the second line back, removed the first line, package runs.

Just finishing up now.

Decipher:
Found one last final bug, the USB device (/dev/ttyACM0) was only getting it's permissions set during install, but losing said permissions when Octoprint was stopped. Now permissions are set on 'start_daemon'.

But, tested, and I am now able to run my Printrbot JR v2 from my DS413 :)

Dr-Bean:
Well, I did mean keep the first line, remove the second :)
Apart from the other flags, one important flag is missing: `-f ${INSTALL_DIR}/share/wheelhouse`. That flag ensures that the wheels are installed from the `wheelhouse` directory, instead of being downloaded from PyPI when the installer runs.
On top of that, you created the wheels via the SPK Makefile, but you're not actually using them now ;)

To debug: Pipe the output of the command to a log file during install, or you can even run it directly in a clean venv.

Decipher:
Yeah, my bad (again ;), still a lot to learn in the SynoCommunity). The issue was that I was still trying to install PyYAML from the wheelhouse, which was the original hold up, and is no longer required.

There was also a few additional requirements that weren't defined, but where obviously getting picked up as dependencies, so that's resolved as well.

Last but not least, updated default config to reflect the correct serial device path, so on a fresh install you can connect straight away. This should hopefully be all there is to it.

Dr-Bean:
Great :) Can't test the functionality myself, but the code looks good to me. If you can squash the commits, we should be able to merge/publish.

Decipher:
I can squash the commit, but I don't think I can push the squashed commit into the same branch and/or PR, at least I've not done before and don't know how to. Thoughts?

Decipher:
Figured it out. Time for bed.

BrianGilbert:
Looking forward to having access to this, is there any idea of time frame until it will be generally available in SynoCommunity?

Diaoul:
I just noticed you miss the config file for the app to be visible in DSM menu. Can you add this in a new PR @Decipher?
See other packages for syntax, such as sickbeard-custom.

Dr-Bean:
Octoprint 20150315-1 has been published on the repository.

rmanalan:
I just tried to install this on my DS213j (which runs on the armada 370). The problem is that this packaged doesn't even show up in my package manager. Is there a reason for [this line](https://github.com/SynoCommunity/spksrc/blob/master/spk/octoprint/Makefile#L31) in the Makefile? I believe this is why my NAS can't see this package.

Decipher:
@rmanalan That line is required to so that SPKSRC doesn't try to cross compile the package as it is a python package. It is not the reason your NAS can't see the package. I suspect the reason you can't see the package is that the package is a Beta package and you need to tell the Synology Package Center to allow Beta packages.

PR #1481, opened 2015-02-08
Update btsync to 1.4.110

hurik:
**Change log:**
- Fixed issue with Sync hanging due to invalid filename characters
- Fixed issue with Sync preventing HDD from sleeping on NAS (*)
- Fixed some incorrect localization text
- Fixed issue with confusing old and new filenames in history when renaming
- Added new advanced parameter for Linux - config_save_interval
- Minor fixes and changes
- Remove asterisks in return values of “get_prefs” API call
- Fixed sporadic Out-Of-Sync issues (**)

(*) To allow NAS sleeping when Sync is running set folder_rescan_interval, config_refresh_interval and config_save_interval to some significant value, for example - 3600 (seconds)

(**) The "Out-of-Sync" issue has a number of root causes and we know that not all of them are addressed at the moment. Link to the topic dedicated to fixing the issue is below.


hurik:
* Fixed issue with Sync preventing HDD from sleeping on NAS (*)

Isn't working :( tested it with `"folder_rescan_interval" : 86400,` in `sync.conf` ....

Dr-Bean:
Thanks :)

Genesius:
hello,

I am unable to update, my synology package center tell me that it's always the 1.4.103-9 in the repository.

Dr-Bean:
Merging PR's doesn't necessarily mean new packages are available immediately.
Check again, packages should show up now.

Genesius:
thank you for the information.

Now the update is available.

PR #1488, opened 2015-02-14
Add OpenZWave support for Domoticz

Decipher:
As per #1478 and about a week of hard work.


Decipher:
I think this is pretty much as done as can be.

bru7us:
I just want to say, I'm totally stoked to see someone getting value out of the kernel work that I did!! It was a hard slog to get LIRC running - and I'm very happy that there is more value to be had.

Decipher:
@bru7us Thanks heaps for the work, I can't even imagine doing it myself. This entire process has just been guess work, before this I'd never really dealt with cross-compilation (or regular compilation) in any serious way. The Kernel stuff was the cherry on top of how good spksrc is.

cross/cp210x/Makefile
```diff
@@ -0,0 +1,35 @@
...
+.PHONY: myCompile
+myCompile:
+	cp $(KERNEL_DIR)/drivers/usb/serial/cp210x.c $(WORK_DIR)/$(PKG_DIR)/cp210x.c
+	CPPFLAGS="" CFLAGS="" LDFLAGS="" \
+	make -C $(KERNEL_DIR) SUBDIRS=$(WORK_DIR)/$(PKG_DIR) modules
+
+.PHONY: myInstall
+myInstall:
+	mkdir -p $(STAGING_INSTALL_PREFIX)/modules
+	install -m 755 $(WORK_DIR)/$(PKG_DIR)/$(PKG_NAME).ko $(STAGING_INSTALL_PREFIX)/modules/
```
Dr-Bean: How about this? (easier for me to just drop the whole file in here in this case)
```
PKG_NAME = cp210x
PKG_DIR = $(PKG_NAME)

DOWNLOAD_TARGET = kernel #any value but `nop` for now, needs a change in spksrc.download.mk

REQ_KERNEL = 1

HOMEPAGE = http://sourceforge.net/projects/dsgpl/
COMMENT  = CP210x USB to UART Bridge VCP Drivers
LICENSE  = GNU GPL

CONFIGURE_TARGET = nope
COMPILE_TARGET = myCompile
INSTALL_TARGET = myInstall

.PHONY: myCompile
myCompile:
	@mkdir -p $(WORK_DIR)/$(PKG_DIR)
	@echo "obj-m += cp210x.o" > $(WORK_DIR)/$(PKG_DIR)/Makefile
	@cp $(KERNEL_DIR)/drivers/usb/serial/cp210x.c $(WORK_DIR)/$(PKG_DIR)/
	$(MAKE) -C $(KERNEL_DIR) SUBDIRS=$(WORK_DIR)/$(PKG_DIR) modules

.PHONY: myInstall
myInstall:
	mkdir -p $(STAGING_INSTALL_PREFIX)/modules
	install -m 755 $(WORK_DIR)/$(PKG_DIR)/$(PKG_NAME).ko $(STAGING_INSTALL_PREFIX)/modules/
```

cross/openzwave/Makefile
```diff
@@ -0,0 +1,30 @@
+PKG_NAME = openzwave
+PKG_VERS = $(PKG_SVN_REV)
+PKG_EXT = tar.gz
+PKG_DOWNLOAD_METHOD = svn
+PKG_DOWNLOAD_VCS_CLEAN = nope
```
Dr-Bean: Couldn't you just download the snapshots directly? http://old.openzwave.com/snapshots/.
Btw, the snapshots are deemed unstable. If possible, stick to the stable releases: http://old.openzwave.com/downloads/

cross/openzwave/digests
```diff
@@ -0,0 +1,3 @@
+openzwave-r1009.tar.gz SHA1 6acc7bd5878b0e68df9bfa139c004c035b2ccf73
```
Dr-Bean: With an svn checkout, this probably won't work for anything else but the file you originally created the digests with ;)

spk/domoticz/Makefile
```diff
@@ -12,7 +12,7 @@ DESCRIPTION_FRE = Domoticz est un système domotique qui vous permet de surveill
 ADMIN_PORT = 8084
 RELOAD_UI = yes
 DISPLAY_NAME = Domoticz
-CHANGELOG = "update to svn revision 2112"
+CHANGELOG = "update to svn revision 2276"
```
Dr-Bean: ...and add zwave support (or something like that)

Dr-Bean:
Left a couple comments. Considering the changes, maybe put the package back in beta?

spk/domoticz/src/dsm-control.sh
```diff
@@ -16,6 +16,14 @@ PORT="8084"
 
 start_daemon ()
 {
+    insmod /lib/modules/usbserial.ko
+    insmod /lib/modules/ftdi_sio.ko
+    insmod ${INSTALL_DIR}/modules/cp210x.ko
```
Dr-Bean: Not sure, but shouldn't there be a corresponding `rmmod` when stopping/removing the package?

Decipher: I originally had that, but it occurred to me, especially with my other pull request for OctoPrint, that removing USB drivers because I'm done with them isn't really very polite.

spk/domoticz/Makefile
Decipher: Yeah, thought I maybe should have that, but technically Domoticz had OpenZWave support, so we haven't added it, we've enabled it. But happy to append that to the message.

cross/openzwave/digests
Decipher: Will remove.

cross/openzwave/Makefile
Decipher: I was mostly following the guidelines of the tutorial on Domoticz.com about how to get OpenZWave working with with Domoticz on Syno, which specifically used the SVN version. But I'm happy to test with the stable version to see if it still works.

cross/cp210x/Makefile
Decipher: Will give it a spin. Again, was mostly just following guides I could find online, flying blind half the time here :) But that looks like it will be much simpler.

Decipher:
Regarding the patches, and essentially the whole thing, it was mostly based on http://www.domoticz.com/wiki/DomoticzSynology, which supplied the patches. I assume they had reasons for most of what the patch does, but I agree that removing comments is unnecessary.

I'll try to action all the comments you've made ASAP, but that will most likely not be for at least a day or two.

Thanks for the review.

spk/domoticz/src/dsm-control.sh
Dr-Bean: You might be right. My concern was mainly if this would have the potential of leaving stuff (specifically, a loaded cp210x.ko module) behind after removing the package. I really don't know the answer to that.

spk/domoticz/Makefile
Dr-Bean: Good thing that it's our package changelog, not a Domoticz changelog ;)

cross/openzwave/digests
Dr-Bean: Fwiw, if you go with downloading the snapshot.tar.gz file, you should use the digests again ofc, it's just the svn checkout method which doesn't work with a digests file.

spk/domoticz/src/dsm-control.sh
bru7us: There's a lot of comments over in the HDHomeRun/TVHeadEnd threads about this kind of issue.
I know with LIRC I did a full rmmod of anything we inserted, but that is not necessarily the best idea, as already mentioned above (shared usb modules etc).

The whole "package" infrastructure makes it a bit tough because ideally things should be completely independent and removing everything from a package installation shouldn't impact anything else on the system - but we have well and truly closed that line! Ideally rather than directly loading and unloading, we could have a system in place where we could package-specific module requirements into config files (e.g. /var/packages/spksrc_drivers/modules.d/...) with a separate package that handles loads/unloads and dependencies as required; however we don't have such a process in place, and that would still only help with our (spksrc) packages and there are many other package providers who wouldn't comply.... can't win!

Best practice here would probably be to unload package-specific modules (cp210x) on exit and leave potentially shared modules loaded. Agree?

cross/cp210x/Makefile
Decipher: With one minor variation this worked fine. The variation being the MAKE arguments, from what I can tell they need to be as they where (which was taken from LIRC I believe) instead of the flags provided by $(MAKE). At least, while it compiled directly with the provided code, it failed to compile when built as a dependency of Domoticz until I re-adjusted the command.

cross/openzwave/Makefile
Decipher: Done

cross/openzwave/digests
Decipher: Done (digests from snapshot)

cross/openzwave/Makefile
Decipher: Worth noting that with this approach the modification to spksrc.download.mk is no longer necessary. Should it be removed?

Decipher:
All changes made (except to put it into Beta), bumped to r2287 and running great.

Re-rolled the OpenZWave patch and improved on it, drastically reducing the size. Only re-rolled the Domoticz patch without any modifications.

Hopefully this is good to go now.

Decipher:
Can I ask why this was closed? I assume (hopefully) it is related to the fact that the 'develop' branch which this was made against no longer exists, presumably due to some sort of cleanup?

Diaoul:
Yes that's the reason. I contacted the support to reopen those PR against the master but I don't know if that's possible.

Decipher:
No stress, should be able to do another PR this afternoon if they're unable to.

PR #1491, opened 2015-02-16
Added "force update" flag to init script, useful for systems that do not stay on 24/7 (sickbeard updates show info at 3AM)

stevanov:
Systems that do not stay online overnight (3AM specifically) will experience a problem where sickbeard will not automatically download new episodes after a while because it is unaware of the existence of new episodes. Adding the --forceupdate flag to the initialization script fixes this issue by making sickbeard update its tv shows on start instead of the scheduled 3AM. Sickbeard does not allow the user to change the time when this happens in the settings (3AM is hardcoded) so this seems like an easy fix for anybody experiencing these issues.

### DIFF
--- a/spk/sickbeard/src/dsm-control.sh
+++ b/spk/sickbeard/src/dsm-control.sh
@@ -18,7 +18,7 @@ LOG_FILE="${INSTALL_DIR}/var/Logs/sickbeard.log"
 
 start_daemon ()
 {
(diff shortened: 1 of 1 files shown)


Dr-Bean:
It would make sense to implement a change like this upstream: e.g. by making SB to automatically do a force-update when it starts (could be an on/off option in the config), or make the scheduled time adjustable from the config. Imo, either would be preferred over adding this to the start script.

PR #1498, opened 2015-02-18
Added sudo pkg

cybfr:
As /root is not permanent against dsm updates, and busybox is not suid root, sudo is very usefull !
NB you need to do
$ sudo make arch-<arch>
or 
$ make arch-<arch>
$ sudo make arch-<arch>
to achieve package building.

Thanks


cross/sudo/Makefile
```diff
@@ -0,0 +1,25 @@
...
+PKG_VERS = 1.8.12
+PKG_VERS_CUT = $(PKG_VERS)
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = http://www.sudo.ws/sudo/dist
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS_CUT)
+
+DEPENDS = 
+
+HOMEPAGE = http://www.vim.org/
```
bru7us: Clean up references to Vim

spk/sudo/Makefile
```diff
@@ -0,0 +1,29 @@
+SPK_NAME = sudo
+SPK_VERS = 7.4
```
bru7us: Should this be SPK version 1.0 as it's never ben released yet? Also revision 1.

spk/sudo/Makefile
```diff
@@ -0,0 +1,29 @@
...
+HOMEPAGE   = http://www.sudo.ws
+LICENSE    = ISC-style license (TODO: check the compliance of the licence)
+
+INSTALLER_SCRIPT = src/installer.sh
+SSS_SCRIPT       = src/dsm-control.sh
+ADDITIONAL_SCRIPTS = 
+
+INSTALL_PREFIX = /usr/local/$(SPK_NAME)
+
+POST_STRIP_TARGET = sudo_extra_install
```
bru7us: More cleanup of all the references to sudo_extra_install that doesn't do anything

bru7us:
Also, don't include the tar.gz file, this gets pulled from the DIST_SITE at make time

cybfr:
Hi,
Thank you for the review and sory for the dust... I think things are cleaner now.

Questions :
- The package does not create /etc/sudoers (since it will be erased on dsm upgrades, I though it's better to let the user manage it). But is there a correct way to manage those conf file ?
- It does not create a symlink in PATH to visudo cause I don't know where to put such a link for sbin execs. Is there a policy in that too ?

Cheers.

cross/sudo/Makefile
```diff
@@ -0,0 +1,25 @@
+PKG_NAME = sudo
+PKG_VERS = 1.8.12
+PKG_VERS_CUT = $(PKG_VERS)
```
Dr-Bean: Looks like you can just use `$(PKG_VERS)` for `PKG_DIR`, right?

cross/sudo/Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+DEPENDS = 
+
+HOMEPAGE = http://www.sudo.ws
+COMMENT  = Sudo allows a system administrator to delegate authority to give certain users (or groups of users) the ability to run some (or all) commands as root or another user while providing an audit trail of the commands and their arguments.
+LICENSE  = Charityware
+
+GNU_CONFIGURE = 1
+CONFIGURE_ARGS = ""
+CONFIGURE_ARGS += ""
```
Dr-Bean: Remove L16 and L17, no need to specify empty args

cross/sudo/PLIST
```diff
@@ -0,0 +1,36 @@
+bin:bin/sudo
+bin:bin/sudoedit
+bin:bin/sudoreplay
+bin:include/sudo_plugin.h
```
Dr-Bean: I _think_ the header files can be removed.

cross/sudo/PLIST
```diff
@@ -0,0 +1,36 @@
...
+bin:libexec/sudo/group_file.la
```
Dr-Bean: libexec: use `lib:` instead of `bin:`

cross/sudo/PLIST
```diff
@@ -0,0 +1,36 @@
...
+bin:libexec/sudo/libsudo_util.so.0
+bin:libexec/sudo/libsudo_util.so.0.0.0
+bin:libexec/sudo/sudoers.la
+bin:libexec/sudo/sudoers.so
+bin:libexec/sudo/sudo_noexec.la
+bin:libexec/sudo/sudo_noexec.so
+bin:libexec/sudo/system_group.la
+bin:libexec/sudo/system_group.so
+bin:sbin/visudo
+shr:share/doc/sudo/ChangeLog
```
Dr-Bean: Remove `man`, `doc` and maybe `examples` too (if the examples are needed to get the package working, they might fit better in a `/var` location)

spk/sudo/Makefile
```diff
@@ -0,0 +1,25 @@
+SPK_NAME = sudo
+SPK_VERS = 1.0
```
Dr-Bean: Should be the same value as `cross/sudo`.
`SPK_VERS` is used for the version of the software you're packaging. The package version is `SPK_REV`, and that should be `1` (and go up with every new package release)

spk/sudo/Makefile
```diff
@@ -0,0 +1,25 @@
...
+SPK_REV = 1
+SPK_ICON = src/sudo.png
+
+DEPENDS = cross/$(SPK_NAME)
+
+MAINTAINER = frv
```
Dr-Bean: SynoCommunity

spk/sudo/Makefile
```diff
@@ -0,0 +1,25 @@
...
+DESCRIPTION = "Sudo allows a system administrator to delegate authority to give certain users (or groups of users) the ability to run some (or all) commands as root or another user while providing an audit trail of the commands and their arguments."
+RELOAD_UI = no
```
Dr-Bean: Minor: I don't think you need to specify this if you don't need the reload

spk/sudo/Makefile
```diff
@@ -0,0 +1,25 @@
...
+DISPLAY_NAME = sudo
+STARTABLE = no
+BETA = 1
+
+HOMEPAGE   = http://www.sudo.ws
+LICENSE    = ISC-style license (TODO: check the compliance of the licence)
```
Dr-Bean: Where does this come from? If nothing else, put a link to the official license in here: (seems to be http://www.sudo.ws/sudo/license.html), because that's what we're referring to here.

Dr-Bean:
Added another couple of comments. I don't have an answer to your questions right now, I'll need to see the package in action for that, I think. Maybe someone else does though :)

bru7us:
In response to your question about sudoers:
Keep the file in your package directory and symlink it to /etc on installation.

It is possible to detect DSM upgrades and re-link, see what I did with LIRC here: https://github.com/SynoCommunity/spksrc/blob/develop/spk/lirc/src/dsm-control.sh#L126
(Each startup does a quick version check to a cakes version captured on last install/upgrade)

It is a shame about PATH - but @Diaoul likes us to not mess with any part of of the system that isn't cleaned up on uninstall. Again you can see some functions that I added in the LIRC installer to update PATH - but they never get called from the package. This lets any knowledgable users update their own path if they wish, with a simple: `. /var/packages/lirc/scripts/installer && fixpath`

cross/sudo/Makefile
cybfr: Just wanted to stick to the model but yes PKG_DIR = $(PKG_NAME)-$(PKG_VERS) is shorter.

cross/sudo/Makefile
cybfr: Same reason as above, done.

cross/sudo/PLIST
cybfr: Done, I did'nt find an extensive description of PLIST  does someone have a pointer ?

cross/sudo/PLIST
cybfr: Done,
We don't need any of them to run the package, but we may need them to understand how to run it (providing we have a man package)

spk/sudo/Makefile
cybfr: Thanks, this was really not clear forme,
Done

spk/sudo/Makefile
cybfr: Honestly, I've no clue on what this is doing and  'have not spent much time to find it. If anyone can give me an explanation on this field ... 
By the way, don't specify it  seems hurtless.

spk/sudo/Makefile
cybfr: At a first glance, I've just put the text of the debian package license term. I replace it with the link to exact license.
Done.

PR #1523, opened 2015-03-04
Update btsync to 1.4.111

hurik:
Change log:
- Possible Out-Of-Sync issue fix (*)
- Minor core fixes

(*) The "Out-of-Sync" can only be fully fixed in 2.0 thanks to core changes introduced in 2.0. If 1.4.111 does not resolve your out-of-sync, upgrade to 2.0 is advised.


hurik:
I added `  "display_new_version" : false,` to `syno.conf` so that there would be now anoying update message ...

Issue #1378, opened 2014-12-07
Security - local authentication broken


bobertato:
Hi,

It's great that GateOne has made it to the Synology DiskStation but perhaps the following security issue should be looked into:

Default installation of GateOne allows anyone to launch sessions. This isn't very secure. It would be better to set authentication as Synology DiskStation local accounts or Admin only to prevent the service being used as a jumping point for malicious third parties who discover the service using network scanners.

Also the ability to enable local "pam" authentication as detailed in official GateOne documentation is broken on the Synology GatePne package (but not in the original GateOne package when installed on other systems) preventing users from securing GateOne on Synology DiskStations using the GateOne configuration files (such as /usr/local/gateone/var/server.conf  [different location on other systems]).

Many thanks (and the next release is much anticipated).

patsissons:
The issue is that the debian package `python-pam` is not available (https://packages.debian.org/stable/python/python-pam). 

Unfortunately the GateOne app does not gracefully report the missing dependency so you're left to guess. You can confirm it is missing by running `/volume1/@appstore/gateone/env/bin/python -c "import PAM"`. Note that this `python-pam` package is not to be confused with the pip package `python-pam` (https://pypi.python.org/pypi/python-pam/1.8.2). I don't know enough about GateOne or this pip package to understand why the pip package was not used instead (I assume there is a good reason). The debian package includes a compiled assembly with python bindings, whereas the pip package is pure python.

I'm trying to research how much effort is required to compile the debian package, if I make any progress I'll try and report it here.

Dr-Bean:
GateOne 20150311-5 has been published on the repository. #1526 adds PAM auth to a certain extent, although I'd suggest to take a look at the PR for the details.

PR #1526, opened 2015-03-06
Gateone

patsissons:
Major update of GateOne to 1.2.0-20150205214929 (master, da393e7)

This update fixes some features that were broken in the old 1.1 version (such as google auth and pam auth) as well as brings a new configuration framework to provide much better configuration options (such as user access limiting). Along with numerous bug fixes, this 1.2 release is all around much more pleasant to work with.

Note that there is no official stable 1.2 release, as GateOne has moved to a continuous integration process for versions 1.2 and up. This version is, however, quite stable and has dealt with a large number of issues reported since November 2012 (when 1.1 was released).

This PR also addresses to some degree issue #1378. While I haven't provided a sample configuration to use PAM auth (which isn't really possible unless you want to run GateOne as root), you can quite easily implement htpasswd file basic auth using PAM as described here: https://wiki.archlinux.org/index.php/Gateone#PAM

I additionally included some sample scaffolding configuration to easily support google authentication (which is a much better alternative than htpasswd file basic auth). 

I don't have a fantastic test environment for the changes I have made, but I have been able to do reasonable testing of the spk build and installation. 

I have run into a few non-critical issues with install. my knowledge of spk installs is not great, so I'm not sure what the best course of action is. Some of the synopkg.log output includes the following:
```
2015/03/06 14:04:34     install gateone 1.2-20150205214929: run postinst  ret=[0]
2015/03/06 14:04:34     gateone stderr: file gateone.py (for module gateone) not found
warning: no previously-included files found matching '.gitignore'
warning: no previously-included files matching '*.kate-swp' found anywhere in distribution
warning: no previously-included files matching '*.pyc' found anywhere in distribution
warning: no previously-included files matching '*.pye' found anywhere in distribution
warning: no previously-included files matching '*.git' found anywhere in distribution
warning: no previously-included files matching '.directory' found anywhere in distribution
file gateone.py (for module gateone) not found
file gateone.py (for module gateone) not found
```
an issue with the tornado.speedups extension not being able to compile (this could be related to my build environment). And generally just extra (what i assume is) stderr output from either the `setup.py install ... requirements.pybundle` or the `setup.py install` commands.

None of these issues appear to impact the installation or operation of GateOne. I am open to trying to fix the issues (if they warrant fixing) if anyone has ideas on where to look. I really don't know why something is looking for gateone.py (this file doesn't exist in the code base), this could be a legacy code issue. The warnings may also be related to my build environment.


spk/gateone/src/installer.sh
```diff
@@ -35,13 +36,16 @@ postinst ()
     adduser -h ${INSTALL_DIR}/var -g "${DNAME} User" -G ${GROUP} -s /bin/sh -S -D ${USER}
 
     # Create a Python virtualenv
-    ${VIRTUALENV} --system-site-packages ${INSTALL_DIR}/env > /dev/null
+    ${VIRTUALENV} --system-site-packages ${APP_ENV} > /dev/null
 
     # Install the bundle
-    ${INSTALL_DIR}/env/bin/pip install --no-index -U ${INSTALL_DIR}/share/requirements.pybundle > /dev/null
+    ${APP_ENV}/bin/pip install --no-index -U ${INSTALL_DIR}/share/requirements.pybundle > /dev/null
```
Diaoul: Please stick with the default in spksrc which is `${INSTALL_DIR}/env`. I agree your way is better but it's better to stick to one convention, it'll make mass maintaining of the code easier.

spk/gateone/src/installer.sh
```diff
@@ -35,13 +36,16 @@ postinst ()
...
 
     # Install GateOne
-    ${PYTHON} ${INSTALL_DIR}/share/GateOne/setup.py install --prefix=${INSTALL_DIR} > /dev/null
+    PYTHONPATH="${APP_ENV}/lib/python2.7/site-packages/:${PYTHONPATH}" ${PYTHON} ${INSTALL_DIR}/share/GateOne/setup.py install --prefix=${APP_ENV} --skip_init_scripts > /dev/null
```
Diaoul: Changing PYTHONPATH isn't necessary here

spk/gateone/src/installer.sh
```diff
@@ -35,13 +36,16 @@ postinst ()
...
+
+    # install initial certificats
+    cp /usr/syno/etc/ssl/ssl.crt/server.crt /usr/syno/etc/ssl/ssl.key/server.key ${INSTALL_DIR}/ssl/
```
Diaoul: What about linking instead?

Diaoul:
Thanks for the PR. I assume this is not backward compatible so we need to make sure uninstall is required before installing. It was done in previous versions of nzbget so you can look how to do that by looking in the history of spksrc on nzbget.

The pybundle must be replaced by wheels, @Dr-Bean can you help with that?

spk/gateone/src/installer.sh
patsissons: No problem, I was doing that for convenience for myself, but I completely understand keeping things consistent.

spk/gateone/src/installer.sh
patsissons: this was necessary in my build env, but i'll rip that change out of the PR and put it in my own build env fixes branch

spk/gateone/src/installer.sh
patsissons: I'm fine with linking, but I was mostly just following the previous install (which, still, copies the cert and key over when the package is started) see https://github.com/patsissons/spksrc/blob/gateone/spk/gateone/src/dsm-control.sh#L21

patsissons:
Yup, definitely not backwards compatible. I would like to hang out to the 1.1's `server.conf` file just because it may have some useful information (that would need to be manually transferred over). Not sure if that is possible with an uninstall?

I'll take a look at nzbget and try and make the rest of the above adjustments either tomorrow or monday.

spk/gateone/src/installer.sh
Diaoul: Well you already pick python binary in the venv so the libraries in the venv's site-packages is included already.

spk/gateone/src/installer.sh
Diaoul: Maybe due to permissions? Idk the reason why it was a cp but ln should be prefered here unless I'm missing some consideration.
Do you agree?

patsissons: yea i realized the issue is due to the cert and key having 400 as the perms. So it must be copied over to the gateone install dir so that we can `chown` the files to gateone.

spk/gateone/src/installer.sh
patsissons: yea i wasn't sure why it wasn't picking it up. I'll try another test build tomorrow and see if it's working without the override. If not, maybe it works properly in a more _normal_ build env?

patsissons:
The GateOne dev accepted my PR to fix the bug that required the patch, so i'll try and push an update sometime today to remove that patch and use today's master.

spk/gateone/src/installer.sh
Dr-Bean: Similar to https://github.com/SynoCommunity/spksrc/blob/develop/spk/syncserver/src/installer.sh#L78 maybe? I thought GateOne relied on a .pem file, or can it handle .crt and .key files too?

Dr-Bean:
I'll take a look at converting the pybundle to wheels...should already have the code somewhere, seeing as I was working it in the past, just can't find it right now.

spk/gateone/src/installer.sh
patsissons: That's a good question, i know the docs use pem's in the samples provided, but my installation seems to have no problem verifying against the crt/key files.

Dr-Bean: Maybe the documentation isn't up to date, but it's stated here: http://liftoff.github.io/GateOne/About/#certificate

patsissons: yes that documentation is for 1.1, it is not up to date.

patsissons: OK, that didn't take long. rather than check his code, I wanted to get a solid understanding of the differences between a pem file and a crt/key file. Turns out they are the same, just different extensions.

reference: http://serverfault.com/a/9717

Diaoul:
@Dr-Bean: OctoPrint SPK use wheels, maybe that helps?

Dr-Bean:
@Diaoul Nah, it's not that, converting the pybundles to wheels is a piece of cake ;) I just can't recall if `WHEELS = tornado` would do it, or whether it needed patches and/or other modules alongside. Easy enough to redo it though.

Dr-Bean:
Did a quick check, we'd need `WHEELS = certifi backports.ssl-match-hostname tornado` which results in this:
```
NAS> ./bin/pip freeze
backports.ssl-match-hostname==3.4.0.2
certifi==14.05.14
tornado==4.1
```
Optionally, we could set up a cross package for tornado to enable speedups. I'll take a proper look tomorrow or so. `configobj` is currently also part of the pybundle, not sure if it's still needed or not.

Diaoul:
@Dr-Bean: I'm working on updating python to 2.7.9, I'm updating all modules too and adding uWSGI. I can take a look and add tornado too while I'm at it.

It's compiling at the moment (without tornado) and will almost certainly require a few fixes before compiling without errors.

Dr-Bean:
Sure, although we can include Tornado as a cross wheel package if you prefer that, either works. It just bumps the compile time (and disk space...) with a cross wheel package.

Btw, if I'm not mistaken, 2.7.9 includes pip now, so we probably should hurry up with removing the pybundle stuff from the remaining packages, otherwise those can't be installed anymore.

Diaoul:
I pushed my changes but for some reason Pillow does not compile. Does it compile for you @Dr-Bean? I tried to use the old version (2.6.2) but same error. Maybe that's due to my environment, I'm running spksrc in Docker.

Dr-Bean:
Started compiling Python with 88f6281-5.1, I'll report back in an hour or so.

Dr-Bean:
@Diaoul Pillow compiles fine here, so seems it's something on your end.

Diaoul:
Yeah I might have missed that one, might need some fixes like that before being able to publish the new version of python to the repo.

I tried to update libgpg-error but we are missing some information (config file for each architecture) so we're stuck with this version for now. I asked Synology for the config files as I don't have all the NAS architectures at home :)

Dr-Bean:
Fyi, #1555 covers the conversion of the GateOne pybundle (as well as the other remaining bundles) to wheels. I simply included Tornado as a wheel without speedups: the pybundle didn't handle the speedups either. 

As mentioned before, we can either use a cross/wheel approach (see cross/scrypt for example), and/or include it as a module in the Python SPK, to be decided.

Diaoul:
I'd rather include tornado with speedups in the Python SPK and rely on that. Otherwise we'd need to create a multi-arch SPK instead of a noarch one.

Dr-Bean:
To be clear, #1555 currently provides the exact same thing as the pybundle, just in wheel format.
But if you prefer the speedups without making this package multi-arch, yep, adding Tornado in the Python SPK is the way to go. I'll correct that next chance I get.

Diaoul:
Well, let's start without speedups and we'll see if that's a problem after the release.

patsissons:
for the record, I've been running GateOne on my 412+ since I opened this PR and have noticed no issues without the speedups compiled in. That said my typical usage of GateOne is basic command line server maintenance, so I'm not sure if my day-to-day usage would benefit.

PR #1528, opened 2015-03-08
Changed default fork entry during installation. (Sickbeard-custom)

neoatomic:
The mr-orange fork doesn't exist anymore. (removed from Git-hub)
Replaced it with the very popular Sickrage fork.
Also added a Dutch install_ufile .

### DIFF
--- a/spk/sickbeard-custom/src/installer.sh
+++ b/spk/sickbeard-custom/src/installer.sh
@@ -22,7 +22,7 @@ FWPORTS="/var/packages/${PACKAGE}/scripts/${PACKAGE}.sc"
 preinst ()
 {
     # Check fork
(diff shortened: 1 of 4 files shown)


Diaoul:
Shouldn't we create a new SPK for SickRage instead as the name changed?

neoatomic:
Agree, however as the mr-orange fork doest exist anymore the url needs changing anyway.

Here is a SPK witch Sickrage, it only needs the Git Url changed. (currently points to a non official Sickrage fork.)
http://cytec.us/spk/download/sickrage_noarch_20150103-1.spk

cytec:
I think we should make a seperate Package for SickRage as it is technically more than just a fork... it started as a fork but evolved to a more complex thing in my opinion so i think it may deserves a seperate SPK as well.

@Diaoul i can change the default installation repo to the original repo and make a PR if you like. I'd rather provide a "custom" SPK with default URL and cloning on installation. So the user can (if he likes) change the URL on installation

Diaoul:
OK lets do that. Thanks

Dr-Bean:
So, to be clear: we're not going to merge this PR as-is, unless it points to the original SickBeard repo instead (same as e.g. the CouchPotato packages), and we're releasing another SB-based package called SickRage?

PR #1542, opened 2015-03-13
#1480: Fixed missing app config file.

Decipher:
Minor issue with missing app config file preventing the App icon from appearing in the DSM menu, as reported in #1480.

Didn't update the changelog or version as I don't believe a release has been built yet.


Dr-Bean:
Thanks, that should do it, I think ;)

PR #1545, opened 2015-03-13
Update c_icap to version 0.3.5

GaetanCambier:


cross/c_icap/PLIST
```diff
@@ -1,7 +1,7 @@
 bin:bin/c-icap
 bin:bin/c-icap-client
-bin:bin/c-icap-config
-bin:bin/c-icap-libicapapi-config
+rsc:bin/c-icap-config
+rsc:bin/c-icap-libicapapi-config
```
Diaoul: Why the change?

GaetanCambier:
these 2 file are not real binary file but script file --> strip failed on these file

PR #1555, opened 2015-03-15
Convert remaining pybundles to wheel

Dr-Bean:
Converted the remaining pybundles to wheel:
- Debian Chroot
- Deluge
- GateOne
- HAProxy

- Subliminal (+update Subliminal to 0.7.5 and prepare for new release)


Dr-Bean:
Note: untested, except for Subliminal, although I don't expect any issues. 
We'll need to release new packages when pip >1.5.6 / Python 2.7.9 is released and check the Python version in SPK_DEPENDS to at least be set to `python>=2.7.6-8`

PR #1556, opened 2015-03-15
Update owncloud to 8.0.2

mirweb:


mirweb:
some merging problems, replaced by pull request #1559

PR #1562, opened 2015-03-17
Fix broken url download

GaetanCambier:


Dr-Bean:
That can't be right. cross/znc was updated to 1.6.0. Maybe update your fork?

PR #1565, opened 2015-03-18
Update apr to version 1.5.1

GaetanCambier:
apr version 1.4.6  not availaible on web site


cross/apr/Makefile
```diff
@@ -1,8 +1,8 @@
 PKG_NAME = apr
-PKG_VERS = 1.4.6
+PKG_VERS = 1.5.1
 PKG_EXT = tar.bz2
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = http://apache.crihan.fr/dist//apr
+PKG_DIST_SITE = http://www.apache.org/dist/apr/
```
Dr-Bean: No trailing slash here

Dr-Bean:
Haven't verified, but seems weird to me that it needs a native build, I would think a `./configure, make, make install` would do it with the relevant flags. Otherwise, looks okay.

GaetanCambier:
the native build is in the pr

GaetanCambier:
done

Dr-Bean:
Might be something on my side, but this isn't working for me. The compile errors out with `bin/bash: tools/gen_test_char_native: cannot execute binary file`. Can you recheck that it's successful on your side after you clean both the native and cross builds?

On that note, how about dropping the native build, and patching out the tools/gen_test_char check altogether? See here for an example: http://lists.openembedded.org/pipermail/openembedded-core/2014-June/093158.html

Dr-Bean:
Ah, figured it out. The native target has to be built first, can't build it as part of the cross make process: probably Make- or environment flags being passed on causing that to happen.
Maybe that can be fixed, either via my previous suggestion, or by correcting the flags in the native's Makefile.

GaetanCambier:
i have compared compile.log from native and cross pakage and the make grogram used is not correct when cross compil the native package. i test a correction

Diaoul:
On that topic, I think we can clear env variables when buiding a native package.

GaetanCambier:
resolbed by compiling all native package

PR #1568, opened 2015-03-20
Add CalibrePHP

demorfi:
This is a simple HTML and OPDS web server to access a database created by the Calibre application. CalibrePHP was written using CakePHP.


Dr-Bean:
Just for my information: We have COPS and BBS on the repository, an open PR for MyReadings: does your package add something we're missing with regards to functionality?

The PR looks quite nice :) One suggestion: Maybe take a look at the COPS package and how the installer and wizard work for that package. Currently, your package doesn't seem to include a way to set the location of the Calibre dir, or the permissions needed to access it. 
That method also has the benefit that you can reuse the path information to set the open_basedir yourself, instead of asking that of the user.

demorfi:
This is an alternative COPS and BBS.
There is work as a guest, and user accounts with the keys to xml reader.
screenshot-192 168 0 11 2015-03-21 07-10-07

You can apply a custom filter by tags, for example to limit the issuance books of certain users.
screenshot-192 168 0 11 2015-03-21 07-10-52

Can be read as epub, pdf books in the browser.

The default used is an empty database to work correctly, the real address is placed on a base sqlite.
In the settings, you can at any time change the address to base. Before change the address is checked for correctness.

If a change of address at the time of installation it is necessary, I can change a little logic database.

PR #1572, opened 2015-03-21
WIP: Implement updated Subliminal API

Dr-Bean:
I've implemented most of the new 0.7.5 API in Subliminal's GUI, to get that back to working order again...along with adding some more missing modules.
Because of the changed language handling, this package isn't backwards compatible. I haven't added that backwards-incompatible code in yet.

@Diaoul, you'll probably want to take a look at this code before I go any further with it... ;)


spk/subliminal/src/app/subliminal.js
```diff
@@ -2685,14 +2640,13 @@ SYNOCOMMUNITY.Subliminal.PanelParameters = Ext.extend(SYNOCOMMUNITY.Subliminal.F
...
                             ["zap", "Zapotec"],
                             ["zbl", "Blissymbols"],
                             ["zen", "Zenaga"],
                             ["zha", "Zhuang"],
-                            ["znd", "Zande languages"],
                             ["zul", "Zulu"],
                             ["zun", "Zuni"],
+                            ["zxx", "No linguistic content"],
                             ["zza", "Zaza"],
                             ["por-BR", "Portuguese (Brazil)"]
```
Dr-Bean: Hm, forgot to take this one out...probably won't work.

Diaoul:
I just updated HAProxy which also has a GUI. I was thinking reworking the UI with DSM 5.x new look and feel but I don't have time for that right now. Subliminal could use the same treatment.

I trust you on this one :)

Dr-Bean:
Well, I wouldn't trust me with it, but thanks for the vote of confidence :P I'll release the new package as a beta anyway, seeing as it's backwards-incompatible.

If we update to the new DSM5 code, any idea what that would mean for DSM4.x support? I haven't ever looked at it, I think.

Diaoul:
No idea how this would behave on DSM 4.x. If that's a problem, lets just drop 4.x.

PR #1036, opened 2014-06-01
Gentoo Chroot

MasterRoot24:
I've created a package called Gentoo Chroot, which is based on the existing Debian Chroot package.

Builds succeed for all architectures defined in the `Makefile`. However, I only have a DS214play (evansport) available to test on, so I'm not sure if this installs/runs on other architectures.

I've squashed down all my commits to 7 major changes, which I hope helps.

PR #1579, opened 2015-03-22
Gentoo Chroot

MasterRoot24:
As per #1036, I've rebased the branch against master.


spk/gentoo-chroot/Makefile
```diff
@@ -0,0 +1,93 @@
+SPK_NAME = gentoo-chroot
+SPK_VERS = 13.0
+SPK_REV = 1
+SPK_ICON = src/gentoo.png
+DSM_UI_DIR = app
+
+DEPENDS = 
+SPK_DEPENDS = "python>2.7.3-3"
+
+MAINTAINER = "MasterRoot24"
```
Diaoul: Please use SynoCommunity as maintainer

spk/gentoo-chroot/Makefile
```diff
@@ -0,0 +1,93 @@
...
+endif
+
+GENTOO_STAGE3 := $(shell curl -s http://distfiles.gentoo.org/releases/$(GENTOO_CPU)/autobuilds/latest-stage3-$(GENTOO_ARCH).txt | sed '/^\#/d' | awk '{print $$1}')
+GENTOO_STAGE3_URL := http://distfiles.gentoo.org/releases/$(GENTOO_CPU)/autobuilds/$(GENTOO_STAGE3)
+GENTOO_STAGE3_FILE := $(shell basename $(GENTOO_STAGE3))
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: gentoo-chroot_extra_install
+gentoo-chroot_extra_install: $(STAGING_DIR)/share/requirements.pybundle
```
Diaoul: pybundles are not working anymore, we need to use wheels here instead

spk/gentoo-chroot/src/dsm-control.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/sh
+
+# Package
+PACKAGE="gentoo-chroot"
+DNAME="Gentoo Chroot"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PATH="${INSTALL_DIR}/bin:/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/usr/syno/sbin:/usr/syno/bin"
```
Diaoul: `PATH="${INSTALL_DIR}/bin:${PATH}`

spk/gentoo-chroot/src/installer.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Package
+PACKAGE="gentoo-chroot"
+DNAME="Gentoo Chroot"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PYTHON_DIR="/usr/local/python"
+CHROOTTARGET="${INSTALL_DIR}/var/chroottarget"
+PATH="${INSTALL_DIR}/bin:${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/usr/syno/sbin:/usr/syno/bin"
```
Diaoul: `PATH="${INSTALL_DIR}/bin:${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:${PATH}"`

spk/gentoo-chroot/src/installer.sh
```diff
@@ -0,0 +1,82 @@
...
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # Create a Python virtualenv
+    ${VIRTUALENV} --system-site-packages ${INSTALL_DIR}/env > /dev/null
+
+    # Install the bundle
+    ${INSTALL_DIR}/env/bin/pip install --no-index -U ${INSTALL_DIR}/share/requirements.pybundle > /dev/null
```
Diaoul: pybundle stuff, see above

Diaoul:
I added a few comments, nice PR :)

MasterRoot24:
Requested changes have been made and pushed.

Diaoul:
LGTM
@Dr-Bean?

spk/gentoo-chroot/Makefile
```diff
@@ -0,0 +1,92 @@
+SPK_NAME = gentoo-chroot
+SPK_VERS = 13.0
+SPK_REV = 1
+SPK_ICON = src/gentoo.png
+DSM_UI_DIR = app
+
+DEPENDS = 
+WHEELS = pyextdirect==0.3.1 flask Werkzeug Jinja2 itsdangerous
+SPK_DEPENDS = "python>2.7.3-3"
```
Dr-Bean: It's actually `python>=2.7.8-10`, Debian package needs that updated too ;)

spk/gentoo-chroot/Makefile
```diff
@@ -0,0 +1,92 @@
...
+
+MAINTAINER = "SynoCommunity"
+DESCRIPTION = "Gentoo is a free operating system based on either Linux or FreeBSD that can be automatically optimized and customized for just about any application or need. Gentoo Chroot allows you to install the Gentoo OS inside your DiskStation, alongside DSM. This package is intended for advanced users only."
+RELOAD_UI = yes
+DISPLAY_NAME = "Gentoo Chroot"
+CHANGELOG = 
+
+HOMEPAGE   = "http://www.gentoo.org/"
+LICENSE    = 
+HELPURL    =
```
Dr-Bean: Drop this one, L20 covers it :) It won't do much, but just leave it in, until we get to fixing that.

spk/gentoo-chroot/Makefile
```diff
@@ -0,0 +1,92 @@
...
+
+COPY_TARGET = nop
+POST_STRIP_TARGET = gentoo-chroot_extra_install
+
+GENTOO_ARCH =
+ifeq ($(findstring $(ARCH),88f6281),$(ARCH))
+GENTOO_CPU = arm
+GENTOO_ARCH = armv7a
+endif
+ifeq ($(findstring $(ARCH),armada370 armadaxp),$(ARCH))
```
Dr-Bean: You can add ` armada375 alpine` in here as well.

spk/gentoo-chroot/Makefile
```diff
@@ -0,0 +1,92 @@
...
+GENTOO_CPU = arm
+GENTOO_ARCH = armv7a_hardfp
+endif
+ifeq ($(findstring $(ARCH),bromolow cedarview evansport x86),$(ARCH))
```
Dr-Bean: Add `avoton x64` here too.

spk/gentoo-chroot/Makefile
Dr-Bean: You might have overlooked this: `$(STAGING_DIR)/share/requirements.pybundle` can be removed.

Dr-Bean:
Added some comments. I can't build/test the package atm, but other than that, the code's looking good.

spk/gentoo-chroot/Makefile
MasterRoot24: Oops. Good spot!

spk/gentoo-chroot/Makefile
Dr-Bean: I keep getting that version wrong. `"python>=2.7.8-10"` would work, but let's make it what it should actually be: 
SPK_DEPENDS = `"python>=2.7.6-8"`

MasterRoot24:
Ok, I've made those extra changes and pushed.

Dr-Bean:
No other comments, so I'll merge next chance I have.

Dr-Bean:
@Diaoul The packages are about 200MB in size (depending on arch), and I'm getting a `413 Request Entity Too Large` when uploading. Along with that, uploading (just) the 5.1 packages would amount to 1,6GB of space needed. Can the repo be adjusted to handle that (and do we want it to, because 200MB is quite the size for a package...)?

Diaoul:
I can increase upload size but we will probably need to upgrade the storage of the repository.

Dr-Bean:
Alternatively, we can move this: `wget -nv -c -O $(WORK_DIR)/$(GENTOO_STAGE3_FILE) $(GENTOO_STAGE3_URL)` to `installer.sh` (with some adjustments here and there to make it work)

The way I'm seeing it, clients need to download those ~200MB anyway, either from the repository, or directly from the Gentoo servers. Might as well be direct from Gentoo, especially since it's probably not going to be a highly-used package.

MasterRoot24:
My concern with downloading the stage3 tarball on installation of the package is that on slow connections it can take some time and with no feedback to the UI (in the install phase). Having a bigger package download size IMO is better because the GUI provides the user with visibility on the download of the package.

Allowing users to download the package with a built in stage3 is less complexity for end users to endure and fairer on Gentoo's mirrors.

Dr-Bean:
Here's some arguments for the opposing side then:
The download time for slow connections is going to be exactly the same, regardless of where the data comes from. So that doesn't make a difference imo.
More importantly, the feedback in the GUI is limited either way, and even more restricted during a package download; at that point you are essentially looking at a 'greyed-out' dialog for the package itself. During that time, you can't interact with that dialog or Package Center at all. Only after the download is completed, the dialog will refresh, and starts with you needing to confirm the installation.

A way around that is to allow the package to download the tarball in the background (ie. maybe something similar to what the Horde package does [here](https://github.com/SynoCommunity/spksrc/blob/master/spk/horde/src/installer.sh#L99).
Fairer on Gentoo's mirror: hm. If the amount of package downloads would be significant, yes, I would definitely agree. But will the amount of users using this package really be a burden on the mirrors? 

As for a complexity for end users, I can't see it making a whole lot of difference for them. Personally, I might even prefer a background installation on a slow connection, at least then I can still access other packages in Package Center while the download is going on.

Dr-Bean:
I should add that one argument against a background installation is providing feedback when the download of the tarball fails. I don't think we can provide that feedback in the package installer once it switches to a background installation, so that would have to be handled in some other way.
Anyway, just providing an alternative way to handling packages of this size.

MasterRoot24:
> As for a complexity for end users, I can't see it making a whole lot of difference for them.

My point here was pretty much what you have said since: if the background download fails, how does the user know?

> I don't think we can provide that feedback in the package installer once it switches to a background installation

Hence, it's current implementation of downloading the stage3 at package build causes the build to fail if the download fails, stopping potentially broken packages going out with no feedback to the user available, causing more issues for end users than we need.

I should point out: I don't have access to a Synology NAS now, so I'm not able to look into how easy it would be to change the installation process to download in the background.

Diaoul:
FYI Debian chroot does provide feedback through its GUI and start script

PR #1591, opened 2015-03-23
Update ruby to version 2.0.0-p643

GaetanCambier:


Dr-Bean:
It would probably make sense to update `spk/ruby` with the updated version too, seeing as that the only reason for `native/ruby` and `cross/ruby`.
Although, while saying that, I'm not completely sure Ruby as a package actually works. Would you happen to know that?

Dr-Bean:
I'm closing this PR because a newer version of Ruby is available, and this PR needs additional work anyway. Thanks for taking the time to open the PR in the first place though.

If you're interested in keeping the Ruby package around, you might want to take a look at (and respond to) #1844 as well.

PR #1592, opened 2015-03-23
Update update_db.sh script

GaetanCambier:
Fix typo error in rsync command

Remove -u in squidguard command to prevent Segmentation fault


Diaoul:
Is the typo fixed? I don't think the command is actually rcync. Moreover, this is far too less modification to the package to be in a PR, please refactor the whole package and submit a PR for that.

Issue #1558, opened 2015-03-15
Mono should not be startable


Diaoul:
For consistency with other packages such as Python and Git

Dr-Bean:
Fixed with https://github.com/SynoCommunity/spksrc/commit/f897fee74f7f8e51550c6c2f80959f1f8336b3cf

Issue #1557, opened 2015-03-15
Mono installer.sh doesn't define USER


Diaoul:
All in the title.

Dr-Bean:
Fixed with https://github.com/SynoCommunity/spksrc/commit/f897fee74f7f8e51550c6c2f80959f1f8336b3cf

PR #1605, opened 2015-03-25
Mono: Removed beta flag & redundancy

maxrogers:


Dr-Bean:
Maybe see if #1557 and #1558 can be addressed too?

maxrogers:
@Dr-Bean, haha thanks! I made those fixes and squashed

Diaoul:
I'll give it a shot but I think it should work just fine :)

Dr-Bean:
Okay, just to be sure: after building the Mono package, test it too (before publishing)
On one of my devices, the current Mono package on the repo works with Sonarr just fine. After installing the package I just built, Sonarr won't start anymore, without any indication as to why. That might be my build environment or that device, but it's strange to say the least.

Dr-Bean:
Oh. #1609? :D

maxrogers:
Yeah, sorry. I tested it and saw that it ran appropriately. However, I didn't test actually running a program that needed libraries (sonarr.) I will make sure to do that next time!

Dr-Bean:
Sure, I was not putting any blame you at all. I'm in the process of rebuilding my build/test environment, so it's hard to pinpoint a cause. Anyway, building (again...;)) as we speak. I'll merge when I can confirm the fix :)

maxrogers:
Awesome! And I will publish once you confirm and merge. Fun stuff! Thanks @Dr-Bean

Issue #671, opened 2013-10-03
Package for NZBDrone [Request]


EvtK:
Hi there,

I don't know if it is even possible to request here, but I would really like to install NZBDrone as package on my synology NAS. Can this be added to the repository?

I read NZBDrone is build on mono, and I have seen an official synology mono package. But I'm no expert and would not know how to use this package to run NZBDrone.

Many thanks!

Dr-Bean:
Mono v3.10.0-2 and Sonarr v20150101-3 have been published on the repository. Note that the packages are still in beta for now, so you'll need to enable seeing beta packages in the Package Center settings.

For people using the last packages as provided by @maxrogers in [this comment](https://github.com/SynoCommunity/spksrc/issues/671#issuecomment-68173062), the update won't show up via Package Center, as it's basically the same version. No need to reinstall. If a package update is released, it'll be recognized automatically.

Please report problems in a [new issue](https://github.com/SynoCommunity/spksrc/issues/new): this one will be closed (finally, yay)
Thanks @maxrogers for your work and being patient with us until spksrc was up for the task :)

Diaoul:
Awesome :tada: 
@maxrogers Once the new spkrepo version will be in production I'll give you permission to publish updates to mono and sonarr. This should happen this month.

Stanzilla:
Looks like the icon is missing though: http://i.imgur.com/ktjDaEM.png

maxrogers:
@Dr-Bean Thanks!

@Diaoul, that will be so awesome!  :+1: Let me know

@Stanzilla, good catch! Fixed here: https://github.com/SynoCommunity/spksrc/pull/1422

SpiderDice:
I'm on Sonarr Ver. 2.0.0.2594 and all of my shows are displaying an air time of 12:00am. Anyone else seeing this?

derik0709:
@SpiderDice Yeah, it is because Sonarr switched to using theTVDB instead of Trakt. theTVDB doesn't have air times for shows. Everything else that has changed is explained here https://forums.sonarr.tv/t/sonarr-now-with-tvdb/3314

SpiderDice:
@derik0709 Ahh..okay. Guess it's not that big of a deal since it's still picking up showing when they are released.

schumi2004:
I thought I saw a list of architectures mono is working for in the past but can't seem to find it. What is current status for Sonarr + Mono, what architectures are excluded?

maxrogers:
[...]

PR #1609, opened 2015-03-25
Mono: Fix- Redundant lines should have been removed from cross not spk

maxrogers:
The essential mono lib files do not make it into the package if these lines are removed from ```spk/mono/Makefile``` (as done in most recent commit.) To fix this I removed them from cross instead. This solves the issue

### DIFF
--- a/cross/mono/Makefile
+++ b/cross/mono/Makefile
@@ -25,13 +25,3 @@ MONO_NATIVE_INSTALL = $(WORK_DIR)/../../../native/$(PKG_NAME)/work-native/instal
 .PHONY: myPreConfigure
 myPreConfigure:
 	$(RUN) autoreconf
-
-.PHONY: myPostInstall
-myPostInstall:
-	install -m 755 -d $(STAGING_DIR)/lib
-	tar -cpf - -C $(MONO_NATIVE_INSTALL)/usr/local/lib . | tar -xpkf - -C $(STAGING_DIR)/lib
-	install -m 755 -d $(STAGING_DIR)/etc/mono
-	tar -cpf - -C $(MONO_NATIVE_INSTALL)/usr/local/etc/mono . | tar -xpkf - -C $(STAGING_DIR)/etc/mono
-	install -m 755 -d $(STAGING_DIR)/bin
-	tar -cpf - -C $(MONO_NATIVE_INSTALL)/usr/local/bin . | tar -xpkf - -C $(STAGING_DIR)/bin
-
--- a/spk/mono/Makefile
+++ b/spk/mono/Makefile
@@ -41,6 +41,12 @@ include ../../mk/spksrc.spk.mk
 
 .PHONY: mono_extra_install
 mono_extra_install:
+	install -m 755 -d $(STAGING_DIR)/lib
+	cp -nR $(MONO_NATIVE_INSTALL)/usr/local/lib $(STAGING_DIR)/
+	install -m 755 -d $(STAGING_DIR)/etc/mono
+	cp -nR $(MONO_NATIVE_INSTALL)/usr/local/etc/mono $(STAGING_DIR)/etc/
+	install -m 755 -d $(STAGING_DIR)/bin
+	cp -nR $(MONO_NATIVE_INSTALL)/usr/local/bin $(STAGING_DIR)/
 	install -m 755 -d $(STAGING_DIR)/var
 	install -m 755 -d $(STAGING_DIR)/app
 	install -m 755 -d $(STAGING_DIR)/app/images


Dr-Bean:
Well, it would seem the error returns with this fix: the files are already created by the cross package, and (from what I see) it looks like the behavior of tar changed to cause the error.

@maxrogers  What exactly should happen there? Should the files which are created by the cross package be overwritten with the files from the native package, or should only the missing files be copied over?
From the flags in the tar command, it looks like `-k` is added to preserve existing files, but maybe that's broken in Jessie?

Dr-Bean:
Fix: Replace the current commands with `tar -cpf - -C $(MONO_NATIVE_INSTALL)/usr/local/lib . | tar --skip-old-files -xpf - -C $(STAGING_DIR)/lib` and so on for all three operations.

Documentation on `skip-old-files` vs `-keep-old-files`: http://lists.gnu.org/archive/html/bug-tar/2011-11/msg00049.html

maxrogers:
@Dr-Bean, So as you guessed, the goal is to put the missing mono libraries and associated files that are only created in native into the cross/package. So yes, it needs to just add the missing files over those created in cross.

Yeah, I am running wheezy where that command still works. Should I update to Jessie? I will try this new command. Thanks!

Dr-Bean:
Well, I'm only running Jessie to test Docker (and I needed to set up a new build env anyway), that's easier on Jessie than on Wheezy (although doable, I suppose)

But you do bring up a good point. The `--skip-old-files` does not exist on tar 1.26, I conveniently overlooked that :/
You can upgrade to tar 1.27, but you'll probably need to add Jessie to `/etc/apt/sources.list`.

Dr-Bean:
@Diaoul Do you think we should put the above in the README?

Diaoul:
Not yet, I need to create a Vagrantfile with docker provisioner so Windows users can uses spskrc too through a VM. Once it's confirmed everything works fine on Docker i'll update the readme

Dr-Bean:
I meant documenting the upgrade to tar 1.27 on Wheezy (or downgrade to 1.26 on Jessie, that works too). Because depending on which fix is implemented, either Wheezy works, or Jessie works (and by extension, Docker), but not both.

Besides that, I don't know how other Debian-based distros are affected, but I assume the same issue might occur.

Anyway, another option is to replace the tar/untar with a more traditional cp command, like `cp -nR` or something similar. I'm currently building a package with that approach, let's see.

Dr-Bean:
How about this then:
```
	install -m 755 -d $(STAGING_DIR)/lib
	cp -nR $(MONO_NATIVE_INSTALL)/usr/local/lib $(STAGING_DIR)/
	install -m 755 -d $(STAGING_DIR)/etc/mono
	cp -nR $(MONO_NATIVE_INSTALL)/usr/local/etc/mono $(STAGING_DIR)/etc/
	install -m 755 -d $(STAGING_DIR)/bin
	cp -nR $(MONO_NATIVE_INSTALL)/usr/local/bin $(STAGING_DIR)/
```

maxrogers:
@Dr-Bean, nice! That works wonderfully! Thanks

I added that to this PR

Dr-Bean:
Perfect :)

Dr-Bean:
For the record, the reason it didn't work in cross/mono/Makefile is most likely because of the PLIST that didn't contain the added files.
That might be something to consider for a future version of Mono, let's leave it as is for now.

maxrogers:
@Dr-Bean, I would love it if you could help me make that work. I set it up that way the first time but ran into this issue: https://github.com/SynoCommunity/spksrc/issues/671#issuecomment-38137601

Incidentally adding to the PLIST resulted in the same error that @jerroen is experiencing here: https://github.com/SynoCommunity/spksrc/issues/1613#issuecomment-87012260

Dr-Bean:
Ah, right, there's always a catch ;)
The stripping takes place on three of the prefixes: `lib:`, `libexec:` and `bin:` (It's mentioned in the wiki too, although I could see how one could miss that: https://github.com/SynoCommunity/spksrc/wiki/PLIST-files). 
When using the prefix `rsc:` does not strip the files.

So, you can do two things from what I see: either copy after the stripping takes place (so that would be `POST_STRIP_TARGET` in the cross/Makefile, or, prefix the entries with `rsc:` in the PLIST (you could presumably do that with the whole dir, although then the original entries would not be stripped while they actually should be stripped)

Dr-Bean:
But #1613 should not be affected by my previous comment, right? If the copy takes place in spk/Makefile, then those aren't stripped, right?

maxrogers:
@Dr-Bean, right, it shouldn't have caused @jerroen's issue. I am trying to figure out what is going on. My best guess right now is armada375 failed to compile properly for some reason. I am running a fresh compile of it to have him try. However, the file size does look right as you mentioned.... not sure what happened as bromolow worked for me just great (from the same batch...)

Dr-Bean:
Yeah, tested the same arch here. I don't have a VM around to check, but where should the mscorlib.dll come from: native, or is it compiled via the cross package?

maxrogers:
@Dr-Bean, mscorlib comes from native

Dr-Bean:
Eugh...just realised that was kind of a dumb question :/ Anway, just wanted to report that 88f6281 does work with the new package.

Diaoul:
Testing on x64 arch is not relevant, the best way to test is to use some arm arch.

Diaoul:
Well I don't know what you published @maxrogers but I just compiled with Docker a mono + nzbdrone package and tested on armada370 and it works fine. I'll push to the repository

PR #1612, opened 2015-03-27
OpenZWave support for Domoticz

Decipher:
As per #1488

### DIFF
--- a/cross/cp210x/Makefile
+++ b/cross/cp210x/Makefile
@@ -1,0 +1,29 @@
+PKG_NAME = cp210x
+PKG_DIR = $(PKG_NAME)
+
+DOWNLOAD_TARGET = kernel
+
+REQ_KERNEL = 1
+
+HOMEPAGE = http://sourceforge.net/projects/dsgpl/
+COMMENT  = CP210x USB to UART Bridge VCP Drivers
+LICENSE  = GNU GPL
+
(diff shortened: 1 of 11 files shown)


Decipher:
Possible something went wrong, this may not be ready for Pull yet. Tried compiling but patch fails. Should have tested more before I re-did the PR against Master.

Will look more at it ASAP.

Decipher:
So the only issue is the line endings in the patch (cross/domoticz/patches/CMakeLists.txt.patch). It could be related to me squashing the commits on my local OS X environment rather than my SPKSRC Vagrant box.

Git doesn't want to deal with the issue, just wants to convert it to the non-working version. Thoughts?

Diaoul:
What should be the line ending or the patch?

Decipher:
Honestly I don't know. I don't really live in a world where I need to worry about line endings very often. Is there an easy way to find out?

gkroone:
Domoticz uses crlf (msdos) line endings. If i have issues with it i run unix2dos to correct them.

In linux, the line endings appear with an additional ^M at the end. Not sure how mac shows it.

In one of the Domoticz versions i noticed that it accidentally switched to unix line endings,  which was then corrected again in a later version.

KaraokeStu:

Thanks,

Stuart

PR #848, opened 2014-02-26
Shairport Airplay Receiver

lost-carrier:
Hi there,

this is my first try to build a Synology package.

Since I have some AirPlay-enabled iDevices in use, as well as a Synology hooked via a USB soundcard to my amp, I wanted to use the storage as an AirPlay receiver. Synology provides various Apps that would stream TO AirPlay receivers, but nothing that would make the NAS act as an AirPlay receiver.

Luckly there's a project called "ShairPort" around (well, it's even @ github: https://github.com/abrasive/shairport). So I compiled shairport using your crosscompile framework. Works fine with on my DS1512+. However, there are some known issues:

1. shairport currently runs as root. I've tried to use a "shairport" user, but the /dev/snd and alsa-config are a bit messed-up in the DSM, so I could not allow any other user than root to play back via alsa without messing around with chgrp and chmod in /dev/snd/. Suggestions welcome.

2. avahi-deamon has to support avahi-publish and dbus. Did not apply any of that stuff http://blog.elsdoerfer.name/2011/05/15/shairport-on-synology-nas/ suggests and it runs fine on my box. However: no guaranty it will run on every box. Beta testers welcome.

3. the icon I used is from http://gnarld.com/how-to-use-airplay-with-your-ipad-2-iphone-4s-apple-tv/ - hope there's no copyright issue... (...graphically gifted contributions welcome. ;-) )

Cheers,
Losty

PR #1617, opened 2015-03-28
Re-base of lost-carrier/spksrc after re-naming of develop branch to master 

thunfischbrot:
See https://github.com/SynoCommunity/spksrc/pull/848#issuecomment-86654427

Please let me know in case I messed something up or this was not what you expected.

Thanks!


cross/shairport/Makefile
```diff
@@ -0,0 +1,23 @@
+PKG_NAME = shairport
+PKG_VERS = 1.1.1
+PKG_EXT = zip
+PKG_DIST_NAME = 1.1.1.zip
```
Dr-Bean: $(PKG_VERS).$(PKG_EXT)

cross/shairport/Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+DEPENDS = cross/openssl cross/alsa-lib
+
+HOMEPAGE = https://github.com/abrasive/shairport
+COMMENT  = Use Synology NAS as an Airplay-Receiver. Requires USB-Soundcard or speakers.
+LICENSE  =
+
+GNU_CONFIGURE = 1
+
+#ADDITIONAL_LDFLAGS = "-static"
```
Dr-Bean: If this is not needed, let's remove it.

cross/shairport/Makefile
```diff
@@ -0,0 +1,23 @@
+PKG_NAME = shairport
+PKG_VERS = 1.1.1
+PKG_EXT = zip
+PKG_DIST_NAME = 1.1.1.zip
+PKG_DIST_SITE = https://github.com/abrasive/shairport/archive
+PKG_DIR = $(PKG_NAME)-1.1.1
```
Dr-Bean: $(PKG_NAME)-$(PKG_VERS)

spk/shairport/Makefile
```diff
@@ -0,0 +1,24 @@
+SPK_NAME = shairport
+SPK_VERS = 1.1.1
+SPK_REV = 2
```
Dr-Bean: Start with `1`, it would be the first SynoCommunity version.

spk/shairport/Makefile
```diff
@@ -0,0 +1,24 @@
...
+SPK_ICON = src/shairport.png
+
+DEPENDS =  cross/$(SPK_NAME)
+
+MAINTAINER = Lost Carrier
```
Dr-Bean: SynoCommunity

spk/shairport/Makefile
```diff
@@ -0,0 +1,24 @@
...
+DESCRIPTION = Use Synology NAS as an AirPlay-Receiver. Requires USB-Soundcard or speakers.
+RELOAD_UI = yes
```
Dr-Bean: I believe this is not needed if there's no UI interaction (e.g. no icon to be placed)

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,96 @@
...
+DNAME="ShairPort"
+# Others
+SHAIRPORT="${SYNOPKG_PKGDEST}/bin/shairport"
+VAR_DIR="${SYNOPKG_PKGDEST}/var"
+PID_FILE="${VAR_DIR}/${PACKAGE}.pid"
+
+
+start_daemon ()
+{
+    # Launch the service in the background.
```
Dr-Bean: Check this file for TABs, and replace with spaces

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,96 @@
...
+	cd ${SYNOPKG_PKGDEST}
```
Dr-Bean: Is this needed? Defining and using a PATH variable might help here.

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,96 @@
...
+	$SHAIRPORT --daemon --pidfile="$PID_FILE" --log=$SYNOPKG_TEMP_LOGFILE
```
Dr-Bean: - Use `${VARIABLE}` for consistency. 
- The logfile should be placed in `${INSTALL_DIR}/var/shairport.log` or something like that.

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,96 @@
...
+    do
+        daemon_status && break
+        let counter=counter-1
+        sleep 1
+    done
+}
+
+stop_daemon ()
+{
+    # Kill the servive.
```
Dr-Bean: `service`

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,96 @@
...
+    kill -9 `cat ${PID_FILE}`
+
+    # Wait until transmission is really dead (may take some time).
```
Dr-Bean: Even though it's just a comment, it's `ShairPort` ;)

spk/shairport/src/installer.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/sh
+
+# Package
+PACKAGE="shairport"
+DNAME="ShairPort"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+SSS="/var/packages/${PACKAGE}/scripts/start-stop-status"
+PATH="${INSTALL_DIR}/bin:/usr/local/bin:/bin:/usr/bin:/usr/syno/bin"
```
Dr-Bean: Use ${PATH}

spk/shairport/src/installer.sh
```diff
@@ -0,0 +1,74 @@
...
+    exit 0
+}
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # persuade asla to let use use that playback device
+    mv ${INSTALL_DIR}/share/alsa/alsa.conf ${INSTALL_DIR}/share/alsa/alsa.original.bak
```
Dr-Bean: Can't this be done in a separate patch/fix in cross or even spk if need be?

spk/shairport/src/installer.sh
```diff
@@ -0,0 +1,74 @@
...
+    cat ${INSTALL_DIR}/share/alsa/alsa.original.bak | grep -v "defaults.pcm.dmix.rate" > ${INSTALL_DIR}/share/alsa/alsa.conf
+    rm ${INSTALL_DIR}/share/alsa/alsa.original.bak 
+    echo "defaults.pcm.dmix.rate 44100" >> ${INSTALL_DIR}/share/alsa/alsa.conf
+    echo "defaults.pcm.ipc_gid    \"root\"" >> ${INSTALL_DIR}/share/alsa/alsa.conf
+
+    # create some /var directory - just for us...
+    mkdir ${INSTALL_DIR}/var
```
Dr-Bean: This belongs in spk/Makefile

Dr-Bean:
Added some comments, haven't actually tested. 

One thing to note: ShairPort is no longer maintained: https://github.com/abrasive/shairport#shairport-is-no-longer-maintained. We've removed packages in the past because of that.

spk/shairport/src/installer.sh
```diff
@@ -0,0 +1,74 @@
...
+{
+    exit 0
+}
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # persuade asla to let use use that playback device
```
Dr-Bean: `alsa`...apologies for the nitpicking, I can't unsee it now ;)

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,96 @@
...
+        daemon_status && break
+        let counter=counter-1
+        sleep 1
+    done
+}
+
+stop_daemon ()
+{
+    # Kill the servive.
+    kill -9 `cat ${PID_FILE}`
```
Dr-Bean: Isn't there a way to tell Shairport to stop, other than killing the process?

spk/shairport/src/dsm-control.sh
Diaoul: Also `kill -9` is really rude. What about using the standard `stop_daemon` here?

thunfischbrot:
Thank you. Your time spent on reviewing is very much appreciated.
I addressed your comments. Just tested this version, compiles and runs.

Will check in the coming week whether the shairport fork at https://github.com/mikebrady/shairport-sync is a maintained alternative.

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/sh
+
+# Package
+PACKAGE="shairport"
+DNAME="ShairPort"
+
+# Others
+SHAIRPORT="${SYNOPKG_PKGDEST}/bin/shairport"
```
Dr-Bean: Define `INSTALL_DIR` like this: https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/src/installer.sh#L8
Then, use `${INSTALL_DIR}` instead of `${SYNOPKG_PKGDEST}` throughout this file. Check how tt-rss does it (although most packages use this approach, tt-rss is just an example)

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,84 @@
...
+VAR_DIR="${SYNOPKG_PKGDEST}/var"
```
Dr-Bean: Add the var dir to the package like this: https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/Makefile#L40. 
Then, your PID_FILE might be defined as done here: https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/src/dsm-control.sh#L14

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,84 @@
...
+PID_FILE="${VAR_DIR}/${PACKAGE}.pid"
+
+start_daemon ()
+{
+    # Launch the service in the background.
+    ${SHAIRPORT} --daemon --pidfile="$PID_FILE" --log=${VAR_DIR}/${PACKAGE}.log
```
Dr-Bean: As Diaoul suggested, maybe use the start-stop daemon Busybox provides. See https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/Makefile#L33-L34.
Usage would look quite similar to this: https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/src/dsm-control.sh#L19.

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,84 @@
...
+        daemon_status && break
+        let counter=counter-1
+        sleep 1
+    done
+}
+
+stop_daemon ()
+{
+    # TODO figure out a cleaner way to stop shairport
+    kill `cat ${PID_FILE}`
```
Dr-Bean: With the `start-stop-daemon` approach mentioned in the previous comment, this would look nicer. Something like this: https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/src/dsm-control.sh#L25-L26

spk/shairport/src/installer.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/sh
+
+# Package
+PACKAGE="shairport"
+DNAME="ShairPort"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+SSS="/var/packages/${PACKAGE}/scripts/start-stop-status"
+PATH="${INSTALL_DIR}/bin:/usr/local/bin:/bin:/usr/bin:/usr/syno/bin:${PATH}"
```
Dr-Bean: Let's remove everything we don't need, which makes it look like this:
`PATH="${INSTALL_DIR}/bin:${PATH}"`

spk/shairport/src/installer.sh
```diff
@@ -0,0 +1,71 @@
...
+{
+    exit 0
+}
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+	# TODO maybe add an spk/shairport/src/alsa.conf with the correct configuration, and place that file during installation.
```
Dr-Bean: The general approach to add a 'standard' file would look like this: https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/Makefile#L42 (put your file in `/spk/shairport/src`), then you can simply replace the current file during the installer.

lost-carrier:
Sry - had to pass an exam during the WE. Will have a look into this, during the week, too...

lost-carrier:
...managed to get shairport-sync-2.2.2 compiled (thanks @thunfischbrot for that link), but "make install" breaks. Raised an issue @ https://github.com/mikebrady/shairport-sync/issues/66

lost-carrier:
@thunfischbrot can you add me or transfer?

```
vagrant@vagrant:~/shairport2/spksrc$ git commit
[develop ba1dd69] Moved from shairport-1.1.1 to shairport-sync-2.2.2. Added libdaemon and popt since required by shairport-sync. Autoconfigure suxx.
 4 files changed, 41 insertions(+)
 create mode 100644 cross/libdaemon/Makefile
 create mode 100644 cross/libdaemon/PLIST
 create mode 100644 cross/popt/Makefile
 create mode 100644 cross/popt/PLIST
vagrant@vagrant:~/shairport2/spksrc$ git push
ERROR: Permission to thunfischbrot/spksrc.git denied to lost-carrier.
fatal: The remote end hung up unexpectedly
vagrant@vagrant:~/shairport2/spksrc$
```

thunfischbrot:
Done. Added you, can also transfer if you like.

lost-carrier:
That was fast - thnx.

Transfer: I don't mind...

Shairport-sync does compile (and work ;-)) now!

However: the handling of the pidfile obviously has changed. That's why the Synology Package Manager does not work properly, yet. @mikebrady used this init script: https://github.com/mikebrady/shairport-sync/blob/2.2/scripts/shairport-sync . Proably we can find something useful in there...

lost-carrier:
I'm currently working on the dsm-control-skript. I wanted to use the start-stop-daemon as it's used in the original project. I've also seen you're using it in other packages like horde, deluge, fengoffice. Unfortunately I cannot see how you get this to work, since it seems this start-stop-daemon-skript is not present on my diskstation. Can you give me a hint?

Dr-Bean:
@lost-carrier You're looking for this part here: 
https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/Makefile#L33,L34

lost-carrier:
Cool, thnx!

lost-carrier:
@Dr-Bean @cytec Thanks for your feedback. Can you re-check?

cross/libdaemon/Makefile
```diff
@@ -0,0 +1,18 @@
...
+PKG_DIST_SITE = http://0pointer.de/lennart/projects/$(PKG_NAME)
+
+DEPENDS =
+
+HOMEPAGE = http://0pointer.de/lennart/projects/libdaemon/
+COMMENT  = libdaemon is a lightweight C library that eases the writing of UNIX daemons.
+LICENSE  =
+
+GNU_CONFIGURE = 1
+CONFIGURE_ARGS = --localstatedir=/var
```
Dr-Bean: I haven't tested, but does using `/var` here have any effects? If so, I'd suggest using `${INSTALL_DIR}/var`

cross/shairport/Makefile
```diff
@@ -0,0 +1,34 @@
+PKG_NAME = shairport
+PKG_VERS = 2.2.3
+PKG_EXT = zip
+PKG_DIST_NAME = $(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://github.com/mikebrady/shairport-sync/archive
+PKG_DIR = $(PKG_NAME)-sync-$(PKG_VERS)
+
+DEPENDS = cross/libtool cross/popt cross/libdaemon cross/openssl cross/alsa-lib
+HOMEPAGE = https://github.com/mikebrady/shairport-sync
```
Dr-Bean: Add a newline between L8 and L9 for clarity :)

cross/shairport/Makefile
```diff
@@ -0,0 +1,34 @@
...
+COMMENT  = Use Synology NAS as an Airplay receiver. Requires USB soundcard or speakers.
+LICENSE  =
+
+FINAL_INSTALL_DIR  = /usr/local/$(PKG_NAME)
+FINAL_VAR_DIR = $(FINAL_INSTALL_DIR)/var
+
+GNU_CONFIGURE = 1
+
+CONFIGURE_ARGS = --prefix=$(STAGING_INSTALL_PREFIX) --with-alsa --with-ssl=openssl --with-tinysvcmdns --with-piddir=$(FINAL_VAR_DIR)
```
Dr-Bean: `$(INSTALL_PREFIX)/var` should work here iirc. Then you can drop the two vars on L13,14.

spk/shairport/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+DEPENDS =  cross/busybox cross/$(SPK_NAME)
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = Use Synology NAS as an AirPlay receiver. Requires USB soundcard or speakers.
+RELOAD_UI = no
+DISPLAY_NAME = ShairPort
+CHANGELOG = "Initial SynoCommunity Release"
+
+HOMEPAGE = https://github.com/abrasive/shairport
```
Dr-Bean: This is the old link

spk/shairport/src/dsm-control.sh
```diff
@@ -0,0 +1,86 @@
...
+DNAME="ShairPort"
+
+# Others
+PORT="4242"
+SHAIRPORT="shairport-sync"
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PATH="${INSTALL_DIR}/bin:${PATH}"
+DAEMON="${INSTALL_DIR}/bin/${SHAIRPORT}"
+VAR_DIR="${INSTALL_DIR}/var"
+PID_FILE="${VAR_DIR}/${PACKAGE}.pid"
```
Dr-Bean: Just use `${INSTALL_DIR}/var` instead of declaring a `VAR_DIR` variable :)

Dr-Bean:
Nice work :)
Two things: Please add a beta flag, and secondly: the fork is called `shairport-sync` if I'm not mistaken. Probably best to rename everything that refers to `Shairport` to the correct name then.

Dr-Bean:
And I forgot this one: is it possible to run this with a different account then `root`?

mikebrady:
Thanks for your work on this!

lost-carrier:
I will have a look into the details later, but one thing about the root-thingy: I've already tried about a year ago to run shairport on a different user. I'm not an expert for this, but in regular linux systems there's a group called "wheel" oder "audio". If you want to use the soundcard, you just have to be a member of that group. On the Synology, those groups do not exist. The problem is that the /dev/snd-stuff is owned by "root". A solution might be to create such a group and/or adapt the permissions on /dev/snd, but I'm afraid that'll be a big hassle to not break other apps and especially to keep those rights when there's DSM updates etc. I took a look into how Audiostation is doing it and it was running as root. If someone has a solution for this I'd be happy to implement it.

lost-carrier:
Done!

Any hints on finding a suitable the port no.?

Dr-Bean:
We've got a wiki page on choosing ports, here: https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports#how-to-choose-a-port.
Because the current port wasn't in our 'normal' range, I thought it was the default. Guess not ;)

lost-carrier:
Nope, Airplay default is 5000 ... ;-)

spk/shairport-sync/Makefile
```diff
@@ -0,0 +1,34 @@
+SPK_NAME = shairport-sync
+SPK_VERS = 2.2.3
+SPK_REV = 1
+SPK_ICON = src/airplay.png
```
Dr-Bean: We're probably going to need a different logo here, because the current one is owned/copyrighted by Apple as far as I can tell.

spk/shairport-sync/src/dsm-control.sh
```diff
@@ -0,0 +1,86 @@
...
+DNAME="Shairport Sync"
+
+# Others
+PORT="4242"
+SHAIRPORT="shairport-sync"
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PATH="${INSTALL_DIR}/bin:${PATH}"
+DAEMON="${INSTALL_DIR}/bin/${SHAIRPORT}"
+VAR_DIR="${INSTALL_DIR}/var"
+PID_FILE="${VAR_DIR}/${PACKAGE}.pid"
```
Dr-Bean: It's minor, but I thought I already mentioned this: Just use ${INSTALL_DIR} here.

Dr-Bean:
Apart from that copyrighted logo, should be good to go. Is there an alternative you can use?

lost-carrier:
@Dr-Bean I think I found a fancy one... see https://github.com/thunfischbrot/spksrc/commit/b5a01ebcc87404afe98053a8ea39596c53b7408d

Dr-Bean:
Please add a firewall rule, see [here](https://github.com/SynoCommunity/spksrc/pull/1000/files) for an example. That should be the last thing...probably ;)

lost-carrier:
@Dr-Bean Hmm... turns out that'll be not that easy... I can create a firewall rule for that tcp port without a problem, but the Airplay protocol uses that one only to announce three random udp port for streaming, control, etc. ... 

@mikebrady ...did I get this three-port-theroy right?
I'm not a C expert, but "0" means "some free port", right?
https://github.com/mikebrady/shairport-sync/blob/master/rtp.c#L497
How familiar are you with the protocol? Do you have any hits for defining a firewall rule?

@Dr-Bean ...or would it be ok to open just all udp ports beyond 1024? Probably not the exact idea of a firewall, but airplay is meant to be used on the local network, not that (evil) internet...

Dr-Bean:
If the ports are actually random, then we might have an issue with people using the built-in firewall, _if_ the firewall does not allow traffic to flow over a different port when initiated internally. I don't have an answer to that atm.
You should be able to quickly test that though: Turn on firewall, allow tcp/8303, leave the rest unchanged, see if the stuff works.

Anyway, another reason for the firewall rule is that the package won't be installed when the main port(s) are in use by other 3rd-party packages (which would result in a broken package). From that point of view, adding `tcp/8303` will at least cover that potential issue.

mikebrady:
Hi @lost-carrier. You are right about the three-port-theory. I suppose I could restrict the range of ports used by  shairport-sync to a range of, maybe, 20. Would that be useful?

lost-carrier:
@Dr-Bean OK, I've added the tcp port now.

@mikebrady Yes, I think that would be very helpful for the firewall rule. However: I think in general picking a random free(!) port is a good idea. So, can you add it somehow as a parameter for the executable?

lost-carrier:
PS: @mikebrady I found out there is already a rule present on the synology for udp 6001-6010. Is that the Apple default for this ports?

cross/shairport-sync/Makefile
```diff
@@ -0,0 +1,31 @@
...
+PRE_CONFIGURE_TARGET = myPreConfigure
+INSTALL_TARGET = myInstall
+
+include ../../mk/spksrc.cross-cc.mk
+
+.PHONY: myPreConfigure
+myPreConfigure:
+	$(RUN) autoreconf -i -f
+
+.PHONY: myInstall
```
Dr-Bean: Do you actually need to override the default INSTALL_TARGET? See: https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.install.mk#L55

spk/shairport-sync/Makefile
```diff
@@ -0,0 +1,36 @@
...
+SSS_SCRIPT       = src/dsm-control.sh
+ADDITIONAL_SCRIPTS =
+
+FWPORTS          = src/${SPK_NAME}.sc
+
+INSTALL_PREFIX = /usr/local/$(SPK_NAME)
+
+POST_STRIP_TARGET = shairport_extra_install
+
+BUSYBOX_CONFIG = daemon nice
```
Dr-Bean: Is `nice` used?

Dr-Bean:
Apart from a couple comments above, some missing `LICENSE` fields, and maybe the firewall rule/ports discussion above, it's just about done I think.

lost-carrier:
Done.

Firewall rule is added for tcp, so port will be reserved. Udp ports are picked totally arbitrary so it won't work when firewall is actually activated (...and blocking udp). Will add those as soon as @mikebrady found a way to restrict those...

Dr-Bean:
Note: some cross packages still don't have a license type mentioned, can be done during merge.

Dr-Bean:
Shairport-sync fails to build with the following toolchains: `bromolow-4.3`, `cedarview-4.3`, `powerpc-4.0`, `ppc824x-4.2`, `ppc854x-4.0`, `x86-4.3`.
Please verify. If there's a fix, go ahead and open a new PR. If not, use `REQUIRED_DSM` and `UNSUPPORTED_ARCHS` to exclude building the arches. 

Seeing the list, I'd propose setting `REQUIRED_DSM = 5.0`, meaning 5.0 is the minimum DSM version we create packages for.

Dr-Bean:
`shairport-sync 2.3.0-1` has been published on the beta channel (what could be compiled obviously ;))

mikebrady:
Hi there. I've just uploaded an update to Shairport Sync that allows it to look for ports from 7000 upwards, which should make it easier to cater for in firewall rules. It's in a temporary branch called `portnumbers` and not extensively tested. Do you want to try it, or should I wait a few days, do a little more testing and merge it into the development branch?

mikebrady:
Okay – I now see that the range of ports should begin at 6001. I'll make that change right away...

Dr-Bean:
@mikebrady Thanks for looking into this :)
From the code, I gather we should set a udp range anywhere up from three ports by setting `udp_port_range` in the config file (along with setting `udp_port_base`)? That should work fine, I think.

As for publishing packages, we usually prefer to stick with official releases, but we could make an exception here (and move back to the next official release when this change is merged with master).
 
@lost-carrier, will you look into this? Please make sure that upgrading from the existing package is handled too :)

mikebrady:
Yes indeed. The `udp_port_base` is set by default to 6001 and the `udp_port_range` is 100, so it will look for available ports starting at 6001 and working upwards for up to 100 ports. As you say, it only needs three ports, so a firewall range of 6001 to 6010 would be fine.

lost-carrier:
Sorry, guys - I was away for a while...

@Dr-Bean thanks a lot for merging!! ...awesome! ;-)

@mikebrady sure, I will test it. Thanks a lot!

@Dr-Bean yes, I will have a look into this. Is there anything special to look for? The packages does not have any configuration to keep, so just kill the old one and install the new one from scratch should be fine. How's you workflow here? New fork & PR?

lost-carrier:
@mikebrady technically I can already confirm it's working... but I still have to figure out the proper syntax for firewall definition... (created rule manually for now...)

mikebrady:
Thanks for the update. I seem to remember there was a firewall  rule from a (?) previous version of Shairport? Anyhow, let me know if I can help.

Dr-Bean:
> The packages does not have any configuration to keep, so just kill the old one and install the new one from scratch should be fine. 

Sure. As long as it's retained during (a future) package upgrade.

>How's you workflow here? New fork & PR?

Probably the easiest, yes. Just make sure you have an up-to-date local fork (update with a `git pull` or something, or rebase on master afterwards)

As for firewall rules, you already have a working example: https://github.com/SynoCommunity/spksrc/blob/master/spk/shairport-sync/src/shairport-sync.sc, the only change is adding the UDP range.
Take a look at the Synology development guide. I'd put the range at 10 ports, starting from the same port as the tcp one (e.g. 8303-8313 on UDP, _providing_ it's not a used range for other packages. See the [wiki article on SynoCommunity ports](https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports), and the [Synology one](https://github.com/SynoCommunity/spksrc/wiki/Synology-Used-Ports))

Issue #1031, opened 2014-06-01
SABNZBD Cannot create directory


Flynsarmy:
It seems a frenshly installed SABNZBD from the newly working Synocommunity repo on a DS1813+ doesn't work correctly. When opening SAB I get warnings

>  ERROR: Cannot create directory /volume2/path/to/downloads

The path is owned by admin:users and has the permissions drwxrw----. Is the SABNDB app not part of the users group? I notice when clicking Browse in Config - Folders - Temporary Download Folder it can see /volume2/files (owned by root:root - created with Control Panel - Shared Folder) but can't see any subfolders (all owned by admin:users).

wickedwally:
Hi everyone,

First and foremost, thank you to all the people responsible for helping to bring Sabnzbd and all the various other goodies to Synology.  Special thanks to DrBean, & Diaoul who's names have come up across many forums regarding Synology.

Although this topic has been closed, it seems the most relevant to the issue I am having; Sab and CP not being able to create or view folders through their respective GUI despite being able to see them in DSM GUI.  After reading a number of posts, I believe it is an issue that has to do with permissions.

I am running 1815+ (DSM 5.1) and have previously been able to install Sab, CP, & SB on a 1511+.  Although I ran into a few issues with the 1511, they have been relatively trivial.

Admittedly, I am a newb so please excuse any rudimentary questions and statements but I have a strong will and a desire to learn.  To date, I have relied mostly on the DSM GUI and have had VERY little experience with ssh.

Having said that, I wanted to turn to you all before attempting to ssh into my 1815 in fear that I may cause more damage than good.

I appreciate any direction and guidance you may be able to provide!

Diaoul:
Try to set permissions for the users group in DSM to the problematic
directories.

To give a little background, Synology changed their default setting for
their shares with DSM 5.0 and did not provide packagers a way to circumvent
permission issues. We have some pending work with dedicated group creation
for Syncthing, that's still a workaround but it works. I'm still waiting
for a response from Synology that doesn't make sabnzbd and others run as
root for obvious security reasons.

wickedwally:
Thanks Diaoul!  I deleted my post because I ended up trying to SSH into the Synology and i chmod 777 /volume1/downloads/

Although I am glad that works, I'm going to try and figure out how to set up a group for SAB, SB, CP, etc so that only those programs will have RW access to necessary folders.  It just seems more secure that way.

Heartfelt thank you again for your reply!

[...]

Issue #775, opened 2013-12-31
SickBeard: Permissions on autoProcessTV.cfg


albachten:
Because SABnzbd and SickBeard use different accounts, SABnzbd cannot read the autoProcessTV.cfg file (mode 600)  It would be nice if this file was installed mode 644 instead:

SickBeard version: 20130103-5

$ cd /usr/local/sickbeard/share/SickBeard/autoProcessTV
$ ls -l autoProcessTV.cfg
-rw-------    1 sickbear root            78 Jan  3  2013 autoProcessTV.cfg
$ chmod 644 autoProcessTV.cfg
$ ls -l autoProcessTV.cfg
-rw-r--r--    1 sickbear root            78 Jan  3  2013 autoProcessTV.cfg
##

PR #1618, opened 2015-03-28
added spk for SickRage

cytec:
took a while but i finally managed to make a PR...


spk/sickrage/src/sickrage.sc
```diff
@@ -0,0 +1,6 @@
+[sickbeard-custom]
```
Diaoul: sickrage here

spk/sickrage/src/wizard/install_uifile
```diff
@@ -0,0 +1,22 @@
+[{
+    "step_title": "Choose your fork",
+    "items": [{
+        "type": "textfield",
+        "desc": "Configure the fork to use",
```
Diaoul: Not sure there are a lot of forks out there, maybe add something like "skip this to install default sickrage"

spk/sickrage/src/wizard/install_uifile_enu
```diff
@@ -0,0 +1,22 @@
+[{
+    "step_title": "Choose your fork",
+    "items": [{
+        "type": "textfield",
+        "desc": "Configure the fork to use",
+        "subitems": [{
+            "key": "wizard_fork_url",
+            "desc": "Fork URL",
+            "defaultValue": "git://github.com/cytec/SickRage.git",
```
Diaoul: why cytec here and SiCKRAGETV elsewhere?

spk/sickrage/src/wizard/install_uifile_fre
```diff
@@ -0,0 +1,22 @@
+[{
+    "step_title": "Choisissez votre fork",
+    "items": [{
+        "type": "textfield",
+        "desc": "Paramètres du fork à utiliser",
+        "subitems": [{
+            "key": "wizard_fork_url",
+            "desc": "URL du fork",
+            "defaultValue": "git://github.com/cytec/SickRage.git",
```
Diaoul: Same

spk/sickrage/src/dsm-control.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PYTHON_DIR="/usr/local/python"
+GIT_DIR="/usr/local/git"
+PATH="${INSTALL_DIR}/bin:${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:${GIT_DIR}/bin:${PATH}"
+USER="sickrage"
+PYTHON="${INSTALL_DIR}/env/bin/python"
+GIT="${GIT_DIR}/bin/git"
+SICKBEARD="${INSTALL_DIR}/var/SickRage/SickBeard.py"
```
Diaoul: Shouldn't that be SICKRAGE?

spk/sickrage/src/wizard/install_uifile_enu
cytec: missed those files...

cytec:
@Diaoul updated, i didn't change the french wizzard file as i don't speek french... should i delete it or leave it as it is?

Diaoul:
Add " (suivant pour installer le sickrage par défaut)"

neoatomic:
@cytec Thanks for adding. :)

For Dutch :

```
install_uifile_nld
```

```
[{
    "step_title": "Installatie SickRage",
    "items": [{
        "type": "textfield",
        "desc": "Klik op volgende om SickRage te installeren. (Of vul een alternative URL in.)",
        "subitems": [{
            "key": "wizard_fork_url",
            "desc": "Fork URL",
            "defaultValue": "git://github.com/SiCKRAGETV/SickRage.git",
            "validator": {
                "allowBlank": false
            }
        }, {
            "key": "wizard_fork_branch",
            "desc": "Fork branch",
            "defaultValue": "master",
            "validator": {
                "allowBlank": false
            }
        ]
    }]
}]
```

spk/sickrage/Makefile
```diff
@@ -0,0 +1,46 @@
...
+SPK_VERS = $(shell date +%Y%m%d)
+SPK_REV = 1
+SPK_ICON = src/${SPK_NAME}.png
+DSM_UI_DIR = app
+
+DEPENDS  =
+SPK_DEPENDS = "python>2.7.3-3:git"
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = Video File Manager for TV Shows, It watches for new episodes of your favorite shows and when they are posted it does its magic.
```
Dr-Bean: Use a `.` instead of a `,` there: [..]TV Shows. It watches[..]

spk/sickrage/src/dsm-control.sh
```diff
@@ -0,0 +1,87 @@
...
+GIT="${GIT_DIR}/bin/git"
+SICKRAGE="${INSTALL_DIR}/var/SickRage/SickBeard.py"
+CFG_FILE="${INSTALL_DIR}/var/config.ini"
+PID_FILE="${INSTALL_DIR}/var/sickrage.pid"
+LOG_FILE="${INSTALL_DIR}/var/Logs/sickrage.log"
+
+
+start_daemon ()
+{
+    su - ${USER} -c "HOME=${INSTALL_DIR}/var PATH=${PATH} ${PYTHON} ${SICKRAGE} --daemon --pidfile ${PID_FILE} --config ${CFG_FILE} --datadir ${INSTALL_DIR}/var/"
```
Dr-Bean: Remove the `-` after `su`.

spk/sickrage/src/wizard/install_uifile_nld
```diff
@@ -0,0 +1,22 @@
+[{
+    "step_title": "Installatie SickRage",
+    "items": [{
+        "type": "textfield",
+        "desc": "Klik op volgende om SickRage te installeren. (Of vul een alternative URL in.)",
```
Dr-Bean: `"desc": "Klik op volgende om SickRage te installeren. (Of vul een alternatieve URL in.)",`

Dr-Bean:
How about using an `sc-sickrage` group here, instead of the `users` group, similar to `sc-syncthing`? /cc @Diaoul.
(note: we might want to put that in the wizard/set up a FAQ entry on that subject)

cytec:
@Dr-Bean i've fixed your comments. about that `sc-sickrage` group. i can add that but i don't know if that's the best solution because you'll end up with like hundreds of package-specific groups... for stuff like sickrage, sickbeard, sabnzb, transmission etc how about adding a `sc-downloaders` group or something like that? so that we can group a few packages together... most likely if you'll provied sabnzb access to your directory you also want to provide access for sickbeard i think? any thoughts?

Dr-Bean:
I agree with that point of view, in fact, I've brought it up before, here: https://github.com/SynoCommunity/spksrc/issues/1031#issuecomment-46164982.
Maybe split off the discussion in a separate issue (although I think we have an open issue discussing the exact same thing), but I brought it up here because if there's ever a time to easily implement it, it's before the package is published...

Dr-Bean:
@cytec Any idea if https://github.com/SynoCommunity/spksrc/issues/775 is still something we need to concern ourselves with?

Diaoul:
I prefer fine grained permissions but we're using groups (not users) so maybe that makes sense to have a more generic "group" for similar applications.
I bring this issue to Synology in september last year but haven't got any real feedback on that issue since despite my numerous follow ups. I'll tell you if I have any news on that side.

Dr-Bean:
Well, then I vote we go ahead with implementing the group approach. We can revisit the approach when Synology chooses to implement the application user solution (who knows how long that might be...and it probably won't have any form of backward compatibility). In the meantime, the group approach should lessen the issues people are having with setting permissions.

So, as per the proposal @Diaoul mentioned in the other issue: 
- Use `sc_download` for downloaders such as SABnzbd, transmission, rutorrent, etc
- Use `sc_media` for Sickbeard(-derivatives), CouchPotato, Sonarr, etc.
- Unrelated packages that only cover their own setup: one group per package (e.g. Syncthing's approach)

Some details might not have been worked out yet, e.g. we need to make sure that the group is not removed on uninstall, unless it's empty etc. @cytec, are you up for that?

Diaoul:
We use `-` for [Syncthing](https://github.com/SynoCommunity/spksrc/blob/master/spk/syncthing/src/installer.sh#L17) so let's stick to that.

cytec:
@Dr-Bean yup i'll take a look at that when i get some time. As @Diaoul said id stick to `-` for the group names as syncthing currently uses `sc-syncthing`...

also if i remember correctly `syncthing` already checks if the group is empty on uninstall and only removes the group if no users are left

gkroone:
a bit offtoopic, but while discussing group permissions here, I'd like to make another suggestion to give some thought.

I see multiple packages now adding support for USB serial devices, and each package handles the permission setting of these devices on it's own, but all I've seen so far is that they set the permission to 666 or even 777?? for /dev/ttyUSB*, which seems to me like it's insecure. 

I know Ubuntu has a default group called "dialout" which gets added to all the serial ports, and users needing access to such ports get added to group "dialout". woudl it be possible and feasible to do something similar on DSM?

Dr-Bean:
Oh right, the underscore was an honest mistake, I didn't even notice we switched. Dash is fine obviously.
@cytec, looks good to me. The description: I don't have anything better, so let's start with that. We can always change it later.
@gkroone, could you open a new issue for that, where you point out the applicable packages? It's offtopic for this PR, so tracking progress for it might get lost.

cytec:
@Dr-Bean yeah i've renamed the functions to `syno_group_create` and `syno_group_remove` i think that describes pretty good what it does, and one can simply copy paste the functions over to new SPK's and just change the `SYNO_GROUP` and `SYNO_GROUP_DESC` values

spk/sickrage/src/installer.sh
```diff
@@ -0,0 +1,139 @@
...
+
+SERVICETOOL="/usr/syno/bin/servicetool"
+FWPORTS="/var/packages/${PACKAGE}/scripts/${PACKAGE}.sc"
+
+SYNO_GROUP="sc-media"
+SYNO_GROUP_DESC="SynoCommunity's media related group"
+
+syno_group_create ()
+{
+    # Create sync group (Does nothing when sync group already exists)
```
Diaoul: That's not sync group here ;)

Diaoul:
A few changes in the comments I suggest you standardize and remove any reference to the name of the group in the comments so it's "portable".

spk/sickrage/src/installer.sh
cytec: damn, thought i got all of those changed to syno group...

cytec:
@Diaoul fixed

Dr-Bean:
Should be good to go, right?

Dr-Bean:
..and to be complete, `sickrage 20150526-1` is published by cytec. Missing a beta tag, but I guess it's too late now ;)

neoatomic:
@Dr-Bean Can confirm it works fine. :)
(However had one small problem, i forget that i already had a user called "sickrage" on the Synology. So got permission errors after the install, but removing the account and reinstalling the spk solved that.)

Thanks to all who made this possible.

Dr-Bean:
@neoatomic Thanks for the feedback, sounds good. Not that we were expecting a lot of issues anyway, plenty of experience with the current Sickbeard packages.

We should probably keep in mind to add a check if users exist and halt the package installation if so. Especially once Synology adds application users it might get more interesting to add that in.

neoatomic:
@cytec @Dr-Bean Could we change the below line: 
https://github.com/SynoCommunity/spksrc/blob/master/spk/sickrage/src/installer.sh#L68 to
`${GIT} clone --recursive -q -b ${wizard_fork_branch:=master} ${wizard_fork_url:=git://github.com/SiCKRAGETV/SickRage.git} ${INSTALL_DIR}/var/SickRage`

The ` --recursive` will allow the download of submodules that where implemented in SickRage. (added NZBtoMedia Repo in the contrub folder.)

Dr-Bean:
Sure that should work for new installs (and when we release a package update. Will you open a PR?

Seeing the linked discussion, you'd probably need a `git submodule update --init --recursive` or something to that effect for future SR updates.

Dr-Bean:
@neoatomic I saw your edit. Note that if the solution still includes working with submodules, we probably will still need to use --recursive, otherwise new installations won't init the submodules. Or will you force a submodule init the first time SR runs? 
Anyway, let us know if we need to make changes.

neoatomic:
Thought we had a solution by using --recursive in the updater but it doesn't seem to work (at least for me) on my Synology. So it still might be necessary to add it to the spk/installer.
Will update this topic as soon as we know more.

Dr-Bean:
Please open a new issue then, that makes the discussion easier to find :)

Fwiw, did a quick test just now (stop SR, `git reset --hard some_old_commit`, verify versionChecker is using `submodule update --init --recursive --force`, remove versionChecker.pyc, start SR, trigger manual update), and I've got content in `contrib/nzbToMedia`.

But, as I said, unless SR forces `submodule update --init` on the very first run of SR, we'll still need to use `--recursive` to cater for people who start off with a fresh install of the package (and that applies to other packagers too, not just SynoCommunity).

Issue #1628, opened 2015-04-02
Syncthing doesn't work after upgrade


cie007:
Hello 

I've upgraded syncthing via package center (v 0.9.4.2 to 0.10.30) and when I click on run, it won't run. 
I't has the status "stopped"
I can't find any log what is wrong so I can't tell more. 

I have uninstalled and installed again, but it won't execute. 

via putty this is what I get:
ds212> "/var/packages/syncthing/scripts/start-stop-status" start
Starting Syncthing ...
ds212> "/var/packages/syncthing/scripts/start-stop-status" status
Syncthing is not running

Is it possible to turn back the previuos version? (v0.9.4.2)

edit: Added Putty

Dr-Bean:
@cytec ARMv5 is broken. The actual error is this: `runtime: this CPU has no floating point hardware, so it cannot run this GOARM=6 binary. Recompile using GOARM=5.`
https://github.com/golang/go/wiki/GoArm has some details on that.

cie007:
A question, Do I have to move this thread to #1011  ?

Dr-Bean:
Nah, it's more likely that #1011 will get closed in the near future, I actually considered doing that after publishing the current package, but forgot to do so.

sdkent:
@Dr-Bean If this helps, I have the same error message on a DS413j with a MARVELL Kirkwood 88F6282 (ARMv5TE), although I had the same behavior before the upgrade. (I didn't check for the floating point hardware error message with the older version, but syncthing never started.)  Is the problem something as simple as the GOARM=5  flag wasn't set?

sebtic:
I found a solution waiting for an update. Just replace the binary at /var/packages/syncthing/target/bin/ by the official binary from https://github.com/syncthing/syncthing/releases/download/v0.10.30/syncthing-linux-arm-v0.10.30.tar.gz. It works like a charm on my DS413j.

cie007:
It works!!!!!!!!! Thank you very much.

sdkent:
Hey, thank you very much, @sebtic !  It worked for me, too.

gpopesc:
thank you. it works for me either.

cytec:
@Dr-Bean thanks i'll have a look at that tomorrow

PR #1631, opened 2015-04-07
fix syncthing arm compile issues

cytec:
this should fix https://github.com/SynoCommunity/spksrc/issues/1628

### DIFF
--- a/cross/syncthing/Makefile
+++ b/cross/syncthing/Makefile
@@ -20,8 +20,13 @@ GOOS = linux
 STATIC_BINARIES = 0
 
 # Define SYNCTHING_ARCH as per SyncThing standards
-ifeq ($(findstring $(ARCH),armada370 armadaxp armada375 alpine 88f5281 88f6281 comcerto2k),$(ARCH))
+ifeq ($(findstring $(ARCH),88f5281 88f6281),$(ARCH))
 SYNCTHING_ARCH = arm
+ENV += GOARM=5
+endif
+ifeq ($(findstring $(ARCH),armada370 armadaxp armada375 alpine comcerto2k),$(ARCH))
+SYNCTHING_ARCH = arm
+ENV += GOARM=7
 endif
 ifeq ($(findstring $(ARCH),evansport),$(ARCH))
 SYNCTHING_ARCH = 386
--- a/cross/syncthing/Makefile
+++ b/cross/syncthing/Makefile
@@ -22,6 +22,7 @@ STATIC_BINARIES = 0
 # Define SYNCTHING_ARCH as per SyncThing standards
 ifeq ($(findstring $(ARCH),armada370 armadaxp armada375 alpine 88f5281 88f6281 comcerto2k),$(ARCH))
 SYNCTHING_ARCH = arm
+ENV += GOARM=5
 endif
 ifeq ($(findstring $(ARCH),evansport),$(ARCH))
 SYNCTHING_ARCH = 386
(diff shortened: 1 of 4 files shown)


Dr-Bean:
With this patch, I get an `illegal instruction [core dump]` when running `/usr/local/syncthing/bin/syncthing`.

If I'm not mistaken, the environment variable should be used when compiling go for linux/arm, not for Syncthing itself (although it wouldn't hurt)
The approach used by syncthing itself might be helpful: https://github.com/syncthing/build/blob/master/Dockerfile. The Syncthing issue tracker has had some discussions on ARM arches as well.

cytec:
so if i got everything right we'll have to set `GOARM=5` in `native/go` to support armv5... that should work on armv6 and armv7 too but will use softfloat instead of the supported hardware floating point (at least for armv6 and armv7)

cytec:
@Dr-Bean i've reverted the `cross/syncthing` changes and patched native go instead, can you test if it is working now?

Dr-Bean:
Ok, did a bit of testing. It looks like both the previous patch and this one need to be combined, as well as not stripping the binary file.

My suggestion: add `ENV += GOARM=5` to the cross package near L52 : it should not affect non-ARM arches anyway. We can't use `GOARM=7` for other ARM arches right now, because that would mean setting up a new linux/arm build, and the current approach does not support more than one.

Two notes on that: 
- Maybe there's a way around that, by setting up a cross/go package and building go per arch, or something like that. If not, we'd be limiting all go packages for ARM arches to ARMv5. Since Syncthing is the only package using go atm, we're not in a rush with that, but still, that would be nice to have.
- The native build needs a fix somewhere to specify the host/default arch, because you can't build native/go from the SPK when starting off with an ARM arch. Probably passing on an environment variable for x86 would help with that.

As for not stripping the binary, use `rsc:` instead of `bin:` for the PLIST.

Dr-Bean:
...although, the web interface doesn't seem to want to work for some weird reason. I'll come back to this later.

Dr-Bean:
Ok, that's on my end: If you manually start Syncthing without the proper arguments, it'll overwrite the predefined config.xml with the defaults.
So, my earlier comments should fix the arm build.

cytec:
@Dr-Bean so... revert the reverted patch again? (include `ENV += GOARM=5` in `cross/syncthing`) and change `bin:` to `rsc:`?

Dr-Bean:
The thing is, the package (for ARMv5 that is) is still broken with those changes. I haven't looked at it further after discovering it's stuck in a loop of stopping/starting over and over. 
There's an error log, but I assume that the issue occurs due to the way we compile Go and Syncthing, so the log wouldn't seem very helpful.

I was considering copying the whole process syncthing uses (use https://github.com/calmh/gonative for starters, and compare their Docker repo with your approach), but I'll probably won't get to that very soon.

As for the 2nd note, what I mean is this: do a `make clean` on `native/go`, then go to `spk/syncthing` and compile a non-x86 arch, e.g. `make arch-88f6281-5.1`. 
You'll see `make` attempt to build `native/go` (which it should) but that does not succeed. Usually that occurs when incorrect makeflags are passed on. A fix in a similar vein has recently been made to the `mono` package: https://github.com/SynoCommunity/spksrc/commit/e8a9af782dc0699ec841b3fac52427fad90ff676, although Go would probably use environment variables.
We should probably clean all flags when starting a native build or something like that, but that's very low prio.

Diaoul:
So this is still broken right? Does it at least work for some archs?

Dr-Bean:
Works for all but ARMv5 arches. Note that Syncthing does provide a working binary for ARMv5, so it's something on our side.

Dr-Bean:
This PR does seem to work in combination with Syncthing 0.11.10 on ARMv5, so I think we'll go ahead and merge this while updating Syncthing too.

Issue #1623, opened 2015-03-30
a few build/depend questions


cytec:
as i just try to compile chromaprint for all supported archs i was wondering if there is a solution for some things i came across:

1. bundle arches by type... currently in some make files there is something like: `ifeq ($(findstring $(ARCH),xyz,$(ARCH))` wouldn't it be nice to have this in a global place? so that not all make files have to be updated everytime a new arch pops up?
2. there is `make all-archs-latest` but is there something like `make all-archs-5.1` to only build with that specific version of the toolchains?
3. is there any way i can depend on a minimum version of the toolchain in compile time? some stuff seems to need at least 5.x to compile and `make all-archs-latest` tries to compile with 4.x too, which fails

@Diaoul @Dr-Bean maybe you can enlighten me a little bit on those issues :)

cytec:
to clear 1. up a little bit:

in `debian-chroot` or `syncthing` there are if's like this:

	ifeq ($(findstring $(ARCH),88f6281),$(ARCH))
		DEBIAN_ARCH = armel
	endif
	ifeq ($(findstring $(ARCH),armada370 armadaxp armada375 alpine),$(ARCH))
		DEBIAN_ARCH = armhf
	endif
	ifeq ($(findstring $(ARCH),bromolow cedarview evansport x86 avoton x64),$(ARCH))
		DEBIAN_ARCH = i386
	endif
	ifeq ($(findstring $(ARCH),qoriq),$(ARCH))
		DEBIAN_ARCH = powerpc
	endif
	ifeq ($(strip $(DEBIAN_ARCH)),)
		$(error Arch $(ARCH) not supported)
	endif

and i was thinking making it something like this:

	ifeq ($(findstring $(ARCH),$ARM_ARCHES),$(ARCH))
		DEBIAN_ARCH = armel
	endif
	ifeq ($(findstring $(ARCH),$ARM7_ARCHES),$(ARCH))
		DEBIAN_ARCH = armhf
	endif
	ifeq ($(findstring $(ARCH),$386_ARCHES),$(ARCH))
		DEBIAN_ARCH = i386
	endif
	...

so that `$ARM7_ARCHES` contains `armada370 armadaxp armada375 alpine` in a more global place and must only be edited once... also that would clear things up a little as i had some struggle to find out where to put `alpine` and had to look for a spk already referencing it

Dr-Bean:
[...]

PR #1635, opened 2015-04-09
add vars for available arches

cytec:
like discussed here: https://github.com/SynoCommunity/spksrc/issues/1623


Diaoul:
Maybe use them somewhere too?

cytec:
originally was planning to use those for `cross/syncthing` and `cross/ffmpeg`

PR #1649, opened 2015-04-13
added dsm groups for various packages

cytec:
as discussed here: https://github.com/SynoCommunity/spksrc/pull/1618

affected packages:

`sc-media`:

	couchpotatoserver
	couchpotatoserver-custom
	flexget
	headphones
	headphones-custom
	lazylibrarian
	sickbeard
	sickbeard-custom
	sonarr

`sc-download`:

	deluge
	nzbget
	nzbget-testing
	transmission
	sabnzb
	rutorrent

@Dr-Bean @Diaoul any thoughts?

EDIT: Sickrage changes will come in SickRage PR :)

### DIFF
--- a/spk/couchpotatoserver-custom/src/installer.sh
+++ b/spk/couchpotatoserver-custom/src/installer.sh
@@ -115,8 +115,6 @@ preupgrade ()
     # Stop the package
     ${SSS} stop > /dev/null
 
-    syno_group_create
-
     # Remove auto-updater stuff so it doesn't get confused
     rm -fr ${INSTALL_DIR}/var/cache/updates/
 
--- a/spk/couchpotatoserver-custom/src/installer.sh
+++ b/spk/couchpotatoserver-custom/src/installer.sh
@@ -19,6 +19,32 @@ TMP_DIR="${SYNOPKG_PKGDEST}/../../@tmp"
 SERVICETOOL="/usr/syno/bin/servicetool"
 FWPORTS="/var/packages/${PACKAGE}/scripts/${PACKAGE}.sc"
 
+SYNO_GROUP="sc-media"
+SYNO_GROUP_DESC="SynoCommunity's media related group"
+
+syno_group_create ()
+{
+    # Create sync group (Does nothing when sync group already exists)
+    synogroup --add ${SYNO_GROUP} ${USER} > /dev/null
+    # Set description of the sync group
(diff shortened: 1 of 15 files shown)


Dr-Bean:
Seems to me the addition of `syno_group_create` to `preupgrade` isn't needed, right?

Dr-Bean:
...and the descriptions still say `sync group` ;) Other than that, quick turnaround :)

cytec:
@Dr-Bean idk if `postinst` is executed when the user updates an spk so i thought i'd put it in there as well

Dr-Bean:
A package essentially is always uninstalled/reinstalled during an update. `preupgrade` allows you to do stuff before the uninstall, e.g. backup files.
Looked it up: page 35 on my copy of the developers guide shows which steps are done during a package upgrade.

cytec:
so should i remove `syo_create_group` from `preupgrade` then? :)

Dr-Bean:
Well, my experience would say yes, remove it. Otoh, the dev guide has been wrong before, so the best thing to do is test it :P 
I might give it a go during the week: reviewing code online is easy enough to do in between other things, testing stuff less so ;)

cytec:
ok, so i'll leave it in there till you tested it, removing it if it's not needed is done quicker then readding it ;)

Diaoul:
@Dr-Bean is right, *inst and *uninst are both executed during upgrade.

hurik:
Can i also add this to btsync now?

Dr-Bean:
@cytec It works without the `preupgrade` addition ;) 
In the interest of releasing new packages, it would need a version bump and changelog (`Group sc-xyz added to allow setting permissions via DSM GUI` or something to that effect)

Question: should we add an install wizard to explain how to use that group? We don't necessarily need an upgrade wizard imo, but that can have the same content if we do.

@hurik Yes, if you can reuse the code in this PR, go ahead and create a PR. Maybe wait until we've fleshed out the implementation details (or, if you have an answer to my question above, go ahead)

Diaoul:
Applications users will be part of next major release of DSM from what I heard. :ear:

Dr-Bean:
With the PR for NZBget (#1728) being ready to merge, I think I'll merge this PR in at the same time (and add a mention of it in the NZBget changelog). We can do the other packages little-by-little, depending on time etc.

When the application users are available in DSM6, I believe a working migration path is to add the newly created application users to the same group, preserving existing permissions for existing installations.

Diaoul:
Seems good

Issue #1654, opened 2015-04-17
Help needed with building an spk for home assistant (python3-based app)


wind-rider:
I'm trying to build an spk for [Home Assistant](https://home-assistant.io/) / https://github.com/balloob/home-assistant.git. It is a promising home automation server written in Python 3.

I've created a basic setup for it at f013a38be5e2e3ec214343ddc608634adabf0a5a.

The dependencies build now, but it starts the "stripping binaries" build step based on the PLIST from the work directory instead of from the spk/homeassistant directory. This causes spksrc trying to find binaries that are not built.

Can anybody give me some hints in the right direction?

Dr-Bean:
Your description isn't completely clear to me, but I believe you might be referring to the difference between [DEPENDS](https://github.com/SynoCommunity/spksrc/blob/03f583a3db38b80021aa9e7c5af641ac2b9fdf0b/mk/spksrc.depend.mk#L9) and [BUILD_DEPENDS](https://github.com/SynoCommunity/spksrc/blob/03f583a3db38b80021aa9e7c5af641ac2b9fdf0b/mk/spksrc.depend.mk#L12).

As an example, the Mercurial package uses that approach: https://github.com/SynoCommunity/spksrc/blob/03f583a3db38b80021aa9e7c5af641ac2b9fdf0b/spk/mercurial/Makefile#L7. 
Fwiw, try to replicate the approach of `cross/mercurial` with your netifaces module: Instead of adding `cross/netifaces` as a module in the Python SPK, make that a (cross-compiled) wheel, and add the resulting wheel to your SPK in the same way as `cross/mercurial` is added to `spk/mercurial`.

wind-rider:
@Dr-Bean you're right, I confused DEPENDS and BUILD_DEPENDS. This helped a lot, thanks!

After changing that, I could build the package (including PyYaml and netifaces as cross-compiled wheels), but adding pyephem as a wheel didn't go well.

I took my netifaces wheel as an example, but I think there is something wrong in the pyephem Makefile that causes spksrc not to understand what it should do. The changes I made are at 31a38b6fa7eb263c287d51f010bc1d6aba8ef1fa / https://github.com/SynoCommunity/spksrc/compare/master...wind-rider:homeassistant

It doesn't matter whether I try to quickly build pyephem from its cross folder or by referencing it from my spk folder.

Could you explain to me what's going wrong?

See the log below:

``````````````
[...]

Issue #1647, opened 2015-04-12
virtualenv in python3-package does not work?


wind-rider:
I'm trying to build a package based on SynoCommunity's python3 package.

Running it does not work well. It seems that the virtualenv directory is not created.

When trying to run virtualenv manually, I get:

```````
-ash: /usr/local/python3/bin/virtualenv-3.4: not found
``````

I also tried the virtualenv (without -3.4 suffix), but it gives the same result.

Dr-Bean:
Confirmed. Not sure why that is right now.
This might be a workaround: `/usr/local/python3/bin/python -m pip install virtualenv`, then try again.

wind-rider:
Thank you for the quick reply, I'll look at it next weekend :) this week is a busy one for me

Diaoul:
Is this something we can fix? Or an upstream bug/feature?

wind-rider:
I don't know, building with python is new for me. Maybe @Dr-Bean knows?

Dr-Bean:
The shebangs of `/usr/local/python3/bin/virtualenv` and `/usr/local/python3/bin/pip` are wrong.

I don't think it's a generic upstream issue. It might be related to https://www.python.org/dev/peps/pep-0453: I believe PEP453 is responsible for placing `/usr/local/python3/bin/pip3.4` and `/usr/local/python3/bin/virtualenv-3.4`, and look at those shebangs.
Otherwise, I haven't done any troubleshooting, so no idea what's up with that.

Diaoul:
Maybe [this](https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.python-module.mk#L33) is wrong?

Dr-Bean:
Yes, after rethinking my previous comment, the PEP can't be related. So, it's likely something went wrong with processing the PLIST entries.
Seems there are two seperate issues then?

- `@PYTHON_VERSION@` isn't processed correctly. That accounts for the unchanged shebang of `pip3.4` and `virtualenv-3.4` files
- The incorrect shebang of `pip` and `virtualenv` seems to be caused by `PYTHON_INTERPRETER` being set to `python` instead of `python3.4` (or a missing `python` symlink to `python3.4`, that'd work too)

Dr-Bean:
My second suggestion above is correct: https://github.com/SynoCommunity/spksrc/blob/master/cross/python3/Makefile#L76. Did Python 3.4 remove the `python` symlink then, or what happened there?

[...]

PR #1655, opened 2015-04-19
Homeassistant

wind-rider:
Basic working version
- Start / stop and log files work
- Pyephem and netifaces are packaged as cross-compiled wheels
- Some optional cross-compilation dependencies are not yet compiled:
 - pychromecast>=0.5 disabled since it currently does not work with Chromecast api
 - pyuserinput>=0.1.9 is excluded for the moment
 - tellcore-py>=1.0.4 is excluded for the moment
 - pydispatcher>=2.0.5 is excluded for the moment
 - psutil>=2.2.1 is excluded for the moment
- pip34 installs the rest of the requirements during installation (spksrc.common.mk in the build system currently contains hardcoded pip for python2).


wind-rider:
The python-libnmap needs the nmap executable, which is not yet included.

spk/homeassistant/src/installer.sh
```diff
@@ -0,0 +1,92 @@
...
+DNAME="Home Assistant"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+SSS="/var/packages/${PACKAGE}/scripts/start-stop-status"
+PYTHON_DIR="/usr/local/python3"
+PATH="${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:${PATH}"
+USER="homeassistant"
+GROUP="users"
+VIRTUALENV="${PYTHON_DIR}/bin/python3 -m virtualenv"			#was "${PYTHON_DIR}/bin/virtualenv-3.4"
```
Diaoul: Why not virtualenv?

spk/homeassistant/src/installer.sh
```diff
@@ -0,0 +1,92 @@
...
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # Create a Python virtualenv
+    ${VIRTUALENV} --system-site-packages ${INSTALL_DIR}/env > /dev/null
+
+    # Install the wheels
+    ${INSTALL_DIR}/env/bin/pip3.4 install --use-wheel --no-deps --force-reinstall -f ${INSTALL_DIR}/share/wheelhouse -r ${INSTALL_DIR}/share/wheelhouse/requirements.txt > /dev/null 2>&1
```
Diaoul: If you made the virtualenv with 3.4 binary you should be able to use `pip` here.

Diaoul:
Please make this package installable without internet connection. The goal is to have a reproducible installation process, if pip downloads stuff from internet at install time it may result in differences between install if a module was updated for example.

spk/homeassistant/src/installer.sh
wind-rider: Because of https://github.com/SynoCommunity/spksrc/issues/1647

spk/homeassistant/src/installer.sh
wind-rider: I think this was the same as above (https://github.com/SynoCommunity/spksrc/issues/1647)

wind-rider:
Thank you for your feedback.

Currently pip is hardcoded in spksrc.common.mk. This results in python2 packages being downloaded during the packaging process. @Dr-Bean noted (https://github.com/SynoCommunity/spksrc/issues/1654#issuecomment-94267324) that the current spksrc wheel mk-files are targeted at python2. In the installer.sh file I had more freedom to specify that pip34 should be used.

Is there a way to make that more flexible during the building / packaging process?

wind-rider:
I fixed your comments except for the `--no-index`-parameter, since issue #1658 should be fixed / explained for that.

spk/homeassistant/src/requirements.txt
```diff
@@ -0,0 +1,43 @@
...
+pushbullet.py>=0.7.1
+
+# thermostat.nest
+python-nest>=2.1
+
+# z-wave
+# pydispatcher>=2.0.5 is excluded for the moment
+
+# sensor.systemmonitor 
+# psutil>=2.2.1 is excluded for the moment
```
Dr-Bean: Fyi, psutil also contains C-code. If you want to include it in the package, cross-compile it.

wind-rider:
I think this set of commits is a good basic version. What do you think?

There are a few dependencies left out:
- pychromecast>=0.5 disabled since it currently does not work with Chromecast v2 api
- pyuserinput>=0.1.9 since there is no keyboard on a NAS
- tellcore-py>=1.0.4 (for tellstick) since I have no hardware to test it
- pydispatcher>=2.0.5 (for zwave) since I have no hardware to test it
- transmissionrpc>=0.11 (bittorrent status) - not done yet.

cross/netifaces/PLIST
```diff
@@ -0,0 +1 @@
+
```
Dr-Bean: PLIST is not needed with BUILD_DEPENDS

cross/netifaces/digests
```diff
@@ -0,0 +1 @@
+netifaces-0.10.4.tar.gz MD5 36da76e2cfadd24cc7510c2c0012eb1e
```
Dr-Bean: Generate digests with `make digests`

cross/nmap/Makefile
```diff
@@ -0,0 +1,19 @@
...
+PKG_DIST_SITE = https://nmap.org/dist
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS = cross/libpcap cross/openssl
+
+HOMEPAGE = https://nmap.org/
+COMMENT  = Nmap ("Network Mapper") is a free and open source (license) utility for network discovery and security auditing
+LICENSE  = GPL
+
+#ncat doesn't compile well
```
Dr-Bean: Loose end?

cross/nmap/PLIST
```diff
@@ -0,0 +1,3 @@
+bin:bin/ndiff
+bin:bin/nmap
+bin:bin/nping
```
Dr-Bean: Could use a newline

cross/psutil/PLIST
```diff
@@ -0,0 +1 @@
+
```
Dr-Bean: PLIST is not needed with BUILD_DEPENDS

cross/pyephem/PLIST
```diff
@@ -0,0 +1 @@
+
```
Dr-Bean: PLIST is not needed with BUILD_DEPENDS

cross/pyephem/digests
```diff
@@ -0,0 +1 @@
+pyephem-3.7.5.3.tar.gz MD5 46b101da152bb109b1137263b150f390
```
Dr-Bean: Create the digests with `make digests`.

spk/homeassistant/src/dsm-control.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/sh
+
+# Package
+PACKAGE="homeassistant"
+DNAME="Home Assistant"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PYTHON_DIR="/usr/local/python3"
+PYTHONPATH="${INSTALL_DIR}/share/${PACKAGE}"
```
Dr-Bean: The description does not match the content here. Either correct that, or add this to PATH and use that in the start command.

spk/homeassistant/src/dsm-control.sh
```diff
@@ -0,0 +1,89 @@
...
+PATH="${INSTALL_DIR}/bin:${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:${PATH}"
+USER="homeassistant"
+PYTHON="${INSTALL_DIR}/env/bin/python3"
+
+CONFIG_DIR="${INSTALL_DIR}/etc"
```
Dr-Bean: I'd suggest to clean up L15-L24. How about (just) this instead:
```
LOG_FILE="${INSTALL_DIR}/var/home-assistant.log"
PID_FILE="${INSTALL_DIR}/var/homeassistant.pid"
RUN_ARGS="-m ${PACKAGE} -c ${INSTALL_DIR}/var --open-ui"
```

spk/homeassistant/src/dsm-control.sh
```diff
@@ -0,0 +1,89 @@
...
+
+PID_DIR="${INSTALL_DIR}/var"
+PID_FILE="${PID_DIR}/homeassistant.pid"
+
+RUN_CMD="${PYTHON}"
+RUN_ARGS="-m ${PACKAGE} -c ${CONFIG_DIR} --open-ui"
+
+start_daemon ()
+{
+    start-stop-daemon -b -o -c ${USER} -S -u ${USER} -m -p ${PID_FILE} -x env PYTHONPATH=${PYTHONPATH} ${RUN_CMD} -- ${RUN_ARGS}
```
Dr-Bean: With the comments I made above covered, the start command probably would look like this:
start-stop-daemon -b -o -c ${USER} -S -u ${USER} -m -p ${PID_FILE} -x PATH=${PATH} ${PYTHON} -- ${RUN_ARGS}

spk/homeassistant/src/dsm-control.sh
```diff
@@ -0,0 +1,89 @@
...
+}
+
+stop_daemon ()
+{
+    start-stop-daemon -o -c ${USER} -K -u ${USER} -p ${PID_FILE} -x ${RUN_CMD}
```
Dr-Bean: I don't think the `-x xyz` is need here, right?

spk/homeassistant/src/dsm-control.sh
```diff
@@ -0,0 +1,89 @@
...
+
+daemon_status ()
+{
+    start-stop-daemon -K -q -t -u ${USER} -p ${PID_FILE}
+    [ $? -eq 0 ] || return 1
+}
+
+wait_for_status ()
+{
+	counter=$2
```
Dr-Bean: Fix indents and replace tabs with spaces, throughout this file.

spk/homeassistant/src/installer.sh
```diff
@@ -0,0 +1,92 @@
...
+PACKAGE="homeassistant"
+DNAME="Home Assistant"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+SSS="/var/packages/${PACKAGE}/scripts/start-stop-status"
+PYTHON_DIR="/usr/local/python3"
+PATH="${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:${PATH}"
+USER="homeassistant"
+GROUP="users"
```
Dr-Bean: Probably should be `nobody`, there's no file access needed, right?

spk/homeassistant/src/installer.sh
```diff
@@ -0,0 +1,92 @@
...
+VIRTUALENV="${PYTHON_DIR}/bin/virtualenv"
+CONFIG_DIR="${INSTALL_DIR}/etc"
```
Dr-Bean: You could just use `${INSTALL_DIR}/var` throughout, see other packages.

spk/homeassistant/src/installer.sh
```diff
@@ -0,0 +1,92 @@
...
+    # Stop the package
+    ${SSS} stop > /dev/null
+
+    # Remove the user if uninstalling
+    if [ "${SYNOPKG_PKG_STATUS}" == "UNINSTALL" ]; then
+        delgroup ${USER} ${GROUP}
+        deluser ${USER}
+    fi
+
+    # Remove firewall config
```
Dr-Bean: Please add the firewall config during postinst too ;)

spk/python3/Makefile
```diff
@@ -9,6 +9,8 @@ DEPENDS += cross/setuptools cross/pip cross/wheel cross/virtualenv cross/six
 DEPENDS += cross/bcrypt cross/pyopenssl cross/pycrypto
 DEPENDS += cross/pycurl cross/sqlalchemy cross/lxml
 DEPENDS += cross/zope.interface cross/markupsafe
+DEPENDS += cross/pyyaml
+#DEPENDS += cross/netifaces cross/pyephem
```
Dr-Bean: Cleanup required here.

spk/python3/Makefile
```diff
@@ -9,6 +9,8 @@ DEPENDS += cross/setuptools cross/pip cross/wheel cross/virtualenv cross/six
...
```
Dr-Bean: If you're preparing for a new version by bumping SPK_REV, please add/update the CHANGELOG entry too.

Dr-Bean:
About those dependencies: are there particular reasons why those wouldn't work, even though you can't  test them yourself? A beta test is probably the best time to receive feedback, and maybe it's worth a shot just leaving them in for now.

cross/netifaces/digests
wind-rider: fixed

cross/nmap/PLIST
wind-rider: fixed

cross/pyephem/digests
wind-rider: fixed

spk/homeassistant/src/installer.sh
wind-rider: I didn't know about the nobody group, but there is a [downloader](https://home-assistant.io/components/downloader.html) component that can download from given urls to the file system. Will that still work?

spk/homeassistant/src/installer.sh
wind-rider: I wanted to keep the pid file separate from the rest of the files

spk/homeassistant/src/installer.sh
Dr-Bean: No, it probably won't then. If that's a useable option, you might want to take a look at #1649 and add that approach in as well.
How can people set that downloadpath, is it available via the interface?

spk/homeassistant/src/installer.sh
Dr-Bean: Why? It's nothing special :) If you're worried it'll be backed up when you upgrade: it won't, because when the package is stopped, the PID file is already removed.

spk/homeassistant/src/installer.sh
wind-rider: done

spk/python3/Makefile
wind-rider: done

spk/python3/Makefile
wind-rider: done

spk/homeassistant/src/dsm-control.sh
wind-rider: done

spk/homeassistant/src/dsm-control.sh
wind-rider: fixed

spk/homeassistant/src/dsm-control.sh
wind-rider: I did this because:
- I wanted separation of (config and runtime files) and the pid file.
- Home Assisant writes all config and runtime to the folder specified with -c
- PYTHONPATH is needed to let python find the homeassistant module

spk/homeassistant/src/installer.sh
wind-rider: I haven't used it myself yet, but according to the documentation it takes a download dir in the configuration file (a yaml file).

wind-rider: I added the download group logic. However I don't understand how that'll give file access if the GROUP is set to nobody?

wind-rider:
@Diaoul thank you for your thorough feedback. I fixed most of the issues.

About the dependencies - it is not only trying to cross-compile them, for example for the zwave and tellstick support there is also something with USB access involved. First of all I can't test that, and also I don't have experience with configuring that in Synology nases.

About the dsm-control.sh file - I'm not sure what is wrong with setting the PYTHONPATH.

spk/homeassistant/src/installer.sh
Dr-Bean: If you need access to the filesystem (outside of the users home dir), then use the users group. If access is not needed, use nobody. Because I didn't know about that component, I suggested to use the nobody group.

But, if you want to make that component useable, maybe best to add a way to set that downloadpath, because otherwise, people will have to dig into the config to change the path manually. That's why I asked if the path could be set via the web interface.
If you can't set it via the interface, maybe add a wizard (there's a lot of packages using the wizard to do that, take a look at couchpotato for instance)

spk/homeassistant/src/dsm-control.sh
Dr-Bean: Imo, let's stick to the approach we use for most packages. ${INSTALL_DIR}/var is the home directory for the user you created (see installer.sh).
Its normally used for config files, (sqlite) databases, log files and all that kind of stuff. There's no harm in putting a PID file in there.
Depending on how the log files are generated, sometimes it makes sense to create a ${INSTALL_DIR}/var/log directory (e.g. when those are rotated)

As for PYTHONPATH, that is usually used to specify the path to the Python executable (not to the executable you want to run). The reason I mentioned it is because you're using PYTHONPATH where others would use PATH. 
Besides that, you're calling (or should be calling) `${PYTHON} ${HOMEASSISTANT}`. It's quite likely you don't even need to specify PYTHONPATH or PATH.

spk/homeassistant/src/installer.sh
wind-rider: Having a UI would be cool, but at the moment I don't have time to pick up that extra thing (getting this basic version to work already takes quite much of my time).

At the moment all configuration is done through configuration files (scripts, setting up groups, etc) and writing a wizard for it will take lots of time, while I think that eventually the configuration will also be done through the web interface.

wind-rider:
@Dr-Bean I enabled more wheels and included their dependencies.

wind-rider:
@Dr-Bean I cleaned up the folder definitions. Setting PYTHONPATH is still needed, otherwise python3 can't find the home-assistant module folder.

I think I processed all your feedback, can it be merged now?

wind-rider:
@Diaoul @Dr-Bean could anybody have a look at this pull request to verify whether it can be merged?

spk/homeassistant/src/installer.sh
```diff
@@ -0,0 +1,125 @@
...
+    # Create user
+    adduser -h ${INSTALL_DIR}/var -g "${DNAME} User" -G ${GROUP} -s /bin/sh -S -D ${USER}
+
+    syno_group_create
+
+    # Correct the files ownership
+    chown -R ${USER}:root ${SYNOPKG_PKGDEST}
+
+    # Add firewall config
+    ${SERVICETOOL} --install-configure-file --package ${FWPORTS} >> /dev/null
```
Dr-Bean: We're missing the FWPORTS definition on L19 or so.

mk/spksrc.common.mk
```diff
@@ -14,7 +14,7 @@ MSG = echo "===> "
 RUN = cd $(WORK_DIR)/$(PKG_DIR) && env $(ENV)
 
 # Pip command
-PIP ?= pip
+PIP ?= pip #define PIP in your makefile if you need pip3
 PIP_WHEEL = $(PIP) wheel --no-use-wheel --no-deps --wheel-dir $(STAGING_DIR)/wheelhouse
```
Dr-Bean: Probably just remove the changes you made to this file, current code is fine as-is.

wind-rider: isn't it a good practice to document such options? frankly speaking I find it not easy to quickly understand all compile options that spksrc offers.

Dr-Bean: Sure. It's on the todo list to document the possible options for Makefiles, there's an open issue on the subject somewhere.

spk/homeassistant/src/installer.sh
wind-rider: Done :)

mk/spksrc.common.mk
wind-rider: What to do with that comment for now - should it be removed?

Dr-Bean: Yes please. I won't forget (that is, once I get to documenting the stuff in the first place)...;)

wind-rider: The comment is also removed

wind-rider: Great that documentation is on the todo list, by the way :)

Diaoul: spksrc is something to learn by example, getting to know the existing packages will teach you all the possibilities spksrc offers.
I agree documentation needs some love but until then, this is how to proceed :)

Dr-Bean: Yes, I agree. The documentation is never going to cover everything, as in: after you read it, you can make a package. It won't be an instruction on how to cross-compile an arbitrary piece of software. That's simply not feasible; I would end up writing a book on cross-compiling: no thanks ;) 
Realistically speaking, the starting point will be to explain the entries in a Makefile, such as content, format, purpose, and that's mostly spksrc specific stuff. We'll see where we go from there.

Back to the subject at hand: I'll probably add it to the Python/wheel wiki article, that's a sensible place to put it (https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-solutions). I was working on an update of the doc, so I'll fit it in when I get to finishing it.

Dr-Bean:
@wind-rider: I've added some fixes to your setup with https://github.com/SynoCommunity/spksrc/commit/11fec5f9b9f2228c27c85e2a3dbea9d2c607cfdc.

In addition, you probably want to do `git checkout` of some sort for homeassistant, to prevent (uncontrolled) changes to the code, (see the pytz module requirement which was added recently). Test to make sure that works correctly for the submodules, those probably need a separate checkout or something.

Also, you might want to switch over to using a cross package for homeassistant, instead of cloning homeassistant in `spk/homeassistant/Makefile`: I'm cloning 13-15MB of data for each package I'm building (20-25). Using a cross package allows that data to be cached in `/distrib`.

Dr-Bean:
`homeassistant 20150526-1` has been published on the beta channel.

wind-rider:
@Dr-Bean great, thanks for all your help (feedback, fixes, building and publishing) :) it is very much appreciated!

I'm looking into your last comment now, it sounds useful to use a cross package then. I found GateOne (https://github.com/SynoCommunity/spksrc/blob/master/cross/gateone/Makefile). It also shows how to fix the git commit hash. Is this a good example of how to do create that cross package?

Also I'm wondering how to deal with the ```--recursive``` option that is needed in the ```git clone command```. Can I just add that to https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.download.mk#L66-67 or will it have an unwanted impact?

Diaoul:
@wind-rider: I think it's safe to add it

Dr-Bean:
Agree, adding `--recursive` doesn't seem to do much if there are no submodules, so should be good. 
Anyway, the GateOne example should work, I actually meant to add it as an example. Take a look at GateOne's PLIST too, in case you notice the dir isn't included in the package.

Decipher:
Installed this, asks me for a password when I access it for the first time.... what would that be?

Decipher:
I could be wrong, but it looks like configuration has to be done manually via ssh: https://home-assistant.io/getting-started/configuration.html

Maybe this could be done via the install wizard? Or maybe I'm missing something.

Decipher:
Tried placing configuration.yaml file in multiple places and rebooting and trying, still can't log in.

Log message seems unrelated to the experience, but definitely shows an error with one of the missing dependencies noted above:

```
11:59 03-06-15 homeassistant: hass.local_api is deprecated. Use hass.config.api
11:59 03-06-15 homeassistant: hass.local_api is deprecated. Use hass.config.api
11:59 03-06-15 homeassistant.components.recorder: Found unfinished sessions
11:59 03-06-15 homeassistant.components.light.hue: Connected to Hue at 10.1.1.200 but not registered.
11:59 03-06-15 homeassistant.components.media_player.cast: Failed to import pychromecast. Did you maybe not install the 'pychromecast' dependency?
  File "/usr/local/homeassistant/share/homeassistant/homeassistant/components/media_player/cast.py", line 31, in setup_platform
    import pychromecast
ImportError: No module named 'pychromecast'
```

wind-rider:
Hi,

You're right, the configuration needs to be done via configuration.yaml using ssh.

This is currently how Home Assistant (the upstream code) works. I have no time to make such an UI for it, but if you would like to contribute to it, you would be very welcome.

The error messages you showed are nothing to worry about - the deprecated warnings don't mean that it doesn't work and the module loading error is because this module was broken at the moment the synology package was built. This results in that chromecast functionality won't work in the synology package, but it won't hurt the other functionality.

About the password - this behavior changed after I committed my code to synocommunity. I think that the default password is empty (just press login), and otherwise you can setup a password in configuration.yaml.

If it still doesn't work, then the package might need to be updated (it was built using a different git revision than I tested with before submitted the code..)

I hope you'll get a bit further with this :)

Decipher:
@wind-rider 

I don't doubt that is how it is in upstream, and I understand that you don't have time to make a UI for it, my point is that out of the box, when you install Home Assistant via the SynoCommunity it will not work, and there's no indication why this is or what to do. At the bare minimum it should put the default configuration in place and set the appropriate permissions to the file.

I played around with the file for a good 20 minutes, and I didn't manage to get access to Home Assistant. I tried with a blank password as well as the password in the configuration file to no end. I will give it another shot though and report back with my findings.

Have you managed to install this package on a fresh test environment and actually log in? Because if so, I would like to hear the steps so I can determine what I'm missing.

stvnrlly:
I'm having a different problem, seemingly related to the SSL configuration. After installation, trying to open the site gets an `Error code: ssl_error_rx_record_too_long` error.

Dr-Bean:
@stvnrlly How did you configure SSL? Because if HomeAssistant runs on HTTP, but you access it via HTTPS, you'll see that message.

stvnrlly:
@Dr-Bean I have HSTS set to force HTTPS. Is there a way to set HomeAssistant to use HTTPS instead?

Dr-Bean:
Homeassistant uses it's own webserver, not Synology's webserver, so the HSTS setting doesn't affect it.

If Homeassistant doesn't provide SSL support (and it doesn't look like an easily accessible option with a quick look), then you might also want to contact the developer (forum/bugtracker/mailing list), and see if they are willing to implement it.
An even better solution is to use a reverse proxy. HAProxy is the go-to package we provide for that.

stvnrlly:
Thanks, @Dr-Bean, I'll see what I can do.

stvnrlly:
I'm able to access Homeassistant, but can't seem to get it to read the configuration file. Is the proper location `share/homeassistant/config/configuration.yaml`? I also see that there's an existing `var/configuration.yaml`, but changing that seems to mess things up.

wind-rider:
Hi,

The one in share/homeassistant/config/ is a default one that comes from git.

The one in var/ is the one you need to change. I checked and I have no problems with that package - it correctly shows my hue lamps and I added via the configuration file a sunlight tracker.

As for me - I did a manual install because I developed the package. Afterwards I upgraded to the version distributed by synology.

stvnrlly:
Thanks, @wind-rider, it appears that my problem was the configuration for my Nest (maybe because of special characters in the password). Everything runs properly once that's removed.

wind-rider:
@Diaoul @Dr-Bean  hi, today I filed a new PR for Home Assistant at https://github.com/SynoCommunity/spksrc/pull/1773. ChromeCast finally works and there are several other new and updated modules. 

@stvnrlly The Nest dependency is updated, but I can't test it myself since I don't have a Nest. Maybe this one fixes your password bug? Also another Nest bug was fixed in https://github.com/balloob/home-assistant/issues/161

@Decipher Wrt. the password - the default password is empty, just click login (unless you manually set a password yourself).

PR #1659, opened 2015-04-22
WIP: SPK for Tor 

cytec:
creates a Tor bridge (unannonced) with SOCKS5 Proxy on `0.0.0.0:9050` (unprotected)

@Diaoul @Dr-Bean i thought about adding a wizzard but that would be a lot of steps to go through if you want to be able to set up correctly... maybe a more basic wizzard? like Tor itself does for OSX/Windows Users? so you basically decide for what you are going to use it and thats it?

So that the wizzard lets you decide between `exit node` and `bridge` on the other hand... i think if someone want's to host an exit node we can assume they know what to do and how to change the settings?...


Diaoul:
Yep, simple settings is good. Anything specific should handled manually by the users.

cytec:
@Diaoul so i was thinking about something like that... any thoughts?

    [
        {
            "step_title":"TOR Relay Type",
            "items":[
                {
                    "type":"singleselect",
                    "desc":"Please select your prefered TOR Relay Type",
                    "subitems":[
                        {
                            "key":"TOR_NONEXIT",
                            "desc":"Non-exit (Configure TOR as an Non-exit Relay)",
                            "defaultValue":false
                        },
                        {
                            "key":"TOR_EXIT",
                            "desc":"Exit (Configure TOR as an Exit Relay)",
                            "defaultValue":false
                        },
                        {
                            "key":"TOR_BRIDGE",
                            "desc":"Bridge (Configure TOR as an anonymous Bridge)",
                            "defaultValue":true
                        }
                    ]
                }
            ]
        },
        {
            "step_title": "TOR Installation Note",
            "items": [
                {
                    "desc": "TOR was installed with an default configuration file depending on your previous selection at /usr/local/tor/var/torrc please verify the settings and   change them to your needs (if needed)"
                }
            ]
        }
    ]

Diaoul:
I'm afraid I don't have enough knowledge to answer you. Stick with default from GUI installs or whatever security distribution settings like Tails

cytec:
@Diaoul i think that'll cover the 3 most basic use cases with some default settings...

spk/tor/Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = Tor is free software for enabling anonymous communication. The name is an acronym derived from the original software project name The Onion Router.
+RELOAD_UI = no
+DISPLAY_NAME = Tor
+
+BETA = 1
+
+HOMEPAGE   = https://www.torproject.org
+LICENSE    =
```
Dr-Bean: https://gitweb.torproject.org/tor.git/plain/LICENSE, found on https://www.torproject.org/docs/faq.html.en#DistributingTor

spk/tor/src/installer.sh
```diff
@@ -0,0 +1,89 @@
...
+# Package
+PACKAGE="tor"
+DNAME="Tor"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+SSS="/var/packages/${PACKAGE}/scripts/start-stop-status"
+PATH="${INSTALL_DIR}/bin:${PATH}"
+USER="tor"
+GROUP="users"
```
Dr-Bean: Is access to filesystem necessary for this package? If not: `nobody` would fit better :)

spk/tor/src/installer.sh
```diff
@@ -0,0 +1,89 @@
...
+
+postupgrade ()
+{
+    # Restore some stuff
+    rm -fr ${INSTALL_DIR}/var
+    mv ${TMP_DIR}/${PACKAGE}/var ${INSTALL_DIR}/
+    rm -fr ${TMP_DIR}/${PACKAGE}
+
+    exit 0
+}
```
Dr-Bean: You might want to look at your editor sometime to fix those line endings ;)

spk/tor/src/wizard/install_uifile
```diff
@@ -0,0 +1,36 @@
+[
+    {
+        "step_title":"TOR Relay Type",
+        "items":[
+            {
+                "type":"singleselect",
+                "desc":"Please select your prefered TOR Relay Type",
```
Dr-Bean: `preferred`, and I'd suggest `relay type`, without the caps (L3 too btw)

spk/tor/src/wizard/install_uifile
```diff
@@ -0,0 +1,36 @@
...
+                "subitems":[
+                    {
+                        "key":"TOR_NONEXIT",
+                        "desc":"Non-exit (Configure TOR as an Non-exit Relay)",
```
Dr-Bean: I'd propose to use `non-exit relay`, `exit relay` and `anonymous bridge`, to keep in line with Tor's usage of caps.

spk/tor/src/wizard/install_uifile
```diff
@@ -0,0 +1,36 @@
...
+                    }
+                ]
+            }
+        ]
+    },
+    {
+        "step_title": "TOR Installation Note",
+        "items": [
+            {
+                "desc": "TOR will be installed with an default configuration file depending on your previous selection at /usr/local/tor/var/torrc please verify the settings and change them to your needs (if needed)"
```
Dr-Bean: Maybe something like the description below? You might also want to check if using quotes work in the wizard (for around the `/user/local` link, helps distinguish it from the text around it), and I can't recall if those need to be escaped or not.

`The configuration file with your settings is located at /usr/local/tor/var/torrc.<br>Please verify the settings and change them to your needs.`

spk/tor/src/wizard/install_uifile_ger
```diff
@@ -0,0 +1,36 @@
...
+                        "defaultValue":false
+                    },
+                    {
+                        "key":"TOR_EXIT",
+                        "desc":"Exit (TOR als Exit Relay betreiben)",
+                        "defaultValue":false
+                    },
+                    {
+                        "key":"TOR_BRIDGE",
+                        "desc":"Bridge (TOR als anonyme Brdige/Proxy betreiben)",
```
Dr-Bean: `bridge` :P

spk/tor/src/wizard/uninstall_uifile
```diff
@@ -0,0 +1,11 @@
+[
+    {
+        "step_title": "TOR Uninstallation Note",
+        "items": [
+            {
+                "desc": "If you like to backup your configuration and ID, please backup /usr/local/tor/var BEFORE you uninstall this package!"
```
Dr-Bean: Another way of putting it, a bit more in line with other package uninstall messages:
`Attention: The Tor configuration and ID will be removed during package uninstallation. Remember to backup /usr/local/tor/var before you uninstall this package`

Dr-Bean:
@transper, you brought up some valid points:
- On the subject of the optional first step for the wizard you mentioned: Synology's wizard approach doesn't allow for that unfortunately.
- Adding translations is always appreciated :) Cytec can probably pick them up if you add them in a comment (after finalizing what the text should be, I guess)
- Portforwarding: @cytec, could the FWPORTS file be added? As for linking to the portforwarding guide: people who run this package should be quite aware of how that works imo :)
- Idle/bandwidth: See below. If we link to the Tor documentation, imo, that's covered. The interesting bits wrt the configuration probably start from here: https://www.torproject.org/docs/faq.html.en#torrc.

In general, I can see the appeal of making the wizard easier to understand for users unfamiliar with Tor, but I'm not sure that's the right approach.
This isn't a package anyone should run without being fully aware of what they are doing and which consequences that has or even risks they take/might be taking. 
If anything, I'd prefer a link to the Tor documentation (SUPPORT_URL? HELP_URL? I always forget which one we're supposed to use), and keeping the wizard in line with that wrt terminology as much as possible.

cytec:
@transper i've seen the install wizard you've mentioned but i think it's a little bit "over the top" my goal was to provide 3 "basic" setups (like tor installer itself does) and then allow the user to edit the settings as I see this as a "advanced user" SPK in the first place... translations are verry welcome once we decided the wording and/or the single steps :)

@Dr-Bean adding a FWPORTS file shouldn't be a problem i think. also there are different ports used on different setups but i think we can just enter them all

singpolyma:
I want to use this to set up a hidden service for my NAS, but I will have no issue adding config files manually so long as tor is installed and starts when the systems boots.

KaraokeStu:

Thanks,

Stuart

cytec:
Not sure if this is still a "thing" afaik there where still some points left open on how we handle that

Safihre:
@cytec Do you think you ever want to continue this PR? Seems like an interesting one to me 👍

cytec:
@Safihre yeah totally forgot about this one... i'd love to continue it if i get some time to refactor everything and move it to the new generic approach :)

Safihre:
@cytec It actually looks so close to being done, right? Just need to sprinkle a little bit of Generic Service on top and voila 🎉 🎈 😄

cytec:
basically... yep. but i'd like to fine tune the wizzard a little bit more and fix the comments that @Dr-Bean mentioned. Also rebase and squashing should be done i guess ^^

cytec:
@Safihre i'm closing this in favor of https://github.com/SynoCommunity/spksrc/pull/3189

PR #1669, opened 2015-05-02
Rule updates

Dr-Bean:
- Move toolchain definitions from spk.mk and cross-cc.mk to common.mk
- Rename `SUPPORTED_TCS/ARCHS` to `AVAILABLE_TCS/ARCHS`
- Add main rules: `all-supported` and `all-legacy`
- Add REQ_KERNEL support to `all-supported` and `all-archs-latest` rules.

Ref #1366

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 
-SUPPORTED_TCS = $(notdir $(wildcard toolchains/syno-*))
-SUPPORTED_ARCHS = $(notdir $(subst syno-,/,$(SUPPORTED_TCS)))
+AVAILABLE_TCS = $(notdir $(wildcard toolchains/syno-*))
+AVAILABLE_ARCHS = $(notdir $(subst syno-,/,$(AVAILABLE_TCS)))
 SUPPORTED_SPKS = $(patsubst spk/%/Makefile,%,$(wildcard spk/*/Makefile))
 
 
@@ -61,8 +61,8 @@ natives:
 	done
 
 .PHONY: toolchains kernel-modules
-toolchains: $(addprefix toolchain-,$(SUPPORTED_ARCHS))
-kernel-modules: $(addprefix kernel-,$(SUPPORTED_ARCHS))
+toolchains: $(addprefix toolchain-,$(AVAILABLE_ARCHS))
+kernel-modules: $(addprefix kernel-,$(AVAILABLE_ARCHS))
 
 toolchain-%:
 	-@cd toolchains/syno-$*/ && MAKEFLAGS= $(MAKE)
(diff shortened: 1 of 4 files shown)


Dr-Bean:
This adds the `all-supported` and `all-legacy` rules that were discussed in #1366. 
The `all-supported` rule only builds for supported/current archs (defined [here](https://github.com/SynoCommunity/spksrc/pull/1669/files#diff-03673f6b8edacbe66b58298f02036ef0R28)), with the latest available toolchain.
When `REQ_KERNEL` is not set (most of the packages), an `x64` package is created, which decreases overall build time due to skipping separate `avoton`, `bromolow` and `cedarview` packages.

The `all-legacy` rule builds for the old archs (defined [here](https://github.com/SynoCommunity/spksrc/pull/1669/files#diff-03673f6b8edacbe66b58298f02036ef0R29), as well as builds for supported/current archs with the previous toolchain (4.3, hardcoded [here](https://github.com/SynoCommunity/spksrc/pull/1669/files#diff-93babb8aca8f10daaac201d6641f6d13R351))

The `all-archs-latest` rule essentially combines the two rules above (via a different, slightly more efficient, approach).

Diaoul:
Awesome, sorry I missed that PR somehow.
Can you update the README with that comment above?

Dr-Bean:
Yep. I was actually considering a separate wiki to explain available building rules for cross and spk, and to refer to that from the README. I could put in all in the README, but imo, it's getting a bit crowded if we add it in.

The README will probably need an overhaul at some point anyway, considering the Docker instructions need to be added too (although those could also be added to a separate article and linked for that matter). What do you think?

cytec:
im with @Dr-Bean on that one... i think putting such stuff in a wiki article would be the best place... basic instructions are in the readme and if one likes to know something deeper/further we can link them to the wiki... Also i'd love to have a summary of the available rules and a short explanation on what they do and what they are good for at some point... it's harder and harder to keep all that stuff in mind :D

Diaoul:
OK with that but at least we need to update the basic instructions of the README. I'll add some Docker documentation and update the README as well.

You can merge when the wiki article is done (please refer to it here)

Diaoul:
See #1727

Dr-Bean:
Here we go: https://github.com/SynoCommunity/spksrc/wiki/Compile-and-build-rules.
That should be about it. Note: hyperlinks in there need changing after merging the PR.

...and while I was looking at it, I can't remember if `all-archs-latest` was supposed to be renamed to the current `all-archs-dsms`, or the other way around.

Diaoul:
Good work. I believe this will make the build process a lot easier.

Dr-Bean:
Thanks for merging :) Just fixed the hyperlinks in the article, so that's done too.

Dr-Bean:
Somewhat related to this subject: I've set the default toolchain to 5.1 (when `make setup` is executed) via https://github.com/SynoCommunity/spksrc/commit/18dcad0457d78244042ffe4883aa21aca21e80a6.

PR #1670, opened 2015-05-03
Added `flexget` to the path since it now runs daemon only.

sebask:
This makes working with flexget from the cli a little bit easier.

### DIFF
--- a/spk/flexget/src/installer.sh
+++ b/spk/flexget/src/installer.sh
@@ -26,6 +26,10 @@ postinst ()
     # Link
     ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
 
+    # Put mc in the PATH
+    mkdir -p /usr/local/bin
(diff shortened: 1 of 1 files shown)


Dr-Bean:
We're probably not going to merge this, sorry. You can add the Flexget directory to PATH yourself:
```
PATH=/your/new/dir:$PATH
export PATH
```

sebask:
Okay no problem. Just wondering, why?

Diaoul:
Because /usr/local is shared within the whole system and this might result in conflicts. It is tolerated for CLI only packages to provide a symlink in /usr/local/bin but for packages with a GUI this is not necessary.

Dr-Bean:
What @Diaoul said. To add to that: Even though the flexget package currently does not have a webui, that's not to say it won't return (eventually). That obviously won't occur with pure CLI-based applications.

I'll close the PR, but nonetheless thanks for your input :)

sebask:
@Diaoul @Dr-Bean I agree, it makes a lot of sense. Thanks for sharing!

PR #1672, opened 2015-05-04
added ImageMagick SPK

cytec:


cross/imagemagick/Makefile
```diff
@@ -0,0 +1,32 @@
...
+myInstall:
+	$(RUN) make install prefix=$(STAGING_INSTALL_PREFIX) \
+ SHARE_PATH=$(STAGING_INSTALL_PREFIX)/share/ImageMagick-6/ \
+ DATA_DIR=$(STAGING_INSTALL_PREFIX)/share/ImageMagick-6/ \
+ CONFIGURE_PATH=$(STAGING_INSTALL_PREFIX)/etc/ImageMagick-6/ \
+ LIBRARY_PATH=$(STAGING_INSTALL_PREFIX)/lib/ImageMagick-6.9.0/ \
+ SHAREARCH_PATH=$(STAGING_INSTALL_PREFIX)/lib/ImageMagick-6.9.0/config-Q16 \
+ DOCUMENTATION_PATH=$(STAGING_INSTALL_PREFIX)/share/doc/ImageMagick-6 \
+ INCLUDE_PATH=$(STAGING_INSTALL_PREFIX)/include/ImageMagick-6 \
+ INCLUDEARCH_PATH=$(STAGING_INSTALL_PREFIX)/include/ImageMagick-6
```
Diaoul: Are all the overrides necessary? Also please indent this?

cross/imagemagick/PLIST
```diff
@@ -0,0 +1,38 @@
...
+shr:etc/ImageMagick-6/log.xml
+shr:etc/ImageMagick-6/magic.xml
+shr:etc/ImageMagick-6/mime.xml
+shr:etc/ImageMagick-6/policy.xml
+shr:etc/ImageMagick-6/quantization-table.xml
+shr:etc/ImageMagick-6/thresholds.xml
+shr:etc/ImageMagick-6/type-dejavu.xml
+shr:etc/ImageMagick-6/type-ghostscript.xml
+shr:etc/ImageMagick-6/type-windows.xml
+shr:etc/ImageMagick-6/type.xml
```
Diaoul: You can do `shr:etc/ImageMagick-6` here, it'll apply to the directory

cytec: not sure if we need the `etc` directory at all...

cross/imagemagick/Makefile
cytec: i tested it without those overrides some time ago... if i remember correctly those where needed but im not 100% sure... can check that again if you want :)

cytec:
@Diaoul @Dr-Bean anything else? or are we good to go :)

spk/imagemagick/Makefile
```diff
@@ -0,0 +1,40 @@
...
+POST_STRIP_TARGET = imagemagick_extra_install
+
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: imagemagick_extra_install
+ntop_extra_install:
+	install -m 755 -d $(STAGING_DIR)/app
+	install -m 755 -d $(STAGING_DIR)/app/images
+	for size in 16 24 32 48; do \
```
Dr-Bean: ...72 ;)

Dr-Bean: Although, now that I'm thinking of it, you don't actually need the DSM_UI_DIR, right? It's a non-startable package, so there's no menu entry needed.

sidenote: we should probably check that sometime...I'm quite sure I add it for packages that don't need it myself...

cytec: @Dr-Bean you are right... so what? remove `DSM_UI_DIR = app` and `imagemagick_extra_install` completely?

Dr-Bean:
Good to go imo, although that broken line-ending stood out in scrolling past it ;)

cytec:
@Dr-Bean @Diaoul i've just worked a little bit with the ImageMagick binaries from this SPK and found out a few things:

1. seems like there are no fonts at all installed on the DS by default at least not in the directorys IM is looking for them (`/usr/share/fonts/type1/gsfonts/s050000l.pfb`) 
2. If you want to install/add some fonts you'll have to install `Pearl` and run http://www.imagemagick.org/Usage/scripts/imagick_type_gen to create a new xml which has to be included in `/usr/local/imagemagick/etc/ImageMagick-6/type.xml`

so i was thinking about something like this:

1. Add `Pearl` to dependencies and add the `imagick_type_gen` script directly to the bin folder.
2. Create a `var/fonts` folder which contains a (empty) `mytype.xml` which is referenced in `type.xml` so one can put fonts in the `var/fonts` dir and update the `mytype.xml` to use those?
3. maybe add `fontconfig` to dependencies as well...?

or alternatively i was thinking about including some google fonts, or make a google fonts package at all... which contains the fonts provided by google and an `googlefonts.xml` which contains the definititions and is added to `type.xml` at install time?

any thoughts?

Diaoul:
Once installed you cannot upgrade fonts in the var/ directory as it will be restored after upgrade. So if you go that way that means once installed fonts are pretty untouchable.
Is there a way to pick fonts in share/ instead? That way we keep control of the directory and it would be nice to have some documentation about how to add a font, tell the users to put them in var/ and do a symlink in share? Or configure imagemagic so it picks fonts in both share/ and var/ ?

I don't know what's possible

cytec:
yeah of course... if we include some fonts by default we should put them in share/... but allow the user to define fonts in var/ as personally i have some fonts i'd like to have installed even when updating ImageMagick...

you can link as many files as you want to in `/usr/local/imagemagick/etc/ImageMagick-6/type.xml` here is an example:

    <typemap>
       <include file="type-ghostscript.xml" />
       <include file="/usr/local/imagemagick/share/synocommunity.xml" />
       <include file="/usr/local/imagemagick/var/userfonts.xml" />
    </typemap>  

so we could provide some default fonts (maybe change the xml name) and the ability to allow the user to add his own in the `var/userfonts.xml` ... this is how i have done it currently on my device (expect the synocommunity.xml ;)) 

example content of `userfonts.xml` generated with `imagick_type_gen`:

    <typemap>
         <type
            format="ttf"
            name="OpenSans"
            fullname="Open Sans"
            family="Open Sans"
            glyphs="/usr/local/imagemagick/var/fonts/OpenSans-Regular.ttf"
            />
    </typemap>

Diaoul:
Awesome.

cytec:
hm best way would be to create a xml in var which includes other xml files in var (if thats possible)
just came to my mind with the google fonts spk:

If the user installs ImageMagick and then Google Fonts, enters the path to the xml in ImageMagick's `type.xml` and then updates ImageMagick the `type.xml` gets overwritten...

my solution would be:
in `type.xml` reference an `user-include.xml` which is located inside the var folder and which then includes other xml files... so basically something like this:

`type.xml`:

    <typemap>
       <include file="type-ghostscript.xml" />
       <include file="/usr/local/imagemagick/share/synocommunity.xml" />
       <include file="/usr/local/imagemagick/var/custom-includes.xml" />
    </typemap>      

`custom-includes.xml`:

    <typemap>
       <include file="/usr/local/googlefonts/share/googlefonts.xml" />
       <include file="/usr/share/fonts/myfonts.xml" />
    </typemap>     

haven't tested if this nested includes would work but if they do that seems to be the best solution for me

cytec:
@Dr-Bean @Diaoul i'm currently trying to compile ImageMagick with OpenMP support so that (at least x64 arches) have multicore support... works pretty well and gives a huge speed increase!

Without OpenMP:

    daenerys> time convert model.png -limit thread 1 -sharpen 0x1 null:
    real    0m 12.43s
    user    0m 12.43s
    sys 0m 0.02s

With OpenMP:

    daenerys> time convert model.png -sharpen 0x1 null:
    real    0m 4.17s
    user    0m 13.70s
    sys 0m 0.03s

but `libgomp` is located and linked against `/home/cytec/syno/toolchains/syno-x64-5.1/work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib` which is of course not available on the DS itself... Copying the file from `/home/cytec/syno/toolchains/syno-x64-5.1/work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib` to `/usr/local/imagemagick/lib` on my DS however, works! 

Is there any way to copy this file over to the ImageMagick lib dir on compilation?

Diaoul:
Never had this kind of issue before where you need to copy some toolchains libs on the NAS. Can you locate that lib on your NAS? Are you using DSM 5.1 on your NAS? Have you tried using the dedicated toolchain, not the x64 one?

cytec:
well `libgomp` is included in gcc >= 3.7 so it's present in the toolchains lib dir... just found out that it's also available on my DS in the `/lib` dir.

Yeah I'm using DSM 5.1 and i've tried to compile with dedicated cedarview toolchain as well...
maybe `ADDITIONAL_LDFLAGS = -Wl,--rpath,/lib` does the trick as there is a `/lib/libgomp.so.1` present on my NAS already

Diaoul:
What if you add LD_LIBRARY_PATH=/lib before launching the command? I'm not a linker expert so I don't know exactly how libraries are searched and what we can do about it.

cytec:
hm seems not to work:

    daenerys> LD_LIBRARY_PATH=/lib convert
    convert: error while loading shared libraries: libgomp.so.1: wrong ELF class: ELFCLASS32

if i get that correctly: there is a 32bit libgomp in /lib but convert is a 64bit process

Dr-Bean:
Another option could be to try and add `-static` to the compile process, from what I'm reading here: http://stackoverflow.com/questions/6351961/specify-openmp-to-gcc.
I don't know if ImageMagick supports it or if that has other side effects though

cytec:
I've tried adding `-static` but that doesn't seem to work i've also tried `-Wl,-Bstatic -lgomp -Wl,-Bdynamic` which also did not work... maybe i'm just doing it wrong :/ but seems like the only way to get it working is/was copying the toolchain files -.-

Dr-Bean:
That's because ImageMagick doesn't have a `-static` option as such. Looking at the ImageMagick configure file, you'd probably end up with something like this: `--enable-openmp --enable-static --disable-shared`. Note that the PLIST will be outdated too.

I don't have a reference, as I didn't compile the original, but `convert --version` shows `Features: DPC OpenMP`

Dr-Bean:
@cytec I've been holding off on merging this PR because of the OpenMP discussion above (and the fonts PR, I don't know how that one interacts with this PR atm).

Am I correct when I say that we can merge this PR as-is, and move the OpenMP stuff to a new issue/PR? If not, let's remove the current `ready-to-merge` label and put back a WIP label.

cytec:
@Dr-Bean yup should be good to go... OpenMP support and also the font stuff will be a new PR once i've got it working somehow :/

Dr-Bean:
Package ImageMagick 6.9.1-1 has been published on the repository.

PR #1679, opened 2015-05-07
Openssl 1.0.2

GaetanCambier:
Update to Major version openssl 1.0.2

i have include an patch to support cipher chacha20_poly1305
this cipher, used by google in chrome, use strong encryption and is fastest than aes
(usefull for in example haproxy to minimise server load)

### DIFF
--- a/cross/openssl/Makefile
+++ b/cross/openssl/Makefile
@@ -1,11 +1,11 @@
 PKG_NAME = openssl
-PKG_VERS = 1.0.1m
+PKG_VERS = 1.0.2a
 PKG_EXT = tar.gz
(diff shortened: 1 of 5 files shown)


Dr-Bean:
OpenSSL has been updated to 1.0.2a with https://github.com/SynoCommunity/spksrc/commit/339c2a3a598863a7676076e02ceca2f232c7e01d.

I did not add the additional cipher though. I don't know exactly what the status of it is (other than some minor discussions on mailing lists), and it's not clear to me how far off the OpenSSL devs are from merging. @Diaoul, your thoughts?

Diaoul:
I'd say we stick with standard stuff, if one needs bleeding edge stuff that's what spksrc is for: build your own packages.
I doubt Synology NAS are made to host heavily loaded websites.

PR #1687, opened 2015-05-15
Htop SPK

terziyski:
Htop v1.0.3 - an interactive system-monitor process-viewer written for Linux (intended for DSM-5.0 and earlier versions)
htop


Dr-Bean:
I believe `/usr/bin/htop` is available by default on DSM5.1 and up.

terziyski:
Yes, it is. My target were the previous versions.

Dr-Bean:
We can't target for a specific DSM version/up to a specific version though, so it shows up on DSM5.0 and newer too.
The problem, fwiw, is mainly with the symlink you added here: https://github.com/SynoCommunity/spksrc/pull/1687/files#diff-e7f2513d27b576f57611a5997a65a9e6R21.
Running `htop` on newer devices will run the built-in `/usr/bin/htop` instead of your package. That might cause unexpected behavior for people, so we probably want to avoid that.

I'd suggest to remove that symlink, and add something to the package description: `This package is intended for DSM4.3, where htop is not available.`, or something like that.

gkroone:
@Dr-Bean , What do you mean with "We can't target for a specific DSM version"? surely you can include a check for the DSM version inside the install script? or do you mean this will break when DSM is upgraded to a newer version with the package still installed?

Dr-Bean:
I meant from a Package Center availability point of view. We can set a minimum DSM version in INFO, but we can't set a maximum version. 
In other words, you can say "This package should be available for DSM5.0 and newer only", but not the reverse: "This package should be available for DSM4.3 and older only"

If you were thinking of something like showing a package in PC and then stopping the installation with a message like "htop is already available on your DSM version, you don't need this package", that just provides a very poor user experience.

terziyski:
@Dr-Bean 
I will fix the description as you suggested.
Although I just tested the integrated htop in DSM-5.1 - it's a v1.0.2 and it won't start instead of the freshly installed SPK (v1.0.3) - basically the installer.sh is the same as for nano SPK.
Or by changing the symlink you meant something that I didn't catch ?

cytec:
@terziyski what @Dr-Bean was refering to is, that in line 21-23 of your installer script (https://github.com/SynoCommunity/spksrc/pull/1687/files#diff-e7f2513d27b576f57611a5997a65a9e6R21) you are linking the binary to `/usr/local/bin` which comes after `/usr/bin` where `htop` is located at DSM 5. So if the users runs `htop` on DSM 5 there will be executed the synology one `/usr/bin/htop` and not your package `/usr/local/bin/htop` which can be confusing if they do not behave the same... so maybe you'll better leave this to the user itself ;)

terziyski:
@cytec Thanks for the detailed answer. It means I understood corectlly what @Dr-Bean meant.
So that's why I tested the freshly created SPK against DSM-5.1 (which I have). Like I wrote earlier @Dr-Bean when I try to run ```htop``` from SSH console (as root) the package which starts is the new SPK (v1.0.3) isntead of the integrated one (v1.0.2). So if you want to execute the integrated (v1.0.2) you have to specify its full path to do that. When the SPK is removed the normal behaviour returns back. 
My ```$PATH``` is as follows: ```/sbin:/bin:/usr/sbin:/usr/bin:/usr/syno/sbin:/usr/syno/bin:/usr/local/sbin:/usr/local/bin```
Am i missing something else here ? Isn't that the behaviour in pursuit ?

Dr-Bean:
@terziyski Your setup seems to behave differently from default :)
When you execute `htop` or `which htop` on a standard device, it should start/show the first `htop` that is encountered on PATH. On a default DSM5.1 setup, that should be `usr/bin/htop`, not `/usr/local/bin/htop`. On an older DSM version, it should not show any result at all.

To be clear, our packages should never take precedence over built-in functionality. The correct way to ensure that is to not add the symlink.
The difference in behavior between your device and mine illustrates exactly why we're reluctant to create symlinks in `/usr/local/bin` in general ;)

terziyski:
@Dr-Bean well I thought I have a pretty standard setup on my DSM 5.1 :) - thanks for clarifications.
So you want me to just remove the lines https://github.com/terziyski/spksrc/blob/master/spk/htop/src/installer.sh#L21-24 ?

Dr-Bean:
If you could, yes please :)

terziyski:
Will do, thanks for the guidance guys.

spk/htop/Makefile
```diff
@@ -0,0 +1,23 @@
+SPK_NAME = htop
+SPK_VERS = 1.0.3
+SPK_REV = 1
+SPK_ICON = src/htop.png
+
+DEPENDS = cross/$(SPK_NAME)
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = An interactive process viewer for Linux (the package is intended for DSM-4.3 or earlier where it's not available)
```
Dr-Bean: This is just nitpicking, but you can change the `(the package[..] available)` to a separate sentence: 
`An interactive process viewer for Linux. The package is intended for DSM-4.3 or earlier where it's not available.`

Btw, I believe that should be DSM 5.0 there, not 4.3, right? If I'm not mistaken, DSM5.1 was the first version with `htop` (but go ahead and verify that if you can)

terziyski:
I've put DSM-4.3 based on your seccond post here :) But sure I could do a research if I can find in which DSM version it appears for the first time.

Dr-Bean:
Oh right, I was just providing input for the description there ;)
Anyway, my DSM5.0 device does not have it, my 5.1 device does.
You can download and extract firmware versions off of Synology's download site if you feel like doing that. It's not all that important though.

terziyski:
Description in Makefiles and paths in installer.sh script modified as discussed.

Dr-Bean:
Htop fails to build on a number of toolchains: `bromolow-4.3`, `cedarview-4.3`, `powerpc-4.0`, `ppc824x-4.2`, `ppc854x-4.0`, `x86-4.3`. 
That's kind of an issue where the targeted DSM versions are 4.x...

As an example, below the error for Bromolow-4.3. Can you verify this, and see if it can be fixed?
```
mv -f .deps/htop-BatteryMeter.Tpo .deps/htop-BatteryMeter.Po
/spksrc/toolchains/syno-bromolow-4.3/work/x86_64-linux-gnu/bin/x86_64-linux-gnu-gcc -DHAVE_CONFIG_H -I.  -DNDEBUG -I/spksrc/spk/htop/work-bromolow-4.3/install//usr/local/htop/include  -pedantic -Wall -Wextra -std=c99 -rdynamic -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR=\"/usr/local/htop/etc\" -I/spksrc/spk/htop/work-bromolow-4.3/install//usr/local/htop/include  -MT htop-Process.o -MD -MP -MF .deps/htop-Process.Tpo -c -o htop-Process.o `test -f 'Process.c' || echo './'`Process.c
Process.c: In function 'Process_updateIOPriority':
Process.c:713: error: 'SYS_ioprio_get' undeclared (first use in this function)
Process.c:713: error: (Each undeclared identifier is reported only once
Process.c:713: error: for each function it appears in.)
Process.c: In function 'Process_setIOPriority':
Process.c:719: error: 'SYS_ioprio_set' undeclared (first use in this function)
Makefile:761: recipe for target 'htop-Process.o' failed
make[4]: *** [htop-Process.o] Error 1
make[4]: Leaving directory '/spksrc/spk/htop/work-bromolow-4.3/htop-1.0.3'
Makefile:292: recipe for target 'all' failed
```

terziyski:
Sure I'll look into it.

terziyski:
So, I can confirm that ```htop v1.0.1``` compiles successfully for ```bromolow-4.3```
This issue is due to an old header files for toolchains in question - their kernel is probably earlier that ```kernel 2.6.13```.
What I can do is to modify Makefiles accordingly and to commit them with ```htop v1.0.1``` instead.

Dr-Bean:
That's just about the only option, so go ahead.
Before opening the PR, can you test the other failing arches as well? Whatever still fails with `htop 1.0.1` should be excluded via setting `UNSUPPORTED_ARCHS`.

terziyski:
I was not able to compile ```htop v1.0.1``` only for these two: ```powerpc-4.0```, ```ppc824x-4.2```.
Another error shows up: 'unrecognized option -Wextra in AvailableMetersPanel.c'.
I think it's reasonable to exclude these two with the setting you mentioned.

cytec:
think so too... @Dr-Bean haven't we dropped support for ppc arches altogether? thought there was something about that and putting them in legacy_arches or something like that?

Dr-Bean:
Sort of. For spksrc itself, legacy arches are arches for which no new toolchains are released. We agreed that support/package maintenance will be turned down a notch or two for those arches, and same for older DSM versions, meaning: if it doesn't work, don't bend over backwards to fix it.

Once #1669 is merged, the default rule to build packages is `all-supported`. But, you can still build packages for legacy arches via specifying it directly or using the [`all-legacy`](https://github.com/SynoCommunity/spksrc/pull/1669/files#diff-03673f6b8edacbe66b58298f02036ef0R27) rule.

We'll still upload legacy arches to the repository for the near term. At some point undecided, we'll stop doing that, and upload only when a certain package is requested.

terziyski:
New PR #1722  opened.

PR #1695, opened 2015-05-18
update monit to 5.13

cytec:
@Dr-Bean @Diaoul i'm still thinking about a proper way to include files for some commonly used packages... but sadly if the path doesn't exist it'll throw out errors so just providing some sample files in `monit.d` won't work...

so i see 2 "options" here:

1. check for every path if it exists in the installer and only install those who exist
2. provide some default files which are all commented out by default


Dr-Bean:
Or option 3: Use a `sample` dir to put the files in, instead of commenting out the content of every file.
I think I'd prefer the options in order: 3, 2, 1...although I'm currently not a user of Monit ;)

Other than that, no comments on the current PR ;)

cytec:
option 3 sounds good for me too... didn't thought about that :D

@Dr-Bean @Diaoul im getting some `502 Bad Gateway` errors when publishing to the repo, anyone else?

Dr-Bean:
Haven't seen those errors in a while. Even with those errors, it's likely that the upload completed succesfully. Should be fine if the packages are visible in the interface.

cytec:
@Dr-Bean nope got those errors for everything except `cedarview`, `avoton`, `comcerto2k` and `bromolow` when running `make all-toolchain-5.1` and `make publish-all-toolchain-5.1` and they don't show up in the backend either...

Dr-Bean:
For the record, x86/64-5.1 based archs fail to upload if x64 was uploaded first, but those show a `Conflict` error, not a 502 error (I need to fix and merge my rules PR to correct that behavior, but I'm lagging a bit).

Otherwise, I don't have a solution other than to retry. I suppose a busy server might be the cause, but that's just guessing on my part.

cytec:
yeah i'm aware of that x86 Conflict errors... got them a few times already ;)
i don't belive in a busy server, i've tried 4 times already (with some time between) and every time with the same results: `cedarview`, `avoton`, `comcerto2k` and `bromolow` got uploaded (or a `conflict` error after the first upload) and all the other arches i'm getting 502 errors

Diaoul:
Maybe we ran out of disk space. I cannot check from work though.

Diaoul:
Or one of the uwsgi process went AWOL

cytec:
@Diaoul any news on this? still getting the 502 Error :/

Diaoul:
There are still a few GB available, I'm rebooting now.

cytec:
@Diaoul i don't get it... still getting the 502 error :/ i've even checked the http command but it seems good to me: `http --auth MYAPIKEY: POST https://api.synocommunity.com/packages @/home/cytec/syno/spk/monit/../../packages/monit_armada375-5.1_5.13-7.spk`

Dr-Bean:
@cytec Is this good to merge as-is, or did you want to add the samples to this PR?

cytec:
@Dr-Bean good to go, the samples don't really work like i thought... monit doesn't skip non existing services by default... so you'll have a lot of "broken" stuff in there if samples are included... still thinking about an option (maybe just provide samples with commented code or something like that)

Dr-Bean:
I probably have no idea what I'm talking about, but if you have a `sample` dir (located somewhere that isn't read by Monit), and you add one sample per package (e.g. one for Sickbeard, one for SAB, that kind of approach), then people can use those samples as a starting point: they can copy the scripts they want to use, to the location Monit *does* read, and edit that particular script to their needs.
Wouldn't that work?

cytec:
@Dr-Bean that would work! I didn't even thought about that :D

cytec:
@Diaoul @Dr-Bean on that sample dir... where should we put it? in `share/examples` or in `var/examples` personally i think `share/examples` would be better as we can always update/change/add them while if we put them if it's needed

Diaoul:
Yes, share/examples is better

Dr-Bean:
@cytec Apart from a possible upgrade to 5.14 (just checked), would this be good to go?

cytec:
@Dr-Bean yup should be good to go :)

Dr-Bean:
Merged with https://github.com/SynoCommunity/spksrc/commit/055de70b69d5f77df08a647826482a731492945c. Some changes to the examples approach via  https://github.com/SynoCommunity/spksrc/commit/20964dfb7808d02c9f1961c6536a14f0a5b3d507.

Issue #1492, opened 2015-02-16
BTSync compilation broken


tmarkwardt:
Hi,

i can't compilation the spk btsync. on make the script checked all files on digest and not found libraries with unused architectures ...

Regards,
Torsten

tmarkwardt:
here a solution:
file: spksrc.checksum.mk
modification: select only $(DIST_FILE) filename from file digist to validate checksum

--- mk/spksrc.checksum.mk_orig	2015-02-16 16:20:25.649627628 +0100
+++ mk/spksrc.checksum.mk	2015-02-16 18:52:43.450749911 +0100
@@ -41,7 +41,9 @@
 checksum_target: $(PRE_CHECKSUM_TARGET)
 	@if [ -f $(DIGESTS_FILE) ] ; \
 	then \
-	  cat $(DIGESTS_FILE) | (cd $(DISTRIB_DIR) && while read file type value ; \
+	  distfile=$(DIST_FILE) ; \
+	  distfilename=$${distfile##*/} ; \
+	  cat $(DIGESTS_FILE) | grep "^$${distfilename}" | (cd $(DISTRIB_DIR) && while read file type value ; \
 	  do \
 	    if [ ! -f $$file ] ; \
 	    then \

Dr-Bean:
Or download all the packages (either manually, or do a `make all-archs && make clean` in cross/btsync. BTSync is a one-off that has this issue.

Overall, I wouldn't mind seeing the functionality implemented, but I'm not sure I like the way this looks. Maybe compare $(DIST_FILE) to $$file would make sense?

Sidenote: please put your code between backticks (`), that preserves the formatting.

tmarkwardt:
on `$(DIST_FILE)` the file is with path and a match or compare is needed. 
the alternate is the loop over all entries without filtering and a additional compare ...
i think, the filter of file is more efficient.
only a little security set: `grep "^$${distfilename} "` with a space on end on filename.

Diaoul:
@Dr-Bean: is there any pending action here?

Dr-Bean:
@Diaoul It's essentially the todo item here: https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.checksum.mk#L9. We've managed without it for a while though ;) What do you think of @tmarkwardt's solution?

Diaoul:
Well if it works as expected and is backward compatible that's all that matters for me.

Diaoul:
[...]

PR #1696, opened 2015-05-18
Fix file permissions

nliaudat:
File must have read/write access for everybody

### DIFF
--- a/spk/btsync/src/dsm-control.sh
+++ b/spk/btsync/src/dsm-control.sh
@@ -15,7 +15,7 @@ PID_FILE="${INSTALL_DIR}/var/syncapp.pid"
 
 start_daemon ()
 {
-    su - ${USER} -c "PATH=${PATH} ${BTSYNC} --config ${CFG_FILE}"
+    su - ${USER} -c "umask 200;PATH=${PATH} ${BTSYNC} --config ${CFG_FILE}"
 }
 
 stop_daemon ()


Dr-Bean:
Personally, I'd prefer to not merge this, for two reasons: I like the default permissions just fine, because that means I can override them as I wish. 
Secondly, this package will no longer be updated: 1.4.111 is the last package of the line, because Bittorrent is maintaining their own set of packages for 2.0 forwards.

Can you explain why this would be a good thing to implement for all package users, instead of just locally for yourself?

nliaudat:
Hi, 
btsync sets it's permission through the custom user btsync, which is not linked to the synology users/groups. 

I mean the only way to sync the content of two or more synology nas with btsync is to set file permissions with umask as proposed.

Btsync is not only for home users, but can be used in a complex topology with more than one server and clients. 

The file access must be done with the common synology user/group (perhaps linked with AD /LDAP) interface and not be restricted with a shadow user (btsync in that case). 

Regards
Nicooas

Dr-Bean:
The `btsync` user is member of Synology's `users` group, same as for all our packages. Synology uses ACL (Access Control Lists) internally, not Unix permissions.

Dr-Bean:
Closing this PR: BTSync package will be removed from spksrc. See e.g. https://github.com/SynoCommunity/spksrc/issues/1492#issuecomment-112913138.

PR #1699, opened 2015-05-19
Google Fonts SPK

cytec:
like already mentioned here: https://github.com/SynoCommunity/spksrc/pull/1672

I've decided to use a different approach and create the needed xml when the spk is created, so the user can easily add `/usr/local/googlefonts/share/googlefonts.xml` to his ImageMagick `type.xml` and automatically gets new fonts (if the Google SPK is updated) and we avoid to depend on Pearl as `imagick_type_gen` must not be executed on the DS itself.

basically this just clones the fonts repo and generates an xml to use with ImageMagick.

Maybe we'll should make a install note where the xml is located and which file has to be modified to make this work with ImageMagick...


cytec:
@Dr-Bean @Diaoul with the `ImageMagick` Package already published i thought i just push this up again. What do you two think?

I was thinking about an Post-Install Note presented to the user on how to include it in his `ImageMagick` installation (would require sshing to DS and change a single file) or maybe even check if it is installed and alter the line automatically (not sure on this one)

Dr-Bean:
I have no clue how the two packages interact with each other, so this might be a dumb suggestion, but why not merge the fonts stuff into the ImageMagick package?

cytec:
well basically this just gives you the additional option to install **all** googlefonts to a directory on your DS which then can be used by different tools like `ImageMagick` for example...

i did not include it in the first place as it's a pretty big package and may be not needed by most users at all...

Dr-Bean:
Makes sense. How about combining your two ideas: use an install wizard with a checkbox, which says: "Would you like to add the fonts config to ImageMagick?" (which essentially makes it extensible to other packages too)
During the fonts installer.sh, use the output to make the necessary changes to ImageMagick's config. I assume an update to the ImageMagick package will be able to preserve those changes.

Might be more challenging to handle the fonts package removal though (as in: remove the changes from the ImageMagick config without breaking anything)

cytec:
technically this shouldn't be a problem... worst case you could even let it there (im doesn't break if you remove the files without changing back the config) i've written something about what has to be done here: https://github.com/SynoCommunity/spksrc/pull/1672#issuecomment-103432523

so basically you would look in `/usr/local/imagemagick/etc/ImageMagick-6/type.xml` for `<include file="/usr/local/googlefonts/share/googlefonts.xml" />` and add/remove it depending on install/remove. problem is when updating ImageMagick this will be removed so on my installation i made the changes in `/usr/local/imagemagick/var/myfonts.xml` which is refered to by `/usr/local/imagemagick/etc/ImageMagick-6/type.xml` not sure if that fix is already implemented in ImageMagick...

KaraokeStu:

Thanks,

Stuart

cytec:
Im getting to this, but i think that'll might not be before the 31st July...

PR #1716, opened 2015-05-24
update nzbget-stable and nzbget-testing

baz1860:
Upgrades nzbget from 14.2-stable to 15.0-stable
Upgrades nzbget-testing from svn-rev 1228 to svn-rev 1286


Dr-Bean:
Note: Looks like v15 contains the built-in installer/updater already: http://sourceforge.net/p/nzbget/code/1287/tree//trunk/ChangeLog?diff=51a7be6ce88f3d037b73b9b5:1286
This might mean more changes are needed (e.g. skip compile steps and use provided binaries)

Dr-Bean:
@baz1860 I think you didn't test these changes before opening the PR? It's missing some bits and pieces unfortunately.

Diaoul:
I suggest we disable this auto update feature. Packaging is more than just providing the binaries.

Dr-Bean:
Yup @Diaoul. After looking into it, it's not something I want to add right now (for one, I can't test it properly, v15 is the first version with the functionality). As it so happens, auto-update is disabled by default when you build from source, so no issue.

Dr-Bean:
Packages nzbget 15.0-16 and nzbget-testing 15.0~r1286-8 have been published.

PR #1719, opened 2015-05-25
Add get_iplayer SPK

hadess:
Hopefully fixing the problems in #822 

This also adds the fdupes cross-compile, but not the SPK, feel free to split off.


cytec:
idk what @Dr-Bean and @Diaoul think, but how about useing KSV's patches for rtmpdump as well? http://stream-recorder.com/forum/customized-rtmpdump-binaries-patch-file-t16103.html

hadess:
@cytec can't KSV get them upstream?

cytec:
seems like upstream doesn't really care about adding the patches... i use rtmpdump in some other projects as well and always used the ksv patch... lot of usefull stuff in there ;)

richardhopton:
I'm keen to get using this - looks like KSV doesn't have the patch file checked in anywhere... it's in the 1.19MB "release" on his github.  Is it worth seeing if K-S-V would setup a repo for this file so we don't need to store a copy in spksrc?

cross/fdupes/Makefile
```diff
@@ -0,0 +1,34 @@
+PKG_NAME = fdupes
+
+NAME = fdupes
```
Dr-Bean: As a general comment, please use the same formatting as other packages (wrt order and layout). 
`NAME` isn't defined in spksrc, no need to add it.

cross/fdupes/Makefile
```diff
@@ -0,0 +1,34 @@
...
+PKG_DIST_SITE = https://github.com/adrianlopezroche/fdupes.git
+# From Makefile.inc/VERSION in the git checkout
```
Dr-Bean: You probably should use official releases instead of using a git checkout for the same version. In this case, check  https://github.com/adrianlopezroche/fdupes/releases to check which releases are available, and use the associated download link, which is https://github.com/adrianlopezroche/fdupes/archive/fdupes-1.51.tar.gz

cross/fdupes/Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = myCompile
+INSTALL_TARGET = myInstall
+
+include ../../mk/spksrc.cross-cc.mk
+
+.PHONY: myCompile
+myCompile:
+	$(RUN) $(TC_PATH)$(TC_PREFIX)gcc $(CFLAGS) -D'VERSION=$(PKG_VERSION)' -o $(SRC_DIR)fdupes $(SRC_DIR)fdupes.c $(SRC_DIR)md5/md5.c $(LIBS)
```
Dr-Bean: Why not use the method as described in the INSTALL document: https://github.com/adrianlopezroche/fdupes/blob/master/INSTALL?

cross/get_iplayer/Makefile
```diff
@@ -0,0 +1,27 @@
+PKG_NAME = get_iplayer
+
+NAME = get_iplayer
```
Dr-Bean: Same comment as above.

cross/get_iplayer/Makefile
```diff
@@ -0,0 +1,27 @@
...
+PKG_DIST_SITE = git://git.infradead.org/get_iplayer.git
```
Dr-Bean: http://www.infradead.org/get_iplayer/html/get_iplayer.html mentions  ftp://ftp.infradead.org/pub/get_iplayer/ as the download site.

cross/rtmpdump/Makefile
```diff
@@ -0,0 +1,26 @@
+PKG_NAME = rtmpdump
+
+NAME = rtmpdump
+PKG_DIST_SITE = git://git.ffmpeg.org/rtmpdump
```
Dr-Bean: http://rtmpdump.mplayerhq.hu/download/rtmpdump-2.4.tar.gz

cross/rtmpdump/Makefile
```diff
@@ -0,0 +1,26 @@
+PKG_NAME = rtmpdump
+
+NAME = rtmpdump
```
Dr-Bean: Same comment as above

Dr-Bean:
Thanks for the addition. I've added some comments for you to look at.
Other than that, probably best that you remove fdupes from this PR, and add it, together with an fdupes SPK via its own PR. We'd end up with unused/unmaintained cross packages otherwise.

cytec:
afaik he doesn't keep a repo and from what i've read i don't think he is going to put one up for that. But you can still download and store the patch in `spksrc` and use the spksrc`s patching ;)

Dr-Bean:
As for the patch, well, good question. You could add the download/unpack/application of the patch in, say, myConfigure. Not sure if I would add a patch of this size directly to spksrc.

richardhopton:
FYI - http://www.openssl.org/source/openssl-1.0.2a.tar.gz no longer works... 

if you need this specific version then the url is now... http://www.openssl.org/source/old/1.0.2/openssl-1.0.2a.tar.gz or https://github.com/openssl/openssl/archive/OpenSSL_1_0_2a.tar.gz

OpenSSL maintain a clone on github too - https://github.com/openssl/openssl with this version being tagged...  https://github.com/openssl/openssl/tree/OpenSSL_1_0_2a

hadess:
All updated in the patch. Right now, I'm not including the KSV patches. They're not needed for get_iplayer, and I'm not interested in providing an rtmpdump for other uses.

If we wanted something that was more generally useful, we'd add a separate cross for rtmpdump KSV, and an SPK for it.

richardhopton:
When running make arch-bromolow I get this error...

wget http://rtmpdump.mplayerhq.hu/download/rtmpdump-2.4.tar.gz/rtmpdump-2.4.tar.gz
http://rtmpdump.mplayerhq.hu/download/rtmpdump-2.4.tar.gz/rtmpdump-2.4.tar.gz:
2015-06-12 14:32:14 ERROR 404: Not Found.

cytec:
@hadess isn't get_iplayer a Pearl script? If so, i think the SPK should depend on the Pearl SPK for installation... I've also added two comments: keep in mind that `PKG_DIST_NAME` gets attached to `PKG_DIST_SITE` automatically when downloading :)

hadess:
> @hadess isn't get_iplayer a Pearl script?

Perl

> If so, i think the SPK should depend on the Pearl SPK for
> installation...

Perl is built into the OS, it's not separate.

> I've also added two comments: keep in mind that PKG_DIST_NAME gets attached to
> PKG_DIST_SITE automatically when downloading :)

Yeah, I noticed that now that this morning's routing problems are fixed.

hadess:
I've been able to build get_iplayer's SPK for qoriq with that PR.

cytec:
@hadess im pretty sure perl isn't part of DSM (anymore?) at least not with 5.X... why would Synology provide a Perl SPK if it's already present?

richardhopton:
Works with bromolow with current code

cross/get_iplayer/Makefile
```diff
@@ -0,0 +1,22 @@
+PKG_NAME = get_iplayer
+PKG_VERS = 2.94
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = ftp://ftp.infradead.org/pub/$(PKG_NAME)/
```
Dr-Bean: Drop the trailing `/` please :)

cross/rtmpdump/Makefile
```diff
@@ -0,0 +1,22 @@
+PKG_NAME = rtmpdump
+PKG_VERS = 2.4
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = http://rtmpdump.mplayerhq.hu/download/
```
Dr-Bean: ...and here ;)

richardhopton:
I can confirm perl isn't installed as part of DSM out of the box

Dr-Bean:
I think Perl was made into a separate package during DSM4.3.

>If we wanted something that was more generally useful, we'd add a separate cross for rtmpdump KSV, and an SPK for it.

I don't see a need for a separate package for that, might as well add it in to the rtmpdump cross package. Mind you, I'm not saying you should add it right now, we might as well wait until it's needed.

@hadess Added two more comments. Please also add a beta tag in the SPK Makefile and add digests for the cross packages (`make digests`)

cytec:
well i can update `rtmpdump` to include KSV Patch once i've got my `ffmpeg` stuff working properly as (if i recall that correctly i was using the patched version there) or if it's needed at some point we can still add the patch later on like @Dr-Bean said

Dr-Bean:
@cytec Didn't realise that you included it in that PR. Is that fork you used up-to-date, or unknown? If it tracks upstream rtmpdump, we can probably use that. Still, shame the patch isn't available as a bare file somewhere (and kept up-to-date, I mean...copying has the big downside it gets outdated)

cytec:
yup was hoping he would put the patch in the repo itself with the latest update...

hadess:
All comments should be all fixed now.

richardhopton:
Have compiled and loaded this onto my DSM 5.2 box. Had to install the perl module as expected, then when trying to use it I got the following error...

INFO: 1 Matching Programmes
WARNING: Please download and run latest installer or install the XML::Simple Perl module for more
 accurate programme metadata.
ERROR: Failed to get version pid metadata from iplayer site

Same version 2.94 works on Windows so it's nothing to do with my network or the BBC feed.

richardhopton:
I have built some additional behaviour onto this branch to enable the get_iplayer webpvr interface to work and be exposed through DSM.  Should I open a second PR?

hadess:
I don't have anything against you taking over, I haven't been able to find a way to get XML::Simple installed without a lot of yak shaving, and I'm already shaving other yaks...

KaraokeStu:
No activity in over 12 months - please reopen or create a new pull request if either of you is still interested @richardhopton or @hadess

hadess:
I'm still interested, but I couldn't get XML::Simple working on the latest DSM releases.

KaraokeStu:
Ok, let me know how it goes and we can look at a new pr once that's all working

PR #1721, opened 2015-05-27
NZG Get 15.0 is stable now

onlineme88:
Updated NZBget version


Dr-Bean:
Someone beat you to it: #1716 ;) So thanks, but I'll close this.

For the future: you also need to update the digests file (use `make digests`, only if you're working on a local spksrc setup), and updating NZBget involves updating configuration files/templates as well.

PR #1722, opened 2015-05-27
Changing Htop SPK version to v1.0.1

terziyski:
Changing Htop version from ```v1.0.3``` to ```v1.0.1``` so the older toolchains can be used to compile the package. This is neccessary for particular architectures on some DSM4.x vesrions.
The package is not compatible with the following toolchains: ```powerpc-4.0```, ```ppc824x-4.2```


Dr-Bean:
Thanks...but I'm missing the `UNSUPPORTED_ARCHS` addition in `cross/htop/Makefile`? To be clear, the purpose of it is to prevent spksrc from building a package for those arches mentioned. As example, see https://github.com/SynoCommunity/spksrc/blob/master/cross/rtorrent/Makefile#L8.

Could you also add a digests file (`make digests` in cross/htop after downloading), and also correct the description in `cross/htop/Makefile` to match the one in `spk/htop/Makefile`? (I changed that because it had an escaping error, didn't fix the cross package though).

terziyski:
Fixed. Just to be clear, I have added in UNSUPPORTED_ARCHS the specific toolchains: ```powerpc-4.0```, ```ppc824x-4.2```. Do I have to put only archs instead or it's fine as it is ?

terziyski:
Fixed. Just to be clear, I have added in UNSUPPORTED_ARCHS the sepecific toolchains: ```powerpc4-0```,  ```ppc824x-4.2```. Is it OK like that or I have to put there just archs ```powerpc```, ```ppc824x``` ?

Dr-Bean:
Just arches please, I would have called it UNSUPPORTED_TOOLCHAINS otherwise ;) 
It only confuses things if you add a version. Example: even if a `powerpc 5.0` toolchain exists, spksrc will still halt compilation because it filters on `powerpc`.

Btw, please take another look at your digests file. It has two sets of digests for the same file.

Dr-Bean:
...and rebase your commits, because you're redoing changes that have already been made.

terziyski:
I hope it's OK now.

Dr-Bean:
Package htop 1.0.1-1 has been published on the repository.

terziyski:
Thanks for your effort.

PR #1726, opened 2015-05-28
Fix 2x superfluous ) and a missing TAB.

kwetal:
This fixes two syntax errors during generation of the INFO file of a package.
Also, I noticed a makefile recipe line that did not start with a TAB.


Dr-Bean:
I don't know that it causes syntax errors, but either way, it should get fixed ;)

PR #1736, opened 2015-06-02
FIX: Debian chroot fails to start up slow scripts

mitom:
The original PR (#1381) was auto closed by GH. All credits to @HakShak.


HakShak:
Sweet, thanks for picking this up!

PR #1741, opened 2015-06-05
Use TT-RSS from Git, as no more releases. Fixes #1739

### LINKED ISSUE #1739: tt-rss now git only, no longer doing periodic releases
See https://tt-rss.org/forum/viewtopic.php?f=10&t=3262
SynoCommunity is using the very old 1.14 release, while there have been 139 commits since then. See https://github.com/gothfox/Tiny-Tiny-RSS/compare/1.14...master
Can the tt-rss spksrc be updated?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/19895569-tt-rss-now-git-only-no-longer-doing-periodic-releases?utm_campaign=plugin&utm_content=tracker%2F282943&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F282943&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nickbroon:
See https://tt-rss.org/forum/viewtopic.php?f=10&t=3262


spk/tt-rss/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = tt-rss
-SPK_VERS = 1.14
-SPK_REV = 7
+SPK_VERS = 20150529
+SPK_REV = 1
```
Dr-Bean: `8`, `SPK_REV` only goes up :)

spk/tt-rss/Makefile
```diff
@@ -12,7 +12,7 @@ DESCRIPTION_FRE = Tiny Tiny RSS est un agrégateur et lecteur web de flux RSS et
 ADMIN_URL = /tt-rss/
 RELOAD_UI = yes
 DISPLAY_NAME = Tiny Tiny RSS
-CHANGELOG = "1. Update Tiny Tiny RSS to 1.14"
+CHANGELOG = "1. Update Tiny Tiny RSS to latest Git as of 20150529"
```
Dr-Bean: Can I propose: `CHANGELOG = Update Tiny Tiny RSS to revision c3dfc1b`?

Dr-Bean:
Thanks, that should speeds things up ;) Does an upgrade from the current published package to this one work?

nickbroon:
I've rebased the branch with your suggested changes.
I did a manual upgrade and it went fine, it prompted to update the tt-rss database on next login, which also went fine. (The usual problem of it blowing away manually installed tt-rss themes and plugins is still there, but that's always been present on upgrade. I should file a bug and look at fixing our upgrade script to not remove these)

Dr-Bean:
Not sure there's a good solution for the themes/plugins with that type of setup. It's kind of hard to implement something like: backup/restore a set of files, *except* tt-rss original files.

Anyway, PR looks okay to me, so thanks for that.

nickbroon:
In the same way that preuninst() moves the config file out way and postuninst() copies it back, it should be possible to mv the plugin/ and themes/ directories out the way and then put them back. I'll looking into this and if appropriate open a new issues/pull request for it. Thanks.

Dr-Bean:
Possible, sure, but the outcome is probably different from what you want.
You backup `plugins` and `themes` during `preupgrade`. During the `upgrade` step, tt-rss may have updated the `plugins` and `themes` dirs. During `postupgrade`, those changes will be overwritten with the old stuff from your backup. That's probably not the result that you want ;)

nickbroon:
Agreed. It would much easier if if tt-rss had separate directories for user plugins/themes and those shipped with the package. I'll perhaps raise this with upstream.

PR #1747, opened 2015-06-08
update HTPC-Manager to choose custom fork

cytec:


spk/htpcmanager/src/installer.sh
```diff
@@ -30,6 +38,9 @@ postinst ()
     # Create a Python virtualenv
     ${VIRTUALENV} --system-site-packages ${INSTALL_DIR}/env > /dev/null
 
+    # Clone the repository and configure autoProcessTV
```
Hellowlol: Whats autoPostprocessTv? Is this some copy paste error from sr or sb?

Hellowlol:
Thanks doing this :)

spk/htpcmanager/src/installer.sh
cytec: yup :)

Hellowlol:
https://github.com/SynoCommunity/spksrc/pull/1747/files#diff-94d669ae2f3faca10e180a3e88c1372eR42

Will this line only clone branch "master"? I thats how i read the git docs but im not git expert, if thats the case it would be a issue as my default branch is master2 and master isnt in use

spk/htpcmanager/src/wizard/install_uifile_fre
```diff
@@ -0,0 +1,22 @@
+[{
+    "step_title": "Choisissez votre source",
+    "items": [{
+        "type": "textfield",
+        "desc": "Paramètres du source à utiliser",
```
Diaoul: -du
+de la

spk/htpcmanager/src/wizard/install_uifile_fre
```diff
@@ -0,0 +1,22 @@
...
+        "subitems": [{
+            "key": "wizard_fork_url",
+            "desc": "URL du source",
```
Diaoul: -du
+de la

spk/htpcmanager/src/wizard/install_uifile_fre
```diff
@@ -0,0 +1,22 @@
...
+            "defaultValue": "git://github.com/styxit/HTPC-Manager.git",
+            "validator": {
+                "allowBlank": false
+            }
+        }, {
+            "key": "wizard_fork_branch",
+            "desc": "Branch du source",
```
Diaoul: -du
+de la

cytec:
@Hellowlol by default it will checkout master yes. but you can adjust it in the install wizard (like you have to do with the url)

Hellowlol:
Thanks, should be no issues then :)

BuSHari:
Hello,

When it will be available?

cytec:
totally forgot about that PR... i've fixed the spelling errors.
@Diaoul @Dr-Bean anything else?

Hellowlol:
I believe there was a discussion regarding psutil but i cant seem to find it. About adding it to the python spk.

BuSHari:
Thank you...
So when do you think it will be available?

Dr-Bean:
@cytec This won't work with package upgrades so a forced reinstall should be added in `preupgrade`: something like https://github.com/SynoCommunity/spksrc/blob/master/spk/salt-master/src/installer.sh#L66.
(Although it's possible to change the approach so a reinstall is not needed, this is the easiest)
Other than that, I don't think I have other comments.

Hellowlol:
Any change this can be merged?

spk/htpcmanager/src/installer.sh
```diff
@@ -30,6 +38,9 @@ postinst ()
     # Create a Python virtualenv
     ${VIRTUALENV} --system-site-packages ${INSTALL_DIR}/env > /dev/null
 
+    # Clone the repository
+    ${GIT} clone -q -b ${wizard_fork_branch:=master} ${wizard_fork_url:=git://github.com/styxit/HTPC-Manager.git} ${INSTALL_DIR}/var/HTPC-Manager
```
Dr-Bean: Imo, if the software can update itself, it makes sense to put this inside an `if [ "${SYNOPKG_PKG_STATUS}" == "INSTALL" ];` statement. The restore during postupgrade would replace the newly-cloned code, so it doesn't do anything worthwhile, and it slows down the upgrade process because repos are cloned with each upgrade.

Sidenote: We do the same with at least some of the other -custom packages (maybe all of them?), so we'll want to use the same approach for those too.

Dr-Bean:
Let's see if we can revive this one, @cytec ;) Not sure if you still plan to work on it, but otherwise, I'll probably pick it up. 

I mentioned in an earlier comment that a forced uninstall would be needed, or some logic to fix updating from version `-1` to `-2`. I'd probably prefer the added logic from an UX point of view, but I suppose it depends on how convoluted it would get.

Dr-Bean:
Additional changes done with bb130a1c5321a34c7f4bce8d31d457fd1c7c20b7, and published it on the repo.

Hellowlol:
Thanks!

PR #1752, opened 2015-06-11
Update syncthing to v0.11.9

### LINKED ISSUE #1674: Add Syncthing 0.11.x
Syncthing has a new major release with lots of improvements and fixes see:

https://github.com/syncthing/syncthing/releases/tag/v0.11.0
https://github.com/syncthing/syncthing/releases/tag/v0.11.1
https://github.com/syncthing/syncthing/releases/tag/v0.11.2

Would love to see it in the repo

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13151652-add-syncthing-0-11-x?utm_campaign=plugin&utm_content=tracker%2F282943&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F282943&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Zetten:
Fixes #1674.

### DIFF
--- a/spk/syncthing/Makefile
+++ b/spk/syncthing/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = syncthing
-SPK_VERS = 0.10.30
-SPK_REV = 3
+SPK_VERS = 0.11.9
+SPK_REV = 4
 SPK_ICON = src/syncthing.png
 DSM_UI_DIR = app
 BETA = 1
--- a/cross/syncthing/Makefile
+++ b/cross/syncthing/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = syncthing
-PKG_VERS = 0.10.30
+PKG_VERS = 0.11.9
 PKG_EXT = tar.gz
 PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/syncthing/syncthing/archive
--- a/spk/syncthing/src/dsm-control.sh
+++ b/spk/syncthing/src/dsm-control.sh
@@ -10,23 +10,22 @@ PATH="${INSTALL_DIR}/bin:${PATH}"
 USER="syncthing"
 SYNCTHING="${INSTALL_DIR}/bin/syncthing"
 CONFIG_DIR="${INSTALL_DIR}/var/"
-PID_FILE="${CONFIG_DIR}/syncthing.pid"
 
 
 start_daemon ()
 {
-    start-stop-daemon -b -o -c ${USER} -S -u ${USER} -m -p ${PID_FILE} -x env HOME=${CONFIG_DIR} ${SYNCTHING} -- --home ${CONFIG_DIR}
+    start-stop-daemon -b -o -c ${USER} -S -u ${USER} -x env HOME=${CONFIG_DIR} ${SYNCTHING} -- --home ${CONFIG_DIR}
 }
 
 stop_daemon ()
 {
-    start-stop-daemon -o -c ${USER} -K -u ${USER} -p ${PID_FILE} -x ${SYNCTHING}
-    wait_for_status 1 20 || start-stop-daemon -K -s 9 -q -p ${PID_FILE}
+    start-stop-daemon -o -c ${USER} -K -u ${USER} -x ${SYNCTHING}
+    wait_for_status 1 20 || start-stop-daemon -K -s 9 -q -x ${SYNCTHING}
 }
 
 daemon_status ()
 {
-    start-stop-daemon -K -q -t -u ${USER} -p ${PID_FILE}
+    start-stop-daemon -K -q -t -u ${USER}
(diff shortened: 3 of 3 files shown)


Diaoul:
You changed mode to 755, please set back to 644

Zetten:
Oops, sorry about that. I was lazy and committed it quickly from my Windows box :disappointed:

All better now.

spk/syncthing/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = syncthing
-SPK_VERS = 0.10.30
-SPK_REV = 3
+SPK_VERS = 0.11.8
+SPK_REV = 1
```
DaftHonk: SPK_REV should only increase

Zetten:
Is SPK_REV not a field tracking updates to each SPK_VERS (i.e. it is reset on each new upstream package version, and reflects changes to the SPK-specific side within that version)? Otherwise the overall package version (0.11.8-1 in this case) holds no information about the package *for that upstream version*. It just says people have modified the SPK Makefile X times in the entirety of its history... Resetting to 1 says unambiguously that this is the first SPK released for upstream version 0.11.8.

This would be more in line with other packaging systems I'm familiar with like .deb (the '[debian_revision](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version)' field) or .rpm (the '[release](http://www.rpm.org/max-rpm/ch-rpm-file-format.html) field) or Arch pkgbuilds (['pkgrel'](https://wiki.archlinux.org/index.php/Arch_packaging_standards#Package_naming)).

I couldn't find anything on the wiki about this, and I'm happy to make the change, but I wanted to make sure as it seems a bit unintuitive to me.

Diaoul:
The package center doesn't use that kind of advanced version tools so it's often confused by complex versions numbers such as dev versions, snapshots, betas, etc.

When the package center is confused, the update isn't detected and you're stuck with an old version of the package.

Another reason is that Synology uses that versioning scheme by appending the build number at the end (same as our SPK_REV) on their packages without reseting it.

Yet another one is that it's easier to do version comparison in database with SPK_REV always increasing as it is split by spkrepo. You don't need to extract all the database and do version comparision in application part, that can be done with a simple query in a reliable way (no "insert_date" comparison).

When Synology will implement a strong version comparison tool in Package Center and document it we will review this. I've already suggested that a few years ago and I've been told that they were working on it recently so that might happen soon (next DSM major?).

Zetten:
All excellent reasons - thanks for the clarification. I didn't realise the Synology-specific differences from everything else I've been used to.

So in this case should I reset it to 3? Or increment it to 4?

Diaoul:
I forgot one reason too. With that it's easier to detect package structure changes and take appropriate migration actions.
See https://github.com/SynoCommunity/spksrc/blob/d97a1063e6de7b08dfb70e92535828afbd97befe/spk/haproxy/src/installer.sh#L88-L96

Diaoul:
Increment to 4

Dr-Bean:
Probably something to put in a FAQ somewhere. It's not the first time it's asked, and all those reasons lined up should answer any questions :)

Zetten:
Done.

cytec:
yup a FAQ would be great... also i was thinking about something like "spksrc in depth" which clears up on a few variables (what they do, how they are used) how downloading from git/svn works (revision based) etc... would be pretty usefull (also for myself as i forget a lot over the time when i don't need it) :D

Kaelber:
I´m not sure, if my issue is right here. I´m using Synology Diskstations DS1512+, 412+, 712+ and some others. This three are x86-models and currently only in the DS712+ model I can choose the installation of Syncthing. In the past, also the two other models are able to install Syncthing over the package center. I don´t know exactly, but it seems since DSM 5.2, is installed in all my machines, there are differences

What can I doe? I´m also search the SPK´s for Syncthing for a manuall installation, but I doesn´t find. Is possible to get a Download adress for this?

Zetten:
Rebased on top of master and tweaked: updated to 0.11.9, and added a second commit fixing the initscript issue discussed in #1674.

Kaelber:
Sorry, but up to now I doesnt read about any solution.

Zetten:
@Kaelber I'm not sure this is the correct place for your question. Someone more familiar with the SynoCommunity backend might be able to help if you raise this as a new issue.

DaftHonk:
Running spk version 0.11.9-4 on a DS1813+ (cedarview-5.1). Auto-updated to v0.11.11 and package center is still showing as running so the initscript fix seems to have worked.

Dr-Bean:
Done with https://github.com/SynoCommunity/spksrc/commit/a7c9609ad4a4f2d665c07b31ec17471d803c6b17. I used your start-stop-status changes, but did the rest separately. 
Package is not yet released.

PR #1757, opened 2015-06-14
fish-shell package, initial release

tmaone:
Initial attempt to bring fish-shell to syno. Have tested the package on avoton (x64) /w latest DSM5.2.


tmaone:
cc @ridiculousfish

spk/fish/src/installer.sh
```diff
@@ -0,0 +1,50 @@
...
+preuninst ()
+{
+    exit 0
+}
+
+postuninst ()
+{
+    # Remove link
+    rm -f ${INSTALL_DIR}
+    rm -f /usr/local/bin/zsh
```
ridiculousfish: zsh ought to be fish?

spk/fish/src/installer.sh
```diff
@@ -0,0 +1,50 @@
...
+{
+    exit 0
+}
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # Put zsh in the PATH
```
ridiculousfish: should read fish, not zsh
also, is it necessary to update /etc/shells?

tmaone:
Sry bout that, used  the zsh template and didn't spot it.

I think the /etc/shells part is not applicable in this case, as the synology system does not allow any other program for a login shell besides ```/bin/sh``` which itself is a busybox shell. And I mean nothing else, not even /bin/bash (which does not even exist in the first place).

Usually we load the intended shell at the end of a ```.profile``` file.

Also, some programs are not readily available in synology, which in turn are required by fish for its completions, such as ```apropos``` and ```manpath```. They can be installed via a 3rd party package manager (```ipkg```).

cross/fish/Makefile
```diff
@@ -0,0 +1,26 @@
+PKG_NAME = fish
+PKG_FULL_NAME = fish-shell
+PKG_EXT = tar.gz
+PKG_DOWNLOAD_METHOD = git
```
Dr-Bean: Why not use the official tarball: http://fishshell.com/files/2.1.2/fish-2.1.2.tar.gz? Probably means you can drop `PKG_FULL_NAME` and `PRE_CONFIGURE_TARGET`.

cross/fish/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+HOMEPAGE = http://fishshell.com
+COMMENT  = fish is a smart and user-friendly command line shell for OS X, Linux, and the rest of the family.
+LICENSE  = GPLv2
+
+PRE_CONFIGURE_TARGET = myPreConfigure
+GNU_CONFIGURE = 1
+CONFIGURE_ARGS = --prefix=$(INSTALL_DIR)$(INSTALL_PREFIX)
+
+ENV += CXXFLAGS="-I$(INSTALL_DIR)$(INSTALL_PREFIX)/include" LDFLAGS="-L$(INSTALL_DIR)$(INSTALL_PREFIX)/lib"
```
Dr-Bean: Please recheck if this line is needed, seems to me these flags are already set as part of tc_vars.mk.

spk/fish/Makefile
```diff
@@ -0,0 +1,25 @@
...
+SPK_VERS = 2.1.2
+SPK_REV = 1
+SPK_ICON = src/ascii_fish.png
+
+DEPENDS = cross/$(SPK_NAME)
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = fish is a smart and user-friendly command line shell for OS X, Linux, and the rest of the family.
+RELOAD_UI = no
+DISPLAY_NAME = fish shell
```
Dr-Bean: Please change to `fish` to keep consistent with the rest.

cross/fish/Makefile
tmaone: official at the moment is quite outdated, with many bugs being fixed since then. soon I think there will be a brand new release. does it make sense to wait for next release and resubmit a pull request which uses the official tarball?

Dr-Bean:
Added a coupl comments. Otherwise, nice work :)

tmaone:
cheers! will revise and push your recommendations later today. :)

cross/fish/Makefile
Dr-Bean: Yep, I'd suggest to wait for the next release, which, hopefully, isn't too far away. For the record, `SPK_VERS` mentions `2.1.2`, and I assumed the git hash is a checkout for that release (right?)

cross/fish/Makefile
tmaone: Yes, I triple checked with the latest avoton toolkit and spksrc. I think due to the configure script of fish-shell, these includes don't get to the actual build step unless specifically defined. Took me quite a bit to realise =)

cross/fish/Makefile
tmaone: correct, with a longer git version which uses git --describe, being : ```fish, version 2.2b1-42-gad23713```

tmaone: I have to double check but I think the next ```2.2``` release is around the corner, maybe a few weeks.
 I will resubmit a pull request with a proper release when available.Thanks

Dr-Bean:
Perfect. I'll leave it to you if you want to keep this open (the new commits could be squashed and force-pushed), or you close this PR and open a new one.

tmaone:
Please leave open, I will commit extra changes and ping this PR when that happens. Cheers.

Dr-Bean:
@tmaone Seeing as fish 2.2.0 was released, maybe you want to revisit this?

Dr-Bean:
@tmaone I wasn't pinged ;) If you can change the last remaining comment I had on DISPLAY_NAME (in the spk's Makefile), I think this is about ready, right?

Dr-Bean:
...and please add digests for the cross package.

KaraokeStu:

Thanks,

Stuart

tmaone:
You are right. I absolutely forgot about this, why not since its almost there, I will work on it and update the PR the soonest possible.

One question, how would you like me to go about external dependencies, specifically apropos from man-db for instance, that the shell uses when processing autocompletions etc? ad-hoc gracefully return (ie not include the lib), or properly introduce it as a dependency, a-la ncurses way?

etcusrvar:
Since this PR is still open, is it possible for this still to be merged? fish itself has been updated significantly in the intervening 3 1/2 years, so the version should be updated.

etcusrvar:
Now that #4366 has been merged, this PR can most likely be closed.

PR #1759, opened 2015-06-15
Update shairport-sync to 2.3.11 and add firewall rule for UDP ports

lost-carrier:
Here we are - continuing from https://github.com/SynoCommunity/spksrc/pull/1617 in my own old fork now...

@Dr-Bean Murphy's law. I was hoping for something like "8303/tcp,6001:6010/udp", but obviously that kind of mix-up was not possible, so I had to split it up into two rules (factory default "AirPlay" rule does about the same). But after all there was an issue in my installer. The firewall rules were not replaced with an package update, so I could try whatever syntax I want - it just wasn't executed... ;-) ...anyhow - fixed now.

@mikebrady thanks, but pure Synology bulls**t here... ;-) Anyhow: some version tag for 2.3.4 would be nice. Using the development branch for now...


cross/shairport-sync/Makefile
```diff
@@ -1,9 +1,9 @@
 PKG_NAME = shairport-sync
-PKG_VERS = 2.3.0
+PKG_VERS = 2.3.4
 PKG_EXT = zip
-PKG_DIST_NAME = $(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = https://github.com/mikebrady/shairport-sync/archive
-PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+PKG_DIST_NAME = development.$(PKG_EXT)
+PKG_DIST_SITE = https://github.com/mikebrady/shairport-sync/archive/
```
Dr-Bean: No trailing slashes please

spk/shairport-sync/Makefile
```diff
@@ -1,5 +1,5 @@
 SPK_NAME = shairport-sync
-SPK_VERS = 2.3.0
+SPK_VERS = 2.3.4
 SPK_REV = 1
```
Dr-Bean: `SPK_REV` goes up with new releases.

spk/shairport-sync/src/installer.sh
```diff
@@ -59,21 +59,29 @@ preupgrade ()
...
+    if [ "${SYNOPKG_PKG_STATUS}" == "UNINSTALL" ]; then
+        ${SERVICETOOL} --remove-configure-file --package ${PACKAGE}.sc >> /dev/null
+    fi
 
     # Save some stuff
     rm -fr ${TMP_DIR}/${PACKAGE}
     mkdir -p ${TMP_DIR}/${PACKAGE}
     mv ${INSTALL_DIR}/var ${TMP_DIR}/${PACKAGE}/
-
+
```
Dr-Bean: Remove unnecessary spaces please

spk/shairport-sync/src/installer.sh
```diff
@@ -59,21 +59,29 @@ preupgrade ()
...
 {
+    # Add firewall config
+    ${SERVICETOOL} --install-configure-file --package ${FWPORTS} >> /dev/null
+
     # Restore some stuff
     rm -fr ${INSTALL_DIR}/var
     mv ${TMP_DIR}/${PACKAGE}/var ${INSTALL_DIR}/
     rm -fr ${TMP_DIR}/${PACKAGE}
-
+
```
Dr-Bean: Remove unnecessary spaces

Dr-Bean:
@lost-carrier What I think I'm missing here is the actual configuration file, forcing Shairport to use a certain set of ports for Airplay.

The thing is, you don't want to use the port 6000-6010. That causes conflicts with Synology's Airplay (part of AudioStation iirc), see https://github.com/SynoCommunity/spksrc/wiki/Synology-Used-Ports. 
Meaning: people can only have AudioStation installed _or_ Shairport-sync installed, but never both. That doesn't seem right, or necessary, to me.

So, instead, I'd propose to add a configuration file in `${INSTALL_DIR}/var/shairport-sync.conf`. I assume there's a way to tell the shairport binary to use a particular conf file (@mikebrady? couldn't find it in the README, but I only skimmed it). In that configuration file, use the tcp port as starting port+10, so 8303-8313.

mikebrady:
Yep, use a command line option list this: `-c /path/to/configfile`. The documentation all needs a good update...

lost-carrier:
@Dr-Bean Hmm... for some reason I had both installed. Not sure what went wrong there.
...anyhow: done now. Is there some more fancy way for adding the config file? Those echos is just the first thing I found with a search over this repo.

@mikebrady thanks - directly switched to 2.3.5... ;-)

spk/shairport-sync/src/installer.sh
```diff
@@ -59,6 +61,11 @@ preupgrade ()
 {
     # Stop the package
     ${SSS} stop > /dev/null
+    
+    # Remove firewall config
```
Dr-Bean: I don't know that there's a reason to do this here? It should be sufficient to add the rules on `postinst` and remove them during `preuninst`.

Dr-Bean:
@lost-carrier Apologies for the delay in answering, must've missed this somehow. Yes, there's a better way to add a configuration file. See the couchpotato-custom package for example. In short: 
- [Create your configuration file](https://github.com/SynoCommunity/spksrc/blob/master/spk/couchpotatoserver-custom/src/settings.conf) as `shairport-sync/src/shairport-sync.conf`
- [Refer to it from the Makefile](https://github.com/SynoCommunity/spksrc/blob/master/spk/couchpotatoserver-custom/Makefile#L40), so that it's placed in `$(INSTALL_DIR)/var` with the same name.

spk/shairport-sync/src/installer.sh
lost-carrier: Not sure how those are executed, but it was meant for allowing to change ports within upgrades...?

lost-carrier:
@Dr-Bean done...

@mikebrady tried to upgrade to 2.3.8, but everything beyond 2.3.6 does not want to start. Went thru the release notes, but could not find anything particular.

When starting, shairport always reports...
```
diskstation> bin/shairport-sync -v
startup
Failed to bind to address 0.0.0.0.
Failed to bind to address ::.
could not bind any listen sockets!
```
...no matter if you specify config file or but. BTW: checked port-ranges - both 8303-range as well as default range should be available...

...did I miss anything?

spk/shairport-sync/Makefile
```diff
@@ -34,3 +34,4 @@ include ../../mk/spksrc.spk.mk
 shairport_extra_install:
 	install -m 755 -d $(STAGING_DIR)/var
 	install -m 644 src/alsa.conf $(STAGING_DIR)/share/alsa/alsa.conf
+	install -m 600 src/shairport-sync.conf $(STAGING_DIR)/var/shairpot-sync.conf
```
Dr-Bean: Please set this to 644 (which is default...for some reason, Couch doesn't adhere to that unfortunately, we'll need to fix that)

spk/shairport-sync/src/installer.sh
Dr-Bean: Postinst is run during a package upgrade as well (check Synology's dev guide), so no need for that. Add/update rules during `postinst`, and remove them during `preuninst` (with a check during `preuninst`, just as you have now)

Dr-Bean:
Added two comments. Could you also please rebase your PR and squash commits? You're removing previously added digests, and have a lot of corrective commits :)

cross/shairport-sync/Makefile
```diff
@@ -13,7 +13,7 @@ LICENSE  = https://raw.githubusercontent.com/mikebrady/shairport-sync/$(PKG_VERS
 
 GNU_CONFIGURE = 1
 
-CONFIGURE_ARGS = --prefix=$(STAGING_INSTALL_PREFIX) --with-alsa --with-ssl=openssl --with-tinysvcmdns --without-initscript --with-piddir=$(INSTALL_DIR)/var
+CONFIGURE_ARGS = --prefix=$(STAGING_INSTALL_PREFIX) --with-alsa --with-ssl=openssl --with-tinysvcmdns --without-initscript --without-configfiles --with-piddir=$(INSTALL_DIR)/var
```
Dr-Bean: Just a question: you're specifying `--without configfiles` here...this doesn't refer to `shairport-sync.conf`?

spk/shairport-sync/Makefile
mikebrady: That message usually happens if another instance of Shairport Sync is running... Is there any possibility?

lost-carrier:
@mikebrady nope - no other instance was running. Also no other application was blocking those ports. Ports were free - checked with netstat. Same procedure that worked for 2.3.5 and 2.3.6 did not work for 2.3.7 and 2.3.8...

cross/shairport-sync/Makefile
lost-carrier: It does. However: it's about installing a default config file to /etc/something which will not work when cross compiling...

mikebrady:
I think I see the problem – you should omit the `port = 8303;` line.
The "port" is the port Shairport Sync listens on – by default it is port 5000. The UDP ports are used when Shairport Sync is establishing a play connection for data, control and timing signals respectively. Port 5000 seems to be the default but, in any case, it must be different from the UDP port range. If I was you, I'd just drop the line and let it go to the default.

lost-carrier:
When I do *not* specify the config file like "-c var/shairport-sync.conf", it should use the default 5000, right? (...because that doesn't work either)

mikebrady:
If you don't specify the config file, it will look in `/etc/shairport-sync.conf`. If that doesn't exist, it will use the default port 5000, yes. The defaults would also be used for the UDP ports, which are 6001 upwards.

If you do `-vv` it will print out the option values.

lost-carrier:
Thanks - I will check that -vv output this evening...

lost-carrier:
@mikebrady Looks like shairport has a problem with the reading the config file:
```
diskstation> bin/shairport-sync -vv -c var/shairport-sync.conf
startup
statistics_requester status is 0.
daemon status is 0.
rtsp listening port is 5000. <--- !!!
udp base port is 6001. <--- !!!
udp port range is 100.
Shairport Sync player name is "Shairport Sync on diskstation".
Audio Output name is "(null)".
on-start action is "(null)".
on-stop action is "(null)".
wait-cmd status is 0.
mdns backend "(null)".
userSuppliedLatency is 0.
AirPlayLatency is 88200.
iTunesLatency is 99400.
forkedDaapdLatency is 99400.
stuffing option is "0".
resync time is 2205.
allow a session to be interrupted: 0.
busy timeout time is 120.
tolerance is 88 frames.
password is "(null)".
ignore_volume_control is 0.
audio backend desired buffer length is 6615.
audio backend latency offset is 0.
Configuration file name "var/shairport-sync.conf". <--- !!!
Failed to bind to address 0.0.0.0.
Failed to bind to address ::.
could not bind any listen sockets!
```
...same result for...
```
diskstation> bin/shairport-sync -vv --configfile=var/shairport-sync.conf
```
...while...
```
diskstation> bin/shairport-sync -p 8303
```
...appears to be working.

Did something change with the parsing of that file or the parameter?

BTW:
```
diskstation> cat var/shairport-sync.conf
general =
{
	port = 8303;
	udp_port_base = 8303;
	udp_port_range = 10;
};

diskstation> cat /etc/shairport-sync.conf
cat: can't open '/etc/shairport-sync.conf': No such file or directory
```

mikebrady:
Thank for the update. I've wondered about the file path `var/shairport-sync.conf``. As it stands it's a relative path. Should it not be an absolute path, like `/var/...`?
AFAIR there was no change in file name parsing.

mikebrady:
Sorry about the formatting. I'm in a phone, away from my machines.

lost-carrier:
n.p. - I got what you wanted to say... ;-)

...but this time it *is* relative, since the the config-file is in the package-dir.

mikebrady:
Sure, but when Shairport Sync is actually *running*, where will that file be?

lost-carrier:
Not sure, what you mean, but...
```
diskstation> pwd
/var/packages/shairport-sync/target
diskstation> ls -la ./bin/shairport-sync
-rwxr-xr-x    1 root     root        125528 Aug  4 23:36 ./bin/shairport-sync
diskstation> ls -la ./var/shairport-sync.conf
-rw-r--r--    1 root     root            75 Aug  5 20:12 ./var/shairport-sync.conf
```

mikebrady:
That's great. As an experiment, if you give Shairport Sync the full pathname for the configuration file, will it start working, i.e. `--configfile=/var/packages/shairport-sync/target/var/shairport-sync.conf`?

If so, then the problem is that Shairport Sync is not making up a proper file path for a relativey-specified file. Also if so, as a workaround, would that absolute file path be stable across all installations? In other words, if it works experimentally now, could it be a general solution?

lost-carrier:
Full path was correct, but unfortunately didn't work either...

lost-carrier:
Nevertheless: not sure if this path is always the same, but: yes - it should be possible to construct that it int he script if needed...

lost-carrier:
@Dr-Bean done! (...no sure about that rebasing-stuff, though)

mikebrady:
Could you let us see the `-vv` log please? Is it the same as before?

lost-carrier:
Yep...

```
diskstation> ls -la /var/packages/shairport-sync/target/var/shairport-sync.conf
-rw-r--r--    1 root     root            75 Aug  5 20:12 /var/packages/shairport-sync/target/var/shairport-sync.conf
diskstation> bin/shairport-sync -vv --configfile=/var/packages/shairport-sync/target/var/shairport-sync.conf
startup
statistics_requester status is 0.
daemon status is 0.
rtsp listening port is 5000.
udp base port is 6001.
udp port range is 100.
Shairport Sync player name is "Shairport Sync on diskstation".
Audio Output name is "(null)".
on-start action is "(null)".
on-stop action is "(null)".
wait-cmd status is 0.
mdns backend "(null)".
userSuppliedLatency is 0.
AirPlayLatency is 88200.
iTunesLatency is 99400.
forkedDaapdLatency is 99400.
stuffing option is "0".
resync time is 2205.
allow a session to be interrupted: 0.
busy timeout time is 120.
tolerance is 88 frames.
password is "(null)".
ignore_volume_control is 0.
audio backend desired buffer length is 6615.
audio backend latency offset is 0.
Configuration file name "/var/packages/shairport-sync/target/var/shairport-sync.conf".
Failed to bind to address 0.0.0.0.
Failed to bind to address ::.
could not bind any listen sockets!
```

Dr-Bean:
I think I can reproduce this with 2.3.8, and I might have some new info:
- With `/usr/local/shairport-sync/bin/shairport-sync -vv --configfile=/usr/local/shairport-
sync/var/shairport-sync.conf`, no start, same error as @lost-carrier.
- With `/usr/local/shairport-sync/bin/shairport-sync -vv`, and `/etc/shairport-sync.conf` doesn't exist, no start, same error (presumably because port 5000 is in use)

Now, I created a config file at `/etc/shairport-sync.conf`, using port 8310/udp 8311 to distinguish it from the other file.
- With `/usr/local/shairport-sync/bin/shairport-sync -vv`: starts fine on port 8310. Reports `Configuration file name "/etc/shairport-sync.conf"`.
- With `/usr/local/shairport-sync/bin/shairport-sync -vv  --configfile=/usr/local/shairport-
sync/var/shairport-sync.conf`: starts fine on port 8310. Reports `Configuration file name "/usr/local/shairport-sync/var/shairport-sync.conf"`.
Expected result: should start on port 8303, not 8311.

So it seems that shairport-sync doesn't want to work with a specified config file if `/etc/shairport-sync` doesn't exist first.

mikebrady:
Thanks for this and for your investigations! I've just pushed an update – 2.3.9 – which [I believe] fixes the problem of using a relative path name to specify the location of the configuration file and which shows in the log file (with `-vv`) the actual absolute file path used when a relative path is specified. So, you should be able to use `--configfile=var/shairport-sync.conf` and it should work (and you should be able to see what file it actually resolves to). It works now for me with your configuration file:
```
mike@ubuntu:~/shairport-sync$ ./shairport-sync -vv --configfile=var/shairport-sync.conf
Looking for the configuration file "var/shairport-sync.conf".
Looking for configuration file at full path "/home/mike/shairport-sync/var/shairport-sync.conf"
startup
statistics_requester status is 0.
daemon status is 0.
rtsp listening port is 8303.
udp base port is 8303.
udp port range is 10.
Shairport Sync player name is "Shairport Sync on ubuntu".
Audio Output name is "(null)".
on-start action is "(null)".
on-stop action is "(null)".
wait-cmd status is 0.
mdns backend "(null)".
userSuppliedLatency is 0.
AirPlayLatency is 88200.
iTunesLatency is 99400.
forkedDaapdLatency is 99400.
stuffing option is "0".
resync time is 2205.
allow a session to be interrupted: 0.
busy timeout time is 120.
tolerance is 88 frames.
password is "(null)".
ignore_volume_control is 0.
audio backend desired buffer length is 6615.
audio backend latency offset is 0.
configuration file name "var/shairport-sync.conf" resolves to "/home/mike/shairport-sync/var/shairport-sync.conf".
metdata enabled is 0.
metadata pipename is "(null)".
get-coverart is 0.
avahi: avahi_register.
avahi: register_service.
Avahi without metadata
```
If this doesn't work for you, could you check file permissions on the intervening directories – i.e. I'm just wondering if Shairport Sync would have permission to see into the directory containing the configuration file, as distinct from having permission to read the file itself. It's a long shot.

Another thing is that, as of 2.3.8, two things have happened with the `--with-initscript`. First, it's been renamed from `--with-initscript` to `--with-systemv` and second, it is no longer automatically included, so you don't have to explicitly exclude it with `--without-systemv`. So, you can drop the `--without-initscript` from the configuration options.

lost-carrier:
Thanks @mikebrady ! I will try 2.3.9 (and 2.3.10) later. Unfortunately I will have only very limited time this evening because I'm heading off for a holiday off-the-grid tomorrow... so plz excuse if there's no reaction during the next two weeks...

mikebrady:
Cool. Happy holidays.

lost-carrier:
Went directly to 2.3.11 - works like a charm! ;-) Thanks @mikebrady !

CU!

mikebrady:
Excellent! Thanks for your help finding the problem.

Dr-Bean:
> (...no sure about that rebasing-stuff, though)

`git rebase -i master` . Then follow instructions, e.g. use `fixup` and `reword` to squash commits and reword commits into a logical structure.
The reason this happens is because you executed `git pull [origin/master]` in this branch. If you do that, follow it up with a rebase.

lost-carrier:
Ups, sorry - must have missed somehow that this is still open...

@Dr-Bean did something ...better now?

...and BTW: we recently got complaints over here https://github.com/mikebrady/shairport-sync/issues/146 that shairport-sync dies with a Segfault on DSM 6.0 beta. Am I correct, that there's new toolchains to come and recompile with?

Dr-Bean:
You're still removing the digests from `libdaemon`, that shouldn't happen. I might correct it during merging the PR, but I wouldn't mind if you fixed that. Otherwise, PR should be good to go.

As for the issue you referenced, that looks like it's this one: https://github.com/SynoCommunity/spksrc/issues/1877. Yours has more info though ;) Let's keep that discussion centered there.

lost-carrier:
@Dr-Bean done.

Dr-Bean:
shairport-sync 2.3.11-2 has been published on the repository. Thanks @lost-carrier :)

PR #1760, opened 2015-06-16
Update tt-rss and preserve local themes/plugins

nickbroon:
Bump to rev 11 is a result of my local testing upgrading and needing increase the rev number.


spk/tt-rss/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = tt-rss
-SPK_VERS = 20150529
-SPK_REV = 8
+SPK_VERS = 20150616
+SPK_REV = 11
```
Dr-Bean: Set to `9` please :)

spk/tt-rss/src/installer.sh
```diff
@@ -125,6 +125,11 @@ preupgrade ()
     mkdir ${TMP_DIR}/${PACKAGE}/feed-icons/
     mv ${WEB_DIR}/${PACKAGE}/feed-icons/*.ico ${TMP_DIR}/${PACKAGE}/feed-icons/
 
+    mkdir ${TMP_DIR}/${PACKAGE}/plugins.local
```
Dr-Bean: Probably cleaner to just do:
```
mv ${INSTALL_DIR}/plugins.local ${TMP_DIR}/${PACKAGE}/
mv ${INSTALL_DIR}/themes.local ${TMP_DIR}/${PACKAGE}/
```
...and put them back the same way ;)

Dr-Bean:
Looks good to me, added two comments. Thanks for the implementation @gothfox :)

nickbroon:
Thanks for feedback. I've pushed a new version that addresses your comment.

I'm actually surprised that @gothfox implemented this, as it would seem to largely benefit packaged installations, and he appears to only care about raw git installations where this would appear to be of little benefit.

I've also been attempting to get the update notice in tt-rss to work with packaged installations, or more specifically installations that don't have the .git directory, but @gothfox has so far not been interested in  the approach I attempted (In his usual abusive and condescending manner) :-(
I'll continue to look at how best to do this. I'm reluctant to maintain a patch against tt-rss that I can't upstream, so simply installing the .git/refs/heads/master file as part of the package might be best, though currently the Git PKG_DOWNLOAD_METHOD generates a tarball that does not contain the .git directory, so it's not easily available.

Dr-Bean:
Could you clarify the last part of your comment? As far as I know, tt-rss does not have a built-in updater, so are you building an updater into tt-rss, or do you want just the notifications to work within our tt-rss package?

If the latter, why? Getting a notification of a newer version without being able to influence it (i.e. updating) gives the wrong signal to users [of our packages]: The only course of action is to open an issue on our repo to update the package. However, it's not feasible for us to release new packages on an almost daily basis (looking at the recent commits upstream). 
All in all, I don't see a point in adding notifications. If you do want to keep up-to-date with upstream, I would suggest not to use the package, and instead, install from source.

If the former: obviously it's up to upstream whether or not something like that will be accepted, but from my point of view, I wouldn't mind it. Couchpotato, Headphones et al use git to check/update on a regular basis, and that works just fine. We can probably figure out a way to bring back the `.git` dir for this purpose if needed.

If I've misunderstood you, let me know.

nickbroon:
Correct, it's just update notification I'd like to get working. As you say tt-rss does not have a built in updater.
With the update notification mechanism working I'd be able to know when i need to submit updates for our packages, but I would propose disabling it (CHECK_FOR_UPDATES in config.php) in the default installation to address your concerns about most package users.

Dr-Bean:
As you say, we would disable it in the package, so there's no need for it on our behalf. 
If you're interested in keeping your local install up-to-date, install from source. 
If you want to keep the package in general up to date (e.g. when interesting updates are implemented), then do as we do for other packages: simply keep an eye on upstream. You could even use an [RSS feed](https://github.com/gothfox/Tiny-Tiny-RSS/commits/master.atom) for that ;)

nickbroon:
If I'm using git source for my local install I'll be less inclined to keep the package up to date
If I was using a packaged version with a working update notifier that I'd enabled so I see the icon I'd more inclined to keep the package periodically updated. :-)
I already subscribe to the github RSS feed for tt-rss commits.

cytec:
Well nobody said you can't do this. But i'm with @Dr-Bean here. Including an update notification function (even if disabled by default) without the option to actually update will get users confused.

Also it would get update notifications really frequently as (if i seen that correctly) there are like daily commits or even several commits a day. 
Personally i would go with the tagged releases anyway instead of git checkouts (if those are still updated/used) https://github.com/gothfox/Tiny-Tiny-RSS/releases

nickbroon:
As mentioned here tt-rss is no longer making releases. https://github.com/SynoCommunity/spksrc/commit/c4c9c7c81ad594a1ca505895b0467baa35a60178

PR #1765, opened 2015-06-19
Update haproxy, support gzip, fix <1024 ports

### LINKED ISSUE #1704: Support gzip for haproxy => save bandwith + time
```
make USE_ZLIB=yes
```

If haproxy is compiled using USE_ZLIB=yes it is possible to use gzip to compress a lot of data.
For example you could add this to your default config section:

```
 compression algo gzip
 compression type text/html text/plain text/css
```

There are many more compression types available.

### LINKED ISSUE #1730: Haproxy config doesn't support binding to privileged ports
A couple small tweaks need to be made to support this:

Raise privileges at runtime to bind to low ports ->
/var/packages/haproxy/scripts/start-stop-status:
```
# Others
PYTHON_DIR="/usr/local/python"
PATH="${INSTALL_DIR}/bin:${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:/usr/local/bin:/bin:/usr/bin:/usr/syno/bin"
RUNAS="root"   <----Modification
PYTHON="${INSTALL_DIR}/env/bin/python"
HAPROXY="${INSTALL_DIR}/sbin/haproxy"
PID_FILE="${INSTALL_DIR}/var/haproxy.pid"
CFG_FILE="${INSTALL_DIR}/var/haproxy.cfg"
```

Drop privileges for security->
haproxy.cfg.tpl:
```
global
        daemon
        maxconn 256
        user haproxy <----Modification
        log localhost user info
        spread-checks 10
        tune.ssl.default-dh-param 2048
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/19006515-haproxy-config-doesn-t-support-binding-to-privileged-ports?utm_campaign=plugin&utm_content=tracker%2F282943&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F282943&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1782: [HAPROXY] 1.5.14 Security Update...
Can you update the SynoCommunity HAProxy Package for security reasons ? 

>July, 3rd, 2015 : 1.5.14 : fixes an information leak vulnerability (CVE-2015-3281)

>A vulnerability was found when HTTP pipelining is used. In some cases, a client might be able to cause a buffer alignment issue and retrieve uninitialized memory contents that exhibit data from a past request or session. I want to address sincere congratulations to Charlie Smurthwaite of aTech Media for the really detailed traces he provided which made it possible to find the cause of this bug. Every user of 1.5-dev, 1.5.x or 1.6-dev must upgrade to 1.5.14 or latest 1.6-dev snapshot to fix this issue, or use the backport of the fix provided by their operating system vendors. CVE-2015-3281 was assigned to this bug. Code and changelog are available here as usual.

source : http://www.haproxy.org/

Thanks for your work...

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23748543-haproxy-1-5-14-security-update?utm_campaign=plugin&utm_content=tracker%2F282943&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F282943&utm_medium=issues&utm_source=github).
</bountysource-plugin>

piejanssens:
Fixes #1704
Fixes #1730 
Fixes #1782
Updates haproxy to latest stable version (from 1.5.11 to 1.5.14)

### DIFF
--- a/cross/haproxy/Makefile
+++ b/cross/haproxy/Makefile
@@ -1,14 +1,14 @@
 PKG_NAME = haproxy
-PKG_VERS = 1.5.12
+PKG_VERS = 1.5.14
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = http://haproxy.1wt.eu/download/1.5/src
+PKG_DIST_SITE = http://www.haproxy.org/download/1.5/src
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 DEPENDS = cross/openssl cross/zlib
 
 MAINTAINER = SynoCommunity
-HOMEPAGE   = http://haproxy.1wt.eu/
+HOMEPAGE   = http://www.haproxy.org
 COMMENT    = HAProxy is a free, very fast and reliable solution offering high availability, load balancing, and proxying for TCP and HTTP-based applications.
 LICENSE    = GNU GPL 2
 
--- a/cross/haproxy/digests
+++ b/cross/haproxy/digests
@@ -1,3 +1,3 @@
-haproxy-1.5.12.tar.gz SHA1 c488f7f108ddefb558bd86e7cbc1b71636ada856
-haproxy-1.5.12.tar.gz SHA256 6648dd7d6b958d83dd7101eab5792178212a66c884bec0ebcd8abc39df83bb78
-haproxy-1.5.12.tar.gz MD5 4b94b257f16d88c315716b062b22e48a
+haproxy-1.5.14.tar.gz SHA1 159f5beb8fdc6b8059ae51b53dc935d91c0fb51f
+haproxy-1.5.14.tar.gz SHA256 9565dd38649064d0350a2883fa81ccfe92eb17dcda457ebdc01535e1ab0c8f99
+haproxy-1.5.14.tar.gz MD5 ad9d7262b96ba85a0f8c6acc6cb9edde
--- a/spk/haproxy/Makefile
+++ b/spk/haproxy/Makefile
@@ -1,5 +1,5 @@
 SPK_NAME = haproxy
-SPK_VERS = 1.5.12
+SPK_VERS = 1.5.14
 SPK_REV = 18
 SPK_ICON = src/haproxy.png
 DSM_UI_DIR = app
@@ -14,7 +14,7 @@ DESCRIPTION_FRE = HAProxy est un produit libre, performant et fiable qui permet 
 ADMIN_PORT = 8280
 RELOAD_UI = yes
 DISPLAY_NAME = HAProxy
-CHANGELOG = "1. Update to 1.5.12<br>2. Added gzip compression support<br>3. Fix listening on <1024 ports"
+CHANGELOG = "1. Update to 1.5.14<br>2. Added gzip compression support<br>3. Fix listening on <1024 ports"
 
 HOMEPAGE   = http://haproxy.1wt.eu/
 LICENSE    = GNU GPL 2
--- a/spk/haproxy/Makefile
+++ b/spk/haproxy/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = haproxy
 SPK_VERS = 1.5.14
-SPK_REV = 18
+SPK_REV = 19
 SPK_ICON = src/haproxy.png
 DSM_UI_DIR = app
 


Diaoul:
LGTM

Dr-Bean:
Does the haproxy.cfg change work for existing installations? Seems like it might be overwritten during postupgrade, no?

piejanssens:
Good point. Where can we add the new line for haproxy.cfg for existing installations?

piejanssens:
I guess I could use "postupgrade ()" in installer.sh.
Would this work?https://github.com/piejanssens/spksrc/commit/46ed6cd33f32caa902f27cd887d1a0959863ba82

Diaoul:
True, that would cause some trouble. Need to add a migration script.

piejanssens:
@Diaoul Should I undo my last commit?

Diaoul:
Yes please remove this and make a migration script that will add the appropriate config section like migrations here: https://github.com/SynoCommunity/spksrc/blob/master/spk/haproxy/src/installer.sh#L88-L96

piejanssens:
@Diaoul @Dr-Bean Done :+1:

piejanssens:
It does add the correct line in the correct spot, however, I couldn't get a TAB printed out by sed, so it hasn't got the correct indentation. I tried yesterday for over an hour using various methods.. 
Do you guys know of an spk where sed was already used to print a TAB?

MattPark:
print the tab with \t
if it doesn't work then you need to add -r like `sed -i -r 's/tab/\t/'

piejanssens:
The \t is not recognised as a tab in the 'replace by' part of sed.
Can you try it yourself with a test file on your syno ?

spk/haproxy/src/haproxy.cfg
```diff
@@ -1,4 +1,5 @@
 global
+    user haproxy
```
Dr-Bean: Seems like it should be a 4-space tab here, Github converts it to 8.

Dr-Bean:
This might work: `awk 'NR == 2 {print "\tuser haproxy"} {print}' haproxy.cfg && mv haproxy.cfg.new haproxy.cfg`
Awk has it's downsides (...like no in-place editing, hence the `mv`), but it can pass a `\t` ;)

piejanssens:
Got it! phew :smile:

Dr-Bean:
That's not a tab?

piejanssens:
@Dr-Bean It sure is. Used the tab key on my keyboard after the backslash. 
Tested the bash script on my syno and the result was a correctly tabbed entry in the haproxy.cfg file.

Dr-Bean:
Yep, that works. Two things remaining: Make sure you entered a tab in the `haproxy.cfg` file (see my open comment on the code), because your editor put spaces in there as well. This handles new installations of the package.
You'll also need to update `haproxy.cfg.tpl` in the migration script, otherwise the change will be reverted as soon as the config is updated via the haproxy UI.

piejanssens:
Indeed, forgot about the default haproxy.cfg. I believe this file is actually copied to haproxy.cfg.tpl, so there is no haproxy.cfg.tpl file itself.

piejanssens:
@Dr-Bean I think we're good now :smile:

piejanssens:
Any idea when this would be merged and deployed?

Dr-Bean:
LGTM, so it'll be merged when possible. We don't give out ETA's...spare time and all.

piejanssens:
@Dr-Bean @Diaoul Guys, can you check something out for me please?
I compiled this package to install on my Syno (DS413), but when I install it using the package manager (manual install) I get the message "Invalid file format".. I have no clue what could be wrong. The architecture is correct and there was no error whilst building the spk...

Here is my package: https://dl.dropboxusercontent.com/u/3691492/haproxy_qoriq-5.1_1.5.14-18.spk

MattPark:
I've been getting that for a lot of packages lately.  It still seems to
install just fine though

piejanssens:
In my case it didn't install unfortunately.
Btw the current haproxy pkg has version 1.5.12-18. 
Is the 18 the spk revision? Because I bumped from 17 to 18 in my commits. 
So I'm confused why the repo would contain spk revision 18 already...

Dr-Bean:
I think @Diaoul updated the package to fix something, but the update wasn't pushed to the repo. @piejanssens, you can set the version to 19...or we do that after merging this PR, either way works.

As for the package installation failures, that's a browser issue then? I'd suggest to open a ticket with Synology for that.

piejanssens:
@Dr-Bean Bumped to 19 - good to go. Don't bother with the installation issues I had, it was a problem caused by my VM (Parallels) while copying the archive over to my host system (OS X).

Diaoul:
Compiling now, this should take quite some time.

Diaoul:
Published.

PR #1767, opened 2015-06-19
remove python from boost

gkroone:
Remove Python dependency. If boost_python is required, it needs to be added as dependency to the calling package. having the dependency in the boost package just results in building python for packages that don't need it.

### DIFF
--- a/cross/boost/Makefile
+++ b/cross/boost/Makefile
@@ -5,7 +5,7 @@ PKG_DIR = $(PKG_NAME)_$(subst .,_,$(PKG_VERS))
 PKG_DIST_NAME = $(PKG_DIR).$(PKG_EXT)
 PKG_DIST_SITE = http://sourceforge.net/projects/$(PKG_NAME)/files/$(PKG_NAME)/$(PKG_VERS)
 
-DEPENDS = cross/bzip2 cross/zlib cross/python
+DEPENDS = cross/bzip2 cross/zlib
 
 HOMEPAGE = http://www.boost.org/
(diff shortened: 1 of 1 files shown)


gkroone:
@Dr-Bean, @Diaoul, I realize the subject of this PR is not very descriptive so might be overlooked, but it does save some compile time for packages depending on boost.(currently libtorrent and domoticz are the only ones I beleive)

Dr-Bean:
You can change the title of the issue to something more descriptive ;)

On the subject: isn't python is needed to compile boost correctly for libtorrent? I assume it's why it's added in the first place.

Diaoul:
See 8418449b82c95a49aab956154f302855d789a087 and 8a9e3dc75f31e2de02d191084ed8358241dd88a7.
Yes it is required for libtorrent, I don't think adding the dependency on the parent package is a good solution as it results in different behaviors for the same cross/

Dr-Bean:
Well, you could add a separate `boost-without-a-python-dependency` cross package? :P

Diaoul:
At least the PR should add some documentation and fix libtorrent.

gkroone:
A while back, domoticz required boost_python to be built, and i could not get boost python to build unless i put the python dependency of domoticz before the cross/boost dependency. There is already a different behaviour for the same cross/boost, as parent package specifies what boost modules are required. 

I dont know anything about libtorrent other then that it seems to be broken for some time already, and the only other package that depends on it is domoticz.

Dr-Bean:
Should no longer be necessary.

PR #1773, opened 2015-06-27
Improved homeassistant package

wind-rider:
Fixed the git revision to last homeassistant release
Updated requirements.txt --> Chromecast works now :)
Added homeassistant cross package so that it will be downloaded only once
Modified spksrc.download.mk to recursive checkout submodules

Please note that this is a beta package. I tested it with Hue and ChromeCast. Other modules I can't test, so please try to contribute if something doesn't work. For bugs in Home Assistant (if not in the Synology package), please file a bug report at https://github.com/balloob/home-assistant/issues.

The default password is empty, just click login. You can make settings in the `/usr/local/homeassistant/var/configuration.yaml` file. The system creates a default configuration.yaml file at the first launch.

Debugging help:
- if home-assistant doesn't start at all, remove the -b option in the `start` section of `/var/packages/homeassistant/scripts/start-stop-status` and run `/var/packages/homeassistant/scripts/start-stop-status start` from the ssh command line. Look for any error messages.
- If a specific module does not load, check whether the dependencies are installed. Run
  `export INSTALL_DIR=/usr/local/homeassistant/` and after that `${INSTALL_DIR}/env/bin/pip install --force-reinstall [module name]` (fill in the module name). Check whether it tries to install additional dependencies. If so, these modules should be added to the `requirements.txt` separately, since spksrc does not install dependencies automatically.
- if it installs fine, it might be that another package was not installed successfully, causing the `pip install` command to abort during installation. Run `${INSTALL_DIR}/env/bin/pip install --use-wheel --no-deps --no-index --force-reinstall -f ${INSTALL_DIR}/share/wheelhouse -r ${INSTALL_DIR}/share/wheelhouse/requirements.txt` to check whether this is the case.

Please note that I do not have time to quickly respond to support requests / issues - I made this in my rare free time. It works for me and I tested it where I could. Everybody is welcome to try it, but please do not expect quick fixes when something does not work. Instead try to contribute. Another source for support is https://github.com/balloob/home-assistant - the upstream project.


refactoredjello:
Thanks for all your hard work @wind-rider. I'll do my best to contribute.

wind-rider:
Great :)

I see that @Dr-Bean noticed the PR so now after some time it will be reviewed and merged if there are no problems in it.

wind-rider:
@Diaoul / @Dr-Bean could you please have a look whether this is good to merge? :)

wind-rider:
Please don't merge this PR yet, currently it might not work well (096c813053a40d59c7abc063119bf38abb102fa6 is still ok but since cd9a2396921ea441720de231d96afebb4f629cf8 and some other changes it doesn't load well. I haven't found the cause yet). Coming two weeks I can't work on it, I'll have a look at it afterwards.

wind-rider:
@Dr-Bean @Diaoul I fixed the issue - since the updated release a dependency needed a newer cross-compiled wheel (psutil). Since it was not found, pip aborted installation of other wheels but DSM does not report this as an error. It only turned up because loading of the modules failed.

The needed wheel is fixed now, so this PR is ready to be merged :)

mk/spksrc.download.mk
```diff
@@ -63,8 +63,12 @@ download_target: $(PRE_DOWNLOAD_TARGET)
 	      if [ ! -f $${localFile} ]; then \
 	        rm -fr $${localFolder}.part ; \
 	        echo "git clone $${url}" ; \
-	        git clone --no-checkout --quiet $${url} $${localFolder}.part ; \
+	        git clone --no-checkout --recursive --quiet $${url} $${localFolder}.part ; \
 	        git --git-dir=$${localFolder}.part/.git --work-tree=$${localFolder}.part checkout --quiet $(PKG_GIT_HASH) ; \
+	        echo "git submodule update --init --recursive" ; \
```
Dr-Bean: Doesn't the `clone --recursive` on L66 imply that the submodules are up-to-date? In other words: what's the purpose of L68-L71?

spk/homeassistant/Makefile
```diff
@@ -11,16 +11,16 @@ BUILD_DEPENDS += cross/pyephem cross/netifaces cross/psutil
 
 SPK_DEPENDS = "python3>=3.4.1-4:git"
 
-DEPENDS = cross/busybox cross/nmap #for python-libnmap
+DEPENDS = cross/$(SPK_NAME) cross/busybox cross/nmap #for python-libnmap
 
 MAINTAINER = SynoCommunity
-DESCRIPTION = Home Assistant is an open-source home automation platform running on Python 3. Track and control all devices at home and automate control.
+DESCRIPTION = "Home Assistant is an open-source home automation platform running on Python 3. Track and control all devices at home and automate control. The default password is empty, just click \"Login\".  Edit your settings via ssh in the configuration.yaml file. This is a beta package. Hue and Chrome are tested. Other modules might work, please test them and try to contribute if something does not work."
```
Dr-Bean: Please keep the description limited to the package, so something like this: `DESCRIPTION = "Home Assistant is an open-source home automation platform running on Python 3. Track and control all devices at home and automate control. The default password is empty, just click \"Login\".`

The rest of the text does not belong here. If official documentation is available explaining what you need, use that (SUPPORTURL). Alternatively, use a help page (HELPURL) to point to a help page or wiki page explaining these details. Realistically, the only necessary sentence is this one: `Edit your settings via ssh in the configuration.yaml file`, the rest is implied as it's a beta package.

spk/homeassistant/Makefile
```diff
@@ -11,16 +11,16 @@ BUILD_DEPENDS += cross/pyephem cross/netifaces cross/psutil
...
 ADMIN_PORT = 8123
 DISPLAY_NAME = Home Assistant
 DSM_UI_DIR = app
 
 BETA = 1
-CHANGELOG = "First release"
+CHANGELOG = "Update to revision bfa8131f4bff33160e08df3a2e0ead76d53f2728. PyISY will now work without having the climate module installed. More release notes at https://home-assistant.io/blog/2015/07/11/ip-cameras-arduino-kodi-efergy-support/."
```
Dr-Bean: You can use the shorthand notation for the revision, no need to repeat the whole sequence. `bfa8131` would do. 
With separate entries, use `1.`,`2.` to separate entries. Use `<br>` for the line ending. See other packages for examples.

spk/homeassistant/src/dsm-control.sh
```diff
@@ -14,7 +14,7 @@ PYTHONPATH="${INSTALL_DIR}/share/${PACKAGE}" #needed for python to find the home
 
 LOG_FILE="${INSTALL_DIR}/var/home-assistant.log"
 PID_FILE="${INSTALL_DIR}/var/homeassistant.pid"
-RUN_ARGS="-m ${PACKAGE} -c ${INSTALL_DIR}/var --open-ui"
+RUN_ARGS="-m ${PACKAGE} -c ${INSTALL_DIR}/var" #--open-ui is not needed since it opens a browser we don't have on a NAS
```
Dr-Bean: Remove the commented-out section there. No need to maintain a running commentary of changes in the code ;)

Dr-Bean:
Added some comments. Did you test that both new installations of the package as well as updates from the current released version work?

spk/homeassistant/Makefile
wind-rider: I plan to make an installation page at home-assistant.io, but it's not finished yet. Since this package is still beta, I thought it wouldn't be a large problem.

mk/spksrc.download.mk
wind-rider: L68-71 were there already and just remove the git metadata from the folder before packaging it.

spk/homeassistant/Makefile
wind-rider: I see. I can do it, but is it a blocker for this revision?

spk/homeassistant/src/dsm-control.sh
wind-rider: I'm not commenting the changes, the --open-ui parameter is recommended in the home-assistant.io documentation.  I like to put a comment when something is not self-explanatory or when something isn't the same as default.

wind-rider:
@Dr-Bean Thanks for your review. I agree with the documentation / help parts, but I think they are not a blocker for this revision since this package is still a work-in-progress. What do you think?

As for testing updates from the current released version - no, I didn't test it. How can I test it? Installing the version from spksrc and then manually installing my own version over it?

A new manual installation I did test; it works well.

mk/spksrc.download.mk
Dr-Bean: I meant [_your_ L68-L71](https://github.com/wind-rider/spksrc/commit/75c9aec62d066f1185669ffed1962bafad8a739c#diff-61dfff1d55cba7939914513abb791f4fR68-R71), not the lines of existing spksrc code :P You added this:
```
+echo "git submodule update --init --recursive" ; \
+    cd $${localFolder}.part ; \
+    git submodule update --init --recursive --quiet ; \
+    cd .. ; \
```
I'm not sure that does anything worthwhile, because the recursive checkout to cover submodules is done just before that (https://github.com/wind-rider/spksrc/commit/75c9aec62d066f1185669ffed1962bafad8a739c#diff-61dfff1d55cba7939914513abb791f4fR66). There's no need to update the submodules right afterwards. Why did you add it?

Dr-Bean:
Maybe a bit of a misconception here. Beta != WIP, as far as SynoCommunity/spksrc goes. 

The main difference between packages with a `beta` tag and packages without one is the amount of feedback that the package does what it's supposed to do. Packages should work [to the best of our knowledge], regardless of stable/beta indicators. 
From a code perspective, there is no difference between the two whatsoever. We don't merge work-in-progress code or publish work-in-progress packages (not knowingly anyway)

So, yes, I'd prefer to see my comments addressed before I merge this :)
As for testing upgrades, yes, you can do that by manually updating an existing package, at least that should give a fair idea that the package still works/config is retained/no errors are logged etc.

Dr-Bean:
Closing. Outdated due to changes in wheel approach, and can't be merged due to conflicts

PR #1776, opened 2015-06-29
Added SABnzbd Beta 0.8.0 Alpha 3

KaraokeStu:
This is a new SABnzbd Beta package, allowing users to run the latest build of SABnzbd.


Dr-Bean:
7z is available by default (at least on 5.0+), so I don't think there's a need to include it as a cross package.

KaraokeStu:
Removed dependency, as requested

KaraokeStu:
I have checked and, without the dependency, sab does not see 7z as available, nor can I unzip a 7z via ssh without first installing p7zip from optware.

Are you sure it is enabled by default?

Dr-Bean:
I would assume that Sab simply calls `7z`, so as long as it's on $PATH, Sab should be able to find it.

Does `which 7z` return `/usr/syno/bin/7z` on your device? Does the binary exist on your device in that location?
What does `echo $PATH` return? It should be `/sbin:/bin:/usr/sbin:/usr/bin:/usr/syno/sbin:/usr/syno/bin:/usr/local/sbin:/usr/local/bin`.

KaraokeStu:
I had a further dig into this, and according to the documentation on the "Introducing 0.8" page of SABnzbd, the issue is as follows:

> Other systems will need to have the "7za" tool available in the $PATH environment variable.

     update-alternatives --install /usr/syno/bin/7za 7za /usr/syno/bin/7z 50
 
Appears to fix the problem.

The question is, should I raise this as a bug with SABnzbd, or should I just include the command within the package?

Dr-Bean:
Hm. It's understandable seeing as there are binaries provided for certain platforms, but imo, allowing the `7z` command as a fallback if `7za` is not available would make a lot of sense.
I wouldn't say it's a bug, but you can certainly raise this as a feature request with SABnzbd. Maybe even consider users to set the path to an executable.

If that doesn't go anywhere, we're better off including the cross package then playing around in Synology's directories ;)

KaraokeStu:
@shypike has confirmed the issue and will be including it in the next alpha :smile:  :+1: 

So that means this pull request is now ready to go in :smiley:

KaraokeStu:
Updated to Alpha 3

KaraokeStu:
@Dr-Bean this is ready to go now, with the fixed 7z support too

spk/sabnzbd-beta/Makefile
```diff
@@ -0,0 +1,58 @@
+SPK_NAME = sabnzbd-beta
```
Dr-Bean: Keeping in line with other packages, use `sabnzbd-testing`. This should be changed for all occurrences of `sabnzbd-beta`...probably the only exception is the changelog if at all.

spk/sabnzbd-beta/Makefile
```diff
@@ -0,0 +1,58 @@
...
+SPK_VERS = 0.8.0
+SPK_REV = Alpha3
```
Dr-Bean: This breaks with package updates. Instead, use:
```
SPK_VERS = 0.8.0Alpha3
SPK_REV = 1
```

spk/sabnzbd-beta/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+DEPENDS  = cross/busybox cross/par2cmdline cross/unrar cross/$(SPK_NAME)
+WHEELS = six configobj
+SPK_DEPENDS = "python>=2.7.6-8"
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = SABnzbd makes Usenet as simple and streamlined as possible by automating everything we can. All you have to do is add an .nzb. SABnzbd takes over from there, where it will be automatically downloaded, verified, repaired, extracted and filed away with zero human interaction.
+DESCRIPTION_FRE = SABnzbd rend Usenet aussi simple et automatisé que possible. Vous n\\\'avez qu\\\'a ajouter un .nzb, SABnzbd s\\\'occupera du reste sans aucune intervention de votre part : téléchargement, vérification, réparation, extraction et déplacement.
+DESCRIPTION_SPN = SABnzbd hace que Usenet sea lo más simple posible, automatizando todo lo que se puede. Todo lo que tienes que hacer es agregar un archivo .nzb. SABnzbd empieza desde ahí. Tus archivos serán automáticamente descargados, verificados, reparados, descomprimidos y archivados.
+ADMIN_PORT = 8080
```
Dr-Bean: This should be changed to a different port, otherwise the consequence is that the sabnzbd package and the sabnzbd-testing package can't be installed simultaneously. See https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports for instructions on how to choose a port.

spk/sabnzbd-beta/src/config.ini
```diff
@@ -0,0 +1,10 @@
+[misc]
+permissions = 777
+auto_browser = 0
+host = 0.0.0.0
+port = 8080
```
Dr-Bean: See my comment on `ADMIN_PORT`

spk/sabnzbd-beta/src/dsm-control.sh
```diff
@@ -0,0 +1,91 @@
...
+PYTHON="${INSTALL_DIR}/env/bin/python"
+SABNZBD="${INSTALL_DIR}/share/SABnzbd/SABnzbd.py"
+CFG_FILE="${INSTALL_DIR}/var/config.ini"
+LOG_FILE="${INSTALL_DIR}/var/logs/sabnzbd.log"
+PID_FILE="${INSTALL_DIR}/var/sabnzbd.pid"
+
+
+start_daemon ()
+{
+    su - ${USER} -c "PATH=${PATH} ${PYTHON} ${SABNZBD} -f ${CFG_FILE} --pidfile ${PID_FILE} -d"
```
Dr-Bean: Please remove the `-` after `su`, we're moving away from using that.

spk/sabnzbd-beta/src/sabnzbd-beta.sc
```diff
@@ -0,0 +1,6 @@
+[sabnzbd-beta]
+title="SABnzbd Beta"
+desc="SABnzbd Beta"
+port_forward="yes"
+dst.ports="8080/tcp"
```
Dr-Bean: See my comment on `ADMIN_PORT`

Dr-Bean:
Added some comments. Two more: 
- Please add a digests file for the cross-package (Use `make digests` in that dir);
- It'd be nice to have a slightly different logo. That makes it easier to distinguish the -testing package from the main one. Maybe change the color (e.g. similar to nzbget), or maybe you have a better idea.

ulrick65:
This is a very interesting package to me...as I am sure it is to others.  I would love to do some testing on it once it is complete.

KaraokeStu:
@Dr-Bean Made the changes, as requested :+1:

ulrick65:
How do give this package a whirl?  Or is this a branch of the original package (in which case I am not sure how to pull a different branch into synology...I have only used the package center for synoCommunity packages).

Thanks.

spk/sabnzbd-testing/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+DEPENDS  = cross/busybox cross/par2cmdline cross/unrar cross/$(SPK_NAME)
+WHEELS = six configobj
+SPK_DEPENDS = "python>=2.7.6-8"
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = SABnzbd makes Usenet as simple and streamlined as possible by automating everything we can. All you have to do is add an .nzb. SABnzbd takes over from there, where it will be automatically downloaded, verified, repaired, extracted and filed away with zero human interaction.
+DESCRIPTION_FRE = SABnzbd rend Usenet aussi simple et automatisé que possible. Vous n\\\'avez qu\\\'a ajouter un .nzb, SABnzbd s\\\'occupera du reste sans aucune intervention de votre part : téléchargement, vérification, réparation, extraction et déplacement.
+DESCRIPTION_SPN = SABnzbd hace que Usenet sea lo más simple posible, automatizando todo lo que se puede. Todo lo que tienes que hacer es agregar un archivo .nzb. SABnzbd empieza desde ahí. Tus archivos serán automáticamente descargados, verificados, reparados, descomprimidos y archivados.
+ADMIN_PORT = 9090
```
Dr-Bean: I'd prefer it if we kept it in the 8000 range, as stated in the wiki article.

Dr-Bean:
Other than the one comment on the port, can you confirm you've checked that the package compiles for all arches and functions correctly for the arches that you can test?

KaraokeStu:
ok, @Dr-Bean hopefully we're all done now

The package compiles for all arches I can currently test on and I've been using it for a few weeks now.

KaraokeStu:
@ulrick65 - What system do you plan to run this on? I'm wondering if you let me know, I can compile it and send you the SPK file to test?

ulrick65:
@KaraokeStu - I have a Synology 1512+, running DSM 5.2-5592 Update 2

THanks.

Dr-Bean:
Thanks ;) Building now, so it'll probably be published later today.

ulrick65:
Any idea when an update will be posted so that I can update sabnzbd from my Synology Control Panel?  I don't know how to update it any other way (I fluent in telnet and the command line and a good understanding of linux, just have never upgraded a package like this before without doing it from the programs interface.).

Dr-Bean:
The Sabnzbd and Sabnzbd-testing packages are completely separate packages, you can't upgrade from one to the other. 
If you want to run the -testing version of Sabnzbd, you can install it just like any other package.

Stanzilla:
beta1 is out now btw https://github.com/sabnzbd/sabnzbd/tree/0.8.0Beta1

PR #1780, opened 2015-07-03
Update Debian-Chroot to Jessie

rmsr:
- Update version to 8.0-1
- Update base distro to Jessie
- Add commented-out testing entries in apt.sources, and comment out unstable
  and experimental. Apt-pinning is never 100% and should not be on by default.
- Clean preferences of references to obsolete third-party repos

### DIFF
--- a/spk/debian-chroot/Makefile
+++ b/spk/debian-chroot/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = debian-chroot
 SPK_VERS = 8.1
-SPK_REV = 1
+SPK_REV = 5
 SPK_ICON = src/debian.png
 DSM_UI_DIR = app
 
--- a/spk/debian-chroot/Makefile
+++ b/spk/debian-chroot/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = debian-chroot
-SPK_VERS = 7.4
-SPK_REV = 5
+SPK_VERS = 8.1
+SPK_REV = 1
 SPK_ICON = src/debian.png
 DSM_UI_DIR = app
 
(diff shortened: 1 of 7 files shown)


spk/debian-chroot/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = debian-chroot
-SPK_VERS = 7.4
-SPK_REV = 5
+SPK_VERS = 8.1
+SPK_REV = 1
```
Dr-Bean: `SPK_REV` only goes up.

spk/debian-chroot/src/sources.list
```diff
@@ -1,31 +1,33 @@
...
+# (unless you change the ordering). See /etc/apt/preferences for more
+# information on how pinning is set up.
 
 ################################################
 ## jessie
-deb http://ftp.fr.debian.org/debian/ jessie main contrib non-free
-deb-src http://ftp.fr.debian.org/debian/ jessie main contrib non-free
-
-## jessie security
+deb http://ftp.debian.org/debian/ jessie main contrib non-free
```
Dr-Bean: It's bad netiquette to use the central servers. Use mirrors where possible, which is why `fr` was used.
Advanced users should probably change the defaults at some point anyway (unless they live in France...)

Dr-Bean:
Added a couple comments. I don't have a particular opinion on the apt-pinning changes, I've never run into issues with it myself. I guess I'm fine with it either way. @Diaoul?

Anyway, thanks for the contribution, much appreciated :)

Diaoul:
LGTM

spk/debian-chroot/src/sources.list
```diff
@@ -1,31 +1,33 @@
...
 ## jessie
 deb http://ftp.fr.debian.org/debian/ jessie main contrib non-free
 deb-src http://ftp.fr.debian.org/debian/ jessie main contrib non-free
-
-## jessie security
 deb http://security.debian.org/ jessie/updates main contrib non-free
 deb-src http://security.debian.org/ jessie/updates main contrib non-free
 
+# jessie-updates - optional, but default
+deb http://ftp.fr.debian.org/ jessie-updates main contrib non-free
```
Dr-Bean: This line and L14 are missing the `debian/` part if I'm not mistaken: `deb http://ftp.fr.debian.org/debian/ jessie-updates main contrib non-free`

Dr-Bean:
While testing this, it seems `aptitude` is no longer available by default, meaning the help file is incorrect. Can't find a source on that, but it's simply not installed afaict. 
We should be able to replace the current commands in the help file with `apt-get <command>`.
@Diaoul On that note, mind if we put the current help file in the Github wiki, and point to that via `HELPURL`? Probably better than having a non-working link in the package.

spk/debian-chroot/src/sources.list
rmsr: Multiple lines missing that, fix pushed.

rmsr:
aptitude is used to maintain and upgrade the debian chroot, so that's another thing that needs fixing.

Dr-Bean:
Thanks :) Made a minor correction in the helpfile with https://github.com/SynoCommunity/spksrc/commit/3282e03de643b10e701c2996136b26d8c3e17722. 
Also, switched over to a wiki page: https://github.com/SynoCommunity/spksrc/wiki/Debian-Chroot. Can be moved over to https://synocommunity.com whenever, but for now, this produces a working link in the package.

PR #1806, opened 2015-08-06
Update owncloud to 8.1.0

mirweb:


Dr-Bean:
Release notes: https://doc.owncloud.org/server/8.1/admin_manual/release_notes.html#changes-in-8-1. Seems there are a number of potentially breaking changes in 8.1.0, looking at those notes. One thing that stands out:
>It is not recommended to upgrade encryption-enabled systems from ownCloud Server 8.0 to version 8.1.0 as there is a chance the migration will break. We recommend to migrate to the first bugfix release, ownCloud Server 8.1.1.

8.1.1 hasn't been released yet, but it might not be a bad idea to wait for that. What do you think?

mirweb:
No Problem. For me I run the package and test it on my nas. The encryption-enabled is no issue for me, because I don't use it.

When 8.1.1 and I have time I will send new pull request

Issue #1510, opened 2015-02-24
[request] Cura SPK


Decipher:
Cura/CuraEngine is a slicing tool used by OctoPrint (#1480); This addition would mean simple 3d printing without the need of a desktop/laptop.

Relevant projects:
- https://github.com/daid/Cura (Python GUI)
- https://github.com/Ultimaker/CuraEngine (C++ engine)

Decipher:
Not 100% sure about the patch, as I had to re-roll some C++11 for loops for C++0x, but it does compile and run on my initial test. Would be grateful for someone to eyeball the patch.

Decipher:
Delved into this a bit more, tried compiling Cura but it has some wheels that wouldn't compile. Ideally I'd like to bypass Cura completely as OctoPrint is moving that way, but you need to use the devel branch of OctoPrint to do that and I was unable to compile the devel branch of OctoPrint as it also had wheels that wouldn't compile :|

Diaoul:
Which wheels?

Decipher:
It seems to be related to Atlast/Scipy/Numpy

* Atlas (http://math-atlas.sourceforge.net/) libraries not found.
* Blas (http://www.netlib.org/blas/) libraries not found.
* Lapack (http://www.netlib.org/lapack/) libraries not found.

And many more errors.

Note that I do have what I believe all the relevant requirements are in requirements.txt, and Atlas is definitely there and run before these errors show up, so I suspect this is a cross-compile related issue where it wants something to exist natively. Would love to proceed further with this one as OctoPrint w/Cura support could be quite nice on Syno.

Dr-Bean:
curaengine 15.04-1 has been published on the repo.

PR #1837, opened 2015-09-08
New version of OctoPrint SPK and add CuraEngine SPK

markbastiaans:
I needed a new version of OctoPrint (#1480), and thought it would be neat to add a CuraEngine SPK, since OctoPrint stable now features a great CuraEngine plugin. With this PR the OctoPrint package now also supports slicing on the NAS, plugins and other goodness.

Here's the progress I've made so far:
- [x] Added CuraEngine 15.04 SPK (originally by @Decipher in #1510, thanks!)
- [x] Fix OctoPrint to cope with Synology symlink madness (foosel/OctoPrint#1051)
- [x] Bump up OctoPrint SPK to latest release (1.2.8), including requirements, configs and Makefile
- [x] Build for all archs (`all-supported`)
- [x] Test on `evansport`

I'd really appreciate any help to get this PR merged!


markbastiaans:
I seem to be stuck at getting the OctoPrint package to run. It seems that some wheels are not being included and / or cross-compiled (in `wheelhouse`, they're included as `x86_64`). I've done some experimenting including them as `BUILD_DEPENDS` along with Python, wheel etc., but this doesn't seem to work. I'm currently using the Docker container as a build environment
Wheels with native C code are - as far as I can see - `markupsafe`, `netifaces`, `psutils`, `pyyaml`, `tornado`.

@Dr-Bean, can you give me a hand getting things to work? Much appreciated!

Dr-Bean:
From the code you're showing I can't see what you have done to add those particular modules, and 'it doesn't seem to work' isn't very detailed ;) Providing relevant output from the build process would help.

In general, if you create and add the necessary cross packages to BUILD_DEPENDS in `spk/octoprint/Makefile`, after cross-compiling [Python and the base modules for building wheels, I don't see a reason that wouldn't work (apart from a possible need of patching the modules, but that's part of the job)

You can consider dropping the Markupsafe and PyYAML requirements, because both are included in the Python package (just make sure the installer uses `--system-site-packages`), and Tornado's C-extensions are optional (as in: if you just include tornado in `src/requirements`, it would still work). That would leave `netifaces` initially, and since that cross package already exists, it should be fairly easy to integrate and test. See e.g. the HomeAssistant package, it uses that module as well iirc.

markbastiaans:
Thanks @Dr-Bean. I've pushed my changes, borrowed from the HomeAssistant package. Building seems to work fine, but I assume I should also change the versions of wheels that need to be cross-compiled in `cross/<wheel>` to match the requirements. Am I right about this?

I'll work on it some more and test if it actually runs at the end of this weekend.

Dr-Bean:
If there's a version mismatch between the cross packages and the requirements file, _something_ needs to be changed, yes ;) 
But, you don't want to break other packages because its associated cross package had a version change. That needs investigating and if need be, a check (upstream issue or even a PR) to make sure a different version than recommended will not cause problems.

markbastiaans:
I've fixed the issues I've having with cross-compilation. Thanks @Dr-Bean for pointing me in the right direction!

The issue was the `tornado` wheel. It has a pure-Python version but the build process included the platform-specific version for the platform I was building on. This made installation fail on the NAS. I've added `tornado` and `psutil-3.2.1` (to prevent version conflicts with other packages) to `cross` and made further modifications to makefile and requirements. Builds for all supported archs and runs great on `evansport`.

Let me know if these changes will make this PR merge-worthy. If not, I'd be happy to continue working on it to make it so.

cross/curaengine/Makefile
```diff
@@ -0,0 +1,24 @@
...
+PKG_EXT = tar.gz
+PKG_DIST_SITE = https://github.com/Ultimaker/CuraEngine/archive
+PKG_DIST_NAME = $(PKG_VERS).$(PKG_EXT)
+PKG_DIR = CuraEngine-$(PKG_VERS)
+
+DEPENDS = 
+
+HOMEPAGE = https://ultimaker.com
+COMMENT  = CuraEngine is a powerful, fast and robust engine for processing 3D models into 3D printing instruction for Ultimaker and other GCode based 3D printers.
+LICENSE  = GNU GPL
```
Dr-Bean: Please verify that that license is correct.

cross/curaengine/digests
```diff
@@ -0,0 +1,3 @@
+15.04.tar.gz SHA1 1a1d2c8f1b3311fbeb5b87fe32b8051508263abe
```
Dr-Bean: I believe setting `PKG_DIST_FILE` in `cross/curaengine/Makefile`, allows for the downloaded file to look like `$(PKG_NAME)-$(PKG_VERS).$(PKG_EXT). Makes it easier to locate in the distrib dir.

spk/curaengine/Makefile
```diff
@@ -0,0 +1,24 @@
...
+DEPENDS = cross/$(SPK_NAME)
+REQUIRED_DSM = 5.0
+
+MAINTAINER = SynoCommunity
+DESCRIPTION  = CuraEngine is a powerful, fast and robust engine for processing 3D models into 3D printing instruction for Ultimaker and other GCode based 3D printers.
+STARTABLE = no
+DISPLAY_NAME = CuraEngine
+
+HOMEPAGE = https://ultimaker.com
+LICENSE  = GNU GPL
```
Dr-Bean: Please verify this license is correct.

spk/octoprint/Makefile
```diff
@@ -1,22 +1,25 @@
-SPK_NAME = octoprint
-SPK_VERS = $(shell date +%Y%m%d)
-SPK_REV = 1
-SPK_ICON = src/octoprint.png
+SPK_NAME   = octoprint
+SPK_VERS   = $(shell date +%Y%m%d)-devel
```
Dr-Bean: `-devel` doesn't belong here, please remove.

If you want to release a testing/develop/nightly version of Octoprint, then you may release a completely separate package as such (see nzbget-testing, sabnzbd-testing etc). In the main Octoprint package, we should use stable releases (or at least as stable as possible)

spk/octoprint/Makefile
```diff
@@ -44,9 +44,9 @@ octoprint_extra_install: $(STAGING_DIR)/share/OctoPrint
...
 	for size in 16 24 32 48 72; do \
 		convert $(SPK_ICON) -thumbnail $${size}x$${size} \
-		        $(STAGING_DIR)/app/images/$(SPK_NAME)-$${size}.png ; \
+			$(STAGING_DIR)/app/images/$(SPK_NAME)-$${size}.png ; \
 	done
 
 $(STAGING_DIR)/share/OctoPrint:
 	install -m 755 -d $(STAGING_DIR)/share
-	cd $(STAGING_DIR)/share && git clone https://github.com/foosel/OctoPrint.git OctoPrint
+	cd $(STAGING_DIR)/share && git clone -b devel https://github.com/foosel/OctoPrint.git OctoPrint
```
Dr-Bean: See my earlier comment. I'd rather see the package using the master branch, and it would be even better to stick with actual releases (https://github.com/foosel/OctoPrint/releases).
At the very least, you need to use a specific commit to checkout.

spk/octoprint/src/config.yaml
```diff
@@ -3,3 +3,13 @@ serial:
 
 server:
   port: 8088
+
+plugins:
+  cura:
+    cura_engine: /usr/local/curaengine/bin/CuraEngine
```
Dr-Bean: I assume this doesn't cause issues if CuraEngine is not installed? E.g. error messages in logs or things like that?

spk/octoprint/Makefile
markbastiaans: It might be a good idea to wait until the upcoming 1.3.0 release. The latest stable release (1.2.6) doesn't work on DSM due to symlink issues. My PR to OctoPrint with a fix has been merged into the `devel` branch.

spk/octoprint/Makefile
markbastiaans: See my reply to your earlier comment.

spk/octoprint/src/config.yaml
markbastiaans: Not unless somebody tries to slice an STL file.

There's also a big "test" button for the CuraEngine binary path in the upcoming 1.3.0 release, so less chance for user error there.

spk/octoprint/Makefile
Dr-Bean: Depends on how long it takes until 1.3.0 is released, I guess. Using official releases is preferred because they are stable, but I don't mind that you use the develop branch if there's a good reason for it. But it isn't a reason to tack `-devel` onto the version :)

If you want (or need) to use the develop branch, you need to specify exactly which commit to use. That way, regardless of when the package is built or who does it, the resulting package is exactly the same as when you tested it.

markbastiaans:
@Dr-Bean I've just updated the PR to use the OctoPrint `master` branch. The 1.2.8 release includes my fix  and I've removed all references to the `devel` branch. Let me know if I can do more to get this PR merged. Much appreciated!

cross/curaengine/Makefile
```diff
@@ -0,0 +1,25 @@
...
+DEPENDS = 
+
+HOMEPAGE = https://ultimaker.com
+COMMENT  = CuraEngine is a powerful, fast and robust engine for processing 3D models into 3D printing instruction for Ultimaker and other GCode based 3D printers.
+LICENSE  = AGPLv3
+
+CONFIGURE_TARGET = nope
+INSTALL_TARGET = myInstall
+
+PATCHES_LEVEL = 1
```
Dr-Bean: Not needed.

cross/psutil-3.2.1/Makefile
```diff
@@ -0,0 +1,14 @@
+PKG_NAME = psutil
+PKG_VERS = 3.2.1
```
Dr-Bean: Fyi, we'll probably include psutil in the next version of the Python package(s), which means some shuffling around of cross packages (`cross/psutil` will be latest available, which would be 3.3.0, if other versions are required, those can be added as separate cross packages with version appended)

spk/octoprint/Makefile
```diff
@@ -1,22 +1,25 @@
-SPK_NAME = octoprint
-SPK_VERS = $(shell date +%Y%m%d)
-SPK_REV = 1
-SPK_ICON = src/octoprint.png
+SPK_NAME   = octoprint
+SPK_VERS   = $(shell date +%Y%m%d)
+SPK_REV    = 1
```
Dr-Bean: `2`, assuming we want to release a new version after merging this PR ;)

Dr-Bean:
Other than the two comments, one last thing: the CuraEngine SPK does not have `SPK_DEPENDS = octoprint=$version"` defined. Was that on purpose?

cross/curaengine/Makefile
markbastiaans: Is this correct? I recon you do need a Makefile for CuraEngine since it really has to be cross-compiled, or am I mistaken?

Dr-Bean: I meant specifically the line `PATCHES_LEVEL`...not the whole file...

markbastiaans:
Yes, this is intentional since the dependency is optional. OctoPrint doesn't depend on CuraEngine, and CuraEngine certainly doesn't depend on OctoPrint.

To clarify: in Octoprint, you can either choose between slicing 3D models somewhere else (in which case you don't need CuraEngine on the NAS) and slicing them on the NAS (in which case you do).

cross/psutil-3.2.1/Makefile
markbastiaans: Cool, I'll check if OctoPrint works with 3.3.0 when the new Python packages are released.

Diaoul:
Too my NAS is far from my 3D printer, I would've used your SPK :-1: 
Got a Pi to do the job with OctoPrint.

How does one use CuraEngine without OctoPrint, no GUI right?

markbastiaans:
@Diaoul Thanks! CuraEngine is usable from the command line, there is an example on the [GitHub Repo](https://github.com/Ultimaker/CuraEngine).

@Dr-Bean I've just pushed the changes addressing your two comments.

Dr-Bean:
Then I'm done with commenting ;)

markbastiaans:
Awesome! Thanks for all your help!

Dr-Bean:
@markbastiaans The new Python package, including psutil 3.3.0, has been released. I don't think it breaks backward compatibility with 3.2.1, but maybe you want to check that.

markbastiaans:
@Dr-Bean Thanks! I've pushed a change to this PR which uses the latest Python package, including `psutil` 3.3.0. I've tested it and so far as I can see, nothing is broken.

I've also changed the OctoPrint updater plugin to track releases instead of commits, which means that users will always have the latest stable version.

Dr-Bean:
Don't you run into issues while `psutil==3.2.1` is hardcoded in Octoprint's setup.py (https://github.com/foosel/OctoPrint/blob/master/setup.py#L35)?

markbastiaans:
Nope, I had the same doubts but the package is installing and running fine on my DS 214 Play.

Dr-Bean:
Huh. That's surprising ;) You sure that the virtualenv doesn't still contain an older version of psutil?

markbastiaans:
I've double-checked and there's no psutil in `/usr/local/octoprint/env/lib/python2.7/site-packages`. Unless I'm missing something this should be okay, right?

Dr-Bean:
I'll take a look at it later, once I get it to compile. Probably a good way to check is to use `pip freeze`/`pip freeze -l`.

You might want to take a look at the curaengine patch, because the current one doesn't work for me. Check the basic instructions I put here: https://github.com/SynoCommunity/spksrc/pull/1852#issuecomment-142917573 for example.

markbastiaans:
Okay, I've narrowed down the patch file failing to `src/util/gettime.h` in CuraEngine. This file has CRLF line endings while the rest of the CuraEngine source (and the patch for it) is LF. I'll re-roll a patch specifically for that file to fix this.

markbastiaans:
Re-rolling the patch didn't work as I expected. I've tested patching the offended file with a patch file with CRLF line endings but it still fails on the line endings.

I have however found a workaround, which I just pushed. It converts the offending file to LF, and patching now works correctly again. @Dr-Bean let me know if the current commit works for you.

Dr-Bean:
Additional change to Octoprint with 36322a7ffcd7d10e7d9d04fa916336ad8ef91d86. Instead of getting ten identical clones of 33 MB a piece, it's only 2,7 MB per.
We should consider moving it to a cross package, that way we can download it once and cache it. Maybe even turn it into a wheel (although then the setup.py would need a patch for psutil ;))

Thanks!

Dr-Bean:
octoprint 20151221-2 and curaengine 15.04-1 have been published on the repository.

markbastiaans:
Nice! I have some other optimizations in mind and will do a new PR for the next OctoPrint / CuraEngine release.

Dr-Bean:
Sure, go ahead :) When you get to it, please split up the changes into separate PR's, one branch for each SPK.

arijaycomet:
OctoPrint & CuraEngine working great on my DS412+ ... thanks for the efforts, guys!  I'll admit I am very much a n00b and such, so pardon my ignorance here.  Is your Synology package version been able to integrate with the Email Notifier?  When trying to install, I get:

"Could not import OctoPrint's setuptools, are you sure you are running that under the same python installation that OctoPrint is installed under?"

Now on my work 3DP, where I setup a RPi with OctoPi, plug-ins work like a champ.  But where the Cura plug-in came pre-installed in your image, and works great, I can't seem to get any other packages to install.  Email is the one I want most.  Meh?  TIA!

markbastiaans:
@arijaycomet looks like a virtualenv issue. This should not occur if you are using the web UI Plugin Manager to install the plugin.

Are you installing from the command line, by any chance? If you are, try the Plugin Manager instead and if that doesn't work, please open a separate issue.

PR #1859, opened 2015-09-30
Fixed gitolite installation

mreppen:
This fixed the gitolite installation which previously failed because of paths and permissions. Should not affect anything else.

The `chown` is moved since `gitolite setup` is run as `git-server`.


Dr-Bean:
@mreppen The last commit looks strange to me. What is the added value of separating gitolite/dropbear into a separate package? Also, doesn't the git-server package need additional changes then?

mreppen:
That was purely for my own sake. This pull request is just to get a working installation for the package as it had been previously developed (gitolite was broken). Apparently I'm not good with github since I thought the pull request would be locked at only 3c9e691.

The reason I pulled out gitolite for my own sake are that
- I didn't want the git daemon running and I couldn't get the UI development package working to create settings for what should and shouldn't run.
- I also didn't want the gitweb (which for me is currently broken, probably because of permission issues).

Therefore I thought I'd make a package for myself with what I need for now. I also added port forwarding, fixed the gitolite installation to a stable commit (v3.6.3), and cleaned up. To the extent I can, I'd be happy to support a gitolite package if you are interested.

TL;DR: I am bad at git(hub). Package was broken. I made a new package with the things I need (gitolite).

Dr-Bean:
You might want to redo/clean up the PR then. Assuming that we want to keep gitolite and dropbear in the same package, then it's best to start in a clean up-to-date branch. I have some comments on the latest commit, but I'm not sure if that was for your own setup, or the SynoCommunity package?

KaraokeStu:

Thanks,

Stuart

PR #1866, opened 2015-10-08
Add cmake and lzip to Dockerfile

CondormanFr:
Add missing packages used to compil some projects (bitlbee for example)


CondormanFr:
Can be test here : https://hub.docker.com/r/rgouyet/spksrc/tags/

PR #1874, opened 2015-10-13
Update nzbget to 16.0.

cxhamilton:


cytec:
Seems like you just copy&pased stuff from the testing package but forgot the new config template ;)
\cc @Dr-Bean

cxhamilton:
Indeed. I have updated the config templates.

Dr-Bean:
I had already prepared the same changes while updating the -testing package. I didn't push it yet, because I wanted to test what happens with active downloads and content in the queue. I seem to remember the queue should be emptied between 15.x and 16.x but I can't recall where that came from or find a reference for it right now.
So, did or can you test that?

cxhamilton:
I have made changes, based on @Dr-Bean 's comments. I also tested upgrading from 15.x to 16.x during an active download. The queue was preserved and the download continued to completion without issue.

Dr-Bean:
Cool, thanks :)

Issue #1872, opened 2015-10-13
Update Node.js to 4.2 LTS


leak:
Could anyone please update the Node.js package to version 4.2 which is marked for Long Term Support and also incorporates the features from the reintegrated io.js fork.
##

Dr-Bean:
According to https://github.com/nodejs/node#unix--macintosh, GCC4.8 is needed. The current toolchains go up to GCC4.7, which is why we can't update the package.
We requested Synology to update GCC (via e.g. https://github.com/SynoCommunity/spksrc/issues/1457#issuecomment-81726526), so maybe that's included in the next toolchain release. 

You might want to consider asking Synology to update their Node package via https://www.synology.com/en-global/form/inquiry/feedback, then we can retire our package ;)

hmalphettes:
Thanks for the feedback link: filed the request.
Here is my vote for others to raise Synology's attention about node support: if you want it go and fill it.

For now in my case I run node-0.12 with the `--harmony` option.
It will give you block scoping (const/let), generators, Map and Set (incomplete).
Avoid buggy arrow functions. Multi-line template strings are unsupported.

More complete list: http://stackoverflow.com/questions/28388885/ecmascript-6-features-available-in-node-js-0-12

leak:
Thanks for the pointers. I also started an inquiry about node and the toolchain.

iharris:
Yes, would be very good to get the latest version running vs 6 months old.

W7PEA:
The new DSM 6 Beta is out, has anyone been able to check we can upgrade NodeJS after 6 is out?

piwi82:
New toolchains for DSM6 beta 2 are out (2 hours ago) : https://sourceforge.net/projects/dsgpl/files/DSM%206.0%20Beta2%20Tool%20Chains/
I don't know how to check the version of GCC.

piwi82:
Synology provides packages for NodeJS [v0.10](https://www.synology.com/fr-fr/dsm/app_packages/Node_js_v0_10), [v0.12](https://www.synology.com/fr-fr/dsm/app_packages/Node_js) and [v4](https://www.synology.com/fr-fr/dsm/app_packages/Node_js_v4).

Dr-Bean:
Seeing as Synology appears to be maintaining their Node packages again, I think we should consider retiring our Node package. 
We don't use Node as a dependency for other packages, and providing an identical base package like this seems unnecessary.

@SynoCommunity/developers?

[...]

PR #1879, opened 2015-10-14
Upgrade node to v4.2.1 (LTS) - no architectures supported?

hmalphettes:
As requested here: https://github.com/SynoCommunity/spksrc/issues/1872

### DIFF
--- a/cross/node/Makefile
+++ b/cross/node/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = node
-PKG_VERS = 0.12.2
+PKG_VERS = 4.2.1
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-v$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = http://nodejs.org/dist/v$(PKG_VERS)/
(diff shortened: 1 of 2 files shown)


hmalphettes:
I would need some help with compilation:
```
vagrant@localhost:~/spksrc/cross/node$ make arch-88f6281
===>  Building package for arch 88f6281
...
error: invalid floating point option: -mfpu=vfpv2
```
I am a newbie with arm but documentation matches 88f6281 with fully ARMv5TE-compliant 
I cannot determine if that implies that the floating point instructions set vfpv2 are available or not.

hmalphettes:
I found out that V8 has dropped arm support for chips that don't have vfpv2: https://groups.google.com/forum/#!topic/v8-users/aSOFbaAQvMk

So this new version of node wont work with 88f6281.

The other new constraint is the requirement for gcc4.8 as V8 requires C++11.
It looks like none of the architectures have a toolchain that meets this requirement: at best we get gcc4.7

Someone knows what to do?

Dr-Bean:
Nothing we can do about it, unless Synology releases new toolchains with an updated GCC version. Maybe the next toolchain release will cover it, we'll have to wait and see.

As for 88f628x: Since Node v0.12, ARMv5 wasn't supported anymore, which is why this line exists: https://github.com/SynoCommunity/spksrc/blob/master/cross/node/Makefile#L9.

hmalphettes:
@Dr-Bean thanks for the explanation. Closing then.

PR #1662, opened 2015-04-24
ffmpeg update to 3.0

cytec:
update ffmpeg for an emby compatible build: https://github.com/SynoCommunity/spksrc/issues/1346

Includes:

    libx264
    libmp3lame
    libfdk-aac
    libspeex
    openssl
    libvorbis
    libass
    libbluray
    libopenjpeg
    libopus
    libtheora
    librtmp
    fontconfig
    libfreetype
    libvpx

`libvpx` currently does not compile when invoked from `spk/ffmpeg` so one has to compile it in `cross/libvpx` and copy the files over atm... seems like there are some variables set which are not overwritten correctly when invoked as `DEPEND`

@Diaoul @Dr-Bean can you have a look on that or provide any informations on how to ... "reset" everything for this depend? seems like there are still some vars set (`CC`, `LD` etc) which collide with the way `libvpx` has to be cross compiled... hopefully there is a way to just "reset" that stuff or overwrite it or something... 

Currently i've got it working and running an my DS1513+ (cedarview) but transcoding is slow and uses a lot of CPU currently, most likely because there aren't any opimizations done yet which should definitively be done in the future... Sadly i don't have any access to my VM from now on for at least 1 Week, so if anyone is willing to try compiling this for some other arches (takes about 1hour depending on your PC) that would be great :)

Like i mentioned before `cross/libvpx` has to be compiled on his own and copied/moved to the `spk/ffmpeg` work folder afterwards. (personally i would recommend doing this first before anything than it's much easier)... 

EDIT: x64 SPK can be found here: http://cytec.us/syno/test-spk/ffmpeg/

PR #1900, opened 2015-10-27
Comskip Initial release

schumi2004:
Thanks for all the help, it all seems to compile just fine now.
I think it needs to be stripped more because size is around 100MB now.

### DIFF
--- a/mk/spksrc.cross-cc.mk
+++ b/mk/spksrc.cross-cc.mk
@@ -110,7 +110,7 @@ $(DIGESTS_FILE): download
 
 dependency-tree:
 	@echo `perl -e 'print "\\\t" x $(MAKELEVEL),"\n"'`+ $(NAME) $(PKG_VERS)
-	@for depend in $(DEPENDS) ; \
+	@for depend in $(BUILD_DEPENDS) $(DEPENDS) ; \
 	do \
 	  $(MAKE) --no-print-directory -C ../../$$depend dependency-tree ; \
 	done
@@ -127,7 +127,7 @@ kernel-required:
 	@if [ -n "$(REQ_KERNEL)" ]; then \
 	  exit 1 ; \
 	fi
-	@for depend in $(DEPENDS) ; do \
+	@for depend in $(BUILD_DEPENDS) $(DEPENDS) ; do \
 	  if $(MAKE) --no-print-directory -C ../../$$depend kernel-required >/dev/null 2>&1 ; then \
 	    exit 0 ; \
 	  else \
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -303,7 +303,7 @@ all: package
 
 dependency-tree:
 	@echo `perl -e 'print "\\\t" x $(MAKELEVEL),"\n"'`+ $(NAME)
-	@for depend in $(DEPENDS) ; \
+	@for depend in $(BUILD_DEPENDS) $(DEPENDS) ; \
 	do \
 	  $(MAKE) --no-print-directory -C ../../$$depend dependency-tree ; \
 	done
@@ -434,7 +434,7 @@ kernel-required:
 	@if [ -n "$(REQ_KERNEL)" ]; then \
 	  exit 1 ; \
 	fi
-	@for depend in $(DEPENDS) ; do \
+	@for depend in $(BUILD_DEPENDS) $(DEPENDS) ; do \
 	  if $(MAKE) --no-print-directory -C ../../$$depend kernel-required >/dev/null 2>&1 ; then \
 	    exit 0 ; \
 	  else \


cytec:
just left you some commits, maybe @Dr-Bean has anything else but besides that. Nice!

schumi2004:
Okay, lots of things to work on. Will check them out later today. Thanks

spk/comskip/src/dsm-control.sh
```diff
@@ -0,0 +1,28 @@
...
+        exit 0
+        ;;
+    stop)
+        exit 0
+        ;;
+    status)
+        exit 0
+        ;;
+    log)
+        echo "${INSTALL_DIR}/install.log"
```
cytec: does that file even exist?

cytec:
except of that one last commit it looks good to me... let's see if @Dr-Bean has anything left :)

spk/comskip/Makefile
```diff
@@ -0,0 +1,38 @@
+SPK_NAME = comskip
+SPK_VERS = 0.81.089
+SPK_REV = 1
+SPK_ICON = src/comskip.gif
```
Dr-Bean: Could you make this a PNG file please? And the file itself doesn't seem to be 256*256, but that would be the recommended size. I can't see the icon right now, but iirc, it was fairly low quality. Maybe take a look at that too.

spk/comskip/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+DEPENDS = cross/ffmpeg cross/argtable cross/$(SPK_NAME)
```
Dr-Bean: I'd do this alphabetical, with `cross/$(SPK_NAME)` always at the end.

spk/comskip/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+RELOAD_UI = yes
```
Dr-Bean: If the package isn't startable, and doesn't have an icon to show in DSM's menu, this isn't needed. 
@cytec's example (and probably a whole heap of others) should be cleaned up at some point :)

spk/comskip/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+POST_STRIP_TARGET = comskip_extra_install
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: comskip_extra_install
+comskip_extra_install:
+	install -m 755 -d $(STAGING_DIR)/var
+	install -m 644 src/comskip.ini $(STAGING_DIR)/var/comskip.ini
+	install -m 755 -d $(STAGING_DIR)/app
```
Dr-Bean: L32-L38 can be removed: I don't think those icons are ever used for pure command-line packages. The packaging process takes care of adding two icons which are used for Package Center.

spk/comskip/src/comskip.ini
```diff
@@ -0,0 +1,78 @@
+; See comskip.txt in the distribution zip file for many settable parameters, read manual.html, tuning.html and debugwindow.html for how to tune and debug comskip
+detect_method=43			;1=black frame, 2=logo, 4=scene change, 8=fuzzy logic, 16=closed captions, 32=aspect ration, 64=silence, 128=cutscenes, 255=all
+validate_silence=1			; Default, set to 0 to force using this clues if selected above.
+validate_uniform=1			; Default, set to 0 to force using this clues (like pure white frames) if blackframe is selected above.
+validate_scenechange=1		; Default, set to 0 to force using this clues if selected above.
```
Dr-Bean: Would it be possible to make the indents the same throughout?

spk/comskip/src/dsm-control.sh
Dr-Bean: Nope ;) If Comskip doesn't maintain it's own log to link to from here, this can be removed.

Dr-Bean:
@Dr-Bean sure has ;) A couple other points: please check that digests are added (`make digests`).
Besides that, how is comskip used? Is the idea to start it from the command line by itself and use a video file as argument, or does it integrate in other packages somehow? That might be usefule to describe somewhere. Maybe a short line in the description (Com

schumi2004:
Yeah i was thinking about the same regarding a short howto description. You could schedule this and launch from cli but the way i'm using it is with TVH, when a recording is finished i trigger comskip to process recording and generate a EDL file which can be used in Kodi. More info about this here http://kodi.wiki/view/Edit_decision_list

Will fix other comments also and i think I need to create a new icon since i couldn't find a proper one also.

spk/comskip/src/dsm-control.sh
schumi2004: I thought this was needed for the package itself (install.log) Comskip creates a logfile next to the processed file thus this can be removed indeed.

spk/comskip/Makefile
schumi2004: How would you know package is installed then? Or am i missing something?

Dr-Bean:
Depending on how long a howto could be (describing various methods to integrate it?), you could consider writing a wiki page, and add a link like this: https://github.com/SynoCommunity/spksrc/blob/master/spk/ffsync/Makefile#L23. It'll show up as an additional link in Package Center for the package.

Dr-Bean:
Another one: I believe you're missing `BETA =1` in the SPK, or did I miss that?

spk/comskip/Makefile
Dr-Bean: Package Center will refresh itself, no problem. The RELOAD_UI is to refresh the DSM interface altogether. That includes animating the placement of the icon in the menu and the `1` showing up on the menu button to indicate a new package is installed.
Seeing as the package doesn't need that, this is unnecessary here.

schumi2004:
Beta isn't there, will add.

spk/comskip/src/dsm-control.sh
Dr-Bean: You may have used Mono as an example? It wrote that information during postinstall. That should be removed with the next release, I actually intended to do that.

spk/comskip/src/comskip.ini
```diff
@@ -0,0 +1,78 @@
...
+remove_after=0				; amount of seconds of show to be removed after ALL commercials
+shrink_logo=5				; Reduce the duration of the logo with this amount of seconds
+after_logo=0		; set to number of seconds after logo disappears comskip should start to search for silence to insert an additional cutpoint
+padding=0
+ms_audio_delay=5
+volume_slip=40
+skip_b_frames=0				; Set to 1 to force Comskip to skip frames for higher processing speed.
+hardware_decode=0			; Set to 1 to enable hardware accelerated video decoding, only available in donator version
+max_repair_size=200			; Will repair maximum 200 missing MPEG frames in the timeline, set to 0 to disable repairing for players that don't use PTS. 
+disable_heuristics=4		bit pattern for disabling heuristics, adding 1 disables heristics 1, adding 2 disables heristics 2, adding 4 disables heristics 3, 255  disables all heuristics
```
Dr-Bean: Is this perhaps missing a `;` in front of `bit pattern [..]`?

Dr-Bean:
You can use SUPPORTURL (official documentation) in your Makefile, along with HELPURL (wiki).

spk/comskip/src/comskip.ini
schumi2004: Yes it is. Fixed it.

schumi2004:
Updated and squashed. Image is still bad but better then nothing. Wiki i will update later today.

spk/comskip/Makefile
```diff
@@ -0,0 +1,34 @@
+SPK_NAME = comskip
+SPK_VERS = 0.81.089
+SPK_REV = 1
+SPK_ICON = src/comskip.png
+BETA = 1
+
+DEPENDS = cross/argtable cross/ffmpeg cross/$(SPK_NAME)
+
+RELOAD_UI = yes
```
cytec: seems like you missed to remove this ;)

schumi2004: Wasn't sure if it really should be removed or not.

Will remove it.

cross/argtable/Makefile
```diff
@@ -0,0 +1,16 @@
+PKG_NAME = argtable
+PKG_VERS = 2.13
+PKG_EXT = tar.gz
+PKG_DIR = $(PKG_NAME)$.2-13
```
Dr-Bean: Weird...I thought I added a comment here, but maybe I forgot to add it. This line should be changed. I'd suggest to replace your L1-L6 with the following:
```
PKG_NAME = argtable
PKG_VERS_MAJOR = 2
PKG_VERS_MINOR = 13
PKG_EXT = tar.gz
PKG_DIST_NAME = $(PKG_NAME)$(PKG_VERS_MAJOR)-$(PKG_VERS_MINOR).$(PKG_EXT)
PKG_DIST_SITE = http://sourceforge.net/projects/$(PKG_NAME)/files/$(PKG_NAME)/$(PKG_NAME)-$(PKG_VERS_MAJOR).$(PKG_VERS_MINOR)
PKG_DIR = $(PKG_NAME)$(PKG_VERS_MAJOR)-$(PKG_VERS_MINOR)
```

spk/comskip/src/comskip.ini
```diff
@@ -0,0 +1,78 @@
...
+validate_silence=1			; Default, set to 0 to force using this clues if selected above.
+validate_uniform=1			; Default, set to 0 to force using this clues (like pure white frames) if blackframe is selected above.
+validate_scenechange=1			; Default, set to 0 to force using this clues if selected above.
+verbose=10				; Show a lot of extra info, level 5 is also OK, set to 0 to disable
+max_brightness=60			; Frame not black if any pixels checked are greater than this (scale 0 to 255)
+test_brightness=40      		; Frame not pure black if any pixels checked are greater than this, will check average brightness (scale 0 to 255)
+max_avg_brightness=25			; Maximum average brightness for a dim frame to be considered black (scale 0 to 255) 0 means autosetting
+max_commercialbreak=600 		; Maximum length in seconds to consider a segment a commercial break
+min_commercialbreak=25			; Mimimum length in seconds to consider a segment a commercial break
+max_commercial_size=125			; Maximum time in seconds for a single commercial or multiple commercials if no breaks in between
```
Dr-Bean: Some indents could still use some work. L12, L33, L34, L41 seem to have one indent too many. Maybe caused by your editor vs Github.

cytec:
how about taking this as image? comscip_03

spk/comskip/src/comskip.ini
schumi2004: When viewing in full it seems to look perfect?
Did change the idents but on github it still looks wierd (using gedit)

spk/comskip/Makefile
schumi2004: Removing this brakes compiling the package, DEPENDS should stay i think?

schumi2004:
I commented on a outdated comment so not sure it it will popup.
But removing this

Dr-Bean:
You only want to remove `RELOAD_UI = yes`, not the lines above it. (see https://github.com/schumi2004/spksrc/commit/23dc962ad40f86d9f22e237f456129473223149a, that commit removed the first SPK-related lines)

Dr-Bean:
@cytec Is that image 256*256? Seems smaller, but that might be my browser. It looks better though, no doubt about that :)

spk/comskip/src/comskip.ini
Dr-Bean: I have no idea what's up with that then. Could be something weird going on with Github's rendering/markdown. We can check when merging, see if it shows up ok locally (I would assume so if you're editing with gedit)

spk/comskip/src/installer.sh
```diff
@@ -0,0 +1,52 @@
...
+{
+    exit 0
+}
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # Put nano in the PATH
```
Dr-Bean: s/nano/comskip

schumi2004:
Thank you guys for all your help so far. Learned a lot.
I have implemented all the suggestions/comments, if there still is anything left I will look at it tonight.

Dr-Bean:
Last comment from me today (probably): I'd prefer the image to be 256*256 to start off, maybe this one can be resized without too much effort.

Other than that, nicely done, and good work keeping up with two people commenting on your code at the same time ;)

cytec:
well if you like i can provide that image in 256*256... i redraw it anyway :D

schumi2004:
If you can give me a 256x256 image I think @Dr-Bean says PR would be ready to be pulled ? ;)

cytec:
@schumi2004 here we go 
comskip

schumi2004:
@cytec thanks for icon ;)
I'm done, it's all yours.

Thanks all ;)

Dr-Bean:
Should be good to go when the wiki article is done :)

This does end up as an 70-something-MB package at the moment, which is ridiculous imo. Most of that seems to be related to the current ffmpeg package though, so @cytec, any near-term plans for #1662?

cytec:
i hope i can get something done towards #1662 this weekend ;)
how about useing ffmpeg only for build and then require the ffmpeg SPK on install time? or get rid of everything not explicit needed by comskip... 

im not really a fan of building a dependency several times as there will be a dedicated ffmpeg SPK which already should contain everything needed by comskip (libraries etc)... so my try/solution would be the same i used for `chromaprint` when i was testing it: installed ffmpeg spk and update `LD_LIBRARY_PATH` to include ffmpeg's lib folder... 

or maybe try making a static build? i don't know if comskip needs access to the libs all the time or if you can get rid of it after building (like chromaprint spk does currently)

schumi2004:
@cytec You mean as in ffmpeg package must be installed first before comskip/chromaprint would install?
Not sure if this will work for comskip. What i have read is that it needs to be compiled against ffmpeg. If we then don't include ffmpeg but link to existing package i'm not sure it still works.
I could give it a try.

Wiki page needs to wait a day, need to testing something regarding configuration file location and if comskip detects it.

schumi2004:
I was trying to reduce the size of the package and remembered my previous question to def on the Kodi forum.
[quote='tmm1' pid='2109398' dateline='1442596915']
[quote='schumi2004' pid='2109383' dateline='1442595850']
Can this also be (cross) compiled for a Synology DS715 (alpine) ?
[/quote]
Yes, but you will have to cross-compile ffmpeg first to link against.
[/quote]
If I understand it correctly I can drop the ffmpeg stuff in package.
Will test a package with and without ffmpeg stuff in it.

Dr-Bean:
Note sure, but the dev could have meant that ffmpeg is a compile-time dependency for comskip, and not a run-time dependency. In spksrc, the functionality is somewhat available as `BUILD_DEPENDS` vs `DEPENDS`.
Using BUILD_DEPENDS means that the PLIST of the dep is not processed, so in this case, the resulting SPK will not have the ffmpeg bins and libs included.

Using this mechanic, you can play around a bit with the setup. For example, if you set `BUILD_DEPENDS = ffmpeg`, and comskip works: perfect. Package is smaller, and there are no downsides.
On the other hand, if it turns out that ffmpeg _is_ a runtime dep, you can still use `BUILD_DEPENDS`, but then you can set `LD_LIBRARY_PATH = /usr/local/ffmpeg/lib` before starting comskip. (Which is what @cytec was suggesting)
Upside: comskip's package is smaller, downside: Not the most user-friendly for cli-based applications where you actually need to type it.

Maybe you can play around with it for a bit, see if it works.

Dr-Bean:
Fyi, back to `status/needs-review` due to an additional commit.

schumi2004:
I needed to move comskip.ini into bin folder otherwise comskip couldn't find it without adding argument to ini file itself which isn't user friendly either in my opinion.
Also noticed after looking at ffmpeg that the PLIST was different then described in wiki. Not sure if it's needed for ffmpeg but just noticed it. (bin: lnk: man: etc)
l.s.: Am i correct ffmpeg doesn't work for evansport architecture?

Regarding including ffmpeg or not i'll leave it this way, it works now and I can always optimize it later.

Dr-Bean:
Hm. I'm not sure I'm willing to publish 80MB*8 arches or so, if we can get away with about half that with a little work. Secondly, the ini file doesn't belong in /bin, and that is going to cause issues with package upgrades. You need to add arguments to comskip anyway, might as well add an ini location. If people don't, I assume comskip is going to complain, or use a set of (incorrect) defaults, yes?

As for ffmpeg: the package needs work, which is what #1662 is for.

schumi2004:
Comskip is indeed going to complain if it can't find the ini file.
I checked code if I could patch it to it would search in correct directory but haven't find the correct place yet.
When ini is placed next to binary you don't need to submit arguments because comskip finds the ini file itself so you can just use comskip somevideofile.mkv otherwise it would be like this comskip /volume1/@appstore/comskip/var/comskip.ini somevideofile.mkv

Maybe a symlink works (if allowed in packages)?

cytec:
i've checked and it seems like comskip searches your `PATH` for `comskip.ini` by default. Not 100% sure but i guess if you want to patch it, you'll have to do it somewhere around here: https://github.com/erikkaashoek/Comskip/blob/8a4d502b0349d9e39179def86a2f681317f76cbc/comskip.c#L12239-L12247 to also check if the file is present in the packages `var` folder...

cytec:
I've just compiled `comskip` and removed `ffmpeg` libs and as i expected it failed running as it needs `libavutil` on runtime... maybe you can use an approach i used a while ago for a different package:

Compile it with `cross/ffmpeg` and then in the Makefile of the spk remove everything not needed... not really the cleanest solution but at least you'll not end up with an 80MB Package... maybe this works :) for testing purpose you could just try to remove some libs from the lib folder and check if it still runs ...

schumi2004:
Maybe not cleanest but works for me, good idea though.
At first i'm going to fix myself a decent demo file to test.
After that i'm going to manually remove stuff and see how it goes.

Also reverted commit for putting ini file in bin folder, it's back in var again (with extra comskip.dictionary file)
Users should always give full path to ini file now until i found a solution.
I noticed those lines in comskip.c also but thought it was for windows builds. I asked a someone at Kodi forum if he could help me with this PATH thing.

cytec:
you could also just append ' /usr/local/comspkip/var` to the PATH ... maybe `putenv` would to the trick?

Diaoul:
The cleanest solution for this is to patch comskip to accept an extra directory to search for the .ini defined in an env variable.
When you compile comskip SPK you set that compile-time env variable to the `/usr/local/comspkip/var` directory. You will end up with a solution that doesn't require any user action and is still flexible enough so you can change that directory without changing the patch file (not hardcoded).

cytec:
+1 for this solution... maybe you'll also append `/usr/local/ffmpeg/lib/` to the `LD_LIBRARY_PATH` on runtime compilation so that you are able to move `ffmpeg` to `BUILD_DEPENDS` and add ffmpeg SPK to the SPK depends...

seems like it needs more than just `libavutil` :/

Diaoul:
+1 for `--rpath` or reusing something existing rather than patching to add a library path.

schumi2004:
I need to figure out the append stuff and BUILD and SPK depends and also --rpath (all new for me)
Hopefully I can fix and test it today.

cytec:
i'm not 100% sure but if i remember correctly setting `--rpath` can be done by useing `ADDITIONAL_LDFLAGS` like this: `ADDITIONAL_LDFLAGS=-Wl,-rpath,/usr/local/ffmpeg/lib/`

@Dr-Bean @Diaoul correct me if im wrong ;)

for the env patch for comskip: maybe @Diaoul can provide some more help/information on this...

tmm1:
The PATH searching code is only enabled on windows. I don't mind adding it for other platforms but we will need a posix implementation of _searchenv in platform.c

schumi2004:
@tmm1  adding PATH for other environments would make things alot easier but don't know how easy/hard it is to implement.

schumi2004:
@cytec 
Can you provide me source code to compile ffmpeg for Evansport architecture? I spoke to someone that said he got a Evansport package from you?

cytec:
@schumi2004 should be the source from here https://github.com/cytec/spksrc/tree/ffmpeg-update

schumi2004:
@cytec 
I compiled ffmpeg after i cloned your fork but it returns an error, any ideas?

1/install//usr/local/ffmpeg/lib -Wl,--rpath,/usr/local/ffmpeg/lib " ./bootstrap.sh
./bootstrap.sh: 2: ./bootstrap.sh: autoreconf: not found
Makefile:21: recipe for target 'myPreconf' failed
make[3]: *** [myPreconf] Error 127
make[3]: Leaving directory '/home/openelec/github/cytec-spksrc/cross/openjpeg'
../../mk/spksrc.depend.mk:44: recipe for target 'depend_target' failed
make[2]: *** [depend_target] Error 2
make[2]: Leaving directory '/home/openelec/github/cytec-spksrc/cross/ffmpeg'
../../mk/spksrc.depend.mk:44: recipe for target 'depend_target' failed
make[1]: *** [depend_target] Error 2
make[1]: Leaving directory '/home/openelec/github/cytec-spksrc/spk/ffmpeg'
../../mk/spksrc.spk.mk:337: recipe for target 'arch-evansport-5.1' failed
make: [arch-evansport-5.1] Error 2 (ignored)

schumi2004:
@cytec could you have a look at rtmpdump?
I'm unable to compile it, is it still needed?

sr/local/ffmpeg/lib -Wl,--rpath-link,/home/openelec/github/cytec-spksrc/spk/ffmpeg/work-evansport-5.2/install//usr/local/ffmpeg/lib -Wl,--rpath,/usr/local/ffmpeg/lib "
/bin/sh: 1: cd: can't cd to /home/openelec/github/cytec-spksrc/spk/ffmpeg/work-evansport-5.2/rtmpdump-ksv
Makefile:20: recipe for target 'myCompile' failed
make[3]: *** [myCompile] Error 2
make[3]: Leaving directory '/home/openelec/github/cytec-spksrc/cross/rtmpdump'
../../mk/spksrc.depend.mk:44: recipe for target 'depend_target' failed
make[2]: *** [depend_target] Error 2
make[2]: Leaving directory '/home/openelec/github/cytec-spksrc/cross/ffmpeg'
../../mk/spksrc.depend.mk:44: recipe for target 'depend_target' failed
make[1]: *** [depend_target] Error 2
make[1]: Leaving directory '/home/openelec/github/cytec-spksrc/spk/ffmpeg'
../../mk/spksrc.spk.mk:417: recipe for target 'arch-evansport-5.2' failed
make: [arch-evansport-5.2] Error 2 (ignored)

schumi2004:
@cytec 
I think this worked ADDITIONAL_LDFLAGS="-Wl,-rpath,/usr/local/ffmpeg/lib/"
(and removing bin files for ffmpeg)
I have a package that is 2,4MB in size now.

Will test ffmpeg package and see how it works on test NAS

schumi2004:
Comskip depends on ffmpeg-3.x these days. Since ffmpeg-3.x isn't in SynoCommunity's repo yet this PR can't be merged.

cytec:
yeah... i should really get my ffmpeg stuff merged -.-

schumi2004:
I have a ffmpeg-3.0.2 package available here. I could submit it so you can review?

ndaversa:
@schumi2004 with this landing: https://github.com/SynoCommunity/spksrc/pull/2375 can we get comskip landed as well?

schumi2004:
I'll finish it today ;)

schumi2004:
Current ffmpeg isn't working by default.
Need to figure out what is different with mine.

PR #1905, opened 2015-11-02
Update Berkeley DB to 6.1.26.

kratz00:
- fixed downloading of package source file


kratz00:
Here is the official change log in case you are interested http://download.oracle.com/otndocs/products/berkeleydb/html/changelog_6_1.html#idp518176

Diaoul:
Thanks

Dr-Bean:
I thought we forced a manual download because of license issues..?

kratz00:
@Dr-Bean The code is distributed under the GNU AFFERO GENERAL PUBLIC LICENSE          Version 3 license. What are your concerns?

Dr-Bean:
I don't have concerns, I asked a question. IIRC, a downloadable file for berkeley-db was removed for some reason or another, which is why the manual download was forced. Either that was because of licensing issue, or there was an unrelated reason. Check the commit history.

kratz00:
I see. No idea about the reasons, I only found some comments following this commit https://github.com/SynoCommunity/spksrc/commit/89b235855839174913256a0d0fb5c459b625dd0a It looks more like downloading it directly was not supported for technical reasons. Hopefully @Diaoul can remember the real reason.

Dr-Bean:
Well, the license did change to AGPLv3 with 29ed311e17cd958e25aa4bc02728d963b6a12ecc.
Oracle does require you to sign in prior to downloading, but you don't need to accept a license agreement (at least not anymore, although I can't remember if that was ever part of the procedure, e.g. as with Java)

Diaoul:
Yes, that was part of the procedure, one had to manually sign up then login, accept the license prior to be able to download berkeleydb. This wasn't something that could be automated. Now it seems they provide a direct link, that's much easier for us.

Dr-Bean:
That direct link might have always been there ;) You can do the same for Java, because the link you get after accepting the license agreement is directly accessible too. It goes against that license agreement though, which is why e.g. Synology's Java package does what it does.

Anyway, that clears it up. I'll probably remove the trailing slash in PKG_DIST_SITE at some point.

Diaoul:
I believe it wasn't available for direct download if you weren't logged back in 2012 but I might be wrong. OE does direct download too so I don't think there is any real issue here. (See [that recipe](http://cgit.openembedded.org/cgit.cgi/openembedded-core/tree/meta/recipes-support/db/db_6.0.30.bb?h=master))

Dr-Bean:
Yeah, I was thinking that too (the login-before-download stuff), so that must've changed at some point I guess. I'm perfectly fine with automating the download, obviously ;)

I'll also take a look at slightly cleaning up the code in download.mk (set `manual`, instead of playing around with `nop[e]`) I've been wanting to do that for a while to make it clear it has a very specific purpose...even though we might never have to use it again.

Issue #540, opened 2013-05-25
Deluge


foXaCe:
The project spk deluge is it stop? I am looking for an alternative to transmission and the official client synology

schumi2004:
@Diaoul
Do you know if anything has changed that would give compiling this package a second chance or is it still useless like last year?

Diaoul:
I think if we stick to 5.1 toolchains we can add libtorrent to Python SPK and/or use wheels. So I'd say yes, we can give this package a second chance.

Bon3:
@Diaoul How long time u think it will be before a beta of Deluge will appear in SynoCommunity repo?

schumi2004:
Any progress made already, maybe toolchain 5.2 is easier?

schumi2004:
I accidentally run against these packages, someone already tried them or know them?
Would it be possible to decrypt them and create compile code for it?

http://ssd.dlinkddns.com/pub/synology/deluge/

Dr-Bean:
Success ;) Got a Deluge package with a standalone version of libtorrent (as in: it relies on the Python SPK, but doesn't include Python, as the linked packages above seem to do).
DSM5.0+ only because of multiprocessing in libtorrent.

Unfortunately, I can't get PPC-based arches (qoriq and ppc853x) to work at this time...it seems there's something between boost and libtorrent that just won't work properly with PPC. @Diaoul Do you recall if libtorrent/deluge could be compiled for PPC in the past?

Diaoul:
Nope it couldn't, that's why I didn't want to include libtorrent in Python SPK.

Dr-Bean:
Right. Well, you can take a look at #1911. I'll probably merge/build it in a couple of days, if you don't see anything standing out.
Fyi, we need to swap over the alpine toolchain, the current one is broken when compiling boost/libtorrent.

Dr-Bean:
Merged the PR with 0279f415ad63a7d4030a45d11c33b20c44ca0e43
deluge 1.3.12-2 has been published on the repository as a beta package.

Seeing as this is quite an old ticket, I'm closing it. Let us know if the package works as expected (you can do that in this issue), that way the package has a chance to move out of beta at some point.

Please [open a new issue](https://github.com/SynoCommunity/spksrc/issues/new) if you encounter problems.
[...]

PR #1911, opened 2015-11-05
Fix and update Deluge and dependencies

Dr-Bean:
See https://github.com/SynoCommunity/spksrc/issues/540

### DIFF
--- a/cross/boost/Makefile
+++ b/cross/boost/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = boost
-PKG_VERS = 1.53.0
+PKG_VERS = 1.59.0
 PKG_EXT = tar.bz2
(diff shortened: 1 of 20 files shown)


Diaoul:
LGTM

Dr-Bean:
Todo: 
- Default password is set to `deluge`, we should mention that in the wizard;
- Add wizard to set download dir/completed/blackhole dir, sets `torrentfiles_location`, `download_location`, `autoadd_location`, `move_completed_path` in `/usr/local/deluge/var/core.conf`;
- _Maybe_ additional config changes in `/usr/local/deluge/var/core.conf` (`random_outgoing_ports": true` to `false` and other random ports, and see how that affects the firewall.
- Check what config.yml is for, it doesn't seem to be used.

Dr-Bean:
Rebased, should be good to go. Only changes are corrections/additions to installer/dsmcontrol/wizard.

PR #1913, opened 2015-11-05
Added fuse-exfat and exfat-utils packages to support exfat

romantymoshyk:
Initial release. Currently i'm working on automounting of exfat, on fail i will design UI app that will allow to mount/unmount exfat partitions.


spk/fuse-exfat/src/installer.sh
```diff
@@ -0,0 +1,51 @@
...
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+    
+    # Put exfat & fusermount in the PATH
+    mkdir -p /usr/local/bin
+    ln -s ${INSTALL_DIR}/bin/mount.exfat /usr/local/bin/mount.exfat
+	ln -s ${INSTALL_DIR}/bin/fusermount /usr/local/bin/fusermount
```
Dr-Bean: This caught my eye, and I'm not talking about the indenting issue :) The `sshfs` package creates the same symlink, but to a different location. That's no good, because it means unexpected results if both packages are installed/uninstalled. Does this symlink need to exist? Note that the symlinks are only a convenience option, not a necessity.

spk/fuse-exfat/Makefile
```diff
@@ -0,0 +1,24 @@
+SPK_NAME = fuse-exfat
+SPK_VERS = 1.2.1
+SPK_REV = 2  
+SPK_ICON = src/fuse.png
+
+DEPENDS = cross/$(SPK_NAME) cross/exfat-utils
+
+MAINTAINER = Roman Tymoshyk 
+DESCRIPTION = Add support of exFAT filesystem
```
Dr-Bean: This is a description of the package as it appears on [our repository](https://synocommunity.com/packages), or in Package Center. You might want to change it to the description the developer provides.

spk/fuse-exfat/Makefile
```diff
@@ -0,0 +1,24 @@
+SPK_NAME = fuse-exfat
+SPK_VERS = 1.2.1
+SPK_REV = 2  
+SPK_ICON = src/fuse.png
+
+DEPENDS = cross/$(SPK_NAME) cross/exfat-utils
+
+MAINTAINER = Roman Tymoshyk
```
Dr-Bean: Probably you'll want to put SynoCommunity or a username here (based on the assumption that the repo can handle usernames)

sh0dow:
Dear @Dr-Bean, i made fixes that you suggest

spk/fuse-exfat/src/installer.sh
```diff
@@ -0,0 +1,64 @@
...
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+    
+    # Put exfat & fusermount in the PATH
+    mkdir -p /usr/local/bin
+    ln -s ${INSTALL_DIR}/sbin/mount.exfat-fuse /usr/local/bin/mount.exfat
+    ln -s ${INSTALL_DIR}/sbin/mount.exfat-fuse /usr/local/bin/mount.exfat-fuse
+    ln -s ${INSTALL_DIR}/sbin/dumpexfat /usr/local/bin/dumpexfat
+    ln -s ${INSTALL_DIR}/sbin/exfatfsck /usr/local/bin/exfatfsck
+    ln -s ${INSTALL_DIR}/sbin/exfatlabel /usr/local/bin/exfatlabel
+	ln -s ${INSTALL_DIR}/sbin/exfatfsck /usr/local/bin/fsck.exfat
```
cytec: seems like you got some indents mixed there :)

spk/fuse-exfat/src/installer.sh
```diff
@@ -0,0 +1,64 @@
...
+    exit 0
+}
+
+postuninst ()
+{
+    # Remove links
+    rm -f ${INSTALL_DIR}
+    rm -f /usr/local/bin/mount.exfat
+    rm -f /usr/local/bin/mount.exfat-fuse
+	rm -f /usr/local/bin/dumpexfat
```
cytec: see commit above

sh0dow:
indents unified

Dr-Bean:
To be blunt: I don't like the dumping all available binaries/executables on PATH at all. If people need them all on PATH, they can do so themselves, or call the binary with the full path.

spk/fuse-exfat/Makefile
```diff
@@ -0,0 +1,24 @@
+SPK_NAME = fuse-exfat
```
Dr-Bean: In my mind, calling the SPK `exFAT` instead makes sense. Same with the filename, for that matter.

spk/fuse-exfat/Makefile
```diff
@@ -0,0 +1,24 @@
...
+SPK_VERS = 1.2.1
+SPK_REV = 3
```
Dr-Bean: Start with `1`. It only goes up when we plan to release a new version of the SPK.

spk/fuse-exfat/Makefile
```diff
@@ -0,0 +1,24 @@
...
+SPK_ICON = src/fuse.png
```
Dr-Bean: As per my earlier comment, how about `exfat.png`? On that note, would it make sense to change the icon to something more fitting?

spk/fuse-exfat/Makefile
```diff
@@ -0,0 +1,24 @@
...
+
+DEPENDS = cross/$(SPK_NAME) cross/exfat-utils
```
Dr-Bean: Probably would need a change here too, if $(SPK_NAME) can't be used.

spk/fuse-exfat/src/dsm-control.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+# Package
+PACKAGE="exfat"
+DNAME="exfat"
```
Dr-Bean: Please use consistent names. `exFAT` here, assuming that's what the project calls it.

spk/fuse-exfat/src/installer.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+
+# Package
+PACKAGE="exfat"
+DNAME="exfat"
```
Dr-Bean: Please use consistent names. `exFAT`, assuming that's what the project calls it.

spk/fuse-exfat/src/installer.sh
```diff
@@ -0,0 +1,64 @@
...
+}
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+    
+    # Put fuse-exfat & exfat-utils in the PATH
+    mkdir -p /usr/local/bin
+    ln -s ${INSTALL_DIR}/sbin/mount.exfat-fuse /usr/sbin/mount.exfat
```
Dr-Bean: Only use `/usr/local`, no links to other locations. Please also reconsider what goes on PATH.

spk/fuse-exfat/Makefile
sh0dow: fuse-exfat is original name of the package in all linux systems, like sshfs

spk/fuse-exfat/src/installer.sh
sh0dow: Links to /usr/sbin/ required for automounting according to next article 
http://stackoverflow.com/questions/1554178/how-to-register-fuse-filesystem-type-with-mount8-and-fstab

spk/fuse-exfat/Makefile
sh0dow: Fully agree

spk/fuse-exfat/src/installer.sh
Dr-Bean: `/usr/sbin` is rewritten during DSM updates, afaik.
In other words, the symlinks are dropped when someone updates DSM (possibly not with all minor versions, but that would depend on what's in the patch).
That's why we're limited to `usr/local`.

spk/fuse-exfat/Makefile
Diaoul: I think you can chose the one you prefer as long as you stick to it and are consistent.

spk/exfat/src/installer.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+    
+    # Put fuse-exfat & exfat-utils in the PATH
+    mkdir -p /usr/local/bin
+    ln -s ${INSTALL_DIR}/sbin/mount.exfat-fuse /usr/local/bin/mount.exfat
```
Dr-Bean: Seems this should be changed to `ln -s ${INSTALL_DIR}/sbin/mount.exfat /usr/local/bin/mount.exfat`, right?

Dr-Bean:
Apart from the one comment, could you please add digests (use `make digests` while in the cross package dir)?
Barring any issues with compiling, I think it should be good to go with those changes.

Dr-Bean:
Ping @sh0dow, still working on this?

sh0dow:
Yes @Dr-Bean , i can't get auto mounting to work

KaraokeStu:

Thanks,

Stuart

noplanman:
Any chance this gets picked up again by a motivated individual? :innocent:

PR #1914, opened 2015-11-06
Added usleep usage to sqlite

BillyNate:
Without HAVE_USLEEP, xSleep on unix is implemented using sleep() which means that sqlite3_sleep() will have a minimum wait interval of 1000 milliseconds regardless of its argument,
causing the [sqlite nightmare](http://beets.radbox.org/blog/sqlite-nightmare.html).

### DIFF
--- a/cross/sqlite/Makefile
+++ b/cross/sqlite/Makefile
@@ -13,6 +13,6 @@ LICENSE  = http://www.sqlite.org/copyright.html
 
 GNU_CONFIGURE = 1
 
-ADDITIONAL_CFLAGS = -DSQLITE_ENABLE_COLUMN_METADATA
+ADDITIONAL_CFLAGS = -DSQLITE_ENABLE_COLUMN_METADATA -DHAVE_USLEEP=1
 
 include ../../mk/spksrc.cross-cc.mk


Dr-Bean:
Have you checked that all arches are able to compile this (assuming that if usleep is not implemented for a particular arch, the build process will break/halt)?

BillyNate:
No, I did not check that all arches are able to compile this.
Would you mind telling me how I can check this?

Dr-Bean:
Enter the cross dir, issue `make all-archs`. See https://github.com/SynoCommunity/spksrc/wiki/Compile-and-build-rules#cross-package-rules

BillyNate:
Thank you for the explanation.
I've tried following the steps on [github.com/SynoCommunity/spksrc](https://github.com/SynoCommunity/spksrc#usage) on a Raspberry Pi but I'm getting several errors.
Also going into the cross dir and issueing `make all-archs` returned `No rule to make target 'all-archs'. Stop.`.
Unfortunately I don't even like setting up build environments and such (and I find it a shortcoming to see spksrc isn't in some kind of CI to check pull-request automaticly).
Isn't there a more easy way to check if all archs are able to compile this?

Diaoul:
Well, sure no CI is a shortcoming but still, you should test your build, that we cannot automate. You can setup an env quite easily with Docker. Just pull the image and run bash in a container after mounting spksrc volume.

cytec:
+1 for CI... if you want to.. go ahead and open a PR ;)

Dr-Bean:
I don't think there's a publicly available CI service that can handle what we need with the space requirements we have (toolchains, kernels etc) or in a timely fashion. We'd probably end up needing to roll our own, and tbh, for the majority of PR's, I don't see how that helps.

Back to the question which started it off: you won't be able to setup spksrc on an rPi, as that's ARM-based. You need x86. Setting up spksrc via Docker would be the quickest way to set it up.

Dr-Bean:
Closing. It turns out that this is already checked during `configure`, and if it is supported, `-DHAVE_USLEEP=1` is passed along.

Issue #1078, opened 2014-06-12
Arch-* targets not recognized by latest bash-completion


moneytoo:
Hiting `<TAB>` after writing `make arch-` doesn't list any target when using latest `bash-completion` 2.1. For now it's fine in Debian stable as it uses 2.0 but 2.1 is already in testing and other distros already have it (Ubuntu for example).

It doesn't work because of [following rule](http://anonscm.debian.org/gitweb/?p=bash-completion/bash-completion.git;a=blob;f=completions/make;h=f8f71939b67640793142b7aac77846a3e88c6161;hb=afe39fd1e171d0ea3d4fb8ec0d8c8c14fa120ed8#l24) in bash-completion: 
`/^# Not a target/,/^$/        d;            # skip not target blocks` 

Running `make --print-data-base --dry-run` confirms the culprit:

```
...

# Not a target:
arch-88f6281:
#  Implicit rule search has not been done.
#  Modification time never checked.
#  File has not been updated.

...
```

Who's to blame? I miss any `make` know how so I don't know if the arch targets are defined correctly.
##

astroganga:
I have the same problem on a virtual machine with debian 8 jessie. I need to continue use old vm with debian 7 to work around. Any news?

Dr-Bean:
Kind of weird to not upgrade just because autocompletion isn't working for some make targets...Anyway, we can fix it with this: `.PHONY: $(addprefix arch-,$(AVAILABLE_ARCHS))`. We have a couple of other rules which could use similar changes.

astroganga:
Actually I am using both vm. I have some uncommited code on debian 7 and I am preparing environment to transfer on debian 8. 
I tried your suggestion and it works. I think we need to add this too:
`.PHONY: $(addprefix publish-arch-,$(AVAILABLE_ARCHS))`

thanks

Dr-Bean:
Problem is that it doesn't seem to actually work as intended, does it? It autocompletes, but it looks like the target is always considered up-to-date (...or something, haven't figured it out yet), because for subsequent commands, you end up with this:
```
/spksrc/spk/testpackage$ make arch-88f6281-5.2
make: Nothing to be done for 'arch-88f6281-5.2'.
```

Running `make arch-88f6281` works because that isn't considered part of the `.PHONY` target...but autocompletion probably won't ever work for that command.

astroganga:
You're right. I was not aware of this serious side effect. 
[...]

PR #1919, opened 2015-11-08
WIP: Attempt to fix autocomplete

Dr-Bean:
As per https://github.com/SynoCommunity/spksrc/issues/1078, autocompleting arches has been broken since bash-completion 2.1.
This PR changes the build rules such that bash autocompletion works again, without losing the current functionality.

We could move the `PREP` variables to spksrc.common.mk and rename them. We can also add the content to the rules themselves, but I don't think that'll help readability (and it's ugly enough as is...)
Also, there are probably ways to shorten/clarify the substitutions/lookups etc in the `PREP` variables.

### DIFF
--- a/mk/spksrc.cross-cc.mk
+++ b/mk/spksrc.cross-cc.mk
@@ -118,9 +118,11 @@ dependency-tree:
 .PHONY: all-archs
 all-archs: $(addprefix arch-,$(AVAILABLE_ARCHS))
 
-arch-%:
-	@$(MSG) Building package for arch $*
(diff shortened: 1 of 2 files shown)


cytec:
@Dr-Bean any news on this? seems like this occurs on the docker container too...

Dr-Bean:
Having not worked on this in ages, no progress to report, but yeah, it'd happen on any platform.

Not sure what I want to do with this actually, because another option might be to scrap the split between `ARCH` and `TCVERSION`. That helps with `smart-clean`, cos that's broken unless you manually set TCVERSION every time, and there was something else that needed fixing...which I can't remember.
Another reason for not pursuing this at the time is it's so darn ugly. Not sure I can stand it :/

Otoh I have no idea if reworking ARCH and TCVERSION actually cleans things up or not...I'd have to dive into it again, and other things seem to be more urgent atm.

KaraokeStu:

Thanks,

Stuart

ymartin59:
Related to #2371

Safihre:
Closing due to age, can be re-opened if someone wants to work on it again.

PR #1926, opened 2015-11-10
Create Gerrit package

Zetten:
Create a package for the Gerrit code review platform. This package
distributes the gerrit.war file and preseeded config. The bouncycastle
libraries are installed automatically (may not be redistributed
directly), then the Gerrit installation is launched in non-interactive
mode.
----
I'd appreciate some further testing with this - it works pretty well in my case (DS414, armadaxp) but the default config is pretty basic and I think any actual user of this would have to make their own adjustments.


Dr-Bean:
Source on not being allowed to redistribute the Bouncycastle libraries? The [license](http://www.bouncycastle.org/licence.html) doesn't say anything about that.

spk/gerrit/Makefile
```diff
@@ -0,0 +1,54 @@
+SPK_NAME = gerrit
+SPK_VERS = 2.11.4
+SPK_REV = 1
+SPK_ICON = src/gerrit.png
+DSM_UI_DIR = app
+
+DEPENDS = cross/busybox
+SPK_DEPENDS = "JavaManager:git"
```
Dr-Bean: Please add restrictions for models/arches that don't support JavaManager. Use `UNSUPPORTED_ARCHS`, like [this](https://github.com/SynoCommunity/spksrc/blob/master/spk/sonarr/Makefile#L8)

spk/gerrit/Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = Gerrit provides web based code review and repository management for the Git version control system.
+ADMIN_PROTOCOL = https
+ADMIN_PORT = 8091
```
Dr-Bean: Can you move out of the 80* range, as per https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports#how-to-choose-a-port?

spk/gerrit/Makefile
```diff
@@ -0,0 +1,54 @@
...
+LICENSE = Apache 2.0
+HOMEPAGE = "https://gerritcodereview.com"
+
+INSTALLER_SCRIPT = src/installer.sh
+SSS_SCRIPT = src/dsm-control.sh
+FWPORTS = src/${SPK_NAME}.sc
+
+INSTALL_PREFIX = /usr/local/$(SPK_NAME)
+
+#COPY_TARGET = nop
```
Dr-Bean: Remove this please

spk/gerrit/Makefile
```diff
@@ -0,0 +1,54 @@
...
+POST_STRIP_TARGET = gerrit_extra_install
+
+BUSYBOX_CONFIG = usrmng
+ENV += BUSYBOX_CONFIG="$(BUSYBOX_CONFIG)"
+
+GERRIT_WAR_URL="https://gerrit-releases.storage.googleapis.com/gerrit-${SPK_VERS}.war"
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: gerrit_extra_install
```
Dr-Bean: This looks convoluted. Just create the dirs you need directly in one target, and move the files you have into them where needed.

spk/gerrit/Makefile
```diff
@@ -0,0 +1,54 @@
...
+gerrit_extra_install: $(STAGING_DIR)/bin/gerrit.war $(STAGING_DIR)/etc $(STAGING_DIR)/lib
+	install -m 755 -d $(STAGING_DIR)/app
+	install -m 644 src/app/config $(STAGING_DIR)/app/config
+
+$(STAGING_DIR)/bin/gerrit.war:
+	install -m 755 -d $(STAGING_DIR)/bin
+	wget -nv -c $(GERRIT_WAR_URL) -O $(WORK_DIR)/gerrit-$(SPK_VERS).war
```
Dr-Bean: Move the download to a cross package (example: cross/rutorrent) and add it to DEPENDS.

spk/gerrit/src/dsm-control.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/sh
+
+# Package
+PACKAGE="gerrit"
+DNAME="Gerrit"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PATH="${INSTALL_DIR}/bin:${GIT_DIR}/bin:/usr/local/bin:/bin:/usr/bin:/usr/syno/bin"
```
Dr-Bean: Use `PATH="${INSTALL_DIR}/bin:${GIT_DIR}/bin:${PATH}`

spk/gerrit/src/dsm-control.sh
```diff
@@ -0,0 +1,65 @@
...
+USER="gerrit"
+PID_FILE="${INSTALL_DIR}/logs/gerrit.pid"
+LOG_FILE="${INSTALL_DIR}/logs/error_log"
+
+start_daemon ()
+{
+    su - ${USER} -c "PATH=${PATH} ${INSTALL_DIR}/bin/gerrit.sh start"
```
Dr-Bean: Remove the `-` after `su`

spk/gerrit/src/dsm-control.sh
```diff
@@ -0,0 +1,65 @@
...
+}
+
+stop_daemon ()
+{
+    su - ${USER} -c "PATH=${PATH} ${INSTALL_DIR}/bin/gerrit.sh stop"
```
Dr-Bean: See previous comment

spk/gerrit/src/gerrit.config
```diff
@@ -0,0 +1,24 @@
...
+[index]
+    type = LUCENE
+[auth]
+    type = OPENID
+[sendemail]
+    enable = false
+    smtpServer = localhost
+[container]
+    user = gerrit
+    javaHome = /volume1/@appstore/JavaManager/Java/jre
```
Dr-Bean: JavaManager, just as other packages, aren't always installed on `volume1`.

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+# Package
+PACKAGE="gerrit"
+DNAME="Gerrit"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+SSS="/var/packages/${PACKAGE}/scripts/start-stop-status"
+JAVA_DIR="/var/packages/JavaManager/target/Java/jre"
+GIT_DIR="/usr/local/git"
+PATH="${INSTALL_DIR}/bin:${GIT_DIR}/bin:${JAVA_DIR}/bin:/usr/local/bin:/bin:/usr/bin:/usr/syno/bin"
```
Dr-Bean: Same as earlier comment

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+BOUNCYCASTLE_SSL_URL="http://www.bouncycastle.org/download/bcpkix-jdk15on-${BOUNCYCASTLE_VER}.jar"
+
+#SYNO_GROUP
+#SYNO_GROUP_DESC
+
+DEFAULT_PLUGINS="download-commands review"
+
+preinst ()
+{
+	exit 0
```
Dr-Bean: Formatting: Change all tabs to (4) spaces in this file.

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+{
+	# Link
+	ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+	# Install busybox stuff
+	${INSTALL_DIR}/bin/busybox --install ${INSTALL_DIR}/bin
+
+	if [ "${SYNOPKG_PKG_STATUS}" != "UPGRADE" ]; then
+		# Create user
+		adduser -h ${INSTALL_DIR} -g "${DNAME} User" -G ${GROUP} -s /bin/sh -S -D ${USER}
```
Dr-Bean: Move outside of the if-statement please

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+
+		# Install prerequsite libraries
+		wget -nv -c "$BOUNCYCASTLE_PROVIDER_URL" -O "${SYNOPKG_PKGDEST}/lib/bcprov-jdk15on-${BOUNCYCASTLE_VER}.jar"
```
Dr-Bean: No downloads on install time.

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+		wget -nv -c "$BOUNCYCASTLE_SSL_URL" -O "${SYNOPKG_PKGDEST}/lib/bcpkix-jdk15on-${BOUNCYCASTLE_VER}.jar"
+
+		# Correct the file ownership
```
Dr-Bean: You're already doing this outside of the if-statement, which is a good thing. No need to do it twice.

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+		chown -R ${USER}:${GROUP} ${SYNOPKG_PKGDEST}
+
+		# Install Gerrit and initialise secondary index (if upgrading, do this after restoring working data in postupgrade)
+		su - ${USER} -c "PATH=${PATH} java -jar ${SYNOPKG_PKGDEST}/bin/gerrit.war init --batch --no-auto-start --site-path=${SYNOPKG_PKGDEST} --install-plugin=download-commands --install-plugin=reviewnotes --install-plugin=singleusergroup --install-plugin=replication"
```
Dr-Bean: - Same comment as earlier (`su -` -> `su`)
- Replace ${SYNOPKG_PKGDEST} with ${INSTALL_DIR}

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+		su - ${USER} -c "PATH=${PATH} java -jar ${SYNOPKG_PKGDEST}/bin/gerrit.war reindex --site-path=${SYNOPKG_PKGDEST}"
```
Dr-Bean: Same as previous comment

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+
+		# Change pid check
+		sed -i "s#ps -p \$PID#ls /proc/\$PID#" ${SYNOPKG_PKGDEST}/bin/gerrit.sh
+
+		# Add firewall config
```
Dr-Bean: Move outside of if-statement please.

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+
+preupgrade ()
+{
+	# Stop the package
+	${SSS} stop > /dev/null
+
+	# Save all the working data
+	rm -fr ${TMP_DIR}/${PACKAGE}
+	mkdir -p ${TMP_DIR}/${PACKAGE}
+	for D in data db etc git index lib logs plugins; do
```
Dr-Bean: I can see this causing problems in the future. Instead of doing this, can you put all the content you want to keep in `${INSTALL_DIR}/var` (or ${INSTALL_DIR}/etc` for files that can't be changed), and create a backup of that.

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+		mkdir -p ${INSTALL_DIR}/${D}
+		mv ${TMP_DIR}/${PACKAGE}/${D}/* ${INSTALL_DIR}/${D}/
+	done
+	rm -fr ${TMP_DIR}/${PACKAGE}
+
+	# Correct the file ownership
+	chown -R ${USER}:${GROUP} ${SYNOPKG_PKGDEST}
+
+	# Re-init and reindex
+	su - ${USER} -c "PATH=${PATH} java -jar ${SYNOPKG_PKGDEST}/bin/gerrit.war init --batch --no-auto-start --site-path=${SYNOPKG_PKGDEST} --install-plugin=download-commands --install-plugin=reviewnotes --install-plugin=singleusergroup --install-plugin=replication"
```
Dr-Bean: This is mostly already included in postinst(), so just use that. Besides, if you restore the db and index, why reindex?

spk/gerrit/src/installer.sh
```diff
@@ -0,0 +1,133 @@
...
+	su - ${USER} -c "PATH=${PATH} java -jar ${SYNOPKG_PKGDEST}/bin/gerrit.war reindex --site-path=${SYNOPKG_PKGDEST}"
+
+	# Change pid check
+	sed -i "s#ps -p \$PID#ls /proc/\$PID#" ${SYNOPKG_PKGDEST}/bin/gerrit.sh
```
Dr-Bean: Already done as part of postinst(), but you'll need to move it outside of the if-statement.

Dr-Bean:
First walkthrough, added some comments.
Could you please trim and resize the icon to 256*256 please?

Nice start :)

spk/gerrit/Makefile
Zetten: Is there any easy way to figure out exactly what the value of `UNSUPPORTED_ARCHS` should be?

spk/gerrit/src/installer.sh
Zetten: Gerrit's init/reindex process includes handling of schema upgrades, which is pretty important. You're right that it should be handled by the same call in postinst, but that is actually called before postupgrade so the necessary have not been restored yet... Any idea how best to tackle that?

spk/gerrit/src/installer.sh
Zetten: The installation of Gerrit is greatly simplified by dumping everything into what it calls the $SITE_PATH which is then effectively managed by the 'init' call (e.g. an upgrade is handled by calling 'init' on an existing install directory).

I'll see how much of this can be configured, but I've left it as-is for now.

spk/gerrit/Makefile
Zetten: Since Gerrit needs several files downloaded (but not extracted or compiled or anything) I made the cross/gerrit Makefile slightly more complex than the other projects in there. It's not _too_ bad but I'm guessing there's probably a better way to do this - any hints?

Zetten:
Right, I finally got another chance to look at this! Thanks very much for the comments; I've made most of the requested changes and amended accordingly.

Still to do:
* Fix the icon
* Double-check behaviour of postinst/postupgrade (I think it's ok though)

Any further suggestions would be appreciated though :smile:

spk/gerrit/src/dsm-control.sh
```diff
@@ -0,0 +1,65 @@
...
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PATH="${INSTALL_DIR}/bin:${GIT_DIR}/bin:${PATH}"
+USER="gerrit"
+PID_FILE="${INSTALL_DIR}/logs/gerrit.pid"
+LOG_FILE="${INSTALL_DIR}/logs/error_log"
+
+start_daemon ()
+{
+    su ${USER} -c "PATH=${PATH} ${INSTALL_DIR}/bin/gerrit.sh start"
```
cytec: where is that file? did i oversee it?

Zetten: This file comes from the gerrit package itself and is created/overwritten by the init call. It's pretty much just a standard initscript in terms of format, but it sets up a bunch of config variables and stuff before launching the executable .war.

After init we have to sed a line in there during postinst, since it makes use of a `ps` arg format unsupported by DSM. Fortunately the start-stop-status commands all seem to work fine after this though, so just passing through to the Gerrit script saves a whole bunch of work.

Zetten:
FWIW, I did consider that the 'correct' way of creating this package might actually be to run the `java -jar gerrit.war init` during the packaging process, and collect the result.

But that has more than a few downsides, and would be significantly more complex than just delegating to Gerrit's own init process:
* The build infrastructure would need Java available
* The init call generates some crypto keypairs, which would need to be stripped and re-generated/restored during installation
* The init call takes care of database schema upgrades, which would need to be re-run on an existing installation anyway

spk/gerrit/src/dsm-control.sh
cytec: Was just asking because it isn't listed in `cross/gerrit` PLIST and the digests file also only refers to the .war file ...

Zetten: I may have misunderstood the purpose of the various manifests and makefiles... What's the best procedure for a package like this, where the actual SPK contains only a limited subset of what's required to run the application (but is automatically bootstrapped into the full application)?

The Gerrit install/upgrade process is not great from the packaging side, I admit... This 'init after install' approach is roughly what is used by the semi-official deb/rpm packages though: https://gerrit.googlesource.com/gerrit-installer/+/refs%2Fheads%2Fmaster

Safihre:
@Zetten Are you still willing to work on getting this package in SynoCommunity?
If so, let us know, otherwise we will close this PR at some point to keep a better overview of work-in-progress 📝

Zetten:
@Safihre I'm happy to try to keep it moving. I guess I should start looking at the DSM 6 work to see what the general pattern is... is there a reference package for inspiration?

Although I would probably wait until the upcoming Gerrit 2.15 release before working on it properly. There is a bit of rejigging regarding database use and storage, so it may get slightly easier to manage.

Safihre:
The simplest packages are `mosquitto`, but more "advanced" ones are for example NZBGet package where there are also upgrade and special install steps included.

hgy59:
closed due to inactivity

PR #1930, opened 2015-11-13
[Duplicity] Add missing 'six' dependency

adrien-delhorme:
six is a missing dependency of pyrax module


Dr-Bean:
There are a lot more missing if you're looking at dependency resolving. The question is if they are they needed or not.
Run `./env/bin/pip install -U --force-reinstall -f /usr/local/duplicity/share/wheelhouse -r /usr/local/duplicity/share/wheelhouse/requirements.txt`, and compare the two sets of requirements.

adrien-delhorme:
I forgot dependencies of dependencies. They are required.
I did a new pull request that fixes that: #1942.

Issue #1886, opened 2015-10-19
Pillow compil failed


CondormanFr:
I'm finishing a new SPK to allow to run WeeWX on Synology.
The last issue is with image creation.
The python project use PILLOW to generate picture and I get an error because 

```"The _imagingft C module is not installed".```

In fact, Pillow need to have libfreetype. (as explain here : https://github.com/weewx/weewx/wiki/Troubleshooting-Pillow)

I didn't success in installing Pillow 2.9.0 or 3.0.0  with WHEELS because this version seems not compatible with 64 bit linux : I get this error : 
```
 ImportError: /volume1/@appstore/weewx/env/lib/python2.7/site-packages/PIL/_imaging.so: wrong ELF class: ELFCLASS32
```

So at the end, it seems that the only way to have a working version of Pillow must be to cross compil it.
Pillow is already in spksrc but the actual version doesn't compil : 

```
===>  Configuring for Pillow
===>  Compiling for Pillow
env: setup.py: No such file or directory
../../mk/spksrc.python-module.mk:28: recipe for target 'compile_python_module' failed
make[1]: *** [compile_python_module] Error 127
make[1]: Leaving directory '/spksrc/cross/pillow'
```
Is-it possible to solve this issue, then I will test to add libfreetype as depends and try to make running WeeWX...

Thanks

Dr-Bean:
Pillow needs cross-compiling, yes. At the moment, it's compiled as a python module, and created as part of the python SPK (hence the `mk/spksrc.python-module.mk`, you can't compile it standalone).
You can do one of two things:
- Leave the Pillow package as-is, and just add freetype as a dependency. Compile a new version of the Python package, and during your package installation, create a virtualenv with `--system-site-packages`, which will make Pillow available in that virtualenv
- Convert the Pillow package into a wheel. Change the include to `mk/spksrc-wheel.mk`, and include the cross-package as BUILD_DEPENDS of your package. Note that you don't want to break the Python package, so make sure that still works.

Dr-Bean:
Closing, I assume my explanation was clear enough.

Issue #1939, opened 2015-11-16
Python didn't compil : issue with uwsgi


CondormanFr:
Python SPK compil failed : 

I ran the /spksrc/native/python make command and I get no error.

When I launch the make command in the spk\python folder, i get this error : 

```
[thread 1][/spksrc/toolchains/syno-avotonIn file included from /usr/include/python2.7/Python.h:8:0,
                   from plugins/python/uwsgi_python.h:2,
                   from plugins/python/python_plugin.c:1:
  /usr/include/python2.7/pyconfig.h:3:51: fatal error: x86_64-linux-gnu/python2.7/pyconfig.h: No such file or directory
  compilation terminated.
  In file included from /usr/include/python2.7/Python.h:8:0,
                   from plugins/python/uwsgi_python.h:2,
                   from plugins/python/pyloader.c:1:
  /usr/include/python2.7/pyconfig.h:3:51: fatal error: x86_64-linux-gnu/python2.7/pyconfig.h: No such file or directory
  compilation terminated.
  In file included from /usr/include/python2.7/Python.h:8:0,
                   from plugins/python/uwsgi_python.h:2,
                   from plugins/python/pyutils.c:1:
  /usr/include/python2.7/pyconfig.h:3:51: fatal error: x86_64-linux-gnu/python2.7/pyconfig.h: No such file or directory
  compilation terminated.

  ----------------------------------------
  Failed building wheel for uwsgi
Failed to build uwsgi
ERROR: Failed to build one or more wheels
../../mk/spksrc.wheel.mk:36: recipe for target 'wheel_target' failed
make[1]: *** [wheel_target] Error 1
make[1]: Leaving directory '/spksrc/spk/python'
```

Diaoul:
If a previous build errored, try again after `make clean`

CondormanFr:
I do the build for the second time with make clean and I get the same error...

Dr-Bean:
UWSGI was updated on PyPI, and since we don't pin the version in WHEEL, we have a mismatch.
Either update the UWSGI cross package to latest, or pin the current version in Python's Makefile 
(`WHEEL = "uswgi==<whatever version>"`)

I already knew about it, but planned to hold off with updating modules until Python 2.7.11, after 5 dec...there's usually very low demand to build the Python SPK.

CondormanFr:
OK, i updated the cross package to latest and it works. Thanks.

Dr-Bean:
[...]

PR #1941, opened 2015-11-18
Update cross package to allow Python to build

CondormanFr:
I updated some cross module to allow Python build.
I added freetype to Pillow to allow pillow to create some picture (See #1886)


Dr-Bean:
Please remove `pyephem` and `psutil` from this PR. If you plan to update those packages, please do so in a separate PR together with updating to the SPK that uses them.

In general: if you update cross packages, split into separate PR's by SPK, especially if those packages are only used by one SPK. If needed, update the SPK alongside.
This applies in particular to python modules and wheels: those are (or ideally: should be) pinned to a specific version in the Makefile/requirements file to avoid compatibility issues.

Dr-Bean:
As per https://github.com/SynoCommunity/spksrc/issues/1939#issuecomment-158625298, I have updated uwsgi separately. 
Could you rebase your PR, and address my previous comment? The updates to libpng and pillow are still welcome of course. Thanks :)

CondormanFr:
OK. After I do the rebase, I need to create new PR , is it right ?

PR #1942, opened 2015-11-18
Add duplicity missing dependencies

adrien-delhorme:
All dependencies of dependencies were missing in duplicity package.


cross/rackspace-novaclient/Makefile
```diff
@@ -0,0 +1,15 @@
+PKG_NAME = rackspace-novaclient
```
Dr-Bean: Are you sure this module needs cross-compiling? According to the source code, there's not much to it.

spk/duplicity/Makefile
```diff
@@ -4,7 +4,8 @@ SPK_REV = 1
 SPK_ICON = src/duplicity.png
 
 BUILD_DEPENDS = cross/python cross/setuptools cross/pip cross/wheel
-BUILD_DEPENDS += cross/boto cross/netifaces cross/wrapt cross/$(SPK_NAME)
+BUILD_DEPENDS += cross/boto cross/msgpack-python cross/netifaces cross/rackspace-novaclient
```
Dr-Bean: `cross/msgpack-python` seems to be a run-time dependency (according to the requirements file, for example). If so, then this probably isn't going to work: `cross/msgpack-python` is compiled as a python module for inclusion in the python package, whereas you'd need to create a wheel to be able to include it in the duplicity package.

cross/rackspace-novaclient/Makefile
adrien-delhorme: If I do not cross compile, the wheel present in the wheelhouse is `rackspace_novaclient-1.5-cp27-none-linux_i686.whl` which is specific to arch i686.
When I cross compile, I get `rackspace_novaclient-1.5-cp27-none-any.whl`.
How could I do without cross compiling it?

spk/duplicity/Makefile
adrien-delhorme: Should I replace `include ../../mk/spksrc.python-module.mk` by `include ../../mk/spksrc.python-wheel.mk` in msgpack-python's Makefile? Will this not affect the python package?

cross/rackspace-novaclient/Makefile
Dr-Bean: I'd suggest to update pip, setuptools and wheel to latest. There was a fix for this type of packages in wheel 0.25: https://bitbucket.org/pypa/wheel/commits/985dbf6d1692980d6f393f2db317f855b6713e6d

spk/duplicity/Makefile
Dr-Bean: Yes, if you change the entry there to `python-wheel`, then you need to adjust the Python package as well. See what we did with uwsgi: https://github.com/SynoCommunity/spksrc/commit/0a3851f7255ace78020d49662ce18e9a910b7491 and https://github.com/SynoCommunity/spksrc/blob/master/spk/python/Makefile#L16.

There are other options too (set --system-site-packages when creating the virtualenv would be the main alternative), but I think I'd prefer the above option.

adrien-delhorme:
I updated `pip`, `setuptools` and `wheel` in PR #1993.
After the `wheel` update, `rackspace-novaclient` will install properly.
And I changed `msgpack-python` into a python wheel.

Dr-Bean:
Very nice :)

Dr-Bean:
@adrien-delhorme Can you please rebase?

adrien-delhorme:
Done. Sorry for the late answer.

spk/python/Makefile
```diff
@@ -13,7 +13,7 @@ DEPENDS += cross/sqlalchemy cross/pyaudio cross/pyalsa cross/lxml cross/twisted
 DEPENDS += cross/zope.interface cross/markupsafe cross/sphinxbase cross/pocketsphinx
 DEPENDS += cross/msgpack-python cross/psutil cross/pymongo cross/pyyaml cross/uwsgi
 
-WHEELS = psutil uwsgi
+WHEELS = psutil uwsgi msgpack-python
```
Dr-Bean: Almost ;) Could you put these in alphabetical order please? (Don't bother with the DEPENDS here though, that could use a cleanup)

Dr-Bean:
With that last comment fixed it should be good to go. For real ;)

spk/duplicity/src/requirements.txt
```diff
@@ -1,15 +1,27 @@
 # Cross compiled packages dependencies
 boto
 duplicity
+msgpack-python
 netifaces
+rackspace-novaclient
```
Dr-Bean: Minor thing: This is no longer cross-compiled, right?

adrien-delhorme:
Done!

Dr-Bean:
duplicity 0.7.05 -2 has been published on the repository. Thanks for working on it :)

adrien-delhorme:
Thank you for all your code reviews!

PR #1949, opened 2015-11-20
update plexivity to 0.9.8

cytec:


cytec:
@Dr-Bean i already merged an published but... technically `0.9.8` is able to support SSL if pyopenssl is installed which (if i remember correctly) ships with our version of python... 

When useing the current approach for installation i'll have to add pyopenssl to requirements.txt and then would need to build openssl on build time? because if there is no .whl pip install will not install anything and therefore the SPK will not start...

that means i/we need to add `--system-site-packages` to `virtualenv` again, right?

Dr-Bean:
Yep, cffi and cryptography both need cross-compiling, so just including them in requirements.txt wouldn't be enough. So yes, set `--system-site-packages`, and be done with it ;)

If we get to converting the current cross packages for cryptography, cffi and the other libs for SSL to wheels, we can reconsider the idea, but there's no need for that for the foreseeable future.

PR #1950, opened 2015-11-21
Update libpng and Pillow

CondormanFr:
As ask in #1941


cross/libpng/digests
```diff
@@ -1,3 +1,3 @@
-libpng-1.6.17.tar.xz SHA1 6394f3c0847c3b9ea69f4f972dddae3786592050
-libpng-1.6.17.tar.xz SHA256 98507b55fbe5cd43c51981f2924e4671fd81fe35d52dc53357e20f2c77fa5dfd
-libpng-1.6.17.tar.xz MD5 430a9b76b78533235cd4b9b26ce75c7e
+libpng-1.6.19.tar.xz SHA1 483D72CED11C9258F9D1119105273D9AF9FF151C
```
Dr-Bean: It seems you're using a different method to create the digests than using `make digests`? When using that command, the contents are in lower case. Maybe easier than your current method, just a suggestion :)

Dr-Bean:
LGTM. We can pick it up when updating the Python package. Pillow itself should be updated too.

CondormanFr:
I didn't know that there were a make digest ... It could be easier that my method.

Dr-Bean:
Did this one manually, libpng 1.6.19 was no longer available upstream. Libpng done with 14830c62f58c6e2dbcae4a2ab2a9b8746db41ae0, pillow via https://github.com/SynoCommunity/spksrc/commit/803ba88a077a88b49221a257c24deb92e307e4e3.
Thanks for opening the PR though.

PR #1951, opened 2015-11-22
Updated rTorrent to 0.9.6, libtorrent to 0.13.6, ruTorrent to 3.7

GoodOmens83:
Updated all the packages.  I not longer use the ruTorrent plugin folder as the plugins are now bundled with the 3.7 zip.  Perhaps you can disable that directory.

Everything seems to work wonderfully.  I tried and successfully complied for the qoriq syno.

In regards to the ongoing openbase dir issue people seem to report:

I noticed in I had /etc/php/conf.d/user-settings.ini with a path set.  Deleting that line caused openbase to start working (e.g. I no longer had to manually add my torrent dir to the basedir path in synology settings).

Perhaps people having issues also have that line as well?  Perhaps the installer script can look for that line and remove it if need be.  I'm not too good at bash to figure out how to do that....

Edit: Didn't realize that my later commits would be included in this pull request.  I've also updated xmlrpc to 1.33.18 and libxml2 to 2.9.3.

For xmlrpc I had to comment out two lines of the config in order for the AR and RANLIB files to pick up correctly.  I am doing this via the patch file.


cross/libtorrent-rakshasa/Makefile
```diff
@@ -14,7 +14,7 @@ LICENSE  = GPLv2
 
 GNU_CONFIGURE = 1
 PRE_PATCH_TARGET = myPrePatch
-CONFIGURE_ARGS = --disable-instrumentation
+CONFIGURE_ARGS = --disable-instrumentation --with-zlib=$(STAGING_INSTALL_PREFIX)
```
Dr-Bean: You should add `cross/zlib` to depends as well. Even though it comes along with openssl, better to be explicit.

cross/rutorrent/Makefile
```diff
@@ -20,6 +20,8 @@ include ../../mk/spksrc.cross-cc.mk
 
 .PHONY: myInstall
 myInstall:
+	mv $(WORK_DIR)/ruTorrent-master $(WORK_DIR)/$(PKG_NAME)
```
Dr-Bean: Please change PKG_DIR to `ruTorrent-master`, and drop this line. You'll need to make other changes accordingly. 
Setting EXTRACT_DIR might also work.

spk/rutorrent/Makefile
```diff
@@ -1,10 +1,10 @@
 SPK_NAME = rutorrent
-SPK_VERS = 3.6
-SPK_REV = 7
+SPK_VERS = 3.7
+SPK_REV = 1
```
Dr-Bean: `8`. The number goes up, not down :)

spk/rutorrent/Makefile
```diff
@@ -12,7 +12,7 @@ ADMIN_URL = /rutorrent/
 RELOAD_UI = yes
 STARTABLE = yes
 DISPLAY_NAME = ruTorrent
-CHANGELOG = "1. Update rtorrent to 0.9.4"
+CHANGELOG = 1. Update rtorrent to 0.9.4 2. Update rtorrent to 0.9.6 libtorrent to 0.13.6 rutorrent to 3.7
```
Dr-Bean: CHANGELOG = "Update rtorrent to 0.9.6, libtorrent to 0.13.6, rutorrent to 3.7"

Dr-Bean:
This is only a first quick look, so I might have other comments along the way. I do believe the open-basedir issues you mentioned work slightly differently than you think, but it requires a bit more explanation that I can't provide right now.

GoodOmens83:
Working on it - I also just noticed I'm getting a "file not found" for openssl 1.02d - looking in the FTP folder the openssl team seems to have deleted it.  I've updated to 1.0.2e on my local machine but won't update that change in my branch.

I don't know how to do separate pulls from the same repo - git newb here :)

GoodOmens83:
I think that took care of what you suggested - tried to compile and everything seemed to work / install on my qoriq machine.  Only change I didn't commit was the change to openssl.

spk/rutorrent/Makefile
```diff
@@ -12,7 +12,7 @@ ADMIN_URL = /rutorrent/
 RELOAD_UI = yes
 STARTABLE = yes
 DISPLAY_NAME = ruTorrent
-CHANGELOG = "1. Update rtorrent to 0.9.4"
+CHANGELOG ="Update rtorrent to 0.9.4"
```
Dr-Bean: Please readd the space here.

Dr-Bean:
I haven't compiled/tested it, but otherwise, looks good. 
Any idea if the open issues we have on ruTorrent are still valid one way or the other?

GoodOmens83:
Ok - I think I finally got the change log file the way you wanted it :)

What outstanding issue(s)?  Open Base-Dir?  Sort of - I found I had another file in my synology with a conflicting open base dir setting that was conflicting (/etc/php/conf.d/user-settings.ini) - commenting out that line resolved all issues I had.

Also just re-looking this over I notice in cross/rtorrent MakeFile a unsupported arch of powerpc is listed.  This compiled/installed fine on my qoriq machine.  Isn't qoriq a powerpc or do you list that seperately?

Dr-Bean:
No, I meant the [open issues](https://github.com/SynoCommunity/spksrc/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+rutorrent) we have on rutorrent.

`cross/rtorrent` sets UNSUPPORTED_ARCHS for a number of ppc-based arches. Qoriq isn't listed there, so it should compile. If it were a collection of arches, we'd probably refer to `$(PPC_ARCHES)` in spksrc.common.mk.

As for the open basedir: we add a particular file during installation (based on DSM version): https://github.com/SynoCommunity/spksrc/blob/master/spk/rutorrent/src/installer.sh#L74. It essentially sets the open-basedir for rutorrent to NULL. That file is applied after loading the default config in `/etc/php/conf.d/user-settings.ini`. 
In short, the contents in `user-settings.ini` apply globally, and the individual files we add apply only to that specific instance, based on the set path.
So either that behaviour is broken since we added it in, or it's specifically your device that doesn't behave as it should. I don't have the time to test it right now, but I'm quite sure it worked when it was initially set up, because we use it for a lot of other packages with the same expectations.

GoodOmens83:
Most issues seem related to openbase-dir issues (creating torrents, auto tools  moving files etc). Rutorrent seems to really require loose permissions. 

There are a few in there - like the ipfilter that can be setup with this version. Not sure about the rss issues as I don't use that feature.  

Besides having to clear out my openbase dir(which I still needed to do on the current reposit version) this is really stable and runs well on my box. 

Going forward possibly bundling lighttpd might be a better solution (instead of the built in dsm web server) as I never had a issue with moving files etc. But that could cause other problems. I can see about working on that.

GoodOmens83:
On further thought regarding openbase - it might not work as you designed if people add their torrent download directories outside the package directory. 

Not sure - would have to test but can't for the next few weeks as work has me away from home.

Issue #1965, opened 2015-11-30
Jackett


Diaoul:
https://github.com/Jackett/Jackett

piwi82:
[Installation on Linux/OSX](https://github.com/Jackett/Jackett#installation-on-linuxosx) :

1. Install Mono 4 or better :weary:

This is a nice app, let's hope the Mono stuff will work flawlessly.

Dr-Bean:
It does ;) Mono 3.x works, unless you count some unconfirmed issues with exceptions. Apart from that, Mono 4.x is available in beta, and Jackett was one reason for adding it.

Sidenote: I'll probably update Mono 4.x to latest in a bit, after I've worked out the PLIST and can confirm it works with Sonarr as well. Current version should work with Jackett without a hitch, several people are already using it.

Diaoul:
Didn't know you had that one on your TODO list ;)

Dr-Bean:
Well, I think you just pulled it off my list, and put it on yours :P

Diaoul:
Did I?

Dr-Bean:
I'm quite happy to give someone else a chance to package this. Think of the everlasting glory and praise you'll receive. @piwi82 sounded quite enthusiastic earlier...how about it?

Diaoul:
Totally agree :+1:

piwi82:
:sweat_smile: I don't use this app, I have just take a look at what it does.
Don't expect me to package this one :wink:

BTW: I have pushed [Adminer](https://github.com/SynoCommunity/spksrc/commit/9aa714df262ca0d7fcca6b9bf11d554933697fc6) to the repository. I don't remember how to publish it as beta (API key ? URL to post stuff ? ...)

Dr-Bean:
Well...back to Diaoul then, I suppose ;)
All joking aside, I should have the code stashed somewhere I think. I still needed to figure out a way to get start-stop-daemon to work correctly, because iirc, it didn't work when I set it up identical to Sonarr, but I haven't looked at it since.

To publish as beta, set `BETA = 1` in the Makefile.

Dr-Bean:
Done with https://github.com/SynoCommunity/spksrc/commit/aeac78cb0b82d6bb5d69aa68867f91a983e2a7da.

Dr-Bean:
Jackett 0.7.33-1 has been published.

PR #1966, opened 2015-12-01
Add Jackett

Dr-Bean:
As per https://github.com/SynoCommunity/spksrc/issues/1965, because we'd never get it done otherwise ;)

The SPK_DEPENDS might not work as-is (better to be explicit and use the full version anyway), but I'll fix that together with all the other comments from @Diaoul ;)


Diaoul:
Awesome work, as always. I have nothing to add here.

Dr-Bean:
Thanks. One other thing: there's a blackhole feature, so we need to add the `jackett` user to the `sc-download` group. Must've been added after I set up the code initially.

PR #1976, opened 2015-12-05
Update TT-RSS to Git commit f61f2a26 and fix on DSM-6

saschpe:
DSM-6.0 moved PHP into a separate package. Both /bin/php and /usr/bin/php
are minimal variants lacking the necessary MySQL PHP modules. Therefore
depend on package PHP5.6 on DSM build 7135 or later and use
/usr/local/bin/php56.


saschpe:
Tested on DSM-6.0-7135. Can't hurt if someone checks on DSM-5.x

spk/tt-rss/src/dsm-control.sh
```diff
@@ -9,7 +9,11 @@ INSTALL_DIR="/usr/local/${PACKAGE}"
 WEB_DIR="/var/services/web"
 PATH="${INSTALL_DIR}/bin:${PATH}"
 USER="$([ $(grep buildnumber /etc.defaults/VERSION | cut -d"\"" -f2) -ge 4418 ] && echo -n http || echo -n nobody)"
+if [ `/bin/get_key_value /etc.defaults/VERSION buildnumber` -ge "7135" ]; then
+PHP="/usr/local/bin/php56"
+else
 PHP="/usr/bin/php"
+fi
```
Diaoul: Is it possible to make this a oneliner like USER right above?

Dr-Bean:
I don't know exactly why Synology chose `php56` as the binary name, but that's going to be fun when Synology decides to package PHP7.0...
Other than the comment Diaoul provided, LGTM.

saschpe:
@Dr-Bean I don't know why we now need to have three php binaries across the system :-)

Dr-Bean:
Probably backwards compatibility to a degree, at least during the beta. My guess is that the other binaries will disappear, or be replaced with a symlink pointing to the binary of the new package e.g. when DSM6 final is released.

One last thing I didn't notice before: could you remove `PKG_FULL_NAME`? That's no longer necessary.

spk/tt-rss/src/dsm-control.sh
```diff
@@ -8,8 +8,8 @@ DNAME="Tiny Tiny RSS"
 INSTALL_DIR="/usr/local/${PACKAGE}"
 WEB_DIR="/var/services/web"
 PATH="${INSTALL_DIR}/bin:${PATH}"
-USER="$([ $(grep buildnumber /etc.defaults/VERSION | cut -d"\"" -f2) -ge 4418 ] && echo -n http || echo -n nobody)"
-PHP="/usr/bin/php"
+USER="$([ $(/bin/get_key_value /etc.defaults/VERSION buildnumber) -ge 4418 ] && echo -n http || echo -n nobody)"
+PHP="$([  $(/bin/get_key_value /etc.defaults/VERSION buildnumber) -ge 7135 ] && echo -n /usr/local/bin/php56 || echo -n /usr/bin/php)"
```
Dr-Bean: And remove the unnecessary space ;)

saschpe:
Note quite sure if it's a TTRSS bug but I had to do this to get the update daemon to work:
 php -r "echo time().\"\n\";" > lock/update_daemon.stamp
Would be great if someone could double-check.

Dr-Bean:
Last question was meant for DSM6? Can't test that myself.
I assume that the new WebStation package is still running as `http`, so it's not a permissions issue?

saschpe:
@Dr-Bean Yes, that was on DSM6. Not it's not a permission issue TT-RSS simply expects the file to be there. I'm pretty sure it's a bug since the update_daemon doesn't create it itself if missing.

So I'm unsure if we should add the above php line to the postinstall function...

Dr-Bean:
That seems to be a bug or a change in behavior in TT-RSS then? The lock files _are_ created with the current package. I only did a quick check on DSM5: I updated to the commit in this PR, and the same problem occurs: no lockfiles.
Can you check if the current TT-RSS package (with the necessary changes to the PHP version) does work on DSM6?

nickbroon:
:+1:

Dr-Bean:
@saschpe Not sure if you read my previous comment?

saschpe:
Seems like the issue was solved recently. The package works for me after upgrading to commit 7f4fff96. Could you please double-check?

nickbroon:
Linking to tt-rss DSM6 discussion. http://forum.synology.com/enu/viewtopic.php?f=260&t=105868

Dr-Bean:
Seems to work as intended on DSM5 ;) Thanks for taking care of it :)

Dr-Bean:
tt-rss 20160102-10 has been published on the repository.

nickbroon:
Using the new tt-rss 20160102-10 I'm seeing the following in the tt-rss logs:

`E_WARNING (2)	include/functions2.php:261	file_get_contents(lock/update_daemon.stamp): failed to open stream: No such file or directory [Last query: SELECT ref_id AS id FROM ttrss_user_entries WHERE owner_uid = 1 ORDER BY ref_id DESC LIMIT 1]
admin	0:23`

and the update daemon fails to update the feeds as a result.

Dr-Bean:
Does the file exist? (`/volume1/web/tt-rss/lock/update_daemon.stamp` presumably)

nickbroon:
Unfortunately no.

Dr-Bean:
Try a fresh install maybe...it does create it here, and iirc I tested with both an upgrade and an actual clean install.
Otherwise, you can create probably it with the command @saschpe mentioned [here](https://github.com/SynoCommunity/spksrc/pull/1976#issuecomment-162310136)...make sure you're located in the correct dir of course.

nickbroon:
It was a fresh install on DSM6 after uninstalling the old package.
I'll try  the command you mention.

Dr-Bean:
Ah...DSM6. Still can't test that, so I'll leave that to @saschpe: he said it worked ;)

saschpe:
Works for me, though I'll re-test by uninstalling and wiping the installation. Either way, we could add the one-liner from above into the post-install function...

saschpe:
Nope, works. Just un-installed the package completely and did a fresh install.

@nickbroon Sure you installed the latest version?

reyman:
I installed the version which appear today. Uninstall change nothing. Where can i find the log ?

nickbroon:
@saschpe Did you uninstall the package and also manually remove the 'tt-rss/lock/update_daemon.stamp' to make sure you have a clean starting state. (Or perhaps completely remove the tt-rss directory)

I'm definitely using 20160102-10, installed from synocommunity using Package Manager.

saschpe:
@nickbroon Quick shot at fixing it: https://github.com/SynoCommunity/spksrc/pull/2056

LawrenceAusten:
FWIW, I did the following and everything, including updating, seems to be working fine:

rm /usr/syno/bin/php
ln -s /usr/local/bin/php56 /usr/syno/bin/php    // Side effects on other packages?

change '/usr/bin/php' to '/usr/local/bin/php56' in /var/services/web/tt-rss/config.php

Restarted package via regular DSM web interface

PR #1983, opened 2015-12-06
Update digests openssl to Version 1.0.2e

roterBulle:
OpenSSL Version has changed to 1.0.2e so Makefile and digests had to be updated to work correctly


Diaoul:
Please make 1 PR with code + digest changes.

PR #1993, opened 2015-12-10
Update setuptools, pip and wheel to latest versions

adrien-delhorme:


Dr-Bean:
I did this one myself, setuptools was updated twice since I looked at it ;) Thanks for the effort though.

PR #1997, opened 2015-12-13
Update zsh to 5.2

alt219:
Update zsh from 5.0.7 to 5.2.


spk/zsh/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = zsh
-SPK_VERS = 5.0.7
-SPK_REV = 2
+SPK_VERS = 5.2
+SPK_REV = 0
```
Diaoul: This is incremental only, so 3.

alt219:
Apologies for not using a topic branch and for getting SPK_REV wrong in the first place.  8f01ec5 (or newer) should be correct now.  If it's easier, I can close this pull request and reissue a new pull request.

Dr-Bean:
Maybe you'd want to correct the account (name) associated with the PR.

alt219:
Thanks, will do.

Dr-Bean:
@alt219 Were you still planning on fixing this?

alt219:
Do you think closing this PR and moving it to a topic branch would be a better approach?  That would be easier/better for me in the long run.

Dr-Bean:
Technically, it's not required, but it is best-practice to do so. Using a topic branch makes it easier to keep your master branch in sync with origin. So if you'd prefer to do that, go ahead.

alt219:
Moved to separate topic branch, see #2070.

PR #2013, opened 2015-12-19
support braswell D716+

yeled:
This brings in support for D716+.

I've tested syncthing, sabnzbd-testing, tmux, transmission... so far.


Dockerfile
```diff
@@ -34,7 +34,8 @@ RUN apt-get update && \
         swig \
         xmlto \
         zlib1g-dev && \
-    apt-get clean && \
+        gcc-multilib
```
Dr-Bean: Are you not running the supplied 32-bit Docker instance or something like that? This should not be necessary, and causes problems for certain packages.

cross/syncthing/Makefile
```diff
@@ -38,7 +38,7 @@ $(error Unsupported ARCH $(ARCH))
 endif
 
 ifeq ($(findstring $(SYNCTHING_ARCH),amd64),$(SYNCTHING_ARCH))
-SYNCTHING_BIN = $(WORK_DIR)/$(PKG_DIR)/bin/linux_amd64/syncthing
+SYNCTHING_BIN = $(WORK_DIR)/$(PKG_DIR)/bin/syncthing
```
Dr-Bean: Please revert this, otherwise you're break things for people running on 32-bit.

kernel/syno-braswell-5.2/digests
```diff
@@ -0,0 +1 @@
+x64-source.txz MD5 3051b37fc390e0be6cebeae9bf569775
```
Dr-Bean: Please use `make digests`

Dr-Bean:
Thanks for the effort. The PR is not complete yet though: there are several packages that need to know about `braswell` still (e.g. debian-chroot, and there are others)
Apart from that, could you check if packages created via x64 work for braswell as well? See https://github.com/SynoCommunity/spksrc/blob/master/toolchains/syno-x64-5.2/Makefile#L3

toolchains/syno-braswell-5.2/Makefile
```diff
@@ -0,0 +1,28 @@
+TC_NAME = syno-$(TC_ARCH)
+
+TC_ARCH = braswell
+TC_VERS = 5.2
+TC_FIRMWARE = 5.2-5565
```
Dr-Bean: Should probably be 5644 here.

yeled: thanks for all your help! well spotted.

cross/syncthing/Makefile
yeled: hrmm ok, I will fix it. I did test on `armada370` and `cedarview` as well. Do you have another favorite to use as a canary build?

Dockerfile
yeled: huh. well spotted. I need to run my docker container with [this 64b->32b trick](https://github.com/docker/docker/issues/611) and then try all this again.

```
root@f4b7d15c7fe6:/spksrc/native/python# uname -a
Linux f4b7d15c7fe6 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt11-1+deb8u6 (2015-11-09) x86_64 GNU/Linux
```

Diaoul:
Done with 3088bdca0e991a22ec3a94d25a8d39baf863aba9

Diaoul:
I did not test (I don't own a braswell NAS) but if it doesn't work we will fix it, better to get things moving.

PR #2017, opened 2015-12-20
Update to flac 1.3.1

vkostas:
Set new download link.
Fix config file to avoid error (as per gihub flac).

### DIFF
--- a/cross/flac/Makefile
+++ b/cross/flac/Makefile
@@ -1,8 +1,8 @@
 PKG_NAME = flac
-PKG_VERS = 1.2.1
-PKG_EXT = tar.gz
+PKG_VERS = 1.3.1
+PKG_EXT = tar.xz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = http://downloads.sourceforge.net/project/flac/$(PKG_NAME)-src/$(PKG_NAME)-$(PKG_VERS)-src
+PKG_DIST_SITE = http://downloads.sourceforge.net/project/$(PKG_NAME)/$(PKG_NAME)-src
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 DEPENDS = cross/libogg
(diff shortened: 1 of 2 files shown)


Dr-Bean:
This should be part of a larger PR, instead of on its own. Are you working on mpd? Maybe you can include it as part of that PR.

If you do, please add digests (`make digests`), update the PLIST, and minimize the patch size (don't comment out, just remove)

vkostas:
Yes, I am working on mpd.
I will include icecast to mpd PR then.
Thx

Dr-Bean:
`flac`, right? ;) Besides that typo, you might want to check that PPC arches actually compile with flac. They probably won't.

vkostas:
Yes, FLAC
I mixed it up with icecast request :-).

vkostas:
ah, I see...
To locate PPC, I had first:

>    ifeq ($(findstring $(ARCH),<b>powerpc ppc824x ppc853x ppc854x qoriq</b>),$(ARCH))
>      CONFIGURE_ARGS += --disable-altivec
>    endif*
>    include ../../mk/spksrc.cross-cc.mk

Then I went for "dynamic" logic
>    include ../../mk/spksrc.cross-cc.mk
>    ifeq ($(findstring $(ARCH),<b>$(PPC_ARCHES)</b>),$(ARCH))
>      CONFIGURE_ARGS += --disable-altivec
>    endif

I could set *CONFIGURE_ARGS* after the *cross-cc.mk* but it doesn't look good to me.

>    include ../../mk/spksrc.cross-cc.mk
>    ifeq ($(findstring $(ARCH),<b>$(PPC_ARCHES)</b>),$(ARCH))
>      CONFIGURE_ARGS += --disable-altivec
>    endif

I see node/Makefile and mono/Makefile use PPC_ARCHES before cross-cc.mk.
Any hint???

Dr-Bean:
That functionality currently only works for spk Makefiles, not cross Makefiles. Just use the arches themselves, all the other cross packages do.

Issue #2007, opened 2015-12-17
Move maintainer, distributor, report_url to local.mk?


Dr-Bean:
Instead of defaulting to SynoCommunity for maintainer (-url), distributor (-url) and report_url, we should consider moving them to local.mk. 
This prevent packages from being published outside of SynoCommunity's repo, but under our name. See #2003 for example. Those packages point to SynoCommunity, causing confusion as to who is responsible for those packages.

In doing so, unfortunately, package authors who wish to be credited won't be, unless they publish their packages on the repo themselves. A possible solution might be to set `maintainer` in the spk's Makefile instead.

@SynoCommunity/developers?

piwi82:
:+1:

Dr-Bean:
I just realised the last part is probably not an issue. We already provide the ability to set the maintainer in the Makefiles obviously...After defining the variables in local.mk, it's should be sufficient to delete all instances of `MAINTAINER = SynoCommunity` from spk/*/Makefile, and we're done.

Diaoul:
I believe there is a maintainer branch that does just that. However that's not implemented in the backend.

Dr-Bean:
It's not the main goal anyway ;) I'll set up a PR in a bit.

Do you still want to have the the maintainer branch merged then (as far as it'll go anyway)?

Diaoul:
I think maintainer should be kept in the Makefile. Ultimately it should be set to the package author in some way or at least point to it as a link. It's easy enough to change if you fork spksrc and won't prevent updates unless conflicting with another changeset upstream.

I agree for the others, spksrc should be repository agnostic (asides from local.mk)

Dr-Bean:
Take a look: https://github.com/SynoCommunity/spksrc/pull/2014.
If MAINTAINER is set in the SPK's Makefile, it'll take precedence. Otherwise, it'll pick it out of local.mk/error out. It's the only required field of the bunch.
All the others should be set in local.mk, all to be set manually.

Diaoul:
Awesome then :+1:

Dr-Bean:
I'll merge this in a bit. For the record, MAINTAINER_URL behavior has changed. The URL should probably be built on the repo side, but we don't use it atm so no urgency.

Dr-Bean:
Merged with 5a17600350866ace6170943668f40403dc9abc1a.
@SynoCommunity/developers, please add the following to your local.mk, or regenerate.
```
MAINTAINER?=SynoCommunity
[...]

PR #2024, opened 2015-12-22
Set package maintainers based on `maintainer` branch

Dr-Bean:
As per https://github.com/SynoCommunity/spksrc/issues/2007, setting the maintainers to their original authors for a number of packages, based upon the `spksrc/maintainer` branch.

New branch due to merge conflicts. We can delete the `spksrc/maintainer` branch after merging this.

### DIFF
--- a/spk/bicbucstriim/Makefile
+++ b/spk/bicbucstriim/Makefile
@@ -6,7 +6,7 @@ DSM_UI_DIR = app
 
 DEPENDS  = cross/$(SPK_NAME)
 
-MAINTAINER = SynoCommunity
+MAINTAINER = Diaoul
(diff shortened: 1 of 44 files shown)


Dr-Bean:
@Diaoul, good with this? After merging, we can start instructing contributors to set their own name instead of SynoCommunity for the MAINTAINER field as well.

Dr-Bean:
Calling this good to merge. If need be, we can adjust the names to remove inactive maintainers, but it doesn't hurt anyone to leave as is (alternative: set an `author` key/value, and use that on the repository as well)

We'll also start instructing people to set themselves as maintainer, see if we can improve taking ownership and instill that there's a certain level of responsibility once a package has been accepted (with or without push access to the repository: as far as I'm concerned, that's a case-by-case basis, depending on maintainer activity)

Issue #2026, opened 2015-12-24
Web UI Plugin Manager Issues (Syn DS412+)


arijaycomet:
Running the latest updated OctoPrint (v20151221-2) on my Synology DS412+ (on DSM 5.2-5644 Update 2).  Things run great, CuraEngine works awesome, etc.  However, I'm unable to add any additional Plug-ins using the Web UI Plugin Manager.  Specifically, Email Notifier is the one I'm trying to get going.

Below is a copy (cut/paste) of the log it shows on screen through the UI when attempting to complete the installation.  My Synology setup is very basic (and I'm an Intermediate user at best, no expert!).  Nothing is rooted or hacked or whatever... so if I need to do some command line changes or something let me know (and being wet behind the ears, I may need hand-holding there LOL)

Please advise/help with any advice to help get this to install correctly.  Thanks!

--- [ copy/paste of on-screen logging during plugin install attempt ] ---

Installing plugin "Email Notifier" from https://github.com/anoved/OctoPrint-EmailNotifier/archive/master.zip...
/usr/local/octoprint/env/bin/pip install https://github.com/anoved/OctoPrint-EmailNotifier/archive/master.zip
Collecting https://github.com/anoved/OctoPrint-EmailNotifier/archive/master.zip
Downloading https://github.com/anoved/OctoPrint-EmailNotifier/archive/master.zip
Command "python setup.py egg_info" failed with error code 255 in /tmp/pip-WKtTUH-build
Complete output from command python setup.py egg_info:
Could not import OctoPrint's setuptools, are you sure you are running that under the same python installation that OctoPrint is installed under?

----------------------------------------

Done!

markbastiaans:
I've managed to reproduce the issue. I've found the cause, am currently working on a fix and will do a new PR for OctoPrint soon.

Dr-Bean:
octoprint 20160103-3 has been published.

PR #2029, opened 2015-12-27
Fix for OctoPrint plugin installation

markbastiaans:
As mentioned in #2026, some OctoPrint plugins which depend on `octoprint-setuptools` are not installing. I've fixed this by doing the following:
* **Added OctoPrint to cross.** It now downloads the latest release from the GitHub archives instead of doing a `git clone`.
* **Patched the hardcoded requirements in the OctoPrint `setup.py`** to the most up-to-date versions of the `netifaces`, `psutil` and `pyyaml` wheels in the repository. The alternative was to add the versions OctoPrint needs to cross, but I ran into problems building `netifaces 0.10.0` (as per [this issue](https://bitbucket.org/al45tair/netifaces/issues/1/gcc-fails-while-installing-0100-0101)).
* **Moved the `tornado` wheel to `tornado-4.0.1`** since this version is not the latest version of `tornado` (which seems to be 4.3).
* **Fixed the OctoPrint requirements and installation** script to use the correct wheels and to run `setup.py`.

I've tested this fix on `evansport` and the plugin issue seems to be fixed with these changes.

@Dr-Bean, could you review this PR? I'm also interested in your preference in navigating dependency hell. My current approach should work until OctoPrint is updated, but it's not as future-proof as including every dependency in cross. Thanks in advance!


spk/octoprint/src/installer.sh
```diff
@@ -33,6 +33,9 @@ postinst ()
     # Install the wheels/requirements
     ${INSTALL_DIR}/env/bin/pip install --use-wheel --no-deps --no-index -U --force-reinstall -f ${INSTALL_DIR}/share/wheelhouse -r ${INSTALL_DIR}/share/wheelhouse/requirements.txt > /dev/null 2>&1
 
+    # Install OctoPrint
+    cd ${INSTALL_DIR}/share/${DNAME} && ${INSTALL_DIR}/env/bin/python setup.py install > /dev/null 2>&1
```
Diaoul: This assumes DNAME of the package is the same name as the git repo. I think I'd prefer hardcoded OctoPrint here instead. This is more readable that using a variable that's not meant for that use.
DNAME refers to Display Name, it's the package name as displayed in the package center and for any GUI rendering.

Diaoul:
Does this download anything at install time?

markbastiaans:
As far as I can see, no.

The requirements in `requirements.txt` are those after a `pip freeze --local` done after running `setup.py`. While debugging, I also did an install with `setup.py` piping its output to a log file and there was no mention of downloading anything in there.

Dr-Bean:
>Patched the hardcoded requirements in the OctoPrint setup.py to the most up-to-date versions

That wouldn't work with Tornado? (As in: Leave cross/tornado at latest and patch Octoprint to use that, instead of adding a separate cross package for an older version of tornado)

As far as future-proofing: I think the approach you took is generally the best thing to do from an spksrc perspective. Adding separate cross packages is going to be the same hell, just a different corner ;)
Imo, the best thing to do would be to help upstream in getting their dependencies/requirements updated, so we don't have to patch things downstream.

Dr-Bean:
Personally, I don't mind leaving the commented-out entries in requirements.txt: it may be easier to notice when updating the reqs from a fresh `pip freeze`, but I'm good either way.

Other than the one comment on the Tornado version, LGTM.

markbastiaans:
My reasoning for using 4.0.1 instead of the latest version of Tornado (4.3) was that Octoprint is currently the only package in the repository using the Tornado cross package, which I included in the previous PR with a really minor version difference (4.0.2 instead of 4.0.1, my mistake).

If you want me to include 4.3 explicitly, let me know. I haven't tried Tornado 4.3 yet with OctoPrint. I can check but that will be next week at the earliest, since my development machine killed itself during the holidays.

Dr-Bean:
Ah, right. I thought it was 4.0.3 for some reason, not 4.3. Lets go with this.

cross/octoprint/Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+HOMEPAGE = http://octoprint.org/
+COMMENT  = OctoPrint is a so called host software for 3D printers that controls your 3D printer and sends it the actual commands to do its job.
+LICENSE  = AGPLv3
+
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = nop
+INSTALL_TARGET = myInstall
+
+PATCHES_LEVEL = 1
```
Dr-Bean: Patch level is 1 by default, so no need to set it.

Dr-Bean:
Merged with https://github.com/SynoCommunity/spksrc/commit/62b6a83b2bcec12d564750ab3f5798660e1600c6, additional change with https://github.com/SynoCommunity/spksrc/commit/73851c511810854cb1207aa702e483ecc6a69c65.
Building, but it'll take a while before it's on the repository.

markbastiaans:
Cheers, thanks!

Dr-Bean:
For next time : could you please create your patches with `diff`, instead of `git diff`? Thanks.

markbastiaans:
No worries, will do!

PR #2032, opened 2015-12-30
Initial attempt at a epic5 build

NathanHowell:
This is enough to get it working on my box, I haven't tried to get the packaging done yet.

### DIFF
--- a/cross/epic5/Makefile
+++ b/cross/epic5/Makefile
@@ -1,0 +1,25 @@
+PKG_NAME = epic5
+PKG_VERS = 1.4
+PKG_EXT = tar.xz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = http://ftp.epicsol.org/pub/epic/EPIC5-PRODUCTION
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS = cross/openssl cross/ncurses
+
+HOMEPAGE = http://www.epicsol.org/
+COMMENT  = Internet Relay Chat client for UNIX like systems
+LICENSE  = BSD
+
+GNU_CONFIGURE = 1
+CONFIGURE_ARGS = --with-ipv6 --with-ssl --with-wserv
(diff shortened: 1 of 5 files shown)


cross/epic5/patches/config.cache.patch
```diff
@@ -0,0 +1,4 @@
+--- /dev/null     2011-08-22 20:38:12.000000000 +0200
++++ config.cache  2011-12-16 23:15:45.000000000 +0100
+@@ -0,0 +1 @@
++ac_cv_func_getpgrp_void=yes
```
Diaoul: Pass this as CONFIGURE_ARGS in cross

NathanHowell: Any idea how? I couldn't figure out a way, but it does work as an environment variable, see the most recent push.

cross/epic5/Makefile
```diff
@@ -0,0 +1,19 @@
...
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS = cross/openssl cross/ncurses cross/libiconv
+
+HOMEPAGE = http://www.epicsol.org/
+COMMENT  = Internet Relay Chat client for UNIX like systems
+LICENSE  = BSD3
+
+GNU_CONFIGURE = 1
+CONFIGURE_ARGS = --with-ipv6 --with-ssl --with-wserv --with-iconv
```
Diaoul: Add `ac_cv_func_getpgrp_void=yes` to CONFIGURE_ARGS here. It should work, if not, there is something wrong elsewhere.
See other cross packages for example, there are other with configure checks overrides.

NathanHowell: It doesn't work in `CONFIGURE_ARGS`. Both other ways do work however.

NathanHowell: Alright, it does work if `configure` is regenerated. I'm pushing a fix shortly.

Dr-Bean:
Could you please open a new PR which includes the rest of the SPK as well as this cross package? It's not very likely we'll merge this on its own.

NathanHowell:
Sure, I won't have time to work on this again for a bit but I'll reopen it when packaging is ready.

PR #2041, opened 2016-01-03
Fix libevent download location

jokey2k:


jokey2k:
Makes the example work again

Dr-Bean:
Duplicate of #2035.

PR #2045, opened 2016-01-04
Added net-snmp dependency

mikejsutherland:
For review, I would like to include net-snmp as it will be a dependency of another pkg currently under development.


Dr-Bean:
I would prefer it that you open one larger PR which includes this package and the SPK itself. Reviewing bit-by-bit doesn't make things easier, and I prefer not to have dependencies between PR's.

Having said that, just a couple comments for now: 
Please take a look at the wiki article on PLIST, specifically on removing the static libs (`.a`, `.la`).
As for `CONFIGURE_ARGS`, any particular reason to use SNMPv1? v3 is preferred for security reasons. Besides that, `--with-persistent-directory should probably be set to $(INSTALL_PREFIX) instead of a hardcoded path.

mikejsutherland:
I initially kept net-snmp as a separate branch/pull request for two main reasons.  Someday net-snmp could become a full spk itself, including the necessary snmp binaries, etc. The other reason being that I was treating these libs similar to other shared libs like openssl & libxml2 which are dependencies in multiple spk projects. I have no problem merging this into my other SPK project branch to make things easier.

I will remove the .a and .la files and adjust the other args are requested. I have no reason for preferring snmp v1 over v3 other than not wanting to exclude other projects which may use these libs.

hgy59:
Closing this outdated request.
There is no dependent package and DSM has integrated snmp support.

PR #2070, opened 2016-01-10
Update zsh to 5.2

alt219:
Update zsh from 5.0.7 to 5.2.


Dr-Bean:
Generally when SPK_REV goes up, a CHANGELOG entry makes sense. We can add it when merging and publishing a new package, but feel free to add one in the PR.
Good to go ;)

Dr-Bean:
zsh 5.2-3 has been published on the repository. Thanks for the PR :)

PR #2071, opened 2016-01-12
Add GNU bc

alt219:
Add GNU bc 1.06, an arbitrary precision numeric processing language.


Dr-Bean:
This, and the `less` PR are getting quite specific. We're not trying to convert the NAS to a full-blown workstation: you might be better off using debian-chroot if that's the case.

Maybe I'll think differently if all those small utilities/tools were consolidated into one package. I was against that in the past, but if there was ever a time...

alt219:
Completely understood.  This and `less` are the last two tools I built for my own use.  Figured I would share the love rather than keep them to myself.

Diaoul:
We're not against this PR but I agree with Dr-Bean, this is really specific. I was thinking of a web tools package but then comes the problem of updating it each time a new release comes out and deciding what it contains. The downside of a dedicated package is that having tons of packages on the repository, overwhelming the users with package that seem to specific.
Now is the good time to decide which way we should go.

Dr-Bean:
Keeping it up-to-date is an ongoing effort either way, so I don't see that as a sticking point myself. A dedicated maintainer for such a package would make a lot of sense though.

Content on the other hand...apart from these two PR's, I have no idea. Probably because I don't use the device for that purpose, so I'm not missing anything in particular.
@alt219 Maybe you have some ideas for that?

cytec:
im in for a "combined" package as well... I've already done that for myself with some tools i use at a regualar basis like `nano` `htop` `ldd` maybe do something like "cli essentials" or stuff like that and combine all those tools together?

Pimmetje:
We can make a list (just out the top of my fingers):

- nano
- vim
- htop
- ldd
- bc
- sudo
- nice
- screen

I think we can add some more. We should group them so you don't have to install all. But making a package for every one of them is overkill if u ask me.

hgy59:
Now that we have synocli-file shall we add `bc` to it?
Or move `bc` from spk to diyspk to be prepared for a synocli-dev package?

ymartin59:
@hgy59 My idea was to rebase that contributed commit (to keep original author and date) and include it into "synocli-dev"

PR #2073, opened 2016-01-12
Fix ncurses compilation bromolow with gcc5

jekkos:
When building ncurses with gcc5 for bromolow following error occurs

````
#define mouse_trafo(y,x,to_screen) wmouse_trafo(stdscr,y,x,to_screen)
                                                        ^
Makefile:1323: recipe for target '../obj_s/lib_gen.o' failed
````

Patch was added in 

https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/sys-libs/ncurses/files/ncurses-5.9-gcc-5.patch?view=markup 

and fixes the issue.

### DIFF
--- a/cross/ncurses/patches/work_around_changed_output_of_GNU_cpp_5.x.patch
+++ b/cross/ncurses/patches/work_around_changed_output_of_GNU_cpp_5.x.patch
@@ -1,0 +1,40 @@
+Building ncurses with GCC 5.1 (or more precisely, with its 'cpp') fails with
+a syntax error, caused by earlier preprocessing.
+
+(I'm not entirely sure whether it's a GCC bug or rather caused by a new
+feature which breaks further processing with 'awk' and 'sed';  I *think*
+at least the 'awk' inline script "AW2" simply isn't prepared for the changed
+output of 'cpp' w.r.t. line directives [1].  Anyway, the patch fixes the issue.)
+
+[1] https://gcc.gnu.org/gcc-5/porting_to.html
+
+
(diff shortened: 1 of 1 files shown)


Dr-Bean:
I don't understand how bromolow and gcc5 are related, but anyway, that patch (or something similar) was included in ncurses over a year ago. It should be part of ncurses 6.0, which is what's included in spksrc. Is your fork up-to-date?

jekkos:
I have gcc5 on my local machine. I'm not sure which toolchain is used when compiling for 64bit linux (bromolow in this case). Isn't this by default the compiler you have installed on the linux system?

Indeed I would expect the same, that when compiling ncurses 6 that fix would already have been included, however after adding the patch I saw int he logs that it it applied successfuly (and compile works after that), so for some reason it might not be present in the source used by spksrc at this moment.

Dr-Bean:
I would suggest to use the spksrc Docker container instead or check the requirements in the README on how to set up a VM. 
Using a standardized build environment makes it a lot easier to verify and reproduce issues like this. Providing support on build-related errors isn't feasible otherwise.

jekkos:
Ok thanks for the info. I'll try to compile using the docker image first and see whether this issue occurs or not.

Diaoul:
Are you compiling using `make arch-bromolow`?

jekkos:
Yes indeed, make arch=bromolow

Dr-Bean:
Type, or are you actually using an `=` sign? Try with `-` instead.

Dr-Bean:
@jekkos I'm closing this PR, as I can't reproduce it. If you continue to run into this problem, bearing in mind the suggestions above, let us know.

PR #2077, opened 2016-01-14
Pyload update

Dr-Bean:
- Update Pyload dependencies
- Add patch to fix issues where updating pyload plugins is not possible. 
The current maintainers have apparently lost control over the pyload.org domain where the updates previously came from. The added patch is a workaround until a new version of pyload is released.


Diaoul:
LGTM

Dr-Bean:
pyload 0.4.9-2 has been published on the repository.

PR #2119, opened 2016-01-31
OpenSSL library update to version 1.0.2f

tedstriker:
version 1.0.2e is not available for download anymore

PR #2129, opened 2016-02-08
Updated the openssl version to the current one 1.0.2f

Pimmetje:
Version 1.0.2e is not anymore on ftp://ftp.openssl.org/source/. This makes it use the new f version


Pimmetje:
Oops i missed the other pull

Pimmetje:
#2119 for reference

PR #2147, opened 2016-02-19
Start syncthing with proper umask

gotson:
busybox’s start-stop-daemon doesn’t support umask
using ‘su’ and ‘nohup’ to achieve the result


Tstassin:
Your proposed fix works really well. Thank you.
Hope this will be pushed to the package soon.

gotson:
Closing since https://github.com/SynoCommunity/spksrc/pull/2646 will allow to use umask with start-stop-daemon

PR #2173, opened 2016-03-05
Update OctoPrint to 1.2.9

markbastiaans:
OctoPrint was recently updated to 1.2.9, so I've updated the OctoPrint SPK accordingly. These are my changes:
* Updated OctoPrint cross to 1.2.9 stable
* Updated OctoPrint SPK version
* Added ```awesome-slugify``` and ```regex``` requirements to OctoPrint SPK
* Added ```regex``` to cross

Built and tested on ```arch-evansport```.
Let me know if I need to do any more changes to get this PR merged.


Diaoul:
LGTM

Dr-Bean:
I've merged this with 615b789081ac2575abeb75042827a7a8de6b3b70. 

Note that psutil included in the Python package was updated to 4.1.0: this will probably affect OctoPrint when the Python package is (re)built. This applies to the armada38x arch, for example.

markbastiaans:
Thanks for merging! Let me know if and when I should do another PR to address the ```psutil``` version.

What I tried to do (in another branch) is include ```psutil``` 3.2.1 in cross. This seems to work while installing Octoprint. Unfortunately, installing plugins fails with this approach: the installation process (specifically ```pip```) runs into version conflicts (as there seem two versions of the library it can choose from it tends to choose the latest version instead of 3.2.1).

I'll try to get to the bottom of this for the next Octoprint release.

Dr-Bean:
You can try two things right now:
- See if psutil 4.x work with Octoprint. Should be relatively easy to test, remove the version for psutil completely in setup.py.
- Add a cross package, and along with that, remove `--system-site-packages` when creating the Python virtualenv [here](https://github.com/SynoCommunity/spksrc/blob/master/spk/octoprint/src/installer.sh#L31). This only works if Octoprint doesn't rely on other global Python packages.

I'd prefer the first, because we will end up with separate cross packages for different spk's, and that's just a hassle to maintain. 
The third option is changing the way cross-compiled wheels are created, and decouple that from cross packages. This branch is more or less what's needed in order to do that: https://github.com/SynoCommunity/spksrc/tree/pip_wheel

Dr-Bean:
Better late then never: Published Octoprint_20160321-4 on the repository.

Issue #974, opened 2014-05-07
[Request] Add MKVToolNix spk


Celeamas:
MKVToolNix shall be usefull for syno users who wants to merge srt to mkv through script.
##

Brontoloo:
You can simply use ffmpeg(which is already included in dsm):
to add sub.srt to movie.mp4 just use
"ffmpeg -i "movie.mp4" -i "sub.srt" -c copy -map 0 -map 1 "movie.mkv"

fn002449:
Ah!!! That is nice!! Thank you! I tried it and it works flawless.

Here is a code to add more than one subtitle to one mkv file
ffmpeg -i "old.movie.mp4" -i "dut.srt" -i "eng.srt" -c copy -map 0 -map 1 -map 2 -metadata:s:s:0 language=dut -metadata:s:s:1 language=eng "muxed.movie.mkv"

you can set -v (verbose level) and -y (if you want to overwrite the existing-file with output-file

Thx.

BenjV:
I tested this on my i7 PC with a 1080 .mkv file.
With mkvmerge it took 45 sec to merge the subtitle and vidfeo into one file.
With ffmpeg it took about 45 minutes to do the same.

I don't think it will work on my DS414, it will take forever to do this with ffmpeg.

BenjV:
It sure is faster but ffmpeg keep crashing with "unknown entry" while mkvmerge does it's job fine.
So still no luck.

Brontoloo:
@BenjV You simply forgot or misspelled the ffmpeg option "-c copy" (without quotes).
That option means that the audio, video and subtitles streams will not be encoded, but copied over the new file

BenjV:
@Brontoloo 
I thought that I dit a copy paste of the command from your posting but I tried it again today en now it was much faster.  So your probably right about that.
Not as fast as mkvmerge that did it in 29 sec and ffmeg in 55 sec but ffmpeg crashed when it was 95% finised.
So still no luck. Fast enough but crashing.

aqzina:
Hi, I would like to second this request. Can someone say anything about the latest status for this package? Can the boost compilation issue have a workaround?

M0UL:
any progress on this?

kratz00:
I gave it a shot. One problem is compiler support, from the official MKVToolNix [README.md](https://github.com/mbunkus/mkvtoolnix/blob/master/README.md#21-requirements)

[...]

PR #2178, opened 2016-03-07
add mkvtoolnix v8.5.2

cytec:
as requested: https://github.com/SynoCommunity/spksrc/issues/974

@Dr-Bean i've only tested with cedarview 5.1 and 5.2 toolchain atm... compiles fine but 2 things:
1. you'll have to set `LC_ALL=C` manually in terminal before executing any of the binaries or they crash with `terminate called after throwing an instance of 'std::runtime_error'
   what():  locale::facet::_S_create_c_locale name not valid
   Aborted (core dumped)`
2. I've put python in build_depends as the lib/bin files are not needed on the device but somehow it copys them over anyway... what's the correct way to remove unneeded stuff from the final package? ex: all bin tools except the `mkv...` ones?

EDIT: i've selected v8.5.2 because it's the last version which compiles with gcc < 4.8


Dr-Bean:
I would have suggested to setting LC_ALL + calling the binary in a small script, and put that on PATH, but presumably all four binaries would need that same treatment :/
Probably best to mention it  in a help page/wiki or even the package description.

As for Python's PLIST, `cross/boost` includes Python as DEPENDS: that way Python's PLIST would still be processed. Maybe see if changing boost to use BUILD_DEPENDS works: Deluge is the only other package that uses boost, and it shouldn't need to include Python either.

PR #2182, opened 2016-03-11
Updated Transmission to 2.92

GoodOmens83:
Updated Transmission to 2.92

In addition to standard version number changes:

1. Had to update the patch file (Is this still needed?) to reflect line changes
2. Remove transmission-cli from the PLIST file - See 2.92 change log: _"[Don't build transmission-cli by default (it's long deprecated)](https://trac.transmissionbt.com/wiki/Changes)"_

Test compiled on Avoton and qoriq and compiled fine.  Test installed on a DS1515+ and ran fine.

### DIFF
--- a/spk/transmission/Makefile
+++ b/spk/transmission/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = transmission
-SPK_VERS = 2.84
-SPK_REV = 11
+SPK_VERS = 2.92
+SPK_REV = 12
 SPK_ICON = src/transmission.png
 DSM_UI_DIR = app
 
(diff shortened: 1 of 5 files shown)


GoodOmens83:
Noticed the patch file was re-introducing a bug fixed sometime in the past.  We were patching in two "-Wcast-align" flags when there should have been only one.

Retested and everything is fine.  Should be good to go unless @Dr-Bean has some Makefile formatting suggestions.

spk/transmission/Makefile
```diff
@@ -13,7 +13,7 @@ DESCRIPTION_SPN = Transmission es un cliente BitTorrent simple y rápido. Puedes
 ADMIN_PORT = 9091
 RELOAD_UI = yes
 DISPLAY_NAME = Transmission
-CHANGELOG = "Enable sc-download group"
+CHANGELOG = "Update Transmission to 2.92"
```
Dr-Bean: Let's include the previous changelog as well: in this particular case, there were only a very select number of arches that received a package with it included (alpine and newer, or something like that):
```
CHANGELOG = "1. Update Transmission to 2.92<br>2. Enable sc-download group"
```

Dr-Bean:
Other than the one comment, LGTM. Thanks :)

GoodOmens83:
Got it!

PR #2186, opened 2016-03-12
Update rutorrent 2.7 / rtorrent 0.9.6 / libtorrent 0.13.6 / xmlrpc-c 1.39.07 

GoodOmens83:
Made a bunch of updates to get rtorrent to 0.9.6 and ruTorrent to 2.7 (Trying this PR again with everything broken down into separate commits to make it less confusing - I'm now getting the hang of using git - yay!)

As these packages all rely on each other (and nothing depends on them) I updated all the packages in one PR. 

Some notes:

- I  no longer use the ruTorrent plugin folder as the plugins are now bundled with the 3.7 zip so I've deleted it
- For xmlrpc I had to comment out two lines of the config in order for the AR and RANLIB files to pick up correctly. I am doing this via the patch file.  Everything else was minor Makefile/digest/patch line number edits

Everything seems to work wonderfully. I tried and successfully complied for the qoriq sync and avoton and it's running on my DS1515+.

Question:

I want to update some other depends but they rely on other packages.  

- Do you want me to create separate PR for those?  Example libxml2 is out of date.  I have them staged so can add to this PR if need be.  libsigc++ is also really old.
- If I do separate PRs how often do you re-compile packages?  I guess I'd have to bump the version of rutorrent to pick up the newer depends correct?  Does it matter??....


Dr-Bean:
Code looks good to me, so on that basis, good to go.

As for your questions: Usually, we can keep the dependencies up to date, but do take a good look at the 'parent' packages that use those dependencies: sometimes there are reasons to be seemingly out-of-date. On that note, libxml2 isn't outdated afaict, latest is still 2.9.3?

Generally speaking, we prefer PR's based around updating an SPK. PR's to update a cross package to latest 'just because it's outdated' have a good chance of not getting accepted. Exceptions to that are very generic packages, e.g. OpenSSL. 

As for separate PR's, packages are built and published by `SPK_REV`. If you create a new PR to update a dependency after I've merged this and published a new set of ruTorrent packages, the additional changes will not get included until ruTorrent's `SPK_REV` is bumped again.

If you want to continue discussing this, let's do that in a separate issue. It's an interesting subject, and it's something we might want to document somehow.

Dr-Bean:
Forgot to ask: did you test updating from the current package to this version?

GoodOmens83:
Yes - I'm running it now on my DS1515+ - I can test a downgrade and a upgrade if you like.

GoodOmens83:
Forgot to comment lol:

You are correct, latest libxml2 is 2.9.3.  

Also makes sense about packages, and that's exactly what I assumed (need to bump the REV to pick up any package changes).  Just more curious about if its something security related (e.g., openssl) but that is where my functional expertise ends (e.g., does a vulnerability translate to a package built on a compromised openssl and thus they would need a version bump to use the new package).  Was hesitant to open a separate issue for the problem but it seems warranted.

Dr-Bean:
@GoodOmens83 Did you test with building this on 64 bit or were you still on 32b? 
On 64b, I'm getting a number of errors on various arches: `AM_PATH_CPPUNIT` and unpatched cross-compile tests.

GoodOmens83:
32bit host. I can try on my 64 bit host and see what changes are needed.

GoodOmens83:
I was able to do make all-supported on my 32B build box.  

Make sure you `apt-get install libcppunit-dev` .  I also had to `cpan intall HTTP::Request` (I think it was HTTP::Request).  

64B seems to have issues with the compiler executable for a lot of arches (88f6281, alpine, armada370, evansport, monaco-, qoriq etc.) but I think that's because their gcc can't run on a 64B host.  I've tested those same arches on other packages and get the same issues.  Only armadaxp and x64 compiled on 64B (but maybe that was expected).

Dr-Bean:
Installing `libcppunit-dev` should be added to the README/dockerfile then. As for installing the Perl module, that makes no sense to me at all.

Wrt to the 64b host you're using, could you please verify that the i386 arch is added and `libc6-i386` is installed? See [this commit](https://github.com/SynoCommunity/spksrc/commit/c4b3097e8475c01989cedc0da301413ddea9ed0f).

GoodOmens83:
Ah crap - trying to squash commits I really screwed this one up lol.  I'm going to delete the branch and start a new pull.

PR #2187, opened 2016-03-13
Updated sabnzbd-testing to 1.0.0 RC4

mhartvig:


mhartvig:
Updated it to RC5 instead, #2188

PR #2189, opened 2016-03-13
Update tvheadend to 4.0.9

GoodOmens83:
Updated tvheadend to 4.0.8, the latest stable release.

Will work on developing tvheadend-testing package if I can find a source of 4.1 tar balls...

E.g., the latest is 4.1-1670~g5a94400 but all I can find are compiled debian packages.


GoodOmens83:
4.0.9 was released this morning.  Updated to reflect the new release.

m4tt075:
@GoodOmens83 
If you are still interested in developing a testing package and I'm not completely mistaken you can simply pull the most recent 4.1.x tarballs from
https://github.com/tvheadend/tvheadend/archive/master.tar.gz

GoodOmens83:
Yea I was running into compile issues with the recent tar balls and ffmpeg. I'll take another stab when I get more time.

cross/tvheadend/Makefile
```diff
@@ -1,7 +1,7 @@
 PKG_NAME = tvheadend
-PKG_VERS = 3.4patch1
+PKG_VERS = 4.0.9
 PKG_EXT = tar.gz
-PKG_DIST_NAME = $(PKG_VERS).$(PKG_EXT)
+PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
```
Dr-Bean: Add a `PKG_DIST_FILE` here (usually along the lines of `PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS)`. Redownload and recreate digests after that.

The result is that the file is stored in DISTRIB_DIR with a name that can be tied to a package. Same applies to the contents of the digests file.

spk/tvheadend/Makefile
```diff
@@ -12,7 +12,7 @@ DESCRIPTION_FRE = Tvheadend est un serveur de streaming pour linux gérant les s
 ADMIN_PORT = 9981
 RELOAD_UI = yes
 DISPLAY_NAME = Tvheadend
-CHANGELOG = "Update to version 3.4.27"
+CHANGELOG = "Update to version 4.0.8"
```
Dr-Bean: `4.0.9`

Dr-Bean:
Couple comments on the code. Have you verified that it runs as expected?

cross/tvheadend/Makefile
GoodOmens83: Is this because `PKG_DIST_NAME` is not `$(PKG_NAME)-$(PKG_VERS)`, e.g., it has the extra `v` in the file name??  Just curious.

GoodOmens83:
Honestly haven't tested because I don't have a tv tuner installed on my Synology.  Was just inspired to update based upon the recent need to be on a more current version.

cross/tvheadend/Makefile
Dr-Bean: More or less. You're right: `PKG_DIST_NAME` is usually `$(PKG_NAME)-$(PKG_VERS)`, but upstream doesn't always use that format for various reasons.
For spksrc, that format ensures the filename is uniquely connected to a cross package: no possibility for duplicates in DISTRIB_DIR for example, and it's easier to recognize too.

See [here](https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.cross-cc.mk#L9). Essentially, we're overriding `PKG_DIST_NAME` (the original/upstream filename) with `PKG_DIST_FILE` wherever it's used.

Dr-Bean:
Well, seeing as I can't test it either, we can do this:
- The package needs some other clean up (e.g. that .cgi.pl script should probably be removed, so we revert to a static shortcut as all other packages are using now)
- Verify that accesscontrol.json still works/regenerate a new version to make sure the contents are still up-to-date
- Add a beta tag, so we don't inadvertently break the current stable package (even if it's outdated, better to keep a version available that we know works)

GoodOmens83:
Something like that?

>Verify that accesscontrol.json still works/regenerate a new version to make sure the contents are still up-to-date
Not really sure what this does.  I've noticed more recent packages from another repository still utilize this file but couldn't find anything in the developers guides referencing this file.

Dr-Bean:
The accesscontrol is this: https://tvheadend.org/projects/tvheadend/wiki/Access_configuration, it's not a DSM feature.

spk/tvheadend/src/app/config
```diff
@@ -5,7 +5,8 @@
             "desc": "Tvheadend Web UI",
             "icon": "images/tvheadend-{0}.png",
             "type": "url",
-            "url": "3rdparty/tvheadend/tvheadend.cgi",
+	    "port": "9981",
```
Dr-Bean: Replace TAB's with spaces, and ensure text is aligned.

GoodOmens83:
Removed the tab - let me know if you want me to squash.

Dr-Bean:
You checked the accesscontrol.json? (for the record, I would think it's auto-generated if you remove it and restart the package, but I'm not sure)
Other than that, good to go. Squashing would be appreciated :)

GoodOmens83:
Squashed the commits but not quite sure about the json file.  Maybe someone who uses it could chime in about deleting the file and seeing if it's re-generated.

I've checked a synology 4.0.8 package from another source and it has the file in there with the same syntax as this one.

Dr-Bean:
Point is, if we add such a file to the package, we should make sure its up-to-date before issues with compatibility pop up. That may or may not happen, who knows.
So here's a suggestion: install the package as normal. Create a new user via the web interface. A new file will be added to `/usr/local/tvheadend/var/accesscontrol`, next to the original one.

Use that file to recreate the 'template' from (e.g. replace the value of `username` with `@username@`, replace `password2` key with `password`, and adjust the value there with `@password`, that should probably do)

Update the package with that file, and do a clean install of the package. Verify that access still works, and there you have it.

GoodOmens83:
Ok - all should be good.  At least it fires up on my DS1515+ :)

Dr-Bean:
Lost track of this one. Should be ready to go, right?

Dr-Bean:
Nope, compile fails for non-x86_64 due to the following:
```
/work-88f6281-5.2/tvheadend-4.0.9/libhdhomerun_static/libhdhomerun/libhdhomerun.a: could not read symbols: Archive has no index; run ranlib to add one
collect2: ld returned 1 exit status
```
` --disable-hdhomerun_static` works, but that has it's own downsides. Presumably can be worked around by putting hdhomerun in its own cross package and adding it as a dependency.

@GoodOmens83, care to take a look?

GoodOmens83:
Strange - I thought it was compiling fine a few weeks ago for all supported arches.  I'll take a look over the weekend.

Dr-Bean:
@GoodOmens83 Just checking if you were able to confirm the issues I had?

GoodOmens83:
@Dr-Bean: Was able to confirm - have had zero chances to work on anything.  Pulling 11 hour days at work so have had little motivation to work on anything....

beralt:
@Dr-Bean Building on non-x86_64 fails because the HDHomerun libraries are compiled statically, and the that specific Makefile uses the CROSS_COMPILE environment variable to determine the path to the compiler, as can be seen [here](https://github.com/Silicondust/libhdhomerun/blob/master/Makefile#L14). Something like this might fix it:
```
ENV += CROSS_COMPILE=$(WORK_DIR)/$(TC_BASE_DIR)/bin/$(TC_PREFIX)-
```

Note that there is [another repo](https://github.com/jdierkse/spksrc) which contains a somewhat working version of the 4.x version of TVHeadend. The package has issues with permissions and it calls `adduser` in its installation script, causing issues on DSM 6.

taleteller:
Just stumbled over this searching tvheadend 4.1 for a DS1812. When is this pull to be expected available for testing?

KaraokeStu:
@taleteller It all depends on when @GoodOmens83 manages to get the x86_64 part compiling, sorry! :)

Skindred:
Why the close?
Just curious :-)

GoodOmens83:
I don't have time to work on this. Feel free to copy and work on it!

m4tt075:
@GoodOmens83: I believe you were almost there! Please see the above pull request and consider re-opening yours again. Just cherrypick what you need. Will happily close my PR again as soon as you do that. This is your work really!

PR #2192, opened 2016-03-15
Update Node.js to v4.4.2 (LTS)

fhemberger:
Current version 0.12.2 of Node.js lacks several important security fixes, so I updated it to the latest LTS (Long Term Support) release.

I ran the build in the docker container (node worked for me), but I'm getting build errors for the unrelated 'ffmpeg' package.


piwi82:
There is an official Node.js package built by Synology : https://www.synology.com/fr-fr/dsm/app_packages/Node_js
I think it would be better to ask Synology to update this one.

Also, take a look at https://github.com/SynoCommunity/spksrc/issues/1872#issuecomment-148114864

fhemberger:
@piwi82 I'm reaching out to them as well, but as you are also providing the package, here I am. ;)
If gcc 4.8 is a problem, I'll add a second PR to at least use the latest 0.12.x, which contains several important security fixes regarding OpenSSL.

fhemberger:
Added #2196 for latest v.0.12.x

fhemberger:
With the release of DSM6, this should work now. Synology itself is offering the outdated version 4.2.6.

fhemberger:
Update:
- **Node.js 4.4.2 (Security release)**
- Required DSM version set to 6 (as official DSM6 builds contain Node.js v4 as well)
- Use new official Node.js logo

Dr-Bean:
@fhemberger It was already partly discussed in earlier comments above, but I think we're going to retire our Nodejs package and start removing the code from spksrc.
See also https://github.com/SynoCommunity/spksrc/issues/1872#issuecomment-204290721.

If Synology's packages are outdated, it's probably best to request them to update the packages (I believe Synology expects a feature request for those: https://www.synology.com/en-global/form/inquiry/feedback).

This also applies to #2196, but thanks for taking the effort to set up the pull requests.

Dr-Bean:
Closing. I think my previous comment and #1872 explain it all. Thanks again.

PR #2194, opened 2016-03-15
Updated sabnzbd-testing to version 1.0.0

mhartvig:
Build-ed with all-supported.
Test upgraded from current version in channel to this without issues

Should beta be set to 0 or 1, since this is the beta channel?

### DIFF
--- a/cross/sabnzbd-testing/Makefile
+++ b/cross/sabnzbd-testing/Makefile
@@ -1,8 +1,8 @@
 PKG_NAME = SABnzbd
-PKG_VERS = 0.8.0Beta5
+PKG_VERS = 1.0.0
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS)-src.$(PKG_EXT)
-PKG_DIST_SITE = http://downloads.sourceforge.net/project/sabnzbdplus/sabnzbdplus-beta/$(PKG_VERS)
+PKG_DIST_SITE = https://github.com/sabnzbd/sabnzbd/releases/download/$(PKG_VERS)
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 DEPENDS =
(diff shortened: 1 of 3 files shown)


Dr-Bean:
All `-testing` packages should set `BETA = 1`, those packages will never move to stable.
 Other than that, LGTM.

mhartvig:
Changed the makefile to have beta = 1.

Dr-Bean:
Thanks for confirming that updating works, should be good to go.

Safihre:
I don't have strong understanding how these packages are installed, but it seems to also be possible to compile stuff on installation?
Because CPU load and thus performance can be greatly increased if the C-version of yEnc decoder is installed.
There is a Python package that provides this on pip: https://pypi.python.org/pypi/yenc
But it needs to be (pre)compiled and then placed in the SABnzbd folder so it can be used.
Would this be possible?

mhartvig:
I too isn't to familiar with how it's done, but it seems like it is possible to install yEnc. It's already located in https://github.com/SynoCommunity/spksrc/tree/master/cross/yenc. What folder does it have to be located in @Safihre?

Safihre:
Aah interesting, seems somebody already took care of this!
You can see if SABnzbd can use it (after installing that package) by going into the Config in 1.0.0, on the first page you see yEnc and it should have a checkmark next to it.

cytec:
i think inlcuding yEnc should be possible by adding it to depends

mhartvig:
Without doing anything:
yenc
It might have been installed by some other package.

Safihre:
Hahah, you just beat me to the punch ;-) 
Indeed already installed, that's great so this is looking good!

cytec:
@SynoCommunity/developers i'm currently building those packages for `all-supported`

cytec:
packages are up @Dr-Bean can you compile one for `armada38x` seems like i need to migrate to a 64bit machine to build with that toolchain :/

Dr-Bean:
Done. Probably a good idea to remove older versions from the repo (keep `n` and `n-1`, remove the rest)

cytec:
yeah... i've disabled everything below `0.8.0Beta5` for now.

PR #2195, opened 2016-03-15
SAbnzbd update to 1.0.0

lysin:


Dr-Bean:
Thanks for making the effort. A couple comments for now: 
- The CHANGELOG entry in the spk's Makefile needs to be set;
- The primary download site moved to Github, prob best to change that;
- I'd be surprised if the configuration file wouldn't need updating;
- The queue cannot be migrated from 7.20 to 1.0.0 (see Sab's changelog), so we need to do something there. Maybe we could use an upgrade wizard for that? An alternative would be putting it in the CHANGELOG, but I don't know if that's enough exposure.

Safihre:
If possible, could you switch the download location to Github? (just like in the other pull request I saw about the update) 
We are trying to move away from Sourceforge!
 
(I'm dev for sabnzbd)

lysin:
@Dr-Bean 

Changed some stuff in order to comply with your requests.

I tested with my current setup (0.7.20 >>> 1.0.0) and also did a clean install. Everything is working fine.

It's my first attempt at upgrading so I'm still learning. 

Cheers

spk/sabnzbd/Makefile
```diff
@@ -14,7 +14,7 @@ DESCRIPTION_SPN = SABnzbd hace que Usenet sea lo más simple posible, automatiza
 ADMIN_PORT = 8080
 RELOAD_UI = yes
 DISPLAY_NAME = SABnzbd
-CHANGELOG = "Enable sc-download group"
+CHANGELOG = "Update to 1.0.0. Note - When you upgrade an existing 0.7.x installation, release 1.0.0 will NOT be able to finish the queue. It's best to empty your queue first using 0.7.x. Alternatively, you can install 1.0.0 and then use queue repair on the Status page. Some information (like the order of the queue) will be lost."
```
Dr-Bean: Use this format: `"1. Update to 1.0.0<br>2. xyz"`
The `<br>` adds a line break when shown in Package Center and shows that there are two distinct entries in the log.

spk/sabnzbd/src/wizard/install_uifile
```diff
@@ -16,4 +16,9 @@
             }
         }]
     }]
+}, {
```
Dr-Bean: You can remove this addition. The install wizard will only show up if it's a clean install of the package (i.e. no upgrade). In other words, it doesn't do anything for us ;)

Dr-Bean:
Very nice. Two comments, good to go after those are resolved :) 
Please consider squashing your commits and force-pushing the result to your remote branch.

lysin:
Ok, I thing I'm starting to understand how this works. :sweat_smile: 

@Dr-Bean Can you please check to see if everything is as expected?

Cheers

Dr-Bean:
If you could 'fix' the commit message to reflect (only) the actual changes, that would be nice. Squashing by default just bundles up all the previous commit messages, but it doesn't look great.

lysin:
Ok, I honestly can't understand how to make git work and I'm giving up. 

I've been messing around with it a lot and it's time someone more experienced takes over.

I actually opened a new pull request, but I don't know how I can delete either one of them.

I'm sorry for wasting everyones time with this noobiness.

Dr-Bean:
Nothing wrong with the other PR though :P Commits and commit messages look fine.

The only thing that I would comment on is that the author of the commit isn't recognized by Github, so you'd need to do this before creating the commit: https://help.github.com/articles/setting-your-email-in-git (so something like: set the user/mail in gitconfig first, then create the commit and finally force-push that to remote to overwrite the current commit so it shows up in the PR).

Anyway, thanks for taking the time, much appreciated.

Issue #1467, opened 2015-02-03
rutorrent create torrent error


mellowpimp:
Hello,

The create torrent function in rutorrent does not work for me. i can enter all settings, but when i press create it will try creating the torrent file but after a few seconds it will say " error." and thats it.

i can't find anything in any logfiles. all permissions on directories seem to be fine as far as i can see. 
i have seem more people with this issue, but im not sure what is causing it. didnt no where else to post so seeking here for advice. 

thanks.

Oleekae:
same situation friend. Any clues would be welcome

mal68:
I have the same issue, have spent some time learning how to make sure the permissions are correct and everything seems to be in order. But it just simply waits a couple of seconds and then says "error"

Dr-Bean:
rutorrent 3.7-8 has been released, and I can't reproduce it with that. I'm closing this one, feel free to open a new issue if you still encounter the problem.

PR #2200, opened 2016-03-19
Update ruTorrent 2.7 / rtorrent 0.9.6 / libtorrent 0.13.6 / xmlrpc-c 1.39.07

GoodOmens83:
Ok Third time should be the charm!  Sorry for all the PR's - still learning new things about GIT.

Some notes:

- I no longer use the ruTorrent plugin folder as the plugins are now bundled with the 3.7 zip so I've deleted it
- For xmlrpc I had to comment out two lines of the config in order for the AR and RANLIB files to pick up correctly. I am doing this via the patch file. Everything else was minor Makefile/digest/patch line number edits
- rtorrent requires libcppunit-dev so I've updated the read-me to reflect this

Everything seems to work wonderfully. I successfully `completed make all-supported` on a 64B host.  @Dr-Bean - the `cpan install HTTP::Request` was not needed as it was already installed on my brand new Debian 64b host (seems to be a default perl module).  No idea why on my 32B host it was missing it - openssl uses it to grab a certificate for certain arches.  Things are running stable on my DS1515+.  

The only thing I can see this needing is the spk/rutorrent/Makefile updated to reflect the non-supported ARCHs.  It wasn't in the original so I left that aspect as is.


Dr-Bean:
First two commits are done with a different username/unassociated with Github: if you want to change that, go ahead.

What do you mean with non-supported ARCHs though? Are there archs that don't compile (mainly 5.2, we care less about earlier toolchains)? Especially since you mentioned that `all-supported` completed succesfully.

GoodOmens83:
@Dr-Bean :

> First two commits are done with a different username/unassociated with Github: if you want to change that, go ahead. 
- Fixed.  DIdn't know I could do that

>What do you mean by non-supported ARCHs though? Are there archs that don't compile (mainly 5.2, we care less about earlier toolchains)? Especially since you mentioned that all-supported completed successfully.

- If you look at rtorrent it lists `UNSUPPORTED_ARCHS = 88f5281 powerpc ppc824x ppc853x ppc854x`  Have not tried removing those flags to see if it compiles.  I can try and report back if you like?  I'm just not sure if the spk/<package/Makefile inherits all unsupported flags from its depends.  There is no UNSUPPORTED_ARCHS in /spk/rutorrent/Makefile

Dr-Bean:
Ok. The build process will recognize and skip any arch listed in `UNSUPPORTED_ARCHS` when that's defined in either the SPK's Makefile or any of its dependencies.

Except for `ppc853x`, none of the arches listed are part of `all-supported`. The reason for removing those arches in the first place was mentioned here: https://github.com/SynoCommunity/spksrc/commit/fdd7c391127ea087d5a955203d549b42ee9a100d. That means that `ppc853-5.2` should work...and that would make sense, seeing as ppc853x is near identical to qoriq.

GoodOmens83:
Good to know - let me test compile those arch's as it seems they where excluded because rtorrent needs GCC > 4.2.

I think everything is >=4.3 now so in theory they should compile (not sure about run). I'll report back.

GoodOmens83:
Compiled fine for ppc853x - unless you want me to incorporate changes for the cacert - I think this is ready to go.

Dr-Bean:
Let's do the cacert separately, outside of this PR anyway. I might take a stab at it between merging your PR, and before I release new packages.

GoodOmens83:
Sounds good.  I'll leave this branch untouched then!

gornostal:
@Dr-Bean is everything merged to master gets automatically built and becomes instantly available in Synology Package Manager?

Dr-Bean:
It's building now, although it'll take a fair amount of time before its done with that. 
Note that this package isn't compatible with DSM6. It probably won't even work when migrating from DSM5 with the package installed, because of the changes to Webstation and the `http` user.

gornostal:
Thanks for the clarifications.

Dr-Bean:
rutorrent 3.7-8 has been published on the repository. Thanks @GoodOmens83 :)

GoodOmens83:
NP - I'm jealous of the changes you made post merge.  I might have to make my own package and install over - too afraid to uninstall/reinstall as I'm on DSM6 now....

Dr-Bean:
Just the unzip thing really, so that's not too bad. I wouldn't do a clean install of the package if you have it working right now. I've tried it on DSM6, but several things seem to be broken wrt WebStation and user permissions, I don't even know if it works as intended.

You can install it manually (you probably want to make a backup of the appdir before upgrading to be sure). Simplest way: edit `SPK_REV` in `/var/packages/rutorrent/INFO`, and an update should show up via Package Center...or recompile it from scratch.

Dr-Bean:
Wrt DSM6, linking to https://github.com/SynoCommunity/spksrc/issues/2215. The errors mentioned are similar to what I've seen during a quick test.

GoodOmens83:
Strange - I haven't noticed any permissions issues yet.  I might have modified /etc/php/php.ini to remove the open_basedir line though

mal68:
Hey guys, I'm currently running the 3.6 package and I just received a notification that saying that a updated package has been released. This is good as it appears it may have fixed this issue I was having last year https://github.com/SynoCommunity/spksrc/issues/1467#issuecomment-207036908

I am by no means an advanced user, I have only done limited linux stuff in the last year so have forgotten most of it. Anyway my plan is to upgrade to DSM 6 in a week or so and I want to continue using rutorrent but obviously the new package would be better,

Does anyone know yet what the process is going from DSM 5.2 & rutorrent 3.6-7 to DSM 6 and rutorrent 3.7-8?

I am going to assume it's not as simple as updating the app and then updating the NAS. But if someone has done it or is planning to do it shortly, I would certainly appreciate a basic guide on how to do this.

Thanks

GoodOmens83:
@Dr-Bean @Diaoul  : I'm working on a larger fix for this utilizing lighttpd instead of relying on Webstation.

In my limited testing it seems to alleviate a lot of issues people are having.  I will wait for a official PR until we figure out how to create users.

I'm creating lighttpd as a separate package in case anyone wants to use it for some things.  Somewhat to learn more about creating packages etc.  It compiles a lot easier then I thought it would have.

Some questions:

1)  Would you even except something like using lighttpd/php as custom compiles?
2) Should I make PHP a separate spk module as well?  Right now I have it as a cross package within lighttpd.  Just thinking if PHP versions bump would be easier for users to update and would mirror what Synology is doing
3) I assume the user should be lighttpd (as instead of utilizing http)
4) For PHP - what modules should I include?  I'm keeping it simple at the moment - just the basics for it to run ruTorrent but also included MySQL in case anyone wants to use it.  `--with-mysql --with-mysqli --enable-json --enable-cgi --enable-simplexml --enable-libxml --disable-mbregex --enable-sockets --enable-fpm --disable-opcache --without-sqlite3 --without-pdo-sqlite --without-pear`

Zizit25:
Hi guys @GoodOmens83 @Dr-Bean 
I'm pretty sure you have everything to work on that issue, but as I'm part of the user which Rutorrent don't work (permissions issue mostly or package don't start), if you need help to test something don't hesitate.
DS415play with DSM6 (fresh installation)

PR #2201, opened 2016-03-21
SABnzbd update to 1.0.0

lysin:
Updated SABnzbd to 1.0.0
Updated digests for SABnzbd 1.0.0
Created upgrade warning
Changed provider to Github

### DIFF
--- a/cross/sabnzbd/Makefile
+++ b/cross/sabnzbd/Makefile
@@ -1,8 +1,8 @@
 PKG_NAME = SABnzbd
-PKG_VERS = 0.7.20
+PKG_VERS = 1.0.0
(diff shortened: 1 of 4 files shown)


Dr-Bean:
Unless I'm missing something obvious, LGTM.

Popsikill:
Does this take a while to flow through to the package center or something? Wondering if I'm not getting the update due to it not being out yet, or its something to do with my recent DSM6 update...

Dr-Bean:
It's not out yet, the reason being some more tests were needed. On that note, if you're doing a fresh install on DSM6, it would be broken anyway...which is one of the things that needed testing.

Issue #2202, opened 2016-03-23
GIT 2.7.4-7 and SickRage update


flyescke:
Hi,

Since I've updated the GIT package to 2.7.4-7 I'm getting errors in SickRage when I try to update the application. Could this be related to the new version or should I log this with SickRage?
I have a Synology DS413 btw.

git: xdiff/xutils.c:358: xdl_hash_record: Assertion `*ptr == '\n'' failed.
error: merge died of signal 6, treat as error for now

Kind regards,

Kevin

GoodOmens83:
Seems that now the build environment is x64 the flag XDL_FAST_HASH in 
https://github.com/git/git/blob/master/config.mak.uname is picking up for everything (because git checks the build environment to use that feature and not the host environment).

That option does not work for QoriQ (or other possible powerpc environments):

http://lists.openembedded.org/pipermail/openembedded-devel/2014-September/098024.html

Patch:

 https://dev.openwrt.org/browser/packages/net/git/patches/200-disable_fasthash.patch?rev=37239

@Dr-Bean : Another option might be to just patch in the host variable instead of uname_M so this will be set properly for x86_64 hosts.  Not really sure how to do this (unless you can apply patches for only certain arches).

Dr-Bean:
How nice. Let's patch it out for now. We could presumably see if replacing `$(uname_M)`  with `$(TC_TARGET)` works, but otoh we've managed without the functionality so far...

GoodOmens83:
Not around to test but do you think replacing `ifeq ($(uname_M),x86_64)` with `ifeq ($(findstring $(ARCH),$(x64_ARCHES)),$(ARCH))` work?  Or would I also need to include `
include ../../mk/spksrc.cross-cc.mk` . 

This is the only place I can find `$(uname_M)` in the git source.

Dr-Bean:
Well, let's translate `uname_M` and `x86_64` to variables spksrc provides. How about this:
Instead of using `uname`, use `$(TC_TARGET)`. Instead of `x86_64`, we can compare against the equivalent value that `TC_TARGET` sets for x86_64 arches: `x86_64-pc-linux-gnu`.
Easy, and probably the most stable solution we can implement.

GoodOmens83:
Hmm I've tried playing around for a while.  `$(TC_TARGET)` is blank in `config.mak.uname`.  `$(ARCH)` produces a value but `$(x64_ARCHES)` does not.  

[...]

PR #2203, opened 2016-03-24
Fix bug with certain arches (e.g., qoriq) and xdl_fast_hash

GoodOmens83:
Quick Fix for https://github.com/SynoCommunity/spksrc/issues/2202

Still not ideal as we are losing functionality for X64 hosts as I'm just patching this out for all build types.

My limitation of Make and configs show.  I tried calling TC_TARGET (to try and and at least do the if statement for X64 arches) but it was a blank variable in git/config.mak.uname (the culprit).  ARCH works but $(x64_ARCHES) was not returning anything.  

I'm open for ideas - I'm trying to replace `ifeq ($(uname_M),x86_64)` (which is always true because uname_M is based on the build environment) with something that will only be true for x64 host builds (e.g., ARCH of type x64, avoton etc.)

### DIFF
--- a/cross/git/patches/git-remove-xdl_fast_hash-check.patch
+++ b/cross/git/patches/git-remove-xdl_fast_hash-check.patch
@@ -1,0 +1,12 @@
+--- config.mak.uname.orig	2016-03-22 11:13:28.842072683 -0400
++++ config.mak.uname	2016-03-22 11:12:15.698069794 -0400
+@@ -17,9 +17,6 @@
+ # because maintaining the nesting to match is a pain.  If
+ # we had "elif" things would have been much nicer...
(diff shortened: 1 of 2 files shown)


GoodOmens83:
@Dr-Bean : Adding ENV += worked.  This flag sets for x64 hosts only now.

Dr-Bean:
I am just about to publish the packages. Let's pick it up with the next release of Git.

GoodOmens83:
I reversed the last commit so you can merge.

Dr-Bean:
Much appreciated :)

Dr-Bean:
Done, and thanks a lot for looking into this :)

PR #2230, opened 2016-04-03
Use sudo to start Transmission on DSM 6.0

jeromegamez:
I stole the if condition from the [spk/transmission/src/installer.sh#L97](https://github.com/SynoCommunity/spksrc/blob/master/spk/transmission/src/installer.sh#L97) :)

:octocat:

### DIFF
--- a/spk/transmission/Makefile
+++ b/spk/transmission/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = transmission
 SPK_VERS = 2.92
-SPK_REV = 12
+SPK_REV = 13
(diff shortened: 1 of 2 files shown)


Dr-Bean:
This probably won't be merged. Using `su` or `sudo` isn't the issue here, it's the way users are created that causs the bulk of the issues.

Besides that, this doesn't scale: there are several packages that (still) won't work with this approach, and we're not going to implement fixes on a package-by-package basis.

jeromegamez:
@Dr-Bean Thanks for the heads up, I'll then close this PR!

Issue #2216, opened 2016-03-27
DSM6: adduser command not working [$100]


Dr-Bean:
Turns out that the reason users aren't created on DSM6 is due to the fact that `adduser` can't be found. It seems that the correct PATH isn't set, so the file(s) can't be found.

We can solve this in a number of ways:
- Set PATH before calling `adduser` and similar commands. E.g. `PATH=${PATH} adduser [..]`. Yeah...probably not ;)
- Untested, but I see no reason why it wouldn't work: Export PATH in `postinst` and `preuninst` (some kind of generic `setpath` call maybe? Could tie in with some type of `common.sh` approach)
- Use the full path to the binaries, e.g. set `ADDUSER = "${INSTALL_DIR}/bin/adduser"`. 
  Somewhat of a pain because we have multiple packages that don't bundle their own busybox, but use e.g. Python's (which keeps those packages `noarch`), but otherwise it should work.

This applies to `adduser`, `deluser`, `delgroup`. There may be other commands that I didn't think of.

@SynoCommunity/developers?
##

Dr-Bean:
Annoyingly, this issue disappears after you run `adduser` once successfully as root. Fix it once setting PATH (I did it via a edited package installer), and every subsequent package installation works fine, without any changes.

I don't know what's up with this. It almost looks like something got corrupted during the migration. Some package users have vanished without a trace. `start-stop-daemon` fails repeatedly, but similar to `adduser`: after you get it running once, all issues with it are gone and packages start working again.

GoodOmens83:
I was looking at the "privilege" file (<root folder>/conf) for the DSM6 Download Station package.  Not sure if this was how the DSM5.x package was but I see things like:

> {
> 	"defaults":{
> 		"run-as": "package"
> 	},
> 	"username": "DownloadStation",
> 	"groupname": "DownloadStation",

See full file: https://gist.github.com/GoodOmens83/2d465ef7adbfce045820

The resource file (in the same folder) seems to have a section for creating a database:

> 	"pgsql-db": {
> 		"create-user": true,
> 		"create-database": [{
> 			"db-name": "download",
> 			"db-owner": "package",
> 			"db-encoding": "SQL_ASCII",
> 			"db-collision": "skip",
> 			"db-enable-lang-c": true
> 		}]
> 	},

Dr-Bean:
More info: 
[...]

PR #2241, opened 2016-04-07
couchpotatoserver-custom : Update dsm-control.sh

simenon:
Fixes startup script to use sudo instead of su for DSM 6.0 and higher.

### DIFF
--- a/spk/couchpotatoserver-custom/src/dsm-control.sh
+++ b/spk/couchpotatoserver-custom/src/dsm-control.sh
@@ -15,11 +15,10 @@ COUCHPOTATOSERVER="${INSTALL_DIR}/var/CouchPotatoServer/CouchPotato.py"
 CFG_FILE="${INSTALL_DIR}/var/settings.conf"
 PID_FILE="${INSTALL_DIR}/var/couchpotatoserver-custom.pid"
 LOG_FILE="${INSTALL_DIR}/var/logs/CouchPotato.log"
-DSM_MAJOR_VERSION=`cat /etc.defaults/VERSION | grep majorversion | grep -o [0-9]*`
 
 start_daemon ()
 {
-	if [ ${DSM_MAJOR_VERSION} -gt 5 ]; then
+        if [ `/bin/get_key_value /etc.defaults/VERSION majorversion` -ge 6 ]; then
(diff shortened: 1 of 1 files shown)


Dr-Bean:
See https://github.com/SynoCommunity/spksrc/pull/2230#issuecomment-205181130.

Fore future reference, please add the package name in the title and add a description that fits the changes.

Dr-Bean:
I meant the actual comment :)

>This probably won't be merged. Using su or sudo isn't the issue here, it's the way users are created that causs the bulk of the issues.

>Besides that, this doesn't scale: there are several packages that (still) won't work with this approach, and we're not going to implement fixes on a package-by-package basis.

simenon:
I have read the actual comment. And i can agree that it does not scale. However if you want to run the process under a user, you'll have to use su (or sudo in newer DSM 6+). 

I can not think of a better solution to solve the problem then just fixing the start-stop script for couchpotato. As the rest of the permissions and user creation are in my opinion correct.

At this point this is the only working solution to get couchpotato to run under DSM6. 

If u have a better solution to tackle it, i'll gladly try to help out.

Dr-Bean:
https://github.com/SynoCommunity/spksrc/issues/2216.

simenon:
@Dr-Bean #2216 relates to creating users, and not running a process under a certain users. This fix #2241 implies running a proces under a certain user (The change from su DSM5 to sudo DSM6)

Dr-Bean:
The point is that if users are created in a way that DSM6 supports, there's no reason to switch from `su` to `sudo`. This PR is not a fix, but a workaround that only applies to certain conditions.

simenon:
Given that if you can create a valid user, then there is no problem, and all works fine. Verified

Personally i prefer to have no login shell for these kind of pkg users, and as such i kinda have to alter the su command to include a shell, or just run sudo. Personally I prefer to use sudo instead of su, And since we have sudo nowadays, we can use it ...

Closing this.

Dr-Bean:
As the other issue shows, DSM6 provides a way to create system users without a valid shell (that can subsequently be passed on via `su -s`, or whatever the switch is to pass on a shell), so that should cover your preference at least in part.
The current method via busybox was always a workaround, seeing as Synology didn't provide a viable way to create a system user with a valid shell. Feel free to contribute to the other issue if you can/want to help out.

PR #2242, opened 2016-04-07
couchpotatoserver : Update dsm-control.sh

simenon:
Fixes startup script to use sudo instead of su for DSM 6.0 and higher


cytec:
see https://github.com/SynoCommunity/spksrc/pull/2241#issuecomment-207396263

PR #2270, opened 2016-04-18
Add lsscsi package

biyiklioglu:
Add lsscsi 0.28, a command to list information about SCSI devices.


hgy59:
Implemented in synocli-disk

PR #2271, opened 2016-04-19
gogs v0.9.13

alexandregz:
Gogs (https://github.com/gogits/gogs/) is a "painless self-hosted Git service"


cytec:
gernerally we'd prefer to use the source code to build our own version of gogs i think the only SPK which used bin's by default was btsync as it isn't open source... so you should probably rework this to use the source and build your own version of gogs on runtime ;)

Besides from that i left some other comments

cross/gogs/Makefile
```diff
@@ -0,0 +1,42 @@
+PKG_NAME = gogs
+PKG_VERS = 0.9.13
+PKG_EXT = zip
+PKG_DIST_SITE = https://dl.gogs.io
+PKG_DIR = $(PKG_NAME)
+
+DEPENDS =
```
cytec: should depend on `native/go` and build from source (no binary downloading/unpacking if we have source available)

cross/gogs/PLIST
```diff
@@ -0,0 +1,1077 @@
+gogs/LICENSE
```
cytec: use prefixes (see https://github.com/SynoCommunity/spksrc/wiki/PLIST-files) and delete stuff not needed (Readme, licence etc)

spk/gogs/Makefile
```diff
@@ -0,0 +1,47 @@
...
+SPK_VERS = 0.9.13
+SPK_REV = 1
+SPK_ICON = src/gogs.png
+DSM_UI_DIR = app
+
+DEPENDS  = cross/$(SPK_NAME)
+
+MAINTAINER = Alexandre Espinosa Menor
+DESCRIPTION = A painless self-hosted Git service.
+ADMIN_URL = admin
```
cytec: i don't think this is needed anymore

spk/gogs/src/dsm-control.sh
```diff
@@ -0,0 +1,99 @@
...
+DIR_GOGS="${INSTALL_DIR}/gogs"
+GOGS="${DIR_GOGS}/gogs"
+PID_FILE="/var/run/gogs.pid"
+LOG_FILE="/var/log/gogs.log"
+
+FILE_CREATE_LOG="${DIR_GOGS}/wizard_create_log"
+
+export HOME=${DIR_GOGS}
+#export PATH=$PATH:~/opt/bin  # to Git. Not necessary with Git Server (Synology)
+export USER=root
```
cytec: running as root is never a good idea...
take a look at some other tools... create a user on install and use that one...

cytec:
maybe @Dr-Bean should have a look at this one... i found some other stuff too but i'm not really sure if and what is needed in which way :D

alexandregz:
thx for the infor, I'm trying to modify with your suggestions :-)

torrentalle:
A new version with source build is published in PR https://github.com/SynoCommunity/spksrc/pull/2533

alexandregz:
See #2533 :-)

PR #2286, opened 2016-04-28
add 6.0 toolchain. still beta for now

spockfish:
Only for armadaxp, as that is the device i have.


cytec:
we don't use BETA toolchains

spockfish:
why not? this means you're users need to wait until Synology is willing to release the final version. so they're stuck. Releasing packages on a beta toolchain from Synology can help those users and an update can be pushed out when Synology releases the final toolchain.

cytec:
yep. but we never used beta toolchains... everyone is free to fork and set up beta toolchains if they want to but i don't think we will merge that. \cc @Dr-Bean @Diaoul

spockfish:
Agreed, that's how it works of course. And i did that to fix the packages that I personally use. But wouldn't it be better if, in stead of people 'forking around', this project get's more contributors? 

And with respect to the beta toolchain: making the beta toolchain available to contributors is something totally different than actually putting packages alive. At least it would provided everything necessary for package contributors to test their packages and prepare for DSM 6 (somewhat). 

Anyways, just my 2 cts.

GoodOmens83:
>making the beta toolchain available to contributors is something totally different than actually putting packages alive. At least it would provided everything necessary for package contributors to test their packages and prepare for DSM 6 (somewhat). 

I would agree having a beta/master branch could make some sense.  Approved merges go to the beta branch for more testing before being pushed to master for building.

cytec:
We had that a while ago... a development branch... which was actually never used as such but instead became the 'master' somehow over time because no one merged stuff back to master branch in the first place (if i remember correctly)

spockfish:
for my idea: are packages directly published from the master branch? So triggering a rebuild in master results in a deploy?

cytec:
@spockfish seems like the Toolchains for DSM 6 are out: https://github.com/SynoCommunity/spksrc/issues/2296

Dr-Bean:
A couple of comments to start off:
- This PR contains several unrelated changes. Those should be split up into separate branches. For each branch, open one PR. If need be, squash your commits, so we have a clean(er) history.
- Please set the default TC back to 5.2: we can't switch over to 6.0 properly if the kernels are not included. You can still compile packages with a different TC using `make arch-{arch}-6.0` (check the wiki on build/compile rules for other rules)
- Please change the title to accurately reflect the contents of the PR.

As for some comments on the earlier conversation, although this probably fits better in its own issue:
Adding the toolchains one-by-one isn't very useful from an spskrc perspective. When opening a PR to add toolchains, please add all of them. You did that already, so this is just a FYI ;)

As far as I'm concerned, the reason that we didn't add the beta toolchains when the beta was running was because of a lack of timely contributions. I myself have had very limited time lately, and that will continue for a while yet. If the additions are made during the beta period, along with the kernels, that's fine with me. We could set up a separate branch, but that should only be necessary if we have/expect backwards compatibility issues.

Adding in the beta toolchains after the DSM final is already released is a different matter: the final toolchains usually don't take too long to be released, so we might as well skip the beta stuff at that point. More importantly, we don't know that the beta toolchains are fully compatible with the released DSM version (I believe there actually was an issue with the Braswell arch for example).
There are a very limited number of packages that _need_ newer toolchains to run on a newer DSM version, and that usually only applies when the kernel versions are updated. For DSM6, Mono, and maybe ZNC and Syncthing come to mind for a subset of arches.
With the amount of open issues we have, you shouldn't have to run out of things to contribute either...;)

Dr-Bean:
Even though this PR was opened earlier, #2306 seems to be cleaner and more in line with what I'd expect.

Dr-Bean:
Closing in favor of #2306, since that one is ready to go (once we get to merging it anyway). 
@spockfish Even though we're not using your PR, thanks for taking the time to create this PR.

PR #2303, opened 2016-05-05
get dotconf archive directly from github

spockfish:


Dr-Bean:
cross/dotconf isn't actually used for any package currently (it is a dependency for another cross package which also isn't used iirc), hence why it hasn't been updated in a while...we might even want to consider removing it.
How did you test this works, just tried compiling?

spockfish:
I updated it because I ran a 'make prepare' which in turn starts pulling all those packages. As the original URL was gone I just changed the URL to the right place.

If the package is not used than by all means reject this PR.

Dr-Bean:
I have removed dotconf and speechd via d268abd6e2d3790dcf35fa6f1b2ccd7ebbd64402.
Thanks for the trigger ;)

PR #2314, opened 2016-05-12
Call -j parameter when compiling

GuillaumeSmaha:
Permit to do parallel compilation


Dr-Bean:
I don't think we'll merge this. For testing/troubleshooting purposes, it makes things a lot harder when the arg is passed by default.
You can pass on the `-j` arg when compiling, like this: `make -j4 arch-xyz`, that should work just fine.

GuillaumeSmaha:
With `make -j`, I encountered some problems during configure call. That why I use `make -j` only on the compile call. If I retrieve this error, I will report you.
Until this error, I close the PR.

Issue #1723, opened 2015-05-28
Enable libwebsockets support in Mosquitto


Diaoul:
It's all in the title.

marcelrv:
Did you already manage to get it in?? How to enable it?

Diaoul:
Well it's compiled with libwebsockets like asked. I don't know what it takes to actually configure mosquitto to use it.

marcelrv:
Indeed, I updated and found the lib. Great!!!

In the mean time I figured it out.. For others looking for the same,

edit the config file with : vi /usr/local/mosquitto/var/mosquitto.conf

add the following for a minimal setup:
```
listener 1883
listener 9001 127.0.0.1
protocol websockets
```

Diaoul:
Should the package include this as default config?

adamotte:
@Diaoul yes it would be easier for standard users (not familiar with ssh)

PR #2316, opened 2016-05-13
Updated for mosquitto 1.4.8 and DSM 6.0

### LINKED ISSUE #2297: mosquitto 1.4.8 update
Its almost a year ago.... :-)
Can you please update mosquitto to 1.4.8 ?
http://mosquitto.org/download/

And please enable websockets by default 
#1723

Thanx

sarfata:
- updated to latest version from upstream (closes #2297)
 - fix a bug in the startup script that prevented the service to
   start on DSM 6.0 (as discussed in #2207)
 - enable webservice listener by default (#1723)
 - updated libopenssl from 1.0.2g to 1.0.2h because the earlier is not
   available anymore for download
 - update the source for libwebsockets because the previous source
   caused TLS error in the docker container (note that the checksums
   have NOT changed)

Tested on a DS414 running DSM 6.0-7321.

NOT tested on DSM 5.x. Not sure if the `dsm-control.sh` will break
anything. Maintainers please advise!

### DIFF
--- a/cross/libwebsockets/Makefile
+++ b/cross/libwebsockets/Makefile
@@ -1,8 +1,8 @@
 PKG_NAME = libwebsockets
 PKG_VERS = 1.4-chrome43-firefox-36
 PKG_EXT = tar.gz
-PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = http://git.libwebsockets.org/cgi-bin/cgit/libwebsockets/snapshot
+PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://github.com/warmcat/libwebsockets/archive
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 DEPENDS = cross/openssl
--- a/spk/mosquitto/Makefile
+++ b/spk/mosquitto/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = mosquitto
-SPK_VERS = 1.4.2
-SPK_REV = 4
+SPK_VERS = 1.4.8
+SPK_REV = 1
 SPK_ICON = src/mosquitto.png
 DEPENDS = cross/busybox cross/$(SPK_NAME)
 
@@ -9,7 +9,7 @@ DESCRIPTION = Mosquitto is an open source message broker that implements the MQ 
 RELOAD_UI = yes
 DISPLAY_NAME = Mosquitto
 STARTABLE = yes
-CHANGELOG = "Add websockets support"
+CHANGELOG = "Update to 1.4.8 - Websocket on by default"
 HOMEPAGE = http://www.mosquitto.org/
 LICENSE  = BSD licensed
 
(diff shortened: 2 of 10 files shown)


timoline:
Great thx, how long will it take it will be merged?

sarfata:
I am not sure what are the next steps in the process. 

@cytec / @Dr-Bean can you let me know if you need anything else from me? Thanks!

cross/libwebsockets/Makefile
```diff
@@ -1,8 +1,8 @@
 PKG_NAME = libwebsockets
 PKG_VERS = 1.4-chrome43-firefox-36
 PKG_EXT = tar.gz
-PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = http://git.libwebsockets.org/cgi-bin/cgit/libwebsockets/snapshot
+PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://github.com/warmcat/libwebsockets/archive
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
```
cytec: use `$(PKG_NAME)-` in `PKG_DIST_NAME` so we get a proper filename ( i think that was the correct one \cc @Dr-Bean )

spk/mosquitto/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = mosquitto
-SPK_VERS = 1.4.2
-SPK_REV = 4
+SPK_VERS = 1.4.8
+SPK_REV = 1
```
cytec: `SPK_REV` Only goes upwards so set to `5`

spk/mosquitto/src/dsm-control.sh
```diff
@@ -7,7 +7,6 @@ DNAME="Mosquitto"
 # Others
 INSTALL_DIR="/usr/local/${PACKAGE}"
 PATH="${INSTALL_DIR}/bin:${PATH}"
-USER="mosquitto"
```
cytec: why? if you do this you'll need to handle start/stop of SPK too: https://github.com/sarfata/spksrc/blob/7e39296a747d48cd019aa47bbcedcb54578ab291/spk/mosquitto/src/dsm-control.sh#L17

cytec:
looks good except those comments i made ;)

cross/libwebsockets/Makefile
Dr-Bean: @cytec, think it's `PKG_DIST_FILE`. Having said that, it's probably a good idea to verify which version of libwebsockets to use. According to the changelog, Mosquitto should be able to support a newer version than 1.4 (1.6, it looks like)

Dr-Bean:
Thanks. Apart from the other comments, I have some more:
- Please verify that the configuration file is still up-to-date, and if not, replace it with an updated version.
- Even though we're not able to create users on DSM6 yet, that doesn't mean removing the user right now. If nothing else, it keeps the package running on DSM5. 
DSM6 user support will be handled via #2216, and will be added later. Unfortunately, for now that means that when we merge this PR and release a new package on the repo, it will not work on DSM6.
- Update your fork, rebase the PR and force-push the result. The OpenSSL version was updated a couple of hours ago, so that doesn't need to be part of the PR.

timoline:
In the mean time, 1.4.9 is released....

KaraokeStu:
Superceded by https://github.com/SynoCommunity/spksrc/pull/2418

KaraokeStu:
Duplicate of https://github.com/SynoCommunity/spksrc/pull/2418

PR #2337, opened 2016-05-22
Fix link issue with libdl

GuillaumeSmaha:


GuillaumeSmaha:
Example of error :

      CCLD     libxml2.la
    /lib/x86_64-linux-gnu/libc.so.6: file not recognized: File format not recognized
    collect2: ld returned 1 exit status

Dr-Bean:
I thought I solved all those issues...Which package were you testing when you saw this?

GuillaumeSmaha:
The issue occurs with `arch-x86-4.3` when I tried to add ICU to libxml2.
So I checked the lib path for all architectures. In most cases, the libdl.so is only defined in the directory /sys(-)root/usr/lib

GuillaumeSmaha:
I push on a branch my draft : https://github.com/GuillaumeSmaha/spksrc/commit/5333e40053167b8769d266c638008ca436f588e4

GuillaumeSmaha:
I am trying a build for all-arch to test it

GuillaumeSmaha:
This is the same for `arch-bromolow-4.3`.
I found a clue :

    echo "  CCLD    " libxml2.la;/bin/bash ./libtool --silent --tag=CC   --mode=link /storage/home/spksrc/toolchains/syno-bromolow-4.3/work/x86_64-linux-gnu/bin/x86_64-linux-gnu-gcc -D_REENTRANT -I/home/spksrc/cross/libxml2/work-bromolow-4.3/install/usr/local/include  -I/storage/home/spksrc/toolchains/syno-bromolow-4.3/work/x86_64-linux-gnu/x86_64-linux-gnu/sys-root/usr/include -I/home/spksrc/cross/libxml2/work-bromolow-4.3/install//usr/local/include -O3 -pedantic -W -Wformat -Wunused -Wimplicit -Wreturn-type -Wswitch -Wcomment -Wtrigraphs -Wformat -Wchar-subscripts -Wuninitialized -Wparentheses -Wshadow -Wpointer-arith -Wcast-align -Wwrite-strings -Waggregate-return -Wstrict-prototypes -Wmissing-prototypes -Wnested-externs -Winline -Wredundant-decls -Wno-long-long   -Wl,--version-script=./libxml2.syms -version-info 11:3:9 -ldl -L/storage/home/spksrc/toolchains/syno-bromolow-4.3/work/x86_64-linux-gnu/x86_64-linux-gnu/lib -L/storage/home/spksrc/toolchains/syno-bromolow-4.3/work/x86_64-linux-gnu/x86_64-linux-gnu/sys-root/lib64 -L/home/spksrc/cross/libxml2/work-bromolow-4.3/install//usr/local/lib -Wl,--rpath-link,/home/spksrc/cross/libxml2/work-bromolow-4.3/install//usr/local/lib -Wl,--rpath,/usr/local/lib  -o libxml2.la -rpath /usr/local/lib SAX.lo entities.lo encoding.lo error.lo parserInternals.lo parser.lo tree.lo hash.lo list.lo xmlIO.lo xmlmemory.lo uri.lo valid.lo xlink.lo HTMLparser.lo HTMLtree.lo debugXML.lo xpath.lo xpointer.lo xinclude.lo nanohttp.lo nanoftp.lo  catalog.lo globals.lo threads.lo c14n.lo xmlstring.lo buf.lo xmlregexp.lo xmlschemas.lo xmlschemastypes.lo xmlunicode.lo  xmlreader.lo relaxng.lo dict.lo SAX2.lo xmlwriter.lo legacy.lo chvalid.lo pattern.lo xmlsave.lo xmlmodule.lo schematron.lo xzlib.lo -L/usr/lib/x86_64-linux-gnu -licui18n -licuuc -licudata    -L/home/spksrc/cross/libxml2/work-bromolow-4.3/install/usr/local/lib -lz   -lm

There is this part `-L/usr/lib/x86_64-linux-gnu`, if I replace it by `home/spksrc/cross/libxml2/work-bromolow-4.3/install/usr/local/lib` then it works

GuillaumeSmaha:
Ok sorry, I found it -_- :
`icu-config --ldflags`  returns the bad value. I think that libxml2 uses it instead of the previous generated icu-config.

Else what do you think about adding these flags `-L ...` ? because there are divided in 2 directories. And currently just one is provided.

I also found an error due to a directory which is different to TC_BASE_DIR (https://github.com/SynoCommunity/spksrc/pull/2337/commits/21e7ce552cfc8ec136d25b4faa5803a62a1afdf4)

Dr-Bean:
For Armada375-5.1, use `TC_PREFIX` instead of creating a new var. I don't think I explicitly tested 5.1 as it's superseded by 5.2. I probably ended up simply replacing with the 5.2 contents.

As for adding the additional flags to LDFLAGS: I'd rather not merge it unless we are sure we actually need it.
I've spent quite some time in verifying and fixing spksrc to work with 64-bit hosts. That's not to say that there may be software out there that needs the additional path, but that looks very unlikely at this point. 
From what I can tell, your example with ICU supports that, because adding the extra path masked issues that should be solved in the (cross) packages.
Do you agree?

Dr-Bean:
Btw, haven't checked or anything, but maybe libxml2 accepts a path for `--with-icu` [here](https://github.com/GuillaumeSmaha/spksrc/commit/a8e25d5898bb6bc8991a8b37a75fae1f87a7469e#diff-076ce56a6f6358706f83cbba905047f0R14) like this?
```
CONFIGURE_ARGS = --without-python --with-icu=$(STAGING_INSTALL_PREFIX) --with-zlib=$(STAGING_INSTALL_PREFIX)
```

GuillaumeSmaha:
Yes, I think it's better to close it and keep it just in case. I will try a full rebuild to test without this PR.
And, I will create a new PR for Armada375-5.1

No, libxml2 doesn't support a link for icu. So I have to add `install/usr/local/bin` in PATH var use a postIntall fix to edit prefix var in `icu-config` (like libxml2 with xml2-config)
What do you think to add in https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.cross-env.mk#L5 this line ?

    ENV += PATH=$(INSTALL_DIR)/$(INSTALL_PREFIX)/bin:$$PATH

It will permit to avoid this case.

Dr-Bean:
Update libxml2 to latest/2.9.4 and try again. See https://git.gnome.org/browse/libxml2/commit/?id=6e3af87055d2c868d20d8b2b9da59bbafe84e5e0

GuillaumeSmaha:
Thanks for the link :)

GuillaumeSmaha:
I close this PR until an error libdl is reported.

Issue #2299, opened 2016-05-04
Add Issue- and Pull Request templates


Dr-Bean:
This hopefully helps people to add sufficient information when opening an issue or pull request, which int turn makes things easier for other contributors.

Setup: https://github.com/blog/2111-issue-and-pull-request-templates. I kind of like the idea of using a separate `github` dir for the templates and CONTRIBUTING.

Content for issues we already have: https://github.com/SynoCommunity/spksrc/blob/master/CONTRIBUTING.md#issue-content
Content for PR's: at least name+description of PR (/description of package), tbd beyond that. 
We could even consider setting up a checklist (either in the template or in a wiki): when adding new packages, add a `beta` key, `SPK_REV` should be set to `1`, stuff like that.

Dr-Bean:
The formatting in Github's editor doesn't allow for working links etc, I suppose we're kind of limited in the way it looks.
Anyway, let's see how this works, I'll leave this open for a bit to see if we can improve on things.

Dr-Bean:
Current thoughts after looking at it for a bit:
- Change checklist in PR template to something like the text below...we don't care about legacy toolchains all that much:
    * Build rule `all-supported` completed successfully
    * Tested successfully
- Remove the package request template. Either integrate the content into the issue template as a separate section, or just keep it in CONTRIBUTING
- Minimize text/explanation in both templates, and just keep pointing to the guidelines in CONTRIBUTING instead.

Dr-Bean:
Done with 600897d7d2e6827bb600de2a4e663b02b3ad581d.

PR #2339, opened 2016-05-23
Add issue and package request templates

Dr-Bean:
Via https://github.com/SynoCommunity/spksrc/issues/2299.

We can move CONTRIBUTING.md to the same dir, and we can adjust its contents to directly point to the templates instead of the link I added in there a while back.


Dr-Bean:
No Pull Request template yet, so that's another TODO

PR #2347, opened 2016-05-26
Allow to use NATIVE_* flags in native builds

GuillaumeSmaha:
_Motivation:_ Allow to use ADDITIONAL_* flags in native builds
If you try to build native/python without this PR, you will not see the additional flags in the gcc output.

    ===>  Compiling for native-python 
    make[1] : on entre dans le répertoire « /storage/home/spksrc/native/python/work-native/Python-2.7.11 »
    gcc -pthread -c -fno-strict-aliasing -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes  -I. -IInclude -I./Include   -DPy_BUILD_CORE -o Modules/python.o ./Modules/python.c

But with the PR you will see this :

    ===>  Compiling for native-python 
    make[1] : on entre dans le répertoire « /storage/home/spksrc/native/python/work-native/Python-2.7.11 »
    gcc -pthread -c -fno-strict-aliasing -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes  -I. -IInclude -I./Include   -DPy_BUILD_CORE -o Modules/python.o ./Modules/python.c


Dr-Bean:
I was thinking about this for a bit, but we can't merge this as-is.

The general idea of the spksrc framework is that you can compile an spk, which in turn compiles a cross package, which in turn builds a native package, right? (broken for python atm, local fix pending)

Now, native packages are built for the host (as in: not cross compiled). We don't want to pass any flags from the cross compile environment to native packages: you could end up with settings in the native package that should only apply to the cross packages. That's why all the flags are explicitly set to nothing in `native-env.mk`.

If we do merge this, the two flags will pass through to the native package if the build is initiated through the cross package (via `tc_vars.mk`)
For the Python package(s), the contents of both flags happen to be the same...but that's not a given at all. So, an alternative could be to use a different variable name.

On an semi-related note, there's a good chance the native Python packages actually don't need to have `_LARGEFILE64_SOURCE=1` and `_FILE_OFFSET_BITS=64` set at all. See `pyconfig.h`.

GuillaumeSmaha:
Yes I agree with you, native build are only used by the host to help the cross-compiling like yasm, python, ...
Using python was only an example because I found the variable ADDITIONAL_* in native/python/Makefile and I didn't know if the flag was important or not. 

I think it could be useful in the future with a case which will need to pass specific variables to the compiler.

Dr-Bean:
I didn't make it very clear in my previous comment, but I do think it's a good addition to the framework, so nice catch :)

As I said, how about using a different variable name? Maybe use `NATIVE` in the name, like `NATIVE_CFLAGS` (using `ADDITIONAL_*` seems unnecessary, as there's nothing 'additional' about them if we clear all previous flags).

Whatever the choice, let's also add it to the wiki [here](https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables), so we have it documented.

GuillaumeSmaha:
Hum ok, I understand. Sorry for the misssunderstanding :)

Dr-Bean:
We should fix the Python Makefiles as well I guess. Can be done in a separate PR, depending if it makes a difference whether or not the arguments are used or not.

GuillaumeSmaha:
I will create another PR to edit Python Makefiles. I have to search if the flags are useful or not.

GuillaumeSmaha:
I update wiki : https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables

Dr-Bean:
Thanks!

PR #2352, opened 2016-05-30
update go to 1.6.2

cytec:

### DIFF
--- a/native/go-1.4/Makefile
+++ b/native/go-1.4/Makefile
@@ -20,7 +20,3 @@ include ../../mk/spksrc.native-cc.mk
 myComp:
 	@$(MSG) "Building Go"
 	cd $(WORK_DIR)/$(PKG_NAME)/src && ./make.bash
-	@$(MSG) "Building Go for arm"
-	cd $(WORK_DIR)/$(PKG_NAME)/src && export GOARCH=arm && export GOARM=5 && ./make.bash --no-clean
-	@$(MSG) "Building Go amd64"
-	cd $(WORK_DIR)/$(PKG_NAME)/src && export GOARCH=amd64 && ./make.bash --no-clean
--- a/native/go-1.4/Makefile
+++ b/native/go-1.4/Makefile
@@ -1,0 +1,26 @@
+PKG_NAME = go
+PKG_VERS = 1.4.2
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)$(PKG_VERS).src.$(PKG_EXT)
+PKG_DIST_SITE = https://storage.googleapis.com/golang
+PKG_DIR = $(PKG_NAME)
+
+DEPENDS =
+
+HOMEPAGE   = http://golang.org/
+COMMENT    = Go is an open source programming language that makes it easy to build simple, reliable, and efficient software.
+LICENSE    =
+
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = nop
+INSTALL_TARGET = myComp
+
+include ../../mk/spksrc.native-cc.mk
+
+myComp:
+	@$(MSG) "Building Go"
+	cd $(WORK_DIR)/$(PKG_NAME)/src && ./make.bash
+	@$(MSG) "Building Go for arm"
+	cd $(WORK_DIR)/$(PKG_NAME)/src && export GOARCH=arm && export GOARM=5 && ./make.bash --no-clean
+	@$(MSG) "Building Go amd64"
+	cd $(WORK_DIR)/$(PKG_NAME)/src && export GOARCH=amd64 && ./make.bash --no-clean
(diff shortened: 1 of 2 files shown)


native/go-1.4/Makefile
```diff
@@ -0,0 +1,26 @@
...
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = nop
+INSTALL_TARGET = myComp
+
+include ../../mk/spksrc.native-cc.mk
+
+myComp:
+	@$(MSG) "Building Go"
+	cd $(WORK_DIR)/$(PKG_NAME)/src && ./make.bash
+	@$(MSG) "Building Go for arm"
```
Dr-Bean: Seeing as this package is purely for bootstrapping 1.6, L21/L22 would suffice, right? No need to build ARM or amd64: amd64 is what's built when running `./make.bash`, and that's all you'd need to bootstrap the other native package.

native/go/Makefile
```diff
@@ -19,8 +19,10 @@ include ../../mk/spksrc.native-cc.mk
 
 myComp:
 	@$(MSG) "Building Go"
-	cd $(WORK_DIR)/$(PKG_NAME)/src && ./make.bash
+	cd $(WORK_DIR)/$(PKG_NAME)/src && GOROOT_BOOTSTRAP=$(WORK_DIR)/../../go-1.4/work-native/go ./make.bash
```
Dr-Bean: From a DRY pov: Can you define `GOROOT_BOOTSTRAP=$(WORK_DIR)/../../go-1.4/work-native/go` on L17 or thereabouts, then set `GOROOT_BOOTSTRAP=$(GOROOT_BOOTSTRAP)` instead?

native/go/Makefile
```diff
@@ -19,8 +19,10 @@ include ../../mk/spksrc.native-cc.mk
...
 	@$(MSG) "Building Go for arm"
-	cd $(WORK_DIR)/$(PKG_NAME)/src && export GOARCH=arm && export GOARM=5 && ./make.bash --no-clean
+	cd $(WORK_DIR)/$(PKG_NAME)/src && export GOARCH=arm && export GOARM=5 && GOROOT_BOOTSTRAP=$(WORK_DIR)/../../go-1.4/work-native/go ./make.bash --no-clean
 	@$(MSG) "Building Go for 386"
-	cd $(WORK_DIR)/$(PKG_NAME)/src && export GOARCH=386 && ./make.bash --no-clean
+	cd $(WORK_DIR)/$(PKG_NAME)/src && export GOARCH=386 && GOROOT_BOOTSTRAP=$(WORK_DIR)/../../go-1.4/work-native/go ./make.bash --no-clean
+	@$(MSG) "Building Go for amd64"
```
Dr-Bean: You might have noticed that the build output at some point looks like this:
```
===>  Building Go for 386
##### Building Go bootstrap tool.
##### Building Go toolchain using /home/bean/spksrc/native/go/work-native/../../go-1.4/work-native/go.
##### Building go_bootstrap for host, linux/amd64.
##### Building packages and commands for host, linux/amd64.
##### Building packages and commands for linux/386.
```
Essentially, every time you build for a specific `GOARCH`, Go starts from the beginning: it checks if there's a bootstrap available for the host, then builds the actual `go` command for host (which is amd64 for spksrc), and then builds the command for the specified `GOARCH`.
That means that L27/28 are not needed: host/amd64 is already built at that point: you actually start off with it on L22.

Dr-Bean:
Added some comments, but otherwise :+1:

native/go/Makefile
Dr-Bean: ...and if you keep L28, please replace the spaces with a TAB :)

native/go-1.4/Makefile
cytec: i'm not sure if thats enough but i'll try

native/go/Makefile
cytec: yup, good point

native/go/Makefile
cytec: not sure about this... currently i'm still useing a 386 env for building/testing thats why that line was there... maybe we should get rid of the first line (autodetected) and hard code all lines? or is amd64 needed in the future?

Dr-Bean: ...see the README for supported environments? ;)
Seriously though, if that's the case, then probably hardcoding is the way to go. Explicitly do amd64 as the first one (which should build 386 first on your env), then 386, then arm? It removes one unnecessary 'rebuild' (fwiw, because it's more of a verification at that point) I guess.

cytec: i guess my build system is pretty old now... i should migrate it to x64 anyways...
so change those lines or not? up to you ;)

Dr-Bean: Go ahead and change. The README says we support x64, but that's not the same as actively preventing x86 from being used.
Iirc, Syncthing was the only package where it mattered which hostarch you were using, and you've largely removed the differences in this PR.
So if we can remove another difference by being explicit in what we're building, why not.

...assuming it works properly on both sides, but that should be easy to verify.

Issue #2191, opened 2016-03-14
sabnzbd 1.0 is out!


Stanzilla:
https://github.com/sabnzbd/sabnzbd/tree/1.0.0

lysin:
+1

mhartvig:
Updated the testing channel to version 1.0.0 in #2194

Stanzilla:
Why testing though? It's a final release?

Dr-Bean:
That's a separate package, specifically for beta versions of Sab and people willing to test with that.
For stable, we're waiting on #2195 (or even a second PR that takes the comments there into account, either works)

mbnn:
Are we getting 1.0 in the 'non beta' syno package pls?:)

Popsikill:
Keep to get this up and running on my Synology :+1:

mbnn:
Any update on the SABNZBD 1.0 release? :)

Veldkornet:
I see #2195 has been closed and #2201 merged. 
Also the site is showing v1.0.0-15 -> https://goo.gl/4jXkQa

What else still needs to happen? My Synology isn't seeing the update yet.

mbnn:
Same here :) Where is the 1.0 version pls? :)

SmokE-PGF:
Doesn't show a way to update under the Package Center am I missing something?

Dr-Bean:
Please be patient. If you want it immediately, feel free to compile it yourself.

techabyte:
Sab on latest DSM removes unzip so sab is currently broken till 1.0 is pushed or releases a fix for 0.7 to use 7zip :/

r33mi:
Rest is working fine, upgraded my SABnzbd myself. unzip is broken though...

SmokE-PGF:
Ran Sab in a docker container seems to have fixed everything even unzips fine

Diaoul:
Well of course, Docker is like another machine. That workaround is limited to NASes able to run Docker in the first place.

jarrod-s:
As Veldkornet noted the individual package entry shows - SABnzbd v1.0.0-15 
However in the main index of all packages shows - SABnzbd v0.7.20-14 
I assume Synology Package Centre compares to the main index so does not believe there is an update.

mbnn:
Says: https://synocommunity.com/packages  0.7.20 aswell, so I guess it's not done yet ;)

loli71:
[...]

PR #2365, opened 2016-06-07
Updated SABnzbd to 1.0.3

KaraokeStu:
Upgraded SABnzbd to 1.0.3 as per @Dr-Bean's request, I have made this its own pull request with a single commit

Fixes https://github.com/SynoCommunity/spksrc/issues/2191

### DIFF
--- a/spk/sabnzbd/Makefile
+++ b/spk/sabnzbd/Makefile
@@ -1,10 +1,10 @@
 SPK_NAME = sabnzbd
-SPK_VERS = 1.0.0
-SPK_REV = 15
+SPK_VERS = 1.0.3
+SPK_REV = 16
 SPK_ICON = src/sabnzbd.png
 DSM_UI_DIR = app
 
-DEPENDS  = cross/busybox cross/par2cmdline cross/unrar cross/$(SPK_NAME)
+DEPENDS  = cross/busybox cross/par2cmdline cross/unzip cross/unrar cross/$(SPK_NAME)
 SPK_DEPENDS = "python>=2.7.6-8"
 
 MAINTAINER = Diaoul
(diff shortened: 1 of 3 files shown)


spk/sabnzbd/Makefile
```diff
@@ -14,7 +14,34 @@ DESCRIPTION_SPN = SABnzbd hace que Usenet sea lo más simple posible, automatiza
 ADMIN_PORT = 8080
 RELOAD_UI = yes
 DISPLAY_NAME = SABnzbd
-CHANGELOG = "1. Update to 1.0.0<br>2. Warning: When you upgrade an existing 0.7.x installation, release 1.0.0 will NOT be able to finish the queue. It's best to empty your queue first using 0.7.x. Alternatively, you can install 1.0.0 and then use queue repair on the Status page. Some information (like the order of the queue) will be lost."
+CHANGELOG = "1. Update to 1.0.0<br>2. Warning: When you upgrade an existing 0.7.x installation, release 1.0.0 will NOT be able to finish the queue. It's best to empty your queue first using 0.7.x. Alternatively, you can install 1.0.0 and then use queue repair on the Status page. Some information (like the order of the queue) will be lost.
```
Dr-Bean: Please leave the changelog as-is (although you should update the version to what it should be, 1.0.3, not 1.0.0 at this point)

Package Center can't handle the formatting, you'd need to convert it to html for it to work properly. That's why we generally put just the minimum amount of info in there: it's a package changelog, not a software changelog. If people want to see the changes in Sabnzbd itself, they can look it up on the sabnzbd website.

Dr-Bean:
One comment, it applies to the sabnzbd-testing package as well.
If you want, you can also add `cross/unzip` to DEPENDS, it solves the warning people currently see. I think I already added it to the -testing package, but not to the stable package yet. Otherwise we can do that in a separate commit when merging.

KaraokeStu:
Updated commit with dependency on unzip and made changelog much shorter with HTML line breaks (as it had before)

Dr-Bean:
The changelog isn't cumulative. This is the text I'd suggest:
```
CHANGELOG = "1. Update to 1.0.3<br>2. Warning: When you upgrade an existing 0.7.x installation, release 1.0.3 will NOT be able to finish the queue. It's best to empty your queue first using 0.7.x. Alternatively, you can install 1.0.3 and then use queue repair on the Status page. Some information (like the order of the queue) will be lost."
```

Dr-Bean:
...and we might as well add in the unzip addition, so it looks like this:
```
CHANGELOG = "1. Update to 1.0.3<br>2. Add unzip<br>3. Warning: When you upgrade an existing 0.7.x installation, release 1.x will NOT be able to finish the queue. It's best to empty your queue first using 0.7.x. Alternatively, you can install 1.x and then use queue repair on the Status page. Some information (like the order of the queue) will be lost."
```

KaraokeStu:
Updated to use correct HTML and remove cumulative changelog.
Made the warning red and italic ;)

Dr-Bean:
Thanks, though I don't know that Package Center allows all HTML tags...

KaraokeStu:
Updated as per @Dr-Bean requested

Dr-Bean:
Multiple entries in the changelog should be numbered (somewhat documented at https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables, but it was the format previously used in the Makefile as well).
We can fix that when merging.

Dr-Bean:
Additional changes with 13150d8e463155f29130294fe9c019bb8e673eb4.
sabnzbd 1.0.3-16 has been published on the repository.

Issue #2368, opened 2016-06-08
p7zip - cross/p7zip refresh needed to v16.x when available


sirjaymz:
### Setup
_Package Name: p7zip
_Package Version: 9.20.1

_NAS Model: DS1813+
_NAS Architecture: cedarview
_DSM version: 5.2-5967

http://www.7-zip.org/

Multiple 7-Zip Vulnerabilities
http://blog.talosintel.com/2016/05/multiple-7-zip-vulnerabilities.html

Concerns for other SPK's that rely on p7zip support could  open DSM device up to vulnerabilities that can be resolved with quick update. 

@Diaoul if you don't have time, I'll see if i can setup a PR.

sirjaymz:
15.14.1 compiles, with a new patch file. 
This new compile however does not get the needed fix though.. 
tried using  "diff -Naur" between the 16.02 source of 7zip and the source of 15.14.1 to make additional patch files between those sources of the code referenced on https://sourceforge.net/p/p7zip/discussion/383043/thread/9d0fb86b/#6ecc without success. 
still working it.

sirjaymz:
success.. setting up PR.

Dr-Bean:
nzbget 16.4-22 	and nzbget-testing 16.0~r1439-11 are the two affected packages, both have been published on the repository.

Closing: If/when p7zip will be updated to 16.x, it'll quite likely contain the exact same patches.

PR #2370, opened 2016-06-10
Hotfix for p7zip #2368.

sirjaymz:
_Motivation:  Latest version of P7zip 15.14.1 with 2 vulnerability patched from 16.02 7-zip src.
_Linked issues: https://github.com/SynoCommunity/spksrc/issues/2368


cross/p7zip/digests
```diff
@@ -1,3 +1,2 @@
-p7zip_9.20.1_src_all.tar.bz2 SHA1 1cd567e043ee054bf08244ce15f32cb3258306b7
-p7zip_9.20.1_src_all.tar.bz2 SHA256 49557e7ffca08100f9fc687f4dfc5aea703ca207640c76d9dee7b66f03cb4782
-p7zip_9.20.1_src_all.tar.bz2 MD5 bd6caaea567dc0d995c990c5cc883c89
+p7zip_15.14.1_src_all.tar.bz2 SHA1 9b15a79f94230fab9b9d4f9f532c723117145c7a
+p7zip_15.14.1_src_all.tar.bz2 MD5 92cca093312b5a71a7be7dc7d1d32509
```
sirjaymz: sha256 checksum is not available on Sourceforge...
So whats the procedure? Am i suppose to generate a sha256? i only provided the 2 checksums that were available by the sourceforge developers, as i would think they 'own' that deliverable of creating the checksums. thoughts?

GuillaumeSmaha: Yes, you just generate the sha256 for the file.

sirjaymz:
added sha256 checksum to the digest

cross/p7zip/digests
sirjaymz: added sha256 checksum to the digest

Dr-Bean:
@sirjaymz Thanks. The individual patch contents look okay to me, but I will probably cherry-pick the correct commits to clean up the PR.

For future reference, please don't remove the checkboxes in the PR template: they're there to make sure we know if you've tested the PR and in what sense. I'll assume you didn't test.

cross/p7zip/digests
Dr-Bean: Use `make digests` to create a digests file (and then verify the outcome against upstream). Of course, that only works if you have a dev environment set up ;)

Dr-Bean:
Redid the update/patches here: https://github.com/SynoCommunity/spksrc/commit/0e8a35efdbf5c07deb26fd94198fbbeb983d3cf9.
Thanks for taking the time to create a PR :+1:

PR #2371, opened 2016-06-10
Allow to define a set of ARCH-DSM as unsupported

GuillaumeSmaha:
_Motivation:_ Allow to define a set of ARCH-DSM as unsupported 
### Checklist

### DIFF
--- a/mk/spksrc.cross-cc.mk
+++ b/mk/spksrc.cross-cc.mk
@@ -41,6 +41,14 @@ ifneq ($(REQUIRED_DSM),)
   endif
 endif
 
+# Check if package supports ARCH-DSM
+ifneq ($(UNSUPPORTED_ARCHS_DSM),)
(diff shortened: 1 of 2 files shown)


Dr-Bean:
This doesn't work as a generic solution, I don't think. `make arch-88f6281-5.0` would not trigger the error even if `88f6281-5.0` is set as UNSUPPORTED. Are you issuing commands like `make ARCH=arch TCVERSION=tcversion`?

We originally split up ARCH and TCVERSION in the [Makeflags](https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.cross-cc.mk#L123) for reasons described [here](https://github.com/SynoCommunity/spksrc/pull/1276).
It might be easier to use a concatenated `arch+tcversion` (or `toolchain` if you will) to do operations, provided that we don't break other functionality (such as the REQUIRED_DSM check). That was my original implementation, and maybe that's something to consider?

hgy59:
@GuillaumeSmaha is there any use case for this?

hgy59:
It looks like there is no need for this any more.

Issue #2420, opened 2016-07-27
Please update NZBGet to 17.0


joeyd2:
Please update NZBGet to 17.0

rvdlaar:
What he said

1shez:
What he said

benjmd:
What he said

KaraokeStu:
Updated as per request, pull request pending

rvdlaar:
Awesome. Thanks.

KaraokeStu:
No problem

KaraokeStu:
I have closed the pull request due to problems with requiring GCC 4.9 or 5.0

KaraokeStu:
Please close this issue - it is a duplicate of https://github.com/SynoCommunity/spksrc/issues/2185

joeyd2:
So it isn't getting updated? :(

KaraokeStu:
It will be supported, but only from DSM 6.x - unless someone knows how to compile GCC 4.9 for DSM 5.x.
And DSM 6.x still has its own set of issues working with custom packages

Thaek:
Any news on this? :-) NZBGet 17 was released almost two weeks now and there is a feature that I requested that was implemented in 17, would be very nice if I could test it.

So if I have DSM 6.x can I expect this to be updated?

Thaek:
Someone backported NZBGET 17 to work with C++ 11, maybe this will help you for DSM 5.x? http://nuxref.com/2016/08/03/nzbget-v17-centos-7/

cytec:
seems like that got already merged (see https://github.com/nzbget/nzbget/pull/256) and is flagged for 17.1 so let's wait for 17.1 to be released (and hopfeully compile for DSM 5x)

Thaek:
Wait for 17.1? ;-/ ugg that could take ages, it's not like they create releases every month.  :-(

KaraokeStu:
Feel free to amend the package yourself and create the relevant patch for it, however I don't think it would be worth the double effort of us doing that, then undoing it once 17.1 is released.

I wouldn't think it'll be too long for 17.1, given that it will be likely a fair few bugs will have crept  into the new codebase

cytec:
@schumi2004 have you tried ncursesw?

schumi2004:
@cytec doesn't seem to make any difference.

cytec:
[...]

PR #2423, opened 2016-08-03
Updated NZBGet to version 17.0

KaraokeStu:
Updated to the latest version as per requested here:
https://github.com/SynoCommunity/spksrc/issues/2420

### Checklist
- [x] Confirmed no configuration changes

### DIFF
--- a/cross/nzbget/Makefile
+++ b/cross/nzbget/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = nzbget
-PKG_VERS = 16.4
+PKG_VERS = 17.0
 PKG_EXT = tar.gz
 PKG_DIST_SITE = https://github.com/$(PKG_NAME)/$(PKG_NAME)/releases/download/v$(PKG_VERS)
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS)-src.$(PKG_EXT)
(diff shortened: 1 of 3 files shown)


evaryont:
I think the changelog in `spk/nzbget/Makefile` should be updated? The changes in v17 is bigger than just updating a library for security reasons.

KaraokeStu:
The general standard is just to include the actual changes in the package, not the software itself - for that you should look at the referenced website

evaryont:
Sure, I figured that was the case. I guess then instead the second line
could be replaced with a URL/link? Just being pedantic in the end though.
Thanks for making the PR!

DaftHonk:
When I try to compile for cedarview-6.0 I receive the following error when configuring nzbget:
```
checking for ncurses... no
configure: error: "ncurses library not found"
../../mk/spksrc.configure.mk:50: recipe for target 'configure_target' failed
```
ncurses-6.0 built correctly in the work directory and libncurses5-dev is installed on my system (ubuntu 16.04 x64).

Also, only 6.0 TCs should support as previous toolchains do not have the required level of GCC (>=4.9); did you successfully compile for any 5.2 toolchains?

KaraokeStu:
Closed as @DaftHonk states, this requires GCC 4.9 or 5.0, which is not currently enabled - we may be able to build it as a cross binary, but not right now!

Issue #2449, opened 2016-09-12
ZSH not built with regex support


forivall:
### Setup

_Package Name:_ zsh
_Package Version:_ 5.2-3

_NAS Model:_ DS214se
_NAS Architecture:_ MARVELL Armada 370 88F6707
_DSM version:_ DSM 6.0.2-8451 (also happened on 5)
### Expected behavior

```
norwegian-blue% autoload -U regexp-replace    
norwegian-blue% url=foo/bar                     
norwegian-blue% regexp-replace url '/' '-SLASH-'
norwegian-blue% echo $url                       
foo-SLASH-bar
```
### Actual behavior

```
norwegian-blue% autoload -U regexp-replace    
norwegian-blue% url=foo/bar                     
norwegian-blue% regexp-replace url '/' '-SLASH-'
regexp-replace:28: failed to load module: zsh/regex
regexp-replace:28: -regex-match not available for regex
norwegian-blue% echo $url                       
foo/bar
```
### Steps to reproduce

_1._ Install zsh
_2._ Run the above commands in zsh

The build option to enable regex needs to be enabled: https://github.com/robbyrussell/oh-my-zsh/issues/227#issuecomment-816676

KaraokeStu:
I'll try to get on this asap for you

KaraokeStu:
Unfortunately, I don't know enough about where ZSH is pulling its modules from - it appears to be a problem dynamically loading the module, rather than not having the right mode built.

We unfortunately need the help of someone who understands ZSH better

x1a0:
@KaraokeStu @forivall 
Hi, I can confirm if you modify `config.modules` file generated in zsh folder during configuration and make `zsh/regex` module `link=static` then it works.
But I don't know what's the best way to patch it during compilation.
(I tested on my 1515+, `make arch-avoton-5.2`)

KaraokeStu:
Great, thanks for the info. I'll need to check with @Diaoul to see how we can patch this into the make file.

@forivall Any ideas if there are any other modules you'd expect to be available too? 

Ideally, it'd be better to get the dynamic module loading working, but I'm not sure how!

forivall:
@KaraokeStu I would need to take a look at what modules _are_ in zsh in the first place; I ran into this first when trying to load my dotfiles. The only one that came to mind are the vcs hooks.

Googling gets me the full list: http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html

I'll try that when I get home.

[...]

PR #2452, opened 2016-09-19
Updated ZSH to build with Regex Support. Also installed ZSH dynamic modules

KaraokeStu:
As requested here: https://github.com/SynoCommunity/spksrc/issues/2449

ZSH was being built without dynamic module or PCRE support, which meant no Regex could be used (and it is used in lots of ZSH scripts)

_Linked issues:_ 
https://github.com/SynoCommunity/spksrc/issues/2449

### Checklist


KaraokeStu:
Unfortunately, I don't know enough about where ZSH is pulling its modules from - it appears to be a problem dynamically loading the module, rather than not having the right mode built.

I'm going to close this PR until someone who understands ZSH better can help

PR #2459, opened 2016-09-22
Removed SSL Repo Verification

KaraokeStu:
Updated to not verify the remote repository API SSL as api.synocomminity.com is currently expired


KaraokeStu:
Could you provide your thoughts please @Diaoul and @Dr-Bean ?

Diaoul:
That's alright, the connection is still secure. I need to renew the certificate with letsencrypt but lack of time to do so.
I don't see this as really being an issue but more an inconvenience.

KaraokeStu:
Try the certificates available here, not wildcard, but you can specify a few subdomains (I used it for my own site)

https://www.wosign.com/english/freessl.htm

PR #2504, opened 2016-10-28
updates with new location of certificate files for DSM 6

joshlawless:
_Motivation:_ DSM changed certificate location (and filenames?) in v.6
_Linked issues:_ This solves ticket 2340, I think
### Checklist

### DIFF
--- a/spk/gateone/src/dsm-control.sh
+++ b/spk/gateone/src/dsm-control.sh
@@ -17,7 +17,9 @@ USER="gateone"
 
 start_daemon ()
 {
-    # Copy certificate
+    # Copy certificate command for DSM5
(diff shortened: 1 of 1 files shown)


Dr-Bean:
While this may work for DSM6, it breaks the behavior for DSM5 and older, so that's no good. 

We need a better way to set the path for those two files. Maybe use the same approach we use elsewhere, see https://github.com/SynoCommunity/spksrc/blob/master/spk/tt-rss/src/dsm-control.sh#L12 for example.

joshlawless:
What if you left in both copy commands, and suppressed any error message
from the one that didn't work?

Dr-Bean:
Personally, don't like that :)

hgy59:
DSM 6 critificate paths are fixed since update to gateone 20171125-7

Issue #2503, opened 2016-10-27
Readme is missing Usage section


mastacheata:
Up until January, the README contained instructions how to use the spksrc framework to build your own packages.
In this commit 3d260cbd4e121f5e93fdf8b2168baf9174bea67a you removed the whole usage section. Was that on purpose? The commit message doesn't mention removing the Usage section, but only says that @Dr-Bean added some stuff on contributing.

Dr-Bean:
Yes, that was on purpose...
The instructions were moved, as part of an ongoing effort to extend the documentation. See the [Pull Requests section](https://github.com/SynoCommunity/spksrc/blob/master/CONTRIBUTING.md#pull-requests), which refers to [spksrc-example](https://github.com/SynoCommunity/spksrc/wiki/spksrc-example). The example is also linked to from the [Developers HOW-TO](https://github.com/SynoCommunity/spksrc/wiki/Developers-HOW-TO) page.

mastacheata:
Maybe you could add a link in the readme then,because that is usually my first place to look for information. Also you kept the requirements section in the readme, but removed any hint on how to use the things,which is very confusing. I wouldn't really go to a section called pull requests on order to use the tool itself. 

That's my opinion at least. 

Am 31. Oktober 2016 16:09:11 MEZ, schrieb Dr-Bean <notifications@github.com>:

Dr-Bean:
Not saying you're wrong ;) We should probably add a link to the Developers How-to, as that should be the central location to start off at. Sound reasonable?

mastacheata:
I wasn't complaining either. Just providing insight from someone less involved with the project 😉
Sounds very reasonable to me. 

Am 31. Oktober 2016 16:48:37 MEZ, schrieb Dr-Bean <notifications@github.com>:

Dr-Bean:
Done.
See https://github.com/SynoCommunity/spksrc/wiki/Developers-HOW-TO, the example stuff is integrated into that. The README refers to it via [Usage](https://github.com/SynoCommunity/spksrc#usage). Also CONTRIBUTING refers to it, here: [Pull Requests](https://github.com/SynoCommunity/spksrc/blob/master/CONTRIBUTING.md#pull-requests)

PR #2505, opened 2016-10-28
Put the Usage section back into README

mastacheata:
See #2503 I believe this was deleted accidentally, as now there's no reference at all how to use spksrc to build any of the packages in any document of this repository.


Diaoul:
Thx

PR #2528, opened 2016-11-20
Increase max icon resolution to 256

joshlawless:
_Motivation:_ Increase max icon resolution (for better support of Retina and other HD displays)

### DIFF
--- a/spk/transmission/Makefile
+++ b/spk/transmission/Makefile
@@ -40,7 +40,7 @@ transmission_extra_install:
 	install -m 755 -d $(STAGING_DIR)/app
 	install -m 644 src/app/config $(STAGING_DIR)/app/config
 	install -m 755 -d $(STAGING_DIR)/app/images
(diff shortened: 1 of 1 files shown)


Dr-Bean:
The value of 128 isn't valid according to the dev guide.

The icon in Package Center is set via https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.spk.mk#L180. 
The docs are a bit confusing, but I think we should rename `PACKAGE_ICON_120.PNG` to `package_icon_120` (it's deprecated, so either it's that, or completely remove so PC will fallback to the 72x72 icon), and add a `PACKAGE_ICON_256.PNG` with the appropriate values.

Dr-Bean:
I didn't notice this when I commented before, but changes like this should be done for all affected packages, not just one package.
#2580 should fix this for the forseeable future once merged, so closing.

Issue #1522, opened 2015-03-04
SPK for Gogs possible?


Snipe3000:
Would it be possible to create a SPK for Gogs for install in Diskstation?
http://gogs.io/
##

cytec:
found that toll myself a few days ago...
afaik it only depends on git and mysql/sqlite/pgsql so it should be doable.

sadly i have very limited time atm, so don't know if and when i can take a closer look

Snipe3000:
No problem, please let me know if you ever get around to looking at it.

Decipher:
It requires [Go](http://golang.org/), so until there is a SPK for that, I can't imagine this would be possible.

cytec:
I've made a native golang package to make cross compiling go code easier. You don't need to have go installed on your DS, it's just needed for compiling the code :)

as far as i can tell from the website, it only depends on git (for which a spk is present already) and MySQL/PostgreSQL/SQLite (sqlite has also a makefile) so it _should_ run i think... still have a lot of stuff around and was not really able to take a closer look :(

cytec:
@Decipher @Snipe3000 good news! had some time to dive into it. I got it working on my DS1513+ (cedarview) i haven't tried any other platforms yet nor did i managed to create a SPK makefile for automation (due to lack of time sadly) but it compiles and after being copied to my DS it runs and installs :) 

So it's only a matter of time till there will be a SPK ready

Decipher:
I also tried to ge t it running myself last night, and also managed to compile a binary that partially ran on my XPEnology, but I did encounter a few issues. I also found that Golang doesn't currently support PPC which my physical NAS is, so it's somewhat less priority for me at the moment... With that said though, if you @cytec commit your work is be happy to help finish it (time permitting).

Snipe3000:
Awesome, sounds like its coming along nicely.

cytec:
@Decipher i've just pushed what i have done so far: https://github.com/cytec/spksrc/commit/b9e21e271b4f2f7cef6caed7b546046e19758b2b

NOTE: i use the `bin/custom/conf` directory for the configuration file due to debugging and testing purpose... in the final SPK this should be moved to the `var` folder to preserve it througout SPK updates. there is a ENV variable called `GOGS_CONFIG` to provide a custom config path, see: https://github.com/gogits/gogs/issues/145

Decipher:
Cheers, will try to take a look at it this evening.

cytec:
[...]

Issue #1344, opened 2014-11-19
[haproxy] Help and howto link in package center is not live


piejanssens:
I was glad to see a "help and howto" link for this package because I want to setup reverse proxy using this tool, but the link is not active: https://synocommunity.com/help/haproxy

wind-rider:
the help is now at http://htmlpreview.github.io/?https://github.com/SynoCommunity/spksrc/blob/master/spk/haproxy/src/app/help/enu/index.html

wind-rider:
@Diaoul why was this one closed? the original question (why does the help link not work) was not resolved

Dr-Bean:
#1037, or any of the other issues pertaining to the help not working. The html preview link was just an example, you could just as well read the raw file.

PR #2533, opened 2016-11-22
Gogs v0.9.97

torrentalle:
This feaure add [Gogs ](https://gogs.io/)package. Gogs is a self-hosted Git service written in Go which is very easy to get running and has low system usage as well. It aspires to be the easiest, fastest, and most painless way to set up a self-hosted Git service.

### Checklist
- [x] Build rule `all-supported` completed successfully. See [Travis build](https://travis-ci.org/torrentalle/spksrc/builds/315074251)

Tested on platforns:
  - DSM-6.0 arch evansport
  - DSM 5.2 on  XPEnology (x64)


torrentalle:
This resolves #1522, and resolves #1344

cytec:
I've added some comments... please note i've only skimmed through this quickly and it's pretty early so i may have missed some things... besides that, i have some more general things i'd like to adress:

im not really used to glide and have to say that i've never used it before... but i used godeps and if that is anyhow similar (which it looks for me) why not install glide via `go get ...` and then in building run something like `$(RUN) glide install` which should do basically the same as adding a new make file and stuff like that but would keep the Makefile more spksrc like... if that won't work for some reason please tell me, but afaik it should...

also either use `PKG_DOWNLOAD_METHOD` to directly checkout from git (see: https://github.com/SynoCommunity/spksrc/blob/master/cross/rtmpdump/Makefile#L3-L6) or use `PKG_DIST_FILE` (i think that's the correct one) to have a name that's not only the version (see: https://github.com/SynoCommunity/spksrc/blob/master/cross/tesseract-ocr/Makefile#L5)

maybe @Dr-Bean will have a look too

torrentalle:
I took these design decisions to solve some these problems:

The command `go get` always download the latest compatible version for each dependencies; this behavior causes non predictable builds, all build would be different. 

When I try to use dependency manager provided by Gogs project ([Glide](https://glide.sh/) or [GoPM](https://github.com/gpmgo/gopm)), I cannot install the dependencies without forcing dependency update, there are unresolved releases. In this case I have the same problem, all builds will be non predictable.

To solve this problem, I decided to parse the `glide.lock` file to generate corresponding Gihub URL's, to ensure that all builds from the dame package version will be equals.

This decision had consequences, each dependency must me extracted in a specific directory based on the package name. For this reason I don't use the provided extract methods, it's easier extract a package to a custom path than parse package name to extract the corresponding location.

I decided to implement this solution creating a new Makefile [`prepare-glide.mk`](https://github.com/torrentalle/spksrc/blob/gogs/cross/gogs/prepare-glide.mk) to make it reusable by other go projects.

All other ideas to solve this design problem are appreciated.

torrentalle:
Merged with current `master` branch and updated [Travis build link](https://travis-ci.org/torrentalle/spksrc/builds/315074251)

Issue #2530, opened 2016-11-20
Use predefined ARCHES before spksrc include


vkostas:
See also this topic [https://github.com/SynoCommunity/spksrc/pull/1662](url)... (Dr-Bean's comment: We should do something about the XYZ_ARCHES in common.mk, that's not working as intended imo. It's strange that it only works after the include, but not before.)

Random example based on a real spksrc cross packages.
`include ../../mk/spksrc.cross-cc.mk`
`ifeq ($(findstring $(ARCH), $(XYZ_ARCHES)),$(ARCH))`
`DEPENDS += ........`
`ENV += .........`
`CMAKE_ARGS += ...........`
`endif`
Any change of DEPENDS value after spksrc.cross-cc.mk is useless since all related code is "executed" in the latter.
I think the solution is to put common.mk before any reference to XYZ_ARCHES and then user the "normal" include spksrc.cross-cc.mk. (Yeah, that means common.mk is executed twice, but it has no other impact).
`include ../../mk/spksrc.common.mk`
`ifeq ($(findstring $(ARCH), $(XYZ_ARCHES)),$(ARCH))`
`......`
`endif`
`include ../../mk/spksrc.cross-cc.mk`

Any ideas?

Dr-Bean:
My original thought there was that include directives were read before further processing the rest of a Makefile, so $(ARCH) would have the proper value. That's not the case unfortunately...as we've seen, it actually matters where the include is located in the Makefile.

So, we might as well keep the approach of that random example you have there. Especially if adding the second include for common.mk is just to replace the other include as far as location is concerned. The location of the include obviously is less important than the functionality ;)

vkostas:
Thumbs up.
Then it would be wise to amend somehow the instructions (somewhere) so other people do not have to re-invent the wheel ;-)

Dr-Bean:
Cool. Feel free to open a PR to sync all the packages that use the approach to have the same layout, we could even keep this issue open for it to serve as a reminder.

Not sure how to describe it in terms of a instruction/wiki though...https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables would probably fit the best, but it doesn't use any type of order that we care about as far as I can tell. Maybe add the include directive for the cross packages stuff, and put it in there?

hgy59:
[...]

PR #2534, opened 2016-11-22
Correct usage of ARCH. Issue # 2530.

vkostas:
_Motivation:_ Issue [2530 - https://github.com/SynoCommunity/spksrc/issues/2530](https://github.com/SynoCommunity/spksrc/issues/2530)
_Linked issues:_ N/A

### Checklist


cross/alsa-lib/Makefile
```diff
@@ -13,7 +13,7 @@ LICENSE  =
 
 GNU_CONFIGURE = 1
 CONFIGURE_ARGS = --disable-python
-ifeq ($(findstring $(ARCH),88f5281 ppc853x ppc854x),$(ARCH))
+ifeq ($(findstring $(ARCH),$(PPC_ARCHES)),$(ARCH))
```
Dr-Bean: Missing `include ../../mk/spksrc.cross-cc.mk` before this line.
Can we agree to put the include after GNU_CONFIGURE (or LICENSE, if GNU_CONFIGURE isn't set), and before CONFIGURE_ARGS?
That approach keeps all the CONFIGURE_ARGS more or less in the same place.

cross/bison/Makefile
```diff
@@ -12,7 +12,10 @@ COMMENT  = General-purpose parser generator
 LICENSE  = GNU GPL
 
 GNU_CONFIGURE = 1
-ifeq ($(findstring $(ARCH),88f5281 powerpc ppc824x ppc853x ppc854x),$(ARCH))
+
+include ../../mk/spksrc.cross-cc.mk
```
Dr-Bean: The include on L24 should be removed.

cross/bitlbee/Makefile
```diff
@@ -11,6 +11,8 @@ HOMEPAGE = http://www.bitlbee.org/
 COMMENT  = BitlBee brings IM (instant messaging) to IRC clients. It\'s a great solution for people who have an IRC client running all the time and don\'t want to run an additional MSN/AIM/whatever client
 LICENSE  =
 
+include ../../mk/spksrc.common.mk
```
Dr-Bean: I thought we agreed that we didn't want to include common.mk, but that we'd use `include ../../mk/spksrc.cross-cc.mk` instead?

cross/ffmpeg-2.2.1/digests
```diff
@@ -0,0 +1,3 @@
+ffmpeg-2.2.1.tar.gz SHA1 cf455dd900cd7749f2c51c644054a9b657b8b589
```
Dr-Bean: While this may be necessary, it doesn't belong in this PR. Can you move it to a separate one?

cross/alsa-lib/Makefile
Dr-Bean: And line breaks in between. So, to be clear, the generic approach would look like this
```
LICENSE...

GNU_CONFIGURE ...if applicable

include ../...

CONFIGURE_ARGS = ...if applicable

ifeq($(findstring $(ARCH)...
  <thing>
endif

.PHONY etc. ...if applicable

vkostas: I was thinking of the same.
Agree!!!

cross/bitlbee/Makefile
vkostas: My fault. :-)

cross/ffmpeg-2.2.1/digests
vkostas: Ok

cross/bitlbee/Makefile
vkostas: hmmm. I think using **cross-cc.mk** before our variables at least for **TARGET_PATCH, TARGET_INSTALL, etc**.
That's why I applied added **common.mk** before the variable declarations used in **cross-cc.mk**.
I have to redo all changes again:
>LICENSE...
>
>GNU_CONFIGURE ...if applicable
>
>include ../../mk/spksrc.common.mk
>
>CONFIGURE_ARGS = ...if applicable
>
>ifeq($(findstring $(ARCH)...
>  <thing>
>endif
>
>include ../../mk/spksrc.cross-cc.mk
>
>.PHONY etc. ...if applicable
>

vkostas:
@Dr-Bean Just in case you have missed my reply .....
Maybe the solution is as simple as to include **common.mk** at the very beginning of cross Makefile (sometimes ARCH is needed to define dependencies).
>include ../../mk/spksrc.common.mk
PKG_NAME = 
PKG_VERS = 
PKG_EXT = 
...
DEPENDS
...
...

.PHONY
>

Dr-Bean:
Can't, because common.mk contains optional values (e.g. the `pip ?=` entry), you wouldn't be able to define those values correctly.

Besides, let's keep this simple. common.mk is included via cross-cc.mk. For the packages where we currently need access to values of `Available Arches` (that set of arches), we already know it works when we put the include above it. There should be a general order that we can establish to put the include for those particular packages, yes? Let's keep the other packages as is for now, because it's more work than it's worth imo :)

PR #2547, opened 2016-12-08
Updated SABnzbd-Testing to 1.1.1 Final

KaraokeStu:
As there is currently no testing version and the 1.1.1 final is ahead of 1.1.1 RC3, I've updated this to final

### Checklist

### DIFF
--- a/cross/sabnzbd-testing/Makefile
+++ b/cross/sabnzbd-testing/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = SABnzbd
-PKG_VERS = 1.1.1RC3
+PKG_VERS = 1.1.1
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS)-src.$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/sabnzbd/sabnzbd/releases/download/$(PKG_VERS)
--- a/cross/sabnzbd-testing/digests
+++ b/cross/sabnzbd-testing/digests
@@ -1,3 +1,3 @@
-SABnzbd-1.1.1RC3-src.tar.gz SHA1 147a3a254baa3a44ce506018a1e83fbb6932d19e
-SABnzbd-1.1.1RC3-src.tar.gz SHA256 1b855efe3a106f7bb0b5d3054ad1c4b1d5eaa50bd83d57aef8c152904a6f7e6f
-SABnzbd-1.1.1RC3-src.tar.gz MD5 cbdd395a2fb86f3e0851ebd6cb4eb05a
+SABnzbd-1.1.1-src.tar.gz SHA1 f0c83b91fca0d19febc04b46156297e14bec6c14
+SABnzbd-1.1.1-src.tar.gz SHA256 001f19c5a98de641da9fc69cff8ee885f3422d511a73e923dcb713aaca763d44
+SABnzbd-1.1.1-src.tar.gz MD5 917a105fbb594bf756551fcc37121d06
--- a/spk/sabnzbd-testing/Makefile
+++ b/spk/sabnzbd-testing/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = sabnzbd-testing
-SPK_VERS = 1.1.1RC3
-SPK_REV = 10
+SPK_VERS = 1.1.1
+SPK_REV = 11
 SPK_ICON = src/sabnzbd-testing.png
 DSM_UI_DIR = app
 BETA = 1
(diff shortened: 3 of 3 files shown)


Safihre:
@Dr-Bean / @KaraokeStu  
First wanted to say great to see so much activity again on the repo :)
For SABnzbd 1.3.0 that's coming after the 1.2.0 releases (beta's for that should start within the week), we will introduce a new Python C-extension that dramatically increases performance.
Is it possible to install pip packages for the SynoCommunity users? 
The extension is available on pip (`pip install sabyenc`) and only requires a basic C-compiler, no other extensions.

KaraokeStu:
I think we should be able to, @Dr-Bean can you confirm?

Dr-Bean:
It's possible, but it means we need to cross-compile Python to be able to create the extension.

So we can add `sabyenc` to `spk/sabnzbd/src/requirements.txt`, and add the string of Python stuff to `spk/sabnzbd/Makefile`, like this: `BUILD_DEPENDS = cross/python cross/setuptools cross/pip cross/wheel`
Rest should be handled by the framework :)

Dr-Bean:
Oh, and add `rsc:share/wheelhouse` to `spk/sabnzbd/PLIST`, we haven't got that in yet.

Safihre:
Great! And this also means that when users update, it also gets installed?

It's still a few months away I guess, but just wanted to make sure!

Dr-Bean:
Yep. It only adds to the build time of the package...which is unfortunate, but otherwise, 'meh' ;)

Dr-Bean:
@KaraokeStu While you're about, mind if I edit the sabnzbd-testing package to add DSM6 compatibility and release a set of packages in a day or two?

KaraokeStu:
Oh absolutely, please do - I'm running DSM 6 so can test it for you too :)

Dr-Bean:
Well, I'm actually hoping to find someone with a vanilla setup to test if the upgrade path works somewhat (package installed on DSM5, upgrade to DSM6, 'my' upgrade). I'm guessing you're not the right person for that :P
I'm reasonably sure we're going to end up recommending reinstalling packages in a lot of cases...

cytec:
so we have a DSM6 "default/support" approach now? yeah!

Dr-Bean:
Cool. Is the new bootloader able to work with DSM5, or do you have to switch over to upgrade? I still need to rebuild my VM's and my physical device, none of them are representative of an average device with the amount of times I bricked and revived them...

Anyway, can you try this: https://www.dropbox.com/s/7wzdtucopidahdj/sabnzbd-testing_x64-5.2_1.1.1-11.spk.
I brought it back to bare minimum wrt code changes, but if it survives an upgrade, then that's all we need. One thing's for sure. Synology's changes, besides being backwards incompatible have...quirks.

Safihre:
@Dr-Bean doesn't Synology work with you guys at all? Or are there projects like yours to install external things on a Synology NAS?

KaraokeStu:
BTW guys, just tried a new kernel and bricked my DSM box - trying to get it back now! :(

Dr-Bean:
@KaraokeStu No rush. I was thinking of adding an upgrade wizard and a step to the install wizard to explain to use `sc-download` for permissions handling, but I'm not finishing it today regardless.

@Safihre Well, at least from my point of view, Synology, as an organization, _tolerates_ third party package developers. I won't say I never see any response or results, because I have, but Synology certainly won't go out of their way to help us out. That applies to the whole chain: from DSM development, kernel files and toolchains to documentation and support.
I can't speak for other developers though, including other members of SynoCommunity, they may have different experiences altogether. /rant ;)

KaraokeStu:
Just so you know that, although my system is broken, I have spun up both a DSM 5.2 and DSM 6 instance for testing. The spk you supplied works correctly.

Dr-Bean:
Thanks. One more question: Can you provide the following from both instances (I had wanted to add the questions earlier, but it slipped my mind):
- Content of /etc/passwd wrt `sabnzbd-testing` and `sc-sabnzbd-testing`
- Content of /etc/shadow wrt `sabnzbd-testing` and `sc-sabnzbd-testing`
- Content of /etc/group wrt `sabnzbd-testing` and `sc-sabnzbd-testing`

Output of the following set of commands (some should not output anything or spit out an error):
```
synouser --get sabnzbd-testing
synouser --get sc-sabnzbd-testing
synogroup --get sabnzbd-testing
synouser --enum all | grep sabnzbd-testing
synogroup --get sc-sabnzbd-testing
synogroup --enum all | grep sabnzbd-testing
synogroup --enum all | grep sc-download
```

Safihre:
@KaraokeStu We just released 1.2.0Beta1. We don't use the pyOpenSSL anymore but cryptography, dunno if that makes a lot of difference :P

Issue #2142, opened 2016-02-17
Include syncthing-inotify in Syncthing


OrangeFoil:
Since I installed Syncthing on my NAS the hard drives don't power down on idle anymore because Syncthing is scanning the folders every 60 seconds to monitor changes.

So instead of relying on a periodical rescan it would be great if the Syncthing package also installed [syncthing-inotify](https://github.com/syncthing/syncthing-inotify), which is also developed and maintained by the Syncthing team.

> This application (syncthing-inotify) uses OS primitives to detect changes as soon as they happen. Therefore, if you save a file, syncthing-inotify will know about it and pass this information to Syncthing such that near real-time synchronisation can be achieved.
##

Dr-Bean:
Done with #2629 thanks to @o0v7r0o. Packages are published on the repository and should show up shortly.

Issue #2261, opened 2016-04-14
Provide a "CLI Essentials" SPK?


cytec:
Already mentioned here: https://github.com/SynoCommunity/spksrc/pull/2071 with all those CLI Tool PR's coming in lately that'll may be a good thing to do in the near future...

Personally i don't care if we provide a single SPK for each of them or just combine them all into one single SPK but i think it's a lot cleaner for the users and may be not as confusing for some of them (expecting that some may not know that this are cli tools and reporting "errors") 

Here is the list from the PR above as well as the new ones that came after that PR:
- nano [SPK Published](https://github.com/SynoCommunity/spksrc/tree/master/spk/nano)
- vim [SPK Published](https://github.com/SynoCommunity/spksrc/tree/master/spk/vim)
- htop [SPK Published](https://github.com/SynoCommunity/spksrc/tree/master/spk/htop)
- ldd
- bc [PR pending](https://github.com/SynoCommunity/spksrc/pull/2071)
- sudo
- nice (contained in Busybox?)
- screen 
- less [PR pending](https://github.com/SynoCommunity/spksrc/pull/2075)
- detox [PR pending](https://github.com/SynoCommunity/spksrc/pull/2199)
- ncdu [PR pending](https://github.com/SynoCommunity/spksrc/pull/2226)
- fdupes [PR pending](https://github.com/SynoCommunity/spksrc/pull/2257)

New:
- lsscsi [PR pending](https://github.com/SynoCommunity/spksrc/pull/2270)
- nmon [PR pending](https://github.com/SynoCommunity/spksrc/pull/2279)
- davfs2 [PR pending](https://github.com/SynoCommunity/spksrc/pull/1974)

there may be others but that's all i currently remember any thoughts @SynoCommunity/developers 

EDIT: maybe we could even add a busybox build and depend on that SPK for other SPK`s that need to use some of the busybox functions (a lot of python only apps use python's busybox already)
##

Dr-Bean:
I do want to point out that there's a difference between CLI-based software that is worth its own package, and CLI 'tools' that don't warrant a separate package (with a grey area in between). 
You feel the same, because you haven't added all the CLI-based packages by quite a margin ;)

There's a grey area where I'm not 100% convinced either way, but I'd say this is my current pov based on your list (it'd be longer if we actually added all CLI-based packages):
Separate package:
- nano
- vim

Combined:
- ldd
- bc
[...]

PR #2553, opened 2016-12-12
Update: Syncthing to 0.14.21, Go (dependence) to 1.7.4

o0v7r0o:
_Motivation:_ My aim was to share integration syncthing-inotify to the package, as I didn't like any other startup options. By the way, I also updated Syncthing and Go to the latest version, just to keep track recent changes in them.
_Linked issues:_ #2142 

### Checklist
- [OK] Build rule `all-supported` completed successfully
- [OK] Package upgrade completed successfully (tested at 88f6281, ds111)
- [OK] New installation of package completed successfully (tested at 88f6281, ds111)

### DIFF
--- a/spk/syncthing/Makefile
+++ b/spk/syncthing/Makefile
@@ -1,12 +1,12 @@
 SPK_NAME = syncthing
-SPK_VERS = 0.13.4
-SPK_REV = 9
+SPK_VERS = 0.14.21
+SPK_REV = 10
 SPK_ICON = src/syncthing.png
 DSM_UI_DIR = app
 
 UNSUPPORTED_ARCHS = $(PPC_ARCHES)
 
-DEPENDS  = cross/busybox cross/$(SPK_NAME)
+DEPENDS = cross/busybox cross/$(SPK_NAME)
 
 MAINTAINER = cytec
 DESCRIPTION = Automatically sync files via secure, distributed technology.
@@ -14,14 +14,13 @@ DESCRIPTION_FRE = Synchronisation automatique de fichiers via une technologie s√
 ADMIN_PORT = 7070
 RELOAD_UI = yes
 DISPLAY_NAME = Syncthing
-CHANGELOG = "Update to 0.13.4"
-
-HOMEPAGE   = http://www.syncthing.net
-LICENSE    = MPLv2.0
+CHANGELOG = "Update to 0.14.21"
+HOMEPAGE = http://www.syncthing.net
+LICENSE = MPLv2.0
 
 INSTALLER_SCRIPT = src/installer.sh
-SSS_SCRIPT       = src/dsm-control.sh
-FWPORTS          = src/${SPK_NAME}.sc
+SSS_SCRIPT = src/dsm-control.sh
+FWPORTS = src/${SPK_NAME}.sc
 
 INSTALL_PREFIX = /usr/local/$(SPK_NAME)
 
@@ -36,6 +35,7 @@ include ../../mk/spksrc.spk.mk
 syncthing_extra_install:
 	install -m 755 -d $(STAGING_DIR)/var
 	install -m 600 src/config.xml $(STAGING_DIR)/var/config.xml
+	install -m 644 src/options.conf $(STAGING_DIR)/var/options.conf
 	install -m 755 -d $(STAGING_DIR)/app
 	install -m 644 src/app/config $(STAGING_DIR)/app/config
 	install -m 755 -d $(STAGING_DIR)/app/images
--- a/spk/syncthing/src/dsm-control.sh
+++ b/spk/syncthing/src/dsm-control.sh
@@ -9,12 +9,17 @@ INSTALL_DIR="/usr/local/${PACKAGE}"
 PATH="${INSTALL_DIR}/bin:${PATH}"
 USER="syncthing"
 SYNCTHING="${INSTALL_DIR}/bin/syncthing"
-CONFIG_DIR="${INSTALL_DIR}/var/"
+CONFIG_DIR="${INSTALL_DIR}/var"
+SYNCTHING_OPTIONS="-home=${CONFIG_DIR}"
 
+# Read additional startup options from /usr/local/syncthing/var/options.conf
+if [ -f ${CONFIG_DIR}/options.conf ]; then
+	source ${CONFIG_DIR}/options.conf
+fi
 
 start_daemon ()
 {
-    start-stop-daemon -b -o -c ${USER} -S -u ${USER} -x env HOME=${CONFIG_DIR} ${SYNCTHING} -- --home ${CONFIG_DIR}
+    start-stop-daemon -b -o -c ${USER} -S -u ${USER} -x env HOME=${CONFIG_DIR} ${SYNCTHING} -- ${SYNCTHING_OPTIONS}
 }
 
 stop_daemon ()
@@ -25,7 +30,7 @@ stop_daemon ()
 
 daemon_status ()
 {
-    start-stop-daemon -K -q -t -u ${USER}
+    start-stop-daemon -K -q -t -u ${USER} -x ${SYNCTHING}
     [ $? -eq 0 ] || return 1
 }
 
--- a/cross/syncthing/Makefile
+++ b/cross/syncthing/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = syncthing
-PKG_VERS = 0.13.4
+PKG_VERS = 0.14.21
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-source-v$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/syncthing/syncthing/releases/download/v$(PKG_VERS)
@@ -9,7 +9,7 @@ DEPENDS = native/go
 
 HOMEPAGE = http://www.syncthing.net/
 COMMENT  = Syncthing replaces Dropbox and BitTorrent Sync with something open, trustworthy and decentralized. Your data is your data alone and you deserve to choose where it is stored, if it is shared with some third party and how ...
-LICENSE  =
+LICENSE  = MPL-2.0
 
 CONFIGURE_TARGET = nope
 COMPILE_TARGET = myComp
@@ -21,7 +21,7 @@ CGO_ENABLED=0
 # set this to 1 to create static linked binaries
 STATIC_BINARIES = 0
 
-# Define SYNCTHING_ARCH as per SyncThing standards
+# Define SYNCTHING_ARCH as per Syncthing standards
 ifeq ($(findstring $(ARCH),88f5281 88f6281 alpine armada370 armada375 armada38x armadaxp comcerto2k monaco),$(ARCH))
 SYNCTHING_ARCH = arm
 ENV += GOARM=5
--- a/cross/syncthing/digests
+++ b/cross/syncthing/digests
@@ -1,3 +1,3 @@
-syncthing-source-v0.13.4.tar.gz SHA1 35fa3d7810f80a6e00be99d410b0cb9d2aa7402f
-syncthing-source-v0.13.4.tar.gz SHA256 19e113fe00f8166f6811b01ffaa06a546884d767eea79bea9b7907a2a35ca67b
-syncthing-source-v0.13.4.tar.gz MD5 eabf036d050fe148f8c8feadd5c5fd08
+syncthing-source-v0.14.21.tar.gz SHA1 824175a895f86b5e86b055ba6041103dee1fef21
+syncthing-source-v0.14.21.tar.gz SHA256 e5fe49b85425679fec3d08127cd4fcf40e474ff70ccec587ee21f306c0c36085
+syncthing-source-v0.14.21.tar.gz MD5 b81c093bf3630d43ed23e6e19163ad82
--- a/native/go/Makefile
+++ b/native/go/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = go
-PKG_VERS = 1.6.2
+PKG_VERS = 1.7.4
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)$(PKG_VERS).src.$(PKG_EXT)
 PKG_DIST_SITE = https://storage.googleapis.com/golang
--- a/spk/syncthing/src/config.xml
+++ b/spk/syncthing/src/config.xml
@@ -1,25 +1,19 @@
-<configuration version="12">
+<configuration version="17">
     <gui enabled="true" tls="false">
         <address>0.0.0.0:7070</address>
     </gui>
     <options>
-        <listenAddress>tcp://0.0.0.0:22000</listenAddress>
+        <listenAddress>default</listenAddress>
         <globalAnnounceServer>default</globalAnnounceServer>
         <globalAnnounceEnabled>true</globalAnnounceEnabled>
         <localAnnounceEnabled>true</localAnnounceEnabled>
         <localAnnouncePort>21027</localAnnouncePort>
-        <relayServer>dynamic+https://relays.syncthing.net/endpoint</relayServer>
         <maxSendKbps>0</maxSendKbps>
         <maxRecvKbps>0</maxRecvKbps>
         <reconnectionIntervalS>60</reconnectionIntervalS>
         <relaysEnabled>true</relaysEnabled>
         <relayReconnectIntervalM>10</relayReconnectIntervalM>
-        <relayWithoutGlobalAnn>false</relayWithoutGlobalAnn>
         <startBrowser>false</startBrowser>
-        <upnpEnabled>true</upnpEnabled>
-        <upnpLeaseMinutes>60</upnpLeaseMinutes>
-        <upnpRenewalMinutes>30</upnpRenewalMinutes>
-        <upnpTimeoutSeconds>10</upnpTimeoutSeconds>
         <urAccepted>0</urAccepted>
         <urUniqueID></urUniqueID>
         <urURL>https://data.syncthing.net/newdata</urURL>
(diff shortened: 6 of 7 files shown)


spk/syncthing/src/dsm-control.sh
```diff
@@ -3,68 +3,94 @@
...
+SYNCTHING_INOTIFY="${INSTALL_DIR}/bin/syncthing-inotify"
+SYNCTHING_INOTIFY_OPTIONS="-home=${CONFIG_DIR}"
+# Include user configurable options
+. ${CONFIG_DIR}/syncthing-addoptions.cfg
 
 start_daemon ()
 {
-    start-stop-daemon -b -o -c ${USER} -S -u ${USER} -x env HOME=${CONFIG_DIR} ${SYNCTHING} -- --home ${CONFIG_DIR}
+    start-stop-daemon -b -o -c ${USER} -S -u ${USER} -x env HOME=${CONFIG_DIR} $1 -- $2
 }
```
cytec: uhm.... why?
why rework everything so it works completely different than any other package? i don't get it...

spk/syncthing/src/dsm-control.sh
```diff
@@ -3,68 +3,94 @@
...
             echo Starting ${DNAME} ...
-            start_daemon
+			start_daemon ${SYNCTHING} "${SYNCTHING_OPTIONS}"
         fi
+        if daemon_status ${SYNCTHING_INOTIFY}; then
+            echo ${DINAME} is already running
+        else
+            echo Starting ${DINAME} ...
+            start_daemon ${SYNCTHING_INOTIFY} "${SYNCTHING_INOTIFY_OPTIONS}"
+		fi
```
cytec: oh... now i get it... but.. im not a fan.
I would rather make a new Package for Inotify if its an own app! you can (if that's a must) still depend on syncthing in the SPK's Makefile. \cc @Dr-Bean what do you think?

cross/syncthing-inotify/Makefile
```diff
@@ -0,0 +1,60 @@
+PKG_NAME = syncthing-inotify
+PKG_VERS = 0.8.4
+PKG_EXT = tar.gz
+PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
```
cytec: add  `PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)` so the dist file get's a nicer name... also have to re run `make digests`

spk/syncthing/src/dsm-control.sh
o0v7r0o: yes, I agree that the script looks weird now. I tried to keep original variable names. I would rather rework it completely (get rid of 'SYNCTHING' in variables) to make it easy adaptable to run any two binaries simultaneously (e.g. 'DAEMON1' and 'DAEMON2').

spk/syncthing/src/dsm-control.sh
o0v7r0o: The point is that inotify makes no sense without syncthing running. Also it provides nearly must-have functionality if you run it on NAS, as you don't want to hang your CPU constantly rehashing shared folders. Moreover, I hope that syncthing maintainers will integrate inotify one day (https://github.com/syncthing/syncthing/pull/2807) making separate SPK for it obsolete.
Thinking about separate SPK, I would prefer to kick busybox out and make one SPK for it, then you don't have to place its binaries in nearly every package.

cross/syncthing-inotify/Makefile
o0v7r0o: okay, overlooked this option.

spk/syncthing/src/dsm-control.sh
o0v7r0o: First, I just wanted to run inotify on my NAS. As soon as it should use the same user account to read config files with '600' rights, its presence breaks your start-stop-status script here:
```
daemon_status ()
{
    start-stop-daemon -K -q -t -u ${USER}
    [ $? -eq 0 ] || return 1
}
```
So, I decided to dig a bit more...

Dr-Bean:
Putting this as a general remark, instead of in the code itself, so I can find it afterwards ;)

After looking at the PR, my first thought was to put inotify in a separate package as well. But I understand why you didn't. Not sure if I have a real preference at this point myself, it'd work both ways.

Having said that, instead of these changes, why just not run start-stop-daemon twice, like this:
```
start-stop-daemon -b -o -c ${USER} -S -u ${USER} -x env HOME=${CONFIG_DIR} syncthing -- $opts
start-stop-daemon -b -o -c ${USER} -S -u ${USER} -x env HOME=${CONFIG_DIR} inotify -- $opts
```
...same for stopping. Makes it a lot clearer what's going on than these changes imo.
Alternatively, if you really want to split up, just use two/four separate functions to start and stop, just change the function name from start_daemon () to start_syncthing () and start_inotify (), each with their own start call, same for stop respectively. No real need though, because as far as I can tell, you always want both of them running, no?

Lastly, since I happened to see it, moving busybox to a separate SPK doesn't work imo. I more or less explained it in #2261. Keeping packages self-contained makes the most sense in the long run, especially for something as low-level as useraccounts or daemonizing options (while DSM6 takes at least the first step to decrease the dependency altogether)

o0v7r0o:
@Dr-Bean, I made so first, but then I catched myself on copying and inserting the same code twice: separate start, stop, status and wait_for_status functions for every daemon.

o0v7r0o:
@Dr-Bean, made separate functions for start/stop/etc.

spk/syncthing/src/dsm-control.sh
```diff
@@ -44,34 +75,53 @@ wait_for_status ()
...
             echo ${DNAME} is not running
             exit 1
         fi
+
+        if syncthing_inotify_daemon_status; then
+            echo ${DNAME_INOTIFY} is running
+            exit 0
+        else
+            echo ${DNAME_INOTIFY} is not running
+            exit 1
```
Dr-Bean: Have you tested that this works with all variations? If, for example, Syncthing is _not_ running, but Inotify _is_, then I think you'll just see `Syncthing is not running`...and never see `Syncthing-Inotify is running`, because the `exit` on L111 would prevent further processing, right?
I'd also like to know that Package Center is able to correctly report the package status (where we probably want to set the status based off of Syncthing itself, not of Inotify)

spk/syncthing/src/dsm-control.sh
```diff
@@ -44,34 +75,53 @@ wait_for_status ()
...
+        fi
         ;;
     log)
         exit 1
         ;;
     *)
         exit 1
-        ;;
```
Dr-Bean: I do think we want this line back...

o0v7r0o:
@Dr-Bean you were right, having two executables in one script somehow causes strange behaviour executing it from DSM Packet Manager. I made separate spk for inotify trying to preserve as much as possible. Hope that it will fit to the concept.

Checklist for new syncthing-inotify package:

[ to go ] Build rule all-supported completed successfully
[OK] Package upgrade completed successfully (tested at 88f6281, ds111)
[OK] New installation of package completed successfully (tested at 88f6281, ds111)

o0v7r0o:
@cytec @Dr-Bean it looks abandoned, no?

Tstassin:
Syncthing-inotify is an important part of syncthing, I hope this isn't abandoned ! Fingers crossed..

cytec:
sorry i'm a little bit busy with RL stuff lately... maybe @Dr-Bean can look into this again? I just looked through the code really quick and it looked good to me... have you verified that updating from previous Syncthing versions works as excepted? since i saw the config file changed a bit?

also normaly we have a one package per PR policity but as @Dr-Bean already said that might be ok in this particular case... still recommend a squash

cross/syncthing-inotify/Makefile
```diff
@@ -0,0 +1,61 @@
...
+
+LDFLAGS = -w -X main.Version=v$(PKG_VERS)
+
+myComp:
+	cd $(SYNCTHING_INOTIFY_DIR) && env $(ENV) go get && env $(ENV) go build -ldflags '$(LDFLAGS)'
+
+myInstall:
+	mkdir -p $(STAGING_INSTALL_PREFIX)/bin
+	mv $(SYNCTHING_INOTIFY_BIN)-$(PKG_VERS) $(SYNCTHING_INOTIFY_BIN)
+	install -m 755 $(SYNCTHING_INOTIFY_BIN) $(STAGING_INSTALL_PREFIX)/bin/
```
Dr-Bean: Is this necessary? We don't do it for the Syncthing package, and this is probably as near a copy of it as you can get.

spk/syncthing-inotify/Makefile
```diff
@@ -0,0 +1,37 @@
+SPK_NAME = syncthing-inotify
+SPK_VERS = 0.8.2
+SPK_REV = 1
+SPK_ICON = src/${SPK_NAME}.png
+BETA = 1
+
+UNSUPPORTED_ARCHS = $(PPC_ARCHES)
+
+DEPENDS = cross/busybox cross/$(SPK_NAME)
```
Dr-Bean: Can you use Syncthing's busybox binary instead? That means you should put `/usr/local/syncthing/bin` on $PATH of inotify's dsm-control.sh, but that cleans up this package a bit (drop `cross/busybox` here and drop L29-30)

spk/syncthing-inotify/Makefile
```diff
@@ -0,0 +1,37 @@
...
+SPK_DEPENDS = syncthing
+START_DEP_SERVICES = syncthing
```
Dr-Bean: Just to check: Syncthing _has_ to run, or no?

spk/syncthing-inotify/Makefile
```diff
@@ -0,0 +1,37 @@
...
+
+MAINTAINER = SynoCommunity
+DESCRIPTION  = Syncthing uses a rescan interval to detect changes in folders. Syncthing-inotify uses OS primitives to detect changes as soon as they happen. Therefore, near real-time synchronisation can be achieved.
+RELOAD_UI = no
+DISPLAY_NAME = Syncthing-Inotify
+CHANGELOG = "Initial release"
+
+HOMEPAGE   = https://github.com/syncthing/syncthing-inotify
```
Dr-Bean: Remove unnecessary spaces and optionally, align entries with each other.

spk/syncthing-inotify/Makefile
```diff
@@ -0,0 +1,37 @@
...
+BETA = 1
+
+UNSUPPORTED_ARCHS = $(PPC_ARCHES)
+
+DEPENDS = cross/busybox cross/$(SPK_NAME)
+SPK_DEPENDS = syncthing
+START_DEP_SERVICES = syncthing
+
+MAINTAINER = SynoCommunity
+DESCRIPTION  = Syncthing uses a rescan interval to detect changes in folders. Syncthing-inotify uses OS primitives to detect changes as soon as they happen. Therefore, near real-time synchronisation can be achieved.
```
Dr-Bean: Double space between `DESCRIPTION` and `=`
Maybe start with `File watcher intended for use with Syncthing.`, then append the current text?

spk/syncthing-inotify/src/dsm-control.sh
```diff
@@ -0,0 +1,83 @@
...
+SYNCTHING_DIR="/usr/local/syncthing"
+PATH="${INSTALL_DIR}/bin:${PATH}"
+USER="syncthing"
+INOTIFY="${INSTALL_DIR}/bin/syncthing-inotify"
+CONFIG_DIR="${SYNCTHING_DIR}/var"
+INOTIFY_OPTIONS="-home=${CONFIG_DIR}"
+
+# Read additional startup options from /usr/local/syncthing-inotify/var/options.conf
+if [ -f ${INSTALL_DIR}/var/options.conf ]; then
+	source ${INSTALL_DIR}/var/options.conf
```
Dr-Bean: Where is `options.conf` located? In the inotify package, or in the Syncthing package. Seems you're mixing the two up ;)

spk/syncthing-inotify/src/installer.sh
```diff
@@ -0,0 +1,71 @@
...
+{
+    exit 0
+}
+
+postinst ()
+{
+    # Link
+    ln -s ${SYNOPKG_PKGDEST} ${INSTALL_DIR}
+
+    # Install busybox stuff
```
Dr-Bean: See earlier comment, if the busybox dep can be removed, this can go too.

spk/syncthing-inotify/Makefile
```diff
@@ -0,0 +1,37 @@
...
+LICENSE    = MPL-2.0
+
+INSTALLER_SCRIPT = src/installer.sh
+SSS_SCRIPT       = src/dsm-control.sh
+
+INSTALL_PREFIX = /usr/local/$(SPK_NAME)
+
+POST_STRIP_TARGET = inotify_extra_install
+
+BUSYBOX_CONFIG = daemon nice
```
Dr-Bean: See earlier comment, if the busybox dep can be removed, these two lines can go too.

spk/syncthing/Makefile
```diff
@@ -36,6 +36,7 @@ include ../../mk/spksrc.spk.mk
 syncthing_extra_install:
 	install -m 755 -d $(STAGING_DIR)/var
 	install -m 600 src/config.xml $(STAGING_DIR)/var/config.xml
+	install -m 644 src/options.conf $(STAGING_DIR)/var/options.conf
```
Dr-Bean: This seems unused, see an earlier comment.

spk/syncthing/src/dsm-control.sh
```diff
@@ -9,12 +9,17 @@ INSTALL_DIR="/usr/local/${PACKAGE}"
 PATH="${INSTALL_DIR}/bin:${PATH}"
 USER="syncthing"
 SYNCTHING="${INSTALL_DIR}/bin/syncthing"
-CONFIG_DIR="${INSTALL_DIR}/var/"
+CONFIG_DIR="${INSTALL_DIR}/var"
+SYNCTHING_OPTIONS="-home=${CONFIG_DIR}"
 
+# Read additional startup options from /usr/local/syncthing/var/options.conf
```
Dr-Bean: Let's keep inotify stuff in the inotify package. This seems unnecessary (as do most changes in this file).

Dr-Bean:
Well, if I see a commit after the last comment, I automatically assume the PR is not done yet and skip it...

Please take a look at my comments, and if you could, please re-squash the PR into two main commits. I'd say one to update Syncthing itself, one for adding Inotify. Thanks.

spk/syncthing-inotify/src/dsm-control.sh
o0v7r0o: They both have options.conf to allow user to append an additional option to syncthing **and** syncthing-inotify startup script. Moreover, these files will be backed up and restored while upgrading as the whole "var" directory.

spk/syncthing-inotify/Makefile
o0v7r0o: Yes, as soon it depends on syncthing, it can use its busybox.

spk/syncthing-inotify/Makefile
o0v7r0o: fixed

spk/syncthing-inotify/Makefile
o0v7r0o: fixed

spk/syncthing-inotify/Makefile
o0v7r0o: fixed

spk/syncthing-inotify/Makefile
o0v7r0o: In general, yes. Otherwise syncthing-inotify will generate a lot of stuff like:

> [WARNING] 23:38:46 Failed to perform request Get http://192.168.0.33:7070/rest/events?since=0: dial tcp 192.168.0.33:7070: getsockopt: connection refused
[WARNING] 23:38:51 Failed to perform request Get http://192.168.0.33:7070/rest/events?since=0: dial tcp 192.168.0.33:7070: getsockopt: connection refused
[WARNING] 23:38:56 Failed to perform request Get http://192.168.0.33:7070/rest/events?since=0: dial tcp 192.168.0.33:7070: getsockopt: connection refused
[WARNING] 23:39:01 Failed to perform request Get http://192.168.0.33:7070/rest/events?since=0: dial tcp 192.168.0.33:7070: getsockopt: connection refused
[WARNING] 23:39:06 Failed to perform request Get http://192.168.0.33:7070/rest/events?since=0: dial tcp 192.168.0.33:7070: getsockopt: connection refused
[WARNING] 23:39:11 Failed to perform request Get http://192.168.0.33:7070/rest/events?since=0: dial tcp 192.168.0.33:7070: getsockopt: connection refused
[WARNING] 23:39:16 Failed to perform request Get http://192.168.0.33:7070/rest/events?since=0: dial tcp 192.168.0.33:7070: getsockopt: connection refused

it will try to connect to syncthing every 5 seconds. Does a user need that?

spk/syncthing-inotify/Makefile
o0v7r0o: fixed

spk/syncthing-inotify/src/installer.sh
o0v7r0o: fixed

spk/syncthing-inotify/Makefile
Dr-Bean: As I said, just checking. Makes sense to keep it ;)

spk/syncthing-inotify/src/dsm-control.sh
Dr-Bean: Ok, then a default options.conf should be added for the Syncthing package, because currently, there isn't one (or I'm missing it completely)

cross/syncthing-inotify/Makefile
o0v7r0o: If you mean the "mv ..." line, that's because syncthing-inotify compiles into "syncthing-inotify-0.8.4" binary, this is just to rename it. It could be also placed in the compile target before.
Otherwise, I didn't get the point.

Dr-Bean: Right, my brain wasn't working, apologies.

You can probably combine the two into one line: `install -m 755 $(SYNCTHING_INOTIFY_BIN)-$(PKG_VERS) $(STAGING_INSTALL_PREFIX)/bin/$(SYNCTHING_INOTIFY_BIN)`

o0v7r0o:
@cytec, yes, you're right, the config is changed a bit, but it will not affect updating, because update script overwrites the config with current user config (as the whole 'var' folder). So if the config is old, syncthing itself will automatically update it to the current version and create a backup.
@Dr-Bean okay, I've changed things you mentioned. If everything looks good now, I will squash commits and create a separete PR for inotify.

Dr-Bean:
Ok. Haven't tested, and apart from some minor nits that I won't mention, no additional comments from me. Go ahead :)

o0v7r0o:
@Dr-Bean, now squashed. Despite syncthing dropped a strict requirement of Go 1.7 somewhen, I left newer version of Go as a separate commit.

Issue #2554, opened 2016-12-13
cross compiling python modules for mitmproxy


cytec:
@Dr-Bean @Diaoul i could need some support. I've recently started working on a `mitmproxy` SPK and therefore need to cross compile some python modules which require some native libraries... namely: `cryptography`, `Pillow`, `pyOpenSSL`, `ruamel.yaml` and `brotlipy` it seems like the first 3 already come with the Python SPK so i just skip those... 

however `brotlipy` seems to use `cffi` which simply won't compile at all... i've searched the web and found out that @Diaoul hat the same issue with `cryptography`

So what i've done now:
* created a new SPK and added `python` to `BUILD_DEPENDS`
* created a cross makefile for `ruamel.yaml` and `brotlipy` with `include ../../mk/spksrc.python-module.mk` based on `cross/cryptography`

I've got some errors when building the new cross packages, so i've installed both of them locally too. idk if that fixes it when moving the SPK to my DS. Next question is... if i require those cross modules do i still need to provide the wheels for those? if so... how? 

Any help would be appreciated since i have a feeling like i'm going nuts :D
My first guess was: hey ... why not just add those two (`ruamel.yaml` and `brotlipy`) to the python SPK in the first place... but again, also not sure if and how that would work...

cytec:
Update: after reading https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-solutions again, i've changed the includes in `cross/ruamel.yaml` and `cross/brotlipy` to `include ../../mk/spksrc.python-wheel.mk` which then fails when compiling `cross/brotlipy` giving the error: `ImportError: /home/spksrc/spksrc/spk/mitmproxy/work-x86-5.2/install/usr/local/mitmproxy/lib/python2.7/site-packages/_cffi_backend.so: wrong ELF class: ELFCLASS64` which is what i remembered @Diaoul came across when building `cryptography`...

Dr-Bean:
That doc is severely outdated, especially if you're on an updated fork: I've made a couple of changes.
It should be easy enough now. Take the Mercurial package as an example. Add those two projects to src/requirements.txt, and start compiling. The 'new' reason to create cross packages for Python stuff is either because the project needs patches, or because it depends on other cross-compiled packages (and other exceptions are when the project internally does funky stuff wrt dependency handling which it then shouldn't)

cytec:
[...]

PR #2557, opened 2016-12-14
add mitmproxy spk

cytec:
_Motivation:_ make mitmproxy available on Synology Devices.
_Linked issues:_ https://github.com/SynoCommunity/spksrc/issues/2554

### Checklist


cross/ruamel/Makefile
```diff
@@ -0,0 +1,14 @@
+PKG_NAME = ruamel.yaml
+PKG_VERS = 0.13.4
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://files.pythonhosted.org/packages/source/r/$(PKG_NAME)
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS = cross/libyaml
+
+HOMEPAGE =
```
Dr-Bean: We could use some info here ;)

spk/mitmproxy/Makefile
```diff
@@ -0,0 +1,29 @@
+SPK_NAME = mitmproxy
+SPK_VERS = 0.18.2
+SPK_REV = 1
+SPK_ICON = src/mitmproxy.png
+
+WHEELS = src/requirements.txt
```
Dr-Bean: To keep the flow in the file, should we perhaps format it as BUILD_DEPENDS, DEPENDS, WHEELS, SPK_DEPENDS, essentially in the order that the entries will be processed?

spk/mitmproxy/src/installer.sh
```diff
@@ -0,0 +1,114 @@
...
+PYTHON_DIR="/usr/local/python"
+PATH="${INSTALL_DIR}/bin:${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:${PATH}"
+USER="mitmproxy"
+GROUP="users"
+VIRTUALENV="${PYTHON_DIR}/bin/virtualenv"
+TMP_DIR="${SYNOPKG_PKGDEST}/../../@tmp"
+
+SERVICETOOL="/usr/syno/bin/servicetool"
+
+syno_group_create ()
```
Dr-Bean: This function can be removed

spk/mitmproxy/src/installer.sh
```diff
@@ -0,0 +1,114 @@
...
+    # Create syno group (Does nothing when group already exists)
+    synogroup --add ${SYNO_GROUP} ${USER} > /dev/null
+    # Set description of the syno group
+    synogroup --descset ${SYNO_GROUP} "${SYNO_GROUP_DESC}"
+
+    # Add user to syno group (Does nothing when user already in the group)
+    addgroup ${USER} ${SYNO_GROUP}
+}
+
+syno_group_remove ()
```
Dr-Bean: ...as can this one

spk/mitmproxy/src/installer.sh
```diff
@@ -0,0 +1,114 @@
...
+PACKAGE="mitmproxy"
+DNAME="mitmproxy"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+SSS="/var/packages/${PACKAGE}/scripts/start-stop-status"
+PYTHON_DIR="/usr/local/python"
+PATH="${INSTALL_DIR}/bin:${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:${PATH}"
+USER="mitmproxy"
+GROUP="users"
```
Dr-Bean: I belive the `users` group isn't correct here. Does the package need file system access? For DSM5, we  generally use the `nobody` group instead.

FTR, DSM6 users will not be member of the `users` group at all (we'll just use our `sc-*` groups if access to the FS is needed), and one thing to check if if there's even a `nobody` group.

spk/mitmproxy/src/requirements.txt
```diff
@@ -0,0 +1,38 @@
+https://github.com/mitmproxy/mitmproxy/archive/v0.18.2.tar.gz
+backports.ssl_match_hostname>=3.5.0.1, <3.6
```
Dr-Bean: All the versions in this file should be pinned to exactly 1 version (`==`) for reproducability reasons, otherwise we will not be able to recreate the exact same package when even 1 of these requirements updates. 
Run a `pip freeze > requirements.txt` when you've got everything installed, and use that as a base instead.

spk/mitmproxy/src/installer.sh
ymartin59: On DSM6, I can confirm `nobody` group is available. On my system, `domoticz` account belongs to that group. I also can mention `subliminal` is in `users` group.

Dr-Bean: @ymartin59 Just to check: are you sure the `nobody` group is available on a clean DSM6 environment, and not that it was created by one of our packages you have installed?
As for the `subliminal` user, yes, that's expected if you installed the user while on DSM5. It's not a DSM6 package...none of our packages are.

PR #2583, opened 2017-01-05
Update zlib PKG_DIST_SITE to new URL

Grimthorr:
_Motivation:_ The zlib website stores all package archives at http://zlib.net/fossils/ - so version 1.2.8 is located at http://zlib.net/fossils/zlib-1.2.8.tar.gz, not http://zlib.net/zlib-1.2.8.tar.gz (error 404 as this is no long the most recent version). This pull request updates the cross/zlib makefile to point to the new URL to avoid wget & build failures.
_Linked issues:_ Not applicable.

### Checklist

### DIFF
--- a/cross/zlib/Makefile
+++ b/cross/zlib/Makefile
@@ -2,7 +2,7 @@ PKG_NAME = zlib
 PKG_VERS = 1.2.8
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = http://zlib.net
+PKG_DIST_SITE = http://zlib.net/fossils
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 DEPENDS =


Dr-Bean:
Instead of just changing the URL, zlib should be updated to 1.2.10. URL will stay the same then (http://zlib.net/zlib-1.2.10.tar.gz).
I have it staged locally, just didn't push it out yet.

Grimthorr:
That makes sense, but it might also be worth changing the URL as all versions are stored at http://zlib.net/fossils (including the latest). Therefore this will stop wget failing each time zlib push out another release unbeknown to us.

Dr-Bean:
Well, zlib indicates that the official download site is the mentioned URL, or to use an SF mirror. Best to use either of those. It might otherwise bring additional load/hosting costs to zlib in a place they don't expect (e.g. could be a sponsored host/behind cloudflare etc)

Seeing as the update frequency of zlib is very low (previous update was three years ago or so?), it shouldn't be that big of a deal, and we'll discover and fix it soon enough anyway. Makes sense?

Grimthorr:
Ok, fine by me.

Dr-Bean:
Done with 1806c4f4c3c18aee708c1c691a81e00d8d530ef9. Thanks for taking the time to contribute though 👍

Grimthorr:
Perfect, thanks! No problem 👍.

Issue #2215, opened 2016-03-27
Rutorrent and DSM6 compatibility


iglooenergyreferral:
Hi,

I seem to be having these issues when seeting up rutorrent, can anyone help?

[27.03.2016 05:42:19] WebUI started.
[27.03.2016 05:42:22] rTorrent user must have read/execute access to the tmp directory. ruTorrent will not work. (/var/services/tmp/)
[27.03.2016 05:42:22] rss: Some functionality will be unavailable. Webserver user can't access external program (curl).
[27.03.2016 05:42:22] rutracker_check: Plugin will not work. Webserver user must have execute access to the rtorrent session directory (/usr/local/rutorrent/var/.session/).
[27.03.2016 05:42:23] Unpack plugin: rTorrent user can't access 'unzip' program.
##

Dr-Bean:
Does `WM6` in the title actually mean `DSM6` by any chance?

christopherjamesgunn:
Hi Dr Mean,

I wrote the article when I was quite tired, I meant DSM6 yes

nex6:
_Model and arch of your NAS_. DS415Play - evansport x15

_DSM version_; 6.0-7321 Update 1 2016/04/14

_Provide log files if available:_ No log files available

_If the package doesn't start, try to start the package via the command line and provide the output:_"Starting ruTorrent ..." Checking status directly after generates: "ruTorrent is not running"

Have had some problems with SSH service not possible to start so I re-installed DSM today, installed Webstation and then directly tried to install ruTorrent from the Synocommunity package, after installation package will not run, is doing something for a couple of seconds and then stopped again. Everything else working as it should.

First try to start manually, I got an error **"unknown user rutorrent"** so I created a rutorrent user and when trying again I get message **"Starting ruTorrent ...**" but nothing happens..

Would really appreciate some assistance, not a complete n00b but pretty close :)

mal68:
Hey

So I have been trying to track this issue as I would like to update my DS415+ to 6.0 but I don't want to do it unless this package will work. I thought I read somewhere the other day that the tool chains have been released, is this true? If so does that get us closer to a working package for DSM 6.0?

dolphinkite:
ruTorrent works on my DS415+ under DSM6.0

dancingnzb:
Hey guys, I'm on ds1815+ DSM 6.0.1.7393,

I installed rutorrent beta and webserver but i keep getting those errors.:

[...]

PR #2591, opened 2017-01-09
Add Web Station/PHP5.6 as dependency

A----:
_Motivation:_ Add required dependencies to ruTorrent
_Linked issues:_ #2215 

### Checklist

Can't `make all` on a Windows host?
```
debootstrap --foreign --arch powerpc jessie /spksrc/spk/debian-chroot/work/staging/var/chroottarget "http://ftp.debian.org/debian"
mknod: '/spksrc/spk/debian-chroot/work/staging/var/chroottarget/test-dev-null': Operation not permitted
E: Cannot install into target '/spksrc/spk/debian-chroot/work/staging/var/chroottarget' mounted with noexec or nodev
Makefile:55: recipe for target 'debian-chroot_extra_install' failed
```

Sorry :(

### DIFF
--- a/spk/rutorrent/src/conf/PKG_DEPS
+++ b/spk/rutorrent/src/conf/PKG_DEPS
@@ -1,0 +1,5 @@
+[WebStation]
+dsm_min_ver=6.0
+
+[PHP5.6]
+dsm_min_ver=6.0


Dr-Bean:
> Can't make all on a Windows host?

I don't understand what this means. See the README for instructions on how to set up a development environment.

Dr-Bean:
This also needs a change to the Makefile. See https://github.com/SynoCommunity/spksrc/blob/master/spk/cops/Makefile#L23, otherwise the dir won't be included. Hence the request to chekc of the package compiles and works correctly :)

A----:
I tried, but doing a `make all` in the container tries to do a `mknod /spksrc/spk/debian-chroot/work/staging/var/chroottarget/test-dev-null`, which it cannot do since its a docker volume and the underlying FS does not support this (Windows, non-hyper-V).

At least that's my guess.

Dr-Bean:
This is a PR for rutorrent, no? Why run a `make all` in the first place? It has a very specific use, and you'll generally have no need to run that command.
Enter the rutorrent dir, and run `make all-supported` instead.

Dr-Bean:
After re-reading your comment, I think I understand the confusion. Take a look at this link for some information on how spksrc works: https://github.com/SynoCommunity/spksrc/wiki/Developers-HOW-TO

A----:
Thanks, I did not find this doc. Currently downloading things!

A----:
The docker method cannot work on a Windows host, when doing a `make all-supported` in the package folder. It tries to create sym and hard-links is the mounted volume `/spksrc`, which it does not know how to handle.
I copied `/spksrc` to another folder insisde the container and its now building.

I'll add the conf and rebuild if it all goes right.

bamtan:
Do you know how to get ruTorrent running for DSM6?

A----:
I'm sorry, but I can't make it work on my Docker environment. It's still the old Docker Toolbox, not the Docker for Windows. Sorry :man_shrugging: 

@bamtan check #2215 . I made it work

Dr-Bean:
@A---- I'm not sure how to interpret your comment, but assuming you're talking about not getting (whatever flavor of) Docker on a Windows platform to work for compiling: Try using the VM approach instead: https://github.com/SynoCommunity/spksrc#virtual-machine. Imo, that's still the preferred method, because of the complete isolation of the host. Even using Docker on *nix has downsides for cross-compiling.

Also, if Docker for Windows doesn't work in general, then we should probably add that info to the README. I haven't a clue, I don't use Windows myself, so I'm going off of your issue description here ;)

bamtan:
@A---- is there anywhere I can contact you? i'd love some help on setting ruTorrent up on a DS216j (no Docker support).

A----:
@Dr-Bean I'm not sure that it does not work, someone would have to confirm that.
@bamtan Please check the issue #2215, it has all the information I have on the subject. Namely, install Webstation prior to ruTorrent and how to pass the PATH env variable for it to work.

bamtan:
@A----  I still get Error 403 when trying to access ruTorrent, could you write a tutorial on setting this up? Reading through #2215 is quite confusing for me. I'd appreciate it a lot. Are you on IRC by any chance? Perhaps I can contact you on there for assistance?

PR #2606, opened 2017-01-17
Update nasm digests for 2.12.02

### LINKED ISSUE #2605: Update nasm digests for 2.12.02
Pull request coming

ymartin59:
Fixes #2605

_Motivation:_ nasm build failed
_Linked issues:_ #2605


Dr-Bean:
@ymartin59 Please open against the master branch at the very least. The dsm6 branch is for testing purposes only at this point.

Dr-Bean:
After considering this, we can't merge this PR unfortunately. Otherwise, keeping the dsm6 branch in sync with master is going to be impossible. As it stands, only DSM6-specific PR's should be made against the dsm6 branch, all other improvements should be made against the master branch.
I'm keeping this open for a bit, until I have time to make the changes to master myself (or unless you open a PR)
Thanks :)

ymartin59:
Hello Dr-Bean
I understand. To provide a PR for master, I would like to know what set of build/test cases I should run to detect impacts of such dependency updates.

Dr-Bean:
Thanks for understanding. Wrt testing: Good question :) It depends on the contents of the PR. 
For complete SPK's, there are three requirements, which are mentioned in the checklist [here](https://github.com/SynoCommunity/spksrc/blob/master/.github/PULL_REQUEST_TEMPLATE.md#checklist).

We (currently) don't have a checklist specific to cross packages, but if we did, it would look something like:
 - (Succesfully) completing rule `make all-archs` for the cross package (Exceptions to be managed via `CONFIGURE_ARGS`/overriding targets / patching / `UNSUPPORTED_ARCHS`, more or less in that order, legacy arches may be skipped);
- Compiling and testing an SPK which includes the cross package as a dependency, preferably _not_ x86-64 based.

For native-only packages, like `nasm`, the situation is different again: native packages only need to be compiled on the host (so x86-64). Generally, a cross package will depend on it, so that can be testing together with the native package. So that would look like this:
- Succesfully run `make` for the native package;
- Succesfully compile a non-x86-based arch, e.g. `make arch-88f6281` for depending cross package.

...Might not be a bad idea to put this in a wiki or something actually.

Having said all that: PR's that update incorrect digests will always be accepted, and generally speaking, 'patch' updates, e.g. the zlib PR you created, won't have backwards-incompatible changes. Still, testing is never bad :)

ymartin59:
Thanks having written all that details. I submit new PRs to master

PR #2615, opened 2017-01-22
Fix Transmission startup function

akalin:
Pass in "-s /bin/sh" explicitly so that it still works even if
the transmission user has a different shell, e.g. /sbin/nologin.

I found that I've had to implement the workaround described
in step 5 of https://forum.synology.com/enu/viewtopic.php?t=87386#p330119
to get the Transmission package to start up correctly. However,
this lasts only until the next reboot, which wipes out the changes
to /etc/passwd.

But instead, it suffices to just pass in "-s /bin/sh" directly to
the start command in dsm-control.sh. Keeping the default
home directory doesn't seem to hurt.

### DIFF
--- a/spk/transmission/src/dsm-control.sh
+++ b/spk/transmission/src/dsm-control.sh
@@ -15,7 +15,7 @@ PID_FILE="${INSTALL_DIR}/var/transmission.pid"
 
 start_daemon ()
 {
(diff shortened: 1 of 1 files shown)


Dr-Bean:
Thanks, but I can't merge this unfortunately.
This doesn't make the package DSM6 compatible: more changes are needed for that (userhandling, permissions management etc). I'm working on that in the dsm6 branch. 
Because I'm working on multiple packages at once, I haven't pushed out my changes yet: Please be patient for a bit until I complete the majority of the work :)

Issue #2140, opened 2016-02-16
Syncthing umask issue?


gotson:
Hi guys,

i installed Syncthing, but all files replicated on the Nas from another device, when file permissions are ignored in Syncthing, are not created with the correct permissions.

Files/folders get created with permissions `-rw-r--r--` / `drwxr-xr-x`, but i expect them to be created with `-rw-rw-r--` / `drwxrwxr-x`
Note that the user:group of the files is correct (syncthing:users)

I tried adding a `umask 002` in the `.profile` file of the user `syncthing`, but it doesn't seem to be taken into account when syncthing is running, but it is applied when i log in using syncthing user and create new files.
Hence i suppose the startup script provided in the package does not make use of the `.profile` file.

Issue also mentionned on Syncthing forums, but not really a Syncthing issue:
https://forum.syncthing.net/t/umask-on-synology/2499

I looked around also in the various issues on github, and also on the other packages installed i use:
- it looks like `start-stop-daemon` from `busybox`does not support `umask`
- `transmission`has a umask setting in its app config, so it works fine
- `btsync`is a Synology package, and doesn't use `start-stop-daemon`, but runs as `admin` via `su` and set the proper `umask`at startup
- `couchpotato-custom` has a specific `chmod`setting in its config, so folders created have correct permissions, and files inside are under `transmission` user, so are correct too
- `sonarr` has an option to `chmod`/`chown`files/folders, and is thus fine
- some people on GH says that the Unix rights should not be used, and that Synology ACL are used instead. However my user account has full rights on the share, but i cannot edit some files/folders created by `syncthing` user, so i guess Unix rights are used after all.

Does anyone have the issue, is there any possible workaround ?
Changing the `start-stop-daemon` to something else would do the trick, is it something you plan to do to fix those issues ?
##

Dr-Bean:
I don't think there's much we (SC) can do here. We have to use busybox' `start-stop-daemon`, seeing as Syncthing does not provide a daemon itself (as opposed to BTSync), and there are no available alternatives that allow us to manage the package otherwise.

All the other options you mention should be built into the software itself. The only option that Syncthing has is `Ignore permissions`. Have you tried using that?

gotson:
I am using IgnorePermissions already.

Why do you have to use  busybox' `start-stop-daemon` exactly ?
Isn't it possible to replicate the behavior of the command with more simpler shell commands, and handle `umask` at the same time ?

Dr-Bean:
[...]

PR #2646, opened 2017-02-06
Syncthing: patch busybox to get additional start-stop-daemon options and set umask

o0v7r0o:
_Motivation:_ By default syncthing uses umask 022, that leads to creating files which are write-protected for everybody except syncthing itself. By setting umask to 002, it will create files with rights to write for users in *users* group. This will be helpful if one has Windows machines syncing with the NAS. Depends on configuration "Ignore file permissions" option could be set on Windows side.
Patch is from here: https://bugs.busybox.net/show_bug.cgi?id=7430, works with 1.25.1 and 1.26.2 (current).

_Linked issues:_ #2140

As soon as it involves basic functionality used in many community packages, thorough testing could be needed.

### Checklist


o0v7r0o:
Resolved conflicts with master using github tool.

KaraokeStu:
Great stuff, thanks for that. Merging now

PR #1852, opened 2015-09-21
Add cross/fuse, cross/sshfs and spk/sshfs

sea3pea0:
Added fuse and sshfs-fuse to spksrc/cross and added sshfs-fuse to spksrc/spk

I tested this with a DS412+, I have a working sshfs mount.

PR #2647, opened 2017-02-06
Update TVHeadend to 4.0.9

m4tt075:
With the proposed changes TVH 4.0.9 compiles again on evansport with 5.2 toolchains.

Starting point was GoodOmens83's [pull request](https://github.com/SynoCommunity/spksrc/pull/2189), which apparently was close to final but unfortunately closed again. I then found an easy fix for the hdhomerun compilation problem in jdierkse's fork [here](https://github.com/jdierkse/spksrc/blob/develop/cross/tvheadend-testing/patches/hdhomerun.patch).

Again, the only thing I did is to make it compile again. Feel free to try it on your platform and report back. Maintaining this package is beyond my skill set, but maybe GoodOmens83 or someone else can take it from here...


Dr-Bean:
If I'm not mistaken, there's a better way. While this probably works, it's definitely not ideal.

The HDHomerun sourcecode is actually downloaded and compiled during the build process of TVHeadend (See https://github.com/tvheadend/tvheadend/blob/master/Makefile.hdhomerun#L37 for example)
For the purposes of packaging with spskrc, it makes a lot more sense to create a separate cross-package for HDHomerun.
Then, instead of applying the fix as a patch to tvheadend Makefile, it should be fixed directly as a patch to HDHomerun's Makefile: L14-15 (maybe more) have some entries that should be dropped:
```
CC    := $(CROSS_COMPILE)gcc
STRIP := $(CROSS_COMPILE)strip
```

m4tt075:
I looked at it, but again, beyond my skill level: libhdhomerun is linked statically into TVH during the build process. There are lots of cross-references in Makefile.hdhomerun to the TVH build environment which I'm unable to resolve when separating out libhdhomerun into a cross-package. Definitely not as straight forward as the openssl or zlib cross-packages.
I was hoping the suggested fix would get GoodOmens PR across the line, as it was already close to being merged according to it's status label changes. But if it takes more, I'm afraid someone else needs to step in...

Dr-Bean:
So what if you look at it from a different perspective? Start by disabling TVHeadends default behavior with `CONFIGURE_ARGS += --disable-hdhomerun-static`, and then create and include a dynamic library for HDhomerun?

m4tt075:
Well, ok, Dr-Bean. Not sure changing perspectives, will increase my skill level, but here you go. I created "some" cross-package of libhdhomerun and made it a dependency of tvh, but have no clue whatsoever whether it is now dynamically linked correctly. I don't even own a HDHR device. Only thing I could verify is that it compiles on evansport with toolchain 5.2. Is that what you were looking for?

Dr-Bean:
The perspective mention was in reference to the argument of statically linking HDHR, not your skill level...I really have nothing to say about that.
Point being that TVHeadend doesn't necessarily care how HDHR support is setup, it's just the devs choice to use a vendored, statically linked lib by default.

So, going forward, you're on the right track, but there's something missing to really link the two together. First off, you'll want to tell spksrc that `cross/libhdhomerun` should be compiled as part of tvheadend. Adding `cross/libhdhomerun` to `DEPENDS` in `cross/tvheadends/Makefile` does that.

Then, if you look at TVHeadends configure options (`./configure --help`), there's another option next to `---disable-hdhomerun-static`, and that's `--enable-hdhomerun-client` (or something, can't remember the exact wording)
What it does is tell TVHeadend that, while you've disabled the static lib for HDHR to compile against, you still want TVHeadend to support HDHR.

When you now compile TVHeadend, it should show that it's trying to locate hdhr stuff, but it probably says 'not found' or something to that effect, and effectively skips it again.
The reason this happens is that TVHeadend is looking for the HDHR header files (`.h` files) in a specific place: $(INSTALL_DIR)/$(INSTALL_PREFIX)/include/libhdhomerun`. In addition, at runtime, it'll need access to the shared library as well (so, the `.so` file that HDHR produces) in `$(INSTALL_DIR)/$(INSTALL_PREFIX)/lib/`.

Looking at HDHR's Makefile, it does nothing to get any of the files to the right place: it just compiles the shared lib in place, and that's it. So, you'll have to do the copying in cross/libhdhomerun/Makefile. The standard way of doing something like that is to copy the files via setting `INSTALL_TARGET` and doing an actual copy. In a more correct Makefile fashion, it looks like this: https://github.com/SynoCommunity/spksrc/blob/master/cross/python/Makefile#L59-L63.
That also should help in producing libhdhomerun's PLIST, see [this](https://github.com/SynoCommunity/spksrc/wiki/PLIST-files) for more details.

As for testing: let's publish a beta package once we're happy with this. Someone who uses HDHR might be able to check if it works or not ;) Sound good?

m4tt075:
Sounds good. And thanks for your detailed guidance and patience, @Dr-Bean. I have followed your advice. After some struggles, I believe the libhdhr files are copied to the right place now. However, the compilation breaks down with the following error:
`[...]/libhdhomerun.so: file not recognized: File format not recognized
collect2: ld returned 1 exit status`
I believe this is the same problem that broke @GoodOmens83's original pull request. Probably I need to introduce specific compiler options into the libhdhomerun Makefile now to fix it. Would you mind taking a look again?

Dr-Bean:
Looks good :)

I think I already mentioned a fix for the compile issue in here, maybe you missed it: https://github.com/SynoCommunity/spksrc/pull/2647#issuecomment-278062322.
Check if commenting out the two values (CC and STRIP) in HDHR's Makefile works: spksrc already sets the correct values, but that Makefile overrides them again: that's no good. 
In case you're interested, the following link has some background on creating a patch: https://github.com/SynoCommunity/spksrc/pull/1852#issuecomment-142917573.

Three other suggestions while I'm at it:
- Set `CONFIGURE_TARGET = nop`, instead of pointing to a non-existent target; 
- Add an `@install -m 755 -d $(STAGING_INSTALL_PREFIX)/lib` in myInstall on L22 or so, to be sure it exists before copying the .so file;
- Add a PLIST for libhdhomerun. It should only contain the .so file (and maybe check if tvheadend's is still correct). See the wiki article on PLISTs for information.

m4tt075:
@Dr-Bean: Seems we are getting somewhere...  :-)

Everything implemented as requested. TVH's PLIST file seems fine to me. If anything the share row can be removed, but wasn't sure so I kept it as is. Again, I only compiled the evansport package on the 5.2 toolchain.

Anything else you want me to look at or adapt? 

Will do some reading on how to clean up the github commit message mess I have left here in the meantime...  ;-)

Dr-Bean:
Nice work, no further comments. If you want to look into squashing the commits, I'll leave this be for a bit, otherwise I'll see what I can do.

m4tt075:
Well, thank YOU, really! I have looked into squashing but learned that it should be done before pushing to a public fork. Therefore, if that's fine for you, I will leave it as is here and keep it in mind for any upcoming projects.

cytec:
@m4tt075 "public fork" is a little bit confusing there... you `should not` squash commits if others already use your repo/fork/branch and follow it.. cause it'll mess up the comment history as you'll do a force push... which will most likely produces problems for everyone useing your fork/branch.

In this particular case i'd assume no other people are using the branch you used for the PR so you should be save to squash the commits ;)

m4tt075:
@cytec, @Dr-Bean: OK, I rebased on master, squashed all of my commits but left @GoodOmens83's original commit intact as he did the work. Anything else you need before merging or do you generally keep PRs idle for some time for one reason or another (other than real life of course)?

PR #2654, opened 2017-02-09
Update SABnzbd-testing to 1.2.1Beta1

Safihre:
_Motivation:_ Update SABnzbd-testing to 1.2.1Beta1

### Checklist

### DIFF
--- a/cross/sabnzbd-testing/Makefile
+++ b/cross/sabnzbd-testing/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = SABnzbd
-PKG_VERS = 1.2.0
+PKG_VERS = 1.2.1Beta1
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS)-src.$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/sabnzbd/sabnzbd/releases/download/$(PKG_VERS)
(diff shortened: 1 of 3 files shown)


KaraokeStu:
LGTM

KaraokeStu:
Published

Safihre:
@KaraokeStu @Dr-Bean 
We would like to upgrade the par2cmdline version to the [multithreaded one](https://github.com/jkansanen/par2cmdline-mt).
On my desktop I can simply run these commands after downloading the `src`: https://github.com/jkansanen/par2cmdline-mt/archive/v0.6.14-mt1.tar.gz
```
aclocal
automake --add-missing
autoconf
./configure
make
make install
```
What would be a similar set of commands for a package here?
I looked at the `unrar` package, would it be as simple as that?

Since SAB is the only one using the par2cmdline package, it seems just upgrading that one should be safe enough.

Safihre:
Oops, should have read the documentation. WIll go and try to figure it out myself first!

KaraokeStu:
Hi, we will want to update the par2 cross package to use the MT source - I'll take a look at that for you

KaraokeStu:
Do we want par2cmdline-mt or par2cmdline-tbb?

Safihre:
Aaah great!
TBB is faster, but par2cmdline-mt is much much easier to install, because TBB requires the Intel Thread Building Blocks (TBB 😃) and this package is often not available or doesn't compile well on non-x86.

KaraokeStu:
OK, I'll take a look at getting that compiling. @Dr-Bean Any objections to upgrading par2cmdline to par2cmdline-mt?

Dr-Bean:
No, assuming the `-mt` source compiles/works fine for current arches.
We'll probably want to set up a new package though, right? If the current one is only used for sabnzbd, might as well rename the existing package (append `-mt` in all the right places, presumably). That enables us to re-add the original one in case we need it for some other SPK in the future.

KaraokeStu:
ok, I'll add a new -mt package and update sabnzbd to pull that cross instead :)

Safihre:
So since it uses the same command `par2`, it will just overwrite the older `par2` link on install?

PR #2655, opened 2017-02-09
Update Syncthing-Inotify to 0.8.5

o0v7r0o:
_Motivation:_ This is an important update because syncthing 0.14.22+ will no longer work with syncthing-inotify 0.8.4 (see https://github.com/syncthing/syncthing-inotify/issues/154)
_Linked issues:_ no

### Checklist

### DIFF
--- a/cross/syncthing-inotify/Makefile
+++ b/cross/syncthing-inotify/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = syncthing-inotify
-PKG_VERS = 0.8.4
+PKG_VERS = 0.8.5
 PKG_EXT = tar.gz
 PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/syncthing/syncthing-inotify/archive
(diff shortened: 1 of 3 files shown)


Dr-Bean:
LGTM. Maybe also update the synchting package to latest, and set SPK_DEPENDS from this package to that version? I realize that this version of syncthing-inotify is probably backwards compatible, but it seems to make some sense...

o0v7r0o:
@Dr-Bean yes, I wanted to update Syncthing anyway, because auto-update restart is broken in 0.14.21 and release candidates are introduced. It makes sense to set release channel to 'stable' by default in config (https://docs.syncthing.net/users/releases.html#how-to-choose). So, it will follow.

Dr-Bean:
Yep, sounds good to me. Will you amend this PR with the SPK_DEPENDS changes after that then? Then we'll leave this PR open for a bit, so we can release both packages together.

o0v7r0o:
@Dr-Bean btw, what do you say about #2646?

o0v7r0o:
Added new syncthing version dep.

o0v7r0o:
@KaraokeStu thanks, could you please trigger update in the repo, because I don't see new version there.

KaraokeStu:
Building and publishing new spk now.

PR #2658, opened 2017-02-10
Remove unused packages in cross and native

Dr-Bean:
_Motivation:_ Cleanup time. No point in having unused, outdated packages in the repo. If need be, packages that will be used can simply be brought back together with an associated SPK.

### Checklist
Checked PKG_NAME against dependency-tree. Checked a lot of entries manually because the list is quite a bit longer than I expected...

@Diaoul, FYI.


GuillaumeSmaha:
@Dr-Bean Could be usefull for your clean :
- https://gist.github.com/GuillaumeSmaha/324a576cbe203610b9a068ab9ef7e6c9 : Draw graph deps with graphviz
- https://gist.github.com/GuillaumeSmaha/74c4d27ea3052cabc746fbfde9c595af : List not used packages in cross and native directory.

Both scripts have an option -d to specify the spksrc directory and an option -b to include BUILD_DEPENDS values in the graph or in the used package.

Dr-Bean:
I'll take a look at those scripts, thanks. I suppose my solution is more quick and dirty:
```
for spk in spk/*/Makefile;do cd spk/$(basename $(dirname $spk)) && make dependency-tree && cd ../..;done >> /spksrc/deptree.txt && for package in $(ls cross); do echo $package >> results.txt && for entry in $(grep ^PKG_NAME cross/$package/Makefile | awk -F' ' '{print $3}'); do if grep -q $entry deptree.txt;then echo "Found" >>results.txt; else echo "NotFound" >> results.txt;fi;done;done
```
...as evidenced by the fact that it has some false positives ;)

`BUILD_DEPENDS` has the same value as DEPENDS, the only difference with DEPENDS is that the PLIST isn't processed. So you can just bunch them up and process them the same.

Dr-Bean:
@GuillaumeSmaha Didn't dive into the code of your scripts, but I noticed that some results aren't correct: `cross/mercurial`, `cross/octoprint`, `cross/sonarr` come up as Not Used. 
Probably partly because `cross/$(SPK_NAME)` isn't processed? That's why I went the `make dependency-tree` route in the first place ;)

GuillaumeSmaha:
@Dr-Bean Yes, some bug with multiple lines in sed replacement.
I update the script if you are interested.

Here the list I found for the unused package : 

	cross/alsa-utils
	cross/apr
	cross/apr-util
	cross/attr
	cross/audiofile
	cross/coreutils
	cross/dbus
	cross/dobby
	cross/enum34
	cross/espeak
	cross/flex
	cross/flite
	cross/gc
	cross/gdb
	cross/gsasl
	cross/guile
	cross/iksemel
	cross/jemalloc
	cross/julius
	cross/libcap2
	cross/libical
	cross/libid3tag
	cross/libidn
	cross/libmad
	cross/libpar2
	cross/libsndfile
	cross/libunistring
	cross/linux-pam
	cross/lzo
	cross/mpd
	cross/neon
	cross/newznab
	cross/openldap
	cross/openssh
	cross/openvpn
	cross/pear
	cross/polarssl
	cross/pyasn1
	cross/serf
	cross/termcap
	cross/tinyproxy
	cross/udns
	cross/umemcache
	cross/virtualenv
	cross/xzdec-static
	cross/zope.interface
	native/guile

GuillaumeSmaha:
@Dr-Bean Nice `make dependency-tree`, I didn't know this command 👍 
There is missing `lirc-0.8.7` in `lirc` spk. But maybe you can define the arch to return the good deps.

Dr-Bean:
Outputs look very similar. Two entries in your list are strange though, maybe your branch is outdated? I think `libunistring`>`slang`>`mc` and `zope.interface` doesn't exist in master anymore.

If you meant that my script doesn't detect `lirc-0.8.7`, yeah, probably optional DEPENDS won't work out of the box. Good thing we don't have a lot of those...Anyway, for a one-liner, it did most of the work ;)

We should probably save these things somewhere, so we can run them occasionally.

Dr-Bean:
Ha..the libunistring is an error on my part somewhere. It's actually not used. I'll have to look into that.

GuillaumeSmaha:
Yes, I will keep the link available.
I create a gist to gather the script :
https://gist.github.com/GuillaumeSmaha/5545f6e48b5d3ce35399d6b3727c4a9c

Safihre:
@ymartin59 Shall we delete this branch?

ymartin59:
@Safihre Sure. File removal is not so hard to replay...

PR #2667, opened 2017-02-15
[Branch dsm6] DSM 6.0.2: Fix TC_FIRMWARE based on version available

GuillaumeSmaha:
Based on version available on http://dedl.synology.com/download/DSM/release/6.0.2/
The toolchains for DSM 6.0.2 should use firmware `8451` instead of `7321`


Dr-Bean:
Unless the DSM6.0.2 toolchains are backwards incompatible with 6.0, the current entries are correct. THe only reason for them to be backwards incompatible is if the GCC version is different.

All DSM6.x toolchain's `TC_FIRMWARE` should be set to the first version of DSM that it works for, and that ends up being the first stable version that Synology released iirc, 7321.
We use the same approach for DSM5.x and DSM4.x.

GuillaumeSmaha:
Ok, I didn't know. Thank for your answer.
I close the PR.

Issue #2664, opened 2017-02-13
Wiki: Document patch creation


Dr-Bean:
Intro:
* Explain when to use a patch (high level, as in: not when there are config options available to do the same thing, other stuff?);
* Location (patches dir) and naming;
* Fixed ordering done with `001-patchname`,`002-patchname` etc;
* Break large patches up into smaller, self-contained patches;
* Don't set `PATCHES_LEVEL = x` unless you can explain why ;) Generally shouldn't be necessary when following the instructions below.

Main text based on [this comment](https://github.com/SynoCommunity/spksrc/pull/1852#issuecomment-142917573)
```
/spksrc/cross/fuse$ mkdir -p patches ##Make sure the patches dir is empty!
/spksrc/cross/fuse$ make clean ##Clean up old stuff
rm -fr work work-*
/spksrc/cross/fuse$ make arch-88f6281-5.2 ##We need fuse's source to be extracted, this is an easy way.
===>  Building package for arch 88f6281
[..]
../../mk/spksrc.cross-cc.mk:122: recipe for target 'arch-88f6281' failed
make: [arch-88f6281] Error 2 (ignored) ##Yep, failed ;)
/spksrc/cross/fuse$ cd work-88f6281-5.2/fuse-2.9.4/ ##Enter $(WORK_DIR)/$(PKG_DIR), which is where the patch is applied from
/spksrc/cross/fuse/work-88f6281-5.2/fuse-2.9.4$ cp util/Makefile.in util/Makefile.in.org ##Create original file to compare with
/spksrc/cross/fuse/work-88f6281-5.2/fuse-2.9.4$ vi util/Makefile.in ##Open the file in your favorite editor
##...and make the necessary changes, meaning: redo manually what the broken patch does.
##util/Makefile.in should end up containing the changes needed for cross-compiling: a "fixed" file if you will.
/spksrc/cross/fuse/work-88f6281-5.2/fuse-2.9.4$ diff -Naur util/Makefile.in.org util/Makefile.in > ../../patches/cross-compile.patch ##This creates the actual patch.
/spksrc/cross/fuse/work-88f6281-5.2/fuse-2.9.4$ cd ../.. ##And we're done here...
/spksrc/cross/fuse$ make clean ##...so clean up again...
rm -fr work work-*
/spksrc/cross/fuse$ make arch-88f6281-5.2 ##...and test.
===>  Building package for arch 88f6281
[..]
[...]

PR #2671, opened 2017-02-15
Rename patches and add information in patches

GuillaumeSmaha:
_Motivation:_ Rename patches to follow instructions in issue #2664 
I don't edit the packages which are deleted in the PR https://github.com/SynoCommunity/spksrc/pull/2658


Dr-Bean:
@GuillaumeSmaha Had a quick look at your PR, and have a couple of comments:
- I realize I didn't explain it in the issue (it was just meant as a 'write down something on this subject in the doc'-type of text), but the numbering is only needed if two patches on the same file are necessary (as in: patch1 changes x in Makefile, patch2 changes y in Makefile. 
When that happens, you want to make sure that patches are applied in a certain order, otherwise the second patch will not apply.
For cases where there's just one patchfile, or there are multiple patchfiles that operate on different files, numbering isn't needed. I don't necessarily _mind_ it, but I'm not sure I'll remember myself when I add a patch in the future ;)
- Correct me if I'm wrong, but it looks like a number of patchfiles have changes in them other than the added comments at the top (spaces, tabs, EOF line breaks removed, from what I could tell). Patchfiles have a very specific format, so any changes like that will prevent them from being applied.
- There are a number of patches that are renamed and now have spelling errors (noticed `depenances`, `defaul`, may be others)
- I'll have to hold off on merging this until the new set of cross packages comes in, because we'll run into merge conflicts when patchfiles are edited: getting the updated packages in has a higher prio from that perspective.

GuillaumeSmaha:
- As you want, in my case I prefer to use a prefix number even if the patches concern different files but I will follow your way. Some patches miss the number prefix in your way (in screen package for example).

- If you want, I can just made a list of the patches where the name should have a number prefix. It will be more simple that try to update and merge this PR. And after your clean, you could use this list to fix the name.

I also do this to do a patch review and I think some patches are may be useless in trying to keep compatibility with old architecture.

Dr-Bean:
Okay, thought about it for a bit, let's keep the numbering. Makes the doc easier to write, and it's unambiguous for someone who reads it. More importantly, I was thinking it had added value to know which patches had an interdependency on each other, but that's probably not true.

As for the PR itself: thanks. A review+added comments certainly adds value. I don't think we'll want to actively remove existing patches (not unless they don't work for their intended purpose), but knowing the reason that the patch exists in the first place is something we've been missing for a while.

KaraokeStu:
Please could you fix the conflicts so that I can merge this?

GuillaumeSmaha:
@KaraokeStu Rebase done

m4tt075:
Thanks a lot, @GuillaumeSmaha!

PR #2679, opened 2017-02-25
Updated default port for Radarr to 8310

KaraokeStu:
_Motivation:_ Updated the default port so it doesn't clash with High Availability

### Checklist


Dr-Bean:
This doesn't change the port in Radarr's config though, right? Just in the package setup. So while it works, Radarr won't be listening at port 8310, unless you manually change the config.

KaraokeStu:
Yeah, gonna sort that out now and re-commit :)

Issue #2652, opened 2017-02-08
Radarr - In app update killed the app. 


madvtak:
Hi all,

Anyone experiencing this issue? There was a new update tonight so i updated, within the app. After the updated downloaded completed, i was presented with the app has been updated click here to reload the page. I reload the page and it never loads. I check the package.

![image](https://cloud.githubusercontent.com/assets/18261123/22724318/2d6aa81a-ed7a-11e6-96d2-c89de4cab960.png)

I try to run the package, it says its starting but it never runs. 

![image](https://cloud.githubusercontent.com/assets/18261123/22724366/6d20b3a0-ed7a-11e6-8b4a-13e484e972e7.png)

I try to refresh the page a min or two later just in case it actually did start.

![image](https://cloud.githubusercontent.com/assets/18261123/22724377/8019b6f0-ed7a-11e6-87e4-c663765aec36.png)

The only way to fix this is to uninstall and start over from scratch again.

```
admin@Syno:/var/packages/radarr/scripts$ ./start-stop-status start
Starting Radarr ...
admin@Syno:/var/packages/radarr/scripts$
```

```
radarr@Syno:/var/packages/radarr/scripts$ ./start-stop-status start
Starting Radarr ...
radarr@Syno:/var/packages/radarr/scripts$
```

I try to start from SSH and this is what i am presented with.

In case someone wants to see the logs, not a lot of info /shrug

![image](https://cloud.githubusercontent.com/assets/18261123/22724561/e79fd33a-ed7b-11e6-9d7a-8af62ffca2b2.png)

KaraokeStu:
Hi, please can you confirm your current installed dsm version?

madvtak:
![image](https://cloud.githubusercontent.com/assets/18261123/22729160/59b06eb6-ed96-11e6-8e8b-32fb37a251bf.png)

Fieze:
Same problem, DSM 6.0.2-8451 Update 9

cytec:
https://github.com/SynoCommunity/spksrc/issues/2216

KaraokeStu:
Your logs should be in /var/packages/radarr/target/var/.config/Radarr/UpdateLogs

KaraokeStu:
@cytec since this installed originally without issue on both DSM 5 and 6, I don't think it is linked to #2216

cytec:
yeah saw that on the second read too... nevermind than

newkind:
[...]

PR #2680, opened 2017-02-26
Fixing invalid port in Radarr package

KaraokeStu:
_Motivation:_ Updated to update the ports on Radarr to 8310 and 8311
_Linked issues:_ Should fix https://github.com/SynoCommunity/spksrc/issues/2652

### Checklist

### DIFF
--- a/spk/radarr/Makefile
+++ b/spk/radarr/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = radarr
 SPK_VERS = $(shell date +%Y%m%d)
-SPK_REV = 1
+SPK_REV = 2
 SPK_ICON = src/radarr.png
 DSM_UI_DIR = app
 BETA = 1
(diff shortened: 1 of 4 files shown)


ressu:
Thanks. Never found the time to do the tests, bit overwhelmed with other stuff at the moment.

KaraokeStu:
No problem @ressu , thanks for creating the package in the first place!

I help where I can!

MostDefiantly:
I am able to install this now, thank you everyone for the effort to get it working.

SuperDOS:
hmm it won't start, any ideas which logs to look for why?

M0UL:
this has not updated the port  for my radarr config. but it says it has updated the port in dsm. but it was stil using the previous defautl port.

KaraokeStu:
Hi, the port is not updated as part of an upgrade (it leaves your config file intact), however it does set it correctly on new installs

SuperDOS:
Noticed this, had to remove and reinstall it, was a bit confusing how to restore the backup but finally managed to find where the DB should reside. :)

KaraokeStu:
It's a beta version, but as a general rule we don't update port numbers our config entries as part of an upgrade. Many people use non standard ports too

ressu:
Also changing port configuration during upgrades causes more issues than it fixes. The only side effect at the moment is that the link doesn't immediately work.

On the other hand, updating the port during upgrade causes integrations to fail and will most likely confuse the user that is already used a different port. Then there is the matter of user set custom ports, we would need to detect whether the port is still the default port or not and we have no way of knowing if the user has manually set the port back to the old default etc etc.

So yeah, beta package and a good policy in general.

Issue #2185, opened 2016-03-12
[REQ] Update nzbget-testing to version 17.x, c++14 required!


schumi2004:
As title says, please upgrade nzbget testing to version 17.x
http://pastebin.com/TaCUYebS
##

Dr-Bean:
This might not be possible. From what I see, C++ 14 is implemented fully in GCC5+. The DSM6RC toolchains are all GCC4.9.

Background of nzbget's reasoning here: https://github.com/nzbget/nzbget/issues/88.

schumi2004:
Ow, to bad then. I was expecting something like this.

Dr-Bean:
I would suggest to have a go at it with the DSM6RC toolchains anyway. Parts of the C++ 14 standard were introduced pre-GCC5, and that might be enough.

hugbug:
GCC 4.9 is OK for NZBGet v17. Although GCC 4.9 doesn't fully support C++14 I'm doing extra efforts to ensure NZBGet works with GCC 4.9.

>checking whether /home/openelec/github/spksrc/toolchains/syno-alpine-5.2/work/arm-cortex_a15-linux-gnueabi/bin/arm-cortex_a15-linux-gnueabi-g++ supports C++14 features with -std=c++14... no

GCC 4.9 supports flag `-std=c++14`. Are you sure this is the output from GCC 4.9? Please post **config.log**, it usually includes full compiler versions.

GoodOmens83:
Tried compiling the latest 17 development version (from github) with the 6.0 beta2 toolchains and everything compiled fine.  Can't test as I don't have 6.0 installed.

`checking whether to test compiler features... yes`
`checking whether /home/xxx/spksrc/toolchains/syno-avoton-6.0/work/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu-g++ supports C++14 features by default... no`
`checking whether /home/xxx/spksrc/toolchains/syno-avoton-6.0/work/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu-g++ supports C++14 features with -std=c++14... yes`

Dr-Bean:
Thanks for the info @hugbug. The toolchain @schumi2004 was using is based on GCC4.6, so that's why those errors occurred (a lot of the NAS devices on the market use old versions of kernels/GCC/whatnot, which is a shame).

[...]

PR #2692, opened 2017-03-03
[branch dsm6] update nzbget-test to version 18

scelestic:
_Motivation:_ update the nzbget-testing package in branch dsm6 to latest 18 final release
_Linked issues:_ this should be tested on other platforms and checked if we have all new options available in the web interface.

### Checklist
- [] Build rule `arch-armadaxp` completed successfully


KaraokeStu:
Can you confirm that you've managed to get this built successfully, as DSM 5.x doesn't support C++ 14 and DSM 6.x is still only GCC 4.9.

Please see https://github.com/SynoCommunity/spksrc/issues/2185

scelestic:
Can confirm it builds and the package is functional. When looking in the config.log I do see it did define HAVE_CXX14, also looking at this bug report it seems gcc 4.9 should be able to support it : [https://github.com/nzbget/nzbget/issues/142]

KaraokeStu:
@scelestic Great stuff - have you had a chance to test on DSM 5.x?

scelestic:
I did not test it before but with the 5.2 toolchain it fails to compile because C++14 is missing, it is still using gcc 4.6.4 so that makes sense. I guess nzbget 18 will be DSM 6.x+

ymartin59:
Please please do not submit PR to `dsm6` branch which does not concern specifically DSM 6.x support.

Procedure consists in regularly merging from `master` to `dsm6`. So to avoid conflicts, packages have to be submitted only to `master`.

Either rewind your PR on `master` only, or wait for me to do so, as someone requested nzbget for DSM 6.1 #2880

bidorffOL:
I just retrieved this PR and I was wondering if there was any reason for the port to have been moved from the default 6789 to 9876 ?

scelestic:
Looks like you're right, the port for nzbget-testing is 9876  and the package nzbget uses 6789.

I based my changes on the nzbget-testing package while not using -testing code and later when I renamed that back I never even thought about the control port. Feel free to change it to whatever you like, I don't think this pull request will be used.

bidorffOL:
I just saw that, I'll try to PR a cross version of NZBGet this week that can be published in the `master` (i.e. NZBGet 16.4 for DSM 5.x and NZBGet 19.1 for DSM 6.x).

Issue #2656, opened 2017-02-10
Automate new upstream version availability detection and processing


ymartin59:
Here is first specification of my idea. Python sounds me a good choice for implementation.

### Expected behavior
- Without package location as argument, loop on packages, either `cross`, `native` or `spk`
- Parse Makefile to get version and project download location
- Test if a newer version is available according to "supported" download sources:
    - For a git or subversion repository, list tags ordered by dates
    - For a file storage, either ftp listing, html download page, parse content to list available files and target URI
    - Maybe some specific support will be required for common code hosting platforms
- Update package version in Makefile and increase revision
- Try to download archive in `distrib` and report error in `FAILED-version-update.log`
- Generate `digests` from archive
- Any previous steps may fail, for instance test algorithm does not support source or failed to compare versions, and if so will generate a `FAILED-version-update.log` with details
- Run `make arch-88f6281` first and by default to fail fast and report compilation issue. Generate `FAILED-arch-99f6281.log` file or `OK-arch-99f6281.log`
- Generate/update `PLIST` or mark it as modified for manual review
- If script has `all-archs` switch turn on, loop over all archs and produce each `FAILED/OK` log file

### How to use
When over, browsing files marked as new or modified in git working copy is straight forward.

With luck and probably for most minor updates, removing log files and review before commit may be the only remaining job

GuillaumeSmaha:
I made a POC to do this, I will share it when I will obtain a good result. Currently, I just checks the link availibility.

@ymartin59 What do you think about try to ping on possible new file by incrementing the version ? It can be lesser specific than parsing html page.

Also parsing Makefile can be complex when there is specific call like boost which use `substr` (https://github.com/SynoCommunity/spksrc/blob/master/cross/boost/Makefile#L4).
Currently, `substr` is the only special case.

GuillaumeSmaha:
You can find the POC here :
https://gist.github.com/GuillaumeSmaha/5545f6e48b5d3ce35399d6b3727c4a9c#file-check_package_availability-sh

The result on master : (SVN/git is not already supported)

	cross/domoticz -> Not Available: svn://svn.code.sf.net/p/domoticz/code/trunk
	cross/iksemel -> Not Available: http://iksemel.googlecode.com/files/iksemel-1.4.tar.gz
	cross/imagemagick -> Not Available: http://www.imagemagick.org/download/ImageMagick-6.9.5-10.tar.xz
[...]

PR #2708, opened 2017-03-18
Add var DOWNLOAD_PAGE to define URL with list of files. 

GuillaumeSmaha:
To detect new version of package, it needs this new var which define an URL on list of files
See https://github.com/SynoCommunity/spksrc/issues/2656

I also updated URL of fdk-aac to change the call to old API of sourceforge


hgy59:
This PR looks rather outdated. If this is required, it sould be implemented in #2656 (still active).

PR #2711, opened 2017-03-21
Update SABnzbd-testing to 2.0.0RC1

Safihre:
Hi it's SAB again 😃


KaraokeStu:
@Safihre I've had to do another PR to correct the missing filename update in the digests file - I should have spotted it first time round!

Safihre:
Oh man you're right! I should have seen that too of course. Will pay extra attention next time.

PR #2723, opened 2017-03-31
updated jackett to v0.7.1225

abienz:
_Motivation:_ Update to latest version of Jackett

### Checklist

### DIFF
--- a/cross/jackett/Makefile
+++ b/cross/jackett/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = Jackett
-PKG_VERS = 0.7.806
+PKG_VERS = 0.7.1225
 PKG_EXT = tar.gz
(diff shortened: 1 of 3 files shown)


abienz:
Yes, i couldn't find any documentation on how to do that, can you point me
in the right direction please?

m4tt075:
There might be simpler ways but this is how I do it:
(1) Go to cross/jackett directory
(2) `rm digests` (unless you have removed it already)
(3) `make 'some architecture'` (does not matter which one)
(4) `make digests` (will create a digests file based on the source you have just built)
(5) `make clean` (if you like)
Done.

abienz:
Thanks @m4tt075 that was much easier than I expected.

@KaraokeStu I have updated the digests file with the new hashes for jacket-0.7.1225

PR #2727, opened 2017-04-03
[dsm6 branch] DSM 6.1 toolchains

ymartin59:
Each one tested by building `nano` package


jreese:
I was able to successfully build a package from this and run it on my braswell unit.  Thanks for posting this!  👍

michailf:
`TC_FIRMWARE = 6.1-15047-1`

Should the "-1" suffix be removed?

ymartin59:
@michailf You may be right concerning firmware version to specify in toolchain but do you know any guideline or documentation describing exactly how to set it and from which source ?

michailf:
@ymartin59 No, I do not know of any guidelines.  I only observed how other TC_FIRMWARE is set.  The last number is an update level, the one before is a build number which is unique across different DSM releases.

TC_FIRMWARE appears to be used only for display purposes at this point so it's not that big of a deal.

m4tt075:
@ymartin59: Thanks for the toolchains. `Tvheadend 4.2.2` (PR #2763) compiles with all of them, except for `hi3535`, which fails building the `ffmpeg-3.2.4` dependency. If I'm not mistaken the only model using this architecture is the `NVR216`, which is a dedicated surveillance system. Fix seems not straight forward. But before I spend a lot on time on this now, Do you think this actually needs fixing or can I safely "ignore" this architecture fo the TVH package in light of an NVR use-case?

KaraokeStu:
I just need confirmation from @Dr-Bean or @Diaoul that they are happy for me to merge these changes in, please hang-fire while I wait to hear from them (They've been a bit quiet lately - I imagine life has got in the way!)

ymartin59:
@KaraokeStu Thanks for the effort. These is still a question about how `TC_FIRMWARE` value should be set, either `6.1-15047-1` or `6.1-15047`. What is your opinion about it ?

KaraokeStu:
Hi @ymartin59 I'd drop the -1 off the end, I think that makes more sense and falls in line with all other versions

ymartin59:
I have replaced `TC_FIRMWARE=6.1-15047`

ymartin59:
@KaraokeStu I think it is now worth to merge this PR into `dsm6` as it has been used to generate some packages by varia guys... by the way, it will ease package generation then.

Issue #2278, opened 2016-04-22
[Request] Fully updated TTRSS


live4truths:
Website: https://tt-rss.org/gitlab/fox/tt-rss/wikis/home
Version: 9e84bab4
Description: RSS Reader
Documentation: https://tt-rss.org/gitlab/fox/tt-rss/wikis/home
Source code: https://github.com/major/tt-rss
License: GNU

I have been trying desperately to figure out how to update/create/all that sort of stuff with synology packages for the past few weeks. Failing miserably. having to reload my NAS a few times. While I love learning, I am hoping to find the community willing to keep the package updated while I continue. It would likely take someone educated here 5 minutes, and me 5 months.

Thank you for reading.
##

live4truths:
Would this get any more interest if I threw $5 at it?

live4truths:
$10? :(

joeyd2:
Yes, could we please get an update? It's been a very long time since it was updated and maybe fix the PHP error when installing on DSM 6?

live4truths:
Anyone?

michailf:
You can try my fork https://github.com/michailf/spksrc/tree/dsm6

It builds tt-rss 17.1

You may have to use DSM 5.2 toolchain.

live4truths:
@michailf Thank you for this, I am probably not smart enough to figure out how to convert it to the 5.2 version, or even install this. TTRSS was one of the few programs I was waiting on to move to DSM6, So if I can figure it out (or you have any insight to help) I will give it a shot.

Thank you for your hard work

michailf:
@live4truths You have to follow the instuctions provided in the README.

First get the dev environment.  You can use Docker for this.

### Clone dsm6 branch from my fork into your home directory 
git clone -b dsm6 https://github.com/michailf/spksrc.git ~/spksrc

### Install Docker, if you don't have it yet

### Download docker image (you may have to execute this command more than once, if download gets interrupted)
sudo docker pull synocommunity/spksrc

### Run the docker image
docker run -it -v ~/spksrc:/spksrc synocommunity/spksrc /bin/bash

At this point you will be logged in to the dev environment.

### Follow instructions here to build your package: 

https://github.com/SynoCommunity/spksrc/wiki/Developers-HOW-TO

cd /spksrc
make setup
cd spk/tt-rss
make arch-x64-5.2

You have to replace "arch-x64-5.2" with architecture appropriate for your environment.

[...]

PR #2753, opened 2017-05-02
tt-rss upgrade to 17.4

michailf:
_Motivation:_ Upgrade tt-rss to version 17.4.
_Solves issues_: #2278

### Checklist

### DIFF
--- a/spk/tt-rss/Makefile
+++ b/spk/tt-rss/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = tt-rss
 SPK_VERS = 20170422
-SPK_REV = 0
+SPK_REV = 1
 SPK_ICON = src/tt-rss.png
 DSM_UI_DIR = app
 
--- a/spk/tt-rss/Makefile
+++ b/spk/tt-rss/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = tt-rss
-SPK_VERS = 20170422
-SPK_REV = 1
+SPK_VERS = 17.4
+SPK_REV = 11
 SPK_ICON = src/tt-rss.png
 DSM_UI_DIR = app
 
--- a/spk/tt-rss/Makefile
+++ b/spk/tt-rss/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = tt-rss
-SPK_VERS = 20160102
-SPK_REV = 10
+SPK_VERS = 20170422
+SPK_REV = 0
 SPK_ICON = src/tt-rss.png
 DSM_UI_DIR = app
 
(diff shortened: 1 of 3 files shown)


spk/tt-rss/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = tt-rss
-SPK_VERS = 20160102
-SPK_REV = 10
+SPK_VERS = 20170422
```
ymartin59: My opinion is to remove this "timestamp" package version concept and instead use upstream project tag - here 17.4 - even if it "temporarily" prevents smooth upgrade...

spk/tt-rss/Makefile
```diff
@@ -1,6 +1,6 @@
...
+SPK_REV = 0
```
ymartin59: Also my opinion, probably starting revision from 1 is more relevant

ymartin59:
Good improvements, thank you for this job

spk/tt-rss/Makefile
michailf: The problem with switching away from timestamp is that the new version will be lower than the existing installs and will not allow downgrades.

If there is a way to tell DSM that the new version is actually an upgrade, I am all for it.

michailf: The remove/install is hardly an option because you will lose data unless you do backups first.

spk/tt-rss/Makefile
michailf: I can switch SPK_REV to 1, unless someone objects

spk/tt-rss/Makefile
michailf: The problem with switching away from timestamp is that the new version will be lower than the existing installs and will not allow downgrades.

If there is a way to tell DSM that the new version is actually an upgrade, I am all for it.
 
The remove/install is hardly an option because you will lose data unless you do backups first.

KaraokeStu: the SPK_REV field actually determines whether the Synology Package manager actually thinks it is a new version or not

spk/tt-rss/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = tt-rss
-SPK_VERS = 20160102
-SPK_REV = 10
+SPK_VERS = 20170422
+SPK_REV = 1
```
KaraokeStu: If the original SPK_REV was 10, please set the new one to 11

spk/tt-rss/Makefile
KaraokeStu: I'd change this to 17.4 if you can, as it'll let the users know which version they are using (even if we're using git hashes in the background)

michailf:
Interestingly, SPK_VERS had changed from tt-rss versioning to date some time ago.

I made the changes requested.

michailf:
@KaraokeStu Is there something done incorrectly as these changes have yet to make it into a new package?

ymartin59:
@michailf Sorry but we are working on DSM 6 support and all packages require updates... Dependency to Mysql/MariaDB will require some additional work too.

Teryon:
Is there any one working on this ?

PR #2791, opened 2017-06-09
up PKG_DIST_SITE

icemanpro:


KaraokeStu:
Hi, is there a particular reason we're not simply moving to a later version?

KaraokeStu:
Also, please include more details about why you want to create the pull request and your reasons for doing so

cytec:
@KaraokeStu i guess he changed the URL because openssl only hosts the latest version on the root directory. don't know what's speaking against an update though

ymartin59:
Updating `openssl` version has impact on `ffmpeg` and `chromaprint` because `rtmpdump` has to be adapted to new API. I gave it a try and it was no easy.
By the way, I am interested in changing to a HTTP download location as FTP is closed on a network I use (https://github.com/ymartin59/spksrc/issues/1). Please switch to https://www.openssl.org/source/openssl-1.0.2l.tar.gz !

PR #2801, opened 2017-06-22
[TVH] Prevent concurrent installation with tvheadend-testing

m4tt075:
_Motivation:_ Running TVheadend concurrently with a TVheadend-testing package can create problems. They might want to use the same ports or lock the same tuners. This might(!) also solve package the presisting publication issues on the Synocommunity build platform (cannot test).

### Checklist
- [x] Build rule `evansport` completed successfully with 6.1-toolchains
- [x] Confirmed that installation is blocked if tvheadend-testing is already installed with above build

### DIFF
--- a/spk/tvheadend/src/conf/PKG_CONX
+++ b/spk/tvheadend/src/conf/PKG_CONX
@@ -1,0 +1,1 @@
+[tvheadend-testing]


m4tt075:
@KaraokeStu: It does make sense to avoid the concurrent installation of non-testing and testing-packages, but even more so, this is an attempt to see whether we can use this to overcome the publication issues we have been facing for months now. Apparently, your build system complained about an empty `conf` folder as the existing (new) `privilege` file is not acknowledged. This PR introduces another `PKG_CONX` file in the `conf`folder to block the concurrent installation issue. With this, your build system should no longer consider the `conf` folder empty. However, and this is important, IF(!) this works and your build system compiles, it will be critical that the final spk files do not only contain the `PKG_CONX` file in the `conf` folder, but ALSO the `privilege` file. If the `privilege` file is not copied into the final spk package `conf` folder, we MUST NOT publish it, as we risk srewing up user permissions, which would not help anybody. I hope this makes sense. If not, please let me know. If it does, is there a way for you to test this?

KaraokeStu:
Hi @m4tt075 Unfortunately there is no way for me to test this now, as my Synology systems have died at the moment.

I'll rebuild some Hyper-V VMs at some point, as I've reinstalled Windows 2016 :)

m4tt075:
@KaraokeStu: Oh, that sounds like fun! :-)  No worries. Whenever You are ready...

m4tt075:
Thanks, @KaraokeStu. Have you had a chance to check whether your build system produces consistent and correct packages with this commit (please see my notes above)? Just want to be careful and not introduce user permission issues via the build system.

PR #2837, opened 2017-07-24
SABnzbd 2.2.0Beta2

Safihre:


Safihre:
@ymartin59 @KaraokeStu There is a security vulnerability in `unrar` from before 5.5.0 Beta 4.
I can submit a PR with new digests, but since I don't have a Synology NAS or a build-system I can't test it. Maybe one of you could do that and push the update?

KaraokeStu:
It all depends on what the security vulnerability is, however I'm not certain that we want to get around it by switching to a beta version.

Any thoughts on waiting for 5.5 stable?

Safihre:
Yeah that's also what we do, wait for the stable. Probably best here to.

KaraokeStu:
I'll merge this request - feel free to open an issue to update to unrar 5.5 once it is available

PR #1948, opened 2015-11-20
SickRage repository update

CristianBB:
All SickRage developers has moved to a new repository due to team problems, more info on:
http://pastebin.com/raw.php?i=XgMreCb2
and
https://github.com/SickRage/sickrage-issues/issues/3#issuecomment-158350619
and
https://wvvw.me/mirror/sickrage/

PR #2848, opened 2017-07-27
Updated sickrage DSM5 package

echel0n:
Removed requirements.txt and wheel support from package 
PIP installs requirements.txt from git repo after clone
Hard-coded git repo url into installer.sh


echel0n:
Please also publish this to SynoCommunity packages

ymartin59:
@echel0n  @KaraokeStu OK so that package has no versioning and content depends on the precise moment of "when" it has been installed. For sure, NOT-reproducable so NOT-maintainable and impossible to provide any support. What is the "upgrade" process for the user so ? Do I guess right uninstall/reinstall package does the trick ? I really more effort should be done here.

KaraokeStu:
@ymartin59 From what I remember when I was running it, it is like couchpotato, sonarr etc and has its own built in updater.

cytec:
@KaraokeStu you merged that already? Personally i think hardcoded git path is a huge step backwards as we are limiting users to just one specific url there... Same thing goes for SickBeard (why we retired it and only use the custom package) there are a lot of forks and ppl might want to install their favorite fork (which is why we did the git/wizzard thing in the first place) 

@SynoCommunity/developers any input on this?

KaraokeStu:
Ah, you see I was unaware of the previous history here - @cytec do you think I should un-publish the package for now?

cytec:
Not sure about that...
**my** stance would be: revert those changes, and go back to how it was (let user choose on install which fork they want to but set a default) but let's see what the others have to say about that 

maybe we should un-publish the SPK till that is clear? 

\cc @Dr-Bean @Diaoul

KaraokeStu:
Deactivated them for now 👍

echel0n:
This last PR request resolved some incorrect directory issues and the reason for hard-coding in the GIT url was due to the way I install requirements.txt via installer.sh after the GIT clone, I'm not a fork of sickrage as I am the founder and original coder of sickrage

KaraokeStu:
Yes, but sickrage is a fork of sickbeard and thus the sickbeard-custom package allows the user to specify their own git repo URL to use

echel0n:
the BIG difference between sickbeard and my sickrage app is that I don't include the libs inside sickrage and migrated those off into requirements.txt which is installed after the GIT clone so sickbeard-cutsom wouldn't work for this and the other forks of sickrage don't use a requirements.txt file either so sickbeard-custom would work for them

Safihre:
I have to note here that @echel0n is currently in a turf-war with long running forks of Sickrage using DCMA's to try and take them down. 
So hardcoding the git URL might have additional motives..
https://www.reddit.com/r/usenet/comments/6oc65t/sickrage_github_repo_dmcad/

echel0n:
hard-coding the url simply ensures it's cloning from the correct git repo that contains the requirements.txt so that after the clone the installer.sh doesn't fail when trying to pip install the requirements as the app does NOT include the libs in the source tree anymore, sickbeard-custom works perfectly fine for all those other forks simply cause they DO include the libs as part of their source tree

cytec:
Well the `requirements.txt` thing is a good point i think. AFAIK we don't have any case in a `-custom` SPK where that might be there and might not depending on the repo. If i remember correctly we don't have any SPK that uses requirements.txt and doesn't come bundled with it.

So from the Package side of view: I'd rather prefer switching back all the way to the "static" thing (cloning on build, bundling requirements as wheels etc) which ensures that it can be installed even if we should have some problems with github or SickRage server (which is how most of the `-custom` Packages worked before they became custom Packages) 

From the Users side: I find this confusing. Given all the -custom Packages i'd think SickRage would work that way to. Do we even need one since we have SickBeard-Custom which can also be used to install most of SickBeards forks inclusive SickRage (well... you'll have to install the requirements manually).

This whole `war` thing that's going on right now isn't something we should get involved with. We decided not to get into this 2 years ago when they split and i guess we take the same way for that again.

Basically what was said by @Dr-Bean here: https://github.com/SynoCommunity/spksrc/pull/1948#issuecomment-158363273 and there: https://github.com/SynoCommunity/spksrc/pull/1948#issuecomment-158382981 seems like most of that is in some kind still relevant. Im not really sure how we should handle this nowadays.

@echel0n well if you hardcode there is no need to get the files on install at all. AFAIK SickRage provides an internal updater and thus can be updated without pushing updates for the SPK so there is no need for git clone on install at all. If you're going that way (like i said above) putting the whole thing together so it doesn't depend on 3rd Party on install would be the way to go

echel0n:
@cytec originally when I submitted my PR a couple days back with the requirements as wheels it actually did not work correctly and caused sickrage to crash with import errors so instead I changed the install script to pip install from the requirements.txt file that is cloned over via git and that resolved the issue which after doing so is why I hard coded the repo url as its unique to my official version of sickrage.

echel0n:
OK nvm about my last comment, I've found out why wheels and code was not properly compiling, I had git cloned into my windows dev machine and it added carriage returns to the files so I re-forked the repo here and now putting together a static package with wheels and proper versioning so we can move forward.

I'll PR once I'm done compiling and testing, thanks and sorry for all the back and forth here guys!

PR #2858, opened 2017-07-29
Updated SiCKRAGE package to version 9.0.44

echel0n:
Updated to version 9.0.44

### Checklist


KaraokeStu:
Waiting on @Dr-Bean or @Diaoul to confirm if this should have its own package

echel0n:
thanks

splattadat:
@cytec might also have something to add to this, seems like a push to promote this fork above others by decoupling away from the well established -custom install method. #2848

echel0n:
@splattadat -custom install method doesn't cover the pip requirements this needs to run nor does this package install via git but installs via pip

cytec:
Im not really a fan of moving away from the -custom packages as they aim to support a lot of forks!

@echel0n Implementing something that will install requirements after git clone shouldn't be to hard (simply check for requirements.txt after git clone and run `pip install -r` if found should do the trick i guess) so in my opinion that would be a nicer way to go.

of course im open to other opinions but as far as i can see there is nothing preventing us from doing it that way to maintain flexibility for the users.

echel0n:
@cytec I'm closing this PR as I've created a new one as requested by you that will work for custom GIT installs and only will pip install requirements.txt if found
https://github.com/SynoCommunity/spksrc/pull/2877

PR #2867, opened 2017-08-09
SABnzbd 2.2.0RC2

Safihre:


Safihre:
@KaraokeStu Did you already build this? Because due to a failure to load the translations, I updated the files and thus the digests are no longer valid.
These are the new ones:
```
SABnzbd-2.2.0RC2-src.tar.gz SHA1 ca7ac85c4332118defce6c7150ad7f54661b0136
SABnzbd-2.2.0RC2-src.tar.gz SHA256 c7c0472dde2247c7de33dbd106fe604434d0c14809589ef5d018f0555530da9f
SABnzbd-2.2.0RC2-src.tar.gz MD5 131fbc9a392a289257a938e58d9e7f49
```

KaraokeStu:
Could you create a new PR with the updated digests please, and I'll get that updated and published.

I haven't published yet, so no need to change revision numbers

Issue #2866, opened 2017-08-08
Can't add the SynoCommunity source to Docker DSM (DDSM)


miguelmmm:
Hi, I setup a Docker DSM (DDSM) and I'm trying to add the SynoCommunity source to the Package Center within the DDSM. Every time I try, I get an "Invalid Location" message.
I contacted Synology about this and this is their response: Unfortunately this is an issue in which the community server http://packages.synocommunity.com/ does not recognize Docker DSM's model, so it cannot return the package list correctly. Essentially synocommunity did not develop packages for Docker platform. Therefore, no available package list is sent. You would need to contact the developers at synocommunity to resolve this issue.
Are there any plans to support Docker DSM or is there a way to make it work?

Thanks for the help!

ymartin59:
First we have to fix support for DSM 6.x before investigating how to extend to Docker DSM or SRM...

ilatimerP:
Is there any update on this issue.  Is there anyway we can install the package manually or is that not an option.
Cheers

ilatimerP:
Hi ymartin59
I was looking at using Transmission, SquidGuard, Adminer.
Thanks for the quick response.
Cheers

ymartin59:
OK toolchain is identical to x64, so it is easy to produce packages. I have no Docker support on my x86 synology, so if there is a way to run "Docker DSM" on a regular Linux distribution, let me know !

Here is `transmission` package from `dsm6` branch for "Docker DSM". Warning: Not tested at all...
[transmission_dockerx64-6.1_2.92-12.spk.zip](https://github.com/SynoCommunity/spksrc/files/1223300/transmission_dockerx64-6.1_2.92-12.spk.zip)

ymartin59:
@ilatimerP  I did a quick response. May I expect any feedback about your tests of the provided package ?

ilatimerP:
@ymartin59 Sorry about the late response.  I did install it as soon as you posted it and it installed without any issues.  I was also able to log in via the Web gui.  I will download the latest CentOS via transmission to confirm.
Cheers
Isaac

ilatimerP:
@ymartin59 I have tested it and it is running well.  Do you think the repo will ever be added to DDSM?

ymartin59:
To be honest, I am working on repository performance tuning and testing to get DSM 6.x support online... and that is time consuming. If DDSM is no different from DSM 6.x, it may be added then easily.

ymartin59:
[...]

PR #2878, opened 2017-08-18
Update UnRAR to 5.5.8

Safihre:
Main reason for updating is the security vulnerability:

>      a) fixed crashes and security vulnerability when unpacking corrupt
>          RAR archives;

Did not test this, hoped you can do that @KaraokeStu 😺


ymartin59:
UnRAR is part of SquidGuard dependencies, also required to be updated because it no longer compiles. It can be tested thanks this package #2866
I am also interested to find a HTTP download location for `unrarsrc` as FTP is closed from a network I uses

Safihre:
It is also via http available:
http://www.rarlab.com/rar/unrarsrc-5.5.8.tar.gz
Seems to just be the same folder as the FTP!

ymartin59:
Thank you for URL. But please do no longer submit such PR without at least compiling (or inform with "WIP" for "work-in-progress" you intend to do so later)... you may simply create an issue requesting for an update and reason. If we had CI/CD (and I hope to have it soon), related packages would have been broken.

Issue #2893, opened 2017-08-28
How to use SPK_CONFLICT


agross:
### Setup

I try to use `SPK_CONFLICT` (which is undocumented, BTW) for a new package that conflicts with say, the `foo` package.

As soon as I add 

```Makefile
SPK_CONFLICT = foo
```

to my `spk/test/Makefile` (copied from `spk/zsh`) the builds fails:

```
$ cd /sprksrc/spk/test
$ make
../../mk/spksrc.spk.mk:146: *** recipe commences before first target.  Stop.
```

I'm using docker to build.

m4tt075:
I have no experience with SPK_CONFLICT, but used a PKG_CONX file to avoid conflicts with another package (see PR #2801 for details). Might be an alternative to achieve what you are looking for...

agross:
Hi @m4tt075,

Thanks for the hint. How is the `PKG_CONX` file evaluated, or rather who evaluates it? `grep -r PKG_CONX .` shows no results for me.

Thanks!

m4tt075:
The `PKG_CONX` file is not processed by the cross-compilation framework (`spksrc`) but by DSM on installation. It is part of the spk specifications (see Synology DSM6 Development Guide).

cytec:
@agross i've not used it before but i guess it'll work like `SPK_DEPENDS`.. so try something like this: `SPK_CONFLICT="foo>1.2"`

ymartin59:
`SPK_CONFLICT` is simply copied as `install_conflict_packages` in `INFO` file. Refer to https://developer.synology.com/download/developer-guide.pdf
According to that documentation, `PKG_CONX` has higher priority over `INFO` fields.

You have to create `src/conf/PKG_CONX` file in your `spk` with content like examples shown page 70.

By the way I do not understand yet why `install_conflict_packages` from `INFO` may not be honored.

agross:
Thanks for the explanation and details.

My current problem with SPK_CONFLICT is not an issue with the info not being honored, it is that the build fails very early with the message in the original post.

cytec:
@agross have you tried changing 

```
SPK_CONFLICT = foo
```
to
```
SPK_CONFLICT="foo"
```

agross:
Yes, I did. It all leads to the same error.

Did you try that syntax and did it work for you?

agross:
[...]

PR #2889, opened 2017-08-26
zsh: update to zsh 5.4.2

agross:
Update zsh to latest release

### Checklist

- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #2896, opened 2017-08-29
zsh: provide variant with statically linked modules

agross:
_Motivation:_
As discussed in #2449, some zsh users need zsh modules to use their scripts. I, for example, would like to use my zsh plugin manager [zplug](https://github.com/zplug/zplug). zplug requires the `zsh/system` and `zsh/regex` modules which are not available in the existing zsh package.

The zsh Makefile (the one shipping with zsh itself) detects that dynamic loading is not supported for the target arch and hence will be compiled without any modules except the ones linked statically by default. So, I propose the addition of a second `zsh-static` package (this one), [providing statically linked zsh modules](https://github.com/zsh-users/zsh/commit/626c580148a58d92515c04df607eaa06d86f15f9#L85). Zsh processes with static modules have a higher memory footprint, hence the addition of a new package instead of modifying the existing package.

The `zsh-static` package tries to reuse as much Makefile logic as possible from the `zsh` package. This is my first GNU make patch, so feedback is very welcome.

I was not able to make the new `zsh-static` package conflict with the existing `zsh` package, please see #2893 and advise what I should do (I will amend the submitted commit).

Please note that this PR also includes the update to zsh 5.4.1 (#2889).

_Linked issues:_
#2449, #2893, #2889

### Checklist


ymartin59:
Hello.
Thank you for investigating this old issue. Your proposal is interesting.
It looks like you used symbolic links to refer to files from `zsh-static` to `zsh`

I propose the following alternate way, that is more in `spksrc` philosophy:
- no need for a `cross/zsh-static`
- control dynamic/static build thanks to Make variable in `cross/zsh`
- `spk/zsh-static` depends on `cross/zsh` (so no need to replicate `PLIST` and `digests`)

May you please give it a try ? It should reduce copy/paste to minimal in my opinion.
Best regards

agross:
Hi @ymartin59,

Thanks for the hints.

> control dynamic/static build thanks to Make variable in cross/zsh

Do you mean I should add `POST_CONFIGURE_TARGET` etc. conditionally in `cross/zsh` like this?

```Makefile
ifndef ZSH_STATIC
POST_CONFIGURE_TARGET = link-zsh-modules
...
endif
```

...where `ZSH_STATIC` is set in `spk/zsh-static` and unset in `spk/zsh`?

ymartin59:
Hello. It looks like a good idea to me, yes... if it works properly, we remove a copy of `cross/zsh`, what I would appreciate as maintainer.

agross:
@ymartin59 I updated the PR and I hope I could fulfull your suggestions.

The only standing issue now is how I can make `zsh-static` conflict with `zsh`. (#2893)

ymartin59:
Is that really necessary ? Usually each package has its own "sandbox" as `/usr/local/${PKG_NAME}`... if you do not enforce both to have same prefix, the only "conflict" concerns `/usr/local/bin/zsh` link...
So I would say, if someone decides to install both... well, last install wins its place under sunset at `/usr/local/bin/zsh`.
What do you think about it ?

agross:
Well, I thought the maintainers here would care. I don't know bother because I probably will never install the non-static version. One could also argue that zsh users know what they're doing.

ymartin59:
That is what I did... By the way, I am interested in answering your question about `SPK_CONFLICT`...

ymartin59:
@agross Sorry I did use "rebase and merge" on my PR... as a result, you have to rebase your branch to discard my "duplicated" commit. May you please do that ?

agross:
Done!

ymartin59:
Thanks.

ymartin59:
Packages published

agross:
👍

agross:
@ymartin59 Any ideas when the new/updated packages will show up in the feed? https://packages.synocommunity.com/ returns Bad Request for me. zsh is still at 5.3.1 and zsh (with modules) doesn't show up in DSM. (I pressed refresh several times over the last couple of days.)

ymartin59:
Hello. No idea what may be wrong - it is OK on my DSM (evansport). What is your arch and DSM version ?

agross:
Avoton and latest.

I found a package source monitoring project somewhere and it listed synocommunity as offline. Are you sure that Bad Request is an OK state for the feed page?

ymartin59:
I have just "right now" installed zsh-static on my system. I guess you looked at http://dottoremoe.com/synomon/ But repository is online and usable: https://synocommunity.com/package/zsh-static

agross:
Yes, it was synomon I was referring to. For me the classic zsh package is still at 5.3.1. After I removed the package source, all the SynoCommunity remain displayed and appear to be installable despite failing any installation attempts. WTF?

agross:
Removed the source, verified that all SC packages are no longer displayed (which took a while), readded the SC source, now zsh is at 5.3.1 and the modules variant is still missing. The last thing I'm going to try is reboot the damn DSM.

agross:
A reboot didn't fix it. Using `tcpdump` and the non-HTTPs package source at `http://packages.synocommunity.com/` I found that my DSM requests `http://packages.synocommunity.com/?package_update_channel=beta&unique=synology_avoton_2415%2B&build=15152&language=enu&major=6&micro=3&arch=avoton&minor=1&timezone=Amsterdam&nano=4` to get the package list.

This list only contains `zsh 5.3.1-4`, could you please test that in your browser?

ymartin59:
Dammed ! "avoton" and other x86_64 architectures fail to publish on repository and I missed error message. Fix available #2914. Packages should be visible now.

agross:
Confirmed. Many thanks!

PR #2902, opened 2017-09-01
Borgbackup6

yuroller:
_Motivation:_ Update borgbackup to 1.0.11
_Linked issues:_

### Checklist
- [] Build rule `all-supported` completed successfully
- [] Package upgrade completed successfully
- [*] New installation of package completed successfully

PR #2906, opened 2017-09-04
Borgbackup

yuroller:
_Motivation:_ New borgbackup package
_Linked issues:_ https://github.com/SynoCommunity/spksrc/pull/2902

### Checklist


cross/acl/PLIST
```diff
@@ -0,0 +1,7 @@
+bin:bin/chacl
+bin:bin/getfacl
+bin:bin/setfacl
+lnk::libexec/libacl.so
```
cytec: seems like there is a `:` to much

cross/acl/PLIST
```diff
@@ -0,0 +1,7 @@
...
+lnk:lib/libacl .so
```
cytec: is the leading space before the extension really needed?

yuroller: The space for sure is not needed: it is better to remove it

cross/acl/PLIST
yuroller: I do not know about this: I used spksrc/cross/attr/PLIST as a reference for building this package and it HAS double ::

cytec: well it's wrong there then too. AFAIK there should always only be one reference: https://github.com/SynoCommunity/spksrc/wiki/PLIST-files#example

yuroller:
Outdate PR: updated PR in https://github.com/SynoCommunity/spksrc/pull/2998

PR #2943, opened 2017-10-04
Adds support for Watcher3

### LINKED ISSUE #2879: [Package Request] Watcher 3
Name: Watcher 3
Description: Watcher is an automated movie NZB & Torrent searcher and snatcher. You can add a list of wanted movies and Watcher will automatically send the NZB or Torrent to your download client. Watcher also has basic post-processing capabilities such as renaming and moving.
Website: https://nosmokingbandit.github.io/
Software documentation: https://github.com/nosmokingbandit/Watcher3
Build/Installation documentation:  https://github.com/nosmokingbandit/Watcher3/blob/master/README.md
Source Code: 
License: ECL 2.0

fredrike:
_Motivation:_ Adds support for [Watcher3](https://nosmokingbandit.github.io/)
_Linked issues:_ Closes #2879

### Checklist


spk/watcher3/Makefile
```diff
@@ -0,0 +1,46 @@
+GIT_REPO = https://github.com/nosmokingbandit/Watcher3.git
+
+SPK_NAME = watcher3
+SPK_VERS = $(shell git ls-remote ${GIT_REPO} HEAD | head -c 8)
```
cytec: `SPK_VERS` for those kind of packages normaly looks like this `SPK_VERS = $(shell date +%Y%m%d)` since they most likely contain an in app update mechanism...

spk/watcher3/src/dsm-control.sh
```diff
@@ -0,0 +1,90 @@
...
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PYTHON_DIR="/usr/local/python3"
+GIT_DIR="/usr/local/git"
+PATH="${INSTALL_DIR}/bin:${INSTALL_DIR}/env/bin:${PYTHON_DIR}/bin:${GIT_DIR}/bin:${PATH}"
+USER="watcher3"
+PYTHON="${INSTALL_DIR}/env/bin/python3"
+WATCHER3="${INSTALL_DIR}/share/Watcher3/watcher.py"
+CFG_FILE="${INSTALL_DIR}/var/config.cfg"
```
cytec: that file doesn't exist does it? afaik the file is created on startup but most tools provide a simple configuration file which defines at least port, host and that the browser shouldn't be started see couchpotato for example: https://github.com/SynoCommunity/spksrc/blob/master/spk/couchpotatoserver/src/settings.conf

i guess you'll have to define the directory where to save the database which should be `${INSTALL_DIR}/var/` where all data of the tool should be stored to make sure that data presists during updates

spk/watcher3/src/wizard/upgrade_uifile
```diff
@@ -0,0 +1,6 @@
+[{
```
cytec: That file can be deleted i guess you missed that when copying from CP ;)

spk/watcher3/Makefile
```diff
@@ -0,0 +1,46 @@
...
+MAINTAINER = Fredrike
+DESCRIPTION = "Watcher is an automated movie NZB & Torrent searcher and snatcher. You can add a list of wanted movies and Watcher will automatically send the NZB or Torrent to your download client. Watcher also has basic post-processing capabilities such as renaming and moving."
+ADMIN_PORT = 9090
+RELOAD_UI = yes
+DISPLAY_NAME = Watcher3
+
+HOMEPAGE   = https://nosmokingbandit.github.io/
+LICENSE    = Apache License 2.0
+
+WIZARDS_DIR = src/wizard/
```
cytec: remove this line if you don't have a install wizard

spk/watcher3/Makefile
fredrike: Yes I know that is how it is usually done. But it makes more sense to show wich version (on git) that is used.

spk/watcher3/src/dsm-control.sh
fredrike: According to check manual of wizard3 this file gets created if it does not exist. I think it is good to specify so we know where the file reside.

spk/watcher3/src/wizard/upgrade_uifile
fredrike: I were not sure what the purpose of this file is so I figured it would be best to keep it.

fredrike:
Updated version where I've removed the wizard references.

I've been running Watcher3 on my DS716+II for 16 hours now and it even updates via git nicely 😄.

spk/watcher3/Makefile
cytec: afaik we use the date approach so that it is clear that the version is not in any kind assosicated with the upstream version... using git version implies that you'll have to update the SPK to get a newer version and we'll get users asking for frequent updates again. that was the main reason why we moved to the date style format. To make clear that that's the version/date of the installer not the actual app...

spk/watcher3/src/dsm-control.sh
cytec: it sure is. but most tools have an option to specify where to save the database and some other stuff (mostly called data-dir or something like that) the way you are currently running the app will result in a deleted database when the SPK is updated!

So either use a "base config" which povides those options. or use the flags provided by watcher: ex: `--log ${INSTALL_DIR}/var/ --db ${INSTALL_DIR}/var/database.sqlite` personally i'd prefer to go with the default config as we do this for most SPK's already

spk/watcher3/Makefile
cytec: in addition to that: it can get kind of confusing when you install the SPK and then update the app from inside the app itself (over git i guess) so you'll have another version then the one specified in the SPK ;)

spk/watcher3/src/dsm-control.sh
cytec: as you have already installed and running it. the simplest way would be to get the config of your setup, strip it down to the most basic usage (data directory, port, host, database etc) and put that in the `src` folder and copy it over on installation like cp and others do it ;)

fredrike: Ok, you win :). I've updated config.cfg with bare config now.

ymartin59:
Is port 9090 the "right" value ? I found no reference for `watch3` port: https://www.speedguide.net/port.php?port=9090

ymartin59:
@cytec I do not like "auto-updated" application inside package which version number is generation timestamp. Please follow "wheels" packaging directives https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-packages and generate package stuck to publicly official version or else it is impossible to keep it stable and provide support... no-one will have same version...

fredrike:
Yes, according to https://github.com/nosmokingbandit/Watcher3#installation is port `9090` the default port.

I've removed the extra icon in 0ce7b6a.

I'm not planning to spend more time on this right now (with support for python wheels) so if it can't be accepted in the current stage (just like couchpotato) I suggest we put this issue on hold.

ymartin59:
@fredrike An option may be to explicit tell users that package provide a "live" update thanks to a prefix like "watcher3-installer"... so that "watcher3" name is reserved for explicit version package. @cytec your opinion about such naming change ?

ymartin59:
For standby, I recommend it to, as I am designing a generic service support that allows you to discard both installer.sh and dsm-control.sh

cytec:
@ymartin59 i don't really thing that this is neccessary... that's how most of the Python based apps work anyway (see couchpotato, sickbreard, sonarr etc.) each of those clones the git repo and leaves it there so the user can update the app via the in app updater. That is because otherwise we get users pop up here every 2 days asking for a new SPK for the new version... 

IDK how "big" watcher is.. but maybe it'll be an option to go with a -custom SPK instead?

fredrike:
If we require git to be installed during the installation of `Watcher3` I could probably wrap together an installer that does an `git pull` as part of the installation and then we would have a "fully automatic" install.

cytec:
@fredrike https://github.com/SynoCommunity/spksrc/blob/master/spk/couchpotatoserver-custom/src/installer.sh#L67-L68 that's what i was talking about... the wizzard files are here: https://github.com/SynoCommunity/spksrc/blob/master/spk/couchpotatoserver-custom/src/wizard/install_uifile idk if watcher has a big enough fork community for that to be relevant... but updating should be handled by the app itself anyway!

fredrike:
Yes exactly what I was thinking of (are using that pkg myself with my own fork). Is that of interest because then I'll update the code.

cytec:
go ahead and change it :)

Issue #2904, opened 2017-09-02
Meta: DSM6 compatibility of packages with user privilege issues


m4tt075:
In #2661 @Dr-Bean provided a complete overview on issues with current packages. There is one class of issues related to how DSM6 handles user priviliges (see #2216 for the related discussion and information on how to fix the issue).

Edit: In the meantime, @ymartin59 has developed and merged generic service support for DSM 5+6 packages (PR #2949). As such, the objective is now to convert all the affected packages to use the newly created generic service support.

Here is an overview of the status of those packages.

~I~ We will ~try to~ tackle them successively, but will be slow. If you want to contribute, please just comment which packages you would like to tackle so that we avoid double work. I will update the table regularly. Please report any errors / updates you find.

| PACKAGE                  | STATUS | ISSUE        | COMMENT                                        |
|--------------------------|--------|--------------|------------------------------------------------|
| ~bitlbee~                  |  Retired      | #2552        |   Last update: Dec 2016, Retired with #4208   |
| couchpotatoserver | Published |   |     |
| couchpotatoserver-custom |    Published   |              |                                                |
| deluge                   |    Published    |              |                                                |
| domoticz                 |        |              |   Last update: Mar 2015      |
| ejabberd                 |        | #2260        |                                                |
| flexget                  |        |  #3035         |                                                |
| full-text-rss            |        |              |                                                |
| gateone                  |   Published    |        |                         |
| git-server | Published |  |  |
| haproxy                  |        | #2712        |                                                |
| headphones               |    Published    |              |                                                |
| headphones-custom        |   Published     |              |                                                |
| homeassistant            |    Published    |   #2728      |                                                |
| htpcmanager              |        |              |                                                |
| icecast                  |        |              |                                                |
| jackett | Published |  |    |
| lazylibrarian            |    Published    |  |                                                |
| maraschino               |        |              |                                                |
| memcached                |        |              |                                                |
| monit                    |        |              |                                                |
| mosquitto | [Published](https://synocommunity.com/package/mosquitto) |  |   |
| museek-plus              |  Retired      |              |                                                |
| mylar                    |    Published    |              |                                                |
| nzbget | Published |   |  |
| nzbhydra | Published |   |  |
| nzbmegasearch            |        |              |                                                |
| plexivity                |        |              |                                                |
| plexpy-custom            |   Published     |              |                                                |
| pyload                   |        | #2259        |                                                |
| python | Published |   |  |
| radarr | Published |  |  |
[...]

PR #2949, opened 2017-10-07
Generic service support for DSM 5 and 6

ymartin59:
- Generic installer and start-stop-status scripts
- User creation and run-as privilege file
- Share folder creation with group permissions
- Target both DSM 5 and 6

Based on BenjV proposal #2904


m4tt075:
@ymartin59 : Thanks for getting this going. I have cherry-picked your commit and started playing around with your demoservice. Stumbled pretty much at the doorstep though. Currently, cross-compiling fails with...
> ===>  Generating service scripts for demoservice
> /bin/sh: 1: cannot create /service-setup: Permission denied
> ../../mk/spksrc.service.mk:44: recipe for target '/service-setup' failed
> make[1]: *** [/service-setup] Error 2
> make[1]: Leaving directory '/home/walter/builds/spksrc/spk/demoservice'
> ../../mk/spksrc.spk.mk:423: recipe for target 'arch-evansport-6.1' failed
> make: [arch-evansport-6.1] Error 2 (ignored)
But you probably know this...

ymartin59:
@m4tt075 This version is now building properly... but generated scripts still need intensive testing...

m4tt075:
@ymartin59 Cheers! OK, I will play around with your demoservice and then try to apply your approach to the TVH package. I have got an XPenology 5.2 and an evansport-6.1 environment for testing. Will report back with the results.

m4tt075:
@ymartin59 Some progress!

1. Cross-compilation goes through smoothly now
2. First tests on XPE 5.2 (basically clean, only TVH and nano packages installed):
- Manual installation correctly opens up the WebUI Interface (spk seems to be stucturally intact!)
- I just accepted all defaults (`/volume1/downloads`, despite non-existent, and `sc-download`, not being existent) and proceeded through the screens
- After confirming the last screen (leaving "Run after installation" checked) the actual installation fails with something like "Demoservice could not be installed" (no further notice)
- I manually created the `/volume1/downloads` directory and repeated the above, including variations of checking / unchecking "Run after installtion", but same result, the installation fails

3. Diagnostics so far (via ssh to the XPE system):
- No log file in the `/tmp` file
- No package-related files in `/usr/local/`
- No new user created in DSM or in `/etc/shadow`
- No new user group created in DSM or in `/etc/group`

This is how far I got this time. Anything (else) you want me to check or look for?

m4tt075:
@ymartin59 Repeated the above test on my DSM 6.1 system. Same result / failure notice. No log file, no further hints why it fails. 

I then unpacked the spk which was generated and realized that there is no conf folder and that the privilege file is missing. Should not be an issue with 5.2, but might well be one with the 6.1 system.

ymartin59:
@m4tt075 Thanks for testing. I have fixed `conf/privilege` generation and I am debugging installer scripts. It should work soon.

ymartin59:
@m4tt075 Some progress... still buggy features but install and log activity

ymartin59:
For the moment, installer script faced the following bug on DSM 5.2: `servicetool --remove-configure-file --package` seems to succeed but without effect, as reinstalling again package fails with message claiming port is already reserved for another application !

I have not testing on DSM 6.x yet

ymartin59:
Tested on DSM 5.2 and works properly except uninstall do not remove service configuration, leaving port marked as used.

m4tt075:
@ymartin59 Great stuff! I can confirm what you have written. Here some more detailed observations from my testing process:

1. Cross-compilation runs smoothly
2. First-time installation on XPE 5.2 (basically clean, only TVH and nano packages installed):

- Manual installation correctly opens up the WebUI Interface
- I just accepted all defaults (/volume1/downloads, despite non-existent, and sc-download, not being existent) and proceeded through the screens
- Installation finishes successfully
- Running the demoservice correctly opens up the package folder and gives me access to the relevant files. Beautiful! 
- The `downloads` folder is created with `root:users` ownership.
- `demouser` appears as user via DSM (this is still 5.2 after all)
- `sc-downloads` appears as group via DSM (as above)
- Manual checking of `/etc/passwd`, `/etc/shadow` and `/etc/group`: Everything looks as it should be
- The `demoservice.sc` file is correctly installed into `/usr/local/etc/services.d/`

2. Uninstalling the package 

- Finishes cleanly
- `Demoservice` user is correctly removed
- `sc-downloads` group is correctly removed
- The `demoservice.sc` file _incorrectly survives_ in `/usr/local/etc/services.d/`, which prevents a re-installation, as you have been describing.
- Manually removing the sc-file from the above directory works around the problem.

2. I repeated step 1) with different (non-default) group name and download directory. Result as above. Everything working as intended.

3. Updating the package (simulated by simply manually re-installing the package without uninstalling it first)

- WebUI confirms that the package has been updated successfully
- However, afterwards the package cannot be executed anymore
- Checking why, I found that the `demoservice` user and `sc-download` group are gone. They are removed, but not re-created. Consequently, the permissions of the package folders are screwed up as well.
- I think this is just because something goes wrong with the sequence of executing commands during the update process.

4. Just an observation: The shared folders that are created survive uninstalling the packages, but I assume this is intended behavior.

5. I manually played around with ´servicetool´ command to figure out what goes wrong. I figured out that it fails if you specify the sc-file including it's path. However, it works if you just skip the path. Correct usage is `/usr/syno/bin/servicetool --remove-configure-file --package demoservice.sc`. This might be the culprit...

Hope this helps. Again, many thanks for taking this on!

ymartin59:
@m4tt075 Many thanks for your testing. I fixed upgrade process and service configuration handling (with DSM 6.x in mind)
You're right about observation "4.", it is intended not to delete share folder.
I now have to adapt and test for DSM 6.x

m4tt075:
@ymartin59 I have started working on adapting the TVH package. This is really cool and fun! :-)

Some questions:
1. TVH runs on two ports (http and htsp). In the `tvheadend.sh` file, I define both as `PORT` and `PORT2` (the latter added by me). Both should be blocked in the `*.sc` file, but only the first (`PORT`) appears there. It is the same with the `service-setup` script. Is there a way to include handling of two (or more) ports?
2. The TVH package contains a `PKG_CONX` file in the `conf` folder. There can be other similar files in that folder (see documentation). These files should be copied into the `conf`folder of the final package but are lost now. Can you fix this, please?
3. The old TVH `privilege` file contained entries for the actions `start` and `stop` to run as `root`. This is gone now. Is this correct behavior / not needed (anymore)?
4. The old TVH pacakge creates a `sc-tvheadend` user (like all the packages following the original @Dr-Bean approach). These are `legacy users` now and should be removed. Can you include those in the corresponding `remove legacy user` routines?
5. What do you want us to PR against, so that you can take a look at our adaptations? Are you going to merge you PR here into DSM6 and we PR against the DSM6 branch then or do you want us to PR against the corresponding development branch in your repository?

ymartin59:
Thanks a lot for feedback.
1. In that case, you still have to use FWPORTS variable with your own made sc file. If I made no mistake it should land to package at proper place
2. Sure a bug I have to fix
3. It is intended, service process is generally not expected to run as root... except if TVH has specific requirements to do so
4. Based on discussions with BenjV, I decided to keep service account with same name as package... I still have work to do for upgrade path from already published packages, some commands to adapt account attributes. As far as I know, none has been published with "sc-xxx" account.
5. As "DSM 5 to 6 upgrade path" is not fully clear, I would ask to wait for it before processing packages...

m4tt075:
@ymartin59 Cheers! All clear! @4: You are right, not really, but pretty much all recent TVH versions floating around the internet are using the `sc-user`. No problem though, I will simply include the removal of the corrsponding `sc-user` and `sc-group` in the `service-setup.sh` script for it to remain proprietary to the TVH package. 

One more observation from working on the `installer` file: The function `syno_group_remove` is defined but never used...

ymartin59:
@m4tt075 I have tested `src/conf` copy into package and it works as expected. I guess `CONF_DIR = src/conf/` is not present in your spk Makefile. It is required as soon as package provides specific content, the same way as for wizard directory and "app" directory.

At package uninstallation, group is removed if there is no members left but I missed to reuse this function... https://github.com/SynoCommunity/spksrc/pull/2949/files#diff-3b7503318148d2dbdff9b9fb590f64c2R255

I will improve it with group as parameter so that it can be reused in package specific hook functions.

Again I really appreciate your testing and feedbacks.

m4tt075:
@ymartin59 Right (again). `CONF_DIR` was missing. Works correctly now.

Having `group` as a parameter for `syno_group_remove` would be great. Are you going to do this with `user` for the corresponding functions as well?

The only piece I'm still struggling with is the `FWPORTS` file. I have reintroduced the variable in the `spk/Makefile` and point it at `src/${SPK_NAME}.sc`, where the sc-file is located. But somehow it is not picked up. It is not copied into the final package anymore. The `installer` file, that is generated through your SERVICE routines, still points at an sc-file in the non-existing `conf/resource` folder. Do I have to put the sc-file in an `conf/resource` folder to make it work or do I have to copy the file manually in one of the installation routines? Or am I missing something else?

ymartin59:
@m4tt075 You probably have an old branch version script. I have fixed `FWPORTS` support, it should work now.

m4tt075:
@ymartin59 I can't get the propiretary `FWPORTS` definition to work. Could you please take a look at the `spk/Makefile` I'm using? Here it is:
```
SPK_NAME    = tvheadend
SPK_VERS    = 4.2.4
SPK_REV     = 12
SPK_ICON    = src/tvheadend.png
DSM_UI_DIR  = app
BETA        = 1

# DEPENDS = cross/busybox cross/$(SPK_NAME)
DEPENDS = cross/$(SPK_NAME)

MAINTAINER   = m4tt075
DESCRIPTION  = Tvheadend is a TV streaming server and recorder for Linux, FreeBSD and Android supporting DVB-S, DVB-S2, DVB-C, DVB-T, ATSC, ISDB-T, IPTV, SAT IP and HDHomeRun as input sources. Tvheadend offers HTTP, HTSP and SAT IP streaming.
# ADMIN_PORT = 9981
RELOAD_UI    = yes
DISPLAY_NAME = Tvheadend
CHANGELOG    = "Update for full DSM6 compliance"
HOMEPAGE     = https://www.lonelycoder.com/tvheadend/
LICENSE      = GPL v3

WIZARDS_DIR  = src/wizard/
CONF_DIR     = src/conf/
FWPORTS      = src/${SPK_NAME}.sc

# 'auto' reserved value grabs SPK_NAME
SERVICE_USER         = auto
SERVICE_WIZARD_GROUP = wizard_group
SERVICE_WIZARD_SHARE = wizard_download_dir
SERVICE_SETUP        = src/service-setup.sh
# Using standard ports 9981 for http and 9982 for htsp. Adapt SERVICE_COMMAND below if needed
SERVICE_COMMAND      = '$$\{SYNOPKG_PKGDEST\}/bin/tvheadend.sh 9981 9982 $$\{PID_FILE\}'

# INSTALLER_SCRIPT = src/installer.sh
# SSS_SCRIPT       = src/dsm-control.sh

# INSTALL_PREFIX = /usr/local/$(SPK_NAME)

PRE_COPY_TARGET = tvheadend_install

# BUSYBOX_CONFIG = usrmng
# ENV += BUSYBOX_CONFIG="$(BUSYBOX_CONFIG)"

.PHONY: tvheadend_install
# Replace standard copy/install targets
tvheadend_install:
	rm -fr $(STAGING_DIR)
	mkdir $(STAGING_DIR)
	mkdir --parents $(STAGING_INSTALL_PREFIX)
	install -m 755 -d $(STAGING_INSTALL_PREFIX)/bin
	install -m 755 src/tvheadend.sh $(STAGING_INSTALL_PREFIX)/bin
	install -m 755 -d $(STAGING_DIR)/app
	install -m 644 src/app/config $(STAGING_DIR)/app/config
	install -m 755 -d $(STAGING_DIR)/var/accesscontrol
	install -m 755 -d $(STAGING_DIR)/var/passwd
	# The following filenames stem from an arbitrary (valid) combination created by TVH
	install -m 644 src/accesscontrol.json $(STAGING_DIR)/var/accesscontrol/d80ccc09630261ffdcae1497a690acc8
	install -m 644 src/passwd.json $(STAGING_DIR)/var/passwd/a927e30a755504f9784f23a4efac5109
```
What am I doing wrong?

ymartin59:
I guess it does not work because of commented `ADMIN_PORT`. My fault...

m4tt075:
@ymartin59 Thanks, and no worries. For every mistake you make, I make 20. I still have not been able to get the `FWPORTS` running properly, but I have done some further testing. Here some more observations:
1. Before I install anything, I always extract the spks on my Windows laptop, to compare the integrity of the package itself. Whenever I do this, 7-zip asks me whether I would like to overwrite the files in the `conf` folder with new ones. The "old" ones contain content, the "new" ones don't (0 kb file size), so I negate the question. Not sure what this means, but wanted to share...
2. XPE 5.2: I tried to upgrade an old standard TVH package, which was installed via the `busybox` method. The upgrade seemed to work. But uninstalling the package afterwards did not, as the `tvheadend` user could not be removed properly. I figured out that users, who were installed via the `busybox` method cannot be removed via the new approach. I had to reinstall an old package to get the commands, before I was able to clean the system again. I believe we have to keep the `busybox` dependency, at least initially, in order to properly remove legacy users and group. 
3. XPE 5.2: A clean installation seems to work well in terms of user / group creation and permissions.

For transparency, I have not yet been able to run my TVH package, but that's likely related to the 20 mistakes I introduce while you make 1...

m4tt075:
@ymartin59 I finally managed to get `FWPORTS` working. The only thing needed was to move the `tvheadend.sc` file from the `src` to the `conf` folder. Now it is picked up correctly. Most files in my package are installed correctly now. Apparently the `FWPORTS` variable in the `spk/Makefile` does not matter anymore.

Two more questions though:
- On a 5.2 system, is the launch command specified in the `SERVICE_COMMAND` executed as `root` or as `[username]` as defied in the package?  
- Should the actual command running the service (as demon for TVH) always be encapsulated in a `[username].sh` script (also to generate the pid-file) or should it rather be inserted as `SERVICE_COMMAND` itself?

Background: I have tried the encapsulated version following your demoservice example, but ran into issues with lacking access rights to certain configuration folders. This is all still on a XPE 5.2 system.

ymartin59:
@m4tt075 On DSM 5.2 installer script `chown`s only `var` folder. If your code requires write access to other location, you have to do chown in postinstall script (ideally only on DSM 5)
Yes, `SERVICE_COMMAND` process runs as `SERVICE_USER` because of `su` command on DSM 5. https://github.com/SynoCommunity/spksrc/pull/2949/files#diff-9fe71615984368bf44c02ed52c3b1fceR26
DSM 5 does not understand `conf/privilege` file, which is only for DSM 6.
If `SERVICE_COMMAND` process is able to produce `PID_FILE`, then there is no need for a shell script like `demoservice.sh` I have only written for demonstration purpose.

ymartin59:
@m4tt075 I just fixed `FWPORTS` with your specific situation: `CONF_DIR` is set and `ADMIN_PORT` is not. I tested with `src/demo.sc` and it gets copied as `conf/demoservice.sc` as expected: https://github.com/SynoCommunity/spksrc/pull/2949/files#diff-c265147645e98b75c07672b914840285R141

Safihre:
So this is ready and @BenjV also likes it?
I can try and adapt the SABnzbd package for it?

BenjV:
@ymartin59 and @m4tt075 
As far I can see at the moment you guys did a great job.

@Safihre
If you create a SABnzbd package I will happly review and test it because it is a noarch package and can be installed on any Synology Nas.

Safihre:
@BenjV Not really noarch due to the compiled components `unrar`, `par2cmdline` and Python C-extension `sabyenc`. But that's off-topic, I'll get started on it 👍

ymartin59:
For information, I have plan to disjoin ADMIN_PORT and an additional SERVICE_PORT to generate service configuration (.sc) for firewall. (found it relevant when processing mosquitto)

ymartin59:
Thanks for this good team work... I think we are on the right path to get easier job with all these packages maintenance.
I have provided reference documentation on wiki with a (draft) migration guide:
- https://github.com/SynoCommunity/spksrc/wiki/Service-Support
- https://github.com/SynoCommunity/spksrc/wiki/Package-Migration

As a first example, I have published two PRs for Mosquitto package migration: #3025 (master) and #3028 (dsm6)

Safihre:
@ymartin59 Thanks!
The `Service user transition` part seems a bit of a problem to me. What if a user doesn't get this intermediate package but goes straight from an old package to the then published newer package?
There is no way to do this in 1 go? From @BenjV comment here I understand that just using the `syno` commands to remove the old `busybox` user/group should work, even if it errors?

ymartin59:
No the right place for such a question. See https://github.com/SynoCommunity/spksrc/issues/2904#issuecomment-348814928

Issue #2793, opened 2017-06-11
Support NPM modules cross-compiling to package NodeJS application


ymartin59:
After request #2769 to package a NodeJS application, it sounds interesting to provide support for NPM modules - close to what is available for Python Wheels.

As a example for implementation, here are instructions to cross-compile targeting OpenWRT (Google lead me there): https://github.com/netbeast/docs/wiki/Cross-Compile-NPM-modules

ymartin59:
Candidate package `aria2` may include webui-aria2 https://github.com/ziahamza/webui-aria2

hgy59:
we don't support nodejs packages anymore

PR #2959, opened 2017-10-12
Add Aria2 build

jreese:
_Motivation:_ Adds build script for `aria2c` download manager.
_Linked issues:_ #3269 

Only tested on DSM 6.1 on the ds916+.

### Checklist


ymartin59:
Thanks for this work. As a small CLI tool, I suggest to include it in one of planned multi-tool packages, see #2261
Such a PR to `dsm6` will not be merged, as this branch is reserved to support for DSM 6.x. Please rebase your branch to `master`.

spk/aria2/Makefile
```diff
@@ -0,0 +1,24 @@
...
+DISPLAY_NAME = aria2
+STARTABLE = no
+BETA = 0
+CHANGELOG = Initial version
+
+HOMEPAGE   = http://aria2.github.io
+LICENSE    = GPLv2/GPLv3
+
+INSTALLER_SCRIPT = src/installer.sh
+SSS_SCRIPT       = src/dsm-control.sh
```
ymartin59: Please remove `SSS_SCRIPT` and `dsm-control.sh` file as application does not run as a service

spk/aria2/Makefile
```diff
@@ -0,0 +1,24 @@
+SPK_NAME = aria2
+SPK_VERS = 1.31.0
+SPK_REV = 1
+SPK_ICON = src/aria2.png
```
ymartin59: What I see on screen is `tmux` and `top`... If aria is included in a CLI multi-tool package, no need for an icon by the way.

cytec:
@ymartin59 not so sure about the CLI Multipackage for that one. It's a pretty solid/big download client... I'd compare it more to something like rutorrent or something like that than "a small cli tool" so i think it has a right for his own SPK. Maybe we could even include https://github.com/ziahamza/webui-aria2 which seems to be a pretty solid webui for aria2 this days...

ymartin59:
@cytec OK I agree with you. It may be interesting to include webui in same package when NodeJS cross-compilation support #2793 will be available for packages.

PR #2965, opened 2017-10-19
Fix package icons generation failure from SPK_ICON

ymartin59:


m4tt075:
@ymartin59 Thanks a lot. PR and commit title might need updating, but I can confirm this fixes the compilation issue I had.

PR #2969, opened 2017-10-21
[TVH] Update to 4.2.4 (Release) and libhdhomerun to 20170930

m4tt075:
### Motivation
Update to latest TVHeadend maintenance release 4.2.4, including bump of libhdhomerun to 20170930

### Checklist
- [x] Build rule `all-supported` completed successfully with `5.2 toolchains`
- [x] (Optional) Build rule `all-supported` completed successfully with `6.1 toolchains` except with `hi3535` (not expected to run) and `rtd1296` (which probably requires care by some platform owner)
- [x] Package upgrade completed successfully (on `evansport-6.1` system)


ymartin59:
@m4tt075 tvheadend package is not declared as "startable" with `STARTABLE = yes` variable in `Makefile`. Do I guess right it should be set ?

m4tt075:
@ymartin59 To be honest, I have never heard of this variable before. It has not been set in the original packages which I started working from. Looking at the development guide, "yes" should be the right value, but "yes" is also what this variable defaults to if it is not set. Therefore, not sure it is needed. If you'd prefer to have it explicit, please just let me know and I'll add it.

PR #2970, opened 2017-10-22
Properly reflect status for Radarr in Package Center

cuki:
_Motivation:_ In case `radarr` user was created, with user home service enabled, prior to installation in order to have __.config__ directory in home directory.
Status is not represented as expected, it always is shown as "Stopped", despite the fact it would successfully run Radarr in the background upon starting the service.
_Linked issues:_ N/A

### Checklist


ymartin59:
Question: why PID file is not written at expected place ? Is there an option lacking in application startup command line ?

ymartin59:
Is that trouble related to a register issue ? Is trouble visible with DSM 5 and/or DSM 6 ?

cuki:
@ymartin59 In fact PID file is in the expected place, which is the home directory of the `radarr` user specified in __/etc/passwd__.

Only difference is that when you create `radarr` user prior to package installation is that the home directory is `/var/services/homes/radarr`, instead of `/usr/local/radarr/var` when user is [added on installation](https://github.com/cuki/spksrc/blob/7bee8d37f8d17f9a6558e927507d6c69edc3d441/spk/radarr/src/installer.sh#L70).

I noticed the issue when I was configuring Radarr (and Sonarr) on my Synology DiskStation 216j (running DSM 6.1.3-15152 Update 8), as far I'm aware there are not any similar (registered) issues reported in the repository.

Difference between Radarr and Sonarr was that only Sonarr properly [reflected the service status](https://github.com/cuki/spksrc/blob/7bee8d37f8d17f9a6558e927507d6c69edc3d441/spk/sonarr/src/dsm-control.sh#L11-L12), despite being installed the exact same procedure.

Only issue with Radarr is that it does not respect predefined [config.xml](https://github.com/cuki/spksrc/blob/7bee8d37f8d17f9a6558e927507d6c69edc3d441/spk/radarr/src/config.xml) which is [copied](https://github.com/cuki/spksrc/blob/7bee8d37f8d17f9a6558e927507d6c69edc3d441/spk/radarr/Makefile#L44) into the installation directory (`/usr/local/radarr/var/.config/Radarr`), but creating a new configuration file in home directory (`/var/services/homes/radarr/.config/Radarr`).
As a result starting the service listening to (default) port 7878, instead of (predefined) port 8310.

BenjV:
The PID file is in the wrong place.

It should be in the installation folder of the package e.g. /@appstore/radarr and not in the home folder of a random user (not even if that user is called "radar").
Sonarr puts it in the correct place and Radarr not.

As of DSM 6 all packages are and should be run as a "hidden" user with no home folder for obvious security reason as prescribed by Synology.

cuki:
@BenjV Both Radarr and Sonarr place the PID file inside the home directory of the (hidden) user.
Home directory depends on how the user was created as explained in the [previous comment](https://github.com/SynoCommunity/spksrc/pull/2970#issuecomment-338782877).

Reason why I created my own `radarr` user prior to installation, as described in pull request motivation, is to separate the __.config__ directory from the installed package.
In case or re-installation that the configuration is persisted, instead of forcing __.config__ to be located in the installation directory.

BenjV:
The start-stop-status script which is used by the DSM package centre does not look for the PID file in the home folder of a user but in the folder where the package is installed.
So putting it elsewhere will not work.
Of course is putting the pid file in the user home has nothing to do with the location of the config files that you want to be persisted.
Normally an applicatie should ask the user for a location where to backup those files to for reinstallation purpose.
The package self could use the environment variable SYNPKG_TEMP_UPGRADE_FOLDER to put such files during the upgrade process, so the application does not has to bother with that.

Creating a user via the etc/passwd file will not work anymore, because DSM 6 will remove those users at a reboot because it has its own user database.

cuki:
@BenjV I'm not sure I follow what you are trying to say or understand what is the problem with my proposed solution.

The whole point of the committed changes in this pull request is to check for PID file in __.config__ directory within the (`radarr`) home directory, similar to [Sonarr start-stop-status script](https://github.com/cuki/spksrc/blob/7bee8d37f8d17f9a6558e927507d6c69edc3d441/spk/sonarr/src/dsm-control.sh#L11-L12) which also checks the (`nzbdrone`) home directory.

As already explained before the home directory (`$HOME`) depends on how the user was created.
If the user was created prior to installation (via _Control Panel / User_) home directory will be `/var/services/homes/radarr`, if user is created at package installation (see [installer.sh](https://github.com/cuki/spksrc/blob/7bee8d37f8d17f9a6558e927507d6c69edc3d441/spk/radarr/src/installer.sh#L70)) home directory will be `/usr/local/radarr/var`, which is in the installation directory (`/usr/local/radarr`).

Both [Radarr](https://github.com/Radarr/Radarr/blob/520836f475cd5ed79e72b663c8b3cceb820a5f4f/src/NzbDrone.Common/EnvironmentInfo/AppFolderInfo.cs#L27) and [Sonarr](https://github.com/Sonarr/Sonarr/blob/1d2b35fdde90bb3dd1d4adfe7b82c3cdd766e457/src/NzbDrone.Common/EnvironmentInfo/AppFolderInfo.cs#L27) use [`Environment.SpecialFolder.ApplicationData`](https://ibboard.co.uk/Programming/mono-special-folders.html), that maps to `$HOME/.config`, which is either `/var/services/homes/radarr/.config` or `/usr/local/radarr/var/.config` depending on how the user was created.

My Synology setup is running DSM 6.1.3-15152 Update 8 on DS216j, which does not remove the `radarr` user from __/etc/passwd__, which was created via _Control Panel / User_.

cuki:
@ymartin59 @BenjV Could one of you please answer my question from the previous post, what is exactly the problem with the proposed solution?

BenjV:
Quite simpel.
All packages are created by a uniform script of the spksrc framework.
It is not possible to deviate from that for one package.

One part of the package is a set of script that is used by DSM to keep track of the package, so the interface with the package centre.
One of them is the "start-stop-status" script which is used by the package centre.
This script uses the pidfile to check the status and so the pidfile for all packages must be in the same folder and not randomly scattered across.
That location is for all packages the same e.g. where the package is installed.

It is just not possible to make this random for all packages.

cuki:
@BenjV That's my point exactly with these changes made, to ensure this uniform behavior.
Especially considering Radarr is based on Sonarr codebase it seems only fair they are the same.

I really do not understand what is the problem with my proposed solution, just see for yourself the non-uniform behavior by comparing these files (on current master):
- [spk/**sonarr**/src/dsm-control.sh](https://github.com/SynoCommunity/spksrc/blob/2908002d380be0ed3c21cc2b353ffe7c6755f86f/spk/sonarr/src/dsm-control.sh#L11-L12)
- [spk/**radarr**/src/dsm-control.sh](https://github.com/SynoCommunity/spksrc/blob/2908002d380be0ed3c21cc2b353ffe7c6755f86f/spk/radarr/src/dsm-control.sh#L11)

BenjV:
It is quite simple.
Lots of people don't use homefolders at all, so those people then cannot use this packages anymore.
Also is it bad practice to create a package that needs for the user to create a homefolder first.

The fact that at the moment there are different implementation make things hard to support.
That is one of the reason there are not DSM 6 packages yest available at the moment, because it is a huge job to make all those package supporting DSM 6.

cuki:
Everything seems to be "quite simple", however every time I ask for explanation or answer I receive an unclear or misunderstood response to the actual question.

I've explained (in detail) already multiple times in previous comments that home directory is dependent on the way how the user was created. Also that both Sonarr and Radarr store their configuration and PID file in the home directory (within __$HOME/.config__ folder).

I understand that DSM6 branch is under development and is a huge job, however this is a small fix and will resolve the issues with Radarr that users have __now__.

@BenjV Below I rephrased my question and please read my previous comments again if possible before answering.
> "What is wrong with the proposed solution which ensures the uniform behavior between Sonarr and Radarr packages?"

BenjV:
Asking again won't get you a different answer.
You don't seem to be able to grasp my answer so that ends the discussion.

cuki:
@BenjV I do understand your answers, which seem to be in the main interest of this DSM6 branch.

Despite the fact that Sonarr package has the __exact__ same solution as changes to the pull request for Radarr package, you are providing conflicting answers that it should not be looking for the PID file in the `$HOME` directory, but yet the packages should have uniform behavior.

That said, you can understand my confusion to "grasp" your conflicting answers.
Either way it seems you are refusing the proposed solution, most likely due to some potential issues in the DSM6 branch, that is still in development and not relevant to the __current__ state of the packages.

What would you suggest as alternative to resolve this issue for the current state of the package that is being used by people __now__?

cuki:
@ymartin59 @BenjV In response to my previous post, what is the alternative proposition you have to address the problem, considering this pull request / solution is not accepted.

I'd like to see if I can address the problem __now__ for users which already have `radarr` profile on their NAS (mainly in order to separate the __.config__ folder from installation folder) and properly reflect the service status.

BenjV:
Change the location of the pid file to the installation folder, this will resolve the status issue, so put the question at the developers of Radarr.
Maybe you should point out to them that a user home is not the correct place to put configuration data and/or pid files in.
A user home folder is for the users and not for applications and as I already said, on DSM 6 package will run as users without a user home and that is a Synology decision.

It is also a bad idea to solve such issue's per package when it should en will be an uniform solution.

Upgrading a package will save the config files and reinstate them afterwards so that is not a problem if they are put in the correct place.

If you want to be able to remove a package and do a fresh install, then the application itself should have a functionality to backup the config file and restore it later on.
I don't see why that should be a concern of the package distribution (eg.g the SynoCommunity).

cytec:
@cuki you are right, Sonarr uses the exact same approach you are using in this PR however @BenjV is right too. There will be no home user in DSM6 and therefore merging this change just to revert it again later on imho doesn't really make sense. Also, like @BenjV mentioned: the user folder isn't the place to store such things, sonarr AND radarr should at least give you the ability to provide such a folder per env flag or cli flag or what ever! 

The storage folder vor DSM Packages is the packages var folder and that should not be changed imho! also since those packages already use `start-stop-deamon` they should use the pidfile option as well, see: https://github.com/Sonarr/Sonarr/wiki/Autostart-on-Linux#method-2-lsb-style-init-script-debianubuntu which also put's the pidfile in a non user folder ;)

PR #2982, opened 2017-10-31
Update digests for comcerto2k toolchain

Filimindji:
_Motivation:_ To be able to compile using the comcerto2k toolchain
_Linked issues:_ N/A


ymartin59:
Superseeds by #3002

Issue #2819, opened 2017-07-12
Package Missing - broadwell architecture


mysteriouss:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_
_Package Version:_

_NAS Model:_ DS3617xs
_NAS Architecture:_ x86_64
_DSM version:_ 6.1.2

### Expected behavior
Tell us what should happen

Packages like tmux will be seen

### Actual behavior
Tell us what happens instead

Only a few package can be found

### Steps to reproduce
_1._ Open Package Center 
_2._ Add Source : packages.synocommunity.com
_3._ Only a few packages..
<img width="1440" alt="screenshot 2017-07-12 4 54 24" src="https://user-images.githubusercontent.com/985230/28109914-4eaaa07a-6723-11e7-8680-cc96c531ca7e.png">

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
```
Insert the package log here
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here
```

---

```bash
$ curl -v 'http://packages.synocommunity.com'
* Rebuilt URL to: http://packages.synocommunity.com/
*   Trying 104.31.94.179...
* Connected to packages.synocommunity.com (104.31.94.179) port 80 (#0)
> GET / HTTP/1.1
> Host: packages.synocommunity.com
> User-Agent: curl/7.43.0
> Accept: */*
>
< HTTP/1.1 400 BAD REQUEST
< Date: Wed, 12 Jul 2017 08:12:34 GMT
< Content-Type: text/html
< Content-Length: 192
< Connection: keep-alive
< Set-Cookie: __cfduid=dbdfb980f435a119efd5c9af4b8dda4ef1499847154; expires=Thu, 12-Jul-18 08:12:34 GMT; path=/; domain=.synocommunity.com; HttpOnly
< CF-Cache-Status: MISS
< Server: cloudflare-nginx
< CF-RAY: 37d289c975a32e21-NRT
<
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>400 Bad Request</title>
<h1>Bad Request</h1>
<p>The browser (or proxy) sent a request that this server could not understand.</p>
* Connection #0 to host packages.synocommunity.com left intact
```

```bash
$ dig packages.synocommunity.com

; <<>> DiG 9.8.3-P1 <<>> packages.synocommunity.com
[...]

Issue #2918, opened 2017-09-19
Update Python to 2.7.14


Safihre:
Maintenance release for the 2.7 series. Not urgent.

Should not break any packages, but maybe the compile patches changed and I don't have a Synology device to test on myself.

Safihre:
I would suggest to also update the `cross/wheel` and `cross/setuptools` packages with the python update, just to be complete.
https://pypi.python.org/pypi/wheel
https://pypi.python.org/pypi/setuptools

Safihre:
Done.

PR #2984, opened 2017-11-01
[WIP] Python 2.7 fixes

m4tt075:
### Motivation
Cross-compilation of python-2.7 fails. This is an atempt to fix this. Currently WIP. Don't merge!

### Linked issues
#2819 

### Status
- Compilation of cryptography module failed with `cffi.api.CDefError: cannot parse "typedef int __dotdotdot__  time_t;"`: Fixed by updaing `cffi` cross package to `1.10.0`
- Source for cross-package `libsodium 1.0.13` cannot be found: Deprecated, fixed by updating cross package to `1.0.15`
- Stripping `lib/python2.7/lib-dynload/*.so` fails.

### Checklist


m4tt075:
@ymartin59 The stripping fails, because the `lib-dynload` folder does not contain any `*.so` files. I don't know whether they should be there and are missing or whether the corresponding stripping entry can just be removed from the `PLIST` file. Any insights appreciated...

Safihre:
While you are doing this, maybe you can also update python itself to 2.7.14 (https://github.com/SynoCommunity/spksrc/issues/2918).
And maybe also setuptools and wheels `cross` packages?

m4tt075:
@Safihre I have just tried, but those updates introduce additional issues as the existing patches don't apply anymore. The python configuration files have changed significantly. The code is complex and I don't understand it well enough to know what needs to be adapted. Sorry...

Safihre:
@m4tt075 Ah, seems we are doing double work. Maybe you saw I managed to get it compiled by hustling a number of modules around?

@ymartin59 Maybe you can make 1 big "DSM6 progress issue" with a checklist? What steps need to be taken and in what order to get things up and running?
To me it seems first your service-support, then get the website to support DSM6 and then we can start modifying packages 1 by 1? Is that correct?
I really want to help but would help to have a plan so we can all do our part :)

m4tt075:
@Safihre Yes, I saw it, and believe you have made way more progress than I did. Do you want to PR your changes? I'd be happy to close my PR here and focus on the DSM6 integration again instead. 

W.r.t. your question to @ymartin59, this is exactly what he has done. Just check the "Projects" tab at the top of this page and you should see it.

Safihre:
My implementation is lacking testing. Similar to your approach I just identified which modules failed and excluded those. This is of course not a solution for the long term. 
I think that to properly help this project I should setup my own DSM VM's instead of just messing around with Travis building. I noticed me compiling packages for DSM 6.1 using Travis might be counter productive and I would like to be of real help instead :) I hope to have time to setup these VM's soon!

m4tt075:
@Safihre I have made exactly the same experience you just went through. Setting up a testing VM certainly makes sense. I'm sure your help is appreciated, whenever you are ready and able to contribute.  :)

Safihre:
So now I am a bit stuck. Maybe @BenjV you know this.
While compiling Python it fails to compile `ctypes` due to this error:
```
In file included from /spksrc/spk/sabnzbd/work-bromolow-6.1/Python-2.7.13/Modules/_ctypes/libffi/src/closures.c:119:0:
/spksrc/spk/sabnzbd/work-bromolow-6.1/Python-2.7.13/Modules/_ctypes/libffi/src/dlmalloc.c:1454:35: error: ‘__PTHREAD_SPINS’ undeclared here (not in a function)
 static MLOCK_T magic_init_mutex = PTHREAD_MUTEX_INITIALIZER;
```
Isn't this something the Synology toolchain environment should provide?

Safihre:
I just can't get it to work. Who knows a lot about compilers? @cytec @Dr-Bean 
There is something different about the DSM6 toolchains, whatever I do it keeps failing to build `_ctypes` due failing to build `libffi`.
I checked the command that compiles `_ctypes` and it's identical between DSM5.2 and DSM6 tool-chains. So something in the update of GCC or glibc breaks the build.. But where to start?

ymartin59:
As Synology publish GPL sources they have used to build NAS, and python is delivered, I would have expected to find "patches" sources at https://sourceforge.net/projects/dsgpl/files/Synology%20NAS%20GPL%20Source/15047branch/x64-source/
But it is not the case... Probably Python is not GPL but BSD and so they do not bother to publish modified sources as license does not compell to do so.

Safihre:
Updating to Python 2.7.14 fixed it!
It also includes support for cross-compiling so we don't need those patches anymore.
Now `cross/python` compiles 💯 
On to `spk/python`!

ymartin59:
Will be superseed by #3060

m4tt075:
Thanks much, @Safihre . Closing as superseeded.

PR #2989, opened 2017-11-07
Correct openssl download location

jelmerj:
The current download location is no longer valid which breaks build.


jelmerj:
This request should replace PR#2791

PR #2998, opened 2017-11-13
Borgbackup 1.1.2

yuroller:
_Motivation:_ borgbackup v.1.1.2 package.
_Linked issues:_ https://github.com/SynoCommunity/spksrc/pull/2906

### Checklist


cross/acl/Makefile
```diff
@@ -0,0 +1,21 @@
...
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).src.$(PKG_EXT)
+PKG_DIST_SITE = http://download.savannah.gnu.org/releases/acl
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS = cross/attr
+
+HOMEPAGE = http://kernel.org/
+COMMENT  = 
+LICENSE  =
```
cytec: should be GPLv2+ as of https://savannah.nongnu.org/projects/acl
Note: might add the Comment there as well ;)

cross/borgbackup/Makefile
```diff
@@ -0,0 +1,17 @@
...
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://files.pythonhosted.org/packages/source/b/$(PKG_NAME)
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS = cross/acl cross/lz4 cross/openssl
+
+HOMEPAGE = https://borgbackup.readthedocs.io/
+COMMENT  = Deduplicating backup program with compression and authenticated encryption
+LICENSE  = Borg
```
cytec: isn't Borg BSD?

cross/lz4/Makefile
```diff
@@ -0,0 +1,22 @@
+PKG_NAME = lz4
+PKG_VERS = 1.7.5
+PKG_EXT = tar.gz
+PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
```
cytec: Please use`PKG_DIST_FILE` for a better tar.gz name. example see: https://github.com/SynoCommunity/spksrc/blob/master/cross/rtmpdump/Makefile#L5
basically something like `$(PKG_NAME )-v$(PKG_VERS).$(PKG_EXT)` should be good ;)

cross/acl/Makefile
yuroller: Ok, I will update License and add Comment for acl package

cross/borgbackup/Makefile
yuroller: I do not know: licence is this https://github.com/borgbackup/borg/blob/master/LICENSE
What shall I put in the Makefile?

cross/lz4/Makefile
yuroller: Ok, I will try to use PKG_DIST_FILE for lz4 package

cytec:
@yuroller normally i'd say: 1 PR per package. So python and borg would be one PR as both is needed for borg to work, the typo fix in an additional one. since it's only a typo that's been fixed i'd be ok with it to have it in this one for now. don't know about others \cc @SynoCommunity/developers

cross/borgbackup/Makefile
cytec: hm.. seems like on here it's mentioned to be BSD 3 Clause.
https://github.com/borgbackup/borg#main-features (Free and Open Source Software)

ymartin59:
Republishing Python 3 package for some native packages would worth for an complete update to 3.6.3 ?

cytec:
agree if we can push python to 3.6.3 that'll be a seperate PR for just python then :)

yuroller:
Please, update python3 package, then I will do a PR for borgbackup with modification needed in python3 for missing modules

Safihre:
@yuroller Are you still willing to work on this? I am trying to understand this PR, since there already is a `borgbackup` package with these cross dependencies.
Would it now only be needed to bump the version to 1.1.4 (latest)?
Seems though that even though the package is present here, it's not published on the repo.
Is this due to DSM6 compatibility @ymartin59 ?

ymartin59:
@Safihre No idea. I just found borgbackup package has been merged from #2694 but probably never published.

yuroller:
On the weekend I try to find the time to submit a new pull request to master branch (single commit if I remember well),

Safihre:
Great, looking forward. Would you also be able to convert it to Generic Service approach?

yuroller:
What is "generic service" approach?
I created synology borg package by copying other similar python/c already present in spksrc repository, so if you point me to an example I can try doing the "generic service".
BTW borg is a command line program, not a daemon/service, so I do not know if that applies.

Safihre:
In that case you can look at Midnight Commander (`mc`). That's very similar.

I looked at this package and was bit surprised that all the modules were added to the base `python3` and not as part of just this package. That should easily be possible to keep the Python packages clean.
But that's something I can do afterwards :)

yuroller:
I did a new pull request with updated borgbackup https://github.com/SynoCommunity/spksrc/pull/3196
I had to add cross/attr and cross/acl to python3 package or else borgbackup is not able to find the .so if I include them just in borgbackup package: can someone explain me why?
About "generic service": the .sh scripts I use are very similar to those in "mc" package so it should be fine.

Safihre:
Let's continue discussions in the new PR then :)
Will look into this Python thing and report back in new PR!

PR #3002, opened 2017-11-19
Update comcerto2k DSM6 toolchain digests

Safihre:
Noticed a checksum error during the TravisCI building of the `mono` package (#2808) for DSM6.1.
https://travis-ci.org/Safihre/spksrc/jobs/303992982

I also noticed (almost) all the digests for the `6.0.2` toolchains are different from the `6.1` digests even though they are the same file. Seems odd to me?


ymartin59:
Even if file names are identically, content is different... I check downloading Synology toolchains directly from sf.net

PR #3010, opened 2017-11-24
fmpeg package enbale for apollolake-6,1 platform

jvega1976:
ffmpeg compiled with the apollolake-6.1 toolchain
It is compiled with --enable-libmfx and -enable-vaapi for hardware acceleration


cytec:
please don't open DSM6 PR's against master, use the DSM6 branch for that!

PR #3050, opened 2017-12-17
Update Tvheadend to 4.2.6 (maintenance release) for DSM 5+6

m4tt075:
### Motivation
Update Tvheadend to 4.2.6 (latest maintenance release), update libhdhomerun to 20171221 and adapt Tvheadend to use generic DSM 5+6 service

### Linked issues
#2904 

### Checklist
- [x] Build rule `all-supported` completed successfully with `5.2` toolchains
- [x] Build rule `all-supported` completed successfully with `6.1` toolchains (exceptions above)
- [x] Package upgrade completed successfully (on `XPE 5.2` and `XPE 6.1`)
- [x] New installation of package completed successfully (on `XPE 5.2` and `XPE 6.1`)


m4tt075:
@ymartin59 Could you take a look at this, please? I think the package is created correctly. It installs fine on XPE5.2, but the service is not executed after the package has been installed. I'm pretty sure that something is wrong with `SERVICE_COMMAND`, which I have now included in `service-setup.sh` just like you did for `Mosquitto`. The `-f` option should execute `tvheadend` as daemon, `-c` defines the home directory for the installation, so no clue why it is looking for a home dir. Another observation: In the install log it says `USER=tvheadend`, whereas the actually installed user is `svc-tvheadend` (XPE 5.2), as it should be. Any advice?

tvheadend_install.log:
```
Sun Dec 17 21:42:22 CET 2017
Step preupgrade. USER=tvheadend GROUP= SHARE_PATH=
Sun Dec 17 21:42:22 CET 2017
Step preuninst. USER=tvheadend GROUP= SHARE_PATH=
Sun Dec 17 21:42:22 CET 2017
Step postuninst. USER=tvheadend GROUP= SHARE_PATH=
Sun Dec 17 21:42:22 CET 2017
Step preinst. USER=tvheadend GROUP= SHARE_PATH=
Sun Dec 17 21:42:22 CET 2017
Step postinst. USER=tvheadend GROUP= SHARE_PATH=
Installing service configuration /var/packages/tvheadend/conf/tvheadend.sc
Invoke service_postinst
```

tvheadend.log:
```
Sun Dec 17 21:42:27 CET 2017
Starting tvheadend command /volume1/@appstore/tvheadend/bin/tvheadend -f --http_port 9981 --htsp_port 9982 -c /volume1/@appstore/tvheadend/var -p /volume1/@appstore/tvheadend/var/tvheadend.pid
su: can't chdir to home directory '/var/services/homes/svc-tvheadend' 
```

m4tt075:
@ymartin59 I could track the problem down for now, but still some way to go. Here my observations:
1. The `chdir` error appeared because I did not have `home directories` activated on my XPE5.2 testing system. Apparently that's needed, although I wonder why.
2. The package can be executed correctly, but only if I hand over the `-f` (daemon) option and `-u svc-tvheadend` (user) option and the `-g users` (group) option to the `tvheadend` command.
As username and group vary depending on the DSM version and user/group creation method used, I will have to adapt the `tvheadend` paramters accordingly. I think you have some variable names for that somewhere in your service scripts, but I still have to find out how I can use them. Work for another day...

ymartin59:
@m4tt075 Correct, I have configured "home directories" and so have not detected this requirement. As `synouser` command does not provide a way to change user home path, I think I will call sed to replace default location to package `var`.

If as I guessed, `tvheadend` package is designed to run as root and then fork/setuid to specific user, I propose you to adapt previous `SSS_SCRIPT` to load variables from `service-setup` and then to provide your own `conf/privilege` (warning it may be replaced by Makefile ! I have to check that) to declare start/stop operations to be run as `root`. Probably you also need `EFF_USER` variable in your script, so I have to set this variable in `service-setup` too.

I know it is more complex but "generic" support cannot fit all situations. What do you think about this proposal ?

m4tt075:
@ymartin59 Cool. Thanks. Yes, that would work, but I might have found a different solution: It seems `SERVICE_COMMAND` is always execute as `su ${EFF_USER} -s /bin/sh -c "${SERVICE_COMMAND}"` in your scripts. But then I should be able to "reconstruct" `EFF_USER` and correct group name within `service-setup` via evaluating `id -un` and `id -gn` inside of the `SERVICE_COMMAND`. Not giving up yet. I will give it a try. If it doesn't work, I will try your suggestion...

m4tt075:
@ymartin59 I have just pushed some fixes, which utilize a combination of your `demoservice` concept and the `id` approach. It works on my XPE 5.2 system. :smiley:
Could you take a quick look, please, whether this approach makes sense? If so, I'd continue testing on my DSM 6 live system...

m4tt075:
@ymartin59 I have installed an XPE 6.1 testing system this morning, and the package installs and runs with correct permissions as well. The only problem I'm still facing is that for some reason the `sed` `postinstall` commands in the `service-setup` script are not executed on the 6.1 system, whereas they run fine on the 5.2 system and on 6.1 if the non-generic approach is used.

ymartin59:
@m4tt075 It looks like `tvheadend` command is designed to be started as root and fork process setuid to user pass as `-u` argument... Option `-c` may have not effect without this `-u` argument.

m4tt075:
@ymartin59 Yikes! See last commit. Took me hours to find that nasty error. It's the same in your `demoservice` script, but does not matter there. In any case, from my tests on XPE 5.2 and XPE 6.1 I think everything is working now. I will have to do some more tests (i.e. on my live system) and build `all-supported` and stuff, but that will have to wait after Christmas...

m4tt075:
@ymartin59 I have adapted the TVH pacakge using your `pre-start` method. It installs nicely on my XPE 5.2 and XPE 6.1 test systems. Log-ins work beautifully as well. 

There is one issue left: Upgrading the package with itself continues to nuke the old configuration. I have found the `to_save_dir` variable in your generic installer script, which should actually allow to keep the configuration folder during upgrades. The old configuration should be saved prior to the upgrade and restored after the upgrade. I simply define the variable in the `Makefile` now, but apparently it is not picked up. I could not find any hints on this in your Wiki documentation either. Can you let me know how to use it correctly and add it in the documentation, please?

Otherwise, the migration should be done. Your review will be appreciated.

m4tt075:
@ymartin59 Contingent to merging of @Safihre 's fixes in PR #3064 , this package now works correctly and upgrades without nuking previous configurations. Without PR #3064 it stays dangerous and should not be merged.

I have cleaned up some additional minors and squashed all changes into a single commit. With this, for your review, please.

m4tt075:
@ymartin59 Minor cleanup in line with your review comments to the analogous Jackett PR #3070 ; squashed and force-pushed. `[WIP]` label removed from the PR title. From my perspective, ready to go. For final review please. Feel free to merge as you deem appropriate.

spk/tvheadend/Makefile
```diff
@@ -9,29 +9,38 @@ DEPENDS = cross/busybox cross/$(SPK_NAME)
...
 
+.PHONY: tvheadend_install
+# Replace standard copy/install targets
+tvheadend_install:
+	rm -fr $(STAGING_DIR)
+	mkdir $(STAGING_DIR)
+	mkdir --parents $(STAGING_INSTALL_PREFIX)
+
 .PHONY: tvheadend_extra_install
 tvheadend_extra_install:
```
ymartin59: I really wonder if this target is really usefull too... With `DSM_UI_DIR`, folder is copied without such install command.

spk/tvheadend/src/tvheadend.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+HTTPP=$1
+HTSPP=$2
+USRN=`id -un`
+USRG=`id -gn`
+HOME_DIR="${SYNOPKG_PKGDEST}/var"
+LOG_FILE="${HOME_DIR}/tvheadend.log"
+PID_FILE="${HOME_DIR}/tvheadend.pid"
+${SYNOPKG_PKGDEST}/bin/tvheadend -f -u ${USRN} -g ${USRG} --http_port ${HTTPP} --htsp_port ${HTSPP} -c ${HOME_DIR} -l ${LOG_FILE} -p ${PID_FILE}
```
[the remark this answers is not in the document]
m4tt075: I have just tried to use `${GROUP}` instead of `${USRG}` but it is an empty variable when used from within `service-setup.sh`. I will retain `${USRG}` for the time being.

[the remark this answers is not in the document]
Safihre: You have to hardcode it as one of first lines in `service-setup.sh` like @ymartin59 says a bit up here in the other comments and like I did for SABnzbd: https://github.com/SynoCommunity/spksrc/commit/2ac3e53e73577c7055f3eac688169b43d1cc55fd#diff-5d6bd92e6f38c326c2344d53c1dfb20aR10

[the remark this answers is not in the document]
m4tt075: Ahh, now I see. Misunderstood it. Thanks for clarifying! Will put it in `sc-media` then...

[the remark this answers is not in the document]
m4tt075: Just tried this, but it screws up all the user permissions again. I will revert to the old `tvheadend` goup and force push to get a cleaner version. I also think you have been reviewing outdated commits, as the `tvheadend.sh` file has already been removed some days ago.

ymartin59:
In fact, DSM 6 framework creates service account with a default group with same name - you have used as process argument.
But on DSM 5, no group will be created if `GROUP` is unset (either from service-setup or from wizard)

m4tt075:
@ymartin59 Thanks. Mostly implemented. See `Fixup` commit. I'm hesitant w.r.t. the following though:
- Specifying a download location via the wizard: TVH does not do downloads. There is something like a recording directory. This needs to be specified in the TVH Web-UI. However, I do not see a viable way to "synchronize" the two. It is way easier and more natural for a user installing TVH to select the recording directory via the TVH interface and assign the user permissions afterwards as needed.
- Enforcing the group-name `tvheadend` within `service-setup.sh` by hardcoding: This would match what we have on 6.1 systems. On 5.2 systems the group name should be `users` though, shouldn't it?

ymartin59:
OK sorry for my mistake. As DrBean has set "GROUP=sc-media" in installer script, I thought there was a download directory. So forgot about this point about share create and group name. Your shell requesting current user group required as service parameter is perfect so.

ymartin59:
As a consequence, please clean wizard steps about permission management, as this is not relevant either.

ymartin59:
... except if `tvheadend` has to be granted access to other existing share... In that case, GROUP=sc-media will have interesting following effect: it will register `EFF_USER` as member of that group. So that TVH users can grant service access to existing shares.

m4tt075:
@ymartin59 You are right. I'm hardcoding `${GROUP}` to `sc-media` now, leaving the group permissions for starting `tvheadend` as daemon untouched. This does allows usage of `sc-media` to grant permissions for recording directories later on. I have successfully tested this on `XPE 6.1` as well as `XPE 5.2`. So far so good...

However, this leaves me in similar situation to what @Safihre encountered for `SABnzbd`. During an upgrade process, we have to make sure that existing recording dirs receive read/write permissions for the `sc-media` group as well. I have been trying, but so far failing, to achieve that. Following @Safihre 's example I try to use the `set_syno_permissions` function to do it, but it seems to fail on `DSM 5.2`. Is it supposed to work on `6.1` only? I will push my latest changes, but we are back to WIP, I'm afraid...

Safihre:
@m4tt075 I didn't test anything on DSM5.. Do you have an indication what is going wrong in DSM5?
By checking the permissions using `synoacltool -get`, are any permissions applied?

m4tt075:
@Safihre Nope, I don't. I tried recording but did no have permissions. I have checked the group permission settings in DSM and could verify that they were not set correctly. I checked the `install.log` for output (and potential errors) from the synoacltool commands and found nothing. I'm running out of ideas...

ymartin59:
@Safihre we are testing DSM 5 with XPEnology virtual machine... but it is quite complex to set up, and there too proper instructions to succeed at first trial are lacking in wiki.
I will add a "write" test case in `demoservice` to confirm service process gets permission to write on share.

spk/tvheadend/src/service-setup.sh
```diff
@@ -1,15 +1,25 @@
...
 
+# Group configuration to manage permissions of recording folders
+GROUP=sc-media
+
+# Service configuration. Change http and htsp ports here and in conf/tvheadend.sc for non-standard ports
+HTTPP=9981
+HTSPP=9982
+
+# Need to enforce correct permissions for recording directories on upgrades
+UPGRADE_CFG_DIR="${TMP_DIR}/var/dvr/config"
```
ymartin59: Probably it is a better place in `service_preupgrade` ?

m4tt075:
@ymartin59 Yep. I have put it in `service_postupgrade` now, where it makes even more sense. I'm pretty sure the code works as intended now. Git pushed the fixup.

Unfortunately, with evey problem I solve, it seems I'm discovering new ones: While testing I ran into several situations where the upgrade got stuck in the `postupgrade` routines. The logs continued to fill up with `(synoacltool.c, 140)It's Linux mode` errors. I have seen this on `XPE 5.2` as well as `XPE 6.1` systems and could reproduce it by creating a shared folder via the DSM web-UI first and assigning some random read or write permissions. If you then apply the `set_syno_permission` routine on that shared folder, the errors appear. Apparently this is a known bug within the `synoacltool` command, which others have encountered before, see e.g. [here](https://forum.synology.com/enu/viewtopic.php?t=87000). Can you confirm that the issue appears on your systems, too?
@Safihre : FYI, as `SABnzbd` might very well be affected too...

m4tt075:
@ymartin59 @Safihre My apologies, the infinite loop I went into was due to a missing apostrophe in my script. I have just pushed the fix. The `It's linux mode` errors do still occur, but because the default recording directory TVH choses, is set to `var/packages/tvheadend/target`. The `set_syno_permissions` routine in the `spksrc.service.installer` script always expects a path to start with `volumeX` and does not resolve links via `realpath`.

In any case, you will likely not be able to reproduce this. My initial analysis was wrong. Sorry for the hassle...

Safihre:
You got me worried 🤗 I also saw this once, also when I made a typo 😄
Maybe you can analyze the path and then check if it needs a prefix before calling the permission fix?

m4tt075:
@Safihre Yeah. Thanks for your empathy!  😏 
I have just proposed PR #3084 as possible fix.

ymartin59:
`set_syno_permissions` is designed to grant service write permissions to volume shares accessible by users.
If you want users to be able to browse recorded files, then configure application to volume shares.
If not, these temporary location can live in package `target` but users probably do not expect them to consume GiBs.

m4tt075:
@ymartin59 I agree with you, especially as  package `target` is just a symbolic link to a directory residing in `/volumeX` anyways. I just cannot pre-configure TVH to use some `/volumeX/YouNameIt` directory from the start, because of how TVH works:
- The first time you run TVH, an installation wizard is started within TVH. This is when most of the configuration files are created, including the one containing the recording directory.
- If you do not specify a recording directory during the first installation process, or you accept defaults or you choose not to use the wizard, the recording directory will be set to the home directory of the installation, which in our case is `/var/packages/tvheadend/target/var` (including `target` as symbolic link)

There is some likelihood that a substantial number of TVH installations will possess this recording directory configuration. The symbolic link issue plus the potential risks with users who specify system directories as "share" folders, led me to propose #3084 .

If you do not want to amend the `set_syno_permissions` function, I can also do the necessary checks and `realpath` conversion within the TVH package upgrading procedure, but then we would have to make sure to do this in every package using that function. Please just let me know what you prefer: #3084 (or something like this) or a package-specific solution?

ymartin59:
I do not object your proposal... but I really want to have a relevant test case in `spk/demoservice` to confirm `set_syno_permissions` operates properly (mainly because I failed to test it on my own).

m4tt075:
Just pushed some further improvements:
- Wizard files upgraded, analogous to what @Safihre did. I just cut down the installation text a bit, because otherwise you have to scroll down to read the (more important) warning on permission changes...
- Thanks to the improved logging functionality, I have uncovered a bug related to fontconfig, which must have been around pretty much forever. Not a big thing, but fixed now as well.

m4tt075:
@ymartin59 As to your `demoservice` request, I really don't know how to do that. `demoservice` is "safe", because you ensure that the download dir is in `/volume` via the wizard implementation. No other dirs can "sneak in". Let me try to explain again, what the problem is that I see:

In TVH the recording directories are not defined at the wizard level, but via the TVH web interface and I cannot run any checks on the inputs there. Those directories are stored in simply configuration files, which my upgrade script is parsing to correct the permission settings. But that means, that a user with access to the TVH web-interface (i.e. any guy watching television) can put any kind of directory in there. Let's assume you chose `/bin`. In that case, the next upgrade process (which runs as `root`) could give the "sc-media" group rwx permissions on the `bin` folder. So, it is not that `set_syno_permissions` does not work, it might work "too well"...

Therefore I'm still struggling to understand what "test-case" in `demoservice` you are looking for. It is just about where and how you enter or obtain the folder name, which is handed over to `set_syno_permissions`. And that's the issue here. Again, I could block it at the parser level, but I feel strongly we should block these options somewhere. Please advise...

Safihre:
@m4tt075 I understand your concern. But now that you included `realpath`, doesn't `/var/packages/tvheadend/target/var` resolve to its actual path in `/volumeXX/@appstore/tvheaded/var/`? 
And thus we can still do a check that the actual path should start with that, in order to potentially prevent 'dangerous' paths?

m4tt075:
@Safihre Yes and yes. From my perspective the implementation in this PR here together with the current implementation of #3084 does exactly what we need and works correctly. I'm just hesitant to remove the volume check from #3084 and I don't know what I'm supposed to do with `demoservice`... 
I would create some test builds with both, so that we get some insights from beta testers moving forward, but I  want to be on the same page with @ymartin59 and you first.

Safihre:
Aah good, I see you left the `/volumeXX` check in #3084 👍  
That PR looks good to me now and will test it in a bit for SABnzbd package :)

m4tt075:
Thanks @Safihre . Much appreciated! As @ymartin59 has just merged #3084 I feel safe enough to start building and distributing some test packages for TVH as well.

ymartin59:
I consider we should deliver packages for `dockerx64` and `rtd1296`... So I will investigate how to built `ffmpeg` there

m4tt075:
This is great! `dockerx64` builds. I have experimented with `rtd1296` recently. Got `ffmpeg` working except of the `x264` cross dependency. You can pull [this branch](https://github.com/m4tt075/spksrc/tree/rtd-test) if you like to...

m4tt075:
@ymartin59 Cleaned up the wizard files a bit more and removed the _enu versions. If you want French translations here, too, please just let me know.

Test builds are available [here](https://github.com/m4tt075/spksrc/releases/tag/Tvheadend-4.2.5).

m4tt075:
PR comment updated.

@ymartin59 I have spent some more hours trying to fix the `rtd1296` compilation within the `ffmpeg` cross-package. Unsuccessfully. I have also tried to remove the `ffmpeg` dependency by enabling the statically linked version, but this just creates different problems. As this PR is not breaking anything, but just improving the situation compared to what we currently have, I propose to merge it.

m4tt075:
@ymartin59 `rtd1296` builds as well now with PR #3119 .

m4tt075:
@Safihre @ymartin59 As I have lost track of some of your recent changes. Could you please let me know what still needs to be done here? I'm aware of three "dependencies" that were still under consideration:

1. Fixing syno permission setting. I understand this is fixed now, but does it have implications for the code in this PR?
2. That TVH is started as a daemon, which prevented me from just using "SERVICE_COMMAND". Is there a better way to do this now and if yes, do you have a reference implementation for this I can look at?
3. That `ffmpeg` did not compile for `rtd1296`, which I fixed and @ymartin59 "broke" again. 😇 How do we proceed in that regard?

For reference, the test package corresponding to this PR in its current state has had 550 downloads in the meantime and I'm not aware of issues.

ymartin59:
@m4tt075 Hello back.
- `ffmpeg` is ready for TVH (all arch OK and patches included) #3136
- @Safihre have worked on syno permission and fixes have been merged on `master` (included in #3136 branch as rebased)
- for service command started as background, my pending PR #3123 still requires some testing... you may give it a try with TVH if you want

m4tt075:
@ymartin59 Thank You, Sir! You guys rock! Seems you are way more productive when I'm gone, though...  :smile:  
Will test over the weekend.

m4tt075:
@ymartin59 @Safihre I ran some tests and have some issues with user permissions now, which might relate to either #3123 or #3161 / #3165. Here is what happens:
1) Clean install works fine. Package runs. I can configure it, record programs, etc. All fine.
2) Then I upgraded the package by itself
3) The upgrade works well, but the package does not start anymore. 
4) I figured out that it can no longer write into the `.../tvheadend/var` folder.

Here the unix permissions:
```
root@myxpe61:/usr/local/tvheadend/var# ls -la
total 25760
d--------- 1 sc-tvheadend tvheadend      244 Feb 24 19:38 .
drwxr-xr-x 1 root         root            70 Feb 24 19:21 ..
d--------- 1 sc-tvheadend tvheadend       64 Jan 16 21:15 accesscontrol
d--------- 1 sc-tvheadend tvheadend    25600 Feb 24 19:17 bouquet
drwx------ 1 sc-tvheadend tvheadend       18 Feb 24 19:13 channel
---------- 1 sc-tvheadend tvheadend     1136 Feb 24 18:55 config
d--------- 1 sc-tvheadend tvheadend       18 Feb 24 19:14 dvr
---------- 1 sc-tvheadend tvheadend   222492 Feb 24 19:17 epgdb.v2
drwx------ 1 sc-tvheadend tvheadend       12 Feb 24 19:06 epggrab
d--------- 1 sc-tvheadend tvheadend       16 Feb 24 19:04 input
-rw-r--r-- 1 sc-tvheadend tvheadend       22 Feb 24 19:39 .lock
d--------- 1 sc-tvheadend tvheadend       64 Jan 16 21:15 passwd
d--------- 1 sc-tvheadend tvheadend      448 Jan 16 21:15 profile
drwx------ 1 sc-tvheadend tvheadend       12 Feb 24 19:13 service_mapper
-rw-r--r-- 1 root         root          3754 Feb 24 19:21 tvheadend_install.log
-rw-r--r-- 1 sc-tvheadend tvheadend 26138908 Feb 24 19:17 tvheadend.log
```

No ACL permissions are set:
```
root@myxpe61:/usr/local/tvheadend/var# synoacltool -get "/usr/local/tvheadend/var"
(synoacltool.c, 359)It's Linux mode
```

The `tvheadend_install.log` can be found [here](http://paste.ubuntu.com/p/kHJj7PCc4Z/).

Any idea why this happens?

Safihre:
I think I have an idea:
TvHeaded saves its data in it's own package folder, right?
So what happens is that the `set_syno_permissions` forces the `000` permissions by setting ACL.
But then afterwards we set Linux persmissions on the `var` folder.
The main problem of course being: why is the data folder inside the package folder..

m4tt075:
@Safihre Thanks for looking into this. I think you may well be right. TVH indeed sets it's "home" directory as default recording folder on initial installation. As long as you do not replace it explicitly by something else from within the TVH web-interface, it will be there. Hmh, do you have an idea how to solve that? I cannot prevent TVH from setting that folder. I could of course skip the home folder from being parsed as a recording folder, which would avoid that the unix permissions are nuked, but also that the ACL permissions are set. In that case I'm not sure I could still record into the home folder. Any advice?

Safihre:
I don't know 😕  It's a choice between the ACL permissions or the Linux permissions...
But how to users access their recordings? Only through the app? 
Is there no way to supply a default config-file to change this even from first start? Like we do for the other packages like SAB/NZBGet/etc.

m4tt075:
Yes. This package is the backend part of a TV streaming service. You schedule recodings via a frontend which can be on a HTPC, some iPad or Android phone, you name it. The backend records into the specified folder and also plays back recordings from there, streaming to the frontend(s). The frontend(s) have no idea about what recording folder has been selected within the backend.

I'm not at all keen on supplying a default config file because the format and config parameters changes from time to time when the software itself is upgraded. I also cannot just replace the folders using `sed`  as the config file is only generated dynamically by TVH once it has been launched for the first time.

However, I just realized that the version I installed cleanly worked. I could record, even without ACL permissions. If that holds true, just excluding the home folder from ACL permission setting is enough. I think I will give it a try.

But please let me know if you have a better / alternative idea. I very much appreciate your thoughts!

Safihre:
@m4tt075 Aaah like that it works 👍 
Well in that case it doesn't need ACL, only Linux permissions! If only the app itself needs the access.

With regards to the saving directory I would be worried how long an upgrade takes when users have a number of recordings, since it copies `var` -> `tmp` -> `var`. But I guess it seems there's no changing this behavior.

m4tt075:
Well, this is just a hypothesis for now. I still have to verify it. We will see, but I will report back...  :smirk:

You are right, the upgrades can take some time for this package because of the copying involved. I don't see a way to avoid this the way the software works now. But maybe I can ask the developer to include an option to specify the default recording folder, ideally as configuration parameter. Then we could simply solve this via the wizard. Would probably be the best option.

m4tt075:
@Safihre Unfortunately, it does NOT work. I went through everything step by step again. I could not solve it, but at least I found the reason, and hope you can take it from there.

1. Clean install, I don't change the recording folder. I configure some tuners and am able to record.

Unix permissions:
```
root@myxpe61:/var/packages/tvheadend/target# ls -la
total 13592
drwxr-xr-x 1 sc-tvheadend tvheadend       92 Feb 24 23:17 .
drwxr-xr-x 1 root         root           166 Feb 24 23:03 ..
drwxrwxr-x 1 sc-tvheadend tvheadend       24 Feb 24 18:27 app
-rw-rw-r-- 1 sc-tvheadend tvheadend 13903164 Feb 24 23:17 BBC News.ts
drwxrwxr-x 1 sc-tvheadend tvheadend      644 Feb 24 23:03 bin
drwxrwxr-x 1 sc-tvheadend tvheadend       10 Feb 24 18:27 etc
drwxrwxr-x 1 sc-tvheadend tvheadend     4574 Feb 24 18:27 lib
-rw-r--r-- 1 sc-tvheadend tvheadend    10835 Feb 24 17:49 openssl.cnf
drwxrwxr-x 1 sc-tvheadend tvheadend       42 Feb 24 18:27 sbin
drwxrwxr-x 1 sc-tvheadend tvheadend       56 Feb 24 18:27 share
drwxr-xr-x 1 sc-tvheadend tvheadend      254 Feb 24 23:17 var
```
ACL permissions = none:
```
root@myxpe61:/var/packages/tvheadend/target# synoacltool -get "/var/packages/tvheadend/target"
(synoacltool.c, 359)It's Linux mode
```
BBC News.ts is the acutal recording. I have double-checked. This works, and it does NOT require ACL permissions after the clean install.

2. I upgrade the package with itself and the recordings do NOT work anymore.

Unix permissions:
```
root@myxpe61:/var/packages/tvheadend/target# ls -la
total 12
drwxr-xr-x 1 root         root         70 Feb 24 23:25 .
drwxr-xr-x 1 root         root        166 Feb 24 23:25 ..
drwxrwxr-x 1 sc-tvheadend tvheadend    24 Feb 24 18:27 app
drwxrwxr-x 1 sc-tvheadend tvheadend   644 Feb 24 23:25 bin
drwxrwxr-x 1 sc-tvheadend tvheadend    10 Feb 24 18:27 etc
drwxrwxr-x 1 sc-tvheadend tvheadend  4574 Feb 24 18:27 lib
-rw-r--r-- 1 sc-tvheadend tvheadend 10835 Feb 24 17:49 openssl.cnf
drwxrwxr-x 1 sc-tvheadend tvheadend    42 Feb 24 18:27 sbin
drwxrwxr-x 1 sc-tvheadend tvheadend    56 Feb 24 18:27 share
drwxr-xr-x 1 sc-tvheadend tvheadend   270 Feb 24 23:27 var
```
ACL permissions = none:
```
root@myxpe61:/var/packages/tvheadend/target# synoacltool -get "/var/packages/tvheadend/target"
(synoacltool.c, 359)It's Linux mode
```

The ACL permissions are not involved, so the unix permissions must be the reason. There is indeed a subtle difference in the unix permissions. After the clean install the "." directory is owned by the package owner, after the upgrade it is owned by root. Manually `chown`ing the `target` directory back to  the package owner, fixes the problem and I can record again. I don't know whether this is because `target` is a link, but maybe that gives you some hints to inquire further. Any advice appreciated...

Safihre:
@m4tt075 Ah, yes that's relativily easily to explain.
Old packages do a `chown -R ${USER}:root ${SYNOPKG_PKGDEST}` during `postinst`, like [here](https://github.com/SynoCommunity/spksrc/pull/3050/files#diff-6df9ec497736dade7163862893595f92L74). 
But in Generic Service this is (on DSM6) [only performed on the `var` folder](https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.installer#L295-L300). This was specifically changed by @ymartin59 in [this commit](https://github.com/SynoCommunity/spksrc/commit/dd1e07073740c51b3466f0acdc6694a56cb4d67c).

I can imagine the assumption that packages shouldn't need to change anything else then their `var` directory, but why the DSM 5 vs 6 difference?

Safihre:
@m4tt075 In that example it seems it's even recording in the main directory? Not even `var`? Is that where those recording stay really? Or afterwards they go into `var`?

m4tt075:
@Safihre I see the DSM5 and 6 difference, still don't understand. I have not yet even tested DSM5 and I have not installed old packages. In a clean install of a new DSM6+ package the `${SYNOPKG_PKGDEST}` folder is apparently owned by `${EFF_USER}`. Upgrading that package by itself changes the ownership of that folder to `root`. This should not be related to differences btw. DSM5 and 6.
As to your question on the default recording directory: Yes, it is `/var/packages/tvheadend/target`. But I don't apply any unix or ACL permissions to that folder in my package-specific scripts, so the unintended ownership change should be introduced in the generic scripts somewhere.

Safihre:
> Upgrading that package by itself changes the ownership of that folder to root. This should not be related to differences btw. DSM5 and 6.

Hmm, that's strange. I don't see any place where the fresh and upgrade paths differ in permissions of the main folder. @ymartin59, any clue?

@m4tt075 what if you remove the DSM5 switch and just apply `set_unix_permissions "${SYNOPKG_PKGDEST}"` instead of just `"${INST_VAR}"` in `spksrc.service.installer`?

m4tt075:
Exactly. I don't see any place either. The only other idea I have is, that this is DSM-specific. DSM changes file ownerships and permissions on installations using the `privilege` file. Maybe, DSM does not do that on upgrades. If so, this might actually be a DSM problem.

Safihre:
@m4tt075 So I tried to reproduce and indeed I see that the symlink `/var/packages/lazylibrarian/target` is owned by `root:root`. But the actual package folder `/volume1/@appstire/lazylibrarian` is owned before and after upgrade by `sc-lazylibrarian:lazylibrarian`.
```
sabtest@XPTest:/$ ls -al /var/packages/lazylibrarian/target
lrwxrwxrwx 1 root root 32 Feb 25 11:11 /var/packages/lazylibrarian/target -> /volume1/@appstore/lazylibrarian

sabtest@XPTest:/$ ls -al /volume1/@appstore/
total 0
drwxr-xr-x 1 root                 root              160 Feb 25 11:11 .
drwxr-xr-x 1 root                 root              138 Feb 23 16:50 ..
drwxr-xr-x 1 sc-lazylibrarian     lazylibrarian      18 Feb 25 11:11 lazylibrarian
```

ymartin59:
@Safihre @m4tt075 I have decided to only grant write permissions to `var` "by defaults" at least for service log files... knowing that `var` folder is the only one handled by `upgrade` process.
As far as I know, a package upgrade consists "almost" in a full uninstall / install process, all files are removed and new package content is extracted again.
I agree I need to document more, but here is the idea: if a package "really" (and I doubt it is a generic requirement) has to write files in its own package content (which is bad, as it is discarded at upgrade) elsewhere then in `var`, then `service_postxxx` functions has to set proper permission and backup/restore process for modified files (that are not in `var` so).
I hope this get things clearer... I am working on documentation but there is so much to do...

m4tt075:
@ymartin59 Seems it is a feature rather than a bug then. In any case, it does make sense. I will try to add `set_unix_permissions` to the `service_postupgrade` function and see whether it works.
On another note, I fully agree that having the `target` folder as default recording folder is far less than optimal. I will ask for a configuration option upstream.

Safihre:
Maybe you can do some testing and see in which step this permissions are set to `root`. By just removing all commands from the different steps and adding them back 1 by 1 (or in groups).

m4tt075:
@Safihre @ymartin59 OK, with setting the link permission in `service_postupgrade` manually it works again. I do not use the `set_unix_permissions` function as I don't want the `chown` command to recurse. Thanks for your help with this.
@ymartin59 I have cherry-picked the background service commit to test this approach. I will drop it again, once your PR is merged. I have also looked into the option to reuse compiled `ffmpeg` libraries. However, unlike `chromaprint` `Tvheadend` does not have a configure option to point at such libraries.

I have tested on XPE 6.1 (clean install and upgrade, including setting up some tuners, channels and testing some recordings) and XPE 5.2 (upgrade only, but same tests as before). Please take another good look at this package and let me know whether you are fine with it. Once you are, I'd trigger another round of test builds that we could offer via @Safihre 's beta list.

ymartin59:
@m4tt075 About default recording location, there is chance TVH uses either user `HOME` directory or current working directory `CWD` you may change before starting process. Changing it to `var` is probably the best option

m4tt075:
@ymartin59 It defaults to PREFIX. I have an inquiry running upstream. Hope to get some feedback tomorrow.

m4tt075:
@ymartin59 Heck, you were right. It defaults to `HOME` indeed. The problem is "just" that on DSM6.1 the `HOME` directory now refers to `${SYNOPKG_PKGDEST}`, which corresponds to the `target` link. This happens for all packages. I didn't expect that. Examples from DSM6.1 `/etc/passwd`:
```
sc-demoservice:x:210722:210722::/var/packages/demoservice/target:/sbin/nologin
sc-jackett:x:276806:276806::/var/packages/jackett/target:/sbin/nologin
sc-syncthing:x:104555:104555::/var/packages/syncthing/target:/sbin/nologin
sc-tvheadend:x:150309:150309::/var/packages/tvheadend/target:/sbin/nologin
sc-umurmur:x:297653:297653::/var/packages/umurmur/target:/sbin/nologin
```
On DSM5.2 systems, however, the home directories correspond to the "expected" `var` folders. Example:
```
svc-jackett:x:1044:100:User running jackett:/var/packages/jackett/target/var:/sbin/nologin
svc-tvheadend:x:1047:100:User running tvheadend:/var/packages/tvheadend/target/var:/sbin/nologin
svc-umurmur:x:1045:100:User running umurmur:/var/packages/umurmur/target/var:/sbin/nologin
```
Is the DSM6.1 behavior intentional?

Safihre:
@m4tt075 I had the same moment of surprise while working on Sonarr/Radarr 😀
Good thing @ymartin59 pointed that out. 
You can just force `$HOME` before calling the command (see Sonarr/Radarr) and then it's the same always.

m4tt075:
@Safihre Crikey! Thanks. Yes, setting `env $HOME_DIR` before calling the command works. But why is that even necessary? I understand we want / have to handle permissions differently, but I don't understand why we would want `target` folders as `HOME` on DSM6.1. Does that make sense to you?

m4tt075:
Now the permissions are srewed up again. This drives me crazy. Enough for the day...

m4tt075:
@Safihre @ymartin59 I have continued debugging and pushed fixes, which partially are package specific, but partially may require a broader implementation. One of the fixes was already discussed above, but I will summarise to give the full picture. Based on the recent framework changes, I ended up with three problems:

1. The change of the `HOME` directory for DSM6.1+ systems from `target/var` to `target` has lead default configurations of TVH to record their programs into the package system folder (`target`). This problem also affects the currently (re-?)published TVH version, so a solution was needed. I tackled the problem in two ways: Firstly, I introduce a patch to hardcode the default recording directory to `var`, which helps with clean installs (`env HOME` eventually did not work). Secondly, for upgrades I backup potential recordings from the `target` directory and restore them to the `target/var` folder, so that they are not deleted in the upgrade process.

2. Permissions on DSM6.1 (might be relevant beyond this package): The package would not start again after an upgrade because of permission issues. This was eventually fixed by setting `${EFF_USER}:${USER}` ownerships on `/var/packages/tvheadend/target`, i.e. the link itself! I still don't know why this is necessary, but might affect other packages too.

3. Permissions on DSM5.2 (might be relevant beyond this package): Again, the package would not start again after upgrades. I ended up plastering the `installer` with `ls -la` and `synoacltool -get` messages to figure out what happened and figured out that DSM (!) set ACL permissions on the `/volume1/@appstore/tvheadend` folder (but ONLY on upgrades, not for clean installs). During `pre_inst` the folder obviously is not yet present, but right at the start of `post_inst` the ACL permissions are already there. This prevents the unix permissions from taking effect. You can look at `service_post_upgrade` to see how I handled it, but it might make sense to include `synoacltool -del` commands into the `set_unix_permissions` function to ensure that ACL permissions are always removed before unix permissions are set. Just like we remove all ACL permissions before we set unix permissions now. 

I have done A LOT of testing on DSM6.1 and DSM5.2 (Xpenologies), but still need more. However, I don't want to run in the wrong direction. Please review these changes and let me know what you think. Thanks.

cross/tvheadend/patches/002-fix-recording-dir.patch
```diff
@@ -0,0 +1,12 @@
+# This ensures that the default recording directory is not placed in package system directories, independent of the DSM version
+--- src/dvr/dvr_config.c.orig	2018-02-27 18:58:45.957196424 +0100
++++ src/dvr/dvr_config.c	2018-02-27 19:00:17.210138337 +0100
+@@ -282,7 +282,7 @@
+ 
+   /* Try to figure out a good place to put them videos */
+ 
+-  homedir = getenv("HOME");
```
Safihre: So setting `$HOME` didn't work? That seems strange based on this code.

spk/tvheadend/src/service-setup.sh
```diff
@@ -0,0 +1,87 @@
...
+    BIN=${SYNOPKG_PKGDEST}/bin
+    $BIN/busybox --install $BIN
+    $BIN/delgroup "${USER}" "users" >> ${INST_LOG}
+    $BIN/deluser "${USER}" >> ${INST_LOG}
+}
+
+service_preupgrade ()
+{
+    # For backwards compatibility on DSM6 systems, backup potential recordings from package root directory
+    if [ $SYNOPKG_DSM_VERSION_MAJOR -gt 5 ]; then
```
Safihre: So this applies to the already published package, or only the testing package?

cross/tvheadend/patches/002-fix-recording-dir.patch
m4tt075: I know. But it didn't. I have tried...

spk/tvheadend/src/service-setup.sh
m4tt075: Both, the already published version and the testing package.

ymartin59:
Just read quickly... I wonder if `/var/packages/tvheadend/target` permission may be broken because you invoke `set_syno_permissions` on `/volume1/@appstore/tvheadend` which is not a relevant "share" like `/volume1/downloads`

m4tt075:
Yes, correct. We just have to fix it on upgrades (as proposed here) since it can occur with the currently published package as well. In this new version `set_syno_permissions` is not executed on any paths including `/@appstore` anymore. So in the future we might be able to remove the fix again.

m4tt075:
@ymartin59 I just pushed improvements to the recording recovery process, which parses recordings from the relevant log files and replaces links so that the upgrade should be seamless for users. Could you please review this full PR again and let me know what you think? If you are fine with this, I'd like to build some new test packages, collect user experience and move this forward...

noncon66:
Is there a chance to get a rtd1296 version of the TVheadend package (@Safihre closed #3083)?

cross/tvheadend/patches/002-fix-recording-dir.patch
```diff
@@ -0,0 +1,12 @@
...
++  homedir = "/usr/local/tvheadend/var";
```
ymartin59: I have a preference for `/var/packages/tvheadend/target/var`

spk/tvheadend/src/service-setup.sh
```diff
@@ -0,0 +1,128 @@
+# Package specific behaviors
+# Sourced script by generic installer and start-stop-status scripts
+
+# Service configuration. Change http and htsp ports here and in conf/tvheadend.sc for non-standard ports
+HTTPP=9981
+HTSPP=9982
+
+# Replace generic service startup, run service in background
+GRPN=`id -gn ${EFF_USER}`
+HOME_DIR="${SYNOPKG_PKGDEST}/var"
```
ymartin59: What is that variable used for? By the way, it should be `HOME`, not `HOME_DIR`... probably the reason why you enforced location with patch

m4tt075: @ymartin59 It's used in the `SERVICE_COMMAND`. The `-c` option sets the configuration directory. This is independent of the default recording directory, just coincides. I haven't been able to get it to work without the patch, despite defining `HOME` and using `ENV` in the `SERVICE_COMMAND`.

cross/tvheadend/patches/002-fix-recording-dir.patch
m4tt075: @ymartin59 `/usr/local/tvheadend` is used consistently for recording directories, not only here but also during the upgrade procedures when moving files from package root into `/usr/local/tvheadend/var`. If we change it here, we should change the upgrade scripts too and that would require a whole new round of testing of the upgrade scripts, which I would very much like to defer (I'm getting tired). I can look at it again for the next update, but if it is just about preference, I'd very much like to leave it as is for now.

m4tt075:
Thanks, @ymartin59 ! Will update the link in @Safihre 's `beta testing` issue and expose it to some bleeding edge users for some more real-world testing.

Safihre:
It was downloaded 367 times, so that sounds like a proper testing!

m4tt075:
@ymartin59 @Safihre Thanks, guys, but in the meantime 4.2.6 has been published. In addition the upgrade procedures are dependent on the actual volume number, which can be avoided. I have got the upgrade and the fixes, but still need to compile and push them. Have been drowning recently. Will try to do it over the weekend...

ymartin59:
No hurry... was just a reminder... backlog is huge !

Safihre:
Unless there's spk specific comments, let's merge and publish by mid next week (18 April). The PR has been open for 4 months.. 
:)

Safihre:
No comments received, so time to get this out! Do you have time to build and publish @ymartin59 ? :)

noncon66:
Hello, 
when I try to install the latest TVheadend (tvheadend_rtd1296-6.1_4.2.6-13.spk) on my DS218play i get this message: "Port undefined configured for this package is either used by another service or reserved for system use. Please disable or modify the conflicting service or contact the developer to modify the package config."

Thanks, noncon66

m4tt075:
Tvheadend should be running on ports 9981 and 9982. Not sure what is blocking it from installing. Any other package that would come to your mind and which you can remove? 
If you don't know, you can `ssh` into your NAS and execute the following commands:
`cd /usr/local/etc/services.d`
`grep -e "ports" *`
Look for the package that is using ports 9981 and/or 9982. That's the one that is clashing.

Safihre:
@ymartin59 No time to build?

ymartin59:
@m4tt075 It is possible that previous package revision does not have properly freed port when uninstalling... it happened for me often when testing "broken installer scripts".
@Safihre Sure I will publish it

m4tt075:
@ymartin59 Possible, as there have been several packages from different sources flying around, but I don't believe in some systematic problem. I have had this as well with another package some time ago, but could not reproduce it. The 4.2.4 version and the 4.2.5 and 4.2.6 test builds together must have had thousands of downloads in the meantime. I'm just aware of this single report and the reporter did not come back either.

noncon66:
@m4tt075 Sorry for not reporting back. I could not find any other package blocking the port. So I tried a restart. After that the package installed without any problem.

ymartin59:
The code of `spkrepo` really suffers of troubles/limitations/bugs... that new version was not expected to be visible as it was not marked as "active"... I just removed it.
@Safihre @m4tt075 When working on translations, I find out that package relies on `sc-media` group instead of `sc-download`. Do you agree to change it so that it matches our "one-group" permissions policy?

Safihre:
This is really a media-package,s o the `sc-media` makes sense?

m4tt075:
Yes, that was my understanding as well. It should not interfere with the `sc-download` packages, which share folders and download locations. Would keep it separate to allow for different permission settings.

PR #3060, opened 2017-12-27
[WIP] Python 2.7.14 - Fix for DSM6 and Generic Service Support

Safihre:
Fixed Python, it took a whole lot of troubleshooting.. Luckily Python 2.7.14 supports cross-compile natively now, removing some of the trouble.

I did decide to remove the `cross/sphinx` dependency since it adds 18MB to the already large Python file. It shouldn't be included in the main Python package but in the packages that need them.

Only problem I discovered is that `pyyaml` doesn't discover the `yamllib` for some reason. But after hours of trying I wasn't able to fix it. It still works without this dependency anyway.

Building now for all platforms. Will share link later for testing.
Supersedes #2984 by @m4tt075.

### Checklist


ymartin59:
Please discard all sabnzbd related commit from this brach/PR. Ideally you may also squash everything in a single commit with detailed message

ymartin59:
@Safihre May you please also confirm you have got all changes from #2984 (@m4tt075) ?

Safihre:
@ymartin59 All changes from #2984 are in this one ✔️ 
Once #3053 is merged in it will be easier to rebase and squash, since the non-SAB related commits are required also for this one to work. 
I do believe we should not just make this 1 single commit. While debugging all day today it was often hard to understand why some decisions were made in parts of the code: big parts have no comments and the commit was just something like `Python wheel support`. Having a few extra commits won't hurt :)

Safihre:
Back to `[WIP]` since I now saw that not all cross-compiles worked. Will report back when fixed.

Safihre:
For documentation purposes I want to describe what I have learned after hours and hours of debugging this.

- The changes from 2.7.14 > 2.7.13 in the cross-building are pretty substantial, previous patches don't work anymore.
- We now rely on supplying `--host`, `--target` and `--build` in the compile. Since `--host` and `--build` are allready automatically added by the framework, we just need to add `--target` .
- There is no need anymore for a separate `PGEN`.
- We do need to supply `_PYTHON_HOST_PLATFORM` to trigger more rules in the `configure`.
- The main problem then is that the `_sysconfigdata.py` data file is not generated correctly. It contains data of the `native` build and not the real cross-platform. This file is later used when all extension modules are build, it contains all the linker/compiler/directory/etc-flags needed. I spend way too much time trying to patch files while it will all work fine, as long as `_sysconfigdata.py` is generated correctly.

Will update this when I find out more.

Safihre:
Getting very close! It's building on all arches now. Updated previous post (will also include in a commit-message).
Also found that `python` only supports `berkleydb` up to v5.3 and it doesn't pick up the newer v6 that's in the repo. Even Python 3 is limited to v5.3. 
I will now downgrade `cross/berkleydb` since the other 2 packages that depend on it also work just fine with `v5.3`.
Seems for the past 4 years since the update to v6.x the `dbm` module in Python didn't work, but nobody noticed!

PR #3063, opened 2017-12-29
Update Python dependencies to build other packages

Safihre:
_Motivation:_ Update Python dependencies to build other packages, it would not build otherwise on DSM6.1 tool chains.

This does not fix `spk/python`, it only allows packages that need `cross/python` (like `spk/sabnzbd`) during the building to work.


ymartin59:
May you please tell me if these changes only impact python2 package or also python3 ?

Safihre:
The `wheel`, `setuptools` and `cryptography` packages are also for `spk/python3` (not `cross/python3`). 
I haven't tried compiling `cross/python3` or `spk/python3`, but I think they might also fail to compile without changes due to the new DSM6 toolchains.
But it won't be due to these dependency changes, all 3 packages are highly compatible between the python versions.

ymartin59:
In fact `spk/python` does not build with 5.2 toolchain either...

PR #3064, opened 2017-12-30
Changes to Generic service

Safihre:
2 bugfixes and the option to use `set_syno_permissions` to correct permissions specific to package.
Tested with `spk/sabnzbd` and works as expected.

Would suggest not to squash into single commit to preserve easier commit-history reading.

### DIFF
--- a/mk/spksrc.service.installer
+++ b/mk/spksrc.service.installer
@@ -112,16 +112,22 @@ syno_group_remove ()
 }
 
 # Sets recursive permissions for ${GROUP} on specified directory
-# Usage: set_syno_permissions "${SHARE_FOLDER}"
+# Usage: set_syno_permissions "${SHARE_FOLDER}" "${GROUP}"
 set_syno_permissions ()
 {
     DIRNAME=$1
+    GROUP=$2
     echo "Granting '$GROUP' group permissions on $DIRNAME" >> ${INST_LOG}
-    VOLUME=$(echo "$DIRNAME" | awk -F/ '{print "/"$2}')
+
+    # Set read/write permissions for GROUP
+    if [ ! "`synoacltool -get "${DIRNAME}"| grep "group:${GROUP}:allow:rwxpdDaARWcC-:fd--"`" ]; then
+        synoacltool -add "${DIRNAME}" "group:${GROUP}:allow:rwxpdDaARWcC-:fd--" >> ${INST_LOG} 2>&1
+    fi
 
     # Walk up the tree and set traverse permissions up to VOLUME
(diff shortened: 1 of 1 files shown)


m4tt075:
Many thanks, @Safihre . This PR fixes the upgrade issues I have been experiencing in PR #3050.

mk/spksrc.service.installer
```diff
@@ -289,20 +296,32 @@ preupgrade ()
...
 {
     log_step "postupgrade"
-    # Restore some stuff
-    $MV "$TMP_DIR" "$TO_SAVE_DIR" >> ${INST_LOG}
+
+    # Restore some stuff, has to be cp otherwise fails on directories
+    $CP "${TMP_DIR}"/* "${INST_VAR}" >> ${INST_LOG}
+
+    # Correct permissions of var folder
+    if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
```
ymartin59: To avoid copy/paste from line 225, please create a new function `set_var_permissions`

mk/spksrc.service.installer
```diff
@@ -203,31 +208,33 @@ postinst ()
             synoshare --add "${SHARE_NAME}" "${SHARE_DESC}" "${SHARE_PATH}" "" "rw" "" 1 0 >> ${INST_LOG}
         fi
 
-        # Add user/group permissions depending if GROUP is set in UI
-        if [ -n "$GROUP" ]; then
-            synoshare --setuser "${SHARE_NAME}" RW + "${EFF_USER}",@"${GROUP}" >> ${INST_LOG} 2>&1
-        else
+        # Add user permission if no GROUP is set in UI
```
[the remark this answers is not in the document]
Safihre: If GROUP is set then `set_syno_permmissions` will add the permissions for group level, removed this to avoid doing things double :)
This command also applies slightlight different set of permissions (no `C` at end to allow for changing of permissions by GROUP), so 2 sets of permissions would be applied.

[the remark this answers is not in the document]
Safihre: That's also in the commit message ;)

mk/spksrc.service.installer
Safihre: At line 225 the function uses a different folder for DSM5, while in the new part we use the same path for both DSM's.
```shell
        chown -R ${EFF_USER}:root "${SYNOPKG_PKGDEST}" >> ${INST_LOG} 2>&1
VS:
        chown -R ${EFF_USER}:${USER} "${INST_VAR}" >> ${INST_LOG} 2>&1
```

Safihre: So to have a `set_var_permissions` we would need to use the same path for both DSM versions in the things at line 225.

Issue #3038, opened 2017-12-09
sabnzbd on Syno EDS14


Toxine68:
Setup
_Package Name:sabnzbd
_Package Version:2.3.1-33
_NAS Model:EDS14
_NAS Architecture: 
_DSM version: DSM 6.0.3-8754 Update 6
Expected behavior
It seems that is not possible to use sabnzbd on the EDS14.
It is not possible to choose the download folder.

Actual behavior
The download folder path must start with / volume1 / downloads.
But there is no Volume1, only things like usbshare2 or sdshare1-1
Steps to reproduce
1. Install the package
2. Set the download folder
3.
Package log
Check Package Center or /usr/local/{package}/var/
Insert the package log here
Other logs
E.g. /var/log/messages or /var/log/synopkg.log
Insert log here

ymartin59:
Next package will provide an installation wizard where download target and group to manage permissions can be provided specifically. It should solve your issue.

Toxine68:
thank you !
do yo have a publish date for the next package ?

ymartin59:
@Safihre To notify you about this issue

Safihre:
@Toxine68 I have a test-version of the new package, do you want to try it?
https://github.com/Safihre/spksrc/releases/tag/sabtest2 -> You need the `armada370` one.

Toxine68:
yes, i would like to try. 
But the syno server says something like "This package does not contain a digital signature"
can i go ahead and ignore this information ?
thank you

Safihre:
You can ignore that, it's because it's not directly from the SynoCommunity domain (it will be in the future).

Toxine68:
It is not possible to install it. I have DSM 6.0.3-8754 Update 6. (the latest) 
it requires a DSM version 6.1-15047

Safihre:
@ymartin59 Seems the DSM6.0 toolchains fail, can't fetch them.

ymartin59:
@Safihre I have just check all 6.0.2 toolchain download and digest. Everything is correct on `master`

Safihre:
Sorry my mistake, I specified the toolchain incorrectly.
@Toxine68 Does this one work? https://github.com/Safihre/spksrc/releases/tag/sabnzbd-dsm61-602-apt

Toxine68:
sorry it does not work. 
i can not set the download path
The download folder path must start with / volume1 / downloads.

Safihre:
[...]

PR #3065, opened 2017-12-30
Update SABnzbd to use Generic service and version 2.3.2

Safihre:
Won't build without #3064 and #3063.


Safihre:
@ymartin59 Now that I have finally got `spk/python` to work I can continue and convert related packages (`sickrage`, `couchpotato`, etc). But I don't want them to be stuck as PR for weeks. How can I make this as streamlined as possible? As in: what would you like to see before you will merge and publish packages?

spk/sabnzbd/src/wizard/install_uifile_fre
```diff
@@ -1,24 +0,0 @@
-[{
-    "step_title": "Configuration de base",
```
ymartin59: For which reason do you discard french translations ? I may provide it.

spk/sabnzbd/src/wizard/install_uifile
```diff
@@ -1,19 +1,31 @@
...
             "defaultValue": "/volume1/downloads",
             "validator": {
-                "allowBlank": false,
-                "regex": {
-                    "expr": "/^\\\/volume[0-9]+\\\//",
-                    "errorText": "Path should begin with /volume?/ with ? the number of the volume"
-                }
+              "allowBlank": false,
+              "regex": {
+                "expr": "/^\\\/volume\\w*[0-9]{1,2}\\\/[^<>: */?\"]*/",
```
ymartin59: Please relax regular expression to support `/usbshareX/` volumes according to #3038

ymartin59:
To notice/warn. I will really squash commits (for that PR and also python one) for the following reason: when investigating why a dependency or any other change has been triggered by a package update, it is really easier to understand with git log or blame if it points a single commit.
So either to do it yourself and I will preserve commit message details, or else I will append them together when squashing PR.

spk/sabnzbd/src/wizard/install_uifile_fre
Safihre: Sure, what should the texts be? 
For install and upgrade.

spk/sabnzbd/src/wizard/install_uifile
Safihre: It is already relaxed, it now has the extra `\w`. Because for USBs the path is: `/volumeUSB1/usbshare1`.
I also put that in the installation message because for new users it's pretty confusing: 
https://github.com/SynoCommunity/spksrc/pull/3065/files#diff-2d2834ae3fa939e71e1208d004e4ec56R24

ymartin59: Strange. My DSM mounts USB devices with `/usbshareX/` and comment from Toxine68 confirms too: https://github.com/SynoCommunity/spksrc/issues/3038#issuecomment-354251192
Reference documentation: https://www.synology.com/en-uk/knowledgebase/DSM/help/DSM/AdminCenter/system_externaldevice_devicelist
Where have you heard about `/volumeUSBx` ?

Safihre: Yes the _share_ is called `usbshare1`, but when you actually SSH into the device you will see that the path is `/volumeUSB1/usbshare1`. At least on my Xpenology. 
There is no `/usbshare1/` (without prefix).

Safihre:
@ymartin59 I understand then why you want to squash, but can we then be more strict about adding enough comments?
Right now almost none of the `Makefile` have any comments and the commits messages are also really limited in terms of information.
For example the `mimetype` dependency of Python, the reason why we have `x265-2.1`, this could all have been clear with more comments in `Makefile`/patches/etc.

spk/sabnzbd/src/wizard/install_uifile
ymartin59: OK, my mistake, only looking at DSM file manager. I should have checked on file system.

ymartin59:
I really really agree with you... but until now, "previous" maintainers consider such comments as useless and requested removal to merge PRs.
Since I am granted permission to merge, I do add comments everywhere and care about information to future maintainers...
By the way, more details can be written in PR description as you do now, or in commit messages.
For complex situation, my point of view is to track trouble in an "private" issue (only maintainer comment) and code comment refers to issue.

Safihre:
I used Reddit to have about 50 people try the SABnzbd and Python packages.
So far the response has been good, new installs work perfect.
In the case of updating I received from users the problems described in #3084 regarding sub-directories and `It's linux mode`. We can discuss that here (@m4tt075): https://github.com/SynoCommunity/spksrc/pull/3084#issuecomment-356062570.

In any case I modified the upgrade screen to look like this, because clearly we cannot catch every case with our `syno`-tools.
Thoughts?
image

m4tt075:
+1
I agree.

ymartin59:
I still have to provide french translations

ymartin59:
@Safihre Here are french translations
[sabnzbd_upgrade_uifile_fre.txt](https://github.com/SynoCommunity/spksrc/files/1636292/sabnzbd_upgrade_uifile_fre.txt)
[sabnzbd_install_uifile_fre.txt](https://github.com/SynoCommunity/spksrc/files/1636293/sabnzbd_install_uifile_fre.txt)

Safihre:
@ymartin59 @m4tt075 What are the current issues that are preventing the new-style packages to be released? So I can get up-to-date with everything again!
- Something is wrong with `set_syno_permissions` (?)
- No app-icon is being created (@ymartin59 probably that JSON that needs fixing)
- Python 2.7.14 needs a fix (I'll dive into that)
- Sonarr/Radarr package should be inspected for correct permissions on `.NzbDrone`
- Anything else?

This is all taking much much longer than I hoped initially... @ymartin59 If `set_syno_permissions` is fixed and the app-icon thing, we can at least release NZBGet, right?
And even SABnzbd actually (will add French texts now)?

m4tt075:
I'm not really sure whether `set_syno_permissions` needs fixing or whether that is more of a style-issue. From all my testing the ACL permissions "override" the unix permissions, so it does not really matter anymore how those are set. I have not heard about any problems with my TVH test builds either. I cannot say how many people have actually installed them, but there have been close to 270 downloads of those packages in the meantime. @Safihre Have you heard anything about permission setting issues with your test builds? And what is your download count? Pretty sure, way higher than mine...  :smirk:

In any case, personally happy to wait until everybody is comfortable. Just waiting until the dust settles before I will start working on further package adaptations.

Safihre:
@ymartin59 I added the French texts and removed that same `dirname` in the `SHARE_PATH` detection that also caused problems in NZBGet.

Safihre:
@ymartin59 Forgive my directness, but what are we waiting for with SABnzbd and NZBGet?
The test packages (#3138) for both have almost 200 downloads each and, as far as I know, have no other dependencies/issues that need to be resolved.
If we are afraid of possible problems (I can understand, after Python-package problems), we could just apply these changes to `sabnzbd-testing`/`nzbget-testing` and release them for more contained real-life tests. 
There is nothing more for these 2 packages that we can do in terms of "lab-testing", we need real-world tests. I hope you agree :)

Safihre:
Pushed update to handle permissions-fixing in `postupgrade`, which was handled before in `postinstall` for some reason.
Will apply the same to NZBGet.

Safihre:
Oke now everything is fixed and ready 🥇 
Also pushed the new builds for SABnzbd and NZBGet to the download-links listed in #3138.

ymartin59:
@Safihre I would like to have `set_syno_permissions` fixed first because releasing.

remirr:
@ymartin59 I have just had to reset up a brand new DS918+ after having major file system problems on my DS415+. Am I better off waiting a few more days to install SABnzbd and python or are the packages stable? From the above it seems like there are some problems with the permissions?

Safihre:
@remirr The problems are related to upgrades, not new installations.

remirr:
Hi @Safihre, thanks for the fast response. So am I ok to go ahead and install the latest beta of Python and SABnzbd? or are the packages imminent on the SynoCommunity store?

Thanks,

Safihre:
I am afraid it might be more time, so to ahead and use these.

remirr:
@Safihre Thanks a lot - I will go ahead and use there. When the final versions come out, will I be able to upgrade or will I need to uninstall and reinstall?

remirr:
@Safihre Installed the beta Python Package and then tried to install the approved Sickbeard and Couchpotato package on the SynoCommunity store and am getting the following error "The following packages are missing in the package server: Please contact Synology for help: python"

Safihre:
You also need to use the test packages for all of them as long as Python isn't on the official package repo yet. These are also listed in #3138.

remirr:
@Safihre Awesome, thanks!

remirr:
@Safihre there doesn't seem to be any Appololake packages for Sickbeard and Couchpotato, am I missing something?

https://github.com/Safihre/spksrc/releases/tag/sickbeardcustom-dsm61-new
https://github.com/Safihre/spksrc/releases/tag/cp-dsm61-new-all

m4tt075:
@remirr "noarch" means they can be used for all platforms. Go for it. You can do it...

Safihre:
@BenjV @ymartin59 I included in this PR the fixes of the removal of users and the fixing of the icon.

It is ready now. Want to merge this very soon so I can base the other packages on it and do better testing.

remirr:
Any updates on when
- Python
- Sabnzbd

Will be released final for apollolake? 
- Sickbeard and Couchpotato seem to be there.

Safihre:
@remirr When they are done and we are sure automatic updates don't break anything on any system. 
Yes, this current round of updates is taking a while, but it's a major design change and after it's confirmed to work well future updates and new packages will be super easy due to the work of @ymartin59 on the new framework.

Safihre:
@ymartin59 Include 2 new commits to set default toolchain to DSM61 and to automatically include a link to the actual package-homepage.
This way users can actually know where to get more information about the actual package.
image

ymartin59:
@Safihre Hum... why these two commits now land in SABnzbd-related PR ??

Safihre:
Because I need a package to test them with and I am already shuffling so many things from branch to branch all the time.
It doesn't really matter in which PR these are, since they are small changes. When adding to `master` they shouldn't of course be squashed into 1 commit.

ymartin59:
Let that as is... even if I do not agree - as these 5 commits pending merge is in a all-or-nothing mode.

Safihre:
@ymartin59 Oke, I removed all the other commits.

PR #3032, opened 2017-12-05
jackett: update to v0.8.490

kaso17:
_Motivation:_ This should fix various jackett specific issues
_Linked issues:_ fixes #3015 #2986 #2881 #1995 https://github.com/Jackett/Jackett/issues/1985 https://github.com/Jackett/Jackett/issues/2217

### Checklist
- [X] Build rule `all-supported` completed successfully
- [X] Package upgrade completed successfully (Only tested on DSM 6.1)
- [X] New installation of package completed successfully (Only tested on DSM 6.1)

Issue #3015, opened 2017-11-26
Compiling jackett for rtd1296


kianrafiee:
I followed the guide and even tested that I could build another package first. I tested builing beets first for rtd1296-6.1 toolchain. It created a package in /packages/

However, after cross compiling jackett. It creates a work directory which I followed the guide to then take the plist there and add it to spk/jackett with rsc: in front of every share.

Then i created the digest in terminal.

Then I copied and put the PLIST and the MAKEFILE into spk/jackett and then tried to compile by running "make arch-rtd1296-6.1"

The problem is it keeps doing the same thing that happens when you cross compile which is create a work directory. It doesnt create a package.

Here is the terminal text:

```
make arch-rtd1296-6.1
===>  Building package for arch rtd1296-6.1
make[1]: Entering directory '/home/kian/spksrc/spk/jackett-rtd1296'
===>  Set up toolchain 
make[2]: Nothing to be done for 'default'.
===>  Downloading files for Jackett
===>    File Jackett-0.7.1483.tar.gz already downloaded
===>  Verifying files for Jackett
===>  No digests file for Jackett
/home/kian/spksrc/spk/jackett-rtd1296/../../distrib/Jackett-0.7.1483.tar.gz
===>  Processing dependencies of Jackett
===>  Extracting for Jackett
tar -xzpf /home/kian/spksrc/spk/jackett-rtd1296/../../distrib/Jackett-0.7.1483.tar.gz -C /home/kian/spksrc/spk/jackett-rtd1296/work-rtd1296-6.1 
===>  Patching for Jackett
===>  Configuring for Jackett
===>  Compiling for Jackett
===>  Installing for Jackett
find /home/kian/spksrc/spk/jackett-rtd1296/work-rtd1296-6.1/install//usr/local/ \! -type d -printf '%P\n' | sort > /home/kian/spksrc/spk/jackett-rtd1296/work-rtd1296-6.1/Jackett.plist.tmp
mkdir -p /home/kian/spksrc/spk/jackett-rtd1296/work-rtd1296-6.1/install/usr/local/share/jackett
tar -cf - -C /home/kian/spksrc/spk/jackett-rtd1296/work-rtd1296-6.1/Jackett . | tar -xf - -C /home/kian/spksrc/spk/jackett-rtd1296/work-rtd1296-6.1/install/usr/local/share/jackett
[...]

Issue #2986, opened 2017-11-02
[Package Request] Update Jackett


ErwanNumb:
Name: Jackett

The latest version of Jackett brings - a lot - of new torrent sources, could you please update it?

Thanks

ErwanNumb:
still no interest in doing this?

Safihre:
From what I understand Jackett can just use the in app updater?

ymartin59:
Related to #3070

m4tt075:
@ErwanNumb Thanks to @kaso17 's fixes, `Jackett 0.8.490` has been merged into master. That means that the packages for `DSM 5.2` systems should be available via the SynoCommunity repo shortly (if they aren't already). For the time being, you can find `DSM 6.1` builds for `Jackett` on [this link here](https://github.com/m4tt075/spksrc/releases/tag/Jackett-0.8.490). Once installed, you can run further upgrades via the Jackett web-interface. These are beta packages. Therefore, please do report any kind of issues you should encounter.

ymartin59:
Mono and Jackett updates published

Issue #2881, opened 2017-08-18
Provider t411.si don't work in Jackett


solspire:
I change the wrong url in jackett in "site link" 
because isn't https://t411.al/ but now is https://t411.si/
I need jackett + mylar for have an automated ebook downloader and is it the only french comics provider !

Can you make T411 work ?

For now I have this for t411 (the others work fine):
"Configuration failed: Unexpected character encountered while parsing value: <. Path '', line 0, position 0." and this also "Configuration failed: Sequence contains no matching element"

Nas synology DS414 last update of DSM and package of synocomunity.

ymartin59:
Related to #3070

m4tt075:
@solspire Thanks to @kaso17 's fixes, `Jackett 0.8.490` has been merged into master. That means that the packages for `DSM 5.2` systems should be available via the SynoCommunity repo shortly (if they aren't already). For the time being, you can find `DSM 6.1` builds for `Jackett` on [this link here](https://github.com/m4tt075/spksrc/releases/tag/Jackett-0.8.490). Once installed, you can run further upgrades via the Jackett web-interface. These are beta packages. Therefore, please do report any kind of issues you should encounter.

ymartin59:
Mono and Jackett updates published

PR #3070, opened 2018-01-01
Update Jackett to v0.8.490 and to use generic DSM 5+6 services

m4tt075:
### Motivation
Contains @kaso17 's Jackett updates and fixes of #3032 and adapts the package to use generic DSM 5+6 services

### Linked issues
#3015, #2986, #2881, #1995, Jackett/Jackett#1985, Jackett/Jackett#2217

### Notes
- ~~Depends on merging of @kaso 's Jackett update PR #3032.~~ Squashed into this PR on request of @ymartin59 and with agreement of @kaso17  
- ~~Depends on merging of #3064. Do NOT merge without it or old configurations will be overwritten on upgrades.~~ #3064 merged in the meantime.
- ~~Further testing needed.~~ Outdated.

### Checklist
- [x] Build rule `all-supported` completed successfully on `XPE 5.2` and `XPE 6.1` systems.


m4tt075:
@kaso17 As to your hints / notes in #3032:
- I have tested the package on an `XPE 5.2` and `XPE 6.1` system. It installs and runs as far as I can judge. At least I can access the web interface in both cases.
- User premissions for the `share` folder seem correct, because I could succesfully upgrade the package using the internal upgrade process on both systems.
- I could not test a migration from a legacy installation as I don't have one, but would rely on the generic service scripts to reset user permissions correctly as this applies to all packages.
- I could verify that all environment variables are set and used correctly.
- The only piece I could not implement from your approach was the logging functionality as this is driven through the generic service scripts. As such, standard logging to `var/jackett.log` is used for now. It works to the extend that both `DSM` and `jackett` log into that file correctly. It can be accessed through the DSM web-UI. However, your (very nice) `log rotation` functionality is lost unfortunately. I don't know how much information is usually written to the logs, but if you foresee an issue here, please let me know.

kianrafiee:
newly installed jackett package from @Safihre repo will not start. it was starting fine yesterday and updated. but now it wont start. not sure if there was a jackett update yesterday that broke it

m4tt075:
Not related to this PR or spksrc development. Therefore, please report this in the `Issues` section instead, either in one of the old threads or open a new one.

spk/jackett/Makefile
```diff
@@ -1,33 +1,38 @@
...
 
 DEPENDS = cross/busybox cross/curl cross/$(SPK_NAME)
-SPK_DEPENDS = "mono>=4.2.1.102-6"
+# depend on mono 5.4 for BTLS HttpWebClient support
+SPK_DEPENDS = "mono>=5.4"
 
-MAINTAINER = SynoCommunity
-DESCRIPTION = "Jackett works as a proxy server: it translates queries from apps into tracker-site-specific http queries, parses the html response, then sends results back to the requesting software. Jackett is a single repository of maintained indexer scraping & translation logic - removing the burden from other apps."
-ADMIN_PORT = 9117
+MAINTAINER = 'kaso17 (Jackett), m4tt075 (spksrc integration)'
```
ymartin59: I wonder if `spkrepo` will accept packages with such maintainer. As far as I understood/remember, it has to match a user account registered in registry

spk/jackett/Makefile
```diff
@@ -1,33 +1,38 @@
...
-SSS_SCRIPT       = src/dsm-control.sh
-FWPORTS          = src/${SPK_NAME}.sc
-CONF_DIR         = src/conf/
+# Service configuration
+SERVICE_PORT = 9117
+SERVICE_PORT_TITLE = Jackett(HTTP)
 
-INSTALL_PREFIX = /usr/local/$(SPK_NAME)
+# Admin link
+ADMIN_PORT = $(SERVICE_PORT)
```
ymartin59: Is it relevant to consider SERVICE_PORT as ADMIN_PORT ? Probably not, but I prefer to keep it for "backward compatibility"

spk/jackett/src/service-setup.sh
```diff
@@ -0,0 +1,65 @@
...
+    PATH="${SYNOPKG_PKGDEST}/bin:${MONO_PATH}:${PATH}"
+    MONO="${MONO_PATH}/mono"
+    JACKETT="${SYNOPKG_PKGDEST}/share/${SYNOPKG_PKGNAME}/JackettConsole.exe"
+    HOME_DIR="${SYNOPKG_PKGDEST}/var"
+
+    echo "Starting Jackett as user ${EFF_USER}" >> ${LOG_FILE}
+    COMMAND="env HOME=${HOME_DIR} PATH=${PATH} LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib ${MONO} --debug ${JACKETT} --PIDFile ${PID_FILE}"
+
+    if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
+        su ${EFF_USER} -s /bin/sh -c "cd ${SYNOPKG_PKGDEST}; ${COMMAND}" >> ${LOG_FILE} 2>&1 &
```
ymartin59: Is it really needed to change process working directory to SYNOPKG_PKGDEST ? (I did it on `demoservice` to browse files...)
I think HOME_DIR may be more relevant, and it is already done by `su`

spk/jackett/src/service-setup.sh
```diff
@@ -0,0 +1,65 @@
...
+    # Replace generic service startup
+
+    MONO_PATH="${SYNOPKG_PKGDEST}/../mono/bin"
+    PATH="${SYNOPKG_PKGDEST}/bin:${MONO_PATH}:${PATH}"
+    MONO="${MONO_PATH}/mono"
+    JACKETT="${SYNOPKG_PKGDEST}/share/${SYNOPKG_PKGNAME}/JackettConsole.exe"
+    HOME_DIR="${SYNOPKG_PKGDEST}/var"
+
+    echo "Starting Jackett as user ${EFF_USER}" >> ${LOG_FILE}
+    COMMAND="env HOME=${HOME_DIR} PATH=${PATH} LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib ${MONO} --debug ${JACKETT} --PIDFile ${PID_FILE}"
```
ymartin59: Is `--debug` flag really relevant ?

spk/jackett/src/service-setup.sh
```diff
@@ -0,0 +1,65 @@
...
+
+    if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
+        su ${EFF_USER} -s /bin/sh -c "cd ${SYNOPKG_PKGDEST}; ${COMMAND}" >> ${LOG_FILE} 2>&1 &
+    else
+        cd ${SYNOPKG_PKGDEST}
+        ${COMMAND} >> ${LOG_FILE} 2>&1 &
+    fi
+}
+
+service_preinst ()
```
ymartin59: Please discard all other hook functions with only logging... I only wrote them for demo, I consider them not relevant for a package

spk/jackett/Makefile
m4tt075: Sorry, not sure I understand. Do you want me to remove the comment and the ADMIN_PORT definition or do you want me to keep it?

kaso17:
@m4tt075 thank you for taking care of this.
Did you test if the jackett internal updating mechanism works?

In case you didn't:
Please go to the jackett admin panel (port 9117) and click on the yellow "Check for updates" button. Wait a little until jackett updates itself and refresh the page. At the bottom of the page it should say version 0.8.539.0 (or newer).

I don't see that the bin/share directory `${SYNOPKG_PKGDEST}/share/${SYNOPKG_PKGNAME}` is chmod/chowned anywhere anymore. If the jackett user can't write to that directory the update won't work. I assume the /var directory is chowned to the new user during the update? Besides that I don't see any obvious issues.

m4tt075:
@kaso17 All fine. Thanks for reviewing...

spk/jackett/src/service-setup.sh
```diff
@@ -0,0 +1,35 @@
...
+    JACKETT="${SYNOPKG_PKGDEST}/share/${SYNOPKG_PKGNAME}/JackettConsole.exe"
+    HOME_DIR="${SYNOPKG_PKGDEST}/var"
+
+    echo "Starting Jackett as user ${EFF_USER}" >> ${LOG_FILE}
+    COMMAND="env HOME=${HOME_DIR} PATH=${PATH} LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib ${MONO} ${JACKETT} --PIDFile ${PID_FILE}"
+
+    if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
+        su ${EFF_USER} -s /bin/sh -c "${COMMAND}" >> ${LOG_FILE} 2>&1 &
+    else
+        cd ${SYNOPKG_PKGDEST}
```
ymartin59: You removed "cd" in `COMMAND` for DSM 5 and so it should be removed here for DSM 6 too.

spk/jackett/Makefile
ymartin59: It depends if link generated in package view is really usable (often `http://ip:ADMIN_PORT/`) whereas `SERVICE_PORT` is used to provide application link in `app/config` which allows finer settings like `https` and context path in url.
So if "admin link" in package view is not usable as-is, please remove it. If it works, let it there.

m4tt075: It is usable and works correctly, so I kept it.

spk/jackett/src/service-setup.sh
m4tt075: Removed.

m4tt075:
@ymartin59 I have addressed your latest comments. Anything else?

ward0:
When can we see an update pls? @m4tt075

m4tt075:
@ymartin59 @kaso17 Following up with your recent comments in #3032, I have...
- rebased
- squashed all commits (including @kaso17 's) into one
- Adapted the PR title
- Adapted the PR description to keep the linked issues from @kaso17 's PR description
- Re-compiled Jackett for bromolow architecture 5.2 and 6.1, just to make sure
- Successfully upgraded the package on XPE 6.1 within the package manager and then again, using Jackett's internal upgrading process
- Successfully installed the package on XPE 5.2 and upgraded it using Jackett's internal upgrading process

Should be ready to go now...

ymartin59:
Marvelous. Merging.
By the way, online repository rejects mono 5.4 package for x64... so it may not be published in a short term.

spk/jackett/src/service-setup.sh
```diff
@@ -0,0 +1,34 @@
+
+# Package specific behaviors
+# Sourced script by generic installer and start-stop-status scripts
+
+service_prestart ()
+{
+    # Replace generic service startup
+
+    MONO_PATH="${SYNOPKG_PKGDEST}/../mono/bin"
```
ymartin59: Bad idea... Prefer to use `/usr/local/mono/bin`

spk/jackett/src/service-setup.sh
```diff
@@ -0,0 +1,34 @@
...
+    PATH="${SYNOPKG_PKGDEST}/bin:${MONO_PATH}:${PATH}"
+    MONO="${MONO_PATH}/mono"
+    JACKETT="${SYNOPKG_PKGDEST}/share/${SYNOPKG_PKGNAME}/JackettConsole.exe"
+    HOME_DIR="${SYNOPKG_PKGDEST}/var"
+
+    echo "Starting Jackett as user ${EFF_USER}" >> ${LOG_FILE}
+    COMMAND="env HOME=${HOME_DIR} PATH=${PATH} LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib ${MONO} ${JACKETT} --PIDFile ${PID_FILE}"
+
+    if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
```
ymartin59: These if statements are same as generic script.
I think moving all variables settings out of `service_prestart` (so removing this function) and provide `SERVICE_COMMAND` at the begging of this script should work.

Safihre: No that won't work because the process never detaches: no `&` at the end when using SERVICE_COMMAND` for processes that don't support deamon.
So DSM will stay in a constant state of `Starting` since the `start` never returns.
So that's why we need this special function. Or integrated in generic approach :)

m4tt075: *LOL* Just came here to write the same! Agree with @Safihre . This is another candidate for the `start-stop-daemon` approach. I propose to keep this package as is for the time being and update once that approach is established...

m4tt075:
@ymartin59 This has been merged 6 weeks ago. Anything preventing this from being published?

ymartin59:
@m4tt075 Yes... Mono update. The fix I designed to relocate `cert-sync` seems to work. My DSM has a pending update, I will test everything is ok with mono+jackett upgrade

Safihre:
@jdeluyck Could you restart your NAS and try again?
What model NAS do you use?

jdeluyck:
I'm using a DS916+. 
Rebooted NAS, retried upgrade, same story.

ymartin59:
Hum hum. Corrupt package ? It remembers me trouble with signing process in repository... I have already fixed it (and there was no issue with recent published packages like mono, ffmpeg...) so probably there is a specific failure with Jackett. I am investigating....

ymartin59:
@jdeluyck As an immediate work-around, open Package Center "Settings" and select "Any publishers" in Trust Level section... sorry for the inconvenience

Safihre:
Simple re-publish is not fixing it? I indeed also see this error..

jdeluyck:
@ymartin59 that indeed works. So the package is ok, signing is borked.

PR #3073, opened 2018-01-02
Various dependency fixes (fribidi, openssl, x264)

m4tt075:
### Motivation
When experimenting with travis-CI on my private `spksrc` repo, I encountered several compile issues due to outdated cross-packages. This PR addresses those issues. 

### Issues
- fribidi: Download location for sources no longer available. Replaced by alternative source (same package).
- openssl: No issue actually, but realized that we are running two releases behind. Updated to 1.0.2n.
- x264: The old download location still worked, but required the `--no-check-certificate` option for `wget` to work. Adapting the download location to the (new) official source URL fixes the issue.

### Note
I conducted runtime-tests but only on individual packages for individual platforms. Based on the nature of the changes I would not expect any kind of issues though...

### Checklist


yuezhu:
👍

PR #3080, opened 2018-01-05
Python 2.7.14 - Fix for DSM6 and Generic Service Support 

Safihre:
For documentation purposes I want to describe what I have learned after hours and hours of debugging this.

- The changes from 2.7.14 > 2.7.13 in the cross-building are pretty substantial, previous patches don't work anymore.
- We now rely on supplying `--host`, `--target` and `--build` in the compile. Since `--host` and `--build` are already automatically added by the framework, we just need to add `--target` .
- There is no need anymore for a separate `PGEN`.
- We should supply `PYTHON_FOR_BUILD` as native/python, but the basic detection in `configure.ac` isn't enough since it doesn't supply `_PYTHON_HOST_PLATFORM` to `setup.py` (needed to enable cross-compiling of Python internal modules).
- The main problem was that the `_sysconfigdata.py` data file was not generated correctly. It contains data of the `native` build and not the real cross-platform. This file is later used when all extension modules are build, it contains all the linker/compiler/directory/etc-flags needed. I spend way too much time trying to patch files while it will all work fine, as long as `_sysconfigdata.py` is generated correctly. When we let `configure` do it's job, it will generate the correct  `_sysconfigdata.py`.
- In the `install` step of Python it's a complicated situation where we need native/python to compile things but using the `_sysconfigdata.py` of the cross-compiled modules. Therefore we need to copy the library directories of native/python but use the cross-compiled `_sysconfigdata.py`, this needs to be patched in `configure.ac`.
- On `x64` and `x86` systems the `setup.py` will detect that the build-platform (Ubuntu/Debian) supports multi-arch compiling, even though we are cross-compiling. This results in wrong `-I` command-line options to be added to `_ctypes` building. So we need a patch to disable this detection in `setup.py`.
- Using the container-based TravisCI there seems to be a version of `ncursesw` installed, which causes the `configure` to pick it up. However, this breaks with the cross-compile which has a newer version of `ncurses` and the `setup.py` only wants v5 of `ncursesw`. So I had to use the (slower and more often failing) `docker`-based setup. Should be able to fix this in future!
- `pip` can also get dependencies, which makes `requirements.txt` easier.
- `M2Crypto` newest version doesn't compile well, kept it on old version.
- Python relies on `berkeleydb` version 5.3, the newer 6.x version in the repo isn't picked up. Therefore reverted it so the `dbm` module of Python is build. Other spk's using `berkeleydb` are also just fine with the older version.

### Checklist


cross/berkeleydb/Makefile
```diff
@@ -1,12 +1,14 @@
 PKG_NAME = db
-PKG_VERS = 6.2.23
```
ymartin59: Please do not downgrade package because of impact on other packages.
I propose you create `cross/berkeleydb-5.3` with same content and make `python` depends on this specific version.

cross/python/PLIST
```diff
@@ -9,5 +9,4 @@ lnk:lib/libpython2.7.so
 lib:lib/libpython2.7.so.1.0
 rsc:lib/python2.7
 lib:lib/python2.7/lib-dynload/*.so
-rsc:etc/mime.types
```
ymartin59: I wonder about this removal impact.
I guess (to confirm with git blame) that `mime.types` has been included here because needed by a module generated as wheel by an application (for instance)

cross/python/patches/001-mimetypes.patch
```diff
@@ -1,19 +0,0 @@
...
--    "/etc/mime.types",
--    "/etc/httpd/mime.types",                    # Mac OS X
--    "/etc/httpd/conf/mime.types",               # Apache
--    "/etc/apache/mime.types",                   # Apache 1
--    "/etc/apache2/mime.types",                  # Apache 2
--    "/usr/local/etc/httpd/conf/mime.types",
--    "/usr/local/lib/netscape/mime.types",
--    "/usr/local/etc/httpd/conf/mime.types",     # Apache 1.2
--    "/usr/local/etc/mime.types",                # Apache 1.3
-+    "@INSTALL_PREFIX@/etc/mime.types",
```
ymartin59: Am I wrong or it is inconsistent with `etc/mime.types` file removal in package ?

mk/spksrc.common.mk
```diff
@@ -15,7 +15,7 @@ RUN = cd $(WORK_DIR)/$(PKG_DIR) && env $(ENV)
 
 # Pip command
 PIP ?= pip
-PIP_WHEEL = $(PIP) wheel --no-binary :all: --no-deps --requirement $(WORK_DIR)/wheelhouse/requirements.txt --wheel-dir $(WORK_DIR)/wheelhouse --build-dir $(WORK_DIR)/wheelbuild
```
ymartin59: Please leave `--no-deps` and use `pip freeze` command to generate requirement file with module explicit versions. That is required to have reproduceable package generation even many months and updates later. 
Refer to https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-packages for details

spk/python/Makefile
```diff
@@ -1,36 +1,35 @@
...
+DEPENDS += cross/pyalsa cross/pyaudio cross/pycrypto cross/pycurl
+DEPENDS += cross/pyyaml cross/pyzmq cross/uwsgi
 
 WHEELS = src/requirements.txt
 
-MAINTAINER = Diaoul
-DESCRIPTION = Python Programming Language.
-DESCRIPTION_FRE = Langage de programmation Python.
-DESCRIPTION_SPN = Lenguaje de programación Python.
+MAINTAINER = Safihre
```
ymartin59: I appreciate you volonteer to maintain package. Seeing job you did on this package, I am confident about it.

spk/python/Makefile
```diff
@@ -1,36 +1,35 @@
...
 DISPLAY_NAME = Python
-CHANGELOG = "1. Update to Python 2.7.13<br>2. Update modules"
+CHANGELOG = "1. Update to Python 2.7.14<br>2. Update modules"
 
 HOMEPAGE = http://www.python.org
 LICENSE  = PSF
 
-INSTALLER_SCRIPT = src/installer.sh
-SSS_SCRIPT       = src/dsm-control.sh
+STARTABLE = no
```
ymartin59: Redundant. See line 24...

spk/python/src/requirements.txt
```diff
@@ -1,49 +1,34 @@
...
 ## Cross-compiled via spksrc.wheel.mk
 cheetah==2.4.4
 markupsafe==0.23
 psutil==5.0.1
 pymongo==3.4.0
 sqlalchemy==1.1.4
-twisted==16.6.0
 zope.interface==4.3.3
+twisted
+pyopenssl
```
ymartin59: See previous comment about explicit module versioning generated by `pip freeze`

spk/python/src/service-setup.sh
```diff
@@ -0,0 +1,17 @@
+PATH="${SYNOPKG_PKGDEST}/bin:${PATH}"
+
+service_postinst ()
+{
+    # Install the wheels
+    ${SYNOPKG_PKGDEST}/bin/pip install --no-deps --no-index -U --force-reinstall -f ${SYNOPKG_PKGDEST}/share/wheelhouse ${SYNOPKG_PKGDEST}/share/wheelhouse/*.whl >> ${INST_LOG}
```
ymartin59: I would add stderr redirection to file too with `2>&1` (here and in following lines)

cross/python/PLIST
ymartin59: @Diaoul What do you think about this `etc/mime.types` removal in `cross/python` package ? Do you remember what which reason that file was required at the first place ? Thank you in advance for your help

Diaoul: I think this file is required for the [`mimetypes` module](https://docs.python.org/2/library/mimetypes.html) to work.

spk/python/Makefile
```diff
@@ -1,36 +1,35 @@
...
+SERVICE_SETUP = src/service-setup.sh
 
 INSTALL_PREFIX = /usr/local/$(SPK_NAME)
```
ymartin59: Please remove `INSTALL_PREFIX` to use default one.

ymartin59: And please remove `busybox` dependency and related variables in `Makefile` as it is no longer used as far as I know.

cross/python/PLIST
Safihre: Aah, I will double check that and re-add if it indeed brakes this module.

mk/spksrc.common.mk
Safihre: Alright, that makes sense.

cross/berkeleydb/Makefile
Safihre: @Diaoul What do you think? I checked and these other 2 packages need v2 or v3 of `berkeleydb`, so they won't be affected.
We could also just leave it out, since this has been broken since 2012 and nobody really noticed. 
Creating an extra cross-package to the already giant list seems a bit much, or maybe that's just me?

Safihre: (It hasn't been noticed because if it's not available the `dbm` python module still works but is slower: https://docs.python.org/2.7/library/anydbm.html#module-anydbm)

Safihre:
Thanks for the double check @ymartin59! 

The (new) patches are all mine 😎 

Compiling [the spk for all archs](https://travis-ci.org/Safihre/spksrc/builds/325510950) revealed minor other issues:
- The `hi3535` doesn't compile due to a problem with `gpg-error` not recognizing the `arm-cortexa9-linux-gnueabi` triplet. I see however that `comcerto2k` also used to be `arm-cortexa9-linux-gnueabi` in the DSM5 toolchains but is now `arm-unknown-linux-gnueabi` in the DSM6 toolchains. So I am thinking of patching the `makefile`/`configure` of `gpg-error` to detect `arm-cortexa9-linux-gnueabi` as `arm-unknown-linux-gnueabi`.
- `libyaml` isn't picked up by `pyyaml` during the building, will have to see how to fix that. From what I can tell this might have been broken for a while.
- Not sure of `cross/readline` is compiling correctly. Python picks it up by seeing some errors in the log of the actual `libreadline` files.

Side-note: Wish I found out sooner that you can let the framework only recompile specific `cross` packages by removing the special "cookie" files (`.XXX_compile_done`, `XXX_patch_done`, etc). I was recompiling all modules all the time a few days ago.. Takes ages! 😭

ymartin59:
Target `all-supported` should now include new arch like broadwell, rtd1296... I just checked "your" python builds them too.
I also discovered "cookie" file too late from my point of view. Wiki is lacking specific developer "tips and tricks" for new comers... I should do that effort.
I also faced this arch triplet change with arm - probably with ffmpeg update... which is quite complex for platform specific supports: https://github.com/SynoCommunity/spksrc/blob/master/cross/ffmpeg/Makefile
As you can see I split `cross/x265` when updating to keep ppc support - what I propose to do for `berkeleydb` too.
I desperate about online repository to accept any new packages... performance is too bad, process crash often. I have to pause package maintenance to investigate there.

cross/python/PLIST
Safihre: Confirmed, this patch is needed. Will re-add!

Safihre:
@ymartin59 Re-based.
All things are now fixed, including the `hi3535` problem (#3085), the `berkleydb`-fix and also `PyYAML` now works as expected.

Squashed in 1 giant commit with an equally giant commit message 📦

Safihre:
@ymartin59 I finally understand why `spk/python` has `busybox` dependency:
This way pure-Python packages like Sickrage/Couchpotato can just use Python's user-management commands and stay `norarch` . 
So I suggest to re-add `busybox` to this package.

Safihre:
@ymartin59 I think this is the source of our problems, a typo:
https://github.com/SynoCommunity/spksrc/pull/3080/files#diff-b4b826d941f5a852505e1a33d59ced36R16
`INSTALL_DIR` should be `SYNOPKG_PKGDEST`.
Resulting in Busybox not being installed, resulting in `start-stop-deamon` not working.

Will you fix it or should I make PR?

ymartin59:
I wonder why packages you have delivered to testers do not suffer from this issue, or nobody has reported it...

Safihre:
Because they also use the new packages, which don't use `start-stop-deamon` but generic service approach 🤒

ymartin59:
And corrupted `mosquitto` packages... I am not lucky with my publications... Have to check and rebuild everything.

ymartin59:
@Safihre So I guess application packages depend on busybox scripts included in python package too.

Safihre:
Yeah, they also use the `adduser`/`removeuser` commands (see a few posts above).  Otherwise all python packages can't be `noarch`.

Diaoul:
This was the solution I came up with, not the best but much more user
friendly than creating a separate dependency on a BusyBox only package.
Plus we already cross compile so meh.

ymartin59:
@Diaoul Sure. Number of architectures is increasing, and now C++ based application has to been built for both DSM 5.2 and DSM 6.x because of `libstdc++` version change. So I try to reduce number of builds and package volumes. For instance both `tvheadend` and `mediainfo` embed `ffmpeg` libraries, so I think of a cross-package compilation to share them so that packages are lighter.

Issue #1932, opened 2015-11-13
Improve directory permissions handling


Dr-Bean:
Over time, we've seen quite the number of issues related to setting permissions. In the majority of cases, it's because people intend to use a subdir of a share as a main location, and the installer does not handle permissions for the intermediate dirs. We don't inform users that this means they need to set permissions manually which doesn't help the situation.

To illustrate: 
- `/volume1/chosen_dir` is handled correctly;
- `/volume1/intermediate/chosen_dir` is not handled, because traverse permissions are required -but not set- for the intermediate dir(s).

The following function can be called from `postinst` with a dir as argument. It sets r/w permissions on the chosen dir itself, and sets traverse permissions for all intermediate dirs.

This also removes the need to set Unix permissions  for the `users` group (as in: chmod/chown which we do in a number of installers), which is something we should start considering imo.
Along with that, with minor changes, the function can be switched to set permissions for an application user instead of a group (if Synology still plans to implement it, I don't think it's in the current beta)

```
syno_set_permissions ()
{
    dirname=$1
    SYNO_VOLUME=`echo $1 | awk -F/ '{print "/"$2}'`
    # Set read/write permissions for SYNO_GROUP
    synoacltool -add "${dirname}" "group:${SYNO_GROUP}:allow:rwxpdDaARWc--:fd--"
    while test "${dirname}" != "${SYNO_VOLUME}"; do
        # Walk up the tree and set traverse permissions up to SYNO_VOLUME
        dirname="$(dirname "${dirname}")"
        if [ ! "`synoacltool -get \"${dirname}\"|grep \"group:${SYNO_GROUP}:allow:..x\"`" ]; then
            synoacltool -add "${dirname}" "group:${SYNO_GROUP}:allow:--x----------:---n"
        fi
    done
}
```

In practice, postinstall () would look like this for a download-type package:

```
    # Create user
    adduser
    # Create SYNO_GROUP
    syno_group_create

    if [ "${SYNOPKG_PKG_STATUS}" == "INSTALL" ]; then
        # Set group and permissions on dir
        if [ `/bin/get_key_value /etc.defaults/VERSION buildnumber` -ge "4418" ]; then
            syno_set_permissions "${wizard_download_dir}"
        fi
    fi
```

/edit: added some quotes here and there to fix potential issues with spaces in dirs and groupnames.
##

Diaoul:
[...]

PR #3084, opened 2018-01-07
Improve permission setting for shared folders in generic service scripts

m4tt075:
_Motivation:_ While working on adapting the Tvheadend package for generic service use, I have encountered two issues with automatic permission setting for shared directories within `mk/spksrc.service.installer`:
1. The paths of shared directories containing symbolic links are not resolved.
2. If the path of a shared directory does not reside in a `/volume[X]` folder, the permission setting function can recurse into system directories, which is certainly not wanted. 

This PR addresses those issues.

### Checklist
- Tested via automatic permission setting on directories with and without symbolic links, within and outside of `/volume[X]` directories.

### DIFF
--- a/mk/spksrc.service.installer
+++ b/mk/spksrc.service.installer
@@ -115,24 +115,31 @@ syno_group_remove ()
 # Usage: set_syno_permissions "${SHARE_FOLDER}" "${GROUP}"
 set_syno_permissions ()
 {
-    DIRNAME=$1
+    DIRNAME=`realpath $1`
     GROUP=$2
-    echo "Granting '$GROUP' group permissions on $DIRNAME" >> ${INST_LOG}
 
-    # Set read/write permissions for GROUP
-    if [ ! "`synoacltool -get "${DIRNAME}"| grep "group:${GROUP}:allow:rwxpdDaARWcC-:fd--"`" ]; then
-        synoacltool -add "${DIRNAME}" "group:${GROUP}:allow:rwxpdDaARWcC-:fd--" >> ${INST_LOG} 2>&1
-    fi
-
-    # Walk up the tree and set traverse permissions up to VOLUME
     VOLUME=$(echo "$DIRNAME" | awk -F/ '{print "/"$2}')
-    while [ "${DIRNAME}" != "${VOLUME}" ]; do
-        # Set execute permissions for GROUP
-        if [ ! "`synoacltool -get "${DIRNAME}"| grep "group:${GROUP}:allow:..x"`" ]; then
-            synoacltool -add "${DIRNAME}" "group:${GROUP}:allow:--x----------:---n" >> ${INST_LOG} 2>&1
+
+    # Ensure directory resides in /volumeX before setting GROUP permissions
+    if [ "`echo $VOLUME | cut -c2-7`" = "volume" ]; then
+        echo "Granting '$GROUP' group permissions on $DIRNAME" >> ${INST_LOG}
+
+        # Set read/write permissions for GROUP
+        if [ ! "`synoacltool -get "${DIRNAME}"| grep "group:${GROUP}:allow:rwxpdDaARWcC-:fd--"`" ]; then
+            synoacltool -add "${DIRNAME}" "group:${GROUP}:allow:rwxpdDaARWcC-:fd--" >> ${INST_LOG} 2>&1
         fi
-        DIRNAME="$(dirname "${DIRNAME}")"
-    done
+
+        # Walk up the tree and set traverse permissions up to VOLUME
+        while [ "${DIRNAME}" != "${VOLUME}" ]; do
+            # Set execute permissions for GROUP
+            if [ ! "`synoacltool -get "${DIRNAME}"| grep "group:${GROUP}:allow:..x"`" ]; then
+                synoacltool -add "${DIRNAME}" "group:${GROUP}:allow:--x----------:---n" >> ${INST_LOG} 2>&1
+            fi
+            DIRNAME="$(dirname "${DIRNAME}")"
+        done
+    else
+        echo "Skip granting '$GROUP' group permissions on $DIRNAME as the directory does not reside in '/volumeX'. Set manually if needed." >> ${INST_LOG}
+    fi
 }
 
 ### Generic package behaviors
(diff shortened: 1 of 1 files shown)


Safihre:
Nice.
With the testing of the SABnzbd package by a number of users I ran into the reverse problem: subfolders. 
For example when a user updates while having a queue, after the update SAB can't access the folders of the items in the queue (inside the `incomplete` folder).
Does the `synoacltool` somehow supports recursion into subdirectories? Like `chown -R`? (@ymartin59)
I forsee the same problem for NZBget/Sonarr/Radarr/CouchPotato/Sickbeard..

Safihre:
Should have Googled it!
http://automatica.com.au/2016/11/recursively-propagate-permissions-on-a-synology-nas-with-synoacltool/

What do you guys think about applying this `find` method?

Safihre:
Second question/problem:
A user send me this install-log because he was having permissions issue:
```
Granting 'sc-download' group permissions on /volume1/downloads/incomplete
(synoacltool.c, 140)It's Linux mode
(synoacltool.c, 140)It's Linux mode
(synoacltool.c, 140)It's Linux mode
Granting 'sc-download' group permissions on /volume1/downloads/complete
(synoacltool.c, 140)It's Linux mode
(synoacltool.c, 140)It's Linux mode
(synoacltool.c, 140)It's Linux mode
```
I have been testing and `synacltool` works as long as at least 1 parent folder has some ACL, but in this case it seems the whole tree doesn't. Using Google I found others here already found that out 1.5 years ago, it happens after upgrades from DSM5:
https://github.com/SynoCommunity/spksrc/issues/1932#issuecomment-209837411

I am not sure what to do about this? There seems to be no `-force` or something..

m4tt075:
@Safihre Yikes! Not sure either, but I'm starting to understand why @Dr-Bean has not been a huge fan of syno-tools...  :smirk:  
The find method could probably be implemented, but no idea how to solve your second problem. Following your link and links therein, apparently @Dr-Bean was aware of the problem when he created his DSM6 POC PR which contained a `syno_permission` function very similar to the one in @ymartin59 's generic installer script, so I'm surprised it does not work. But I also don't have a test environment for a DSM6 system which has been upgraded from DSM5. 

Maybe we just cannot cover every use case. Maybe, we have to accept a certain amount of user permission breakage for shared folders on upgrades. Worst case, the users have to set the permissions manually via the DSM interface. We could introduce a warning message for upgrades of potentially affected packages (those with share folders) via the wizard. Not ideal, just a fallback, but might be worthwhile in any case...

Safihre:
I would suggest (including your changes):
```shell
set_syno_permissions ()
{
    DIRNAME=`realpath $1`
    GROUP=$2

    VOLUME=$(echo "$DIRNAME" | awk -F/ '{print "/"$2}')

    # Ensure directory resides in /volumeX before setting GROUP permissions
    if [ "`echo $VOLUME | cut -c2-7`" = "volume" ]; then
        echo "Granting '$GROUP' group permissions on $DIRNAME" >> ${INST_LOG}

        # Set read/write permissions for GROUP
        if [ ! "`synoacltool -get "${DIRNAME}"| grep "group:${GROUP}:allow:rwxpdDaARWcC-:fd--"`" ]; then
            synoacltool -add "${DIRNAME}" "group:${GROUP}:allow:rwxpdDaARWcC-:fd--" >> ${INST_LOG} 2>&1

            # Copy permissions to sub-directories
            find ${DIRNAME} -exec synoacltool -enforce-inherit {} \; >> ${INST_LOG} 2>&1
        fi

        # Walk up the tree and set traverse permissions up to VOLUME
        while [ "${DIRNAME}" != "${VOLUME}" ]; do
            # Set execute permissions for GROUP
            if [ ! "`synoacltool -get "${DIRNAME}"| grep "group:${GROUP}:allow:..x"`" ]; then
                synoacltool -add "${DIRNAME}" "group:${GROUP}:allow:--x----------:---n" >> ${INST_LOG} 2>&1
            fi
            DIRNAME="$(dirname "${DIRNAME}")"
        done
    else
        echo "Skip granting '$GROUP' group permissions on $DIRNAME as the directory does not reside in '/volumeX'. Set manually if needed." >> ${INST_LOG}
    fi
}
```

ymartin59:
As I wrote elsewhere, it would be interesting to add test cases in `demoservice` to validate expected permissions set on volume shares.

m4tt075:
@ymartin59 Not sure this would be helpful. It is not the lack of a package applying this function that is the problem. I have been checking and validating user permissions on my `XPE 5.2` and `XPE 6.1` test systems for several hours yesterday. It is just that I cannot emulate a "DSM 5 -> 6 upgrade system" environment. The rest is just a design decision: Do we want to amend the `set_syno_permission` function or not? And if so, how? 
- `realpath`: Yes or no?
- Ensuring the path starts with `/volume`: Yes or no?
- Transferring ACL permissions to sub-folders (@Safihre proposal): Yes or no?
Any advice appreciated...

ymartin59:
This method first version come from DrBean installer script, amended by BenjV... and I agree with your improvements.
If `synoacltool` is able to process other locations than `/volumeXXX`, depending on underlying file system type, I think it may be worth to remove that restriction - I propose to give it a try, and wait if anyone reports trouble with errors in installation log.
I found multiple references about "ACL permission propagation" like http://automatica.com.au/2016/11/recursively-propagate-permissions-on-a-synology-nas-with-synoacltool/

As a result, I propose to give a try to this following combination:
- With `realpath`
- Without `/volumeXXX` restriction
- With ACP permission propagation

m4tt075:
I have update this PR, adding @Safihre 's proposal plus additional quotes to allow better handling of directories with spaces and stuff. I have kept the volume check for the time being though, as I consider it a potential security risk. Please see #3050 for another attempt to explain what concerns me. I'd prefer to keep it here, because it could affect more packages than just TVH. If you still think it is not needed, I will remove it and add it to the parser in #3050 instead. Just want to be 100% sure...

m4tt075:
@ymartin59 Surprised to see this merged without further comment. Hope it wasn't "by accident"...  :-)
Do you agree with keeping the volume check in the service script now?

ymartin59:
Yes, OK for me... by the way, we have to go further in testing...

m4tt075:
I agree, and it's better to test deeper now before we scale up...

ymartin59:
I have many troubles with that `set_syno_permissions` code... I have to fix before publishing any package using it.

Safihre:
Tell us more? What's the cases that go wrong?

PR #3085, opened 2018-01-08
Add hi3535 to ARMv7 archs

Safihre:
As described in #3080 the `hi3535` arch couldn't compile Python because of a problem with `libgpg-error`. 
Turns out all that was needed was to add it to the ARMv7 arches. Now Python builds for `hi3535` [without a problem](https://travis-ci.org/Safihre/spksrc/builds/326366000).

_Not a Python-specific issue, so separate PR._


m4tt075:
Amazing! I have just tried your fix with `tvheadend` out of curiosity. It compiles. Which means that `ffmpeg` and a ton of other cross packages compile as well. Not able to test anything of course. Still, very nice!

ymartin59:
I really wonder what submitted `hi3535` support forgetting that point. By the way I will not waste time to answer it

Issue #3031, opened 2017-12-05
ADD [Debian Chroot] IN PACKAGES FOR DS118


suil1128:
ADD [Debian Chroot] IN PACKAGES FOR DS118, THANKS VERY MUCH!

### Setup
_Package Name:_
_Package Version:_

_NAS Model:_
_NAS Architecture:_
_DSM version:_

### Expected behavior
Tell us what should happen

### Actual behavior
Tell us what happens instead

### Steps to reproduce
_1._
_2._
_3._

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
```
Insert the package log here
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here
```

ymartin59:
Complex and dangerous package... Not sure it will ever come back

christophe-lyon:
I did manage to build it on my side (for DS218play), but couldn't install it because DSM says there's a dependency on python, which I wasn't able to build so far because pip complains:
pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
(Which is probably a problem on my side)

@ymartin59 what do you mean by dangerous?

ymartin59:
See #2758 - according to many issues registered around debian-chroot, there is risk you face an unstable DSM, and uninstall package lead to corrupt DSM up to a full re-install is required

Issue #2886, opened 2017-08-23
Add support for rtd1296


bidorffOL:
Because of the new rtd1296 architecture supported by the 418J, I cannot add the SynoCommunity repository to my NAS. From what I have read in the [FAQ](https://github.com/SynoCommunity/spksrc/wiki/Frequently-Asked-Questions#i-have-a-new-nas-model-what-has-to-be-done-to-provide-packages-for-it), one of the prerequisite for such support is the availability of the toolchain for this architecture (which has been added yesterday by Synology [here](https://sourceforge.net/projects/dsgpl/files/DSM%206.1%20Tool%20Chains/Intel%20x86%20Linux%204.4.15%20%28Rtd1296%29/)).

Is there any hope to see this new architecture supported in a near future ?

davebooi:
I second this and posted a 30 bucks bounty on [bountysource](https://www.bountysource.com/issues/48574365-add-support-for-rtd1296)

Diaoul:
Off topic: This is really annoying that Bountysource doesn't update the issue... It suddenly stopped working, I'll get in touch with them.

ymartin59:
Adding toolchain in spksrc is a five minute job and it should be enough for anyone to build its own packages... but job to get any DSM 6.1 package available through Package Center is still long: https://github.com/SynoCommunity/spksrc/wiki/Frequently-Asked-Questions#which-dsm-versions-do-you-support

bidorffOL:
I actually tried to do it myself on my own [fork](https://github.com/bidorffOL/spksrc/) and I managed to get NZBGet and GIT to build and run, still no luck on Python, but I'm not sure if the toolchain is responsible for the broken build.

I'll try to provide a Pull Request for my changes (I will need to filter those for the toolchains and those for the packages I updated and / or modified).

albino71:
@bidorffOL Got me a DS418J..great device, but the nzbget package is not working and cannot be installed...

Any chance you could slip me your installable package?

bidorffOL:
You can try [this version](https://github.com/bidorffOL/syno-packages/raw/master/nzbget_rtd1296-6.1_19.1-23.spk) (tell me if it works, I'm not sure that this is the file I actually used).

bidorffOL:
That's what I was afraid of...

[...]

PR #3088, opened 2018-01-09
Apollolake update

hgy59:
_Motivation:_ I have a DS918+ and with the initial spksrc copy 31 of 110 packages failed to build.  Apart of apollolake some fixes for hi3535 (armv7) and rtd1296 (aarch64) are included in this PR.
_Linked issues:_ partial fixes for #3031, #2886

### Packages with newly fixed build for apollolake arch
- [x] debian-chroot
- [x] gentoo-chroot
- [x] lirc

other packages will follow...


Safihre:
👍 
Seems in these packages it's better to use the `ARM7_ARCHES`/`ARM8_ARCHES`/etc: https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.common.mk#L36-L38
Should probably have been done long ago already!

Safihre:
@hgy59 could you try using the general variables instead of the specific archs?
For this you might need to move the `include xxx.mk` above the if statements. Otherwise the variables are not known.
Had to do similar for `syncthing`: https://github.com/SynoCommunity/spksrc/pull/3094/files

hgy59:
Replaced by PR #3099

PR #3091, opened 2018-01-10
NZBGet general package - General Service approach

Safihre:
For the NZBGet package I propose something a bit different: 
Why would we try to be smarter then NZBGet's creator @hugbug?
Instead, let the SynoCommunity package just download the latest installer and make sure that the permissions are set correctly. This allows users to update NZBGet from within NZBGet itself. No more need for us to update the package 🎉.
The NZBGet installer also tries to determine optimal performance settings during the installation, which can make a _substantial_ difference in performance on NAS devices. I now changed the package so that these settings are used (for new installs).
For backwards compatibility for users that upgrade from older package I had to still keep some Synology-specific paths and options.

**Note:** I was not able to test any platforms myself besides my XPEnology or test the upgrade process from older-packages, since the current package doesn't work.. I propose user-testing for this. 
Looking at the NZBGet forums it seems @hugbug's installer can handle all platforms. If not, we can always redirect users to him and a general fix can be implemented in NZBGet itself.

**Note 2:** @ymartin59 @m4tt075 Also on this package I think it's important that we first implement the sub-directory-permissions for `set_syno_permissions` [discussed here](https://github.com/SynoCommunity/spksrc/pull/3084#issuecomment-355948494), otherwise upgraded packages will still not work well since they can't access files made by the old package.

### Checklist


ymartin59:
Sorry to disappoint you. Such an idea is interested for non-arch application like full python, perl or java applications... In that case, I recommend package is named `-installer` and its version number is not related at all to application version itself.

By the way, this "installer" concept will not work with `nzbget` as mainly written in C and dowload location you use only retrieve a `x86_64` binary. What about other architectures like ARM and PPC ?

So again, sorry, but it is really required to rebuilt package from sources to target architecture, and also DSM version (5 or 6) when C++ is involved.

Safihre:
The nzbget installer is not just an `x86` installer, it has inside the binaries for all platforms and will install the detected one. Even PowerPC is supported and of course all ARM versions. 
See here: https://github.com/nzbget/nzbget/blob/develop/linux/installer.sh#L91
https://nzbget.net/installation-on-linux

Safihre:
@ymartin59 I hope you'll reconsider your stance on this since I really think this is the best solution:
- **All archs are supported** by the NZBGet installer: `x86`, `x64`, `armv5/6/7/8`, even `ppc8xx`/`qoriq` (these are `e500` cores).
- All it requires is Linux 2.6, no other dependencies. All the toolchains, both DSM5 and 6, are Linux 2.6+ so they are supported.
- The optimized settings that the installer determines specific to platform and compute-power can make a huge difference in performance of NZBGet.
- The users can have in-app updates, meaning we will never have to update the package again since `https://nzbget.net/download/nzbget-latest-bin-linux.run` will always point to the latest stable release.
- The same way we can change `nzbget-testing` to point to the latest `https://nzbget.net/download/nzbget-latest-testing-bin-linux.run` and never have to update the package ever again.

The NZBGet package hasn't been updated in 2 years... That's a really long time. So online you can find many references to Synology users updating their existing NZBGet installation in exactly this way, by downloading the installer and just running it.

So why not use it? 🥇

ymartin59:
Here is my proposal:
- name your new package as `nzbget-installer`
- provide clone URL as wizard field, allowing user to install default "stable" or switch to "testing" binaries

Then we may publish a "final" update for `nzbget` and `nzbget-testing` to invite (either explain or automate) user to migrate to this new package installer.

What do you think about it ?

Safihre:
Yes, that would be possible. But.. why? 
There are many self-updating packages that don't have a special suffix, like Sonarr or all the `git clone` packages.  For users it's not really convenient to have to switch: they would have to manually transfer settings between current and new `installer` package and have to do this before the upgrade (since can't have 2 instances running on same port). Or we have to write a transfer-script. It would also result in having 3 `nzbget` packages on the repo for a while.

We could release it as a `BETA=1` package if you fear it won't work well for all platforms?

ymartin59:
@Safihre You're right... I just dislike such auto update packages, I consider them too difficult to diagnose if anything gets wrong.
I agree not to rename package and keep `nzbget` as far as either description and/or a wizard step inform users about this packaging change and new versioning timestamp, for example in wizard step allowing to change download URL (to switch to testing - you can look at sickbeard-custom for instance)
OK to set `BETA=1` flag too.

Safihre:
Pushed the updates.
The following screen will now be shown during install and upgrade.
Since there is really only 1 source of NZBGet with 2 options, a simple choice seems most clear and sufficient to me 👍 

image

ymartin59:
@Safihre French translation for <<The installer will now download the desired version of NZBGet from NZBget.net and install it. <br>In the future you should update from within NZBGet (Config > System).>>:

<<L'installateur va maintenant télécharger la version voulu de NZBGet depuis NZBget.net et l'installer.<br>Vous pouvez ensuite mettre à jour depuis NZBGet (Config > System).>>

brassica77:
Small typo here : it should be "la version voulu**e**"

Safihre:
@ymartin59 don't want to add translations anymore?

spk/nzbget/src/service-setup.sh
```diff
@@ -0,0 +1,84 @@
...
+# Force-overwrite the PID-file and WebDir setting
+# These could change depending on previous package settings
+SERVICE_COMMAND="${NZBGET} -c ${CFG_FILE} -o WebDir=${WEBDIR} -o LockFile=${PID_FILE} -D"
+
+# Needed to force correct permissions, during update
+# Extract the right paths from config file
+if [ -r "${UPGRADE_CFG_FILE}" ]; then
+    DOWNLOAD_FOLDER=`grep -Po '(?<=MainDir=).*' ${UPGRADE_CFG_FILE}`
+    if [ -n "$(dirname "${DOWNLOAD_FOLDER}")" ]; then
+        SHARE_PATH=$(dirname "${DOWNLOAD_FOLDER}")
```
ymartin59: This trick has a side effect that may be an issue later. `SHARE_PATH` is shorten to `volume1` (for default value) and is stored that way in `etc/installer-variable`.
Is there really no `service_xxx` hook function that may handle a specific call to `set_syno_permissions` ?

ymartin59:
@Safihre I have tested NZBGet upgrade path from current published package in DSM 6. Installation is OK but service does not start (whereas it starts properly after fresh install)
Probably related to `SHARE_PATH=/volume1` as file permissions seems to be broken on my xpenology system:
```
bash-4.3# ls -l /var/packages/nzbget/target/
total 0
d--------- 1 sc-nzbget nzbget  24 Jan 20 21:27 app
d--------- 1 sc-nzbget nzbget 368 Jan 20 21:47 bin
d--------- 1 sc-nzbget nzbget  10 Jan 20 21:27 lib
d--------- 1 sc-nzbget nzbget  12 Jan 20 21:27 share
d-----S--- 1 sc-nzbget nzbget  78 Jan 20 21:46 var
bash-4.3# ls -l /volume1/
total 0
d--------- 1 root  root   52 Jan 20 21:46 @appstore
d--------- 1 admin users  44 Feb  5  2017 @database
d--------- 1 root  users  34 Jan 20 21:34 downloads
d--------- 1 root  root   18 Nov  5 17:57 @eaDir
d--------- 1 root  root   24 Feb  5  2017 @S2S
d--------T 1 root  root  382 Jan 21 08:07 @tmp
```
May you please investigate and propose a new PR with fix ? Thank you in advance

Safihre:
Hmm, could you maybe explain bit more what is going wrong? I don't fully understand?
The `dirname` is the problem? Wonder why I added that in first place.. 🤔

ymartin59:
@Safihre There is chance `dirname` is the problem... Let's try without it

Safihre:
Ok! Now that I think of it, I think I did this previously because I only wanted to apply these group permissions to 1 folder "up" and apply the user permissions to the folder itself. But with the changes to Generic Service, that won't be needed anymore. 

Sabnzbd PR also has this code, I have to change it there to and test again.

PR #3093, opened 2018-01-11
SiCKRAGE - Generic Service approach

Safihre:
### Checklist


Safihre:
For backwards compatibility I guess these (`sonarr/sickrage/couchpotato`) should all be `sc-media`.. 
Still kinda strange, since they clearly couple to `sc-download` packages.
I will change this.

ymartin59:
I also wonder what are expected usage of these two distinct groups `sc-media` and `sc-download`. A quick usage inventory in packages should help to guess a policy and then fix those that do not match.

ymartin59:
I have to submit a french translation too

Safihre:
It seems `sc-media` was assigned to packages doing something with media files, regardless if they came from download providers.
It's too late to change now, because if previous version of any of these packages (Sonarr/CouchPotato/Sickrage/etc) was already installed, switching to `sc-download` would not allow these packages to access the files they manages before anymore. Bit annoying and not so user friendly situation, both cases actually..

Safihre:
Added `sickbeard-custom` to this one. It's 98% the identical code.
Minus the `requirements` installation of SickRage that is not needed for SickBeard.

Issue #3027, opened 2017-12-03
using upstart scripts in packages


kaso17:
Hi,

Since DSM 5 it seems like upstart is used as init system. By default the syno pkg installer  generates a corresponding upstart config file /etc/init/pkgctl-$PACKAGE.conf.
Someone knows how the package can supply it's own upstart config (instead of the auto generated one) so we can make use of upstart features like automatic respwan?

https://neerajcodes.wordpress.com/2017/03/08/creating-a-package-for-synology-nas-device/ mentions that it can be done via the conf directory but no details on how to use it.
The official documentation doesn't mention this at all.
https://developer.synology.com/developer-guide/synology_package/conf.html

ymartin59:
I am curious to know why you are interested in writing your own upstart service configuration when Developer Guide explains how to provide a `start-stop-status` script: https://developer.synology.com/developer-guide/synology_package/scripts.html

For DSM 5 and 6 support, a new generic script will be soon available for package design:
https://github.com/SynoCommunity/spksrc/pull/2949/files#diff-9fe71615984368bf44c02ed52c3b1fce
Here is our reference documentation about it: https://github.com/SynoCommunity/spksrc/wiki/Service-Support#service-startup

So what are your needs requiring a specific upstart script for your package ?

kaso17:
The primary reason is the respawn feature of upstart (automatic restart of the process) and the simple logging.
I would like to update the jackett package init scripts, jackett uses it's own update mechanism (due to the very high release/update rate). The preferred way on linux is to use systemd/upstart to start it with automatic restarting enabled. Then the updater just replaces the files and kills the old process, systemd/upstart takes care of restarting it.
Without the init system restarting jackett the updater takes care of it (fork and restart the original command again).
With the updater instead of the init system restarting jackett the problems begins. The PID file doesn't get updated (not supported by jackett) breaking the stop/status feature (this can be fixed by using `pidof` or `ps | grep` to get the pid based on the program name. Besides that logging is a problem. logging is currently handled by the start-stop-status redirecting stdout/stderr. The jackett process/updater isn't aware of this. Due to this after an update logging is broken.

Besides that I'm not sure how logrotate is handled (is it handled at all?). You can're send a signal to reopen the stdout/stderr redirect so you've to use a copy/truncate solution which is bad for performance and you can loose log entries.

With a direct upstart script instead of the ductape solution via a traditional start-stop-status wrapper script all of these problems would be gone.

[...]

PR #3094, opened 2018-01-11
Syncthing - Update 0.14.44 - Generic Service approach

Safihre:
### Checklist


m4tt075:
@Safihre @ymartin59 I have just realized that `Syncthing` is a package which used a `start-stop-daemon` prior to the conversion. If I'm not mistaken (and @Diaoul might be able to comment here) `start-stop-daemon` was used for _all_ services, which cannot be toggled to run as daemon themselves. Do you think it makes sense to keep this? And if so, would you implement this on package or `spksrc.service` script level?

Safihre:
Ah, I thought you did the same for tvheaded? With the special `service_prestart`?
https://github.com/SynoCommunity/spksrc/wiki/Service-Support#run-service-process-in-background

But that's indeed not really deamon, for example when the service restarts itself it will be shown as Stopped in the Package Center.

Safihre:
I kinda just improvised, maybe the real `start-stop-deamon` is better idea!

m4tt075:
`Tvheadend` has a daemon option (`-f`), so I did not need it. The special `service_prestart` vodoo was needed to workaround permission issues. I only encountered the `start-stop-daemon` stuff when I started looking into `Syncthing`. I `grepped` the `dsm-control` files within `spk` and realized quite a lot of packages utilize this approach, namely:
- deluge
- fengoffice
- homeassistant
- horde
- icecast
- museek-plus
- owncloud
- radarr
- rutorrent
- selfoss
- shairport-sync
- sonarr
- snycthing-inotify
- syncthing
- tt-rss

This is why I was asking...

Safihre:
I found that `sonarr` and `radarr` actually do have a PID file option, but just place it in a custom location.

But indeed for other ones, maybe a general service provider would be great!

ymartin59:
If application produces `PID_FILE` and command is "fixed", then Makefile `SERVICE_COMMAND` is enough.
If command has to be built, or `PID_FILE` has to be created thanks shell... then `service_prestart` is the option I proposed.
In that case, sourcing `options.conf` in `service_prestart`

ymartin59:
In fact `start-stop-daemon` comes from `cross/busybox`... and as far as it can be replaced by shell background execution and echo PID in a file, I prefer that later option as it avoid to make package arch.

ymartin59:
Please copy wizard french translations from `spk/demoservice`

Diaoul:
It is also possible to patch the app so it produces a pidfile as we need.
This is to be preferred if easy to do. There were benefits using the daemon
thing over a simple shell but I can't remember exactly. Go with whatever
that works, just make sure to test special cases like when the app restart
itself.

Le 13 janv. 2018 10:36 PM, "Yves Martin" <notifications@github.com> a
écrit :

Please copy wizard french translations from spk/demoservice

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/SynoCommunity/spksrc/pull/3094#issuecomment-357469135>,
or mute the thread

ymartin59:
To solve automatic restart after update for jackett, someone asks how to integrated into DSM `upstart` system: #3027. It may be an option but I fear about documentation and support in case of troubles... What do you think about it ?

Safihre:
@ymartin59 Discussion in #3027 is interesting. It seems you are also right that the `.conf` scripts don't survive an update of DSM. 
So doesn't seem a viable solution, with the update-rate of DSM 😢 

http://majikshoe.blogspot.nl/2014/12/starting-service-on-synology-dsm-5.html
image

Safihre:
While this would benefit from #3123, it actually needs `start-stop-deamon` as discussed before.
@ymartin59 Shall I manually include that in this package or will you add that to #3123 also?

ymartin59:
My idea is to create a clean / alternate `start-stop-status` script in generic support (there are already two of them https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.mk#L120) with a `Makefile` switch to declare package requires `start-stop-daemon`

Safihre:
Sounds good! :+1:

Safihre:
@cytec Can you help me explain `native/go` has this special rules for `arm`? Why does it need Go-1.4?
And do we need to also add this for `arm64`?
https://github.com/SynoCommunity/spksrc/blob/master/native/go/Makefile#L26

spk/syncthing/src/service-setup.sh
```diff
@@ -0,0 +1,29 @@
...
+CONFIG_DIR="${SYNOPKG_PKGDEST}/var"
+SYNCTHING_OPTIONS="-home=${CONFIG_DIR}"
+
+# Read additional startup options from /usr/local/syncthing/var/options.conf
+if [ -f ${CONFIG_DIR}/options.conf ]; then
+    source ${CONFIG_DIR}/options.conf
+fi
+
+# Add special options to start-stop-daemon command
+SERVICE_OPTIONS = SYNCTHING_OPTIONS
```
ymartin59: This should not work. I propose to rename `SYNCTHING_OPTIONS` into `SERVICE_OPTIONS`

Safihre: Indeed it doesn't work, in my test-branch I already updated that but just didn't push it yet.
But the name needs to stay `SYNCTHING_OPTIONS` because users could have setup some custom options during previous versions via `${CONFIG_DIR}/options.conf`. 
That `options.conf` is kept during the upgrade, so for backwards-compatibility it needs this.

```sh
#!/bin/sh
# This file will be added to start-stop-status script
# For possible options see syncthing --help

# Example: uncomment this to start syncthing with all devices paused
# SYNCTHING_OPTIONS="${SYNCTHING_OPTIONS} -paused"
```

ymartin59: OK. Have not read inside `options.conf`. So I wait for your review request.

cytec:
@Safihre i try my best: go 1.4 is needed because if you install from source and want to build the toolchains you need a go compiler installed (go 1.4 was the last version with a compiler written in c) so you need to bootstrap it with go 1.4

The rules for `arm` and `i386` are needed to get valid go compilers for those plafroms too... basically what you have to do is: export `GOARCH` and run `./make.bash` for each arch you want to use... `--no-clean` just keeps everything and builds the new stuff on top instead of clearing all previous builds.

More info about all that can be found on the [install go from source site](https://golang.org/doc/install/source)

So if you want to enable `ppc64` you'll have to add the same line with `ppc64` same goes for `arm64` and every other ARCH go is able to compile to (see https://golang.org/doc/install/source#environment)

ymartin59:
Really instructive... Sounds like wiki would be the right place for this information.

ymartin59:
By the way, `qoric` PPC last architecture cannot be deprecated as it still has DSM 6.2 support.

Safihre:
Indeed good info, I will add this information as comments to `native/go` otherwise this might get lost! 
 I have to correctly incorporate this and also test this on my DS112J, how the latest Go updates affect this.
Will report back when I do.

ymartin59:
@Safihre In https://github.com/SynoCommunity/spksrc/pull/3094#discussion_r171697661 you told about changed not pushed yet. What is this PR status ?

Safihre:
Waiting for your changes to https://github.com/SynoCommunity/spksrc/pull/3186, you wrote there that you would propose changes: https://github.com/SynoCommunity/spksrc/

ymartin59:
@Safihre OK. I have updated PR. So you can remove `STARTABLE=yes` but still requires to setup `SERVICE_EXE` in `Makefile`, it will be copied at the beginning of generated `scripts/service-setup`

Safihre:
How do you feel about what I did for Syncthing where I add `sc-syncthing` to `users` (if it was before part of `users`)?
Without that I had some feedback of users that Syncthing can't really access any existing files and even applying the right permissions doesn't fix things. Like Transmission..

ymartin59:
@Safihre I agree with you. Any service account expected to read files should be in `users` account.
When producing content in "restricted" folders (not accessible to all users), I have found culprit when trying to access from another user account shell... More "read" permissions are required to browse folders from libc - even if previous settings was enough from File Station.

ymartin59:
@Safihre Here is a "preview" PR just for you to review/test... to be honest I have done "no testing", probably there are typo preventing script to run... so to use with care...

ymartin59:
@Safihre Have you got any local pending commits/updates concerning Syncthing?

Safihre:
Aah yes, all permissions and deamon stuff is merged now so I can finish this.

Safihre:
@ymartin59 I need help 😨 
I adapted the files to be like this:
[Makefile](https://github.com/Safihre/spksrc/blob/travis-branch/spk/syncthing/Makefile)
[service-setup.sh](https://github.com/Safihre/spksrc/blob/travis-branch/spk/syncthing/src/service-setup.sh)

Which then creates the `service-setup`-file below in the package.
But it doesn't work, probably because `SERVICE_EXE` is defined before `SYNCTHING` variables etc.
How do I fix this?

```shell
### Package specific variables and functions
# Base service USER to run background process prefixed according to DSM
USER="syncthing"
PRIV_PREFIX=sc-
SYNOUSER_PREFIX=svc-
if [ ${SYNOPKG_DSM_VERSION_MAJOR} -lt 6 ]; then EFF_USER="${SYNOUSER_PREFIX}${USER}"; else EFF_USER="${PRIV_PREFIX}${USER}"; fi
# Service port
SERVICE_PORT="7070"
# start-stop-status script redirect stdout/stderr to LOG_FILE
LOG_FILE="${SYNOPKG_PKGDEST}/var/${SYNOPKG_PKGNAME}.log"
# Service command has to deliver its pid into PID_FILE
PID_FILE="${SYNOPKG_PKGDEST}/var/${SYNOPKG_PKGNAME}.pid"
# Service command to execute with start-stop-daemon
SERVICE_EXE="env HOME=${CONFIG_DIR} ${SYNCTHING}"
SERVICE_OPTIONS="${SYNCTHING_OPTIONS}"

# Invoke shell function if available
call_func ()
{
    FUNC=$1
    if type "$FUNC" | grep -q 'function' 2>/dev/null; then
        echo "Invoke $FUNC" >> ${INST_LOG}
        eval ${FUNC}
    fi
}

# Setup environment
PATH="${SYNOPKG_PKGDEST}/bin:${PATH}"
SYNCTHING="${SYNOPKG_PKGDEST}/bin/syncthing"
CONFIG_DIR="${SYNOPKG_PKGDEST}/var"
SYNCTHING_OPTIONS="-home=${CONFIG_DIR}"

# Read additional startup options from /usr/local/syncthing/var/options.conf
if [ -f ${CONFIG_DIR}/options.conf ]; then
    source ${CONFIG_DIR}/options.conf
fi

LEGACY_GROUP="users"

service_postinst ()
{
    # Add also to "users" group in case it was there
    # This way it keeps any permissions it used to have
    syno_user_add_to_legacy_group "${EFF_USER}" "${USER}" "${LEGACY_GROUP}"

    # Discard legacy obsolete busybox user account
    BIN=${SYNOPKG_PKGDEST}/bin
    $BIN/busybox --install $BIN >> ${INST_LOG}
    $BIN/delgroup "${USER}" "users" >> ${INST_LOG}
    $BIN/deluser "${USER}" >> ${INST_LOG}
}
```

ymartin59:
@Safihre Please push your recent changes, it will be easier for me to understand.
I propose you set everything once from spk `Makefile`:
```
SERVICE_EXE="env HOME=$${SYNOPKG_PKGDEST}/var $${SYNOPKG_PKGDEST}/bin/syncthing"
SERVICE_OPTIONS="-home=$${SYNOPKG_PKGDEST}/var"
```
Cheers

Safihre:
I don't think this is ready yet, never got it to work properly. That's why this was the only one I didn't put it in the checklist at the end of  #3138.
Will try to do some work on it today to finish it!

Safihre:
@ymartin59 I pushed some updates, now the command is successfully passed by using a dummy that I then later overwrite in `service-setup.sh`
But it still won't start and I'm not sure why. Maybe you can take a look? 
```
admin@DiskStation:/$ cat /volume1/@appstore/syncthing/var/syncthing.log                                                 
Sat Jun  2 12:39:32 CEST 2018
Starting syncthing command env HOME=/volume1/@appstore/syncthing/var /volume1/@appstore/syncthing/bin/syncthing -home=/volume1/@appstore/syncthing/var
```

ymartin59:
@Safihre From my point of view, there is no need for "dummy" as command and options are not dynamic. Setting these in `Makefile` instead of dummy should be exactly equivalent:

```
SERVICE_EXE = "env HOME=$${SYNOPKG_PKGDEST}/var $${SYNOPKG_PKGDEST}/bin/syncthing"
SERVICE_OPTIONS = "-home=$${SYNOPKG_PKGDEST}/var"
```

I will have to grab your branch to diagnose...

ymartin59:
@Safihre I am in trouble with `start-stop-daemon` support... it works properly from command line but not from Package Center. Still investigating...

ymartin59:
Superseed by #3391

PR #3097, opened 2018-01-12
 NzbHydra - Generic Service approach

adchevrier:
First draft of the conversion of the nzbhydra package to the new architecture.
Not at home so I couldn't test the build/install but I will later.

Not sure about the user group to use, since nzbhydra doesn't need to access shares so I went with sc-media.

### Checklist


Safihre:
Looks good. You can also just leave out the `GROUP `, since it doesn't need any access.

Only comment I would have is that `LOG_FILE` variable is already supplied by the generic service approach. So don't define your own one, the variable is already available.

ymartin59:
Please take opportunity to update package to latest version

Safihre:
I see I should have looked at this PR a but better, indeed essential stuff missing (like the update of `Makefile`). Sorry! 
But nzbhydra does use `git` to update internally, just like `couchpotato` and others. So it should follow the same style, with a `git clone` during the `service_postinst`.

adchevrier:
My bad, I did forget to include the updated Makefile (`git add .` in the wrong folder), fixed now.
As Safihre stated, NzbHydra has a built-in updater using git, that's why the dependency is needed.

The commit I checkout is the commit coresponding to the latest release (https://github.com/theotherp/nzbhydra/commit/e56c6364ea0d65a13c240d8b2cfd4c3eca5b78f7). As it is not the latest commit, clone option `--depth=1` won't work, that's why I need this `git clone/checkout` sequence (I'm open to suggestions on this).

adchevrier:
Just noticed I lost my nzbhydra icon in DSM launcher, did I miss something?

Safihre:
@adchevrier Where should the icon appear? I indeed don't have an icon for any of the updated apps.
The Synology dev guide says it needs icons in `/usr/syno/synoman/webman/3rdparty` and a `app/config` file, which all my packages have..
@ymartin59 do you maybe understand?

ymartin59:
@Safihre If `SPK_ICON` and `SERVICE_PORT` is present, then `app/config` is generated and `DSM_UI_DIR` is automatically set.
References:
- https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.spk.mk#L90
- https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.mk#L186
- https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.icon.mk#L42

So please check again:
- is ImageMagick `convert` tool available on your system ?
- please review `make` output for warning/error lines and submit relevant excerpts for analysis

Safihre:
@ymartin59 Yes, the files are all there in the `.spk` archive, the `app` folder with the `config` file and the `images` folder with the correct images.
So what I meant is that even though the `.spk` for each package has all the files, I don't see the icons for any of the apps in the app-drawer (the icon in the top-left?). 
When I install an older package directly from the repo, it also doesn't show an icon. But then again, those repo packages don't start on my XPenology and maybe the icons are only shown if a package is running?

ymartin59:
OK so let's get it start properly. What is in log files in package `var` ?

Safihre:
@ymartin59 @adchevrier I double checked this requested, build it and tested it on DSM 5 and 6.
It works perfectly.
Since apparently I am allowed to push to your branch @adchevrier I used that to update it to the latest version!
I suggest a merge and a publish!

adchevrier:
Thanks!

PR #3099, opened 2018-01-14
add apollolake kernel for lirc (plus other 6.1 kernels)

hgy59:
_Motivation:_ I have a DS918+ and with the initial spksrc copy 31 of 110 packages failed to build.  Apart of apollolake some fixes for hi3535 (armv7) and rtd1296 (aarch64) are included in this PR.
_Linked issues:_ partial fixes for #3031, #2886

### Packages with newly fixed build for apollolake arch
- [x] lirc

Replacement for PR  #3088 
- regarding the hints of @Safihre 
- unrelated commits removed (libvpx, update kernel source digests, debian-chroot, gentoo-chroot)


spk/debian-chroot/Makefile
```diff
@@ -26,28 +26,37 @@ INSTALL_PREFIX = /usr/local/$(SPK_NAME)
 
 POST_STRIP_TARGET = debian-chroot_extra_install
 
+
+include ../../mk/spksrc.spk.mk
+
+ifeq ($(findstring $(ARCH),88f5281 powerpc ppc824x ppc853x ppc854x dockerx64 grantley kvm64),$(ARCH))
```
Safihre: I am a bit surprised these specific `x64` archs don't work?

@ymartin59 Seems to me all reference to `88f5281` should be removed from the repo, since it doesn't even have a toolchain anymore.

hgy59: just reverted the chroot updates to get rid of this dependency.
The real problem is that ARCH is not defined in this context.

Safihre: Huh? `ARCH` is not defined due to the `include` moving up?

hgy59: No, it has nothing to do with the include. `ARCH` is even not defined with the source of the current master.
The only difference to working packages is, that the `Makefile` located below /spk has no dependency to any cross package.
Perhaps the arch-%: target with indirect call of make is the key (see spksrc.cross-cc.mk line 121, spksrc.spk.mk line 420) as far as I understand the `MAKEFLAGS` construct.

Safihre: Yes the `ARCH` only gets defined when you run any of the `make arch-xxx-6.1` or any of the other commands. Then it will deduct the  `ARCH` from it. 
So when we get to package `Makefile` it will be defined and you can use it, just like you were doing.
But still not sure why those specific `x64` arch's needed to be excluded?

ymartin59:
@hgy59 Would it be possible for you to add same kernel and `lirc` support for architectures that lacks it like recent `rtd1296`  and maybe `grantley`, `deverton` and `dockerx64` if relevant. It would really help, even if you cannot test at runtime.

hgy59:
@ymartin59 Have renamed the downloaded kernel source from linux-4.4.x.txz to apollolake-linux-4.4.x.txz
otherwise there are naming conflicts with different archs on similar branch.
I will create another PR for the remaining 6.1 kernels...

hgy59:
add some 6.1 kernels too
due dependency on fix for mk/spksrc.kernel-modules.mk I did not create a new PR
6.1 kernels added with this PR
rtd1296, monaco, hi3535, grantley, dockerx64, denverton, broadwell, braswell, apollolake, armada370

ymartin59:
Thanks a lot for this effort

spk/lirc/Makefile
```diff
@@ -12,7 +12,7 @@ $(error Sorry, this package does not support the $(ARCH) architecture.)
 endif
 
 # Use older LIRC build for ARCHs with kernel < 2.6.36
-ifneq ($(findstring $(ARCH),alpine armada370 armada375 armada38x armadaxp avoton braswell bromolow cedarview evansport x64 x86),$(ARCH))
+ifeq ($(findstring $(ARCH),88f6281 $(PPC_ARCHES) armada370 armadaxp bromolow cedarview evansport qorig),$(ARCH))
```
ymartin59: Kernel version does not only depend on architecture but also target DSM version... `qoriq` (not `qorig`) is included in `PPC_ARCHES`.
If you agree, I will get your commits and rework some details to get it compile on all architectures and at least on DSM 5.2 and 6.x

hgy59: sorry for the qoriq misstype,...
yes please add the rework for TC version dependencies.
The only arch with kernels that make a difference for lirc in this PR is in armada370 (2.6.32 until DSM 5.2 and now 3.x, I didn't analyze DSM 6.0.2)
Other candidates for different lirc versions are only 88f6281, bromlow, cedarview and evansport. Maybe you make the fix here for armada370 and will adapt it for the archs not in this PR.

hgy59:
finally worked it out.
There are no kernel updates with the given archs that would cause different lirc versions for a specific arch.
Only ppc_arches and 88f6281 must use older lirc version (for all TC versions). Built successfully the lirc 0.8.7 packages - exept arch-powerpc-4.0, due failing kernel build (kernel 2.4.x) and build successfully all-supported.

spk/lirc/Makefile
hgy59: there are no TC version dependecies...
but I made a mistake here:
`ifneq ($(findstring $(ARCH),88f6281 $(PPC_ARCHES)),)`
is not the same as
`ifeq ($(findstring $(ARCH),88f6281 $(PPC_ARCHES)),$(ARCH))`
fixed with commit 528ffe60ef9e568225dafcea0cf631d08edc6360

ymartin59:
Thank you for that great job. I understand why you added `PKG_DIST_FILE` support but I have to carefully check impacts of `spksrc.native-cc.mk` on other tools compilation (mono, go...)

hgy59:
@ymartin59 As you probably know, the `PKG_DIST_FILE` is used to give a downloaded file a different name. I suppose that this should work for all makefiles and thats why I modified `spksrc.native-cc.mk` too, even it is not related to the new kernels. `PKG_DIST_FILE` is already handled in `spksrc.cross-cc.mk` and `spksrc.download.mk` and IMHO it is a bug, that it was not handled in `spksrc.kernel-modules.mk` and `spksrc.native-cc.mk`. 
Until now `PKG_DIST_FILE` is only set in some Makefiles located under /cross/ and with this PR in the Makefiles for 6.1 kernels. Therefore no inpacts can occure with Makefiles located under /native/. 
The fix in `spksrc.native-cc.mk` is just for the case someday a native module needs to have different names for downloaded files, the use of `PKG_DIST_FILE` will work as expected.

Issue #2945, opened 2017-10-05
Add apollolake support to some packages (Sonarr/Transmission/Radarr/etc) for DSX18+


ThomasClemenceau:
Setup

_Package Name: Sonarr/Transmission/Radarr/SABnzbd/Python V2/nano/Jackett/Git Server/Python
_Package Version: 

_NAS Model: DS218+, DS718+, DS918+
_NAS Architecture: Apollo Lake
_DSM version: 6.1

Expected behavior

Support for NAS devices (DS218+ specifically) with apollolake architecture in some packages (Sonarr, Radarr, Jackett and Transmission in my case).

Actual behavior

Sonarr, Radarr, Transmission, SABnzbd, Python v2, Jackett, Git Server, Python and nano are currently not available for apollolake-based NAS devices in SynoCommunity package repo.

Thank you for your time.

fanaticDavid:
I've found the following additional packages to lack support for the Apollo Lake architecture (DS918+ in my case):

- SABnzbd
- Python (v2, which SABnzbd also depends on)
- nano

ThomasClemenceau:
I add them to the list

ThomasClemenceau:
Jacket added to the list

exxie:
Here, Git and python packages from the synocommunity apps. These packages are different than the onse from Synology itself which are named Git Server and Python. These packages from synocommunity are required to make other packages work (such as couchpotato).

ThomasClemenceau:
Added to the list.

ThomasClemenceau:
I see the apollolake DSM 6.1 toolchains is available for download, what can we do with this ?

Ugator:
Syncthing seems to be affected by this too

m4tt075:
Thanks for flaggging all this, but the issues are known. Please take a look at the [FAQs in the Wiki](https://github.com/SynoCommunity/spksrc/wiki/Frequently-Asked-Questions) w.r.t. support of new architectures as well as DSM6.1.

The Apollolake toolchains have been added to the DSM6 development branch recently (see PR #2942). However, the packages you are looking for require more than this (see e.g. issue #2904 and discussions therein). Work is in progress, but will likely take more time and lots of testing. You can track it via the references above. If you can and want to contribute, more than wellcome. Otherwise, please don't expect a quick fix.

ThomasClemenceau:
[...]

PR #3102, opened 2018-01-15
Sonarr and Radarr - Generic Service approach

Safihre:
A small change was needed to `spksrc.service.installer`, as it wouldn't copy "hidden" folders like `.config` used by these packages from the `/var/` folder.
Both packages are 99% identical, so updated them at once.

Compiled and posted in the various issues [for testing](https://github.com/SynoCommunity/spksrc/issues/2945#issuecomment-357008518).

### Checklist


ymartin59:
OK. What about still using `start-stop-daemon` ?

Safihre:
Sonarr and radarr don't need it. They do generate a PID file and update it correctly when self restarted. The location of it is just non-standard and so we need to specify it. 
Syncthing is the one that really doesn't generate one and needs the `start-stop-deamon` to function.

Safihre:
@ymartin59 We could always do that?
Just release a new `mono` and then set the new `noarch` Sonarr/Radarr packages to have `SPK_DEPENDS` for this specific version?

PR #3117, opened 2018-01-20
Wizard french translations for Sonarr and Radarr

ymartin59:


Safihre:
After the Python problems, I would suggest to first publish Radarr. The userbase of Sonarr is much larger so if there are problems we would affect a lot less users and gives us time to respond :) Since they share 95% of the code it's a good real world test.

PR #3120, opened 2018-01-21
Wizard french translations and permissions handling for NZBGet

ymartin59:

### DIFF
--- a/spk/nzbget/src/service-setup.sh
+++ b/spk/nzbget/src/service-setup.sh
@@ -2,7 +2,6 @@ PATH="${SYNOPKG_PKGDEST}/bin:${PATH}"
 NZBGET="${SYNOPKG_PKGDEST}/bin/nzbget"
 CFG_FILE="${SYNOPKG_PKGDEST}/var/nzbget.conf"
 TEMPLATE_CFG_FILE="${SYNOPKG_PKGDEST}/share/nzbget/nzbget.conf"
-UPGRADE_CFG_FILE="${TMP_DIR}/nzbget.conf"
 WEBDIR="${SYNOPKG_PKGDEST}/bin/webui"
 NZBGET_INSTALLER="${SYNOPKG_PKGDEST}/var/nzbget.run"
 GROUP="sc-download"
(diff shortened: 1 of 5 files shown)


ymartin59:
@Safihre As I planned to publish NZBGet yesterday... I have merged your PR and prepare translations on my own. By the way, upgrade path is broken, so I postponed publication

Safihre:
Did you test if the error reporting also works without these redirects to the temp log? 
Because I tested it and it would just show `Failed to install` but not the custom message without these redirects..

ymartin59:
OK. I will check again. Look at `/var/packages/nzbget/scripts/` generated scripts like `postinst`, all of them already redirect already to `$SYNOPKG_TEMP_LOGFILE`

Safihre:
@ymartin59 Anything I can do to help?
You said you had problems with `set_syno_permissions`?

ymartin59:
Right. Permissions visible with `ls -l` on various directory levels are strange after `set_syno_permissions`. I have planned to create automated test cases for various situations as shell scripts to run in `demoservice` package to validate this method works properly. You may propose such test scripts if you are willing to work on it.

Safihre:
I also noticed that, which makes sense since the Synology ACL-system is kind-a-but-not-completly seperate from unix permissions. So as long as the right users have access, this seems to be a Syno-side-effect?

ymartin59:
OK. But managing share from webui fixes unix permissions back to a+rwx... so my point of view is that `set_syno_permissions` should keep original unix permissions, only adding expected ACLs

Safihre:
@ymartin59 I think we should do it the same for NZBGet as I did for SABnzbd.
Instead of using `SHARE_PATH` I would suggest not to remove this section:
https://github.com/SynoCommunity/spksrc/pull/3120/files#diff-f316fe34eeaa89bb81c7bc0becd4fdc2L74
But to remove this whole `IF` section (so only leave the `grep` to detect `DOWNLOAD_FOLDER`):
https://github.com/SynoCommunity/spksrc/pull/3120/files#diff-f316fe34eeaa89bb81c7bc0becd4fdc2L18

ymartin59:
I removed that section because this is already processed by generic `postinst` in `installer` when `SHARE_PATH` and `GROUP` are set.

Safihre:
Ah yes, sorry it should be moved to `postupgrade` (like in sabnzbd).

Safihre:
@ymartin59 There are errors in your changes.
Typo:
```
if [ ! -e "${${CFG_FILE}" ]; then 
``
This:
```
 su ${EFF_USER} -s /bin/sh -c "${NZBGET_INSTALLER} --destdir ${SYNOPKG_PKGDEST}/bin" >> ${INST_LOG} 2>&1
```
Gives: `sh: /volume1/@appstore/nzbget/var/nzbget.run: Permission denied`

I will submit a PR to your branch with fixes.

**Comment by Safihre** [comment]
Needs additional fixes for DSM5, because it won't let us install as the package user.
```
Installer for nzbget-19.1
Verifying package...
Checking system...
CPU-Architecture: x86_64
Unpacking...
/volume1/@appstore/nzbget/var/nzbget.run: line 497: can't create /volume1/@appstore/nzbget/bin/installer.tmp: Permission denied
```

Added a fix 👍 
Now I really tested _everything_ 🤣

PR #3092, opened 2018-01-11
CouchPotato - Generic Service approach

Safihre:
### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #3123, opened 2018-01-22
Service Support: add background command execution

ymartin59:
Adapt Radarr and Sonarr to use background support
Fix package conf/ support for mosquitto

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -3,9 +3,9 @@
 #   scripts/installer
 #   scripts/start-stop-status
 #   scripts/service-setup
-#   conf/privilege            if SERVICE_USER
-#   conf/resource/SPK_NAME.sc if SERVICE_PORT
-#   app/config                if DSM_UI_DIR
+#   conf/privilege        if SERVICE_USER
+#   conf/SPK_NAME.sc      if SERVICE_PORT
+#   app/config            if DSM_UI_DIR
 #
 # Target are executed in the following order:
 #  service_msg_target
@@ -102,6 +102,13 @@ endif
 	@echo "# Service command to execute (either with shell or as is)" >> $@
 	@echo 'SERVICE_COMMAND="$(SERVICE_COMMAND)"' >> $@
 endif
+ifneq ($(strip $(SERVICE_EXE)),)
+	@echo "# Service command to execute with start-stop-daemon" >> $@
+	@echo 'SERVICE_EXE="$(SERVICE_EXE)"' >> $@
+ifneq ($(strip $(SERVICE_OPTIONS)),)
+	@echo 'SERVICE_OPTIONS="$(SERVICE_OPTIONS)"' >> $@
+endif
+endif
 	@cat $(SPKSRC_MK)spksrc.service.call_func >> $@
 ifneq ($(strip $(SERVICE_SETUP)),)
 	@cat $(CURDIR)/$(SERVICE_SETUP) >> $@
@@ -135,8 +142,9 @@ ifneq ($(strip $(SPK_USER)),)
 $(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege
 	$(create_target_dir)
 	@sed 's|USER|sc-$(SPK_USER)|' $< > $@
-	$(eval SPK_CONTENT += conf)
-
+ifneq ($(findstring conf,$(SPK_CONTENT)),conf)
+SPK_CONTENT += conf
+endif
 SERVICE_FILES += $(DSM_CONF_DIR)/privilege
 endif
 
@@ -159,13 +167,17 @@ else
 endif
 	@echo "port_forward=\"yes\"" >> $@
 	@echo "dst.ports=\"${SERVICE_PORT}/tcp\"" >> $@
-	$(eval SPK_CONTENT += conf)
+ifneq ($(findstring conf,$(SPK_CONTENT)),conf)
+SPK_CONTENT += conf
+endif
 SERVICE_FILES += $(DSM_CONF_DIR)/$(SPK_NAME).sc
 endif
 else
 $(DSM_CONF_DIR)/$(SPK_NAME).sc: $(filter %.sc,$(FWPORTS))
 	@$(dsm_script_copy)
-	$(eval SPK_CONTENT += conf)
+ifneq ($(findstring conf,$(SPK_CONTENT)),conf)
+SPK_CONTENT += conf
+endif
 SERVICE_FILES += $(DSM_CONF_DIR)/$(SPK_NAME).sc
 endif
 
--- a/mk/spksrc.service.start-stop-daemon
+++ b/mk/spksrc.service.start-stop-daemon
@@ -1,0 +1,96 @@
+#!/bin/sh
+
+# Default display name
+DNAME="${SYNOPKG_PKGNAME}"
+
+# Source package specific variable and functions
+SVC_SETUP=`dirname $0`"/service-setup"
+if [ -r "${SVC_SETUP}" ]; then
+    . "${SVC_SETUP}"
+fi
+
+start_daemon ()
+{
+    if [ -z "${SVC_QUIET}" ]; then
+        if [ -z "${SVC_KEEP_LOG}" ]; then
+            date > ${LOG_FILE}
+        else
+            date >> ${LOG_FILE}
+        fi
+        echo "Starting ${DNAME} command ${SERVICE_COMMAND} " >> ${LOG_FILE}
+    fi
+    call_func "service_prestart"
+    if [ -n "${SERVICE_EXE}" ]; then
+        if [ -n "${SERVICE_OPTIONS}" ]; then
+            start-stop-daemon -b -o -c ${EFF_USER} -S -u ${EFF_USER} -k 002 -x "${SERVICE_EXE}" -- "${SERVICE_OPTIONS}"
+        else
+            start-stop-daemon -b -o -c ${EFF_USER} -S -u ${EFF_USER} -k 002 -x "${SERVICE_EXE}"
+        fi
+    fi
+}
+
+stop_daemon ()
+{
+    start-stop-daemon -o -c ${EFF_USER} -K -u ${EFF_USER} -x "${SERVICE_EXE}"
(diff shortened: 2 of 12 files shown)


ymartin59:
@m4tt075 @Safihre Here is a preview of a small improvement to avoid copy/paste block for background execution when command does not need to be computed in `service_prestart`
Just a preview, as I have not tested it (at all... will come soon)

ymartin59:
To help you to review, here is the most important part: https://github.com/ymartin59/spksrc/blob/af6a098d2f304368a48e89dc6d06ab57e8d3c55f/mk/spksrc.service.start-stop-status

Safihre:
Great!
There were more reports of the new packages not generating DSM  shortcuts even though the requirements (Config file and images) were present.  Could that be because they are not from "trusted sources"? As in: when you install Mosquitto from the repo or from manual upload, do both give a shortcut?

ymartin59:
@Safihre Mosquitto is a "backend" service without shortcut... But `demoservice` works properly. I thought first of `RELOAD_UI=yes` Makefile option but it is present. Do you know if it happens for fresh install or upgrade from previous package (or both) ?

Safihre:
Both it seems to happen with.

ymartin59:
I wonder if it may come from JSON syntax around value `"allUsers": true,` because I added quotes in my generator https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.mk#L201
What do you think about it ?

Safihre:
@ymartin59 Yes, there should *not* be quotes to be valid JSON. Can you fix that? :)

(My internet is finally installed so expect more from me in the next few days 🤗)

Safihre:
@ymartin59 I have quite some time free (starting new job only in few weeks) but the number of my open PR's and unpublished packages that I put so much time in is kind of depressing me and not very motivating for me..
In order to keep things going I believe these things should be fixed (soon):

1. `set_syno_persmissions` problems that you described.
2. Have the `stop-start-deamon` part of Generic Service.
3. The application shortcut not showing (possibly due to this JSON formatting).
4. We should switch all download-processing packages to be `sc-download` and if nessecary also `sc-media` if package-user was part of that before. I can write a conversion function that does this and add it to Generic service so we can upgrade Sonarr/Radarr/CouchPotato/Sickbeard.

Can you fix 1 and 2 so I can continue again with the packages? 
I can do 3 and 4 and then we can get all of these new `sc-media` packages published?

As I also wrote in the other thread, the failing Python package seems ARM specific and very hard to debug, but I will try.

m4tt075:
Nice list. Only additional item I can think of is...

5. Update `ffmpeg` in line with licensing requirements

ymartin59:
@m4tt075 Hum. Why not, but I expect users complains about many features removal

Safihre:
@ymartin59 Great!
I found one more thing for the list: I noticed with a number of packages that when uninstalling a package, the package user doesn't get removed from the `$GROUP` and (maybe due to group-membership?) also don't get removed at all. They are still in `/etc/passwd`.

Safihre:
Put the code for item 4 in PR #3092.
There I also noticed these wrong re-directions in some `if` statements. Maybe it will help with the users/groups getting removed correctly upon uninstall.

BenjV:
@ymartin59 @Safihre 
I looked at the scripts and there is no code that removes the user from the group when is user is removed.
To fix this you can simply add the following command after removing the user to the routine that remove the user.
This will automatically remove all no existing users from all groups.

synogroup --rebuild all

ymartin59:
@BenjV Welcome back and many thanks for your feedbacks

Safihre:
@ymartin59 Got myself a second-hand DS112J (`88f6281`), hopefully next week find out what's wrong with that stupid Python package.

Safihre:
Included the suggestions by @BenjV in #3065, both the fixing of the icon and the fixing of user removal on uninstall.
_(for reference, too many PR's and issues)_

Safihre:
I can now confirm 1) in some way, after the update all folders have the correct permissions for Synology (output below).
However, all folders are set to Linux permissions `000` (output below).
Now when SABnzbd starts it forces the permissions of the download folder using Linux permissions. This seems to invalidate the Synology permissions of the sub-directories, as then it cannot access the sub-folders anymore.
I'm not sure what to do about that.. Should we do `chown -R` instead of `synoacltool -force-inherit`?

Synology permissions (before SABnzbd start after upgrade):
```console
admin@DiskStation:/volume1$ sudo synoacltool -get "/volume1/test1"
ACL version: 1
Owner: [root(user)]
---------------------
         [0] group:sc-download:allow:--x----------:---n  (level:0)

admin@DiskStation:/volume1$ sudo synoacltool -get "/volume1/test1/incomplete"
ACL version: 1
Owner: [root(user)]
---------------------
         [0] group:sc-download:allow:rwxpdDaARWcC-:fd--  (level:0)

admin@DiskStation:/volume1$ sudo synoacltool -get "/volume1/test1/incomplete/ubuntu iso"
ACL version: 1
Archive: is_inherit,is_support_ACL
Owner: [root(user)]
---------------------
         [0] group:sc-download:allow:rwxpdDaARWcC-:fd--  (level:1)
```
Linux permissions (before SABnzbd start after upgrade):
```console
admin@DiskStation:/volume1/test1$ ls -ltr
total 12
drwxrwxrwx+ 3 root  root  4096 Feb 21 11:33 @eaDir
d---------+ 3 root  root  4096 Feb 21 11:42 incomplete
```
SABnzbd start after upgrade:
```
2018-02-21 11:20:08,468::INFO::[__init__:929] Traceback: 
  File "/volume1/@appstore/sabnzbd/share/SABnzbd/sabnzbd/__init__.py", line 916, in load_data
    with open(path, 'rb') as data_file:
IOError: [Errno 13] Permission denied: u'/volume1/sabtest/incomplete/ubuntu iso/__ADMIN__/SABnzbd_nzo_WG1Axc'
```

Safihre:
Fixed it!
We need to change permissions first from `root:root`/`root:users` to the new `{EFF_USER}:${USER}` and then let `synoacltool` do it's thing. But only when it's in Linux mode so we don't remove any Synology permissions.
Put in #3161.

Now we only have item 2 left (`start-stop-daemon`)!

ymartin59:
@Safihre @m4tt075 Here is background execution and `start-stop-daemon` support

Safihre:
Lol, now you merged things without anyone having the time to look at it at all 😉 
So can you explain the new options or add them to the wiki?
`SVC_BACKGROUND` doesn't make it super clear to me that it will use `start-stop-deamon`, but as long as it's in the wiki that's fine.
What does `SVC_WRITE_PID` do in combination with other settings?

BenjV:
@ymartin59 and @Safihre 
I am not sure if this is the right place to place this text and also not sure if it applies to the framework but I want to share my findings with you guys.
I noticed some strange behaviour on a package that I created myself.

It is for a pyhon application and the applicatie itself creates a pidfile.
Because of the nature of the application it takes a second or two after starting before the pidfile is created.
In the meantime DSM conclude via the "start-stop-status status" command that the application is not running and fires the "start-stop-status stop" command.
This will stop this application.
The effect is the the application start, is running a little time and then stops again.
I can fix it by adding a little sleep time in the "start-stop-status status" command but that does not feel ok. 
I am tring to figure out to solve this problem.

So maybe this is also happening to other packages.

Safihre:
@ymartin59 Are you sure this works?
I don't see code anywhere that includes the `spksrc.service.start-stop-deamon`?

Safihre:
I tried to use it for Syncthing (PR #3094) but it just gives me the regular `start-stop-status`. Or did I implement it wrongly?
https://github.com/SynoCommunity/spksrc/pull/3094/commits/d13a324b9bfb6cc5ba3dd1237c2f83f96393cc3b#diff-2da771f505835963832df94a7742c007

ymartin59:
@Safihre You're right... I merged it to publish Sonarr and Radarr. I told myself it was good enough and visible/advertise here since weeks.
About `start-stop-daemon`, its usage is not enabled yet as I am looking for a "simple" package to test code.
Here is updated documentation: 
- https://github.com/SynoCommunity/spksrc/wiki/Service-Support#run-service-process-in-background
- https://github.com/SynoCommunity/spksrc/wiki/Service-Support#service-running-thanks-to-busybox-start-stop-daemon
Do not hesitate to request for more details if needed to understand.

BenjV:
@Safihre 
Sounds good, because applications that deamonise themselves will always need some time before the pidfile is written.
I don't know if it is needed to be configurable, just build a loop testing if the pifdfile is there every second en exit with a fault after say 15 seconds, but configurable is also ok.

ymartin59:
It was a mistake of mine: as far as stopping service still invokes `wait_for_status 1 20`, I failed to report original `wait_for_status 0 20` in `start_daemon`... Because of transmission report, I will have to re-publish it (again)...

ymartin59:
Here is my fix proposal: https://github.com/SynoCommunity/spksrc/pull/3186/commits/c6aab73c4e2d7c7648f3b3fdda12594fc8f6cbdb

Safihre:
Hmm, but isn't this the fault of the application?
As in: an application should only demonize after writing the PID (or simultaneously)? 
Since the `daemon_start` only returns after the command returns, only after that DSM will check its status.

ymartin59:
@Safihre I agree with you. I discard that change, there was no such logic in DSM 5 `dsm-control.sh` scripts.

BenjV:
Sorry but first writing a pid file and then deamonize is not possible for a linux application.
If a process demonize it becomes a new pid, so the pidfile can only be written after it is deamonized.

If it would write the pidfile before deamonizing it would be the wrong pid.

Safihre:
The parent process should only return after it has forked. When it does the fork, it receives the child's PID and can write that to the file. For example with Python: https://docs.python.org/2/library/os.html#os.fork
This is exactly what we do in SABnzbd. 
That way the `status` is always only ran after the process is definitely active.

BenjV:
Why did you write before the that an application should only demonize after writing the PID?
This is the opposite of what you are saying now(and which is correct).

Safihre:
Got confused initially while reading Sab code, sorry!

BenjV:
Ok no problem, just wanted to verify the contradiction.

Issue #3041, opened 2017-12-10
uMurmur package for DS3018xs


mikhnal:
### Setup
Package Name: uMurmur
Package Version: v0.2.16-6

NAS Model: DS3018xs
NAS Architecture: broadwellnk
DSM version: 6.2-22259 beta

### Expected behavior
Please make a package for new model DS3018xs and CPU architecture broadwellnk.

### Actual behavior
Any current package is not working.

hgy59:
@mikhnal Still looking for this?
I have built it with the 6.2-22259 toolchain. Can you test please?
download at:
https://ziba-systems.net/downloads/umurmur_broadwellnk-6.2_0.2.16-6.spk

mikhnal:
Hi!
It works perfectly, thank you for the effort!

ymartin59:
@hgy59 Not really a good idea as this package has not been ported to DSM 6 generic service support

m4tt075:
Just created PR #3132 adding generic service support.

m4tt075:
DSM6-compatible test packages are available [here](https://github.com/m4tt075/spksrc/releases/tag/umu-0.2.17) now. Please test and report back.

ymartin59:
uMurmur has been published. Please report any trouble in new issue

PR #3127, opened 2018-01-28
toolchain of broadwellnk arch

hgy59:
this is the toolchain of broadwellnk for the new DS3018x and FS1018 that support DSM 6.2 only.
even the DSM 6.2 is still beta, the toolchain on sourceforge.net is not named beta

_Motivation:_ DS3018x, FS1018 support DSM 6.2 only.
_Linked issues:_ #3041

### Checklist
- [x] Build rule `all-supported` completed successfully (umurmur)
- [x] New installation of package completed successfully (umurmur)

### DIFF
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -39,7 +39,7 @@ ARM8_ARCHES = rtd1296
 ARM_ARCHES = $(ARM5_ARCHES) $(ARM7_ARCHES) $(ARM8_ARCHES)
 PPC_ARCHES = powerpc ppc824x ppc853x ppc854x qoriq
 x86_ARCHES = evansport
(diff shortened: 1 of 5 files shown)


ymartin59:
Please add this arch in aliases of `syno-x64-5.2` and `syno-x64-6.1` build chain so that a single package built with these targets and deployed to all of them

hgy59:
Where are these aliases defined?

hgy59:
Think I got it. syno-x64 looks like an alias-arch for several x64 archs like apollolake, broadwell, ...
And apollolake is defined as TC_ARCH in syno-x64-5.2/Makefile - even there is not Diskstation with apollolake and DSM 5.2.

PR #3133, opened 2018-02-01
Update htop to 1.0.3

4sag:
_Motivation:_ Explain here what the reason for the pull request is.
_Linked issues:_ Optionally, add links to existing issues or other PR's

### Checklist

### DIFF
--- a/cross/htop/Makefile
+++ b/cross/htop/Makefile
@@ -1,8 +1,8 @@
 PKG_NAME = htop
-PKG_VERS = 1.0.1
+PKG_VERS = 1.0.3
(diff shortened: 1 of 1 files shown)


m4tt075:
@4sag Thanks for contributing. If you upgrade the cross package version you have to update the spk package too, otherwise the versions will not match. Please also update the digests file in the cross folder. Finally, it seems that the latest htop version is 2.0.2. Any reason, to stay with 1.0.3 instead?

hgy59:
htop 2.0.2 is now in synocli-file (but for DSM < 6.0 only, as htop is contained in DSM 6)

Issue #2718, opened 2017-03-25
Synology DS3617, Missing packages


MACE-Zer0:
Hello , dont know if its the right place , but i dont see many of the packages of https://packages.synocommunity.com/ , in my new ds3617 , i have a similar one here the ds3615 , and i see all the packages ,  do i need to do something or is the ds3617 not compatible with these packages , hope not because i need a few of those , 

if im not in the right place i apologise

A few i see missing :
chromaprint 
mediainfo 
disktutils
git 
gitserver
ZNC
Zshell,

many more

Thanks

MACE-Zer0:
Can anyone help me out , have the nas on hold , because i need a lot of missing packages that are only missing in ds3617xs , have tired reinstalling nas  , changed dns servers , but still nothing , in other nas i have  all the packages ..... help !!!!

MACE-Zer0:
The only packages i see 

![image](https://cloud.githubusercontent.com/assets/9946471/24331431/45f20b36-122c-11e7-8a51-2aa1a15cd0ad.png)


![image](https://cloud.githubusercontent.com/assets/9946471/24331434/517e1dfa-122c-11e7-9826-5cb1ed3e7a0f.png)

ymartin59:
What is your DSM version on new NAS ? It is possible missing packages are only available for DSM 5.2 - supposing packages you are interested in on your previous NAS when running DSM 5.2. Is my hypothesis relevant/right ?

MACE-Zer0:
Both nas has the most recent version , 6.1-15047, and  iv always had all the packages no mather the version installed , this has something to do with the DS3617 because its very recent model of synology , but i need help from someone in synocommunity , because iv tried all that i can remember

ymartin59:
Almost no package for DSM 6 is available because of #2216 . Compilation toolchain is not available for your platform yet #2509 

Sorry but nothing can be done for the moment, and there is no other reason that packages are not available for your sytem. Do not spend more time trying to "fix" it... except if you are interested in helping testing packages for instance.

MACE-Zer0:
@ymartin59  , thanks for the info , havent seen those issues , guess only have to wait !!!

ymartin59:
@MACE-Zer0 Toolchain for your system is available but I have no idea when packages may be published to package repository. As your system supports Docker, it may be possible for you to build packages you are interested in directly on your Synology. If interested, I may provide details about this process.

nicky1038:
[...]

PR #3136, opened 2018-02-04
Updates for ffmpeg, mediainfo and chromaprint

ymartin59:
_Motivation:_ Deliver packages to new architectures `rtd1296`, `broadwell`, `broadwellnk`, `apollolake`...
_Linked issues:_ #2718 #3119 

- [X] Build rule `all-supported` completed successfully (except `chromaprint` 1.4.3 on `88f6281` because of GCC version and C++ standard)


m4tt075:
@ymartin59 Thanks for picking this up! Much appreciated. 

I have two questions: 
1. This is related to HW transcoding: Would it be possible to enable VAAPI support on appropriate platforms (I assume that there are no Syno platforms utilizing NVENC, MMAL or OMX)?
2. The statically linked `ffmpeg` version `Tvheadend` is using is patched. See [here](https://github.com/tvheadend/tvheadend/tree/master/support/patches) for the patches applied. I would like to add them if (and only if) `ffmpeg` is cross-compiled as a dependency of `Tvheadend`. Could you create some hook to achieve that? Happy to add the patches via PRing against your branch, if you are fine with it. Just let me know...

ymartin59:
@m4tt075 I am working to get `mediainfo` and `chromaprint` dynamically linked against `ffmpeg` package libraries... If OK, sure it sounds relevant to "improve" ffmpeg so that tvheadend can also use that libraries.
I have reviewed https://www.ffmpeg.org/legal.html because of "License: nonfree and unredistributable" reported after configure... and I am quite uncomfortable about it

m4tt075:
OK, interesting, and very relevant. Tvheadend is already linking ffmpeg dynamically and has been for some time. It is just not utilizing all optimizations. But I see your point, and that it is way more important to figure out. I will just keep on tracking this PR. Maybe we can get back to the optimizations at a later stage. No urgency...

ymartin59:
@m4tt075 I have investing `--enable-vaapi`, adding https://github.com/intel/libva which requires libdrm... and I just guess all this stuff is intel only. Am I right ?

m4tt075:
@ymartin59 I believe you are right that vaapi is for Intel only. I'm not sure which libraries are needed for transcoding support though. Looking at the [ffmpeg configuration for TVH](https://github.com/tvheadend/tvheadend/blob/master/Makefile.ffmpeg), libva does not seem to be required. But I'm purely guessing here. I don't know enough about ffmpeg to judge. Maybe @cytec knows? I think he has been working on ffmpeg quite a lot...

ymartin59:
@m4tt075 Following these instructions https://gist.github.com/Brainiarc7/eb45d2e22afec7534f4a117d15fe6d89 
I fear VA API requires building and release additional kernel modules...

m4tt075:
@ymartin59 Yikes! That looks way more complicated than what it takes for TVH. Look, don't spend too much time on this, please. If it works, great, if it doesn't, please don't worry...

ymartin59:
Just found one issue with:
```
xpenology52> /usr/local/ffmpeg/bin/x265 -version
/usr/local/ffmpeg/bin/x265: error while loading shared libraries: libx265.so.130: cannot open shared object file: No such file or directory
```
But works with `export LD_LIBRARY_PATH=/var/packages/ffmpeg/target/lib/`.
It looks like binary is linked without install prefix !

ymartin59:
@m4tt075 About vaapi, I think `configure` is looking for VA availability on hardware thanks to loaded modules like `drm` or `/sys`. Of course, there is troubles with Docker environment and cross-compilation context. I have postponed until I have new ideas: https://github.com/ymartin59/spksrc/issues/6

ymartin59:
Now I have investigated `x265` support... I face another issue: multibit support is quite an odd compilation process: https://github.com/ymartin59/spksrc/issues/7
I cannot left behind defaults 8 bits to replace by 10 or 12... it does not make sense. Package should be provided with these three options available.

cytec:
@ymartin59 regarding the chromaprint makefile: why are you checking for the spk to be present on the build machine? personally i don't care if it gets rebuilt everytime, sure it takes a while but most tools don't need to be published that often (et least not those who link against ffmpeg) so i'd be fine with them building it again... i don't know if dynamic linking makes any sense there. The reason i didn't do it back then was to keep it simple. One should be able to install chromaprint without the need of installing ffmpeg first which he might not even need except for the libs that are needed by chromaprint ... 

PS: afaik vaapi is intel only and not needed for tvh

ymartin59:
@cytec Thanks for your answer. I investigated this shared library concept because of compilation time and package volumes managed in repository - ffmpeg libraries are required in chromaprint and tvheadend. Now chromaprint package are small and built faster. In general both are delivered together when ffmpeg package is updated. vaapi was requested by ffmpeg users.

ymartin59:
Now I have enable multibit `x265` support... it no longer builds for `rtd1296`. A work-around with `-DENABLE_ASSEMBLY=OFF` but not satisfying.

cytec:
im perfectly fine with using shared lib's for ffmpeg ;)
but im not really sure how the best approach would be for building those SPK's i think the way you currently doing it is fine for now... another way would be to just use shared libs but rebuild them everytime which might help getting rid of the makefile stuff... but imho that's something we'd have to consider: smaller packages which depent on each other with slightly less beautiful make files or bigger packages with longer build times and a cleaner makefile...

just a random thought: what about doing something like the native depend stuff for those cases if we get more of them? so something like "platform shared" or similar?

m4tt075:
@ymartin59 I just ran some tests. It seems all arm-platforms are broken. At least I could not compile them. As ppc-platforms are broken too, basically only x86/x64 platforms are left. Do you want depth or breadth? Or we compile 10 and 12 bit for Intel platforms only, but then it is getting more and more cumberrsome and complex which is not ideal either.

m4tt075:
@ymartin59 Thanks for adding all those `Tvheadend` patches. I don't understand enough of the code, so I'd rather ask: Is it ok to apply them independent of which `spk` is actually compiled? Or should we only apply them conditionally for the case that the `ffmpeg` cross-package is compiled as part of the `Tvheadend` `spk`? Just worried we might break non-TVH packages by applying those patches across the board. What do you think?

ymartin59:
@m4tt075 I have reviewed TVH patches to `ffmpeg` and related libraries. They are "harmless" and only add some switches and events for progress notification. That is why I applied them unconditionally.
I invite you to reuse make command to link against ffmpeg package instead of rebuild and include it in TVH package: see `spk/chromaprint/Makefile` and `cross/chromaprint/Makefile` from https://github.com/SynoCommunity/spksrc/pull/3136/commits/cbf9d26c97d2f6c348aaf7c8c19e53e5a431a610

m4tt075:
@ymartin59 Cool! Thanks. Great idea to link to the `ffmpeg` package. Should save a considerable amount of time building `Tvheadend` again and again.

PR #3143, opened 2018-02-09
Midnight Commander 4.8.20 and use Generic Service

Safihre:
This package wasn't available yet for newer archs like apollolake. Did the minor upgrade to Generic Service, but wasn't even needed really. 

Also relates to #3095, but seems it wasn't published after that change.
Tested by @rusty1281.

Such a simple package... will it be possible to get published soon please?
 
### Checklist


ymartin59:
Just hint for HTTP download location I already have looked for: https://sourceware.org/pub/libffi/libffi-3.2.1.tar.gz

rusty1281:
Just last night I noticed something regarding mc application. version 4.8.20 that was build is not using the move command as it should. My current one on 412+ (4.8.15-3) is working as it should.

Problem: when I select to move multiple files on the same volume but from one shared folder to another it actually copies data 1st and then deletes the source. If I however do the same thing in the same shared folder (from one folder to another) then it moves data as it should (fast).

exmp:

source: /temp/folder
destination: /destination_shared_folder/folder

In this case folder from Temp that needs to be moved to my destination shared mount point is copied and not moved. Again, this does not happen on my old 412+ with its version of MC. Both NAS only have one volume with multiple shared folders on it so i guess I expect the same behaviour.

Can anyone verify this as a bug or is this something thats expected?

tnx in advance

Safihre:
@ymartin59 Since `ncurses` and `libffi` are also part of Python-build I hope it doesn't break anything. Will do my best.. Have not been able to reproduce the problems people have reported with Python...

@rusty1281 the OS decides this, we also see it in Python. If the OS thinks it's 2 separate volumes/disks or other things, it will do a copy-delete instead of a move. It also happens if for some reason the file is (semi) locked when it tries to do the move.

ymartin59:
@rusty1281 It is possible `mc` has changed its way to handle symbolic links and no longer dereference/follow them before invoking `rename` libc function. I am also used to such "trouble" with `bind` mount points... So the best option is to look for absolute canonicalized file path yourself to avoid this copy/remove when you know both source and destination are located on same file system.

rusty1281:
@ymartin59 & @Safihre - thank you both for your reply on this matter. Not sure what to say on this considering the layout of shared folders on my new NAS and old one are the same. Both NAS run the same OS version and like I said before, older one does move data and new one copy/deletes in some case. @Safihre tried to run the 1st build that you did a few days ago but the result is the same.

Not sure if it matters, but the nas is using an SSD cache thats attached to that single volume, the old nas doesnt have a cache od any sort.

Also the old nas is running etx4 and new one is on btrfs fs.

@ymartin59 there are no symlinks in any of my scenarios so thats why thats really puzzling me.

mostly I have paths like /volume1/share_folder/folder1,2,etc... So moving content of one shared folder to another (with no sym links) shouldn't invoke copy/delete... 

Even the folder structure on both NAS is the same (considering I'm replacing the old). Guess I'll get used to it.

Also, one more thing, if I try and move data using their file explorer (web or mobile app) its moves instantly, so thats one more reason why I question this mc behaviour.

Again, thanks for your input and help on this matter

ymartin59:
You're right to question. Reason may comes from btrfs file system... which one I would not recommend as even RedHat has discarded support of this file system from its Linux distribution, looks like it cannot be considered as "stable" enough.
By the way, I will test such move operation with updated version on `ext4` to get additional clues about it.

Safihre:
Updated cross packages, no HTTPS available for `slang` (what's wrong with FTP actually?).

Safihre:
Hmm, the `ncurses` update is giving problems, I remember a similar thing when I was working on Python. 
Will try to fix now.

Safihre:
Sigh. Too much changed in `ncurses(w)` 6.1 and now it has some cross-compilation problem on non-x86 platforms. 
I will try for another hour or so, otherwise I suggest to just stick to version 6.0.  I just don't know enough about this all and don't want to get stuck in it.

Something is wrong with the `tic` binary that's used by ncurses to compile the terminfo, which normally uses the just compiled `tic` but since we are cross compiling it's for the wrong platform. Or something like that.
The strip-process after the compile fails with `strip: Unable to recognise the format of the input file "/spksrc/spk/mc/work-rtd1296-6.1/install/var/packages/mc/target/bin/tic"`

Safihre:
Fixed the problem. [Compiles fine now with updated dependencies](https://travis-ci.org/Safihre/spksrc/builds/340540130).
Will[ also test-compile Python again](https://travis-ci.org/Safihre/spksrc/builds/340540729) and see how that works out.

Safihre:
@rusty1281 could you maybe re-download and install the newest Midnight Commander package here? https://github.com/Safihre/spksrc/releases/tag/mc-dsm61-all

Even though it ha she same name, it's a newer version that has these dependency updates inside. Just want to make sure it really works as intended (apart from this copy behavior).

rusty1281:
Yep, runs fine. All commands are good, and copy/move 'problem' is still here, but lets ignore that. Apart from that everything else is working as it should.

Safihre:
Cool, the copy problem is hard for me to diagnose if it's specific to the package or the system. Seems @ymartin59 could maybe be of better help with that :)

rusty1281:
Well getting used to it by now, but its really annoying considering that its working as it should via Synology DSM UI. 

Tnx for your effort on this, really appreciate it.

PR #3144, opened 2018-02-09
Transmission 2.93 and Generic Service approach

Safihre:
### Checklist


ymartin59:
Great. May you please take opportunity to update dependencies too: `cross/libevent` to 2.1.8-stable, and `cross/curl` to 7.58.0.
Thank you in advance for this effort

ymartin59:
For other packages, `_enu` wizard translation files have been removed to keep only default one. What do you think about their removal ?

Safihre:
@ymartin59 I never understood the `enu` files, I can't find any reference to `enu` in the Synology documentation.
So maybe indeed they can go, just in this case there was no non-`enu` version which kind of surprised me.

Stanzilla:
@Safihre looks like the package is using sc-transmission:transmission instead of sc-transmission:sc-media?

Safihre:
Yes. But it is part of `sc-download` (not `sc-media`) so this captures other permissions.

Stanzilla:
meh, something is weird though, Sonarr complained again: 

`Import failed, path does not exist or is not accessible by Sonarr: /volume1/downloads/x.mkv`

Safihre:
Yes, because Sonarr uses `sc-media` (we will change this in the next release to also be `sc-download`).

Stanzilla:
I see, are there testbuilds for that yet?

Safihre:
Not yet, since we have to implement proper changes for existing users that upgrade for the `sc-media` -> `sc-download` transition.

Safihre:
@ymartin59 Removed the `enu` files.
Since I am no real Transmission user, inclusion of other skins is not my priority and maybe someone else can do this (later).

alex-111-gh:
This is very much needed. [2.93 fixes](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-5702) some some serious vulnerabilities. How do we move this forward? As far as I can tell this change looks good.

ymartin59:
I will merge and publish it today

PR #3149, opened 2018-02-10
Cross cleanup

hgy59:
_Motivation:_ When I checked all the source downloads of cross Makefiles, I found that some modules are outdated and not referenced by any spk and therefore can be removed from the repository.
_Linked issues:_ Just one year ago there was a cleanup proposed in PR #2658. 

The PR #2658 deleted 157 files and some are still in use or under development. It was too much for me to review all the dependencies there, so I deleted only cross modules after manual checking that no dependency exists anymore.

- cross/serf, cross/apr-util and cross/apr
should have been removed when subversion spk was removed.
- cross/dbus, cross/libcap2, cross/iksemel and cross/libical
where never used
- cross/tinyproxy
should have been removed when when spk was removed

PR #3152, opened 2018-02-12
Fix downloads

hgy59:
_Motivation:_ Some download URLs are outdated and download of cross source code fails
_Linked issues:_ In addition to #3149 delete obsolete cross module openvpn.

~~As far as possible the version of the packages did not change.~~

### Checklist
- [x] Build rule `all-supported` completed successfully ~~(exept squidguard, see below)~~
### Fixes:
- delete ~~cross/jemalloc and~~ cross/openvpn
- fix cross/jemalloc URL
- adjust pyload patch (even the source version did not change ?)
- add package name to downloaded archive files to avoid naming conflicts (dvb-apps, lz4, plowshare, rutorrent) - avoid version only filenames in distrib folder.
- dnscrypt 1.9.5 URL moved to fossies.org, developper is working on 2.x (still beta) and does not provide archive for 1.9.5 anymore.
- 'eng' removed from archive-name of tessdata-eng since it contains data for a lot of languages. May be we will support data for other languages in tesseract-ocr and extract it from the same archive.
- jappix.org domain is not in use anymore, use github; no need of patch, the lib rt is already in CMakeLists even the version did not change (?)
- octoprint and cops digests (hash) update
- ~~squidclamav updated to 6.16 since version 6.10 is not available on sf.net anymore - but squidguard does not compile (squidclamav.c:1569:21: error: 'ci_membuf_t' has no member named 'flags')~~
- memcached URL update (googlecode.com is not in use anymore)
- imagemagick URL update (download contains current version only, but download/releases has all the released versions)


ymartin59:
Hum, sorry but there is too much stuff here...
- `cross/jemalloc` is a development tool, replacing standard allocator to diagnose memory leaks. So it is not included in any package, except to produce a test package for troubleshooting 
- updating version requires a package specific PR to review application and dependencies (There is a pending PR for squidclamav full update and DSM 6 support)

hgy59:
@ymartin59 thanks for the hints.
Have reverted all version updates and restored jemalloc (and added comment about usage of jemalloc)

PR #3155, opened 2018-02-19
Remove references to deprecated arch 88f5281

Safihre:
The `88f5281` arch was depcreated a while ago, now all reference to it can be removed.
Also replaced some arch-specific selections to the definitions from `*_ARCHS`.

@ymartin59: I was added to the SynoCommunity group by @Diaoul (to my surprise). Will try to contain my desires to just commit-commit-commit and use the PR's to get double checks! I might close some issues, since I really like cleaning 🥇

### DIFF
--- a/cross/alsa-lib/Makefile
+++ b/cross/alsa-lib/Makefile
@@ -13,7 +13,7 @@ LICENSE  =
 
 GNU_CONFIGURE = 1
 CONFIGURE_ARGS = --disable-python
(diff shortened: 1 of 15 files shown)


ymartin59:
Seems correct. Difficult to test everything but I do not expect troubles
I requested membership for you and @m4tt075 - as reward and comfort when working on project.

Safihre:
👍 😃

PR #3156, opened 2018-02-19
Python correct Busybox and re-introduce pyasn1

Safihre:
I now have my DS112j (`88f6281`) and it's significantly slower than an XPEnology but at least I was able to do some proper testing.
So I have tested: `sabnzbd`, `deluge`, `couchpotato` and `sickbeard` and all of them work with this new package. I tested it properly, so with a number of test downloads.
Everything works, no errors.

Only arch that worries me is `88f6281`, it has more warning messages in the build-logs than the other archs. But hard to know if it's actually bad..

### Checklist
- [x] [Build rule `all-supported` completed successfully](https://travis-ci.org/Safihre/spksrc/builds/343248822)


ymartin59:
OK. An option may be to deliver python updates to all architectures except 88f6281, but then application packages depending on it may be in troubles. What do you think about it ?

Safihre:
I tested this yesterday to, _all_ new packages work also with 2.7.13 that's now on the repo. So they don't need this update 💯 👍 
I am not even sure if `88f6281` will be a problem, would need a real tester for that I guess. 

Could also try a staged-approach: first upload new build for `armv5` (same as my test system), if no problems reported after ~2 days, upload all `x64`, again if no problems 2 days later, upload the rest.

ymartin59:
I agree with the process. I merge and publish armv5 packages only for the moment.

Safihre:
:+1:

ymartin59:
Just find out that `spk/python/Makefile` revision should have been increased to 18... I will commit that

Safihre:
The Python package on my NAS just automagically updated, all seems to be working perfect 💯

Safihre:
So, no problems reported. Let's publish `x64`/`x86`?

Safihre:
Let's try another arch? We haven't had any reports.

ymartin59:
I suspended publication because of #3179.

Safihre:
@ymartin59 Please publish the `x64` and `rtd1296` archs for the new Python package. This way at least people with the new NAS models can install Python. 
#3201 does not seem to affect these archs, both of the test-packages have 100+ downloads each without any reported errors.
#3179 is unrelated and closed for a while.
Thank you.

ymartin59:
@Safihre Python 2.7.14-18 has been published for 88f628x, x64, rtd1296 now

Safihre:
Thank you 💯

Alph2:
@ymartin59 I cannot seem to find the release of Python 2.7.14-* for rtd1296 you mentionned. Am I missing something?

retrography:
@ymartin59 I do not see it either. Here is what I see, in Package Manager and on Synocommunity website:

screenshot 2018-12-14 at 15 21 55
screenshot 2018-12-14 at 15 22 33

PR #3151, opened 2018-02-11
add kvmx64 arch (DSM 6.1 toolchain and kernel)

hgy59:
_Motivation:_ Synology Virtual DSM is not yet supported by SynoCommunity
_Linked issues:_ #3142

As commented in #3142, synology names the arch of Virtual DSM kvmx64 (fixed in mk/spksrc.common.mk)
Maybe the synology DSM package manager is not yet ready for this...

### Checklist
- [x] Build rule `all-supported` completed successfully
- [n/a] Package upgrade completed successfully
- [ ] New installation of package completed successfully: TODO @rheckly

TODO:
- Test kvmx64 arch with DSM package manager

PR #3131, opened 2018-01-29
add more DSM 6.1 kernels

hgy59:
here comes the remaining DMS 6.1 kernels
still missing: kvmx64,  alpine4k

### Checklist
- [x] Build rule `all-supported` completed successfully (lirc)

PR #3157, opened 2018-02-19
Fix digests creation for 6.1 kernels

hgy59:
_Motivation:_ "make digests" did not work with the DSM 6.1 kernel downloads
_Linked issues:_ kernels added with PR #3151, #3131 and #3099

### Checklist
- [n/a] Build rule `all-supported` completed successfully

### Notes
- update some digests that differ in whitespace or in uppercase
- add rules for kernel-clean and digests creation to main Makefile
- avoid duplicate kernel download for x64 and x86
- update kernel 5.2 digests since synology updated the files 2017-09-12


hgy59:
@Safihre `make digests` worked only in the folders with the Makefiles in the subfolders of cross/kernel/toolchains. Therefore I added the *-digests rules to the main Makefile. 
I did not add an all digest rule to the main Makefile, because it would download all files to distrib (with 90/110 spk packets it is currently ~72 GB with 431 files in 19 folders).

If you updates e.g. libev you cd to cross/libev and call make digests, or for a toolchain cd to toolchains/syno-broadwellnk-6.2 and call make digests.

The `make digests` is documented in the [developpers how to](https://github.com/SynoCommunity/spksrc/wiki/Developers-HOW-TO) under create a package (but for cross packages only).

All the toolchains have already digests files (as you are right, it is important). But when synology updates a toolchain file for an already existing arch and branch what shall we do but adjust the digests?
It is hard to find the diff of the new version and whether there is a need to update packages (all cross packages depend on toolchains).
The `spksrc.tc.mk` for toolchains is not changed here, because the download is simple and does not support different destination file names.

I have another 25 digests files of cross packages that have not yet such a file.
These are ready to add to the repo.

Does anybody know why digests use three different hash codes?
I suppose a SHA256 would be enough for integrity check. Additional SHA1 is useful too, since often provided by the download pages. But MD5 could easily be left away.
On my own branch I removed the MD5s, but reverted it for this PR and want to discuss it first.

**Is is OK to remove MD5 hashed from digests files?**

Safihre:
Thanks for that info, I read the documentation wrong about making digests.
I see I also was a bit confused, I meant why the `native/*` don't have any digests. Indeed the toolchains do have digests. I feel we should really add digests to `native/*` and add a `native-digests` rule? I can add that or maybe you want to do that as part of this PR? 

The MD5 seems historical due to [Sourceforge providing those](https://github.com/SynoCommunity/spksrc/commit/167d2485bc21d0d58e6c8ce8f3cf875396c38856), but indeed not really needed anymore. 
But they also don't cause any harm and it would also invalidate all PR's since "unknown" digest-lines cause the build to stop. So since it's so easy to create, why not just keep them? 
What do you think @ymartin59 ?

ymartin59:
@hgy59 @Safihre As there is no specific reason to remove md5 checksum from digests, I would let it there too. I also agree to have digests on `native` tools.

About your question concerning Synology toolchain updates... I really wonder if there is any change in them which may concern `spksrc` binary packages.

In fact, these toolchains provides Synology specific packaging tooling (https://developer.synology.com/developer-guide/create_package/build_and_create_package.html) which is not used by `spksrc`, as it only operates provided cross-compiler.

So my hypothesis is that updates only impact SDK python and shell scripts but not gcc compiler. That is why most toolchains are often updated all together.

Here is a small shell one-liner to generate again selected (here DSM >= 6) toolchain digests:
```
find toolchains -name "syno-*-6.*" -printf "\$(cd %p; make clean digests)\n" | bash
```

ymartin59:
A good test case would be to rebuild `lirc` which provides kernel module.

hgy59:
Source packages from github referenced by the ´PKG_GIT_HASH´ do have constant digests. Maybe there is a timestamp dependent content inside the tarball (?)
Therefore it would be better to reference all source packages by a version number and not by a commit/hash.

cross packages with git hash:
- comskip
- gateone
- rtmpdump
- tt-rss

The source tarball of the package cross/pear is not related to a version at all (go-pear.phar) but the digests do not change with every download. For traceability it would be better to have a versioned tarball too.

Safihre:
Seems we are stuck with these `PKG_GIT_HASH` ones then, since they don't really have a version to link usually (otherwise it would have been used).

Thanks for adding the new digests 👍

PR #3161, opened 2018-02-21
Add set_unix_permissions and syno_user_add_to_legacy_group

Safihre:
Support functions needed to:
- Correct permissions on folders during upgrade in Linux mode.
- For transfer for download-related packages from 'sc-media' to 'sc-download'

@ymartin59 This separate PR to keep package PR's clean. But all my other PR's kind of depend on it.
After this is merged, I can rebase #3092 #3153 #3053 on `master` after that so they are package-only.

### DIFF
--- a/mk/spksrc.service.installer
+++ b/mk/spksrc.service.installer
@@ -124,10 +124,15 @@ set_syno_permissions ()
 
     # Ensure directory resides in /volumeX before setting GROUP permissions
     if [ "`echo ${VOLUME} | cut -c2-7`" = "volume" ]; then
-        echo "Granting '${GROUP}' group permissions on ${DIRNAME}" >> ${INST_LOG}
-
         # Set read/write permissions for GROUP for folder and subfolders
         if [ ! "`synoacltool -get \"${DIRNAME}\"| grep \"group:${GROUP}:allow:rwxpdDaARWcC-:fd--\"`" ]; then
+            # First Unix permissions, but only if it's in Linux mode
+            if [ ! "`synoacltool -get \"${DIRNAME}\"| grep \"Linux mode\"`" ]; then
+                set_unix_permissions "${DIRNAME}"
+            fi
+
+            # Then fix the Synology permissions
+            echo "Granting '${GROUP}' group permissions on ${DIRNAME}" >> ${INST_LOG}
             synoacltool -add "${DIRNAME}" "group:${GROUP}:allow:rwxpdDaARWcC-:fd--" >> ${INST_LOG} 2>&1
             find "${DIRNAME}" -type d -exec synoacltool -enforce-inherit "{}" \; >> ${INST_LOG} 2>&1
         fi
@@ -144,6 +149,39 @@ set_syno_permissions ()
         echo "Skip granting '${GROUP}' group permissions on ${DIRNAME} as the directory does not reside in '/volumeX'. Set manually if needed." >> ${INST_LOG}
     fi
 }
+
+# Sets recurivse permissions using chown
+set_unix_permissions ()
+{
+    DIRNAME=$1
+    echo "Granting '${EFF_USER}' unix permissions on ${DIRNAME}" >> ${INST_LOG}
+    if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
+        chown -R ${EFF_USER}:root "${DIRNAME}" >> $INST_LOG 2>&1
+    else
+        chown -R ${EFF_USER}:${USER} "${DIRNAME}" >> $INST_LOG 2>&1
+    fi
+}
+
+# If package was moved to new group, we need to add the new package user
+# also to the old group. Only if the legacy user was in the old group.
+# Usage: syno_user_add_to_legacy_group "${NEW_USER}" "${LEGACY_USER}" "${LEGACY_GROUP}"
+syno_user_add_to_legacy_group () {
+    NEW_USER=$1
+    LEGACY_USER=$2
+    LEGACY_GROUP=$3
+
+    # Check if user in old group
+    if synogroup --get "$LEGACY_GROUP" | grep "^[0-9]:\[${LEGACY_USER}\]" &> /dev/null; then
+        # Add new user and remove old one
+        echo "Adding '${NEW_USER}' to '${LEGACY_GROUP}' for backwards compatibility" >> ${INST_LOG}
+        MEMBERS="$(synogroup --get $LEGACY_GROUP | grep '^[0-9]' | sed 's/.*\[\([^]]*\)].*/\1/' | tr '\n' ' ')"
+        MEMBERS=${MEMBERS//$LEGACY_USER}
+        # The "synogroup --member" command clears all users before adding new ones
+        # so all the users must be listed on the command line
+        synogroup --member "$LEGACY_GROUP" $MEMBERS "${NEW_USER}" >> ${INST_LOG}
+    fi
+}
+
 
 ### Generic package behaviors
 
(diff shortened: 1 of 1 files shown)


mk/spksrc.service.installer
```diff
@@ -124,10 +124,15 @@ set_syno_permissions ()
 
     # Ensure directory resides in /volumeX before setting GROUP permissions
     if [ "`echo ${VOLUME} | cut -c2-7`" = "volume" ]; then
-        echo "Granting '${GROUP}' group permissions on ${DIRNAME}" >> ${INST_LOG}
-
         # Set read/write permissions for GROUP for folder and subfolders
         if [ ! "`synoacltool -get \"${DIRNAME}\"| grep \"group:${GROUP}:allow:rwxpdDaARWcC-:fd--\"`" ]; then
+            # First Unix permissions, but only if it's in Linux mode
+            if [ ! "`synoacltool -get \"${DIRNAME}\"| grep \"Linux mocde\"`" ]; then
```
ymartin59: Looks like a typo: "mocde" !

mk/spksrc.service.installer
```diff
@@ -145,6 +150,39 @@ set_syno_permissions ()
...
+# also to the old group. Only if the legacy user was in the old group.
+# Usage: syno_user_add_to_legacy_group "${NEW_USER}" "${LEGACY_USER}" "${LEGACY_GROUP}"
+syno_user_add_to_legacy_group () {
+    NEW_USER=$1
+    LEGACY_USER=$2
+    LEGACY_GROUP=$3
+
+    # Check if user in old group
+    MEMBERS="$(synogroup --get $LEGACY_GROUP | grep '^[0-9]' | sed 's/.*\[\([^]]*\)].*/\1/' | tr '\n' ' ')"
+    if [[ $MEMBERS = *"$LEGACY_USER"* &&  $MEMBERS != *"$NEW_USER"* ]]; then
```
ymartin59: Hum. I would take care about such expression which may only run on `bash`... so only on DSM 6. I would check against on DSM 5.2 which runs `busybox` as default shell.

mk/spksrc.service.installer
```diff
@@ -145,6 +150,39 @@ set_syno_permissions ()
...
+        # Add new user and remove old one
+        echo "Adding '${NEW_USER}' to '${LEGACY_GROUP}' for backwards compatibility" >> ${INST_LOG}
+        MEMBERS=${MEMBERS//$LEGACY_USER}
```
ymartin59: Here again, probably requires `bash` too

mk/spksrc.service.installer
Safihre: So I came to this, because (at least on DSM61) this `if` statement in `postinst` always returns true and the user _always_ gets added even if already present:
```sh
if ! synogroup --get "$GROUP" | grep '^[0-9]:\[${EFF_USER}\]' &> /dev/null; then
```
So something is wrong with this syntax, but I couldn't figure out what. 

I just tested and indeed seems busybox does not support this syntax.. Will come up with a new one!

mk/spksrc.service.installer
Safihre: This one does seem to work:
image

mk/spksrc.service.installer
Safihre: Aah, my bash/shell-noobness that it took me an hour to figure out. It should be of course:
```sh
grep "^[0-9]:\[${EFF_USER}\]"
```
and not
```sh
grep '^[0-9]:\[${EFF_USER}\]'
```
In both commands.

Safihre:
Now it is working as expected.
Before it would add `svc-download` twice and not detect legacy membership correctly, now it does:
```
Step postinst. USER=nzbdrone GROUP=sc-download SHARE_PATH=
Installing service configuration /var/packages/nzbdrone/conf/nzbdrone.sc
Creating group sc-download
Group Name: [sc-download]
Group Type: [AUTH_LOCAL]
Group ID:   [65537]
Group Members:
0:[svc-nzbdrone]
Invoke service_postinst
Granting 'svc-nzbdrone' unix permissions on /volume1/@appstore/nzbdrone/var/.config
Adding 'svc-nzbdrone' to 'sc-media' for backwards compatibility
Group Name: [sc-media]
Group Type: [AUTH_LOCAL]
Group ID:   [65536]
Group Members:
0:[svc-nzbdrone]
Thr Feb 22 09:56:18 CET 2018
```

Safihre:
Just found out that `grep` on DSM 5.2 doesn't support the `-P`. Have to update all packages to some fancy `sed`. 😢

Safihre:
Removed Sonarr/Radarr specific things from this PR, creating separate one.

Safihre:
Another small bug: all code assumes `EFF_USER` is set. But for packages like Python it's not set since they don't need a user.
Introduced checks to all functions to only be performed if there actually is  `EFF_USER`.

ymartin59:
OK. Do you consider it as ready to merge ?

Safihre:
I would say so yes.
Have been testing all day today  on the 3 systems (Xpenology 5.2, Xpenology 6.1 and real 6.1 NAS) and also the "old style" package to "new style" upgrade paths.
That revealed all these problems that I fixed today also in the other PR's. 
What a hassle these packages are :|
Never imagined it to be so complex..

Safihre:
If this is merged I'll do a proper rebuild of _all_ the testing packages and test it again on all 3 systems.

m4tt075:
@ymartin59 @Diaoul Thanks a lot for inviting me to this elect club. More than happy to join, although I will consider myself as a "junior" member for the time being. @Safihre I have been on a business trip the last couple of days and am amazed about the progress you have made in the meantime. This is great, great work. Thanks so much! Looking forward to review "my" packages as well as once this is in!

mk/spksrc.service.installer
```diff
@@ -145,6 +154,41 @@ set_syno_permissions ()
...
 }
 
+# Sets recurivse permissions using chown
+set_unix_permissions ()
+{
+    DIRNAME=$1
+    if [ -n "${EFF_USER}" ]; then
+        echo "Granting '${EFF_USER}' unix permissions on ${DIRNAME}" >> ${INST_LOG}
+        if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
+            chown -R ${EFF_USER}:root "${DIRNAME}" >> $INST_LOG 2>&1
```
m4tt075: @Safihre @ymartin59 I have continued to debug the upgrade problems on DSM5.2 systems that I ran into yesterday (packages did not start anymore after upgrade due to permission issues). I suspect the culprit is that the group is set to `root` here rather than `users` for DSM lt 6. Do you agree? If so, we probaby have to re-build all packages which were built since.

ymartin59: Most test packages have only built for DSM 6... Group "users" is used for share folders accessible from other processes. For a package specific folder, group "root" should be far enough as owner matches service user.

Safihre: Agreed with @ymartin59, but unfortunately in the case of Tvheaded everything is mixed. Package and "user" folders blend together.

But @m4tt075, do you need one use `set_syno_permmissions` at all in your case? Since everything stays in the Package directory, do you need to call any permissions command at all?

ymartin59: Isn't really impossible to configure TVH download location to a share folder as other applications ? I think that may be the key. If download locations are "hardcoded", it is still possible to create symbolic link from expected/original locations to DSM share folder(s)

m4tt075: OK, thanks for clarifying. @Safihre I agree with you. Unix permissions need to be sufficient for all folders within `@appstore`. I have been able to implement that approach successfully for DSM6.1 now. I have tested clean installs, upgrading by itself and upgrading from the currently published package. But it does not work on DSM 5.2, when upgrading from the currently published package. In that version, ACL permissions were still set on the `target` and traversal folders. Is there a way to "remove" those ACL permissions again?

Safihre: @m4tt075 Why are ACL permissions applied? Only if `set_syno_permissions` is called this happens, but that shouldn't be needed at all, right?

PR #3163, opened 2018-02-22
Sonarr and Radarr fixes

Safihre:
Needs #3161, separated for clarity.
Uses the functions and changes there to make Sonarr/Radarr work on DSM6 but also 5.

### Checklist

### DIFF
--- a/spk/radarr/src/service-setup.sh
+++ b/spk/radarr/src/service-setup.sh
@@ -2,12 +2,17 @@ PATH="${SYNOPKG_PKGDEST}/bin:${PATH}"
 MONO_PATH="/usr/local/mono/bin"
 MONO="${MONO_PATH}/mono"
 
+# Check versions during upgrade
 RADARR="${SYNOPKG_PKGDEST}/share/Radarr/Radarr.exe"
 SPK_RADARR="${SYNOPKG_PKGINST_TEMP_DIR}/share/Radarr/Radarr.exe"
 
 # Radarr uses custom Config and PID directories
-CONFIG_DIR="${SYNOPKG_PKGDEST}/.config"
+HOME_DIR="${SYNOPKG_PKGDEST}/var"
+CONFIG_DIR="${SYNOPKG_PKGDEST}/var/.config"
 PID_FILE="${CONFIG_DIR}/Radarr/nzbdrone.pid"
+
+# Some have it stored in the root of package
+LEGACY_CONFIG_DIR="${SYNOPKG_PKGDEST}/.config"
 
 GROUP="sc-download"
 LEGACY_GROUP="sc-media"
@@ -16,7 +21,7 @@ service_prestart ()
 {
     # Replace generic service startup, run service as daemon
     echo "Starting Radarr as daemon under user ${EFF_USER} in group ${GROUP}" >> ${LOG_FILE}
-    COMMAND="env PATH=${MONO_PATH}:${PATH} LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib ${MONO} ${RADARR}"
+    COMMAND="env PATH=${MONO_PATH}:${PATH} HOME=${HOME_DIR} LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib ${MONO} ${RADARR}"
     echo "${COMMAND}" >> ${LOG_FILE}
 
     if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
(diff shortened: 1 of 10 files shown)


spk/sonarr/src/service-setup.sh
```diff
@@ -9,20 +9,15 @@ SPK_SONARR="${SYNOPKG_PKGINST_TEMP_DIR}/share/NzbDrone/NzbDrone.exe"
 CONFIG_DIR="${SYNOPKG_PKGDEST}/.config"
 PID_FILE="${CONFIG_DIR}/NzbDrone/nzbdrone.pid"
 
-GROUP="sc-media"
+# Sonarr uses the home directory to store it's ".config"
+# But there is a problem:
+# On DSM5 this is: PKG/var/.config/
+# On DSM6 this is: PKG/.config/
+# So we have to do upgrade magic
```
ymartin59: I do not like such magic. Isn't it possible to set `HOME` to `PKG/var/` in process environment at Sonarr startup with `env` for instance ? it is already done for other packages...

ymartin59: I do not like such complexity which is error prone and requires intensive testing.
Isn't it possible to set `HOME` to `PKG/var/` in process environment at Sonarr startup with `env` for instance ? it is already done for other packages...

spk/radarr/src/service-setup.sh
```diff
@@ -9,20 +9,15 @@ SPK_RADARR="${SYNOPKG_PKGINST_TEMP_DIR}/share/Radarr/Radarr.exe"
 CONFIG_DIR="${SYNOPKG_PKGDEST}/.config"
 PID_FILE="${CONFIG_DIR}/Radarr/nzbdrone.pid"
 
-GROUP="sc-media"
+# Radarr uses the home directory to store it's ".config"
+# But there is a problem:
+# On DSM5 this is: PKG/var/.config/
+# On DSM6 this is: PKG/.config/
+# So we have to do upgrade magic
```
ymartin59: Some request as for Sonarr. Please keep it simple and change `HOME` at process startup.

Safihre: Ah, I didn't know that was possible. I hope Sonarr listens to that variable. Will test.
Wonder why that approach wasn't taken before when they did this overhaul: https://github.com/SynoCommunity/spksrc/commit/5a0b6dabf2c86b4ea3ec50fb80aba1f93c9f7456

Safihre:
@ymartin59 Thanks for the hint. The `HOME` works well!

Safihre:
Checked on DSM 5.2 and DSM6.1:

- [x] Fresh install.
- [x] Upgrade from old-style-package.
- [x] Upgrade from older new-style-package (those in #3138).

💯 💯 🥇

spk/radarr/src/service-setup.sh
```diff
@@ -54,16 +50,15 @@ service_postinst ()
...
-    # We have to account for legacy folder under /var/
-    # It should go, after the upgrade, into /.config/
+    # We have to account for legacy folder in the root
+    # It should go, after the upgrade, into /var/.config/
     # The /var/ folder gets automatically copied by service-installer after this
-    # So we need to move our new /.config/ folder to /var/ (if present)
-    if [ -d "${CONFIG_DIR}" ]; then
-        echo "Moving ${CONFIG_DIR} to ${INST_VAR}" >> ${INST_LOG}
-        mv ${CONFIG_DIR} ${INST_VAR}/.config
+    if [ -d "${LEGACY_CONFIG_DIR}" ]; then
```
ymartin59: I still do not understand why there is some code to handle `.config` "move" whereas there is no longer "move" as its location has never changed between published package for DSM 5 in synocommunity repository.
May you explain ?

ymartin59: OK so I guessed right but that is not my point of view... beta/test packages are "as is"... that is why I avoid to publish when it is not really ready. It is far enough to "support" packages published in repository, not to add complexity to support/track "not ready" packages. Imagine I discard that useless code and let you explain users how to upgrade manually ?

ymartin59: I really think you are "pushing" too fast content, before it has been at least reviewed or tested by someone else. You merged a lot of PRs bulk... but be sure I will not publish any of them without reviewing again, and testing locally.

Safihre: > OK so I guessed right but that is not my point of view... beta/test packages are "as is"... that is why I avoid to publish when it is not really ready. It is far enough to "support" packages published in repository, not to add complexity to support/track "not ready" packages. Imagine I discard that useless code and let you explain users how to upgrade manually ?

They will get the update via DSM package center automatically, how do you want to instruct them to upgrade properly? In just the `sonarr` package has more than 400+ downloads among the different versions I released.

Safihre: Your comments here disappoint me. You seem to suggest that it was wrong of me to supply testing packages. That those testing packages were not really ready and that if their setup brakes because you remove this codd, it's kind of my fault. 
While in reality, without testing packages we wouldn't be able to identify problems with this and other  packages. 

And yes, I merge stuff, because I am tired of my code just being left there in those PRs. Yes of course I want you to review and  test them, but when will you actually do that? 
The code just sits there, after I have spend so much time on it. 
I have spend countless hours already since last November, the last 2 weeks even fulltime! And now you make it sound like I just add code without careful consideration. While in fact it was all my careful testing on all platforms that revealed so many previously unknown problems.
I respect your position to handle the publishing and I am glad you do it. So I hope you will also do the job of inspecting and testing my code, not just let it sit there for weeks waiting for.. Well I am not even sure what everything is constantly waiting for here.

ymartin59:
@Safihre You're right, I am aware of and I really appreciate your investment here. My point is: even if you feel confident  in your code and testing - you should wait for (at least) a first review to have chance to eliminate "first level" mistakes and get some hints (like the HOME variable trick). I am OK you build and publish test packages after such a first review.

A pending PR is not "lost work"... I even plan to resurrect interested closed PR that have never been merged like squidguard.

I have not as much bandwidth as you have to work on it, but we did great progress together... and I am relieved that my PR can be reviewed too. Again, there is no need to hurry, even if PR is merged, package will not be published "faster".

PR #3166, opened 2018-02-23
Headphones -  Generic Service approach

Safihre:
Simple Python packages. 
Issues: https://github.com/SynoCommunity/spksrc/issues?q=is%3Aissue+is%3Aopen+headphones

### Checklist
- [x] DSM 5.2 - Package upgrade completed successfully
- [x] DSM 6.1 - Package upgrade completed successfully
- [x] DSM 5.2 - New installation of package completed successfully
- [x] DSM 6.1 - New installation of package completed successfully


ymartin59:
Would it be possible to improve `headphone` to support custom selection in wizard... so that to merge both package into `headphone`.

Here is a concept:
- `headphone-custom` declares `headphone` as package dependency (so it is installed at upgrade)
- wizard inform it is last update version of the package
- wizard ask for automatic migration and propose user to uninstall obsolete `headphone-custom` package
- if answer for automated migration is yes, installer postupgrade step move configuration and other package specific files into `headphone` package proper locations

What do you think about it ?

Safihre:
That would definitely be possible. But these and similar packages (`sickbeard-custom`, `couchpotatoserver-custom`, maybe more) won't really need updates anymore after this version, since they self-update from within the apps. For  these packages the previous release was years and years ago and the user-base relatively small, is it worth adding complexity for such a mini-packages?

ymartin59:
You're right... my motivation is to clean up package list visible at https://synocommunity.com/packages (also to avoid user mistake - and also to reduce number of packages to publish/maintain). I agree about "complexity" argument but if migration script from "custom" (or "-testing") to "standard" is easy and reusable over packages, it may worth the work.

PR #3168, opened 2018-02-24
lame: fixed download URL

yuezhu:
The sourceforge which hosts lame project has been down for several days, which blocks build for lots of packages. The package lame is also available at https://download.videolan.org/contrib/lame/.


yuezhu:
The sourceforge download URL works again, so this change is not needed for now.

Issue #3124, opened 2018-01-24
Firefox Sync Server doesn't run : files missing after install


MysteryWoMan:
### Setup
_Package Name: ffsync_
_Package Version: 1.5.2_

_NAS Model: 415+_
_NAS Architecture: x86_64 (avoton)_
_DSM version: 6.1.4-15217 Update 5_

### Expected behavior
all files installed and ffsync can be run

### Actual behavior
files are missing and ffsync cannot be run

### Steps to reproduce
_1. install ffsync from synology package center in the DSM webinterface_
_2. run it_

#### Package log
_/usr/local/ffsync/var/install.log_
```
Ignoring indexes: https://pypi.python.org/simple
Collecting cef (from -r /usr/local/ffsync/share/wheelhouse/requirements.txt (line 5))
  Could not find any downloads that satisfy the requirement cef (from -r /usr/local/ffsync/share/wheelhouse/requirements
.txt (line 5))
```
- source url from installer not configured correctly? I think the installation already stops/aborts there

#### Other logs
_/var/log/synopkg.log_
```
2018/01/23 22:41:05     install ffsync 1.5.2-3 Begin postinst
unable to write 'random state'
2018/01/23 22:41:12     install ffsync 1.5.2-3 End postinst ret=[0]
```
- Probably this comes from `installer.sh` --> `postinst ()` --> `Line 66`
When i run `openssl rand -base64 20` (in real environment) as any user, no error is thrown. 

```
2018/01/23 22:41:13     install ffsync 1.5.2-3 Begin start-stop-status start
Starting Firefox Sync Server 1.5 ...
/var/packages/ffsync/scripts/start-stop-status: line 19: /usr/local/ffsync/env/bin/pserve: No such file or directory
2018/01/23 22:41:14     install ffsync 1.5.2-3 End start-stop-status start ret=[0]
2018/01/23 22:41:14     install ffsync: start version 1.5.2-3 successfully, result 0
```

- Last lines of ffsync/share/wheelhouse/requirements.txt are:
`mozsvc==0.8
tokenserver==1.2.7
SyncStorage==1.5.5
syncserver==1.5.2`
~~This should be updated with _requirements.txt_ from the github repository as these packages are not on _https://pypi.python.org/simple/_ anymore.~~ 
_UPDATE: Makefile changes this at the end, so there is nothing wrong with requirements.txt_

**Second try** : Download github repository and install
```===>  Processing dependencies of ffsync
[...]

Issue #2873, opened 2017-08-15
Can't uninstall ffsync


ghost:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ Firefox sync Server 1.5
_Package Version:_

_NAS Model:_ ds414
_NAS Architecture:_
_DSM version:_6.1.3 update 1

### Expected behavior
Can't unistall firefox sync server , can't make run firefox sync server. When try unistalling say error mysql passoword. i try reset it in mariadb,  try make it new one ... problem persist.
 I try look online for problem but some post don't solve me.
Thank's

ymartin59:
Known issue with some DSM 5.2 packages after upgrade of DSM to 6.x
From my point of view, this will wait package updates to both DSM 5.2 and DSM 6.1 when available to do so.

ioio123:
Thank's for answer,

so i have to wait?, little annoying have service that use resource of nas and can't unistall.

Thankl's.


??
??
??
??
Questo messaggio e-mail ?? riservato unicamente al/ai soggetto/i destinatario/i e pu?? contenere informazioni riservate e/o coperte da segreto professionale utilizzabili esclusivamente dal/i destinatario/i sopraindicato/i. Il ricevente, se diverso dal destinatario, ?? avvertito che qualsiasi utilizzazione, divulgazione o copia di questa comunicazione, nonch?? di eventuali suoi allegati, comporta la violazione delle disposizioni di Legge sulla tutela dei dati personali (D.Lgs 196/2003). Qualora il messaggio fosse ricevuto per errore, si prega di cancellarlo e distruggere eventuali copie o stampe e di darne informazione immediata al mittente. Grazie.


________________________________
Da: Yves Martin <notifications@github.com>
Inviato: martedì 15 agosto 2017 18:49
A: SynoCommunity/spksrc
Cc: ioio123; Author
Oggetto: Re: [SynoCommunity/spksrc] Can't unistall (#2873)


Known issue with some DSM 5.2 packages after upgrade of DSM to 6.x
From my point of view, this will wait package updates to both DSM 5.2 and DSM 6.1 when available to do so.

eddieirvine:
I just bought a RasPi and will install the sync server there.
I'm fed up of waiting for a package update.

Fx Sync server in the Synology package is out of date for Android devices:
https://bugzilla.mozilla.org/show_bug.cgi?id=1382666

ymartin59:
[...]

Issue #2828, opened 2017-07-17
Compile ffsync


Martijn085:
I can't compile ffsync anymore.

I run this command from the folder: **/spksrc/spk/ffsync** : **make arch=x86-6.0.2**

This is the error message i get:

`===>  Compiling for bzip2
cd /spksrc/spk/ffsync/work/bzip2-1.0.6 && env PKG_CONFIG_LIBDIR=/spksrc/spk/ffsync/work/install//usr/local/ffsync/lib/pkgconfig WORK_DIR=/spksrc/spk/ffsync/work INSTALL_PREFIX=/usr/local/ffsync make bzip2 bzip2recover PREFIX=/spksrc/spk/ffsync/work/install/usr/local/ffsync
make[3]: Map '/spksrc/spk/ffsync/work/bzip2-1.0.6' wordt binnengegaan
make[3]: 'bzip2' is up-to-date.
make[3]: 'bzip2recover' is up-to-date.
make[3]: Map '/spksrc/spk/ffsync/work/bzip2-1.0.6' wordt verlaten
cd /spksrc/spk/ffsync/work/bzip2-1.0.6 && env PKG_CONFIG_LIBDIR=/spksrc/spk/ffsync/work/install//usr/local/ffsync/lib/pkgconfig WORK_DIR=/spksrc/spk/ffsync/work INSTALL_PREFIX=/usr/local/ffsync make -f Makefile-libbz2_so PREFIX=/spksrc/spk/ffsync/work/install/usr/local/ffsync
make[3]: Map '/spksrc/spk/ffsync/work/bzip2-1.0.6' wordt binnengegaan
cc -shared -Wl,-soname -Wl,libbz2.so.1.0 -o libbz2.so.1.0.6 blocksort.o huffman.o crctable.o randtable.o compress.o decompress.o bzlib.o
/usr/bin/ld: blocksort.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
blocksort.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
Makefile-libbz2_so:32: recept voor doel 'all' is mislukt
make[3]: *** [all] Fout 1
make[3]: Map '/spksrc/spk/ffsync/work/bzip2-1.0.6' wordt verlaten
Makefile:24: recept voor doel 'myCompile' is mislukt
make[2]: *** [myCompile] Fout 2
[...]

PR #3170, opened 2018-02-24
fix ffsync build and wizard

hgy59:
_Motivation:_ ffsync does not build
_Linked issues:_ This PR does not fix install/uninstall ffsync (#3124, #2873). It maybe related to #2828

Before I will analyze the installation errors (apollolake-6.1) this fixes the build.

### DIFF
--- a/cross/libev/Makefile
+++ b/cross/libev/Makefile
@@ -2,7 +2,7 @@ PKG_NAME = libev
 PKG_VERS = 4.22
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = http://dist.schmorp.de/$(PKG_NAME)
+PKG_DIST_SITE = http://dist.schmorp.de/$(PKG_NAME)/Attic
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 DEPENDS =
--- a/spk/ffsync/Makefile
+++ b/spk/ffsync/Makefile
@@ -19,6 +19,8 @@ HOMEPAGE     = https://docs.services.mozilla.com/howtos/run-sync-1.5.html
 LICENSE      = MPL/GPL/LGPL
 HELPURL      = https://github.com/SynoCommunity/spksrc/wiki/Firefox-Sync-Server-1.5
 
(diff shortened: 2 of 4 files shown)


hgy59:
In Makefile ffsync_extra_install the $(STAGING_DIR)/share/wheelhouse/requirements.txt is patched (this file was missing).
Since I don't understand the wheelhouse yet, can please someone check, whether the requirements.txt file in the staging dir is used for building the wheel...?

hgy59:
And this is the solution for #2873
The src/wizard directory was removed from Makefile 2017/02/01. Therefore no wizard is displayed neither for install nor uninstall.
Without the wizard the mysql root password is always empty.

Update works without wizard, but a new install was not possible. To uninstall a ffsync that has no wizard, you can add the missing password in installer.sh like this:
`wizard_mysql_password_root=your_mysql_root_pw_here`

installer.sh is located in /var/packages/ffsync/scripts

OR  (not veryfied)
you can copy the missing wizard files to
/var/packages/ffsync/WIZARD_UIFILES

Safihre:
@hgy59 thanks for this :) Will investigate the `requirements.txt` but pretty sure it's not used from staging but from the original file.

But I am pretty sure @ymartin59 won't accept it if it doesn't get converted to Generic Service package. But that's a relatively easy transition :)

hgy59:
There is still an issue in ffsync.ini. The user for mysql is hard coded as ffsync, but installer.sh configures ffsync for build numbers < 7321 and sc-ffsync for newer.
Hope this will be easy with generic service implementation.

Safihre:
@hgy59 Yes you can use `$EFF_USER` which contains the right user for the platform it's being installed on
 👍

hgy59:
@Safihre already tried this, but `$(EFF_USER)` was empty

Safihre:
It's not available in the `Makefile` no, only as part of the service-scripts in `src/service-setup.sh`.

Safihre:
Can always push work-in-progress or link to working-branch and we can see what's going wrong :)

Safihre:
@hgy59 how is this going? :)

hgy59:
Installed the existing package some time ago, but it did not work.
And without a working installation it is boring to test migration (and it will still be some work to migrate to the new service approach)

hgy59:
outdated due #3849

Issue #49, opened 2012-03-23
Allow parallel make execution


Diaoul:
This is possible with the make -j option.
I tried but it fails at some points of the code because not all targets can be parallelized.

I see two solutions to that:
- Build our own parallel stuff (by adding PARALLEL=1 to local.mk for example) and make the code parallel ourselves in the code
- Use the builtin -j option that comes with `make` but we have to specify .NOTPARALLEL where needed

I recommend the second option, that's the "standard" way.
##

ymartin59:
Even if "make" sequence for each package/architecture cannot be parallelized easily, I find it possible/interesting to parallelize "per architecture" build to speed up at least `make all-supported`.

Dr-Bean:
Parallel make is supported with all rules (iirc anyway), _except_ `make all-supported` ;) Unless you know a way to check if `REQ_KERNEL` is set in dependencies and filtering out a subset of arches based off of that without a loop.

Safihre:
Seems not to be super applicable anymore. We tried in #3171, but it doesn't work well for all packages.

PR #3171, opened 2018-02-25
Add support for parallel make compilation

ymartin59:
_Motivation:_ when building package for a single architecture, top level `-jX` option is useless
_Linked issues:_ #49 

### Tests
- [X] FFmpeg
- [X] Mono
- [ ] Python
- [ ] ...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,6 +126,7 @@ local.mk:
 	@echo "DISTRIBUTOR_URL=" >> $@
 	@echo "REPORT_URL=" >> $@
 	@echo "DEFAULT_TC=" >> $@
+	@echo "#PARALLEL_MAKE=max" >> $@
 
 dsm-%: local.mk
 	@echo "Setting default toolchain version to DSM-$*"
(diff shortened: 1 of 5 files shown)


ymartin59:
@Safihre What do you think about it ? Are you aware of packages to be tested with that support ?

Safihre:
I was thinking something similar  when I was compiling Python over and over again.
But then I read about how many programs cannot handle it, that I abandoned it.
I just tested a simplified version of this PR for Python: https://github.com/Safihre/spksrc/commit/87ed05a43b8604180d05459af4f5f681a5fdc76d#diff-bdafcb533e01bd848bdff3d9f857a58d

But it seems (at least on Travis, which gives you access to 2 cores) not to matter, build times showed similar lengths and variations as before. 
Did I not incorporate it properly?

ymartin59:
@Safihre In fact my opinion is that Python build chain prevents parallel make option to operate...

Safihre:
That's probably true. I noticed already that OpenSSL forces `-j1`, so not much we can gain there.

Safihre:
I would suggest the following:
Use a `cat /proc/cpuinfo` to get the number of cores automatically. 
Then have a `Makefile` variable that triggers the `-j` to the `make` command. For example `PARALLEL_MAKE=1`.

ymartin59:
Documented as last section of https://github.com/SynoCommunity/spksrc/wiki/Compile-and-build-rules

PR #3172, opened 2018-02-25
improve icon generation

hgy59:
_Motivation:_ Each time a package is built, it is binary different.

I figured out, that the only difference in the packages is the content of the icons generated with convert.
The icons differ in metadata only (creation datetime).
With the option -strip the metadata is removed from the images. Unfortunately this works with newer versions of convert only.
The current spksrc Docker image based on jessie comes with convert based on imagemagick 6.8.9-9 and this version does not strip.
With an updated Docker image based on stretch and with convert based on imagemagick 6.9.7 icons are stripped as expected.

### DIFF
--- a/mk/spksrc.icon.mk
+++ b/mk/spksrc.icon.mk
@@ -42,7 +42,7 @@ pre_icon_target: icon_msg
 $(ICON_DIR): $(PRE_ICON_TARGET)
 	@mkdir -p $@
 	@for size in 16 24 32 48 64 72 256; do \
-	  convert $(SPK_ICON) -thumbnail $${size}x$${size} \
+	  convert $(SPK_ICON) -thumbnail $${size}x$${size} -strip \
 	          $@/$(SPK_NAME)-$${size}.png ; \
 	done ; \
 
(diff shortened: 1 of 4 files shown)


ymartin59:
@hgy59 May you please explain motivation to "fix" this repeatable packaging constraint you expects ?
Please also include `Dockerfile` changes to switch to stretch or to include imagemagick backport to jessie (in case toolchains may not work properly on stretch)

Safihre:
Shouldn't the Icon Cookie stop the regeneration of icons? Only when you do a `make clean` this is removed, but when you do that im sure more timestamps change?

hgy59:
When working on different branches it is very usefull to (binary) compare created packages to find unexpected modifications.
E.g to compare packages built with original jessie Dockerfile vs. stretch-slim.

Safihre:
@hgy59 how about this PR? I actually don't know if we can update the Docker image, seeing the amount of problems we were getting on Travis when not using the Docker.
But the stripping won't work without it?

hgy59:
This is more or less an investment for the future.
Still working on the Dockerfile.
And there are a lot more differences in the spk than the icons (*.a, *.whl, ...) ;-(

hgy59:
@ymartin59 with the docker image now based on debian-stretch this PR can be merged.

verified:

```
convert --version
Version: ImageMagick 6.9.7-4 Q16 x86_64 20170114 http://www.imagemagick.org
```

PR #3173, opened 2018-02-26
New package cleanup

Safihre:
I used the great script by @GuillaumeSmaha that scans all [dependencies](https://gist.github.com/GuillaumeSmaha/5545f6e48b5d3ce35399d6b3727c4a9c#file-not_used_packages-sh). 
It has some false-positives mostly due to `cross/$(PKG_NAME)` references, so I manually checked them all. 
Here is the first set of packages, I could do many more but I want to keep it possible to do a good review (@ymartin59)! I can do a second PR with the rest of the packages after this one is merged.
Replaces #2658.

Would like to note that I specifically did not include `cross/pocketsphinx` and related packages because these were removed from `spk/python` in the recent update (saved ~30MB per download). So they might need to be re-added to other packages later on.

This is the whole output of the script:
```
cross/alsa-utils
cross/audiofile
cross/bison
cross/borgbackup
cross/coreutils
cross/dobby
cross/duplicity
cross/enum34
cross/espeak
cross/flex
cross/flite
cross/gc
cross/gdb
cross/gevent
cross/greenlet
cross/gsasl
cross/guile
cross/jemalloc
cross/julius
cross/libev
cross/libid3tag
cross/libidn
cross/libmad
cross/libpar2
cross/libsndfile
cross/libunistring
cross/linux-pam
cross/lzo
cross/mercurial
cross/mpd
cross/newznab
cross/openldap
cross/openssh
cross/pear
cross/pocketsphinx
cross/polarssl
cross/rdiff-backup
cross/sphinxbase
cross/termcap
cross/udns
cross/umemcache
cross/virtualenv
cross/yenc
native/guile
```


Safihre:
Added back `cross/gdb` as it should become part of `CLI Dev Tools`, discussed [here](https://github.com/SynoCommunity/spksrc/issues/2261#issuecomment-334773089).

ymartin59:
`cross/jemalloc` is also a development tool to diagnose native memory leak as an alternative allocator.

ymartin59:
Here are pending PRs with `borgbackup` and `duplicity` package updates...

ymartin59:
I remember there is a python wheel `unistring` that depends on `libunistring`. Probably there are other libraries in that list that are built as wheel dependencies for some packages.

ymartin59:
From `spk/mercurial`, I guess script is not able to handle Makefile expression like `cross/$(SPK_NAME)`... why `cross/mercurial` has been triggered for removal.

Safihre:
:+1: will remove and rebase. I'm not in a hurry to remove packages, I just noticed indeed those Python packages like `virtualenv` and `enum34` that are now handled via `requirements.txt`.

Safihre:
@ymartin59 as mentioned in the first post, I didn't actually remove all the packages that were returned by the script. Only a subset which I really couldn't find a link to. So the packages listed in your comments I did not remove.
Of the packages that I actually removed in this PR, which ones should stay?

vkostas:
@Safihre I would keep cross/mpd (working on it) and its dependent packages, cross/alsa-utils and cross/audiofile.
Still no big deal if you remove. I will submit a new version when finished.

Safihre:
Thanks for reporting, will remove them from this PR!

cytec:
i'd keep `cross/coreutils` for the `CLI Package` you mentioned as well not sure about `libid3tag` thought that would be used somewhere? same goes for `virtualenv` `libpar2`, `yenc` aren't they used by python?

pretty sure there is a `borgbackup` Package... or at least there was a while ago, same for `duplicity` so i'm not sure about those too

Generally speaking: as long as they build and work properly... why remove them? maybe someday someone needs them and can those as they are... or as starting point for updates etc? I get that we want to save space for the repo and SPK stuff, and it might be a little bit cleaner to only have `cross/` packages that we really do use...

vkostas:
_libid3tag_ might be used in _MPD_. Still not big issue if deleted.

Safihre:
@cytec `yenc` was used by SAB, but now replaced by SABYenc. 
`virtualenv` (and `enum34`) is now done via `requirements.txt`.
`libpar2` was used by `par2cmdline` but hasn't in a long time, hasn't been updated either since 2006.

ymartin59:
@Safihre I just discovered your PR branches come from spksrc main repository itself instead of your personal fork. Please care about it, to avoid to show branches publicly here.

Safihre:
Yes this is on purpose, since it's a repo-related PR and not a package update.

Safihre:
@ymartin59 how do you feel about merging this one? I went through all PR's and the packages in this PR seem really unreferenced.
Follow up PR's can be made for other packages on that initial list in the future.

Issue #2974, opened 2017-10-24
Update sslh to version 1.18


Yasston:
It would be great if possible, version 1.18 contains very useful features like tls and SNI options.

Thank you very much

Safihre:
@Yasston Are you still interested in this?
I updated the package, you can get it from here #3138.

ymartin59:
Published on repository

PR #3178, opened 2018-02-26
sslh - Generic Service approach and update to 1.18

Safihre:
Very simple package, update was requested in #2974.

### Checklist


ymartin59:
@Safihre How does service writes its `PID_FILE` here ? If there is no support to do it, option may be to use `SVC_BACKGROUND=y` and `SVC_WRITE_PID=y`, as in https://github.com/SynoCommunity/spksrc/commit/8e9fb5883077939558e6a9d93266db5613f2516f#diff-786e0e526428f5c7eff2293096e26429R5

Safihre:
It write's a PID file based on the location specified `sslh.cfg` (which we provide). So it's supported and working.

PR #3180, opened 2018-02-27
NZBGet-testing - Generic Service approach

Safihre:
@ymartin59 NZBGet is the simplest packages to test really Generic Service and the functions to transfer permissions.
I am getting more and more frustrated that my code for popular packages just sits here on the repo for weeks without being published. 
What is it waiting for? You must have some time to test some of _my_ packages?
If you prefer to do a smaller test first I suggest to release `nzbget-testing` which is in this PR. It's the same code, just different port and "Testing-release" selected by default in wizard.

### Checklist
- [x] Build rule `all-supported` [completed successfully](https://travis-ci.org/Safihre/spksrc/builds/346665557)


ymartin59:
And please stop rating about time. With full time job, wife, two children and house... I do my best to progress here.

Safihre:
We could obsolete the package, but especially now the package can be updated from within NZBGet itself, why not just keep it? We won't have to update/publish new one after this one anymore really.
Based on the experiences/feedback in #3138 people seem to skip over the permissions information so I doubt they would read anything else we put :/

ymartin59:
@Safihre so if they do not read, process should be fully automated. Package description should be replaced by "Obsolete package. Replaced by X. Should be removed". Package should no longer show application link, log or "running" status (always stopped). Some kind of "black hole" package setup. After few months, it can be deactivated from repository

Safihre:
I just don't see the usefulness in making that effort for packages that will most likely never really be updated again anyway.
Yes for SABnzbd-testing this could be relevant, but I just don't see it for NZBGet and these Python-packages..
 Just to shorten the list on the website?

Safihre:
It's really simple, just download the version for your architecture here #3138 and update by clicking Manual Upload in Package Center. It will keep your settings (if you don't switch between nzbget and the nzbget-testing packages).
If you want to switch between the packages, you indeed need to backup Config first.

Donkeyfumbler:
That simple? OK, sounds good - just your comment about people skipping over the permissions information earlier in this thread made me think there was something else that might be needed and I also read this thread in my travels - https://forum.nzbget.net/viewtopic.php?f=3&t=3049&start=10 - that made it sound as though some paths had changed between versions as well as making it all look quite complex. You've sorted all of that out though in the SPK I presume then?

Safihre:
The new installer package now does exactly what Hugbug suggests to do: it downloads the official installer of nzbget and installs it. This installer also determines optimal settings for the platform (only for new installations, not upgrades of the Synology package). 
Now if you do an upgrade (so not when switching packages!!) it will also detect the path you have set in nzbget as download path and apply the right new permissions to it. These permissions mean that everything should now be done via the `sc-download` group. All other packages (like Sonarr) will also use that group from now on.

Donkeyfumbler:
Perfect. Just updated and everything seems to be working fine and all my settings and history are still there. From what I understand then, it is possible to update to newer versions, when released, through the NZBget web interface itself?

I'm still confused though :) - if loads of people have confirmed it works and it updates the existing DSM package with a single click, why is it not published as an update in DSM? Is there some kind of disagreement on whether to create it as a totally new package rather than as an update to the existing one? Seems a shame to get stuck on that when the work has been done and everything is ready to go.

Many thanks for updating this (and others by the looks of it). Thanks also for your work on Sabnzbd, which I use on the PC (the DS213J is a bit underpowered for it I found though). Really appreciate as well as the help here.

Safihre:
> Perfect. Just updated and everything seems to be working fine and all my settings and history are still there. From what I understand then, it is possible to update to newer versions, when released, through the NZBget web interface itself?

Correct 👍 

> I'm still confused though :) - if loads of people have confirmed it works and it updates the existing DSM package with a single click, why is it not published as an update in DSM? Is there some kind of disagreement on whether to create it as a totally new package rather than as an update to the existing one? Seems a shame to get stuck on that when the work has been done and everything is ready to go.

@ymartin59 makes the decisions about what and when to publish.

Donkeyfumbler:
Just looking at what changes have been made to users and groups by this update, I can see that the user account has been changed from nzbget to sc-nzbget. I can also see this user is still in the sc-download group. However I can also see that a group called nzbget appears to have been created but has no members in it - is that normal? To be honest, being a windows person, the whole linux permissions thing still mystifies me but I'd really like to understand it better. Everything is working fine by the way - this is me just trying to peer behind the curtain to gain a better understanding.

Safihre:
@ymartin59 Could you maybe tell why you don't want to publish NZBGet (testing and/or normal)?
What is the blocking issue? The permissions?
This is not meant to offend, I am just trying to understand 👍

ymartin59:
@Safihre Now `SHARE_PATH` permission with `sc-download` is fixed (in `transmission`), I am publishing `nzbget`... For next version, in addition to `busybox` removal, moving  `p7zip` and `unrar` to a "command line" package would probably convert this package into a simple download-and-configure script.

ymartin59:
@Safihre Forget a question/confirmation. There is no need for service user of `nzbget` (like `transmission`) to be member of `users` group, as they only produce content in `SHARE_PATH` - and do not consume content from "anywhere". Am I correct?

Safihre:
That seems to be correct, indeed no consumption.
If used with updated Sonarr/Radarr/CouchPotato/etc which are part of `users` AND `sc-download` of course.

PR #3182, opened 2018-02-27
Mylar - Generic Service approach

Safihre:
Last one I think.

### Checklist


ymartin59:
@Safihre Do you think there is a reason Mylar requires Python upgrade or it may run on old version?

Safihre:
It worked for me on old and new Python. Shouldn't make a difference in any platform :)

PR #3183, opened 2018-02-28
Correct logging redirection

Safihre:
Due to #3138 I noticed not all packages have correct `LOG_FILE` set and as such users can't get the log from Package Center.
@ymartin59 This includes Transmission, maybe this can go in before you publish?

### DIFF
--- a/mk/spksrc.service.start-stop-daemon
+++ b/mk/spksrc.service.start-stop-daemon
@@ -54,6 +54,7 @@ wait_for_status ()
     return 1
 }
 
+
 case $1 in
     start)
         if daemon_status; then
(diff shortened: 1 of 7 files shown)


Teryon:
NZBGET Logs in packages center crash DSM

Safihre:
I added new code that will only show the last 100 lines in Package Center and the path to the full log. 
This makes more sense, as the main purpose is to identify start-up problems and not really to track full program-functioning.
What do you think @ymartin59 ?

Teryon:
@Safihre When I click to show the log, it shows 'loading" nothing happens and it get stuck and sometime DSM reload itself.

ymartin59:
As it is difficult to tune log verbosity for each application, I agree your code is a relevant work-around. And full log remains available too. Does Synology has documented some restriction about this package log file volume ?

PR #3186, opened 2018-03-01
WIP Service: enable start-stop-daemon support

ymartin59:
Enable busybox start-stop-daemon generic support

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -55,8 +55,10 @@ endif
 # Recommend explicit STARTABLE=no
 ifeq ($(strip $(SSS_SCRIPT)),)
 ifeq ($(strip $(SERVICE_COMMAND)),)
+ifeq ($(strip $(SERVICE_EXE)),)
 ifeq ($(strip $(STARTABLE)),)
 $(error Set STARTABLE=no or provide either SERVICE_COMMAND or specific SSS_SCRIPT)
+endif
 endif
 endif
 endif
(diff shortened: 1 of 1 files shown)


Safihre:
I'll give it a try with Syncthing :+1:

mk/spksrc.service.mk
```diff
@@ -131,10 +131,15 @@ ifeq ($(STARTABLE),no)
 $(DSM_SCRIPTS_DIR)/start-stop-status: $(SPKSRC_MK)spksrc.service.non-startable
 	@$(dsm_script_copy)
 else
+ifneq ($(strip $(SERVICE_EXE)),)
```
Safihre: So this only works if `SERVICE_EXE` is defined in the package's `Makefile`. If it's in `service-setup.sh` then this variable is empty since that file and the variables not available when compiling.

Similarly when I just started Generic Service it took me a longgg time to understand why I kept getting `Set STARTABLE=no or provide either SERVICE_COMMAND or specific SSS_SCRIPT`, until I realized I was forced to put `STARTABLE=yes` in the `Makefile` for it to skip this error.

Is that the intended way or a design-flaw?

ymartin59: You're right, this error message should not be thrown. I will fix in my WIP PR.

As `SERVICE_EXE` is supposed to be absolute path to executable, I do not think it is relevant to set it in `service-setup.sh`. You have multiple options: either use symbolic links like `/var/packages/PACKAGE/bin/exe` or `/usr/local/PACKAGE/bin/exe`. If you prefer to "generate" it from package location, this should work too: `SERVICE_EXE=$${SYNOPKG_PKGDEST}/bin/exe`

Safihre: Aah, so don't want to keep it all in `service-setup.sh`? Since so far we have been doing that for all packages. For me it's all good, just wanted to know 😉

ymartin59: Yes I know... but I have to design some kind of "API" so that build chain is able to select "start-stop-status" base script... If you have a better proposal, do not hesitate.

Safihre: How about a `SERVICE_DAEMON=1` for the `Makefile` and then keep the rest in the `service-setup.sh`?

Safihre: In that regard, `STARTABLE` in `Makefile` maybe isn't that bad after all!

ymartin59: A single switch like `SERVICE_DAEMON` will not inform `Makefile` about which `start-stop-status` script to use...  `mk/spksrc.service.start-stop-daemon` or generic `mk/spksrc.service.start-stop-status` or maybe/probably yet another future alternative. I will think about it and propose something here

PR #3189, opened 2018-03-01
added tor spk

cytec:
_Motivation:_ replaces https://github.com/SynoCommunity/spksrc/pull/1659

### Checklist

@ymartin59 @Safihre maybe one of you can check if i did the generic service part correctly?


spk/tor/src/service-setup.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+# Package
+PACKAGE="tor"
+DNAME="Tor"
```
Safihre: The service-part looks good!
Only these first 6 lines are not needed and can be removed!
I didn't compile it yet, but I can start a Travis-build and add it to #3138 if you like?

ymartin59:
I have found in other package (probably `owncloud` if I remember well) a specific `uninstall` process: package generates a backup at expected location and wizard message inform where backup can be found.
It may be a better user experience for tor too.

spk/tor/src/service-setup.sh
cytec: woops. yeah those should be gone.

I

$USER is still the right user? was a little confused with $USER and $EFF_USER \cc @ymartin59

cytec:
i think there are some packages which do this. Most likely those who use a MySQL database which is backed up at uninstall. For tor i'm not 100% sure what files where exactly stored by tor in the var folder (has been some time since i've tested this) but i think it was just config and some ID File which identifies the node.

The warning imho could be deleted completely as EVERY SPK removes it's configuration on uninstall. I'm not really sure why it was added in the first place but there must have been a reason...

Safihre:
Seems indeed bit unnecessary to have this uninstall note.
I saw there is also `geoip` in the PLIST, does that need the `cross/geoip`?

cytec:
nope. afaik tor downloads it's own geoip database stuff on build.

Safihre:
Triggered a build of the package for all archs, will test tonight.
How do I test this actually? Just run it and see if there's no errors?

cytec:
most likely. i guess the easiest way is to install a bridge/proxy and try setting it up as a proxy in chrome or something else

Safihre:
@cytec something is wrong, all except the `x64`toolchains fail.
https://travis-ci.org/Safihre/spksrc/builds/349162017  (ignore commit-message, it's really compiling tor)

cytec:
strange... i didn't test anything but cedarview on my box since downloading the toolchains didn't work correctly... i'll have a look asap

cytec:
@alexey-pimenov would be great! i dont have a lot of time lately and my buildbox went to hell again -.-

noplanman:
I know this is quite an old PR, but I'm just checking in to see what the progress is on this :blush:

Issue #3185, opened 2018-02-28
Permission issue while updating Sonarr


jelmerj:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ Sonarr
_Package Version:_ 201802280-12

_NAS Model:_ DS214play
_NAS Architecture:_ 
_DSM version:_ 6.1

### Expected behavior
Sonarr advertised an update. I expected it to update without problems.

### Actual behavior
The update failed with this error: Error occurred while executing task ApplicationUpdate: Access to the path "/tmp/nzbdrone_backup/config.xml" is denied.

Output of ls -la /tmp at the time:
drwxr-xr-x  2         100 users       100 Feb 27 19:58 nzbdrone_backup
drwxr-xr-x  3 sc-nzbdrone nzbdrone     80 Feb 28 21:00 nzbdrone_update

As you can see the user/group of nzbdrone_backup are incorrect. I decided to just remove the dir and have sonarr recreate it itself. That worked like a charm and the update was succesful afterwards.

I'm not sure if the package handles the permissions on this directory? 
The same appears to be true for radarr:

drwxr-xr-x  2 radarr      users        80 Feb 27 20:10 radarr_backup
drwxr-xr-x  5 radarr      users       480 Feb 27 20:11 radarr_update

As you can the the dirs are owned by radarr/users instead of sc-radarr/radarr

Safihre:
The problem here is more Sonarr related because it doesn't clear out the `/tmp/nzbdrone*` after the updates that it does. It relies on the OS to do that. 
So the files there were created during a previous backup using the older version of the Sonarr package that used the old permissions system. 
I am not sure when DSM clears the `/tmp` folder, but definitely on reboot/startup.

jelmerj:
ah okay, that definitely makes it less of a priority, as you could just ask users to reboot their system as a fix for this problem. However, this problem is related to the new package version ofcourse, because of the user/group which come along. Are you guys detecting an upgrade from old-style to new-style package? If so, you might as well purge these 2 directories at the point (or fix the user/group)

Safihre:
@ymartin59 what do you think?
A simple `rm` in postupgrade of these 2 folders could solve this indeed.

ymartin59:
I have to look in detail. From my point of view, as installer script run as root, `preupgrade` should clean up target location before generating backup... and `postupgrade` has to apply permissions on files after copying files into `var`.

Safihre:
[...]

PR #3190, opened 2018-03-01
Correct permissions of Sonarr/Radarr backup upon upgrade

Safihre:
@ymartin59 I think the release of _just_ Sonarr/Radarr is creating problems. Because SABnzbd/NZBGet/Deluge are not updated they still create files as `sabnzbd:users` and the new Sonarr/Radarr packages are no longer part of `users`.  I think this is the source of the problems described in #3185 . Now that you published Sonarr/Radarr, we need the downloaders to be updated ASAP.. ⏩⏩
And it also means that also other processors like Sickrage/CouchPotato also need updates right away! This whole switch needs to happen at once, we can't do it one package at a time 😨 

Regarding this upgrade problem:
While I first thought that removing these files would be an option, I now see that the Sonarr/Radarr Wiki actually refer to this `/tmp/` location [as a valid source of recent backups](https://github.com/Sonarr/Sonarr/wiki/Backup-and-Restore). I inspected the backup and update process and these are the only folders created. 

### Checklist


ymartin59:
@Safihre The same way we decided to keep busybox in packages at least to clean up previous service user, wouldn't it be possible to enlist new service user to both `sc-` group and `users` group if available in system?

Safihre:
Sure we could, I actually did something like that for #3094. 
But doesn't that sort of remove the purpose of having this whole service-user way? Could just run all packages as 1 user basically. Or am I over-generalizing?

Safihre:
@ymartin59 when are you publishing all these things like this fix and the download packages? Because in the weekend there wil be a big wave of updates since that's when users have time for it.
Otherwise we will be flooded with people complaining about permissions!

ymartin59:
@Safihre I do not think so, as I have unpublished/desactived radarr/nzbdrone packages from repository... Now I decided to do so as soon as an "big" issue is reported on "just published" packages. Sorry but I have a "real" life too.

PR #2872, opened 2017-08-14
Add Docker DSM and update digests of 6.1 toolchains

ymartin59:
_Motivation:_ Produce package for target platform "Docker DSM" as `dockerx64`
_Linked issues:_ #2866

I have already built `nano` and `transmission` packages from `dsm6` branch. Still looking for a proper way to test generated packages - for the moment, I submit them to beta-testers

PR #3195, opened 2018-03-03
add dockerx64 to generic toolchain

hgy59:
_Motivation:_ dockex64 arch is compatible to syno-x64
_Linked issues:_ #2872

### Checklist (validated with mc, midnight commander)

as supposed by @ymartin59 dockex64 arch is compatible to syno-x64
after testing with mc (based on /cross/mc) I saw that the toolchains of syno-x64 and syno-dockerx64 are almost binary the same - only 3 header files (.h) are different and one of them is autogenerated kernel configuration.


ymartin59:
@hgy59 I am really pleased to hear about your confirmation... yet another (often large) binary not to build and upload. Again, many thanks for your work/help/support.

Issue #2351, opened 2016-05-29
[Package Request] Borg Backup


Obruut:
Description: BorgBackup (short: Borg) is a deduplicating backup program. Optionally, it supports compression and authenticated encryption.
Website: http://borgbackup.readthedocs.io
Documentation: http://borgbackup.readthedocs.io/en/stable/usage.html
Build/Installation documentation: http://borgbackup.readthedocs.io/en/stable/installation.html
Source Code: https://github.com/borgbackup/borg
License: https://borgbackup.readthedocs.io/en/stable/authors.html#license

artelse:
I just installed Borg onto debian-chroot and no problems compiling/installing. I am only having problems mounting volumes as posted in previous issue.

Obruut:
True, that's possible. But I think more people could use Borg if it was possible to install as a package.

The past months Borg Backup has grown to one of the best dedup backup programs I've know, and would love a 'native' solution.

kBite:
@yuroller I'm very interested in this package. I have a DS218play and if possible would like to help. Would you please list all open tasks here?

yuroller:
If you are able to build packages for you ds unit, borgbackup should build and (hopefully) run correctly.
Some time ago I successfully built borgbackup for all supported architectures by spksrc project, but I was able to test just on my ds212+ 6.0.2.
You can find build instructions on wiki.
Borgbackup depends on python3 packages, so you need both.
Ask if you need help.

ymartin59:
Do not bother to build it yourself. I am publishing that package

noplanman:
@ymartin59 Are you publishing others too? Where are they listed?

noplanman:
:tada: Looks like Borg is now officially on the repo, great!

But it still needs a manual build/install of Python3, as the version on the repo is outdated.

Will that be updated soon as well?
@ymartin59 Is there a way to contribute builds to the project?

ymartin59:
@noplanman What is your DSM architecture?

noplanman:
`qoriq` and `armada370` are the ones I have.

I've built Python3 for both of them and installed Borg successfully, so all good :smiley:

noplanman:
@ymartin59 I just noticed that the Python 3 package on the SynoCommunity packages page is at 3.5.2-5, but the constraint for Borg is set at 3.5.2-6.
As I still only see 3.4.1-4 in my DSM package center, I can't test to see if it works.

[...]

PR #3196, opened 2018-03-03
borgbackup: update to 1.1.4

yuroller:
_Motivation:_ Update package borgbackup to 1.1.4.
_Linked issues:_ https://github.com/SynoCommunity/spksrc/pull/2998 https://github.com/SynoCommunity/spksrc/issues/2351

### Checklist

Tested on Synology DS212+


Safihre:
Let's see if we can get it to pickup the .so files without including them in Python3.

yuroller:
Thanks. borgbackup requires libacl which in turn requires libattr.
The package I provided has this .so libs problem, but it works on my synology unit: over the weekend I did a remote backup of some GB of data in order to test it.

spk/python3/Makefile
```diff
@@ -10,7 +10,7 @@ DEPENDS += cross/setuptools cross/pip cross/wheel
 DEPENDS += cross/cffi cross/bcrypt
 # Cross-compiled wheels
 DEPENDS += cross/lxml cross/pycrypto cross/pycurl cross/pyyaml
-DEPENDS += cross/msgpack-python cross/attr cross/acl cross/lz4
```
ymartin59: I do not understand why removing `lz4` from Python 3... I may impact other packages running over Python 3.
What is the reason / motivation ?

yuroller:
Lz4 was added because the previous version of borgbackup needed it and, as far as I know, no other package is using it.
Starting from borgbackup 1.1.4, libz4 is bundled with borgbackup, so it is not needed anymore in python3.

ymartin59:
So it means it should run with already published Python 3 package. I will test and publish if OK.

yuroller:
Yes, it should. But just if you lower the python3 required version in borgbackup package or else it would not install.

ymartin59:
So please remove changes to `cross/python3` and change dependencies. When done, I will merge and publish.
Just a question as a confirmation: it is not expected to run as service, it is only a command line utility, right ?

yuroller:
It is not a service: it runs from commandline.
About python3 dependencies: borgbackup requires libacl (which in turn requires libattr), liblz4 is not required anymore (it is bundled in borgbackup source package): I was not able to put those dependencies inside borg package, but instead had to put inside python3 package or else borg it is not able to find them.
Can somebody explain me why is it like that?

ymartin59:
So please:
- remove changes to `cross/python3`
- remove useless `spk/borgbackup/src/dsm-control.sh` and `spk/borgbackup/src/installer.sh`
- remove package dependency to `SPK_DEPENDS = "python3>=3.5.2-6"`
When done, I will merge and publish.

ymartin59:
@yuroller May you please review my request for changes from https://github.com/SynoCommunity/spksrc/pull/3196#issuecomment-371613824 - I really find it good so it will be merged and published

ymartin59:
Superseed by #3280

PR #3197, opened 2018-03-03
Validate digests

hgy59:
_Motivation:_ Add check whether the digests file contains the right distribution file
_Linked issues:_ #3157

- without the check, there was no error if an new version is referenced in the Makefile without updating the digests. The error was not shown until the old distribution files where deleted or a clean spksrc clone was used.
- fix download URL for dnscrypt
- fix digests files without line break at last line (md5)
- as introduced in #3157 the mk/spksrc.tc.mk uses now LOCAL_FILE for the name of the downloaded distribution file


Safihre:
Looks good to me, what do you think @ymartin59?

PR #3200, opened 2018-03-04
Fixed 3 issues with compiling native packages

eelcohn:
_Motivation:_ I found three issues with compiling native packages. This pull request fixes these issues.
_Linked issues:_ Don't think these issues have been found yet

### Checklist
- [] Build rule `all-supported` completed successfully: all native packages build successfully
- [] Package upgrade completed successfully: N/A
- [] New installation of package completed successfully: N/A

Issue 1:
When a Git-based native package is added, the make will fail:
```
===> Downloading files for native-tgl
/bin/sh: line 4: -gitffb04ca: command not found
```
This is caused by an erroneous extra space character on line 11 in the spksrc.native-cc.mk file. Because of this extra space character, line 61 in spksrc.download.mk `localFolder=$(NAME)-git$(PKG_GIT_HASH)` is interpreted as two seperate commands; hence the -gitffb04ca: command not found error.
Fixed by removing the character.

Issue 2:
After fixing issue 1, the make still fails, because the Git archive file is downloaded with the prefix `native-`, while the tar commands tries to extract the file without the `native-` prefix.
Fixed by removing the `native-` prefix on line 11 in the spksrc.native-cc.mk file.

Issue 3:
When testing if all other native packages would compile successfully, I found that glib could not compile correctly. This is because the version in Makefile is 2.55.2 while the versions in the digests file was still at 2.50.2.
Fixed by updating the digests file for the glib package.


eelcohn:
See also #3157

Safihre:
Are you sure you added this fix for 1? I only see 2 and 3 in the code.

eelcohn:
If you look closely, the fix for 1 is also included. Take a look at the changed files tab of this pull request, specifically at line 10 of the spksrc.native-cc.mk file. The original version has an extra space character _**after**_ the `$(PKG_NAME)`. And yes, it's very hard to notice this extra space character. Took me hours of debugging to find this one :)

Safihre:
I see 👍 
And this doesn't break anything for the other `native` packages?

eelcohn:
I've tested this by compiling all of the `native` packges one-by-one (this is how i found issue no.3 because it wouldn't compile). Some packages give me warnings, but I've checked this by comparing these warnings with the warnings produced by the current branch. No new warnings are added by this pull request, and compiling seems to be exactly the same for each of the `native` packages.

Safihre:
Cool.
Looks good to me, @ymartin59 ?

hgy59:
fix of issue 3 is included in #3197

Issue #2358, opened 2016-06-03
[Package Request] stunnel 


sirjaymz:
Description: stunnel
Website: https://www.stunnel.org/index.html
Documentation: https://www.stunnel.org/docs.html
Build/Installation documentation: hmmm
Source Code: https://www.stunnel.org/downloads/stunnel-5.32.tar.gz - ftp://ftp.stunnel.org/stunnel/
License:  Stunnel is a free software authored by Michał Trojnara. Although distributed under GNU GPL version 2 or later with OpenSSL exception, stunnel is not a community project. We retain the copyright of the source code. Please contact us for commercial support or non-GPL licenses. Free, community-based support is also available via stunnel-users mailing list.

sirjaymz:
https://www.bountysource.com/issues/34820100-package-request-stunnel

StoneISStephan:
Did you know that you can install stunnel with bootstrapper?
ipkg (optware) contains stunnel4
opkg (entware) contains stunnel5

cytec:
@Dr-Bean so im currently working on this as i had some need for myself... seems like there isn't any sort of Icon or something available ... what should i do? can't think of any other app where we don't have an icon at all...

Dr-Bean:
@cytec Just a quick response due to holidays: Similar to ncdu package, maybe use the screenshot from https://www.stunnel.org/screenshot.html? Could even put a watermark with text `stunnel` over it ;)

hgy59:
@sirjaymz are you still waiting for this package? 
You can download all packages created by the github build action [here](https://github.com/SynoCommunity/spksrc/actions/runs/351496045) (click the packages link) for manual installation.
These are all generic archs. You find the matching generic arch for your specific Diskstation Model  in the wiki page [Package Archs](https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures).

Please report any feedback in  #4261.

PR #3211, opened 2018-03-07
Added new package: Lua v5.3.4

eelcohn:
_Motivation:_ This pull request will add the [Lua v5.3.4](https://www.lua.org/) package to the SynoCommunity
_Linked issues:_ N/A

### Checklist

**Build rule `all-supported` completed successfully**
Build rule is successfully completed for DSM6.1 on 88f6281, alpine, armada370, armada375, armada38x, armadaxp, comcerto2k, denverton, dockerx64, evansport, hi3535, kvmx64, monaco, ppc853x, qoriq, rtd1296 and x64 architectures.

**Package upgrade completed successfully**
N/A; this is the first package release for Lua.

**New installation of package completed successfully**
Tested on armadaxp and x64.


ymartin59:
May you please motivate why SynoCommunity should publish and maintain Lua language ? Except as a requirement for an interesting application, I will not "extend" volume under maintenance in `spkrepo` with yet another language... we already have a large enough set of them.

eelcohn:
I've created a package for [telegram-cli,](https://github.com/vysheng/tgl/) a command line tool / daemon for interacting with the [Telegram messenger](https://telegram.org/). This tool requires the Lua language as a dependency.
In order to keep the pull request clear, I've split the pull request in two: one for the Lua language (this one) and one for the telegram-cli tool.
Can you please give me some pointers on what an 'interesting application' is to SynoCommunity? I've been working on two other packages as well, [stunnel](https://www.stunnel.org/) (see #2358) and [Eggdrop](https://eggheads.org/), and I'd like to know if I can publish them on SynoCommunity, or that I need to find another platform for distributing these packages.

ymartin59:
That is a good question... I think about it and write a FAQ entry.

eelcohn:
Ok, thank you.

spk/lua/src/service-setup.sh
```diff
@@ -0,0 +1,52 @@
...
+		rm /usr/local/bin/$(basename "$executable")
+	done
+
+	exit 0
+}
+
+service_postuninst ()
+{
+	# Remove permanent config store for this package
+	rm -rf /usr/syno/etc/packages/${PACKAGE}/
```
ymartin59: What is requirement to do so ? As package wrote nothing there, is it really important to clean this ? From my point of view, it is not package responsibility to clean this but DSM framework

spk/lua/src/service-setup.sh
```diff
@@ -0,0 +1,52 @@
...
+
+	exit 0
+}
+
+service_preupgrade ()
```
ymartin59: Please remove empty methods from `service-setup` script... `installer` script test function exists before calling it if available.

cross/lua/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+DEPENDS =
+BUILD_DEPENDS = cross/ncurses cross/readline
+
+HOMEPAGE = https://www.lua.org/
+COMMENT = Lua is a powerful, efficient, lightweight, embeddable scripting language. It supports procedural programming, object-oriented programming, functional programming, data-driven programming, and data description.
+LICENSE = MIT
+
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = myCompile
```
ymartin59: Please use more recent naming convention: "lua_custom_compile" and "lua_custom_install" - which helps to understand faster what fails at build !

cross/lua/patches/Makefile.patch
```diff
@@ -0,0 +1,45 @@
+--- Makefile.orig	2017-04-26 18:12:27.442079208 +0200
```
ymartin59: Please use recent file naming convention, for instance "001_Makefile_add_shared_library.path"
If this patch comes elsewhere, please add a comment as patch first line with description why and original source

cross/lua/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+.PHONY: myCompile
+myCompile:
+	@$(RUN) $(MAKE) linux \
+		CC="$(TC_PATH)$(TC_PREFIX)gcc -fPIC \$$(ALLFLAGS) -I$(STAGING_INSTALL_PREFIX)/include/readline" \
+		CXX="$(TC_PATH)$(TC_PREFIX)g++ \$$(ALLFLAGS) -I$(STAGING_INSTALL_PREFIX)/include/readline" \
+		AR="$(TC_PATH)$(TC_PREFIX)ar rcu \$$(ALLFLAGS)" \
+		RANLIB="$(TC_PATH)$(TC_PREFIX)ranlib \$$(ALLFLAGS)" \
+		CPPFLAGS="$(CPPFLAGS)" \
+		LDFLAGS="$(LDFLAGS)" \
```
ymartin59: Please add `-rpath,/var/packages/lua/target/lib` to avoid wrapper scripts with `LD_LIBRARY_PATH`

cross/lua/Makefile
```diff
@@ -0,0 +1,35 @@
+PKG_NAME = lua
+PKG_VERS = 5.3.4
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://www.lua.org/ftp/
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS =
+BUILD_DEPENDS = cross/ncurses cross/readline
```
ymartin59: Please move these to `DEPENDS`, so that there is no need for copy files to lua `PLIST`

spk/lua/PLIST
```diff
@@ -0,0 +1,7 @@
+lib:lib/libncurses.so.6.*
```
ymartin59: Useless file if libraries are declared as `DEPENDS` instead of `BUILD_DEPENDS`

spk/lua/src/service-setup.sh
```diff
@@ -0,0 +1,52 @@
...
+service_preinst ()
+{
+	exit 0
+}
+
+service_postinst ()
+{
+	echo "Creating startup scripts for Lua executables" >> ${INST_LOG}
+	for executable in ${SYNOPKG_PKGDEST}/bin/lua*; do
+		echo "LD_LIBRARY_PATH=/var/packages/${PACKAGE}/target/lib:\${LD_LIBRARY_PATH} $executable" > /usr/local/bin/$(basename "$executable")
```
ymartin59: I would not do such wrapper scripts. Please use adequate options to generate RPATH in binaries https://en.wikipedia.org/wiki/Rpath

cross/lua/patches/src.Makefile.patch
```diff
@@ -0,0 +1,31 @@
+--- src/Makefile.orig	2017-04-26 18:39:24.655503489 +0200
```
ymartin59: Same comment: rename patch file and add first line comment with description

eelcohn:
Thank you for your feedback, I'll edit the pull request. Might take me a couple of days though.

spk/lua/PLIST
eelcohn: I'll declare the libraries as `DEPENDS` instead of `BUILD_DEPENDS` as you requested, but please consider the following:

If I'd move the ncurses dependency from `BUILD_DEPENDS` to `DEPENDS`, all the files from the ncurses library will be incorporated into the package, not just the library file. This will add all sorts of unnecessary files to the package, while only the `libncurses.so` file is needed. IMO this is a bad idea, not only for security reasons, but it also makes the package unnecessary large.

The real problem is though that the latest version of DSM (6.1.5 update 1) is shipped with older versions of libncurses (v5.9) and libreadline (v6.3). spksrc is already updated to v6.0 for libncurses and v7.0 for libreadline. Because of this, any package compiled with spksrc cannot use the default libraries shipped with DSM. Using the default library versions shipped with DSM is the most preferred solution IMO.

As soon as DSM updates libncurses and libreadline to the latest version, packages can use the default libraries. Until then, packages compiled with spksrc have to be shipped with custom libraries.

eelcohn:
I've implemented the changes you requested.

cross/lua/Makefile
```diff
@@ -5,31 +5,31 @@ PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://www.lua.org/ftp/
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
-DEPENDS =
-BUILD_DEPENDS = cross/ncurses cross/readline
+DEPENDS = cross/ncurses cross/readline
+BUILD_DEPENDS =
```
ymartin59: So it can be removed

cross/lua/patches/001_Makefile_add_shared_library.patch
```diff
@@ -0,0 +1,45 @@
+--- Makefile	2017-04-26 18:12:27.442079208 +0200
```
ymartin59: Please add a comment with explanation what these patches are used for. And if coming from an external source, URL to its source.

spk/lua/src/service-setup.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
ymartin59: Please remove shebang and also PACKAGE and DNAME variables, which are useless (see `demoservice` or `mosquitto` as example...)

spk/lua/src/service-setup.sh
```diff
@@ -0,0 +1,28 @@
...
+
+
+service_postinst ()
+{
+	echo "Creating startup scripts for Lua executables" >> ${INST_LOG}
+	for executable in ${SYNOPKG_PKGDEST}/bin/lua*; do
+		ln -s "$executable" /usr/local/bin/$(basename "$executable")
+	done
+
+	exit 0
```
ymartin59: Do not invoke `exit` in such "hook" shell function... it stops script and prevent remaining code in framework `installer` to run.

spk/lua/src/service-setup.sh
```diff
@@ -0,0 +1,28 @@
...
+}
+
+service_preuninst ()
+{
+	# Remove startup scripts for the executables
+	for executable in ${SYNOPKG_PKGDEST}/bin/lua*; do
+		rm /usr/local/bin/$(basename "$executable")
+	done
+
+	exit 0
```
ymartin59: Do not invoke `exit` in such "hook" shell function... it stops script and prevent remaining code in framework `installer` to run.

ymartin59:
@eelcohn May you please take some minutes to improve `service-setup.sh` so that I can merge and publish.

hgy59:
@eelcohn are you still working on this PR?

eelcohn:
Sorry for my late reply. Since the original repository on which this pull request was based is lost, I've created a new pull request with all the changes as requested by ymartin59, see #3780 .
This pull request can be closed.

hgy59:
With #3780 this PR is obsolete.

Issue #3215, opened 2018-03-10
Permissions Issue Sonarr


forepj:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_Sonaar
_Package Version:_20180303-13

_NAS Model:_DS-1515+
_NAS Architecture:_
_DSM version:_DSM 6.1.5-15254 Update 1

### Expected behavior  Choose folder to import tv shows
Tell us what should happen  select folder from finder

### Actual behavior Cannot select folder
Tell us what happens instead 
Invalid request Validation failed:  -- Folder is not writable by user sc-nzbdrone

### Steps to reproduce
_1._I have given the shared folder volume1/video/tv shows r/w permission to sc-media and sc-download in Local Group via Control Panel and File Station  
_2._I have given the shared folder volume1/video/ r/w to sc-nzbdrone in  System internal user
_3._

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
```
Insert the package log here
```Invalid request Validation failed:  -- Folder is not writable by user sc-nzbdrone
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here

forepj:
Am I doing something wrong or should I wait for another update? I've uninstalled and reinstalled several times making sure to restart the NAS every time to no avail.  Other Syno packages SAB Transmission, Couchpotato, Headphones, Lazy Librarian are all currently up and running with r/w access to shared folders.

ymartin59:
Please try to give full "read" access to `/volume1/video` to group `sc-download`... it should work. Please submit some screenshots to confirm how you set permissions.

alxlaxv:
I had the same issue because I was pretty sure Sonarr and radio were in the “sc-media” group.. and as “users” has no rw access it went crazy.
Changing “sc-download” permissions fix it.

forepj:
![28d31c3c-611d-4101-bca3-54e72158e65c](https://user-images.githubusercontent.com/8423708/37250092-6f3a0bac-24a9-11e8-953d-99fdb6705563.png)
![d0c230bb-eae7-4f52-bfe8-431dfa2a1b36](https://user-images.githubusercontent.com/8423708/37250093-6f63659c-24a9-11e8-816d-aa2b438add99.png)
[...]

Issue #3212, opened 2018-03-08
Sonarr permissions no longer working with Sabnzbd


JZNine:
Hi all, I am having permission problems since the latest Sonarr update, I am now getting the access denied errors on post download processing files provided by Sabnzbd. Any help would be greatly appreciated as I've tried everything I can see here, including a fresh Sonarr install, but still no luck.

I have set permissions on the download location to RW for sc-download, but Sabnzbd seems to always set downloads to the Users group. 

### Setup
_Package Name:Sonarr
_Package Version:2.0.0.5153

_NAS Model:DS916+
_NAS Architecture:x64
_DSM version:6.2-23511

### Expected behavior
Sonarr picks up downloaded task from Sabnzbd, and renames and files correctly. 

### Actual behavior
Sonarr sends download to Sabnzbd, download completes, Sonarr fails to process with an access denied error

### Steps to reproduce
1: Select wanted episode to download and send to Sabnzbd (manually or automatically)
2: Let download
3: Check in Sonarr for error

#### Package log
Import failed, path does not exist or is not accessible by Sonarr: /volume1/media/Videos/TV/This.Country.S02E02.Threatening.Letters.720p.iP.WEB-DL.AAC2.0.H.264-RTN-Obfuscated.1/fdef9d629ed74c96a552581cdcc7cac2.mkv
![image](https://user-images.githubusercontent.com/37171600/37138087-d0be7c2c-22fc-11e8-9179-a73f3a6543f8.png)
![image](https://user-images.githubusercontent.com/37171600/37138131-fd18a07c-22fc-11e8-98fd-9138a5311886.png)

GoodOmens83:
Getting similar issues.

ymartin59:
May I ask you to please test this new package http://dl.free.fr/h4mI2xmkm with a "fix tentative" ? Thank you in advance for your feedback

JZNine:
Thanks. Still fails to process in Sonarr after downloading, unfortunately.

ymartin59:
May you please review your Shared Folder permissions according to the following documentation: https://github.com/SynoCommunity/spksrc/wiki/Permission-Management

JZNine:
Thanks for that comprehensive wiki article. After following that, Sonarr is now correctly processing, renaming and storing Sabnzbd downloads again.

Issue #3192, opened 2018-03-02
Permissions gone since Sonarr/Transmission update


WolfH:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ Sonarr / Transmission
_Package Version:_ Latest update

_NAS Model:_ DS214
_NAS Architecture:_ ARM
_DSM version:_ DSM 6.1.5-15254 Update 1

### Expected behavior
Sonarr should be able to move items from the Transmission download-folder to the media-folder as it did before the update.

### Actual behavior
It seems that permissions for Sonarr have been set wrong since updating to the newest version using the sc-download group. I'm getting errors like:

`Couldn't import episode /volume1/homes/admin/downloads/NAME.mkv: Access to the path is denied.`

I've checked the permissions in File Station and the downloads folder is owned by user sc-transmission and group sc-download (to which sonarr belongs if I'm not mistaken).

It worked fine before, this all started since updating to the new permissions-structure.

### Steps to reproduce
_1._ Install the latest update.
_2._ Try to download and move a file.

Safihre:
How are you with SSH? We need to identify where this is going wrong. Maybe you can also help @SX86?
If you SSH into the device, can you post the output of each folder in the path up to the file of:
```
ls -al /volume1/
sudo synoacltool -get "/volume1/" 
```
And then also for any of the folders up to the file's folder?

WolfH:
I'm not super with SSH, but here goes nothing. I mentioned the wrong folder above, it's the bittorrent folder.

```
admin@TheTARDIS:~$ ls -al /volume1
total 10956
drwxr-xr-x  16 root  root       4096 Feb 28 08:26 .
drwxr-xr-x  21 root  root       4096 Feb 28 08:26 ..
drwxr-xr-x  19 root  root       4096 Mar  2 07:59 @appstore
-rw-------   1 root  root      14336 Feb  6 08:17 aquota.group
-rw-------   1 root  root      15360 Feb  6 08:17 aquota.user
drwx------   2 root  root       4096 Feb  6 08:12 @autoupdate
drwxr-xr-x   5 admin users      4096 Feb 27  2017 @database
drwxr-xr-x   2 admin users      4096 Oct 14  2013 @download
drwxrwxrwx+  4 root  users      4096 May 12  2015 @eaDir
-rw-------   1 root  root   10395648 Jun 14  2014 @heartbeatd.core
[...]

PR #3216, opened 2018-03-10
Fix various permission issues

ymartin59:
Grant GROUP (sc-download) permission to browse subfolder in Share Folder
Conditionally enlist service user into "users" group to access contents

_Linked issues:_  #3192 #3215 #3212 #3185

### DIFF
--- a/mk/spksrc.service.installer
+++ b/mk/spksrc.service.installer
@@ -117,6 +117,22 @@ syno_group_remove ()
     fi
 }
 
+# Add user to existing group
+syno_user_add_to_group ()
+{
+    ADD_USER=$1
+    ADD_GROUP=$2
+    # Check user already in group
+    if ! synogroup --get "$ADD_GROUP" | grep "^[0-9]:\[${ADD_USER}\]" &> /dev/null; then
+        # Add user, not in group yet
+        echo "Adding '${ADD_USER}' to '${ADD_GROUP}'" >> ${INST_LOG}
+        MEMBERS="$(synogroup --get $ADD_GROUP | grep '^[0-9]' | sed 's/.*\[\([^]]*\)].*/\1/' | tr '\n' ' ')"
+        # The "synogroup --member" command clears all users before adding new ones
+        # so all the users must be listed on the command line
+        synogroup --member "$ADD_GROUP" $MEMBERS "${ADD_USER}" >> ${INST_LOG}
+    fi
+}
+
 # Sets recursive permissions for ${GROUP} on specified directory
 # Usage: set_syno_permissions "${SHARE_FOLDER}" "${GROUP}"
 set_syno_permissions ()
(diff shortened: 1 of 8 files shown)


Safihre:
So we would use this on all download related packages or just processing packages like Sonarr?
Because if we add all packages, then why use `sc-download` at all? Add everything to `users` and we have no problems at all. 
If we only add to processing packages, the objective is just backwards compatibility right? Because for example the `sc-nzbget` user wouldn't be part of `users`, only `sc-download`.

ymartin59:
Here is my idea:
- service user is member of "users" to read existing content coming from other users
- download services write in there own folder, which requires to be in "sc-download" to access them "by default"
- protocol service will run in no specific group, neither "users" nor "sc-download" to avoid "leaking" content

If user want "download" folders (from transmission for instance) to be accessible to everyone, they can add "users" read permission on it.

ymartin59:
I just wonder if it is a good idea to set as "owner" folders in Shared... when package is removed, some strange effects may happen:
```
xpenology52> synoacltool -get /volume1/downloads                       
ACL version: 1 
Owner: `������� Y��(user) not found 
--------------------- 
	 [0] group:administrators:allow:rwxpdDaARWc--:fd--  (level:0)
	 [1] group:sc-download:allow:rwxpdDaARWcC-:fd--  (level:0)
```

Safihre:
> download services write in there own folder, which requires to be in "sc-download" to access them "by default"

If all services (both downloaders and professors) are already in `users`, we could just remove `sc-download`. Don't you think?

ymartin59:
Maybe I have guessed what DrBean idea was:
- segregate service accounts from standard end-users account
- "sc-download" group permission for file producers/writers
- "sc-media" group permission for content consumers/readers

Probably sc-download/sc-media split is too complex for most end-users... But I think it is relevant for security and some advanced users (like some small companies for instance) to keep permission segregation.

If a service account is in "users", it does not necessarily mean it will write files readable for "users". My idea is to enlist service account into "users" to be able to read "public" files (readable to anyone)

ymartin59:
I have even imagine in framework to propose users to choose group name from wizard thanks to `SERVICE_WIZARD_GROUP`... I do not remember who rejected this idea, but it would allow to choose between old-style "users" (open-bar) and standard segration (sc-download) or even custom segregation with one group per application (at worst)
For application with both producer and consumer roles, two groups are required... as service B may read content from another service A and write content to service C. That is exactly our user case at the moment, with "users" group to read content and "sc-download" to provide it to others.

Safihre:
I am pretty sure it was me who suggested to remove the group option from wizard. Because I think  SynoCommunity package's should be here to make things easier for regular end-users.. Not expect them to be permissions and command line experts. 

I don't think I completely understand how  the proposed approach would work in practice: how things can be just readable for one group but read and writeable for the other. Doesn't a consumer like Sonarr need also to write (remove) files? 
But if it works and makes things easier for the end-users: let's do it :)

ymartin59:
I agree with you... but I really think there is no reason to give any application full access to "users" group. I am not sure if it makes things clearer but I wrote a concept as a specification/design: #3217

alxlaxv:
Being part of a group like media or download is pretty fine and convenient.

ymartin59:
@Safihre Are you OK with this PR ?

Safihre:
Yes looks good to me.
Now we only add the to `users` when upgrading 👍 
One final improvement would be to detect non-ACL shares and warn users before installing. If this is possible?
But that's for another PR!

ymartin59:
@Safihre I agree. I have no idea how to check and test it, as I only have recent DSM version where Share Folder have ACL enabled by default... We may ask @BenjV for tips about it - with hope its system may still contain non-ACL-enabled Share Folder.

ymartin59:
@Safihre @BenjV `synoshare` command has no convert action, so my best bet is probably to test error return from `--list_acl` if any

BenjV:
synoshare --list_acl does not give you this info.
You have to use: `synoshare --get [sharename]
`
This wil produce a datadump and the `[ACL] `attribute will tell you with a `Yes` or `No` if the share has acl capabilities.
If No then the user has to convert it via the wizard.

Safihre:
Doesn't `synoacltool` also show it? For all paths?

BenjV:
Not to my knowledge.
Can you give an example of the synoacltool command you think would work so I can try it on my last "Non-Acl" share.

ymartin59:
@BenjV May you please report output of `synoacltool -get /volume1/downloads` if you have a "downloads" non-ACL Share Folder on "volume1".

BenjV:
I am sorry, I made a mistake.
On my two Nasses I have a share with the same name and when I was testing yesterday I must have mistakingly used the wrong Nas for my test with the synoacltool -get command.

You were right it gives this output when the share has no ACL support:
`(synoacltool.c, 359)It's Linux mode`
and the return code is 255

Safihre:
That's pretty general.. I also see that sometimes on DSM6 when old-style packages did some `chown` commands. 😢

BenjV:
Yes it is general but enough to determine if the share has to be converted to support ACL's.
If I give that command on a share that is converted, I get a whole list of ACL's and a return code of 0.

ZeroZorro999:
Sorry to intrude, but reading this thread, will this ACL-requirement, when implemented, has as side-effect that external devices (usbshares) won't be able to be the destination shared-folder for Sonarr/Radarr?

Issue #3201, opened 2018-03-04
Deluge 1.3.15-9 not starting - OpenSSL problem


tonabnehmer:
### Setup
_Package Name:_ Deluge from #3138 
_Package Version:_ 1.3.15-9

_NAS Model:_ DS215j
_NAS Architecture:_ armada375
_DSM version:_ DSM 6.1.5-15254 Update 1

--

### Expected behavior
Deluge starting

### Actual behavior
Deluge not starting, with Python from #3138, version: 2.7.14-18 see attached log
Seems like it does not find the imports for openssl or crypto packages? 

#### Package log
```
[INFO    ] 17:05:30 configmanager:70 Setting config directory to: /volume2/@appstore/deluge/var/
[ERROR   ] 17:05:31 main:248 /var/packages/python/target/lib/python2.7/site-packages/cryptography/hazmat/bindings/_openssl.so: undefined symbol: pthread_atfork
Traceback (most recent call last):
  File "/volume2/@appstore/deluge/env/lib/python2.7/site-packages/deluge-1.3.15-py2.7.egg/deluge/main.py", line 240, in start_daemon
    from deluge.core.daemon import Daemon
  File "/volume2/@appstore/deluge/env/lib/python2.7/site-packages/deluge-1.3.15-py2.7.egg/deluge/core/daemon.py", line 47, in <module>
    from deluge.core.rpcserver import RPCServer, export
  File "/volume2/@appstore/deluge/env/lib/python2.7/site-packages/deluge-1.3.15-py2.7.egg/deluge/core/rpcserver.py", line 47, in <module>
    from twisted.internet import ssl, reactor, defer
  File "/var/packages/python/target/lib/python2.7/site-packages/twisted/internet/ssl.py", line 59, in <module>
    from OpenSSL import SSL
  File "/var/packages/python/target/lib/python2.7/site-packages/OpenSSL/__init__.py", line 8, in <module>
    from OpenSSL import crypto, SSL
  File "/var/packages/python/target/lib/python2.7/site-packages/OpenSSL/crypto.py", line 16, in <module>
    from OpenSSL._util import (
  File "/var/packages/python/target/lib/python2.7/site-packages/OpenSSL/_util.py", line 6, in <module>
    from cryptography.hazmat.bindings.openssl.binding import Binding
  File "/var/packages/python/target/lib/python2.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 13, in <module>
[...]

PR #3218, opened 2018-03-11
fix openssl arches

hgy59:
_Motivation:_ fix OPENSSL_ARCH for apollolake and other archs
_Linked issues:_ A lot of packages use openssl

- Add the available arch macros to Makefile.
- Fix for different intel archs (apollolake, broadwell, dockerx64, ...) to use `linux-x86_64` (was `linux-generic32`) 
- Fix for ARM8 arch (rtd1296)
- Use `linux-x32` for x86 arch (was `linux-generic32`) 
Published packages for all kind of x64 archs use the general x64 arch, and therefore are not affected.

There are still some older archs using `linux-generic32` that are handled as before, to not break existing builds.

### DIFF
--- a/cross/openssl/Makefile
+++ b/cross/openssl/Makefile
@@ -19,16 +19,34 @@ STAGING_INSTALL_PREFIX := $(INSTALL_DIR)
 
 include ../../mk/spksrc.cross-cc.mk
 
-OPENSSL_ARCH = linux-generic32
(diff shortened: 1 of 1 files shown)


Safihre:
Maybe this will help with #3201!

Safihre:
Unfortunately unrelated to #3201, but still seems to work fine for compilation.

Issue #3207, opened 2018-03-06
ffmpeg vaapi intel braswell? native / cross package ?


MrSpoocy:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ ffmpeg
_Package Version:_ master

_NAS Model:_ DS416Play
_NAS Architecture:_ braswell-x64
_DSM version:_ 6.1

Hi everyone,

currently i try to extend the ffmpeg pack with vaapi support for intel. But the HOW-TO is incomplete. I can't find anything about "native" package. What is the difference to "cross"?

I build successfully libdrm-intel, cmrt, libva on Ubuntu 16 native. But when i try to do it with spksrc  there is a huge dependence chain. Do i really have to implement every dependence chain with own cross/XY ?

E.g. libpicaccess, libdrm2, libc6, lib ...

ymartin59:
I wish you good luck.. I have started that path and run away for the main reason that I will not be able to deliver such support for any other platform than Intel ones. And loading an additional kernel module is not piece of cake. That is why I suspended investigation about vaapi: https://github.com/ymartin59/spksrc/issues/6

MrSpoocy:
What about my question native vs. cross? I think every user with a Intel Synology will be happy about vaapi. Plex & Synology was able to build a ffmpeg with vaapi. We can do it! ;)

cytec:
The difference between native and cross is that native is build natively on the build box and will never ever even come close to your DS... so basically native is stuff you'll need to have on you Buildbox in order to be able to build some of the cross packages... 

cross packages are packages which are compiled using the coresponding synology toolchains of the selected arch

Example: go
There is go native, which installs go on the buildbox and bootstraps it so you'll have a working compiler to create the syncthing package.

m4tt075:
@MrSpoocy I like your attitude! Synology publishes their sources, see here for instance:
https://sourceforge.net/projects/dsgpl/files/Synology%20NAS%20GPL%20Source/22259branch/apollolake-source/
It contains `ffmpeg`. If you log into your NAS, you can get the compile options from the binary, but you probably know this already. Go for it...    👍

Safihre:
@m4tt075 so why can't we use it?

m4tt075:
[...]

PR #3222, opened 2018-03-13
Update ffmpeg to 3.4.2 with vaapi support for intel x64

hgy59:
_Motivation:_ Support video accelleration API for intel chips
_Linked issues:_ #3207

### Checklist

### Features
- ffmpeg is built with --enable-vaapi (for x64 archs only)
- libva-utils is included (vainfo and other tools)
- i965 video driver is included
- tested with DS918+ (build for arch-apollolake-6.1 and arch-x64-6.1)


ymartin59:
OK interesting. But why not implement my proposal to link against Video Station shared libraries already available ?

ymartin59:
@hgy59 My idea https://github.com/SynoCommunity/spksrc/issues/3207#issuecomment-371965875 was to build ffmpeg thanks to "libva*.so" libraries from VideoStation to avoid to rebuild them including kernel modules.
I would try following sequence:
- download VideoStation package from Synology
- extract it into `install` so that `ffmeg` CMake finds `libva*.so`
- add RPATH to CMake options to get binaries able to load VAAPI libraries from `/var/packages/VideoStation/target/lib`

Just an idea, it may not be successful... Your trial may face troubles to have libraries matching kernel modules.

hgy59:
@ymartin59 I presume that linking to *.so will not be possible without *.la file (not sure for linux, but on Windows you need the *.def file to link to *.dll)
any linux c/c++ link/libtool guru here?

hgy59:
@ymartin59 and how to download spk from synology from the build environment?

m4tt075:
@hgy59 Just googled, but maybe from here?
https://usdl.synology.com/download/Package/spk/VideoStation/?C=M;O=D

ymartin59:
@hgy59 About libraries, C compiler only requires `.h` file and `.so` to link dynamically: https://stackoverflow.com/questions/12237282/whats-the-difference-between-so-la-and-a-library-files
(`.a` and `.la` are for static linking)

MrSpoocy:
I think (if we can) we are avoid to use any lib from Synology. We can see that they use often old lib (ffmpeg, php 5.6) as core. Its better to create own "same" lib, to get the fresh version and newest bugfix.

ymartin59:
@MrSpoocy I can only agree with you, but concerning hardware-related libraries like "libva*" and "drm" kernel drivers, it sounds wiser to keep in-line with what Synology provides in DSM.

m4tt075:
@ymartin59 Have you ever had a chance to look into this PR? I think it could be a stepping stone for further updates. Obvisouly biased, benefitting from vapi integration...  ;-)

m4tt075:
@ymartin59 This PR looks very logical to me, but I'm also aware that you looked into an alternative implemenation method. From what you know now, would you still advocate an alternative approach?

ymartin59:
I still consider that if version provided in DSM "works", there is no need to rebuild or to update - I faced it recently with fuse for sshfs.
If lacking libva headers and so files to compile ffmpeg, spksrc may build same version from sources but no include them in package - so that binaries runs over libraries provided by DSM.

m4tt075:
@ymartin59 It takes character to arrive at conclusions like this, and even more so to state them publically. I enjoy being part of this project!
I fully agree with including this PR in our update.

@hgy59 Many thanks for your contributions and your patience!

ymartin59:
@hgy59 Thanks for your work. It will be included with FFmpeg 4.1 PR #3575 I will take care to keep your author in commits after merge. So I close this one.

PR #3232, opened 2018-03-19
Fix start-stop-status default status wait timeout

ymartin59:
Here is why script not longer works


m4tt075:
*LOL* Congratulations, @ymartin59 ! That one must have been tough to find...

Safihre:
Aaah. Makes sense.
Was actually a bit surprised you added this waiting back? Was there a package that needed it really?

PR #3260, opened 2018-03-27
Adding NZBHydra 2

maxrogers:
_Motivation:_ Adding NZBHydra 2 to synocommunity

### Checklist


Safihre:
Cool :+1:
(didn't test yet)

Safihre:
Actually, does it really need a `cross` package? Can just all be in the `spk`?

maxrogers:
@Safihre, it can if there is a means to download and verify via digest files in spk. I only know how to do that in cross. Point me to an example and I can change that.

Safihre:
@maxrogers Sorry, you are right indeed!

Safihre:
Would it be possible for you to maybe modify the logo and add a "2" for example?
I know both packages have the same logo, but we want to make some distinction in the package-center.

cross/nzbhydra2/Makefile
```diff
@@ -0,0 +1,25 @@
+PKG_NAME = NZBHydra2
+PKG_VERS = 1.4.8
+PKG_REPO = theotherp
+PKG_PLAT = linux
```
Safihre: Not needed, just use it directly.

spk/nzbhydra2/Makefile
```diff
@@ -0,0 +1,36 @@
+SPK_NAME = nzbhydra2
+SPK_VERS = $(shell date +%Y%m%d)
+SPK_REV = 1
+SPK_ICON = src/nzbhydra2.png
+
+DEPENDS = cross/$(SPK_NAME)
+SPK_DEPENDS = "Java8>8.0.151:git"
```
Safihre: So if I read the documentation of NBZHydra it seems this installer only supports x64 linux.
https://github.com/theotherp/nzbhydra2
Therefore we should specify this using: 
```
UNSUPPORTED_ARCHS = $ARM_ARCHES $PPC_ARCHES $x86_ARCHES
```

However, since this is a `noarch` package (see few lines below) that wouldn't work since this variable is checked only during compilation (which there is none).
Therefore we need to specify a custom `INFO` file where you can set `archs`, but I don't think this is possible using `spksrc`?
https://developer.synology.com/developer-guide/synology_package/INFO_necessary_fields.html
What do you think @ymartin59?

Safihre:
I compiled it and it's working on my system 👍 
https://github.com/Safihre/spksrc/releases/tag/nzbhydra2-test

spk/nzbhydra2/Makefile
maxrogers: Added

cross/nzbhydra2/Makefile
maxrogers: removed PKG_PLAT and using "linux" directly now

ymartin59:
I am really reluctant to create a package for only `x64` arch... which already has benefits of Docker support for running anything.

This application is expected to run on any arch where Java is available - BUT installer is delivered as compiled Python. Package may runs python installer script from source but means it depends on both Java and Python (weird, isn't it?). I wonder what is service launcher made from, probably python too.

From my point of view, even installer and service launcher should have been written in Java.

Safihre:
Good point, let's ask the creator @theotherp!
Can this program really only run on `x64`? Or can we clone the repo and run it manually on any platform that Synology provides Java8 for? 
This also raises the question: what platforms does Synology support Java8 for?

Because @ymartin59 has a point: if it's really `x64` only, then these archs can also run Docker and that might be a much better option in this case.

theotherp:
I've never tested it but 32 bit should be fine.

Safihre:
@theotherp How about ARM?

theotherp:
If Python and JRE run so will NZBHydra2.

ymartin59:
@theotherp Is Python really needed for runtime? My opinion is no. Would it be possible to document manual installation process and service startup command line and arguments, so that we can create a package with only shell scripts and JRE dependency.

Iristyle:
@maxrogers any chance you're going to finish this off soon? Would rather install this than the dead nzbhydra.

Thanks!

maxrogers:
@Iristyle It sounds like the synocommunity admins haven't added it to the repo because we are unsure if it will run on ARM or 32bit systems. I have been running it since Mar 27th without any issue.

theotherp:
V1 runs on 5075, V2 on 5076.

schumi2004:
I checked nzbhydra2.log and i see the following error:
Fri Oct 12 12:19:26 CEST 2018
Starting nzbhydra2 command /volume1/@appstore/nzbhydra2/share/NZBHydra2/nzbhydra2 --daemon --nobrowser --datafolder /volume1/@appstore/nzbhydra2/var/
data --pidfile /volume1/@appstore/nzbhydra2/var/nzbhydra2.pid 
/var/packages/nzbhydra2/scripts/start-stop-status: line 49: /volume1/@appstore/nzbhydra2/share/NZBHydra2/nzbhydra2: cannot execute binary file: Exec 
format error

Looks like the binary didn't compile correctly or is Alpine arch (DS715) not supported maybe?

Running the binary by itself isn;t working either.

theotherp:
I'm out...

maxrogers:
> @maxrogers
> Any ideas or can you share a pre-compiled package?

Hmmm, give this a shot. Just compiled and tested fresh (includes the latest nzbhydra 2 2.0.9 release:)
https://www.dropbox.com/s/q5v18kxhuzjbenz/nzbhydra2_noarch-all_20181012-1.spk?dl=0

This worked on a fresh bromolow 6.1.7-15284 update 2 test machine I just set up. I Installed Java8-8.0.161-0015 + Git 2.15.1-11 then nzbhydra2 (above linked spk.)

@Iristyle

maxrogers:
@schumi2004 Also, if NzbHydra2 actually tried to start you should have a log here that might be helpful: `/volume1/@appstore/nzbhydra2/var/data/logs/wrapper.log`

A healthy start should look like this (in /volume1/@appstore/nzbhydra2/var/nzbhydra2.log:)
`Fri Oct 12 11:05:58 MST 2018
Starting nzbhydra2 command /volume1/@appstore/nzbhydra2/share/NZBHydra2/nzbhydra2 --daemon --nobrowser --datafolder /volume1/@appstore/nzbhydra2/var/data --pidfile /volume1/@appstore/nzbhydra2/var/nzbhydra2.pid 
Logging wrapper output to /volume1/@appstore/nzbhydra2/var/data/logs/wrapper.log
2018-10-12 11:05:58,200  INFO - Daemonizing...`

I will try running it side-by-side with nzbhydra 1 in a bit (to see if that causes issue.)

maxrogers:
I messed up this pull request while trying to clean it up (squash.) I am closing this one. Here is the corrected/new pull request: https://github.com/SynoCommunity/spksrc/pull/3486

schumi2004:
@maxrogers 
Installed your compiled package without luck, same error regarding binary execution.
https://pastebin.com/X0vQJFLB

I think Alpine architecture has issues with this.

maxrogers:
> I think Alpine architecture has issues with this.

Agreed. It seems that the `cannot execute binary file: Exec
format error` is a [common error](https://askubuntu.com/questions/648555/bash-program-cannot-execute-binary-file-exec-format-error) when attempting to run software that has either been compiled for a different architecture or is attempting to run a 64bit binary on a 32bit system. I believe your [Alpine (DS715)](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#alpine) is 32bit only?

@theotherp, NzbHydra2 [currently requires](https://github.com/theotherp/nzbhydra2#how-to-run) 64bit, correct? That being the case, we need to limit this package to only x64 systems. We could do that and make the package beta status.

schumi2004:
No, I didn't need to modify anything. At first it complained about Java 8 being needed but once fixed it installed straight forward.

ymartin59:
As package is declared a "no-arch", it is installable on any DSM. `UNSUPPORTED_ARCHS` is only used by build system to restrict generated target architectures.
I wonder why a Java-based application depends on architecture details like 32/64 bits - I fear it loads a native .so library...

funzzy1:
Hi,

I have created a repo with my spk.

You just need to download and install it. It is based on the nzbhydra spk from synocommunity.

I tested it on a DS415play NAS and it is working correctly:

[nzbhydra2-2.20.7.noarch.spk](https://github.com/funzzy1/nzbhydra2_spk/releases/tag/2.20.7)

schumi2004:
Hi @funzzy1 
I tested it and it works perfect but may i ask you a favor?
That is to distribute version 2.10.2 which contains migration tool from Nzbhydra1.
This migration tool has been removed in version 2.10.3 and up.
When migrated people can upgrade NZBHydra2 in-app to newest version.

funzzy1:
Hi @schumi2004,

This is my releases page: https://github.com/funzzy1/nzbhydra2_spk/releases
I have create the new 2.22.1 package and also the 2.10.2 as requested. I have just created the package, unfortunately I didn't test it. I hope it will work correctly.

schumi2004:
I see the 2.10.2 but no spk asset?
Once available I will test. So far nzbhydra2 itself runs without issues ;)

PR #3265, opened 2018-04-02
Add DNScrypt-proxy v2.0.10

publicarray:
_Motivation:_ I've always wanted to add dnscrypt to my router someday. So today is the day :)
Version 2 has many [improvements](https://github.com/jedisct1/dnscrypt-proxy/wiki/Differences-to-v1) over version 1. For example V2 supports DoH (DNS over HTTPS) and tries to pick the best server on lowest  round trip-time. 

I know there is more work to be done (adding a GUI) but I like to get some feedback, especially on how to handle multiple architectures thanks.

### Checklist


Safihre:
Thanks for this contribution. Looks usefull to me, but also want @ymartin59's thoughts on it.
However, could you please change this to use the Generic Service approach? 
https://github.com/SynoCommunity/spksrc/wiki/Service-Support
Examples can be found in various packages, like `demoservice`, `sonarr`, `sabnzbd` or `nzbget`.

publicarray:
Thanks, I'll deferentially look into it.

ymartin59:
@publicarray Thanks for your work. If you do not declare `SERVICE_PORT`, package will not contain service port description for firewall rule... so port access is probably rejected by defaults - and so it requires documentation to explain user how to manually declare its own port to get it opened.

From my point of view, it is not an issue to enforce port number https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports
as far as it is generally used as defaults and it eases package usage.

You really have to build from sources with Synology toolchain to have confidence in binary compatibility. Another detail as consequence, the `digests` file is irrelevant for most architectures.

publicarray:
@ymartin59 What port do you suggest? The standard DNS port is 53 which is already taken by Synology's dnsmasq. There is a chance that a high port number (e.g 10053) in the dynamic rage is already in use.

The application is written in Go. I need to read up on https://github.com/golang/go/wiki/GccgoCrossCompilation first. One thing I can tell you that it works flawlessly on my Router.

publicarray:
Thanks for that!

spk/dnscrypt-proxy/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+RELOAD_UI = no
+STARTABLE = yes
+BETA = 1
+
+WIZARDS_DIR = src/wizard/
+
+SERVICE_USER = auto
+SERVICE_SETUP = src/service-setup.sh
+# SERVICE_PORT = 10053
```
ymartin59: From my point of view these three variables should be set...

spk/dnscrypt-proxy/Makefile
```diff
@@ -0,0 +1,29 @@
+SPK_NAME = dnscrypt-proxy
+SPK_VERS = 2.0.9b2
+SPK_REV = 0
+SPK_ICON = src/dnscrypt-proxy.png
+
+UNSUPPORTED_ARCHS = $(PPC_ARCHES)
```
ymartin59: Sounds inconsistent with https://github.com/SynoCommunity/spksrc/pull/3265/files#diff-141db1dbf1b0d78677131440ebf146edR38
From my point of view, as Go is available for PPC, there is no reason why application does not run on PPC. May you please investigate/confirm? And so remove `UNSUPPORTED_ARCHS` accordingly

ymartin59:
@publicarray I am interested to know how this application is expected to be setup and use. Probably there is some firewall rule to enable, or maybe original DNS service has to be stopped to forward request to dnscrypt-proxy... Sincerely I am not used to DNS services, a small FAQ may help.

spk/dnscrypt-proxy/Makefile
publicarray: Go for ppc is not compiled right now: https://github.com/SynoCommunity/spksrc/blob/master/native/go/Makefile#L22 The problem is that ppc needs extra bootstrapping to compile https://dave.cheney.net/2015/10/16/bootstrapping-go-1-5-on-non-intel-platforms

spk/dnscrypt-proxy/Makefile
publicarray: If there are no objections I'll set it the port to 1053 and enable the varables. Note that it is registered for some remote desktop protocol: https://www.speedguide.net/port.php?port=1053. On a NAS however this should not be an issue.

cross/dnscrypt-proxy/Makefile
```diff
@@ -0,0 +1,66 @@
...
+DEPENDS = native/go
+
+CONFIGURE_TARGET = nope
+COMPILE_TARGET = myCompile
+INSTALL_TARGET = myInstall
+
+include ../../mk/spksrc.cross-cc.mk
+
+ifeq ($(findstring $(ARCH),$(ARM5_ARCHES)),$(ARCH))
+SYNCTHING_ARCH = arm
```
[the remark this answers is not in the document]
publicarray: Ups, yea I missed those :)

spk/dnscrypt-proxy/src/service-setup.sh
```diff
@@ -1,15 +1,15 @@
 SVC_CWD="${SYNOPKG_PKGDEST}"
 DNSCRYPT_PROXY=${SYNOPKG_PKGDEST}/bin/dnscrypt-proxy
 PID_FILE=${SYNOPKG_PKGDEST}/var/dnscrypt-proxy.pid
-CFG_FILE="${SYNOPKG_PKGDEST}/var/dnscrypt-proxy.toml"
```
ymartin59: Please let configuration file in "var"  which is the only location to survive to upgrade process "by default".

ymartin59:
About configuration file management, I recommend to place it in `var` (which simplify `service-setup.sh`) but in case you cannot pass configuration location as service argument, you may inspire from https://github.com/SynoCommunity/spksrc/pull/3281/files#diff-3e72e4e2d95c62afb4fe354ce011a0f2

cross/dnscrypt-proxy/Makefile
```diff
@@ -0,0 +1,63 @@
+PKG_NAME = dnscrypt-proxy
+PKG_VERS = 2.0.9b2
+PKG_EXT = tar.gz
+PKG_DIST_SITE = https://github.com/jedisct1/dnscrypt-proxy/archive
+PKG_DIST_NAME = $(PKG_VERS).$(PKG_EXT)
+PKG_DIR =  $(PKG_NAME)-$(PKG_VERS)
+
+HOMEPAGE = https://github.com/jedisct1/dnscrypt-proxy
+COMMENT  = A flexible DNS proxy, with support for modern encrypted DNS protocols such as DNSCrypt v2 and DNS-over-HTTP/2.
+LICENSE  = ISC License
```
ymartin59: Please keep it short as "ISC"

spk/dnscrypt-proxy/src/service-setup.sh
publicarray: Interesting. I always put configs in [/etc](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#purpose6). Generally  [/var](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#purpose31) is for data that changes while running the software. Ok I'll change it back.

publicarray:
Regarding the documentation, I'll add one to the wiki once we are happy with this package :)

ymartin59:
It definitely looks "perfect" ! You have take benefits from generic script... I appreciate the effort.
After merge, I will add french translation to wizards.

publicarray:
Nice! I'm actually attempting to [add a GUI](https://github.com/publicarray/spksrc) but I think I'll leave that as a separate PR.

publicarray:
My progress so far:
screen shot

publicarray:
I've attempted to add a bit of documentation to my wiki https://github.com/publicarray/spksrc/wiki/DNSCrypt-proxy-2 If you are happy with it I can add it to the wiki here.

publicarray:
I thought I give you guys an update on the GUI: 


(syntax highlighting done by codemirror)

Safihre:
Pretty cool! We could use that for more projects I imagine.

publicarray:
@Safihre  Thanks 😃. Maybe, there are some caveats though.

* The cgi program is written in go (my first ever go program) so devices with PPC will have to wait until the native/go script is can compile on PPC.
* cgi runs as the same user as the web-server (system) and needs extra permissions to write files in the /var/packages/your-awsome-package/target folder.
* I haven't documented anything yet

publicarray:
@ymartin59 it's been a while, Do you want to merge this PR first or go straight to the GUI once it's reviewed?

ymartin59:
@publicarray Correct. I propose you rebase #3296 (and do some squashing) to ease that second PR review. I propose to publish all-in-one when ready.

Issue #2951, opened 2017-10-08
[Request] Support RT2600AC router


thermoska:
Hello

Would it be possible to support router RT2600ac by synocomunity ?
This router has a bump up in specs from previous version

I have seen some standalone recompiling here:
https://synopackages.wordpress.com/2016/02/21/synocommunity-packages-for-rsm/

and small thread here:

https://forum.synology.com/enu/viewtopic.php?t=114206

Thanks in advance

PS: I do have a root access to OS there. If it will help the community to support and migrate packages to SRM. I can provide information from this OS.

CoskunSunali:
I would also like and wish to see the support for the RT2600ac which has higher specs than RT1900ac, including but not limited to CPU and memory.

> I have seen some standalone recompiling here:
https://synopackages.wordpress.com/2016/02/21/synocommunity-packages-for-rsm/

The newer versions of those packages (for RT2600ac) are here:
https://synopackages.wordpress.com/2017/04/23/synocommunity-packages-for-srm-1-1/

ymartin59:
@publicarray According to https://www.synology.com/en-global/knowledgebase/SRM/help/SRM/PkgManApp/configure it is also possible to register a package source to SRM.

I have published Nano 2.9.5 for `ipq806x` in SynoCommunity repository with a firmware version `1.1-6931`.
I have no idea if SRM can grab it. May anyone please test it and report feedback?

publicarray:
Sigh, the documentation is wrong. There is no "Package Sources" tab: 
![screen shot](https://user-images.githubusercontent.com/5497998/38782789-fef9b3b0-413b-11e8-8759-d9bf06ad7af5.png)

ymartin59:
@publicarray Is it worth to search Synology forums or submit a support request to Synology about it? ... for instance in case it is an experimental feature that may be enabled? (I doubt they may release an update only for that reason...)

publicarray:
I imagine they already know, but just in case I've send an email with a reference to this issue.

publicarray:
Their reply:

> Sorry for the confusion.
>
>For SRM there's no "package source" tab, and it is wrong information on our website.
>
>Thank you for bringing it to our attention, and we've informed related department.
>
>Sorry for the inconvenience.

ymartin59:
Sounds relevant to publish SRM packages too in repository, so a manual download is required: https://github.com/SynoCommunity/spkrepo/issues/13

ymartin59:
[...]

PR #3268, opened 2018-04-02
Add ipq806x architecture for SRM 1.1

publicarray:
_Motivation:_ Make it easier to get started and build for the RT2600ac Router (ipq806x).
_Related issues:_ #2951

Not sure if this is wanted due to it being listed as [No longer recieving packages updates](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#models-no-longer-recieving-packages-updates). But I thought a pull request at least gives visibility and can help people if they want to do it themselves.

### Checklist

### DIFF
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -34,7 +34,7 @@ ARCHS_DUPES = $(filter-out apollolake% avoton% braswell% broadwell% bromolow% ce
 
 # Available Arches
 ARM5_ARCHES = 88f6281
-ARM7_ARCHES = alpine armada370 armada375 armada38x armadaxp comcerto2k monaco hi3535
+ARM7_ARCHES = alpine armada370 armada375 armada38x armadaxp comcerto2k monaco hi3535 ipq806x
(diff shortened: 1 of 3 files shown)


ymartin59:
@publicarray I agree to include SRM support...
@hgy59 May you please check if binaries produced for `ipq806x` arch may be "aggregated" together the same way with virtual `x64`?

hgy59:
@ymartin59 I don't think that `ipq806x` arch can be aggregated. 
Synology provides different packages for `ipq806x` (and `northstarplus`) too, and does not merge it with other ARMv7 packages.

Safihre:
I was the one that listed it as `No longer recieving packages updates.` because we didn't support it at the time. Sorry for that bad wording, probably good to support it now.
Looks good to me.

publicarray:
@Safihre I think the wording is fine :) because technically the only way to get packages onto the SRM is to manually select and upload a spk file. Sadly there is no way to add "Package Sources" onto the SRM. I just wasn't sure if the project would support toolchains that could not be used via "Package Sources"

Issue #3274, opened 2018-04-06
[Package Request] rsnapshot (synology_Apollolake)


antonyk72:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name: rsnapshot
_Package Version: v1.4.2-1 (seems to be the latest on SynoCommunity)

_NAS Model: DS218+
_NAS Architecture: Apollolake  (Intel Celeron J3355)
_DSM version:  6.1.6-15266

I've tried out Active Backup for Server, but it just doesn't cut it compared to rsnapshot. ymartin59 was good enough to create a package for mooks42, I'm hoping I might be lucky too! Would be a huge help. Many thanks in advance.

ymartin59:
There is many small things to improve in `rsnapshot` package... so I apply them even if not recent release is available for a really update.

antonyk72:
Hi Yves, thanks for your message - I'm not quite sure what you mean though? Are you saying you will release and older version for Apollolake? Or there is work to be done before making any new releases? I would be happy with any release for Apollolake! Thanks, Antony.

ymartin59:
@antonyk72 Here is a test package: 
[rsnapshot_noarch-all_1.4.2-2.spk.zip](https://github.com/SynoCommunity/spksrc/files/1900409/rsnapshot_noarch-all_1.4.2-2.spk.zip)
As I find original configuration does not fit DSM well, probably you may contribute your own tuned configuration so that package user has some guidance.

antonyk72:
Hi Yves, thank you so so much for providing the test package! Greatly appreciated. I have been setting it up over the last few days, and it is working perfectly so far. I have installed the package on the Synology DS218+ and can do local rsnapshots. I can also connect to a remote RHEL server and do remote rsnapshots via SSH. I'm not very experienced so am taking my time to learn and test everything properly. I will be pleased to contribute my "tuned configuration". What is the best way for me to contribute this? (sorry for my newbiness!) Thanks again, Antony.

ymartin59:
I propose you to:
- comment standard options/path that do not fit DSM
- enable DSM specific options when required
- document recommended options for DSM (user only have to uncomment lines)
- provide commented example setup with DSM standard paths and "usually available location" like "/volume1/downloads"... users have to uncomment and adapt
If you are not a Git user, please simply upload your configuration here as attached file to a comment.

antonyk72:
Hi Yves, is it too late for me to provide the above .conf file? 

Thanks, Antony.

[...]

PR #3281, opened 2018-04-11
Refactor rsnapshot as no-arch

ymartin59:
_Motivation:_ Refactor with generic installer as no-arch package
_Linked issues:_ #3274 

### Checklist


ymartin59:
Waiting for requester's feedback

Safihre:
So for my understanding: it's a Perl-application and can be ran without anything else needed? Since there's still the `make install` in the `cross/` package?

ymartin59:
I was waiting for a DSM specific configuration file but no answer from user yet. I will publish as-is.

PR #3292, opened 2018-04-25
New package: dante-sockd-1.4.2

Anonym-tsk:
_Motivation:_ Sockd is lightweight and simple socks proxy server with authentication support

### Checklist


ymartin59:
Thanks for contributing... sockd package is available in Entware https://github.com/Entware/Entware and I would recommend this source instead of extending again spksrc volume.
Have you got reasons to motivate this package addition?

Anonym-tsk:
@ymartin59 Entware is difficult to install, difficult to configure, and it breaks after every DSM update.
I think that installing spk from the syncommunity repo is the best way.
It gives easy installation and configuration with wizard.

Anonym-tsk:
@ymartin59 @Safihre Do you have any comments on this PR? What are we waiting for?

ymartin59:
Dante seems to provide two components: server (sockd) and client.
As only "sockd" is delivered (and seems too generic as a name), I would rename package as `dante-sockd`
In that case, rename folders too as `cross/dante-sockd` and `spk/dante-sockd` for consistency.
If you prefer to provide both server and client, then package can be named as `dante` (so that corresponding folders)

Anonym-tsk:
@ymartin59 Ok, I renamed it

Anonym-tsk:
@ymartin59 @Safihre How i can speed up this review?

Safihre:
@ymartin59 Technically it looks good to me.

cross/dante-sockd/Makefile
```diff
@@ -0,0 +1,17 @@
+PKG_NAME = dante
```
ymartin59: Please be consistent with folder name and package name. So `dante-sockd`

spk/dante-sockd/src/sockd.conf
```diff
@@ -0,0 +1,39 @@
+# sockd configuration
+# https://www.inet.no/dante/doc/1.4.x/sockd.conf.5.html
+
+logoutput: @socks_log@
+
+internal: eth0 port = @socks_port@
```
ymartin59: I wonder if configurable "socks_port" should be opened in firewall.
If so, it may be interested that `service_preinstall` generates `conf/dante-sockd.sc` so that it can be registered as a known service port in DSM: https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.installer#L245
What do you think about it?

Anonym-tsk: I think, if you are using firewall, you can open port manually, if needed. It's not a problem.

Anonym-tsk:
@ymartin59 thank you. Please merge this pull request, i can't do it.

ymartin59:
May you please document "quick start" usage as a wiki page, typically how to open port in firewall if needed.

Anonym-tsk:
@ymartin59 I will do it.

Issue #3246, opened 2018-03-24
Python 2.7.14 fails to load for ARM


ymartin59:
@Safihre I would like to track this specific issue here to get it fixed

Related issues: #3201 

<<
It's not clear to me if it's in `cross/cryptography` or in `cross/openssl`.
The compilation of `cross/cryptography` is supposed to pick-up the same flags as Python (for example -`pthread`), but for some reason it doesn't.
>>

References:
- https://github.com/pyca/cryptography/issues/3344

My understanding is that `cryptography` wheel build from `requirements.txt` dynamically links with DSM original `openssl` library (.so) instead of package specific version - due to lacking `rpath` option in binary.

My proposal is to either be able to pass arguments to pip when building wheels, or to get back `cross/cryptography` you have removed when upgrading.

Safihre:
I already re-added the `cross/cryptography` in all the test builds that I created for people #3201. It doesn't work.
But at least then you can see the compilation steps (it doesn't show when using the `requirements.txt`) which shows that it's not picking up the same flags as the compilation of Python.
The error is specially in the `_openssl` of the Cryptography files, not in Openssl itself it seems to say in the errors. The `_openssl.c` file gets dynamically created when `setup.py` is run.
I tried forcing `-pthreads` everywhere using the flags, both in Python and in the cross compile of Python packages but the `setup.py` just wasn't picking them up.
It should be picked up from the `_sysconfigdata.py` that's generated during Python compilation. But even when I get those flags in there during Python compilation, it ignores them during the module building.

Safihre:
Or maybe indeed it needs the 'rpath`, didn't think of that! I also don't see that being picked up during module compilation. 
Not an expert really in these compilation things :/

ymartin59:
My understanding:
- there is no trouble at compilation, custom "openssl" is used and expects symbol "pthread_atwork"
- at runtime, libraries fails to find symbol probably because linking with DSM openssl

To confirm with a test case, you may try to run Python with `LD_LIBRARY_PATH` to `/var/packages/python/target/lib`. If failure wanish, then adding `RPATH` to cryptography binary is the right way to fix

ymartin59:
@Safihre Have you got any "simple" test case to run as basic "python" command line to reproduce trouble?

Safihre:
[...]

PR #3305, opened 2018-05-05
python 2.7: fix cryptography wheel for ARM

ymartin59:
Add WHEELS_CFLAGS and WHEELS_LDFLAGS to pass additional cross-compilation options to "pip wheel"

_Motivation:_ Update and fix cryptography wheel
_Linked issues:_ #3246 #3201

### Checklist

### DIFF
--- a/mk/spksrc.wheel.mk
+++ b/mk/spksrc.wheel.mk
@@ -55,14 +55,14 @@ build_wheel_target: $(PRE_WHEEL_TARGET)
 		$(foreach e,$(shell cat $(WORK_DIR)/python-cc.mk),$(eval $(e))) \
 		if [ ! -z "$(CROSS_COMPILE_WHEELS)" ] ; then \
 			$(MSG) "Force cross-compile" ; \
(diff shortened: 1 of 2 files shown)


Safihre:
I/we can compile it for all archs on Travis and let some users test it? :)

ymartin59:
@Safihre I prefer to engage that process with 2.7.15 update !

ymartin59:
@Safihre I changed my mind. I will publish Python 2.7.14 package with fix "as is", as many users wait for it to get applications run properly

Safihre:
Of course! Great you found it and that the fix is relatively easy!

Safihre:
One last note is that there is a "core" problem of the `_sysconfig.py` settings not being picked up during compilation of wheels. Because the compilaton of the cross-compiled (for target) Python will detect that these extra flags are nessecary and save the settings in that file. But for some reason the `cryptography` install just ignores them. 
If we get this problem (or other flags missing) another time we should investigate why it is just not picking up the Python-compile-detected flags. For now this works 👍

PR #3316, opened 2018-05-12
Translations and permissions documentation link

ymartin59:
_Motivation:_ add link to Permissions Management FAQ and french translations
_Linked issues:_ 

### Checklist


Safihre:
Great. 
Maybe also add the other ones from my `python-packs` collection and get these releases out?
https://github.com/Safihre/spksrc/releases/tag/python-packs

Safihre:
I think we should also update `nzbget-testing`  so we never have to think about them again. Clearly the `nzbget` package is working very well. Just need to add the new translations/links 👍 
Only minor difference in the packages is that the `testing` one by default downloads the beta-release of NZBGet.

PR #3330, opened 2018-05-30
Update SABnzbd to 2.3.4

Safihre:
### Checklist
- [x] Build rule `all-supported` [completed successfully](https://travis-ci.org/Safihre/spksrc/builds/385533764)

### DIFF
--- a/cross/sabnzbd/Makefile
+++ b/cross/sabnzbd/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = SABnzbd
-PKG_VERS = 2.3.3
+PKG_VERS = 2.3.4
 PKG_EXT = tar.gz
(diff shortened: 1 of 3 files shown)


ymartin59:
@Safihre Is there a way to get sabnzbd "no-arch" package? If "sabyenc" is the only native dependency, why not move its wheel to python 2 package too?

Safihre:
@ymartin59 There is also `unrar`, `7zip` and `par2` that need native, so not really possible to make `no-arch`.

ymartin59:
@Safihre OK. So maybe when these tools will be delivered in a common dedicated "files" package...

Safihre:
@ymartin59 We discussed that before, but didn't do that because the only package that uses these dependencies is SAB and NZBGet (only `unrar` and `p7zip`), so not really worth the effort it seems.
Plus we want to update `sabyenc` from time to time (few minor releases in past months), which would then require updates of the whole Python package.

PR #3346, opened 2018-06-17
[WIP] ffsync - overhaul and compatibility with DSM6

vigri:
_Motivation:_ Make ffsync compatible with DSM6 and the new generic service layout. Moreover, the goal is to make ffsync compatible with mysql and mariadb 10 by extending the wizard interface to allow the user to enter custom port numbers etc.

As suggested by @ymartin59 I've created a PR for my current WIP regarding ffsync and a problem I'm facing.
Lot of work still needs to be done.

**The problem**
When installing the package I get a "not available" message.
https://imgur.com/LrhmgWD

**Synopkg.log**
2018/06/17 13:24:13     install ffsync 1.5.2-4 Begin /bin/rm -rf /volume1/@tmp/pkginstall
2018/06/17 13:24:13     install ffsync 1.5.2-4 End /bin/rm -rf /volume1/@tmp/pkginstall ret=[0]
2018/06/17 13:24:13     install ffsync 1.5.2-4 failed

I've been able to isolate the problem:
If I comment the first three lines or / and the last line out (makefile) the installation works.

```
#SERVICE_USER = auto
#SERVICE_PORT = 8192
#SERVICE_PORT_TITLE = ffsync(TCP)
#FWPORTS = src/${SPK_NAME}.sc
```

@ymartin59 told me that the error could be caused by a syntax error inside /src/service-setup.sh
But if that would be the case the installation should fail regardless if I comment out the 4 lines above or not....

Maybe someone can give me a hint how to solve this error?

Thanks and best regards

### Checklist


ymartin59:
Hello. "service-setup" is far from previous "dsm-control" script. Many common variables and tasks are handled in "generic" installer script: https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.installer
Functions "service_xxx" are only "hooks" for package specific stuff.

You can inspire from: https://github.com/SynoCommunity/spksrc/blob/master/spk/headphones/src/service-setup.sh

You should look at package "scripts/" content, set some variables like https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.testcase  to run shell script with "bash -x ./preinst" for instance.

MySQL database handling is sure a complex part for ffsync. My preferred way would be to use DSM 6 "resource" framework, but it will require users to migrate their database... I keep that option for brand new package only. Reference: https://developer.synology.com/developer-guide/resource_acquisition/maria_db.html

ymartin59:
Thanks for contribution. I am working on it as #3849

PR #3352, opened 2018-06-18
WIP: Add emacs 26.1

timvisher:
Problems
--------

- Though I am depending on ncurses, I get an [error during the build][].
  Some variable appears to not be being set to allow emacs to know where
  to find libncurses.so.6

  ```
  ./temacs --batch  --load loadup bootstrap
  ./temacs: error while loading shared libraries: libncurses.so.6: cannot open shared object file: No such file or directory
  ```

This feels like something I'm doing wrong here rather than something
that I'm not understanding about the emacs build process.

[error during the build]: https://gist.github.com/bc2855136b3f9933599d1368c5892325

_Motivation:_ Adding emacs 26.1, need help with `cross-cc` tooling.

### Checklist


timvisher:
ping :)

Safihre:
After it fails, did you try to find `libncurses.so.6` file on your system in the `spksrc` folder after it stops? You probably need to somehow tell it where to look for linking. 
I'm not a true expert on this myself, but managed by tweaking some of the `FLAGS` or command-line parameters.

cross/emacs/Makefile
```diff
@@ -0,0 +1,17 @@
...
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS = cross/gnutls cross/ncurses
+
+HOMEPAGE = https://www.gnu.org/software/emacs/
+COMMENT  = Emacs is the extensible, customizable, self-documenting real-time display editor.
+LICENSE  = GPLv3
+
+GNU_CONFIGURE = 1
+CONFIGURE_ARGS = --without-x --with-gnutls --prefix=/usr/local
```
ymartin59: Please remove `prefix` option. Framework set it already at expected location

spk/emacs/Makefile
```diff
@@ -0,0 +1,26 @@
...
+RELOAD_UI = no
+DISPLAY_NAME = Emacs
+STARTABLE = no
+BETA = 1
+CHANGELOG = Add emacs 26.1
+
+HOMEPAGE   = https://www.gnu.org/software/emacs/
+LICENSE    = GPLv3
+
+INSTALLER_SCRIPT = src/installer.sh
```
ymartin59: These two lines and corresponding files can be removed and replaced by generic installer support - controlled by `STARTABLE = no`

spk/emacs/Makefile
```diff
@@ -0,0 +1,26 @@
...
+SSS_SCRIPT       = src/dsm-control.sh
+
+INSTALL_PREFIX = /usr/local/$(SPK_NAME)
```
ymartin59: As already default value, please remove `INSTALL_PREFIX`

spk/emacs/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+POST_STRIP_TARGET = tmux_extra_install
```
ymartin59: Please discard if no `emacs_extra_install` target is useful yet.

timvisher:
I applied all the changes you asked for. Sadly it didn't resolve my problems. FWIW, it appears that emacs is maybe [using a non-standard method of determining where to look for ncurses](https://git.savannah.gnu.org/cgit/emacs.git/tree/configure.ac?h=emacs-26.1#n4023), but I've been unable to locate any instructions on libncurses specifically regarding emacs builds.

I've picked apart the `configure` run command to look for potential errors in the `LDFLAGS` variable but it all looks good to me and the directory it mentions indeed contains `libncurses.so.6`.

I've reached out for help on the mailing list so maybe someone there will be able to help me out.

timvisher:
Someone got back to me on [the emacs list](https://lists.gnu.org/archive/html/help-gnu-emacs/2018-08/msg00041.html). They think this has a somewhat dubious chance of success unfortunately. But if there's anything in their message that you can speak to please do. :)

timvisher:
Got some news! I was able to get the temacs step to run by hand using the advice that Noam gave me on the mailing list. Can you give me any ideas how I could work this into the spksrc build process?

```
vagrant@debian-8:~/spksrc/spk/emacs/work-apollolake-6.1/emacs-26.1/src$ LD_LIBRARY_PATH=../../install/var/packages/emacs/target/lib/ ./temacs --batch --load loadup
bootstrap
Loading loadup.el (source)...
Using load-path (/home/vagrant/spksrc/spk/emacs/work-apollolake-6.1/emacs-26.1/lisp /home/vagrant/spksrc/spk/emacs/work-apollolake-6.1/emacs-26.1/lisp/emacs-lisp /home/vagrant/spksrc/spk/emacs/work-apollolake-6.1/emacs-26.1/lisp/progmodes /home/vagrant/spksrc/spk/emacs/work-apollolake-6.1/emacs-26.1/lisp/language /home/vagrant/spksrc/spk/emacs/work-apollolake-6.1/emacs-26.1/lisp/international /home/vagrant/spksrc/spk/emacs/work-apollolake-6.1/emacs-26.1/lisp/textmodes /home/vagrant/spksrc/spk/emacs/work-apollolake-6.1/emacs-26.1/lisp/vc)
Loading emacs-lisp/byte-run...
…
Finding pointers to doc strings...done
Dumping under the name emacs
87479 pure bytes used
```

timvisher:
This also made it all the way to the stripping step.

```
vagrant@debian-8:~/spksrc/spk/emacs$ LD_LIBRARY_PATH=$PWD/work-apollolake-6.1/install/var/packages/emacs/target/lib/ make arch-apollolake
```

cross/emacs/PLIST
```diff
@@ -0,0 +1 @@
+bin:bin/emacs
```
ymartin59: You should include here all files required to be packaged with proper "type": https://github.com/SynoCommunity/spksrc/wiki/PLIST-files
So "lnk" has to be used for `bin/emacs`

timvisher: I will give that a try. Any ideas on the `LD_LIBRARY_PATH` problem?

timvisher:
With that change in place I was able to run the following without error on a clean tree.

```
vagrant@debian-8:~/spksrc/spk/emacs$ LD_LIBRARY_PATH=$PWD/work-apollolake-6.1/install/var/packages/emacs/target/lib/ make arch-apollolake
```

I'm unsure exactly what I need to do to verify that emacs can launch cleanly though, and I'm still not sure how to configure `LD_LIBRARY_PATH` correctly.

ymartin59:
You have to include required libraries or other resources in PLIST file so everything is embedded in package.
- Refer to documentation: https://github.com/SynoCommunity/spksrc/wiki/PLIST-files
- Inspire from ffmpeg: https://github.com/SynoCommunity/spksrc/blob/master/cross/ffmpeg/PLIST

It is expected you cannot run binary "as is" because dynamic linkage expects some specific location you may inspect with `ldd`.

timvisher:
I added 

```
lnk:lib/libncurses.so
lnk:lib/libncurses.so.6
lib:lib/libncurses.so.6.1
```

to the changeset and ran:

```
vagrant@debian-8:~/spksrc/spk/emacs$ make arch-apollolake
```

and failed again at the `temacs` step.

And again, `make clean` then

```
vagrant@debian-8:~/spksrc/spk/emacs$ LD_LIBRARY_PATH=$PWD/work-apollolake-6.1/install/var/packages/emacs/target/lib/ make arch-apollolake
```

worked. Any thoughts on how to inform spksrc that `LD_LIBRARY_PATH` needs to be set for this build to include the target lib dir or similar?

timvisher:
Ping. :)

timvisher:
Any new thoughts here?

m4tt075:
@timvisher Just looked into this, trying to help. Unfortunately, I've not been able to find a quick fix either. This does not mean anything. Others might be able to help you. In any case, I found some insights that cross-compilation of `emacs` might not be straight-forward in the answer linked here:
https://stackoverflow.com/questions/46746976/how-to-cross-compile-emacs-for-ios
Maybe(!), there is is way building `emacs` natively first (like for the `mono` package for instance), but frankly, not sure it is worth the effort, if you can use `nano` instead.

timvisher:
@m4tt075 Thanks for trying to help! In this instance I have basically no interest in emacs as an editor and much interest in emacs as an IRC client. :)

`nano` is fine for a readline based editor, as you pointed out, and I really don't want to take the plunge into `irssi`.

m4tt075:
@timvisher Ah, ok. Interesting. And `znc`?

timvisher:
@m4tt075 Yeah. I just really like my emacs set up. :)

https://xkcd.com/1782/

m4tt075:
@timvisher Ahhh, that is YOU!  *LOL*
Very well. You have my empathy! Good luck with your package. Maybe someone else can step in...

ymartin59:
I have found out what is wrong. After emacs binary is cross-compiled, build process goes on compiling lisp thanks to temacs binary - which may not be runnable on compile system (if for another architecture like ARM).
My option is that lisp compilation has to be postponed to package postinstallation where temacs binary is runnable - the same way Python compiles its libaries https://github.com/SynoCommunity/spksrc/blob/master/spk/python3/src/service-setup.sh#L18

ymartin59:
References: https://stackoverflow.com/questions/46746976/how-to-cross-compile-emacs-for-ios
So there is chance a package can be created "easily" for x64 platforms but other architecture would require to tweak Emacs build process.

ymartin59:
Even try to go further building x64 for amd64 only adding `export LD_LIBRARY_PATH=/spksrc/spk/emacs/work-x64-5.2/install/var/packages/emacs/target/lib/`
but then "dumping" complains:
```
Dumping under the name emacs
**************************************************
Warning: Your system has a gap between BSS and the
heap (12297400 bytes).  This usually means that exec-shield
or something similar is in effect.  The dump may
fail because of this.  See the section about
exec-shield in etc/PROBLEMS for more information.
**************************************************
```
Because building in Docker context if I guess right.

hgy59:
closing due to lack of support for cross compilation (see https://stackoverflow.com/questions/46746976/how-to-cross-compile-emacs-for-ios)

Issue #3334, opened 2018-05-31
[Update Request] borgmatic 1.1.15 (part of borg)


kBite:
`borg` `v1.1.5-2` bundles `borgmatic` version `1.0.3`, latest is `1.1.15`. Can someone please update the package?

I was able to successfully build `borgbackup_rtd1296-6.1_1.1.5-2.spk` with Docker, so bumping `borgmatic` version seems uncomplicated:

```
diff --git a/spk/borgbackup/src/requirements.txt b/spk/borgbackup/src/requirements.txt
index 93406ef..2136059 100644
--- a/spk/borgbackup/src/requirements.txt
+++ b/spk/borgbackup/src/requirements.txt
@@ -3,4 +3,4 @@
 #msgpack-python==0.4.8
 
 # Downloaded dependencies
-borgmatic==1.0.3
+borgmatic==1.1.15
```

ymartin59:
@yuroller Hello Yuri. May you please propose a PR with borgmatic update? Thank you in advance.

PR #3355, opened 2018-06-21
borgbackup: update to 1.1.6

yuroller:
_Motivation:_ Update borgbackup to 1.1.6
_Linked issues:_ #3334

The update requires update to python-msgpack, so I had to modify also python and python3 packages that are using msgpack.

### Checklist

Little testing on DS212+ with DSM-6.0.2:
* verify data of a backup repo done with previous version of borgbackup
* append new data to backup repo

### DIFF
--- a/cross/borgbackup/Makefile
+++ b/cross/borgbackup/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = borgbackup
-PKG_VERS = 1.1.5
+PKG_VERS = 1.1.6
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://files.pythonhosted.org/packages/source/b/$(PKG_NAME)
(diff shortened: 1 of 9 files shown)


ymartin59:
@yuroller I think you missed to update borgmatic version in `requirement.txt`, or do I mistaken?

yuroller:
requirements.txt has been updated, but just for reference, because it mostly contains comments

ymartin59:
@yuroller Please check this line `borgmatic==1.0.3` in requirements.txt... Requester expects a more recent version.

ymartin59:
Thanks for details. From my point of view, framework "wheels" support should do the trick... the only required change is to remove the "no-arch" switch. Let's have a try, it should not be long. If so, no need to split packages for borgbackup/borgomatic.

yuroller:
Sorry, but I am very busy in this period: I will try to update borgmatic as soon as I can.

yuroller:
I was able to upgrade to latest (=1.2.1) borgmatic package which depends on ruamel.yaml python package.
I had to create a cross package configuration for ruamel.yaml, because it includes a C language extension, then add it as a dependency to python3 package.
Tomorrow I try to clean a bit the modifications and do some more tests, then I will propose a pull request.

Safihre:
@yuroller Since borgbackup package is already arch-specific, why add the new dependency to the `python3` package? 
That would require _everyone_ to update their Python 3. You should add it to the `borgbackup` package. Just like `sabnzbd` does.

yuroller:
Till yesterday I was not able to correctly have the cross compiled version of the ruamel.yaml extension module (dependency of borgmatic), but looking at sabnzbd package I spotted a difference: I am missing 

Probably if I add that line to borgbackup Makefile, the extension module gets built correctly.
I am trying this evening and report back.

yuroller:
See #3418

PR #3363, opened 2018-06-28
Improve package INFO generation

### LINKED ISSUE #2966: Support for DSM 6.2 (firmware version): jackett...
Probably additional behaviors has to be added in synospk repository code.

ymartin59:
Base of SDK Developer Guide updates.

Resolves #2966

Idea is to generate DSM 5.2 package also compliant for DSM 6.x, including both `firmware` and `os_min_ver`, when building with 5.2 toolchain

### DIFF
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -126,12 +126,18 @@ else ifneq ($(strip $(OS_MIN_VER)),)
 	@echo os_min_ver=\"$(OS_MIN_VER)\" >> $@
 else ifneq ($(strip $(TC_FIRMWARE)),)
 	@echo firmware=\"$(TC_FIRMWARE)\" >> $@
+	@echo os_min_ver=\"$(TC_FIRMWARE)\" >> $@
 else ifneq ($(strip $(TC_OS_MIN_VER)),)
 	@echo os_min_ver=\"$(TC_OS_MIN_VER)\" >> $@
 else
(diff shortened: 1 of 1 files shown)


ymartin59:
@publicarray @BenjV @Safihre What do you think about it?

publicarray:
If this will fix things go ahead but synology doesn't like having both `firmware` and `os_min_ver` in the INFO file at the same time, It's one or the other.

Safihre:
> If this will fix things go ahead but synology doesn't like having both firmware and os_min_ver in the INFO file at the same time, It's one or the other.

We know that for sure?

ymartin59:
I consider it as "Synology (developers) do not like having both"... but DSM Package Center probably does not care. I will do tests with sickbeard-custom to confirm that both DSM 5.2 and DSM 6.2 agree.

publicarray:
> Synology (developers) 

Yea good point and that is what I should have to said.

BenjV:
I have tested an installation of a package with both **firmware** and **os_min_ver** in the INFO file under DSM 6.2 and it worked fine.
This can be put in the INFO file and it worked flawless

> firmware="3.1-1594"
> os_min_ver="3.1-1594"
> 
> 

I did not test it on lower DSM versions, but normaly the package centre ignore everthing it does not recognize so I expect not problems there.

Safihre:
I think something broke, all my builds fail: https://travis-ci.org/Safihre/spksrc/jobs/398592185

Issue #3251, opened 2018-03-25
Design: new package "SynoCli disk"


ymartin59:
Create a package for small command line utilities dedicated to disk/file system manipulation.
Revival of #2261
Tools candidates:
- e2fsprogs (from testdisk)
- testdisk (replace standalone package)
- ncdu #2226
- davfs2 #1974
- lsscsi #2270 
- fuse [cross package](https://github.com/SynoCommunity/spksrc/tree/master/cross/fuse)

Any other ideas?
Wanted contribution: an generic icon for "SynoCli" package group, with "disk" text addition for category

PR #3364, opened 2018-06-28
DAR Disk ARchive

spacifici:
_Motivation:_ adding DAR, the Disk ARchive. dar is a shell command that backs up from a single file to a whole filesystems, taking care of hard links, Extended Attributes, sparse files, MacOS's file forks, any inode type (including Solaris Door inodes), etc.

### Checklist


ymartin59:
Thanks for contributing. This tool is probably interested for future "SynoCli disk" package: https://github.com/SynoCommunity/spksrc/issues/3251

spacifici:
It doesn't seem a bad idea, do you need help with #3251? Maybe later I will have time to draft something out.

Safihre:
We would really like help with #3251! :)

hgy59:
@spacifici thanks again for your great work.
DAR is now integrated into synocli-disk and available as single package under /diyspk/dar.

Issue #3387, opened 2018-07-17
Improve start-stop-status status


publicarray:
Specificity the following exit status codes should be used when running `start-stop-status status`

> 0: package is running.
> 1: program of package is dead and /var/run pid file exists.
> 2: program of package is dead and /var/lock lock file exists
> 3: package is not running
> 4: package status is unknown
> 150: package is broken and should be reinstalled. Please note, broken status (150) is only supported by DSM 4.2 and later.

https://developer.synology.com/developer-guide/synology_package/scripts.html

Currently we exit with 0 (package is running) or 1 (package is not running)

BenjV:
Actually the return code should be 0 or 3

Status 1: does not occur, because the start-stop-status script removes the pidfile when the process is no longer running because various applications do not start if a (old)pidfile is already there.
Status 2: does not exists, no lockfile is used.
Status 4: does not exists it is either running(0) or not running(3).
Status 150: is not possible to determine in the start-stop-status script.

publicarray:
Agreed but that’s not what I think our current script does

ymartin59:
Thanks for reviewing documentation. I will fix generic scripts:
- `spksrc.service.start-stop-status`
- `spksrc.service.start-stop-daemon`
- `spksrc.service.non-startable`

PR #3390, opened 2018-07-19
framework: fix busybox start-stop-daemon support for service

ymartin59:
- discard non relevant options, add pid file
- run as root in conf/privilege for DSM 6
- fix status return codes

_Linked issues:_ #3387 #3094


Safihre:
So I understand correctly that the problem was that it doesn't run as root?

ymartin59:
Not only. `HOME` environment variable was not set even with use of `env`. So my usage of `export` in `service_prestart`.

Safihre:
:+1:

ymartin59:
Sorry. I tried to publish two PRs at the same time and one has just "vanished" (don't know where/why) and explain why you cannot understand what was behind. So I now have to rebase your branch from #3094 with my changes for review...

PR #3391, opened 2018-07-19
Fix syncthing startup and use default webui port

ymartin59:
Minor changes: busybox start-stop-daemon usage and translations
Depends on #3094 and #3390


acolomb:
@Safihre maybe I should create separate issues for these, but since you seems to be currently working on the Syncthing package, I wanted to ask first whether you have this on your radar:

1. Apparently your new version uses the `*_ARCHES` variables correctly, so the missing build for rtd1293 (DS418j) I've been waiting for should arrive when this new version gets published?
2. Current Syncthing upstream version is 0.14.48 as of writing.  Will you update to that before this PR gets published?  Or should I try to dig into spksrc and prepare a separate PR?
3. According to https://docs.syncthing.net/users/firewall.html, the standard GUI port is TCP/8384.  Do you have any intentions to switch from the current TCP/7070 for new installs on DSM?
4. The same document states that 22000 (sync) and 8384 (GUI) are TCP, while 21027 (discovery) is UDP.  That is not how I read the current list in `spksrc/spk/syncthing/src/syncthing.sc`.  Does it allow to be more precise?
5. The Syncthing GUI uses its own self-signed SSL certificate.  Can that be integrated into the DSM Certificate management GUI, where you can choose among installed certificates for different services?  Or at least point it at the system "default certificate".

Thanks in advance for answering these questions.  If you'd like, I have a DS418j and a DS218j, both up-to-date on DSM 6.2 available for testing.  Haven't got around to learn building spksrc packages myself yet, but I would gladly help you in testing your builds of Syncthing.

Safihre:
@acolomb I only did a quick conversion to the new DSM5+6 service framework and the fix of your point 1.
The rest will have to be corrected after that. The ports is something that I just copied from the existing package, I actually know very little about that.

acolomb:
I could not get the package to work yet with this PR.  Setup Docker, built using `make arch-rtd1296-6.1` and installed from the built spk file.  DSM Version: DSM 6.2-23739 Update 2 (DS418j, rtd1293)

It does not start from GUI, always remains at "Stopped" state.  When trying to start via SSH, I get:

```
admin@restbox:/var/packages/syncthing/scripts$ sudo ./start-stop-status start
./service-setup: line 6: [: : integer expression expected
./start-stop-status: line 43: start-stop-daemon: command not found
Starting  ...
./start-stop-status: line 27: start-stop-daemon: command not found
```

Seems like `SYNOPKG_DSM_VERSION_MAJOR` is not defined, and the start-stop-daemon command is missing.  What am I missing?  Do I need to install additional tools on the DiskStation?

Safihre:
You also can't just call the start script anymore from the commandline. You have to use `synopkg` command.

acolomb:
Thanks for the hint.  Then https://github.com/SynoCommunity/spksrc/blob/master/CONTRIBUTING.md#issue-content needs updating I guess :-)

ymartin59:
@acolomb Sure. I have created a specific FAQ entry: https://github.com/SynoCommunity/spksrc/wiki/Frequently-Asked-Questions#how-to-query-package-status-from-command-line

acolomb:
@ymartin59 Sorry for not responding earlier, I've been busy.

I just tested the published package on both my DS418j and DS218j, and it works perfectly after adding setting it up under Settings > Privileges.

Thanks for your work on this!

agsfbm6:
@acolomb I am a not very linux literate user and am trying to get Syncthing running on my Synology DS213+, DSM 6.2. After installing the Syncthing package via DSM package manager I still do not manage to get Syncthing to run via the DSM GUI. Under Control Panel -> Privileges I edited Syncthing application privileges and allowed group 'sc-syncthing'. There is no specific Syncthing user set up. Still I cannot get the service to start. Could you advise how to proceed? When I click on the Syncthing Icon it tries to open the webui at port 8384, I thus guess that the latest version with the above discussed changes is installed? Installed version is displayed as 0.14.49-15.

Thank you guys for your support!

acolomb:
Did you also hit "Run" from the Package Center page? What is the status saying there? 

If you cannot reach the Web GUI, check if your firewall is enabled. If so, you can choose the appropriate port from the predefined application list.

Have you run the previous package version before? The package doesn't update Syncthing's config for the new GUI port. Try reaching it over the old default port 7070 and change the GUI listen address within its settings to the new port 8384, then restart Syncthing.

agsfbm6:
> Did you also hit "Run" from the Package Center page? What is the status saying there?

Yes. It flickers for ~10sec and states 'Loading...' but status remains unchanged.

> If you cannot reach the Web GUI, check if your firewall is enabled. If so, you can choose the appropriate port from the predefined application list.

Firewall is running and I allowed both the webui and sync ports.

> Have you run the previous package version before?

No. I did a first clean install via DSM packet manager yesterday. UI cannot be reached via old port either but I think the cause is rather that the service just doesnt start.

I restarted the diskstation and am logged in as administrator. Syncthing didnt start automatically and cannot be started manually.

PR #3411, opened 2018-08-06
New package - iOS Access (udev scripts to mount iOS devices)

bxxxjxxg:
Motivation: create a package to mount and browse an iOS device directory in FileStation.

Developing package by spksrc is quite happy experience here.
Now, I’d like to submit the second package. 
It’s derivative work of first package to interact with FileStation.

### Checklist

This package depends on itools (libimobiledevice). It provides
udev rules and scripts to detect the hotplug events of iOS devices
and run python scripts to mount/umount into/from FileStation.

Signed-off-by: BingJing Chang <bxxxjxxg@gmail.com>


ymartin59:
Hello
Thanks for your contribution.
I would really prefer you include these scripts and service into "itools" package itself.

By the way, `dsm-control.sh` and `installer.sh` are old stuff now. Framework now provides generic installer and generic service support to run application as non-privileged user. 
Refers to:
- demoservice example: https://github.com/SynoCommunity/spksrc/tree/master/spk/demoservice
- documentation: https://github.com/SynoCommunity/spksrc/wiki/Service-Support
- umurmur as another example https://github.com/SynoCommunity/spksrc/blob/master/spk/umurmur/Makefile

You should create "service_postinst" function in a `service-setup.sh` script, setting `SERVICE_COMMAND` to point to your background application. If script is expected to run as `root` (which is not recommended) I have to propose a specific support in framework...

ymartin59:
Replaced by #3438

PR #3412, opened 2018-08-06
Syncthing: More granular firewall port definitions

acolomb:
_Motivation:_ Syncthing's service port definitions are too broad and mix up UDP and TCP.

Define an additional service for Syncthing's **local** discovery mechanism, which is pretty much orthogonal to the externally accessible port for the actual file synchronization.  Opening up the former should only apply locally and only to UDP.  The latter is TCP and must possibly be port-forwarded for the **global** discovery mechanism and direct connections to work.

Adjust the descriptions to match the official Syncthing terminology from https://docs.syncthing.net/users/firewall.html, https://docs.syncthing.net/specs/ and https://github.com/syncthing/syncthing/blob/master/etc/firewall-ufw/syncthing

_Note:_
~~No testing so far, as my time is limited and the Docker environment on Ubuntu 18.04 does not really work.  There's a lot of tweaking involved to make DNS work within Docker, and it kills DNS for the rest of my system along the way.  Please standby for version bump and testing, but do provide feedback to the idea in general.~~

### Checklist


ymartin59:
Thanks to have polished what I initiated...

acolomb:
As stated in the checklist above, I've now tested building the `all-supported` target successfully.  Changes have been verified after a package upgrade, but I did not do a clean installation to avoid losing my Syncthing config.  In this case I guess it does not make enough of a difference.

I don't have the local firewall enabled on my two NAS, so I didn't actually test whether the port rules work for the expected traffic.

ymartin59:
I propose to wait for 0.14.50 release so that it is worth for a new publish process

acolomb:
Syncthing will auto-update right after install anyway. But if you prefer to wait for the next release, I will push an update here when it's ready.

ymartin59:
Pending waiting for #3436 and update

acolomb:
@ymartin59 so are you working on #3436? From my side it was just to record the issue, I haven't got far with my trial and error experiments for the certificate resource config.

Safihre:
@acolomb  I think @ymartin59 indeed hopes you can fix the certificates ;)

acolomb:
I'll work on that when I find some spare time, but don't hold your breath for it.

Therefore I think this should be merged and released when Syncthing is updated, independent of #3436.

acolomb:
Just pushed the new upstream release from two days ago.  So from my side, this should be merge-ready.

I'm still trying to collect puzzle pieces for the certificate integration (#3436).  No ETA available so far :-(

ymartin59:
I propose to reuse logic from https://github.com/SynoCommunity/spksrc/blob/master/spk/gateone/src/service-setup.sh#L8-L18
and base on info at https://forum.syncthing.net/t/ssl-certificate-setup/2901
create symbolic links at expected locations in `/var/packages/syncthing/target/var` to point to system key and certificate.

acolomb:
Pushed new upstream release and tested after merging current master branch. `make all-supported` succeeded and the "aarch64" package seems to work on my DS418j after both clean installation and after upgrade.  Note that it used to get the "rtd1296" architecture package, but I think the "aarch64" is equivalent?

acolomb:
Might as well jump to the most current release when this gets published.

@ymartin59 Anything holding up merge & publish, that I could help with?

acolomb:
Now included in, and superseded by #3649.

PR #3447, opened 2018-09-07
Fix duply script in duplicity package

kukoarmas:
_Motivation:_ Fix an unneeded and erratic condition in duply script

The duply script included in the duplicity packages includes this condition:

if gpg_symmetric && ( [ -z "$GPG_PW" ] || [ "$GPG_PW" == "${DEFAULT_GPG_PW}" ] )

With the bundled bash, the evaluation of this condition is erratic. It could be rewrited to fix it (with [[]]), but the last check is not needed because it is checked some lines earlier

### Checklist


ymartin59:
@kukoarmas Have you submitted your fix upstream at Duply team? It would make senses instead of maintaining a specific patch here

hgy59:
not relevant for out repository.

PR #3449, opened 2018-09-07
Fix GnuPG cross-compile on rtd1296 and other ARM architectures

kukoarmas:
_Motivation:_ Fix GnuPG package cross-compile on ARM architectures

The Makefile in three GnuPG dependencies didn't define the correct **host** and **target** to cross-compile on several ARM architectures (and maybe non-ARM also)
The affected libraries where: gmp, libgpg-error and nettle

This PR fixes the Makefiles to allow cross-compile on (hopefully) all archs

### Checklist

### DIFF
--- a/cross/gmp/Makefile
+++ b/cross/gmp/Makefile
@@ -14,15 +14,15 @@ LICENSE  = LGPL
 include ../../mk/spksrc.cross-cc.mk
 
 GNU_CONFIGURE = 1
-ifeq ($(findstring $(ARCH), $(ARM_ARCHES)),$(ARCH))
+ifeq ($(findstring $(ARCH), $(ARM5_ARCHES)),$(ARCH))
+CONFIGURE_ARGS = --target=arm-unknown-linux-gnueabi --host=arm-unknown-linux-gnueabi --disable-assembly
+endif
+ifeq ($(findstring $(ARCH), $(ARM7_ARCHES)),$(ARCH))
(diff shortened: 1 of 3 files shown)


kukoarmas:
Done. Last commit removes specific arch and uses arch group variables
"make all-supported" completed successfully with these changes

hgy59:
@ymartin59 what was this PR for?
IMHO spksrc supports cross compilation **on** i686 only. `With GNU_CONFIGURE = 1`  the --build and --host arguments for configure are provided by the framework.
May be I missunderstand the the term cross compiling **on** and it is meant as cross compiling **for**.
I have no problem cross compiling the mentioned libraries without these --target and --host definitions.

If really `--target` must be defined we should use `CONFIGURE_ARGS += --target=$(TC_TARGET)`

th0ma7:
Hi @hgy59 and @ymartin59,

Note that with this recent PR https://github.com/SynoCommunity/spksrc/pull/4010 I might have fix this issue as I have updated much of the dependency list along with `gnupg` ...

kukoarmas:
> May be I missunderstand the the term cross compiling **on** and it is meant as cross compiling **for**.
> I have no problem cross compiling the mentioned libraries without these --target and --host definitions.

Yes, I meant crosscompile "for", sorry
Maybe it's already fixed in mentioned PR because last time I did it, it worked

PR #3452, opened 2018-09-07
Add safekeep 1.4.4 package

kukoarmas:
_Motivation:_ Add safekeep 1.4.4 package

SafeKeep is a python frontend to simplify backups using rdiff-backup

### Checklist


ymartin59:
Thanks for contributing. We have lots of packages already, so my idea would be to include "safekeep" in "rdiff-backup" package itself, or create a new "SynoCli Backup" package to group some backup utilities all together
By the way, I would only provide crontab file as an example and document how user has to register it with its own preferences.

kukoarmas:
For me it would be kind of confusing having the safekeep script included in rdiff-backup. Safekeep uses rdiff-backup to do the real job, so I would expect Safekeep to depend on rdiff-backup (or maybe even include it) but not the other way (rdiff-backup including safekeep), since you can user rdiff-backup by itself

BTW, for me it was kind of confusing finding the duply script included in the duplicity package

Also, a generic SynoCli can be less confusing, but would have many dependencies...

Also, where should I include documentation about the package?

hgy59:
closed due to inactivity
_and spksrc integration is really outdated_

PR #3459, opened 2018-09-15
Determine and build affected spks

chickenandpork:
_Motivation:_
(I think) @Safihre mentioned that it can be difficult to determine what SPKs are affected by a change.  This diff gives the build target and script support to determine the SPKs affected by the changes in the most recent commit.  The build target "all-affected-spks" leverages this to build all the SPKs that directly and indirectly use an edited file in `cross/*`, `native/*`, and of course `spk/*`.

Why aren't I building the world with this?  It's not a change to any shipping SPK, but a tool to make building them trivial.  Consider:

`docker run ... SynoCommunity/spksrc:latest -w /spksrc  make   all-affected-spks-from-master`

...I could just push all rebuilt SPKs, but that might be biting off too much at once.

@Safihre @ymartin59 looking for feedback

### Checklist
- [n/a] Build rule `all-supported` completed successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,3 +126,14 @@ setup-synocommunity: setup
 		-e "s|REPORT_URL=.*|REPORT_URL=https://github.com/SynoCommunity/spksrc/issues|" \
 		local.mk
 
+all-affected-spks-from-master:
+	for spk in $(git diff --stat upstream/master...HEAD | bash mk/build-target-determinator.bash | sort) ; \
+	do \
+	    env $(MAKE) $${spk}-all-supported ; \
+	done
+
+# "make python3-all-supported" causes "make all-supported" to run in that spk
+%-all-supported: spk/%/Makefile
+	cd $(dir $^) && env $(MAKE) all-supported
+
+
--- a/mk/build-target-determinator.bash
+++ b/mk/build-target-determinator.bash
@@ -1,0 +1,52 @@
+#!bash
+#
+# A fairly-reliable method of determining the SPKs that need to be rebuild based on the files that
+# changed in the last commit.  Only looks at cross/* native/* and spk/* because, well, mk/* might
+# be a bit harder, and toolchains/* would require a rebuild of the world.
+#
+# I did this as a script rather than inside a mk/spksrc.spk.mk target to ensure I'm not evaluating
+# this on every run, only when needed.
+
+# test via:  git diff --stat HEAD^^^ | bash mk/build-target-determinator.bash
+# expected output is a list of subdirs of spk dir
+awk '
+func getdeps(depdir) {
+	checked[depdir]++
+
+	# somewhat reliable shell-exec grep for Makefiles referring to what we have
+	command = "grep -l \"DEPENDS .*"depdir"\" {cross,native,spk}/*/Makefile"
+	while ((command | getline line) > 0) {
+		split(line,path,"/");
+		dep[path[1]"/"path[2]]++;
+	}
+	close(command)
+}
+
+BEGIN {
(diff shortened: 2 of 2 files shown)


ymartin59:
Thanks for this contribution. Really appreciated.

Getting information from the only last commit is a good starting point and your code is useful to produce packages after squashed PR merge to master.
In fact I am also interested in inspecting all commits up to "master" common ancestor, so that a PR submitted with multiple small commits can be validated the same way in CI platform.

@chickenandpork I should do the same IP process with my employer...

Safihre:
Cool! I remember there was another effort for a similar feature somewhere in the issues. Didn't someone else also wrote a script for this @ymartin59?

Makefile
```diff
@@ -126,3 +126,14 @@ setup-synocommunity: setup
 		-e "s|REPORT_URL=.*|REPORT_URL=https://github.com/SynoCommunity/spksrc/issues|" \
 		local.mk
 
+all-affected-spks-from-master:
+	for spk in $(git diff --stat upstream/master...HEAD | bash mk/build-target-determinator.bash | sort) ; \
```
chickenandpork: Sorry, I force-pushed a rebase rather than stack a commit below this.

Previously, this was just a `for spk in $(bash mk/build...); \` and did the `git diff` inside the script.  The suggestion on that was to look for a common revision between `master` branch and the PR branch, and determine the changed spikes from that.  As you can see, this script now does that.

mk/build-target-determinator.bash
```diff
@@ -0,0 +1,52 @@
+#!bash
+#
+# A fairly-reliable method of determining the SPKs that need to be rebuild based on the files that
+# changed in the last commit.  Only looks at cross/* native/* and spk/* because, well, mk/* might
+# be a bit harder, and toolchains/* would require a rebuild of the world.
+#
+# I did this as a script rather than inside a mk/spksrc.spk.mk target to ensure I'm not evaluating
+# this on every run, only when needed.
+
+# test via:  git diff --stat HEAD^^^ | bash mk/build-target-determinator.bash
```
chickenandpork: This "test via" comment might need edits -- this is how I test but it's now driven by the delta from the most recent diff common to both master and current branch.

chickenandpork:
Update: as noted in code, followed @ymartin59 's request that we consider the diff from where current branch forked from current master branch.  This should allow a PR Build to do the same.

There is a risk in `all-supported`: this build target is leveraged by `${spk}-all-supported` vut it might be a bit weak: I find that when we run this, the unsupported architectures are still tried, then fail, failure being ignored.  I think this needs to be strengthened in future to actually fail when the build target arch fails, but **only attempt** the architectures that are supported.  This would allow `all-supported` to confirm that they all actually build rather than blindly accepting that the supported arches built and the failures are all unsupported arches.  **CAVEAT** I could be misinterpreting something I'm only seeing on my end.

chickenandpork:
Perhaps what we need is not an `all-supported` but an `all-supported-until-failure` that changes this:
```
      for arch in $(sort $(basename $(subst -,.,$(basename $(subst .,,$(ARCHS_DUPES)))))) ; \
      do \
        $(MAKE) latest-arch-$$arch ; \
      done \
```

... into this :

```
      for arch in $(sort $(basename $(subst -,.,$(basename $(subst .,,$(ARCHS_DUPES)))))) ; \
      do \
        $(MAKE) latest-arch-$$arch || exit $$? ; \
      done \
```

...this way, a failure stands out and is harder to ignore.  Should also bubble up to automated tools.

If you merge this PR I'll gladly make the change to a special `all-supported` target that fails on failure.

chickenandpork:
This PR is too stale.

PR #3479, opened 2018-10-05
dnscrypt-proxy: update to 2.0.17

publicarray:
_Motivation:_ To update dnscrypt-proxy to 2.0.17 and fix some issues
_Linked issues:_ 

### Checklist


ymartin59:
@publicarray Is this PR ready to merge? Looks like OK for me.

spk/dnscrypt-proxy/src/service-setup.sh
```diff
@@ -29,44 +32,56 @@ pgrep () {
...
 }
 
 service_prestart () {
-    disable_dhcpd_dns_port "yes"
+    echo "service_preinst ${SYNOPKG_PKG_STATUS}" >> "${INST_LOG}"
+
+    # This fixes https://github.com/SynoCommunity/spksrc/issues/3468
+    # This can't be done at install time. see:
+    #  https://github.com/SynoCommunity/spksrc/blob/e914a32600e65f80131ae09913f1b6f6a2dd8b13/mk/spksrc.service.installer#L307-L319
+    chown root:root "${SYNOPKG_PKGDEST}/ui/index.cgi"
```
publicarray: @ymartin59 Yes I'm happy as long as you are ok with the chroot here ^

publicarray: Sorry for the delay, I've been working on fixing bugs. The CGI program uses [/usr/syno/synoman/webman/modules/authenticate.cgi](https://originhelp.synology.com/developer-guide/integrate_dsm/web_authentication.html) to check if the user is logged-in and `/usr/syno/synoman/webman/initdata.cgi` to check that the user is an admin (only on the SRM because I have not found `/initdata.cgi` on the DSM). As far as I can tell root is required to execute those CGI programs. On the SRM the permissions given here are the same as from the web server user (system). The DSM executes CGI programs differently but I haven't had the time to test all of the permission gotchas/differences between the 2 systems. Having root makes this easier. I think there must be a better way, but I havn't found it yet.

publicarray: Oh and `/usr/syno/synoman/webman/login.cgi` is also used for the CSRF tokens.

Issue #3468, opened 2018-09-24
dnscrypt-proxy - unable to Start Packages


Moonshinelunacy:
Hello,

I am running DSM 6 on a DS918+ with Apollo Lake architecture. When installing new packages, the URL to actually run the package does not appear in the package center window. I have tried uninstalling and reinstalling the packages, and the problem persists across multiple packages.  Is there a way to find the URL's in another location?

Moonshinelunacy:
Just double checked that permissions are all set to allow as they should. Additionally, I am logged in as the only account

s1mwest:
I have similar issue:  "Failed to run the package service" after install DNScrypt-proxy and each time I attempt to run it from package ceter.  When I open the app, I see the following in the app window "fork/exc /user/syno/synoman/webman/login.cgi: permission denied"
DSM 6.1 Armada375 logged in as admin/only account
dnscrypt-proxy_armada375-6.1_2.0.16-0.spk

BenjV:
Dat DNScrypt is een beta package en ik heb er even naar gekeken maar dit package kan helemaal nooit draaien op DSM.
Alle scripts om het te installeren en te kunnen starten en stoppen zijn leeg.

s1mwest:
Are they both beta?
I used this one:  https://github.com/publicarray/spksrc/releases but there's no mention of it being beta - although I am a GitHub newbie and not very clued up (can you tell?)

The one at https://synocommunity.com/packages (dnscrypt.v1.f5004[armada375].spk) shows as beta - whilst this installed, there is no interface to work with to configure it.

Really appreciate the efforts by those who create these - really hope something final available

ymartin59:
Publishment of dnscrypt-proxy is in progress: #3479

publicarray:
Hi, sorry I'm late to the party. Yea I have't been able to test the package on the DSM until today. Thanks to Xpenology I can now run DSM in Virtualbox. And yea please @ me or send issues to the relevant gitub repo so I get notified.

I think the problem is that the generic script for the DSM `chown`'s all files:
![](https://user-images.githubusercontent.com/5497998/47199172-feae9700-d3bb-11e8-9fba-0aa44606fe18.png)

I'll have to chown the cgi script somehow

publicarray:
[...]

PR #3501, opened 2018-10-25
Update dnscrypt-proxy to 2.0.19

publicarray:
_Motivation:_ There was an issue when wget could fail and write a 0 byte file. wget could fail when the device had dnscrypt-proxy configured as it's main DNS without fallback. During a package upgrade/reinstall wget could not resolve the domain name for `raw.githubusercontent.com`
Also remove an unnecessary chmod (it's a no op)

_Linked issues:_ Fixes https://github.com/SynoCommunity/spksrc/issues/3468#issuecomment-431942059

### Checklist


ymartin59:
Wouldn't be preferable to download file from Makefile and include it once and for all users in package itself...

publicarray:
I not fond of downloading any file from within make. Having the file in git gives us version control and create deterministic builds. I've includes the files in the package here: https://github.com/SynoCommunity/spksrc/tree/master/spk/dnscrypt-proxy/src/offline-cache

ymartin59:
@publicarray Does it make sense to copy that files from original upstream release (https://github.com/jedisct1/dnscrypt-proxy/blob/master/utils/generate-domains-blacklists/) ?
May you please fix "cross/dnscrypt-proxy" Makefile and digests to include package name in downloaded archive to avoid collision. You may set `PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)`

publicarray:
@ymartin59 yes it does sorry I missed that they are included in the git archive. I'm sill leaving the offline-cache folder for the server lists. Those 2 files (`public-resolvers.md` and `public-resolvers.md.minisig`) are automatically updated when `dnscrypt-proxy` considers the files out of date.

publicarray:
@ymartin59 Sorry for the mess above. Do you think this can finally be published? 🙏

ymartin59:
Hello. Yes I think so. But at the moment, I work on python3+borgbackup+homeassistant+armada37xx...

publicarray:
Thanks no worries

ymartin59:
@publicarray Publish is pending because of troubles with spkreo. Your package is really complex but you did a great improvement with go compilation preventing multiple downloads.

A proposal for next update: try to reuse https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.privilege-startasroot with additional Makefile variables - even without SERVICE_EXE for start-stop-daemon script: https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.mk#L156 maybe with SERVICE_USER = root as I guess there is no benefit to create non-privilege user for dnscrypt-proxy

publicarray:
> Your package is really complex
No objections from me. At the start I was not aware of the subtle differences between the DSM and RSM. Also I wanted the package to work with dnsmasq (DHCP server). Next add a GUI and help files... and here we are.

>  guess there is no benefit to create non-privilege user for dnscrypt-proxy

It is true that dnscrypt-proxy starts as root because it binds to a privileged port (unless port 53 is already in use). After the files are read and the sockets are opened the process (by the default that I have set) will drop privilege to the `${EFF_USER}` user.

publicarray:
@ymartin59 Isn't having the `SERVICE_EXE` variable the trigger to get the `spksrc.service.privilege-startasroot` permission file?

ymartin59:
Correct but it replaces start-stop-status script too, so that SERVICE_EXE is passed as argument to busybox start-stop-daemon

publicarray:
🎉  Thank you for publishing the package! 🎉

---

I gave it a go but the service can't start. It exits with something like `start-stop-daemon can not be found`

```diff
- CONF_DIR = src/conf
+ SERVICE_EXE = $${SYNOPKG_PKGDEST}/bin/dnscrypt-proxy
+ SERVICE_OPTIONS = --config $${SYNOPKG_PKGDEST}/var/dnscrypt-proxy.toml --pidfile $${PID_FILE}
```

Issue #3202, opened 2018-03-05
Mono/Sonarr Debug high cpu


jonskoglund:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ Mono
_Package Version:_ 5.8.0.108-11

_NAS Model:_ DS414
_NAS Architecture:_ Armada XP MV78230
_DSM version:_ DSM 6.1.5-15254 Update 1

--
### Expected behavior
Normal CPU load. 

### Actual behavior
High CPU load with spikes in the range of 90-115%
![cpu high](https://user-images.githubusercontent.com/593324/36966662-ca78604a-205d-11e8-8adb-86402a4a5337.png)

### Steps to reproduce
_1._ Install Mono and Sonarr SynoCommunity packages with DSM Package Center
_2._ Make sure both are running after installation.
_3._ Update Sonarr from the Web UI: System>Updates

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
-- Can't find any log for Mono. Only install.log in location /usr/local/mono/var/. Any tips?
```
Insert the package log here
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here
```

Safihre:
Yes I saw this too, but this only happens after Sonarr restarts itself after an update.
When we start the package it's not run with debug.

jonskoglund:
Restarting Sonarr doesn't fix the issue. I'm looking for a way to restart Mono form the terminal. DSM doesn't let me stop it from the DSM Package Center.

Safihre:
I am able to shut it down from Package  center, it just takes a while. You can also use the command line `synopkg` command.
Not sure where this Debug is coming from, it's not anywhere in our package code. Maybe it's in the `mono` somewhere? Do you know @m4tt075 / @ymartin59?

m4tt075:
Hmh, strange. I don't think this comes from `mono`, but I could reproduce what you wrote, @Safihre . On clean install `Sonarr` starts `mono` without `--debug` option, after updating it automatically restarts with `--debug` enabled. Stopping and restarting `Sonarr` again from the DSM package center, and the `--debug` option is gone again. @jonskoglund You don't really stop and start `mono`. It is started by the applications using it, `Sonarr` in this case. Maybe just try to stop and start it from the package center again?

jonskoglund:
[...]

PR #3518, opened 2018-11-16
Adding Sonarr v3 compatibility

maxrogers:
_Motivation:_ This PR adds compatibility to upgrade successfully to Sonarr v3 (when released.) Sonarr v3 changes the executable name (NzbDrone.exe --> Sonarr.exe) and data folder/filenames (NzbDrone --> Sonarr.) This package allows proper package upgrades when installing to a v2 or v3 install (or vice versa.)

### Checklist

### DIFF
--- a/cross/libmediainfo/Makefile
+++ b/cross/libmediainfo/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = libmediainfo
-PKG_VERS = 17.12
+PKG_VERS = 18.08.1
 PKG_EXT = tar.xz
 PKG_DIST_NAME = $(PKG_NAME)_$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = http://mediaarea.net/download/source/$(PKG_NAME)/$(PKG_VERS)
--- a/cross/libmediainfo/digests
+++ b/cross/libmediainfo/digests
@@ -1,3 +1,3 @@
-libmediainfo_17.12.tar.xz SHA1 928650e8ee3b131eb1a7ad9b8ec71985788b1f95
-libmediainfo_17.12.tar.xz SHA256 b482bb34df5d59ae8b735b9631ad0ad150dc71b715e851f1d3155ceb4921a452
-libmediainfo_17.12.tar.xz MD5 18556e9e6b14e537aeab6eaac7b9b3f3
+libmediainfo_18.08.1.tar.xz SHA1 13a6ea8d83e4fd4d6e239d5a6b3f71f4766c0599
+libmediainfo_18.08.1.tar.xz SHA256 7a237f35257fadf2796c02161e87330dedca74d0fb914259be3d8d88d2cbcbc0
+libmediainfo_18.08.1.tar.xz MD5 c5fba877164fa5654373b59fa68b9590
--- a/cross/mediainfo/Makefile
+++ b/cross/mediainfo/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = mediainfo
-PKG_VERS = 17.12
+PKG_VERS = 18.08.1
 PKG_EXT = tar.xz
 PKG_DIST_NAME = $(PKG_NAME)_$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = http://mediaarea.net/download/source/$(PKG_NAME)/$(PKG_VERS)
--- a/cross/mediainfo/digests
+++ b/cross/mediainfo/digests
@@ -1,3 +1,3 @@
-mediainfo_17.12.tar.xz SHA1 dfbdbf61ab8304579669de434557d0f28482639c
-mediainfo_17.12.tar.xz SHA256 ce14ae3b15fb36e9cd21c73ac1b4c58bcc7d9f52b1fd110a7d68b0e12d70addf
-mediainfo_17.12.tar.xz MD5 6d29665de2630b0d1b0f4e4bc242f0d2
+mediainfo_18.08.1.tar.xz SHA1 d2d663b8850208b22c9b0db3ab18fa08a010e7df
+mediainfo_18.08.1.tar.xz SHA256 49773d0b13cb13818a7727b59f0124186fd96b3ffd22a6402ccf082c3f6b0267
+mediainfo_18.08.1.tar.xz MD5 3b0ce00cbf110e6c496c14641df59737
--- a/cross/sonarr/Makefile
+++ b/cross/sonarr/Makefile
@@ -1,11 +1,11 @@
 PKG_NAME = NzbDrone
-PKG_VERS = 2.0.0.4689
+PKG_VERS = 2.0.0.5252
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME).master.$(PKG_VERS).mono.$(PKG_EXT)
 PKG_DIST_SITE = http://download.sonarr.tv/v2/master/mono
 PKG_DIR = NzbDrone
 
-DEPENDS = 
+DEPENDS =
 
 HOMEPAGE = https://sonarr.tv
 COMMENT  = Sonarr is a PVR for newsgroup users. It can monitor multiple RSS feeds for new episodes of your favourite shows and will grab, sorts and rename them. It can also be configured to automatically upgrade the quality of files already downloaded if a better quality format becomes available.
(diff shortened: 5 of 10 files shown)


ymartin59:
Looks interesting. May you please update version in spk/mediainfo to match with cross/mediainfo

maxrogers:
@ymartin59 good catch! I added spk/mediainfo. Thanks!

ymartin59:
@maxrogers Thanks for improvement. Please rebase. I am interested in test reports about running v3 with current Mono version and future one from #3561

M0UL:
maybe add the mono debug switch to the sonarr startup script. this recommended by sonarr
image

maxrogers:
> maybe add the mono debug switch to the sonarr startup script. this recommended by sonarr

The debug switch used to be in the sonarr package: https://github.com/SynoCommunity/spksrc/commit/7db4ec087962094e8285aa03b7fadd325adaca27#diff-d94600ee9255b9e01269842f7d537893 However, it was removed by someone (not sure when.) I believe it was removed after complaints regarding high CPU usage when using the debug flag. example: https://github.com/SynoCommunity/spksrc/issues/3202 

I think it is probably wise to not add the debug flag again unless this potential issue has been resolved (considering many are using nas devices with limited resources.) According to the link I referenced above, if you restart sonarr using the sonarr web interface (not using package manager) it will use the debug flag (I have not confirmed.) Perhaps that or a manual start from command line could be used for debug purposes...

maxrogers:
> @maxrogers Thanks for improvement. Please rebase. I am interested in test reports about running v3 with current Mono version and future one from #3561

Rebased and squashed. However, if you want to run v3 now you will want to use this for your testing: https://github.com/maxrogers/spksrc/tree/sonarr_v3

You cannot successfully upgrade to the correct/up-to-date `phantom` (v3) branch using the sonarr upgrade interface in the current v2 package (or using my v3 upgrade compatible PR.) My PR is still loading a v2 binary (2.0.0.5252.) If try to upgrade using the sonarr upgrade interface in current v2 binaries you will get a really old and strangely numbered version of phantom (v3.) The following changes to my PR will get you the correct v3 which will also allow you proper upgrades to the latest v3/phantom betas: https://github.com/maxrogers/spksrc/compare/sonarr...maxrogers:sonarr_v3

Obviously we would not want to add the phantom branch binary to our package. We should wait until it becomes one of the major sonarr branches (develop.) When that happens, my current PR should allow everyone to successfully upgrade at that time (via the sonarr interface on the develop branch.)

M0UL:
the sonarr v3 branch changed from phantom to phantom-develop
http://download.sonarr.tv/v3/phantom-develop/

maxrogers:
@

> the sonarr v3 branch changed from phantom to phantom-develop
> http://download.sonarr.tv/v3/phantom-develop/

@M0UL, thanks for the heads up! I just switched my branch to phantom-develop. However, this PR just gives v3 upgrade compatibility (it actually still installs with a v2 binary.) I should switch the binary in this v3 test package though: https://github.com/maxrogers/spksrc/tree/sonarr_v3 

Thanks again!

m4tt075:
@maxrogers *Argh* I've overlooked this when sending my PR yesterday. Have you tested your `sonarr` builds with the updated `mono` I've been working on? Does it (still) work?

maxrogers:
@m4tt075 Awesome work! I just tested on a 6.1 system with the updated mono build you posted here: https://github.com/m4tt075/spksrc/releases/tag/mono-5.18-11

It worked great! Thankss

kc6108:
I know we are trying to get away from this, but I'm beginning to think adding a new (yet temporary in the long run) package (ie Sonarr Phantom Testing) so we can begin testing with the v3 binaries is in order. Some of us have been waiting for several months to begin our testing, and the release date for Sonarr v3 keeps getting pushed back. I've been helping test on a Windows machine, but Synology is where I really want to focus my efforts.

Thoughts?

Many Thanks,
-kc6108

maxrogers:
@kc6108, if you are comfortable compiling packages in spksrc you can create a v3 Sonarr package (to test out) by modifying three files (as shown below) then compiling your test package:

`cross/sonarr/Makefile`

> PKG_NAME = Sonarr
> PKG_VERS = 3.0.1.497
> PKG_EXT = tar.gz
> PKG_DIST_NAME = $(PKG_NAME).phantom-develop.$(PKG_VERS).linux.$(PKG_EXT)
> PKG_DIST_SITE = http://download.sonarr.tv/v3/phantom-develop/$(PKG_VERS)
> PKG_DIR = Sonarr

`cross/sonarr/PLIST`

> rsc:share/Sonarr

delete (or create proper digest:) --> `cross/sonarr/digests`

Also, I am closing this PR since it was merged here: https://github.com/SynoCommunity/spksrc/pull/3657

Taloth:
@maxrogers What should I put here for instructions? https://sonarr.tv/#downloads-v3-nas-synology
Also, there's also the option now to list the package author/provider in System->About for v3, we should probably talk about that separately.

maxrogers:
@Taloth, thanks for stopping by! The current synology sonarr package Is compatible with v3 but installs with v2. 

Upgrading to v3 via the built-in sonarr interface in v2 is not currently allowed, correct? For example: Adding `phantom-develop` to the branch in a v2 install. If that was allowed, that would be the easiest method for synology users to upgrade to v3 (if it worked.) I assume it is not currently allowed for compatibility reasons with other systems?

The next best option for synology users who want to test the v3 beta would probably be for me to compile a list of beta v3 synology packages and post them here on github until a built-in v2 -> v3 upgrade option is offered.

Taloth:
_We don't plan to offer a v2->v3 update from the UI. There are some breaking changes so we don't want to auto update and also want to avoid ppl doing a blind 'hey update available, _click install_'.
But also because v3 install is a bit more tightly managed, for debian we moved to our own systemd startup script, for example. That doesn't apply here of course. But we don't want too much magic._

----
### Directory Structure and package_info

The new v3 directory structure is slightly different than v2 was:
Binaries were moved to `/usr/lib/sonarr/bin`.
An optional `/usr/lib/sonarr/package_info` file was added (NOT in `sonarr/bin/`)
The package owns that file, the bin dir is owned by sonarr. _The builtin updater only updates the bin folder. Although at one point i'd want to use `sonarr/bin_tmp` and  similar for safer and quicker updates._

For the docker image by linuxserver.io the package_info file contains:
```
UpdateMethod=docker
Branch=phantom-develop
PackageVersion=3.0.1.364-ls5
PackageAuthor=linuxserver.io
```
The PackageVersion + PackageAuthor shows up in System->About, and best contains a combo of sonarr's version and the package version.
The UpdateMethod can be the default 'builtin'. Or alternatively `apt`, `docker` or `external`, which forcibly disables the built-in updater and enforces the specified branch. The only diff is a specific update message in System->Updates.
The user is always able to switch to `Script` but setting that option in package_info is pointless since there is no method to provide the script path at this time.

I'd like you to add a similar package_info to the syno package (but then with UpdateMethod=BuiltIn) and of course the appropriate Version and Author. 
Adjust the binary directory as well or it won't work, Sonarr only looks for ../package_info if it's Sonarr itself resides in bin (so `${SYNOPKG_PKGDEST}/share/Sonarr/bin`).
You can safely add package_info to v2, it's be ignored atm. 

----
### v3 Beta

I can think of a few options:

1) If you create a beta package and have instructions for beta users to install, then we'd be able to point to that until a real v3 package goes live. The package wouldn't need to be updated constantly since the built-in updater should function.

2) Include a script that experienced users can run from the cmdline, which downloads and install v3.
You already know the services url https://services.sonarr.tv/v1/download/phantom-develop/latest?version=3&os=linux which redirects to the appropriate download url.

3) The presence of the package_info could be used by Sonarr to detect whether the package is v3 compatible, but you aren't truly compatible since you're moving from share/NzbDrone to share/Sonarr.
If you were to install v2 in share/Sonarr/bin instead, include share/Sonarr/package_info then we might consider detecting that.
I'm still not sure if we'd want to offer a UI option for it, but it at least becomes a possibility.

Each option has pros and cons. I think Option 2 is the least work.

We do not provide a v3 release date (and we never have so it was never 'pushed back' either). However we did transition from alpha to beta a couple of months ago.
It's still slightly below v2 develop in terms of stability. Especially since we still do breaking changes.

Safihre:
@maxrogers I think it's time we provide a package voor V3. Even if we only provide it here, for users to install manually. I also did that when we were in the DSM6 transition and it worked well.
Could you maybe create a V3 package and open a PR? Then we can build all the releases. I have a TravisCI pipeline for that, so they can be hosted by GitHub (for example [like this](https://github.com/Safihre/spksrc/releases/tag/sabnzbd-239)).

rp4k:
is it possible or am i missing it somewhere but why is the synocommunity repo still showing sonnar v2

Issue #3523, opened 2018-11-21
Error extracting toolchain for armada38x


obourdon:
### Setup
_Package Name:_ ALL
_Package Version:_ ANY

_NAS Model:_ DS218j
_NAS Architecture:_ armada38x
_DSM version:_ 6.1/6.2

### Expected behavior
tmux/git/... should build using make arch-armada38x

### Actual behavior
Error due to bad extraction of toolchain

### Steps to reproduce
_1._ use the official docker dev container 
_2._ follow instructions to setup and build any package
_3._ try building git for instance

#### Other logs
_E.g. `command output`_
```
===>    Extracting for syno-armada38x
tar -xpf /spksrc/toolchains/syno-armada38x-6.1/../../distrib/toolchains/6.1/armada38x-gcc493_glibc220_hard-GPL.txz -C /spksrc/toolchains/syno-armada38x-6.1/work 
tar: arm-unknown-linux-gnueabi/share: Cannot mkdir: Permission denied
tar: arm-unknown-linux-gnueabi/share: Cannot mkdir: Permission denied
tar: arm-unknown-linux-gnueabi/share/locale: Cannot mkdir: No such file or directory
...
# find /spksrc/toolchains/syno-armada38x-6.1/work -ls
42733955    0 drwxr-xr-x   5 root     root          170 Nov 21 23:29 /spksrc/toolchains/syno-armada38x-6.1/work
42733960    0 -rw-r--r--   1 root     root            0 Nov 21 23:29 /spksrc/toolchains/syno-armada38x-6.1/work/.syno-armada38x-checksum_done
42733956    0 -rw-r--r--   1 root     root            0 Nov 21 23:28 /spksrc/toolchains/syno-armada38x-6.1/work/.syno-armada38x-download_done
42733961    0 dr-xr-xr-x   2 1000     1000           68 Jun 24  2016 /spksrc/toolchains/syno-armada38x-6.1/work/arm-unknown-linux-gnueabi
```

BenjV:
You should fix the permissions yourself.
The user which you are using does not have enough permissions to create folders.

obourdon:
@BenjV nope, you can try this by yourself in the docker container, there is no way root user can extract properly when dirs are not writable. The only other option is to fix the toolchain archive but as this kind of issue can occur for any kind of archive the fix proposed in [Pull request #3524 ](https://github.com/SynoCommunity/spksrc/pull/3524) is generic enough to cope with most situations I have tested.

[...]

PR #3524, opened 2018-11-21
Fix potential permission errors

obourdon:
When the toolchain is extracted, current ownership and
permissions can lead to errors (mostly due to non-writable dir).
Only extraction into /tmp can be successful.

_Motivation:_ Fix for [issue #3523](https://github.com/SynoCommunity/spksrc/issues/3523)
_Linked issues:_ See [https://github.com/SynoCommunity/spksrc/issues/3523](https://github.com/SynoCommunity/spksrc/issues/3523)

### DIFF
--- a/mk/spksrc.extract.mk
+++ b/mk/spksrc.extract.mk
@@ -8,14 +8,14 @@
 #  post_extract_target  (override with POST_EXTRACT_TARGET)
 EXTRACT_PATH ?= $(WORK_DIR)
 # Extract commands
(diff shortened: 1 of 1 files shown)


ymartin59:
Container image is designed so that disk space is consumed at /spksrc/ which is expected to be mapped to a user's host folder.
If large volumes are written to /tmp/ in the container, this will increase layer storage at /var/lib/docker and it may difficult to free space - with aufs, the only option is to fully delete /var/lib/docker/ - with overlay2 "docker system prune" is there to do so...
As a conclusion, I prefer to find an option that will not allocate storage with garbage.

PR #3520, opened 2018-11-17
beets: refactor to use generic installer

ymartin59:

PR #3527, opened 2018-11-23
Add tcpdump package

obourdon:
_Motivation:_ Adding packaging for tcpdump and upgrading libpcap version to latest
### Checklist


publicarray:
In case you don't know, tcpdump can be installed on a NAS with synogear -> https://github.com/SynoCommunity/spksrc/wiki/FAQ-synogear

ymartin59:
Because of generic installer provided by spksrc, there is no longer need for SSS_SCRIPT and INSTALLER_SCRIPT. See #3520 for example.

ymartin59:
Because available with synogear package, I prefer not to include it here - except if there are good reasons to do so.

PR #3528, opened 2018-11-25
framework: fix DSM_UI_DIR defaults to app

ymartin59:
_Motivation:_ even with generic service usage, DSM_UI_DIR = app is still required for menu item and icons.Default value is now set at right place.

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -35,6 +35,10 @@ else
 $(PRE_SERVICE_TARGET): service_msg_target
 endif
 
+ifeq ($(strip $(DSM_UI_DIR)),)
+DSM_UI_DIR=app
+endif
+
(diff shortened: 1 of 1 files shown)


m4tt075:
@ymartin59 Does this fix the problem with missing menu items, which have been reported randomly recently (e.g. for Gateone)?

ymartin59:
Almost sure if DSM_UI_DIR variable is not set explicitly in spk Makefile

m4tt075:
Hmh, `DSM_UI_DIR = app` is explicitly set in the `Gateone` spk Makefile. This is really strange. I'm not able to reproduce the missing menu items.

ymartin59:
You should unpack .spk (tar file) and package.tgz to confirm app/config file is present and icons too.
(I suppose you have checked permission granting to use application for the non-privileged user)

PR #3526, opened 2018-11-23
Add nmap package

obourdon:
_Motivation:_ Adding packaging for already existing cross compilation chain for nmap

### Checklist
- [x] New installation of package completed successfully

Issue #3489, opened 2018-10-14
[Package Request] sshfs for DS418j (arch rtd1296)


DrSpaldo:
Name: sshfs
Description: filesystem client based on the SSH File Transfer Protocol

These packages currently exist on the SynoCommunity site:

https://synocommunity.com/package/sshfs

However, I am not sure if another one needs to be built for the rtd1296 Architecture or if I can use one of the current ones. 

If anyone can help it would be much appreciated.

DrSpaldo:
Hi ymartin59. Thanks for the post. Excuse my ignorance but is your spk one that has a bundle of commands/programs in it? And you are going to add sshfs as well?

 If so, great! 

Is it available already?

ymartin59:
Yes that is exactly what is planned. I try to get it available soon.

DrSpaldo:
Great, looking forward to it. Thank you

ymartin59:
You will find sshfs for your arch in packag https://synocommunity.com/package/synocli-net
Not the latest version but should work.

DrSpaldo:
Great, thank you 👍

PR #3267, opened 2018-04-02
Add nmap spk

publicarray:
_Motivation: Add nmap for running network scans from the router/NAS (good for troubleshooting things)

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

Issue #3254, opened 2018-03-25
Design: new package "SynoCli net"


ymartin59:
Create a package for small command line utilities dedicated to network/remote access.
Revival of #2261
Tools candidates:
- screen (from testdisk)
- tmux (replace standalone package)
- mosh #2936 #2483
- socat #3249 
- nmap 
- sshfs
- ~~aria2~~ dedicated package planned #2959 

Any other ideas?
Wanted contribution: an generic icon for "SynoCli" package group, with "net" text addition for category

publicarray:
How about [iftop](http://www.ex-parrot.com/pdw/iftop/), [mtr](https://github.com/traviscross/mtr), [nethogs](https://github.com/raboof/nethogs) ? or should those be a separate group (including nmap)?

I've used created an Icon but I've used Font Awesome so inclusion of their licence is needed: https://fontawesome.com/license

[icons.zip](https://github.com/SynoCommunity/spksrc/files/1892667/icons.zip)

ymartin59:
@publicarray Thanks for these icons... I still wonder if `spksrc` is really the right place to provide "simple" command line tools than may be already available with entware-ng #3255.
There is already a lot of volume to maintain here, so I would be pleased to "delegate" to keep our list of applications "relevant" and as short as possible.

publicarray:
I see your point. Hm the packages that I suggested are more aimed for router users/network people anyway. I don't have an answer either. However I do think that we could combine all current CLI packages together into one meta CLI package (or maybe just two, one that is more lightweight version). The current proposal with CLI categories is also a good idea. I don't know..

Another option for NAS users (I don't think the hardware in the Router is quite enough) is to install [Debian-Chroot](https://github.com/SynoCommunity/spksrc/wiki/Debian-Chroot)

publicarray:
So here is how I see it. People can install Debian-Chroot and have access to "everything". Or they can install Entware (once ready) to get the vast majority of things a dev would need. Or they can install SynoCommunity's own "limited" selection of existing CLI packages (whether a meta, category or individual package, I wouldn't worry too much about duplicates. Worrying about duplicates can course more of a maintenance burden than it tries to solve).

ymartin59:
I partly agree. I would care about "duplicates" only when a "new package request" triage... to avoid volume to grow.

ymartin59:
@publicarray I am back here and I wonder if Font Awesome may be used according Free license: https://fontawesome.com/license/free

[...]

PR #3544, opened 2018-12-09
Package synocli-net

ymartin59:
_Linked issues:_ #3254 #3267 #3526 #3489 

### Checklist
- [N/A] New installation of package completed successfully


m4tt075:
@ymartin59 Very nice implementation! Great stuff, including @publicarray 's icon!  👍 
Only part I don't understand is: Why do you keep the "spk/tmux/*" files, allowing separate installation of the `tmux` package AND include it in the `synocli-net` bundle? I would have guessed it to be one "xor" the other. Is there a particular reason for this?

ymartin59:
@m4tt075 You're right - I have decided to remove "spk/mosh" from proposed PR but kept "spk/tmux" - in case anyone may want to build its own and so get a lighter package that synocli-net which is already 12 MiB heavy.
Another idea I have is to replace "spk/tmux" soon to publish a wizard only package and dependency to synocli-net to notify users about replacement.

nbarrientos:
Sorry for the noise, but have you by any chance removed `tmux` standalone in view of adding this bundle? Can't find it now in the list of available packages to install on my DS418 (rtd1296) running 6.2 latest. Thanks :)

nbarrientos:
Sorry, didn't have beta packages enabled.

Issue #3546, opened 2018-12-13
Problem - Compile Python rtd1296-6.1 ̶a̶a̶r̶c̶h̶6̶4̶-̶6̶.̶1̶


KosakAlmighty:
Hi guys,
I tried to compile Python for new architectures, using aarch64 toolchain.

But after couple minutes it ends on this: 

```
===>  Processing dependencies of cffi
make[3]: Entering directory '/spksrc/spksrc/cross/pycparser'
make[3]: Leaving directory '/spksrc/spksrc/cross/pycparser'
make[3]: Entering directory '/spksrc/spksrc/cross/libffi'
===>  Configuring for libffi

<skiped part, some configuration is going here>

checking build system type... i686-pc-linux-gnu
checking host system type... aarch64-unknown-linux-gnueabi
checking target system type... aarch64-unknown-linux-gnueabi
continue configure in default builddir "./aarch64-unknown-linux-gnueabi"
....exec /bin/bash .././configure "--srcdir=.." "--enable-builddir=aarch64-unknown-linux-gnueabi" "linux
gnueabi"
.././configure: line 2356: config.log: Invalid argument
.././configure: line 2366: config.log: Invalid argument
../../mk/spksrc.configure.mk:50: recipe for target 'configure_target' failed
make[3]: *** [configure_target] Error 1
make[3]: Leaving directory '/spksrc/spksrc/cross/libffi'
../../mk/spksrc.depend.mk:44: recipe for target 'depend_target' failed
make[2]: *** [depend_target] Error 2
make[2]: Leaving directory '/spksrc/spksrc/cross/cffi'
../../mk/spksrc.depend.mk:44: recipe for target 'depend_target' failed
make[1]: *** [depend_target] Error 2
make[1]: Leaving directory '/spksrc/spksrc/spk/python'
../../mk/spksrc.spk.mk:452: recipe for target 'arch-aarch64-6.1' failed
make: [arch-aarch64-6.1] Error 2 (ignored)
```

So the problem is seems to be with libffi, but even after couple hours of digging I wasn't able to find any error. 
```
.././configure: line 2356: config.log: Invalid argument
.././configure: line 2366: config.log: Invalid argument
```

I took look in configure file for libffi and at these lines is this: 
```
2356 cat >config.log <<_ACEOF
2357 This file contains any messages produced by compilers while
2358 running configure, to aid debugging if configure makes a mistake.
2359
2360 It was created by libffi $as_me 3.2.1, which was
[...]

PR #3552, opened 2018-12-20
python: 2.7.15 update with dependencies

ymartin59:
_Motivation:_ Publish for aarch64 and vulnerability fixes
_Linked issues:_ #3546

### Checklist


Safihre:
Cool. Was just about to push a new SABnzbd update so this would be a nice fit!

Safihre:
So what is going wrong here? I tried to build it on Travis but getting some download-error.

ymartin59:
Hello. Sorry I forgot to comment status: build fails on "cryptography" wheel, probably because of versions mismatch in requirements.

Safihre:
Pushed libsodium update, since listed version [was no longer available](https://download.libsodium.org/libsodium/releases/). Continue testing on now, got a new VM set up.

ymartin59:
Thanks. Sorry but I am busy on FFmpeg v4 at the moment.

Safihre:
Seems we are having the exact same problem @Diaoul had 5 years ago: https://github.com/pyca/cryptography/issues/1325
But I don't really understand the solution.. Should we go back to having `cryptography` as a `cross` maybe? Or just keep it at the older version, if that works?

Safihre:
@ymartin59 I can confirm that lowering the `cryptography` version (and the rest in that section of `requirements.txt`) back to what they were for 2.7.14 fixes the compile.
I would suggest to just go back to those versions.

ymartin59:
Hum. Something is strange - I have just clean "native/python" and "spk/python" and package has just been built properly `arch-x64-6.1`. Maybe the cryptography error comes from having left previous version compiled in "native/python"...

Safihre:
It's the 32bit archs that fail to build.

ymartin59:
I wonder if option would be to restore cross/cryptography removed recently https://github.com/SynoCommunity/spksrc/commit/6a071f8c24e2d067041fa4c362387668dfae9d46

Safihre:
With the old requirements file I'm able to build all archs it seems. But can't test them all because for some reason it can't get all the sources when I run it on Travis. Maybe some source is blocking Travis..

Safihre:
It seems Travis switched to some special NAT, resulting in the rotating of the external IP's of the build VM **during** the building: https://twitter.com/travisci/status/1085214062152159232

That's probably why the build process is failing with these download-errors. (CC @m4tt075)

ymartin59:
@Safihre may you please push your additional commit on my PR here - you have permissions granted I guess - I recommend "hub" to checkout PR: https://hub.github.com/ then it cannot be easier.

ymartin59:
@Safihre in fact it is also easy to cherry-pick commit from your branch...

ymartin59:
I have been told it may be useless to "rebuild" python in each package only to build wheels for application requirements (what spksrc does right now) and that "crossenv" is a tool designed for such usage. Have you heard about it?
As far as I understand, when this python update will be ready, there is a high chance that all applications have to be rebuilt and pubilshed again, right?

ymartin59:
OK, first steps. Compilation of "cryptography" wheel fails for ARMv8 (aarch64), I am investigating.

About "crossenv", proposal concerns "home assistant" - and probably "borgbackup" and "beets" can take benefits for it (when moved to python3)

Safihre:
@ymartin59 any luck with `aarch64`?

ymartin59:
@Safihre Sorry I am busy on framework refactoring #3595 to get ffmpeg and company published.

ymartin59:
@chickenandpork May I request your help to achieve this PR? I do not understand why recent cryptography wheel no longer builds on ARM platforms and I am stuck here. Thank you in advance

Safihre:
It does build on some ARM platforms, just not all..

Safihre:
@chickenandpork are you looking into this or should I give it a try over the weekend?

chickenandpork:
I’m not sure the problem, but I’ll have time I think around 06:00UTC (11 hrs from now).  If I’m going to get time to look, it tends to be 0600Z daily for a few hours

chickenandpork:
Manually did a similar patch, trying builds now to see what it looks like.  I don't think I'll be able to debug the failure until tomorrow.

I'm worried about a few related changes we'll need: for example, because the `requirements.txt` has exact version matches (ie `virtualenv==16.1.0` not `virtualenv>=16`) that anything requiring `python` will need to have compatible `requirements.txt` or else `pip` will complain.

Eventually, might need to change the bare minimum of wheels versions, merge/ship that, and try after to bump versions across the board (ie bump all virtualenvs at once).  Dunno.

I'll still need to build after as well with my detection for all affected SPKs.  Dependencies sneak in, and there will probably be a number of SPKs that depend on `python`

Safihre:
Since we make a virtualenv for each Python application within the SPK, the different versions listed in requirements don't need to clash.
The reason for the set version is because we need to list both the package and the dependencies, since we go around the regular pip system. And each new build of the python SPK should produce reproducible SPK's

chickenandpork:
Thanks.  I’ve worked in embedded so I’m aware of why to use exact versions; my concern was mismatched exact versions, but I’d forgotten about virtualenv isolation.  Oops . :(  I kinda do that with containerd now :-/  (pre-k8s world)

chickenandpork:
I've got builds running slowly since last night, so far it's built:
```
python_x64-6.1_2.7.15-20.spk
python_avoton-6.1_2.7.15-20.spk
python_88f6281-6.1_2.7.15-20.spk
python_aarch64-6.1_2.7.15-20.spk
python_alpine-6.1_2.7.15-20.spk
python_armada370-6.1_2.7.15-20.spk
python_armada375-6.1_2.7.15-20.spk
python_armada38x-6.1_2.7.15-20.spk
```

Where did you get the failure?  This is with a hand-done approximation of @ymartin59 's hard work, but it's not yet failing.

Safihre:
Continuing in #3628

PR #3556, opened 2018-12-24
Add syno-manager v1.0 package

neykov:
Download Stations magnet links handler

_Motivation:_ Add a package that will let you use Download Station as a magnet:// link handler - add magnet links to Download Station by directly clicking on the link.

### Checklist
* [x] Build rule `all-supported` completed successfully
* [x] Package upgrade completed successfully
* [x] New installation of package completed successfully


neykov:
I actually started this way, but gave up after I hit a wall with `checksum` being a target for both checking the `digests` file and generating the checksum for the spk package. Any idea how to work around it?

neykov:
I think that's much better now - I renamed `checsum` to `info-checsum` in `spksrc.spk.mk` as you suggested.

PR #3557, opened 2018-12-26
Update go-1.4 to latest version in order to build on Debian Stretch

alexanderharm:
_Motivation:_ Go compiler will not build on Debian Stretch with version 1.4.2
_Linked issues:_ none

### Checklist

According to the official build from source instructions on <https://golang.org/doc/install/source>, version `go1.4-bootstrap-20171003` is needed for `Go compiler` to compile on current systems. Also `CGO_ENABLED=0` should be set. Since the build environment is based on `Debian`, this update is needed for `Go` to compile successfully on `Stretch`.

### DIFF
--- a/native/go-1.4/Makefile
+++ b/native/go-1.4/Makefile
@@ -1,8 +1,8 @@
 PKG_NAME = go
-PKG_VERS = 1.4.2
+PKG_VERS = 1.4-bootstrap-20171003
 PKG_EXT = tar.gz
-PKG_DIST_NAME = $(PKG_NAME)$(PKG_VERS).src.$(PKG_EXT)
-PKG_DIST_SITE = https://storage.googleapis.com/golang
+PKG_DIST_NAME = $(PKG_NAME)$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://dl.google.com/go
 PKG_DIR = $(PKG_NAME)
 
 DEPENDS =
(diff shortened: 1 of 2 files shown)


ymartin59:
I also have to update Docker image to Debian Stretch to get Meson build tool for sshfs in #3544

hgy59:
Sorry, did not find this PR before.

- Just made #3649 with similar fix
- My PR uses version 1.4.3 instead of go1.4-bootstrap-20171003 (who knows the difference?)

alexanderharm:
go1.4-bootstrap-20171003 is the latest version provided by Google. AFAIK it contains some fixes compared to 1.4.3. It is the version they recommend in their official documentation.

cytec:
@alexanderharm wouldn't `CGO_ENABLED=0` disable cgo support? afaik we need cgo for at least one package... might be wrong there but since we have a valid C compiler available i don't see any advantages in disableing CGO...

hgy59:
@cytec no, cgo support is not disabled for the target packages. The native/go-1.4 is the bootstrap version of go, i.e. go 1.4 ist the last go version that is build from c/cpp sources. Newer golang/go versions are built from go sources - and this is the bootstrap - we need go to build go. So the build chain builds go (currenlty version 1.11.1) from go 1.4. So the packages with go source code are compiled with go 1.11.1 that is full featured.

There are currently only two packages with go source code (syncthing and dnscrypt-proxy) and these build well with this solution.

ymartin59:
Closed because included in #3649

PR #3566, opened 2019-01-05
syno-magnet: Update package location because of renamed package name

neykov:
The package name has been changed so we need to update the web app url. Override the 'config' coming from the app with the proper location.

_Motivation:_ The package name changed in  https://github.com/SynoCommunity/spksrc/commit/4eacfb652c96339a992af705d43677b9bf7bc9a7 which caused the url to change. The app assumes a different URL so the PR overrides the app assumption with the correct location.

### Checklist

### DIFF
--- a/spk/syno-magnet/Makefile
+++ b/spk/syno-magnet/Makefile
@@ -8,7 +8,7 @@ PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 SPK_NAME = $(PKG_NAME)
 SPK_VERS = $(PKG_VERS)
-SPK_REV = 1
+SPK_REV = 2
 SPK_ICON = src/magnet.png
 
 MAINTAINER = neykov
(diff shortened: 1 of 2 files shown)


neykov:
cc @ymartin59 - see changes here for further updates needed for https://github.com/SynoCommunity/spksrc/commit/4eacfb652c96339a992af705d43677b9bf7bc9a7.

ymartin59:
Sorry for that side-effect... I have not taken care enough. I changed name for homogeneity with all other packages, package name is expected to be consistent with "spk/" folder name and only minor cases and dashes are used.

spk/syno-magnet/Makefile
```diff
@@ -44,10 +47,24 @@ include ../../mk/spksrc.extract.mk
 
 .PHONY: syno_magnet_copy
 
-# Icons already included in download
-$(ICON_DIR): ;
+# Override because we need type: legacy
```
ymartin59: Please add SERVICE_TYPE support with default value "url" in spk.service.mk directly at https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.mk#L214

ymartin59: I changed my mind about variable name, SERVICE_URL_TYPE seems more appropriate to me.

ymartin59: Correct.

PR #3567, opened 2019-01-06
One shot spk arch convenience func

chickenandpork:
_Motivation:_ 

In looking at build instructions, it makes me think that there must be a more trivial set of instructions.

Instead of:

```
# sudo docker run -it -v $(pwd):/spksrc -w /spksrc synocommunity/spksrc /bin/bash
$ make setup
$ cd spk/python3
$ make arch-XXXX
$ exit
```

The resulting command is:

`sudo docker run -it --rm -v $(pwd):/spksrc -w /spksrc  synocommunity/spksrc  make  python3-avoton-6.1`

_Linked issues:_

Can make home-assistant/home-assistant.io#7861 more trivial

### Checklist  (none apply, so checked all that were required)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,14 @@ spk-clean:
 %-clean: spk/%/Makefile
 	cd $(dir $^) && env $(MAKE) clean
 
+# define a template that instantiates a 'python3-avoton-6.1' -style target for
+# every ($2) arch, every ($1) spk
+define SPK_ARCH_template =
+$(1)-$(2): spk/$(1)/Makefile setup
+	cd spk/$(1) && env $(MAKE) arch-$(2)
+endef
+$(foreach arch,$(AVAILABLE_ARCHS),$(foreach spk,$(SUPPORTED_SPKS),$(eval $(call SPK_ARCH_template,$(spk),$(arch)))))
+
 prepare: downloads
 	@for tc in $(dir $(wildcard toolchains/*/Makefile)) ; \
 	do \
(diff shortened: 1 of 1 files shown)


ymartin59:
@chickenandpork Good proposal. I would recommend to prefix target with "spk-" to avoid any collision with existing targets, like "spk-PACKAGE-ARCH". Please review your PR branch, commit list is unexpected.

chickenandpork:
Prefix: makes sense :)

Commit-list: hmm.  Surprising indeed.  I’ll rebase from upstream and push, —force if I need to.  Probably tonight.

Allan

CharlieDowns:
In the case it might be useful to you guys, I combined the two commands that used to run:
sudo docker run -it --rm -v ~/spksrc:/spksrc -w /spksrc synocommunity/spksrc bash -c 'make setup && make -C spk/python3 arch-XXXX'

chickenandpork:
Force-pushed a re-do based on spksc/master; let me address the `spk-` comment in a second commit to be clear.

CharlieDowns:
Nvm, looking at the commit, the command in my guide is safe, but that one-shot is quite nice, as I can remove  `make setup` and `make -C spk/python3 arch-XXXX'` and simply write `make spk-python3-XXXX`.   
So when you release it, in my guide I change it to:
```bash
sudo docker run -it -v ~/spksrc:/spksrc -w /spksrc synocommunity/spksrc bash -c 'make clean && spk-python3 arch-XXXX'
```
Do not mind the `make clean`, I decided to add it for safety reasons.

PR #3568, opened 2019-01-08
[ffmpeg] Remove libfdk-acc and openssl dependencies and fix licensing information (fixes #3564)

### LINKED ISSUE #3564: [ffmpeg] package is configured as non-redistributable
`--enable-nonfree` is required for some components, including libfdk_aac, that are not considered to be compatible with the (L)GPL .

It appears that your ffmpeg package(s) are being distributed with `--enable-nonfree` and `--enable-gpl` configure options and are therefore not in compliance with the license.

To fix this do not include `--enable-nonfree` (and therefore `--enable-libfdk-aac`) when compiling the distributed packages. This will remove support for the libfdk-aac AAC encoder, but users can use the built-in AAC encoder instead (`-c:a aac`).

These compile options can remain for users who use spksrc to compile the package for themselves as long as they do not distribute it.

Secondly, `Makefile` shows `LICENSE  = LGPL 2.1`, but `--enable-gpl` and `-enable-version3` are being used. It should be changed to `LICENSE  = GPLv3`.

m4tt075:
### Motivation
the `ffmpeg` package has been distributed with `--enable-nonfree` and `--enable-gpl` configure options and is therefore not in compliance with the license. Linked issue: #3564 

### PR Description
- The licensing information for `ffmpeg` is updated to GPLv3
- The incompatible `--enable-nonfree` configuration option is removed. Consequently, `libfdk_aac` and `openssl` can no longer be combined with `ffmpeg`. 
- Initially, the `libfdk_aac` cross package is kept to allow spksrc users to include it manually, but for personal use only (not for distribution)
- Another commit removes the `libfdk_aac` cross package completely, assuming that we want to drop maintainance of code that we cannot publish 
- `libfdk_aac` does not need a replacement, as an appropriate `aac` decoder is already built into `ffmpeg`
- The `openssl`  dependency is replaced by `gnutls` instead. All remaining commits address compilation issues with `gnutls` and upgrade `gnutls` and its dependencies.

### Notes
- The commit removing `libfdk_aac` completely can be dropped, if required or wanted
- `gnutls` is updated to version 3.6.4, as version 3.6.5 (the most recent release) depends on `nettle 3.4.1` which contains a regression and fails with `x64` platforms 
- `gnutls` itself and some cross-packages failed cross-compilation for various platforms. The introduction of appropriate toolchain-triplets together with explicitly specifying installation prefix, libdir and includedir information fixes those issues. 

### Checklist
- [x] Build rule `all-supported` completed successfully (with DSM5.2, 6.1 and 6.2 toolchains)
- [x] Package upgrade completed successfully (on XPE6.1 test system)
- [x] New installation of package completed successfully (on XPE5.2 system)


ymartin59:
Well done. For testing https, just found an hit at https://trac.ffmpeg.org/ticket/5915
I know Linux From Scratch since ages - but I just find out that its instruction may be a great help: http://www.linuxfromscratch.org/blfs/view/8.3/multimedia/ffmpeg.html

cytec:
so we remove the fdk-aac package completely now? even for people who want to compile it themself? besides that: LGTM! nice!

m4tt075:
@cytec: No strong feelings from my side, which is why I separated out the corresponding commit. @ymartin59 : Any explicit preference on this from your side? Otherwise, fine with your proposal. I'll start reviewing the `ffmpeg-4` project related issues and look into the update.

ymartin59:
I propose to add a "FFMPEG_NONFREE" switch so that it may be set at compile time as make "-D" argument. If set, binary is built with "non-distributable" compile flags as before.

ymartin59:
I wonder which ffmpeg version we should target 4.0.3 or 4.1 (2018-11-06): http://ffmpeg.org/download.html#releases
Maybe there is less "risk" with 4.0.3

llogan:
Always use the latest if you can. Then you won't be dealing with bugs that have already been fixed and your users will appreciate the additional features. Note that new features post-release are not backported from the master branch into point release updates.

4.0 and 4.1 have the same major numbers ([more info](https://ffmpeg.org/doxygen/trunk/index.html)), so there should be no backward incompatible changes for dependent packages. I would have expected the 4.1.1 point release to be available by now, so if you may catch that in time if it is released soon.

llogan:
> Always use the latest if you can. Then you won't be dealing with bugs that have already been fixed and your users will appreciate the additional features. Note that new features post-release are not backported from the master branch into point release updates.
> 
> 4.0 and 4.1 have the same major numbers ([more info](https://ffmpeg.org/doxygen/trunk/index.html)), so there should be no backward incompatible changes for dependent packages.

That wasn't worded well. What I should have said is that since you're already dealing with a major version change, from 3.4 to 4.x, and since 4.0 and 4.1 are the same major version, then you might as well use 4.1.

m4tt075:
Yes, have been thinking about latest stable (= 4.1 currently) too. I've started working on the cross packages by bumping to latest stable as well. There are A LOT of them. I'm starting to truly appreciate the linuxfromscratch documentation as well...  ;-)

ymartin59:
OK for 4.1 but I think it is worth to check compilation instructions to ensure all "users" applications (tvendhead, comskip, chromaprint...) have been already updated to support FFmpeg 4 API or options (I guess there is significant changes as a major version)

ymartin59:
Abandoned. Replaced by #3575

Issue #3250, opened 2018-03-25
Design: new package "SynoCli file"


ymartin59:
Create a package for small command line utilities dedicated to file manipulation.
Revival of #2261
Tools candidates:
- less #2075, #2818 
- Tree #3242
- fdupes #2257 or jdupes #3030
- RHash #2823 #2944
- mc
- nano

Potential extension (TBD):
- rmlint #3482 (proposed by cytec) <- Requires `scons` build package (python3 based)
- detox #2199 

Not to be included:
- coreutils [cross package](https://github.com/SynoCommunity/spksrc/blob/master/cross/coreutils/) as almost all commands available already or installable via `synogear install` 
- rename (proposed by adithyabenny) <-- Part of `linux-util` package. Maybe later as part of a separate package.

Package icon: Kindly provided by @publicarray. Many thanks!

MrSpoocy:
mc ;)

hgy59:
@MrSpoocy mc is not a command line tool IMHO
I would like to keep the SynoCli packages small.

h-q:
I cannot wait for these SynoCli packages. I missed them since the Bootstrap days. Thank you so much!

ymartin59:
I also would like to provide "entware" first... to avoid to introduce (new) duplicated tools in `spksrc`

m4tt075:
Chaps, I just remembered the "synogear" package, which is actually provided by Synology. See
http://www.synology-wiki.de/index.php/Synogear:_weitere_Tools
for documentation and the list of commands included (only on German, but I did not find another source). We could provide an English translation on our Wiki page and refer to it. Not sure it covers everything, but I feel a lot of the tools requested. Just for consideration...

ymartin59:
@m4tt075 Thanks for the hint. Here it is: https://github.com/SynoCommunity/spksrc/wiki/FAQ-synogear
Interesting because some tools are painful to build... but some will complain most tools versions are quite old.

jbruchon:
I'm the maintainer of `jdupes` and the one who opened #3030. I'd like to voice my support for integrating `jdupes` by whatever means is necessary. A Synology user requested my effort to make an SPK for it 4 months ago and I've had other users since express interest in it for their NASes. As far as the age concern, I maintain the code regularly and try to be responsive to issue reports. If I need to do anything to assist in inclusion, please let me know how I can help.

adithyabenny:
`rename`? I use it often to batch rename files with a regex.

[...]

PR #3581, opened 2019-01-16
Added rsync version 3.1.3 to synocli-net

fredrike:
_Motivation:_ Synologys built-in rsync is old and have features that prohibits receiving data over ssh.

### Checklist

### DIFF
--- a/cross/rsync/Makefile
+++ b/cross/rsync/Makefile
@@ -1,0 +1,16 @@
+PKG_NAME = rsync
+PKG_VERS = 3.1.3
+PKG_EXT = tar.gz
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+PKG_DIST_SITE = https://download.samba.org/pub/$(PKG_NAME)/src
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+
+# DEPENDS = cross/ncurses
+
+HOMEPAGE = https://rsync.samba.org/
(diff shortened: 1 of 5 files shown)


ymartin59:
I am just wondering if synocli-net is the right place... What about synocli-files #3250 ?

ymartin59:
I made my mind. It is OK for me in synocli-net

ymartin59:
Waiting for Mosh fix to merge and publish: #3582 and #3583

acolomb:
So what about my [comment](https://github.com/SynoCommunity/spksrc/pull/3581#issuecomment-454948452)? How does it interact / possibly interfere with Synology's built-in rsync?

fredrike:
@acolomb I'm not sure how it will work. You must have ssh working for your user but you should not need to have configured rsync the Synology way. 
I'll get back to you once I've tested it and know exactly how it will work.

ymartin59:
I will test and check how it works. Probably user has to prepend its PATH variable in $HOME/.bashrc so that package binary has precedence over system version. By the way, should be documented in synocli-net package FAQ

fredrike:
My idea is to use `--rsync-path=` to the synocli-net version when sending files to the Synology (https://www.computerhope.com/unix/rsync.htm).

ymartin59:
Thanks for hint. I will test and document relevant options.

acolomb:
I would also think the `--rsync-path` option would be the right way to use it. I'm not sure whether rsync even invokes a shell over the remote SSH connection, so setting PATH in `.bashrc` will not even work. Plus Synology forces the users' shell to `/bin/sh` which is a symlink to bash but causes it to not evaluate the bashrc file.

So installing as `rsync-gnu` or even `rsync-community` sounds like a good idea. A symlink into `/usr/bin` is optional IMHO.

My background for the non-admin SSH question is that I gladly discovered Synology allows them to run rsync remotely, but prevents shell access. Whatever mechanism is used for this, it is enabled with the rsync service in Control Panel. I would try to replicate that permission setting for your `rsync-gnu` binary. And what I'm after is actually producing a package for [Unison](https://www.cis.upenn.edu/~bcpierce/unison/) that also works for non-admin users over SSH. Though it is still a lot of work and my time is quite limited currently.

Erwane:
Hi.

Do you known when rsync will be added and pushed toSynoCommunity please ?

PR #3575, opened 2019-01-11
Update ffmpeg to 4.1 and all dependencies to latest release, including build fixes

m4tt075:
### Motivation
`ffmpeg` is a core package for SynoCommunity, as standalone package and as key dependency for the `Chromaprint`, `Tvheadend` and `Comskip` packages. The `ffmpeg` package and external libraries (e.g. x264 and x265) are under continuous development and regularly improved. This PR bumps `ffmpeg` as well as all dependencies to latest release. 

### PR Description
- [x] Updates `ffmpeg` to 4.1 and all dependencies to latest release
- [x] Fixes HDR10+ support within `x265` 
- [x] Fixes ppc cross-compilation issues and reestablishes cross-platform support for `x265-2.9`
- [x] Enables VAAPI support for `x64` Intel platforms

_Linked issues:_ #2952, #3413, #3206, #3222 

### Checklist
- [x] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

PR #3587, opened 2019-01-20
Update comskip ffmpeg 4.1

ymartin59:
_Motivation:_ Do not re-compile and embed ffmpeg libraries
_Linked issues:_ #3575 

### Checklist


ymartin59:
@m4tt075 I am working on linking to ffmpeg package libraries. In Chromaprint, I have enforced RPATH to include ffmpeg installation libpath - but with comskip I find it easier to take benefits of pkgconfig support.

BUT ffmpeg does not generate these files properly where RPATH should be created from  INSTALL_PREFIX as /var/packages/ffmpeg/target/lib - in fact current content is /spksrc/spk/ffmpeg/work-x64-6.1/install/var/packages/ffmpeg/target/lib from objdump -x work-x64-6.1/staging/bin/comskip | grep RPATH
So it builds but will fail at runtime.

I will investigate what options are required in ffmpeg build so that pkgconfig files are as-expected.

m4tt075:
@ymartin59 Sounds good. Gee, it feels like we are getting deeper and deeper into the rabbit hole. The more you look, the more issues you find... ;-)

ymartin59:
Now requires #3595

PR #3589, opened 2019-01-21
framework: fix make install for pkgconfig files

ymartin59:
_Motivation:_ have proper information in pkgconfig to allow cross-package dependencies and library runtime loading

Reference: https://www.gnu.org/prep/standards/html_node/DESTDIR.html

### DIFF
--- a/mk/spksrc.install.mk
+++ b/mk/spksrc.install.mk
@@ -13,7 +13,7 @@
 # Variables:
 #  INSTALL_PREFIX          Target directory where the software will be run.
 #  INSTALL_DIR             Where to install files. INSTALL_PREFIX will be added.
-#  STAGING_INSTALL_PREFIX  Where to instll files, in extenso. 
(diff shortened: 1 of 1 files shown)


m4tt075:
@ymartin59 I ran into issues. Trying to compile `less`, it's required dependency, `cross/ncurses` installed in wrong directories. Reverting this commit fixed the issue. Do you want to revert or fix all cross and spk packages at once?

ymartin59:
@m4tt075 Hum... I am aware my last merge "DESTDIR in makeinstall" has impact to packages where "incorrect" prefix is explicitly set in "install" or "configure" steps.
My first idea was to identify and fix them when applying PR for updates... But I now understand it may be annoying for contributors.

So I agree with you, please revert last commit "DESTDIR in makeinstall" and I will find a way to detect and fix most (I fear I cannot expect "all") later. I am AFK for 30h...

m4tt075:
Done.

PR #3595, opened 2019-01-26
framework: fix pkgconfig rpath and rpath-link

ymartin59:
Following DESTDIR make install addition, most path are as expected.
Add rpath-link for compilation against dependant package binaries.

_Motivation:_ allow compilation and runtime linkage across different SPK packages
_Linked issues:_ #3587

### DIFF
--- a/cross/bzip2/Makefile
+++ b/cross/bzip2/Makefile
@@ -12,21 +12,21 @@ COMMENT  = High-quality data compressor
 LICENSE  = BSD License
 
 CONFIGURE_TARGET = nop
-COMPILE_TARGET = myCompile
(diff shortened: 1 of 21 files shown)


ymartin59:
@Safihre I wonder if these options may allow package wheel compilation without having to build again and again python package in each python2 application...

Safihre:
How would that work? (I'm not very well experienced in all the linking and compiling details)

ymartin59:
@m4tt075 I am working on fixing proper usage of INSTALL_PREFIX, INSTALL_DIR and STAGING_INSTALL_PREFIX in cross/ Makefiles... When ready, I will merge and publish ffmpeg and related packages.

PR #3605, opened 2019-02-03
Minetest server for synology

toffee:
_Motivation:_ Compile the [minetest](https://www.minetest.net/) server for synology.
_Linked issues:_ No issue or PRs

### Checklist
- [x] Build rule `all-supported` completed successfully. The generation of packages was successfully: for following: 88f6281 aarch64 alpine armada370 armada375 armada38x armadaxp comcerto2k evansport hi3535 ipq806x monaco northstarplus x64
- [x] Package upgrade completed successfully. All program generated files are placed into /var/packages/minetest/target/var which is not modified by update script. Package was tested on armadaxp.
- [x] New installation of package completed successfully. 

Some information:
 * The package include [minetest_game](https://github.com/minetest/minetest_game) as mentioned into compilation instructions. The script just download and extract the files to the right location.
 * Because the minetest and minetest_game download URL does not include the package name into the file name I was forced to use different versions (0.4.17.1 for minetest and 0.4.17 for minetest_game).
 * To build the server only IrrLicht library headers are needed. To build a separate package for IrrLicht libary will be inconvenient considering the custom (and not configurable - you have to comment/uncomment defines in *.h files) build system. 
 * The server does not contains internationalization support (gettext library) and no "terminal": the option --terminal will not work because was built without (n)curses support.


digitalbox94:
The package works fine on my NAS DS716+II : client 0.4.17.1 from windows connecting well on my NAS.

Do you think last version 5.0.1 can be proposed as well ?

digitalbox94:
After a few investigations, I've got game server version 5.0.1 compiling and being able to run it on my NAS :)

cross/sqlite/Makefile
```diff
@@ -15,8 +15,15 @@ GNU_CONFIGURE = 1
...
 ADDITIONAL_CFLAGS  = -O
 ADDITIONAL_CFLAGS += -DSQLITE_ENABLE_COLUMN_METADATA
 ADDITIONAL_CFLAGS += -DSQLITE_ENABLE_UNLOCK_NOTIFY
 
 include ../../mk/spksrc.cross-cc.mk
+
+.PHONY: sqlite-pre-configure
+
+sqlite-pre-configure:
+	$(RUN) autoreconf -f -i
```
hgy59, line 29: please revert the changes in this file.

sqlite was updated on the master to version 3.39.3 (3390300, 2022)
your PR would downgrade it.

BTW: If you use the SynoCommunity docker image as dev. environment (or set it up regarding the README.md) there is no need to call autoreconf to build this.

toffee: I've reverted both commits.

hgy59:
@toffee this PR has a git tree mismatch
it is probably the best to create a new PR based on the current master branch.

PR #3620, opened 2019-02-10
Accelerate build: ensure upstream pip/wheel downloads are cached

chickenandpork:
_Motivation:_ 

Building large targets such as `all-supported` takes a long long time.  One thing that bugs me is when things are unnecessarily done: like re-downloading the same wheel via PIP for each and every build.  That doesn't seem like the right thing.

Now, this won't cleave the build time in half (unless you're building is slow shared wifi like coffee shops), but avoiding some downloads will speed things up, reduce unexpected download costs, and let us do builds in coffee shops.

_Linked issues:_ none

### Checklist
- [x] Build rule `all-supported` completed successfully  (doesn't apply)
- [x] Package upgrade completed successfully  (doesn't apply)
- [x] New installation of package completed successfully   (doesn't apply)

Checklist does not apply, but I've checked them off to ensure a "3 of 3" marker on the PR.  There's no specific package to build; rather, success is recognizing that cache is populated when clean and used on subsequent builds of similar content.

Tested with a `make -C spk/python arch-avoton-6.1` to ensure that the cache is created, owned by the builder, and populated; second build leverages the cache and downloads no new wheels.

-- boilerplate -- (sorry)

### DIFF
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -351,6 +351,9 @@ publish-all-archs: $(addprefix publish-arch-,$(AVAILABLE_ARCHS))
 ####
 
 all-supported:
+ifneq ($(PIP_CACHE_OPT),)
+	@mkdir -p $(PIP_DIR)  # to ensure caching
+endif
 	@$(MSG) Build supported archs
 	@if $(MAKE) kernel-required >/dev/null 2>&1 ; then \
 	  for arch in $(sort $(basename $(subst -,.,$(basename $(subst .,,$(ARCHS_DUPES)))))) ; \
@@ -365,6 +368,9 @@ all-supported:
 	fi
 
 publish-all-supported:
+ifneq ($(PIP_CACHE_OPT),)
+	@mkdir -p $(PIP_DIR)  # to ensure caching
+endif
 	@$(MSG) Publish supported archs
 	@if $(MAKE) kernel-required >/dev/null 2>&1 ; then \
 	  for arch in $(sort $(basename $(subst -,.,$(basename $(subst .,,$(ARCHS_DUPES)))))) ; \
@@ -389,6 +395,9 @@ publish-all-legacy: $(addprefix publish-arch-,$(LEGACY_ARCHS))
 ####
 
 all-archs-latest:
+ifneq ($(PIP_CACHE_OPT),)
+	@mkdir -p $(PIP_DIR)  # to ensure caching
+endif
 	@$(MSG) Build all archs with latest DSM per FIRMWARE
 	@if $(MAKE) kernel-required >/dev/null 2>&1 ; then \
 	  $(MSG) Skipping duplicate arches; \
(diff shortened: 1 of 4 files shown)


ymartin59:
@chickenandpork Thanks. I was also wondering why cache has not been enabled - and my hypothesis was that it has been disabled to avoid to re-use binaries built for one architecture when building another. So I guess I was wrong...

chickenandpork:
I don’t think you we’re wrong; avoiding error is key.  This should cache just the original downloads.

The same issue bothers me with git-hash-based downloads, for which I’ve got some benefit with a cross/* section just to ensure the hash-based download is cached — it’s cleaner than in an spk but can appear redundant

mk/spksrc.spk.mk
```diff
@@ -351,6 +351,9 @@ publish-all-archs: $(addprefix publish-arch-,$(AVAILABLE_ARCHS))
 ####
 
 all-supported:
+ifneq ($(PIP_CACHE_OPT),)
```
ymartin59: Please move this block to `mk/spksrc.wheel.mk` in `pre_wheel_target`, it seems the best place to me.

chickenandpork:
Makes sense.  I’m back on a keyboard in a few hours, I can clean that up.

mk/spksrc.spk.mk
chickenandpork: done; the copies are reverted

ymartin59:
In fact, caching already works when running make targets in the same docker container without restart. But I agree it is interesting to move cache in host volume.

chickenandpork:
Caching should not be a choice between _clean build_ vs _faster build_: it's possible to choose to build faster without needing a pre-used buildenv to do so

Caching currently would rely on populating builduser's $HOME/.pip/cache (or similar, I don't remember) from a previous build.  We've all been hit by some odd thing hanging around from a build tainting the next unrelated build.  If you could have the benefit (caching) without the risk (unclean container), wouldn't you ?

What frustrated me was seeing the 3rd or 4th `python` build in the same run of `all-supported` re-downloading the same wheels: **same docker container without restart**, and I see it ticking off the hashmarks to mark download progress.  To me, this would imply that caching is failing **at least some times**.  ...so I tracked down why, and ensured there's a logical reason it should always work.

1. This patch clearly shows caching on multiple builds (ie `all-supported`) and restarted builds
2. **obvious**: I know where it's caching
3. **consistent**: it's caching where the other downloads are caching
4. **control**: I can nuke the cache across the board easily (`rm` command)
5. **shareable**: I can persist and re-use just the cache by mounting a common volume onto `distrib` in multiple volumes, or set `DISTRIB_DIR` to a common location

My long-term goal for building without network access is exactly that persistence of the distrib path.

chickenandpork:
(sweet!  thanks!)

hgy59:
Sorry, but the changes brought with this merge in the file `mk/spksrc.directories.mk` breaks the toolchain build (at least for apollolake-6.1 targets).

I get a lot of errors telling:
`recursive variable DISTRIB_DIR references itself (eventually) - mk/spksrc.tc.mk (16)`

please test pull requests with build of `all-supported` on a clean environment before merging - and don't ignore this on the checklist!

chickenandpork:
Hey, let me repro later tonight.

I tend to run a “make -C spk/{something} clean all-supported” to confirm a clean build but maybe if this is happening then it’s not clean enough?  :-/

Issue #3621, opened 2019-02-11
[TVheadend] Permissions fault on playing and recording


kekukui:
#### **Platform**
TVH 4.2.8.-15
DS111 / DSM 6.2.1-23824 Update 4

#### **Issue description**

1.  I cannot record anything from the tuner through the TVH web U.I.  (The default connection limits were not changed.)

2.  I can start recordings through Kodi, but the Kodi PVR client claims that recording cannot be stopped due to a permissions error.  (However, the recordings did stop according to the TVH web U.I.)

3.  When I logout from the TVH web interface, I cannot log back in without closing the tab and opening a new one

4.  If I try to play a recording in the web interface, it opens Windows Media Player and prompts for a network password.  When I enter the admin credentials, they are rejected.
______
#### **Package logs**

```
===> Step preuninst. USER=tvheadend GROUP=sc-media SHARE_PATH=
Removing service configuration tvheadend.sc
Fri Feb  1 06:46:57 EST 2019
===> Step postuninst. USER=tvheadend GROUP=sc-media SHARE_PATH=
Removing user sc-tvheadend
Sat Feb  9 03:27:31 EST 2019
===> Step preinst. USER=tvheadend GROUP=sc-media SHARE_PATH=
Sat Feb  9 03:27:34 EST 2019
===> Step postinst. USER=tvheadend GROUP=sc-media SHARE_PATH=
Installing service configuration /var/packages/tvheadend/conf/tvheadend.sc
Adding 'sc-tvheadend' to 'sc-media'
Group Name: [sc-media]
Group Type: [AUTH_LOCAL]
Group ID:   [65536]
Group Members:
0:[sc-tvheadend]
Invoke service_postinst
Fixing fontconfig links
```
> Starting tvheadend command /volume1/@appstore/tvheadend/bin/tvheadend -f -u sc-tvheadend -g tvheadend --http_port 9981 --htsp_port 9982 -c /volume1/@appstore/tvheadend/var -p /volume1/@appstore/tvheadend/var/tvheadend.pid

m4tt075:
First of all, please let me know which version you were running, before you've upgraded. As to your questions:

1. As you can record through Kodi but not through the Web UI, I assume you are connecting with different users / user credentials. Please go into the TVH Web UI, ensure "View Level" is set to "Expert"  in the upper right area of the screen. Then go to "Configuration" -> "Users" -> "Access entries". Check whether the "Video Recorder Flags" are set correctly. 
[...]

PR #3623, opened 2019-02-12
Tvheadend 4.2.8 fixes

m4tt075:
### Motivation
- CA Module support lost when upgrading to TVH 4.2.8 or TVH 4.2.7 (see https://tvheadend.org/boards/5/topics/quote/36328)
- Updates from legacy packages fail because legacy users cannot be removed anymore (see #3622)
- Creation of "admin" account does not create expected TVH server permissions  (see #3621)
- Evansport does not build anymore

### Pending
- Problems with internal HW VAAPI transcoding on x64 platforms have been reported (https://tvheadend.org/boards/5/topics/quote/36351).
- Apparently only works with development builds
- Using spawn profile might be an option (under evaluation)

### Checklist


m4tt075:
@ymartin59 I assume removing the beta flag, increased usage. At least I've seen a lot of activity and reports on our recent `Tvheadend` publication on various forums. That's good. There are users!

Two regressions popped up:  One was mine, one yours. Perfectly fair :-) The third commit is a usability improvement. 

Keeping this WIP in light of HW transcoding problems that have been reported. Might be related to our `ffmpeg 4.1` implementation as well. Let's see. Will come back as soon as I learn more...

m4tt075:
@ymartin59 Realizing that the recent `ffmpeg`-related publications do not contain any `evansport` packages, I've reintroduced my earlier fix for `nettle` here. I believe it is required.

cross/dvb-apps/patches/build-libs-only.patch
```diff
@@ -0,0 +1,13 @@
+--- Makefile.orig	2017-03-27 21:10:20.023847566 +0200
++++ Makefile	2017-03-27 21:10:46.282317463 +0200
+@@ -10,8 +10,8 @@
+ 
+ all clean install:
+ 	$(MAKE) -C lib $@
+-	$(MAKE) -C test $@
+-	$(MAKE) -C util $@
++#	$(MAKE) -C test $@
```
ymartin59: Patch will be simply shorter if lines removed

cross/nettle/Makefile
```diff
@@ -33,3 +33,6 @@ ifeq ($(findstring $(ARCH), $(x64_ARCHES)),$(ARCH))
 CONFIGURE_ARGS = --target=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu
 endif
 CONFIGURE_ARGS += --prefix=$(INSTALL_PREFIX) --enable-shared
+
+# Fix for evansport builds to find gmp libraries
```
ymartin59: I really have to understand why on hearth evansport is the only one to fail !

spk/tvheadend/Makefile
```diff
@@ -1,17 +1,17 @@
...
-DEPENDS = cross/$(SPK_NAME)
+DEPENDS = cross/busybox cross/$(SPK_NAME)
 
 MAINTAINER = m4tt075
 DESCRIPTION = Tvheadend is a TV streaming server and recorder for Linux, FreeBSD and Android supporting DVB-S, DVB-S2, DVB-C, DVB-T, ATSC, ISDB-T, IPTV, SAT IP and HDHomeRun as input sources. Tvheadend offers HTTP, HTSP and SAT IP streaming.
 RELOAD_UI = yes
 DISPLAY_NAME = Tvheadend
 STARTABLE = yes
-CHANGELOG = "1. Update TVH to 4.2.8<br/>2. Fix software transcoding with x264<br/>3. Fix ppc build<br/>4. Remove beta flag"
+CHANGELOG = "1. Fix CA module support regression<br/>2. Fix regression of upgrading legacy pacakges<br/>3. Simplify and fix initial login credential setting"
```
ymartin59:
```suggestion
CHANGELOG = "1. Fix CA module support regression<br/>2. Fix regression of upgrading legacy packages<br/>3. Simplify and fix initial login credential setting"
```

spk/tvheadend/src/wizard/install_uifile_fre
```diff
@@ -1,13 +1,6 @@
 [{
-    "step_title": "Configuration du service",
+    "step_title": "Configuration du passe administrateur",
```
ymartin59:
```suggestion
    "step_title": "Configuration du mot de passe administrateur",
```

cross/dvb-apps/patches/build-libs-only.patch
m4tt075: Sorry, I don't understand. Do you mean that you want me to remove the lines instead of commenting them out?

cross/nettle/Makefile
m4tt075: I'm pretty sure this has nothing to do with us / our framework. See here:
http://lists.lysator.liu.se/pipermail/nettle-bugs/2017/006951.html
This was also the source of inspiration for the fix.

cross/dvb-apps/patches/build-libs-only.patch
ymartin59: yes I do... less lines to maintain/review...

m4tt075:
All fine. No worries. We'll tackle everything step-by-step. I appreciate your review. And please excuse my French...  ;-)

cross/nettle/Makefile
ymartin59: These libraries gmp, mpfr, mpc are lacking proper support for cross-compiling... does not generate pkgconfig files, and I wonder when it expects to honor prefix when you read `spk/ffmpeg/work-evansport-5.2/install/var/packages/ffmpeg/target/lib/libgmp.la`
```
# Directory that this library needs to be installed in:
libdir='/usr/local/lib'
```
Really ?

m4tt075: :-)

ymartin59:
About nettle->gmp dependency, I noticed that prefix was lacking in gmp configure... what I fixed for homeassistant PR: https://github.com/SynoCommunity/spksrc/blob/5021707efe02fc816ee1e02617e02817a300b084/cross/gmp/Makefile#L17

ymartin59:
@m4tt075 What is current status? (My workspace is ready to publish tvheadend fixes...) Do you mind if I check all-supported compilation and try to find a better fix if any for nettle?

m4tt075:
@ymartin59 I do mind, if it keeps breaking my Virtualbox build environment. In the last two weeks I kept running into the same problems over and over again. I don't see the benefit in finding alternative fixes, if the ones we have are reasonable and have proven to work across platforms. Don't get me wrong, please. If you find something that works for Virtualbox as well, by all means, go for it. But I believe there is lots of stuff, that would benefit more from your attention...

m4tt075:
As to my current status: I had tested this PR with x64-6.1, which worked. As you know I could not ensure cross-compilation on all platforms (via travis and the old docker image). So that piece has been missing...
I'd also squash before merging...

ymartin59:
@m4tt075 If you are satisfied about this PR, it is ready to be published from my point of view.

m4tt075:
@ymartin59 OK. Thanks. I've squashed and force-pushed the fixups. I've seen that your new docker image has been merged, so I should be able to cross-compile using travis again. I'll now redo the PR checklist to ensure that everything still works as intended. I'll let you know as soon as all of this has succeeded.

ymartin59:
@m4tt075 Image is not ready yet in Docker Hub... upload is still in progress. I will let you know

m4tt075:
@ymartin59 I've cross-compiled all 6.1 platforms in the meantime. `x64` keeps failing via travis / docker and I can't figure out why. For details, see:
https://travis-ci.org/m4tt075/spksrc/builds/494582094
It does build on my Virtualbox environment. Do you have any idea why this is or even how I could debug the issue?

ymartin59:
@m4tt075 I wonder if libnuma has been running test cases for 10 minutes without any progress output... just an idea from log "parallel-tests: installing 'build-aux/test-driver'

ymartin59:
@m4tt075 About your travis job configuration, may you please run again without ">/dev/null"
```
docker run -it -v `pwd`:/spksrc synocommunity/spksrc:latest /bin/bash -c "cd /spksrc && make setup && cd spk/$PKG && make $SYNOARCH>/dev/null";
```
It would be easier to understand what happens. For sure, more verbose but at least travis will no longer it is inactive.

m4tt075:
@ymartin59 OK. Here the `.configure` options from `libnuma`:
```
Optional Features:
  --disable-option-checking  ignore unrecognized --enable/--with options
  --disable-FEATURE       do not include FEATURE (same as --enable-FEATURE=no)
  --enable-FEATURE[=ARG]  include FEATURE [ARG=yes]
  --enable-silent-rules   less verbose build output (undo: "make V=1")
  --disable-silent-rules  verbose build output (undo: "make V=0")
  --enable-shared[=PKGS]  build shared libraries [default=yes]
  --enable-static[=PKGS]  build static libraries [default=yes]
  --enable-fast-install[=PKGS]
                          optimize for fast installation [default=yes]
  --enable-dependency-tracking
                          do not reject slow dependency extractors
  --disable-dependency-tracking
                          speeds up one-time build
  --disable-libtool-lock  avoid locking (might break parallel builds)

Optional Packages:
  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
  --with-pic[=PKGS]       try to use only PIC/non-PIC objects [default=use
                          both]
  --with-aix-soname=aix|svr4|both
                          shared library versioning (aka "SONAME") variant to
                          provide on AIX, [default=aix].
  --with-gnu-ld           assume the C compiler uses GNU ld [default=no]
  --with-sysroot[=DIR]    Search for dependent libraries within DIR (or the
                          compiler's sysroot if not specified).
```
Any idea what to disable to avoid running tests or do I have to look somewhere else?

ymartin59:
According to last lines... I would bet for "--disable-libtool-lock  avoid locking (might break parallel builds)"

m4tt075:
@ymartin59 I've been trying to debug this issue all evening, but am stuck:
1) First, I've tried to introduce the "--disable-libtool-lock" configure option and to remove the ">dev/null" redirection in travis to see what actually happens. I failed because travis produces too many log lines and aborts the build. See here: https://travis-ci.org/m4tt075/spksrc/builds/495604755
2) Then I've reintroduced the ">/dev/null" redirection, but keeping the "--disable-libtool-lock" option, to see whether the build succeeds. It doesn't, so the "--disable-libtool-lock" option does not seem to fix it. I'm not 100% sure though, as I still cannot see what really causes the issues. See here for details: https://travis-ci.org/m4tt075/spksrc/builds/495629591
4) Then, I dug deeper and checked in my Virtualbox environment what is supposed to build after `libnuma`. It is (only) `x265`. Therefore, I've tried to build `ffmpeg`, removing all cross dependencies except of `x265`. It fails, but with an error that travis outputs, not because it stales. See here for details:
https://travis-ci.org/m4tt075/spksrc/builds/495646729
5) Lastly, I thought the problem might occur because `libnuma` is only built as a dependency of a dependency, so I removed all cross-dependencies from `Tvheadend`, except `ffmpeg`,  within `ffmpeg` I only kept `x265`. Again, the build fails, but is not stalling:
https://travis-ci.org/m4tt075/spksrc/builds/495661277

I really don't know how to proceed. I cannot make sense of the outputs above. I cannot understand why I don't encounter the issue in my Virtualbox environment, where `x64` compiles perfectly fine. I cannot understand why you seem to be able to compile `x64` on Docker. And why - still - Docker via Travis does not work. Those bloody log size limitations are preventing me from seeing what the actualy issue is. I'm really stuck. Any advice appreciated...

ymartin59:
@m4tt075 Many thanks for this detailed report. May I propose to publish package even with that issue in Docker on Travis? I will be honest: I prefer to let this alone until we deploy a completely automated PR->travis->repo build chain here.

m4tt075:
@ymartin59 I don't feel comfortable. I used to be able to develop, test and validate cross-compilation with the tools available to me. I'm not able to do that any longer. As we keep implementing package and framework changes at the same time, I cannot distinguish where issues are coming from anymore. So, how can I guarantee the quality of the fixes I propose? I'm unable to tick the boxes that we request for each and every pull request. I don't think it's the right thing to push ahead then, is it?

ymartin59:
@m4tt075 For information, compilation process looks stuck for "minutes" for each produced object file without >/dev/null
```
make[7]: Leaving directory '/spksrc/spk/tvheadend/work-x64-6.1/x265_2.9/build10bit'
make[7]: Entering directory '/spksrc/spk/tvheadend/work-x64-6.1/x265_2.9/build10bit'
[ 30%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/pixel-a.asm.o
[ 32%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/const-a.asm.o
[ 33%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/cpu-a.asm.o
[ 34%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/ssd-a.asm.o
[ 35%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/mc-a.asm.o
[ 37%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/mc-a2.asm.o
[ 38%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/pixel-util8.asm.o
[ 39%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/blockcopy8.asm.o
[ 41%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/pixeladd8.asm.o
[ 42%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/dct8.asm.o
[ 43%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/seaintegral.asm.o
[ 44%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/sad16-a.asm.o
[ 46%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/intrapred16.asm.o
[ 47%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/v4-ipfilter16.asm.o
[ 48%] Building ASM_NASM object common/CMakeFiles/common.dir/x86/h4-ipfilter16.asm.o
```
I recommend to implement proposed work-around to generate log activity and prevents build to timeout: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

ymartin59:
For compilation time improvement, I have proposed to measure time in each target... Probably this would show that x264/x265 builds are really really slow and probably does not output during 10 minutes...

m4tt075:
@ymartin59 Thanks for the hint and your patience. You were right! I've used `travis_wait` and could verify that the build indeed continues eventually. I still couldn't successfully finalize the build, as the maximum build time for public repositories of 50 minutes is exceeded. This build is just very, very complex. However, as I've been able to build `x64-6.1` and `x64-5.2` successfully via Virtualbox
(https://github.com/m4tt075/spksrc/releases/edit/tvh-4.2.8-fixes-12) AND I could see that the travis / docker build does not break in the `x64`-specific (VAAPI-enabled) `x265` part, I'm comfortable the build "would" indeed build fine as well. I've also successfully compiled all DSM5.2 archs in the meantime (`x64` again via Virtualbox), tested an update and a new installation on my XPE6.1 test system (`x64`). From all I can see, we are fine. 

As such, FINALLY, please feel free to merge and move ahead. Sorry for the hold-up...

schumi2004:
Is it possible that 1 of the changes in this commit broke the tvheadend-android for live channels app in Android? 
It now constantly crashes after upgrading to this build. I could downgrade for testing but then I need to rebuild package first using same version as used here because Synology won't let me downgrade? But I'm willing to do that if we think something here could be an issue?

ymartin59:
To downgrade: you may backup /var/packages/tvheadend/target/ content first, before uninstall and reinstall package and then restore configuration from backup... But I do not know specifically which files have to be retrieved/restored. @m4tt075 May you have detailed instructions about such process?

schumi2004:
Not going to reinstall, that's to much work imo. Rebuilding package using same version number should do it also. It's also possible that my Phillips Android TV is the bad guy here but it started to happen after upgrading tvheadend on my nas few days back. I'll rebuild package and report back.

m4tt075:
Difficult to judge. Could be related to the `ffmpeg 3.x` to `ffmpeg 4.x` upgrade or an upstream `Tvheadend` problem as well. I agree with you, @schumi2004. Let me know what you find, ok?

schumi2004:
@m4tt075
Should I also downgrade ffmpeg if I undo latest commit?

m4tt075:
If you want to get rid of `ffmpeg 4.x`, you'd have to downgrade back to the `tvh-4.2.7` version we had. If everything was fine with `tvh-4.2.8` and the problem really only appeared with `tvh-4.2.8-fixes`, I'd actually assume the problem to be upstream with `tvh` and the changes from 4.2.7 to 4.2.8. In that case I'd inquire with Jaroslav via opening a ticket on tvheadend.org.

schumi2004:
I'm currently building package with version tvh-4.2.7.
If that works i'll build package version tvh-4.2.8-15 and after that i'll try the tvh-4.2.8-fixes.
It's also highly possible it's the TV (Philips 49PUS6412) itself (Nougat) in combination with the pretty old app https://github.com/kiall/android-tvheadend.
Building and testing can take a while since my system isn't that fast but will report back asap.

m4tt075:
Yeah, same here, Randy! Still...  :-)

schumi2004:
Took a while but last night package tvh-4.2.7 was compiled but it was to late (01:30) to test.
Just installed it on my NAS (DS715) 1 hour ago and validated functionality with other clients, all okay.
Started from scratch with TVHeadend Live Channel for Android TV on my Philips with several reboots in between and deleting data were possible and it only produced a error on first run after setup and didn't crashed yet. The first run error was there also the previous time. 
I'm not yet convincend it has something to do with the tvh-4.2.8 (fixed) package but the app didn't crashed yet.
I wouldn't spend much time on resolving this though and contacting Jaroslav for it. The TVH android app is not updated recently and although most apps (even Kodi) run fine on the Philips Android TV it sometimes can be a bitch running things and throwing errors. It's just coincidental that it happened when I upgraded to tvh-4.2.8-fixes.
I'll compile tvh-4.2.8 and tvh-4.2.8-fixes later today but can't test them any time soon, next week probably.

m4tt075:
@schumi2004 Thanks for the feedback. Appreciated. And happy it works for you for the time being. Just let me know if you need help. I still remember you helping me a looooooong time ago when I was new to all of this. Cheers!

PR #3627, opened 2019-02-16
ffmpeg: fix GMP install prefix

### LINKED ISSUE #3625: Standard Docker build environment broken?
### Setup
_Package Name:_ `Tvheadend` (for example)
_Package Version:_ 4.2.8-fixes (see PR #3623)

_NAS Model:_ Various
_NAS Architecture:_ Various
_DSM version:_ 6.1

### Expected behavior
Should compile in standard Docker build environment
[Edit: Formerly reported problems with VirtualBox builds had different rationale. Removed as obsolete.]

### Actual behavior
Builds on Standard Docker image (pulled via travis) fail with:
```
CMake Error at CMakeLists.txt:4 (cmake_minimum_required):
  CMake 3.1 or higher is required.  You are running version 3.0.2
``` 
See https://travis-ci.org/m4tt075/spksrc/builds/492410154 for more examples

### Steps to reproduce
Try building `tvh-4.2.8-fixes` with standard Docker image (pulled in via travis)

ymartin59:
Closes #3625

_Motivation:_ GMP framework configure "prefix" option is lost somewhere in the process...
_Linked issues:_ #3625 #3623 #3462

### DIFF
--- a/cross/gmp/Makefile
+++ b/cross/gmp/Makefile
@@ -11,8 +11,6 @@ HOMEPAGE = http://gmplib.org/
 COMMENT  = GMP is a free library for arbitrary precision arithmetic, operating on signed integers, rational numbers, and floating point numbers
 LICENSE  = LGPL
 
-include ../../mk/spksrc.cross-cc.mk
-
 GNU_CONFIGURE = 1
 ifeq ($(findstring $(ARCH), $(ARM5_ARCHES)),$(ARCH))
 CONFIGURE_ARGS = --target=arm-unknown-linux-gnueabi --host=arm-unknown-linux-gnueabi --disable-assembly
@@ -32,3 +30,5 @@ endif
 ifeq ($(findstring $(ARCH), $(x64_ARCHES)),$(ARCH))
 CONFIGURE_ARGS = --target=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu
 endif
+
+include ../../mk/spksrc.cross-cc.mk


m4tt075:
Just tested this but it does **not** fix the nettle compilation for `evansport` platforms. I also remember having trouble with compiling `gmp` across platforms. And I believe I played around with the order of arguments in the `Makefile`, like you do here. Please don't merge at this stage.

ymartin59:
@m4tt075 By the way, it fixes ffmpeg compilation. I have published ffmpeg/comskip/chromaprint for evantsport

m4tt075:
@ymartin59 Then it is build environment related. It does NOT fix the problem in my Virtualbox environment...

ymartin59:
@m4tt075 OK sorry - I think I understand now - as I build with docker image, it does not depend on any "build environment" condition - and I guess in your specific case, nettle "configure" does find "libgmp" at default location in your VM Linux system first. (but I also agree it is my fault to have removed options I considered useless after confirming build was ok with docker)
Cross-compilation is already complex and adding constraints from any Linux distribution running in an home-made VM make things even worse. Even if we face also troubles with "docker on macos x", I think docker provides the most stable environment possible so that all contributors have same conditions to generate packages.
Is there any option for your to run docker container? Even recent windows 10 provides docker support...

ymartin59:
Technically speaking that PR grants that "libgmp.la" content is correct (without it, libdir path is /usr/local/lib which is irrelevant) and then allows other "configure" to generate variables with proper value to build code depending on libgmp.
I do not think it may break any other package compilation.
May I merge it and then confirm "all-supported" passes for tvheadend #3623?

m4tt075:
Merge it, but keep my nettle gmp fix in #3623, PLEASE!

PR #3628, opened 2019-02-16
Python update to 2.7.15

chickenandpork:
_Motivation:_ This is basically @ymartin59's #3552 but with:
   - builds successfully (I wasn't able to repro the problem is with #3552)
   - rebased to latest SynoCommunity/spksrc/master
   - `cross/python/PLIST` is parameterized
   - updates to libsodium's patches used instead of rewriting
   - some text/descriptions slightly different (shouldn't affect build)
   - `Apachev2` license is `ASL2`

The changes above are minimal, and reflect the kind of cleanup that @ymartin59 tends to do to mine anyhow :)  So, honestly, I haven't really done anything to his work except redo it by hand, which can avoid maybe quotation marks that look like other quotes or something else.

Maybe the rebasing to recent version is what did it.

_Linked issues:_ #3546 

### Checklist

Checked upgrade of `2.7.14-19` to `2.7.15-20`

Basic in-place test is:
```
/var/packages/python/target/bin/python   -c 'import _ctypes; import lxml; import sqlite3; print("x");'
```


ymartin59:
@chickenandpork Thanks Allan. Looks great. You are correct when you wrote all Python 2 based applications have now to be rebuilt and published... That was in my plans since vulnerability alerts about "requests" in almost all "requirements.txt": https://github.com/SynoCommunity/spksrc/network/alerts

chickenandpork:
Hmm.  Another big build job.

It’s like we need to have some sort of a fan-out on validation of builds and acceptance.

Safihre:
So what is different from the other PR #3552? Looks exactly the same. Strange it builds now.

The only reason all packages need to be rebuild is the `requests` package vulnerability?

chickenandpork:
I think the biggest thing was rebasing.

I hand-replicates the same changes, skipping what was rebased out.  Bumped bcrypt as a side-effect.

I parameterized the PLIST after but it should not affect compile.

Safihre:
@ymartin59 Did you actually read the vulnerability that was fixed in `requests`?
https://github.com/kennethreitz/requests/issues/4716

This is not something that requires republishing of all packages, as it requires a potential attacker to be able to perform man-in-the-middle attacks on networks.

Safihre:
Python 2.7.16 will come in the near future. 
https://github.com/python/cpython/blob/2.7/Misc/NEWS.d/2.7.16rc1.rst

I am not sure if this will affect this PR.. I certainly hope not because users have been waiting for months to get the new SABnzbd and this is all blocked by this update.

chickenandpork:
I would suggest merging this now and doing 2.7.16 when it’s available.

Always waiting for the next release means we never move ahead.

This builds, I think the builds are already in my beta space (spk.chickenandporn.com), rather work on 2.7.16 when it drops.

Safihre:
@chickenandpork this is a bit off topic: what email can I reach you on? I would also like to setup maybe a spk server. You can also reach me at safihre@sabnzbd.org :)

PR #3560, opened 2018-12-27
Update openssl to 1.0.2q

STemplar:
Motivation: update OpenSSL 1.0.2 branch to version q.
Linked issues:

### Checklist
- [] Build rule `all-supported` completed successfully
- [] Package upgrade completed successfully
- [] New installation of package completed successfully

PR #3629, opened 2019-02-16
WIP: (building) Bump openssl to 1.0.2q

chickenandpork:
_Motivation:_ Security
_Linked issues:_ 

According to `all-affected-spks`, this will require an extensive build:

```
make -C spk/mutt all-supported
make -C spk/irssi all-supported
make -C spk/jackett all-supported
make -C spk/ejabberd all-supported
make -C spk/python all-supported
make -C spk/lidarr all-supported
make -C spk/radarr all-supported
make -C spk/museek-plus all-supported
make -C spk/python3 all-supported
make -C spk/homeassistant all-supported
make -C spk/octoprint all-supported
make -C spk/sonarr all-supported
make -C spk/synocli-net all-supported
make -C spk/itools all-supported
make -C spk/rdiff-backup all-supported
make -C spk/rutorrent all-supported
make -C spk/sabnzbd all-supported
make -C spk/ffsync all-supported
make -C spk/umurmur all-supported
make -C spk/boxbackup-client all-supported
make -C spk/mercurial all-supported
make -C spk/links all-supported
make -C spk/deluge all-supported
make -C spk/borgbackup all-supported
make -C spk/sabnzbd-testing all-supported
make -C spk/duplicity all-supported
```

... see you next week!  I honestly assume something in this list will fail to build but for unrelated reasons, so there might be a pre-PR with some version-bumps that are necessary to build this one.

### Checklist

### DIFF
--- a/cross/openssl/Makefile
+++ b/cross/openssl/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = openssl
-PKG_VERS = 1.0.2n
+PKG_VERS = 1.0.2q
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://www.openssl.org/source
--- a/cross/openssl/digests
+++ b/cross/openssl/digests
@@ -1,3 +1,3 @@
-openssl-1.0.2n.tar.gz SHA1 0ca2957869206de193603eca6d89f532f61680b1
-openssl-1.0.2n.tar.gz SHA256 370babb75f278c39e0c50e8c4e7493bc0f18db6867478341a832a982fd15a8fe
-openssl-1.0.2n.tar.gz MD5 13bdc1b1d1ff39b6fd42a255e74676a4
+openssl-1.0.2q.tar.gz SHA1 692f5f2f1b114f8adaadaa3e7be8cce1907f38c5
+openssl-1.0.2q.tar.gz SHA256 5744cfcbcec2b1b48629f7354203bc1e5e9b5466998bbccc5b5fcde3b18eb684
+openssl-1.0.2q.tar.gz MD5 7563e1ce046cb21948eeb6ba1a0eb71c


ymartin59:
@chickenandpork Large effort. What is specific motivation for this version upgrade? I have already python3 and borgbackup built and published waiting for activation when homeassistant will be ready too...

ymartin59:
@chickenandpork You have proposed a PR to include your "all-affected-spks" target. This is closed today. Have you got a more recent version to submit?

chickenandpork:
That PR was dirty from “all-affected-spks” testing.  I’ll push a cleaner one.

chickenandpork:
Keeping SSL libraries updated tends to be a safe course of action.

The testing will require more than a week of build time, but we cannot very easily have different versions of dependencies.

I’ll need more horsepower to do builds faster, so it’ll be a few days.

Allan

ymartin59:
For information, some  websites (and probably many more soon) follow a "rude" recommendation to only allow TLS 1.3 because of flaw in handshake in previous TLS versions. And this requires at least openssl 1.1.1...

chickenandpork:
So, due to big build requirements, would you suggest to make a bigger jump to OpenSSL-1.1.x ?

Corporate side, we flip between different TLS libraries: the rudeness you observe might also be turning off vulnerable TLS versions due to static analysis tools flagging them as errors.

This is to say: they might not be doing it to be obnoxious, but because a tool they’ve committed to honor to avoid legal risk is now alerting.  It’s a hard choice between supporting customers despite risk.

Given the build demand this PR entails, I don’t want to do builds on it yet until I’ve got a bigger machine up.  I would estimate the dependency to cause over a week of builds on my current environment and I need to ensure at least the first build involves an empty cache.

m4tt075:
For what it's worth, it seems someone tried to introduce `openssl-1.1.x` earlier and ran into issues with `rtmpdump`, and, thus, `ffmpeg`. At least that is written in the "Constraints" chapter of our "Update Policy" Wiki entry here:
https://github.com/SynoCommunity/spksrc/wiki/Update-Policy
With the latest `ffmpeg` update, `rtmpdump` has been removed. So at least that contraint should no longer exist. Still a major task of course. Thanks for looking into this!

ymartin59:
@m4tt075 I did try to upgrade to openssl 1.1 and wrote that section - happy it has been useful. So we may try again and check "most" applications' compatibility.

chickenandpork:
Possible dupe of #3560 .. working on bigger HW (for the massive build demand) so I’ll back off and see if 3560 merges.

There’s perhaps a need for automated canary and federated acceptance/bless of versions across architectures

hgy59:
as openssl is updated to 1.0.1.u this PR got obsolete.

PR #3631, opened 2019-02-17
Revert "framework: accelerate build, cache upstream pip/wheel downloa…

hgy59:
…ds (#3620)"

This reverts commit 416f001e72902577074ff9259fd9c610f2156d6c.

Sorry, but the changes brought with this merge in the file `mk/spksrc.directories.mk` breaks the toolchain build (at least for apollolake-6.1 targets).

I get a lot of errors telling:
`recursive variable DISTRIB_DIR references itself (eventually) - mk/spksrc.tc.mk (16)`

please test pull requests with build of `all-supported` on a clean environment before merging - and don't ignore this on the checklist!

### DIFF
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -15,9 +15,7 @@ RUN = cd $(WORK_DIR)/$(PKG_DIR) && env $(ENV)
 
 # Pip command
 PIP ?= pip
-# Why ask for the same thing twice?  Always cache downloads
-PIP_CACHE_OPT ?= --cache-dir $(PIP_DIR)
-PIP_WHEEL = $(PIP) wheel --no-binary :all: $(PIP_CACHE_OPT) --no-deps --requirement $(WORK_DIR)/wheelhouse/requirements.txt --wheel-dir $(WORK_DIR)/wheelhouse --build-dir $(WORK_DIR)/wheelbuild
+PIP_WHEEL = $(PIP) wheel --no-binary :all: --no-deps --requirement $(WORK_DIR)/wheelhouse/requirements.txt --wheel-dir $(WORK_DIR)/wheelhouse --build-dir $(WORK_DIR)/wheelbuild
 
 # Available languages
 LANGUAGES = chs cht csy dan enu fre ger hun ita jpn krn nld nor plk ptb ptg rus spn sve trk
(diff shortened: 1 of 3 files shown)


ymartin59:
@hgy59 Thanks for reporting. Instead of reverting commit, I will try to fix it today. CC @chickenandpork

ymartin59:
@chickenandpork Error happens when downloading/extracting toolchain if not available at `toolchains/syno-xxx-y.z/work` - I have reproduced with "work" folder removal in any toolchain.

ymartin59:
Reviewing #3620 I find no clue how this PR would have introduced such failure... It should come from older commit/merge...

chickenandpork:
Does it actually stop a build?

I know it sucks to have warnings, and I haven’t seen them myself yet, but the three checkboxes don’t talk about warnings, and the build takes really long on my hardware (1515+ running docker/Jenkins/gogs) so I don’t watch it.  Clearly I missed something that makes us look a bit clowny.

I also look at everything with an eye towards automation.

Can we consider this a non-show-stopper issue, and discuss instead how to detect it during a build?  Count that the number of warnings of a certain reflex doesn’t exceed some whitelist value?  That would allow a PR build someday to audit a build and flag recurrence of this.

I’m sorry this happened.  I feel responsible.  I still need to repro it but I hope to fix it this weekend.

ymartin59:
Oh yes. If toolchain is not yet ready in working copy, build fails immediately.
I have the patch ready and I am testing "download", "digests" and other compilation target... No need to be sorry, framework variables are not so clean in many places - you will understand when I publish PR.

ymartin59:
Here it is: #3632

chickenandpork:
> please test pull requests with build of `all-supported` on a clean environment before merging - and don't ignore this on the checklist!

So, I *do* that, every time.  My build is typically:

```
docker run -it --rm -v $(pwd):/spksrc -w /spksrc synocommunity/spksrc:latest make -C spk/{package} clean all-supported
```

Note:
 - `--rm` deletes the container when complete, so the next invocation is pristine
 - `clean` cleans things (to clean the work environment on the backing disk mapped into `/spksrc`)

The reason this didn't show up for me is that -- like the PR -- I re-use caching as much as possible for pristine upstream content to avoid needlessly re-downloading things.  My bandwidth is barely able to support Netflix :)  Also survives things like when Julian let slip the `bzip` domain name, making `bzip2` non-downloadable hence unbuildable -- remember that?  `clean` doesn't strip the cache.

Maybe I need to cache upstream one hop, at a proxy or something.

Thanks for fixing, @ymartin59

PR #3632, opened 2019-02-17
framework: fix DISTRIB_DIR recursive definition

ymartin59:
DISTRIB_DIR and DISTRIB_FILE are input parameters for spksrc.download.mk
To avoid mistake, "distrib" folder is now identified as BASE_DISTRIB_DIR

_Motivation:_ Fix toolchain operations (download, digests, extract, patch)

### DIFF
--- a/mk/spksrc.directories.mk
+++ b/mk/spksrc.directories.mk
@@ -11,11 +11,13 @@
 
 PWD := $(shell pwd)
 
-DISTRIB_DIR  = $(PWD)/../../distrib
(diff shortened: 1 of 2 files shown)


hgy59:
LGTM not reverting the PR.

I was testing the new docker image for the build environment against my custom image (debian:9.3) and as often used a clean environment without any download.

Was able to fix the build by reverting the changes in the spksrc.directories.mk only and therefore thought it was introduced with this PR.

PR #3642, opened 2019-02-28
Bump imagemagick to stable version

stefaang:
_Motivation:_ Imagemagick failed to build as the package is no longer available on http://www.imagemagick.org/download/releases/
The alternative is to fetch the source code from github: https://github.com/ImageMagick/ImageMagick/releases/tag/7.0.7-30
I know there's more recent builds but they don't keep the source of the recent builds on their release page.

### Checklist

### DIFF
--- a/cross/imagemagick/Makefile
+++ b/cross/imagemagick/Makefile
@@ -1,8 +1,9 @@
 PKG_NAME = ImageMagick
-PKG_VERS = 7.0.7-30
-PKG_EXT = tar.xz
-PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = http://www.imagemagick.org/download/releases
+PKG_VERS = 7.0.8-53
+PKG_EXT = tar.gz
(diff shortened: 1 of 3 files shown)


ymartin59:
Thanks for contributing. Why not replace download location from site release by github releases location - if it becomes new reference?

stefaang:
I had the impression most projects here prefer to stick to the original website, even when github releases are available. I don't mind to update the MR to only change the download location, it's probably less effort for you guys.

m4tt075:
That's not a problem. Updating to recent stable releases is certainly appreciated. I believe that @ymartin59 meant that there is no need to shy away from changing the download location to obtain those.

hgy59:
The source of imagemagick is still available for download on the existing source location.
Today they published release 7.0.8-33 so there is no need to switch to the not yet current github release location.
It looks like on the original site they publish a kind of stable releases.
Don't know why the 7.0.7-30 release we used for the spk was removed from the original site (but there is release 7.0.7-39 to download)

stefaang:
But next week they may release 7.0.8-34 and version 7.0.8-33 will be gone (for the same reason 7.0.8-32 is already gone). This is why it's better to use github as a source.

PR #3646, opened 2019-03-03
Wireguard [WIP]

publicarray:
_Motivation:_ I think wireguard is a good use-case to install on permanent on devices such as NASs and Routers
_Linked issues:_ https://www.reddit.com/r/synology/comments/8ra6io/running_wireguard/

### Checklist

Todo (help needed)
- [] Survive restarts
- [] Start-stop scripts
- [] Possibly a convenience script to make it easier to add new clients to the server configuration and download the client-configuration as a file.


fbartels:
Just stumbled upon https://github.com/runfalk/synology-wireguard not sure how helpful this is here.

publicarray:
@fbartels Thanks, yes this could be very helpful!

fastcat:
Is there anything folks (like me) could do to help move this PR along?

publicarray:
Hi @fastca, the binaries compile and work. The only problem is to make it work with Synology's network, the iptable rules and overall better integration. I suppose just merging this to have the binaries would be enough. But I feel users could accidentally break their network connection.

publicarray:
Supporting both NASs and Routers proved to be a challenge when I attempted to script this.

fastcat:
Well I'm reasonably handy with iptables & shell scripts, though I'm currently ignorant of how Synology manages their front-end/wrapper around iptables. I will try to take a look and see what I can figure out.

But I only have a NAS, no router, so if routers are the pain point I may not be able to help all that much :/

kernel/syno-ipq806x-1.2/digests
```diff
@@ -1,3 +1,3 @@
-synogpl-7742-ipq806x.tbz SHA1 a335c22c8bc67a795181c976b43ffee0ea07317d
-synogpl-7742-ipq806x.tbz SHA256 e7a8bbc3129ce4dbaac5c1c054c710015d33cec448c74209805152554e9f95c3
-synogpl-7742-ipq806x.tbz MD5 c92af5840b1296d6485fddd94ead4308
+ipq806x-linux-4.4.x.tar.xz SHA1 7cad56bfbbd5071f63e671166046ae97b2c87bbc
+ipq806x-linux-4.4.x.tar.xz SHA256 c44f7139bc416f0f8b96879b5718e2d33be96adf7a4cb38ae65b04394298aaf2
+ipq806x-linux-4.4.x.tar.xz MD5 8bf294be92f7d1e3ea04b41d915a08e7
```
ymartin59, line 3: This mistake should be fixed in a dedicated "urgent" PR. @hgy59 @th0ma7

ymartin59:
@publicarray LGTM. What do you expect as help to remove "work-in-progress" marker?

kernel/syno-ipq806x-1.2/digests
publicarray: If I remember correctly I extracted the kernel from the huge archive on a VPS and only downloaded the kernel to my machine, so the new digests are probably wrong here too.

AlexPresso:
Hey there,
Is this PR still on its way ? 🥹 It would be so nice if we could use Wireguard on DSM
Thank you very much

PR #3601, opened 2019-01-30
Add toolchain ipq806x-1.2 for Synology SRM 1.2

andrew-grechkin:
This toolchain can be used to build packages for Synology RT2600ac with SRM 1.2

### Checklist
- [*] New installation of package completed successfully

PR #3647, opened 2019-03-03
Add SRM 1.2 toolchains (not all are tested, hence no digest file)

publicarray:
_Motivation:_ I think WireGuard needs a newer kernel so I updated to the toolchain to SRM1.2 #3646
_Linked issues:_ #3601

### Checklist


ymartin59:
Interesting. As these three architectures seems to be really closed to each other, I wonder if a single binary may work properly on all of these...
CC @hgy59 Have you got any opinion about it?
@publicarray Which arch are you running? I am curious to know if a "dsm" package for any of already available ARM archs may run on SRM (typically aarch64...)

publicarray:
Hi @ymartin59
My router is ARMv7 and as far as I know is not compatible with AArch64 / ARM64 or I think sometimes referred to as ARMv8 (at least in go lang)

I don’t know about arm but I like to refer to Wikipedia: https://en.m.wikipedia.org/wiki/ARM_architecture

publicarray:
Maybe a binary compiled on ARMv7 works on ARMv8?

hgy59:
@publicarray digest files have nothing to do with testing the build. The digests are for verifying the downloaded files are original (complete and not tamperd).
And it is easy to create the digest files with `make digests` in the toolchain directories.
To verify the digests compare the values in the generated digest file (SHA1, MD5) with the respective values on the download page on sourceforge.net (click the i-symbol to view details)

hgy59:
@ymartin59 we already have the ipq806x and the northstarplus archs in synocommunity.

The dakota is new and introduced with MR2200ac (see in the [wiki ](https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures) under Routers)

What is different ist the fact that on the routers not DSM but SRM is running, and SRM has a different Version (current 1.2) but current DSM version is 6.2.

The SoC in MR2200ac is a Qualcomm IPQ4019 with ARM Cortex A7 processor.
IMHO the ARM Cortex A7 is of the ARMv7 family.

I would propose to build and ARMv7 and tagging the spk for MR2200ac with SRM 1.2 and arch dakota.
As with other generic archs I suppose that the spk is build only once for a generic arch and than "tagged" with multiple specific archs (to document the generic archs outside the mk-files I started the architectures wiki page).

But afterwards we need a user who owns such a device to verify the installation.

ymartin59:
@hgy59 Thanks. I have just compared compilers and headers files to find out northstarplus and ipq806x are strictly identical... probably generic ARM compiler without any generation optimisation.
I sometimes build and publish packages for SRM and I would be pleased to support that third new target, decreasing work from two builds to a single one.
I will generate some test packages to validate with @publicarray

hgy59:
@ymartin59 what is planned for the future of SRM support?
There are some packages that use REQUIRED_DSM 5.0 (or 5.2) and those packages will never build for SRM (northstarplus/ipq608x/dakota) since TCVERSION 1.x is less than 5.x.

I have not yet analyzed the reason to use REQUIRED_DSM, but my suggestion is to remove it and use UNSUPPORTED_ARCHS instead.
When building with `make all-supported` the only arch that does not support TCVERSION 6.1 ist ppc853x (DMS 5.2 only). 
Since there is no supported arch now, that does not fulfill the requirement of DSM 5.2, we can remove the REQUIRED_DSM from the whole build chain. Can't we?

ymartin59:
@hgy59 REQUIRED_DSM may be useful if package depends on specific components like nginx or recent php only available from DSM >= 6... I am not sure such application is expected to run on SRM. But for sure I agree to discard REQUIRED_DSM if < 6

ymartin59:
@hgy59 @publicarray I would like to progress there.
Based on https://github.com/SynoCommunity/spksrc/pull/3695/files#diff-f3784b95b6bddec9da0aa7033e4c7539R12
it looks like SRM 1.2 is based on DSM 5.2 and declares as is... So I consider my idea to set TCVERSION==1.x is no longer relevant.
- What about to rename toolchains with version 5.2?
- Why not test to reuse binaries for SRM and DSM when architectures are closed or identical? (it will change almost nothing to SRM users - who still have to download and install packages manually)

ymartin59:
@hgy59 Do you think it may be possible to create an syno-armv7-1.2 toolchain for the three router architectures?

hgy59:
> @hgy59 Do you think it may be possible to create an syno-armv7-1.2 toolchain for the three router architectures?

I have tried to do this (see #3940)

Issue #3652, opened 2019-03-10
[synocli-net] not available for ppc538x


hgy59:
When building synocli-net with `make all-supported`, there is no package generated for ARCH=ppc538x (and TCVERSION=5.2)
The dependency /cross/fuse fails to build for ARCH=ppc538x and the build is aborted.

On the [download site](https://synocommunity.com/package/synocli-net) the package for ppc538x is missing too.

cytec:
Thought we dropped ppc support all together a while ago?

hgy59:
I dont think so
`ppc538x-5.2` and `qoriq-6.1` are still built with `make all-supported`

Active Models with ppc538x:
- DS110+, DS210+, DS410

Active Models with qoriq:
- DS213+, DS413

cytec:
OIC... we did drop support for older ppc models which didn't get the 5.x update... nevermind

PR #3663, opened 2019-03-24
Fix build of sshfs for ppc853x (#3652) and itools

hgy59:
_Motivation:_ It started with the missing `synocli-net` package for `ppc853x`
_Linked issues:_ #3652

An update of `fuse` from version 2.9.4 to 2.9.9 (and disabling the build of the examples code that failed for ppx853x) was required to fix the build of the `sshfs` package that is part of `synocli-net`.
So I verified the packages depending on cross/fuse
- sshfs
- synocli-net
- itools

The `itools` package did not compile for any arch without the fix of `cross/fuse`.
The build of itools for armv5 (88f9621) failed due a bug that was fixed about 5 months ago in the repo of [usbmuxd](https://github.com/libimobiledevice/usbmuxd/commit/96e4aabe0b9a46ea9da4955a10c774a8e58fe677#diff-67e997bcfdac55191033d57a16d1408a).

Finally the package `sshfs` was updated with correct version information.
And the /cross/libusb uses now the source from github with the same version as the former sf.net repo.

### Checklist
- [x] Package upgrade completed successfully (tested with synocli-net only)

The only impediment found, is for DSM 6.x. To use the updated fusermount command, the path must be specified (/usr/local/bin), otherwise /usr/bin/fusermount used that is already installed by synology in the former version.

### Proposals
1. As the build of `synocli-net` is fixed with this PR, I propose to remove the dedicated `sshfs` and `tmux` packages that are contained in synocli-net.
2. Since `itools` depends on `fuse` it could be integrated into `synocli-net` (**or** both `itools` and `sshfs` may be moved to synocli-file).


ymartin59:
@hgy59 Well done ! What do you think about my proposal to move "spk/xxx" to a new "diyspk/" folder where we keep per-application package for advanced users which expects to create a dedicated package?

hgy59:
I think we can freely delete such spk-definitions. By moving them to a special folder they will not be maintained anymore (as spk/sshfs was). Anyone interested in such sources can find them after deletion in the history of the repo.
May be it is worth to tag the master branch with special labels before the deletion of such spk definitions, to make it easy to retrieve it for interested developpers.

ymartin59:
@hgy59 It is not a so great effort to maintain "diyspk/xxx" Makefile variables consistent with "cross/xxx"... I doubt any "interested developper" will really have interest in building a dedicated package for an old version...
May you please fix your "root" Git commit author here too.

hgy59:
@ymartin59 sorry for the wrong user, may be I made the commits from within the docker container ;-(

ymartin59:
@hgy59 Do you recommend/expect some packages to be published again? (lacking architectures in repository for instance)

hgy59:
@ymartin59 yes the ppc853x arch is still missing in the repository.
And with synocli-net published for ppc853x you may close #3652.
itools (iOS Access) in the current version is not yet published for 6.1 88f8621, 1.1 northstarplus and 1.1 ipq806x.

PR #3669, opened 2019-03-26
Update templates

hgy59:
_Motivation:_ The checkboxes of PR description look ugly if not filled with a space (or x)

- add space to make [ ] look like checkbox
- add space as separator from _italic_ text

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,12 +1,12 @@
 _For new Package Requests, see the guidelines_  
 
 ### Setup
-_Package Name:_
-_Package Version:_
+_Package Name:_ 
+_Package Version:_ 
 
-_NAS Model:_
-_NAS Architecture:_
-_DSM version:_
+_NAS Model:_ 
+_NAS Architecture:_ 
+_DSM version:_ 
 
 ### Expected behavior
 Tell us what should happen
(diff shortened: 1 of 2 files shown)


chickenandpork:
@hgy59 I don't mean to bike shed or ask a dumb question, but are the trailing spaces at end-of-line in `.github/ISSUE_TEMPLATE.md` necessary to improve readability?  It seems unusual that a space at EOL alters the line appearance, so I'm curious.

Otherwise, do the spaces `[]` --> `[ ]` still show as, for example, "1 of 3" when summarizing that 1 of three check boxes are checked (and the others have spaces)?  If that's the case, that makes this change risk-free, and having no vested interest either way, I'd recommend merging in the next merge of build logic.

hgy59:
@chickenandpork no, **without** spaces it is not shown as checkbox and is not counted as check to be done.
often I see PRs with
- [] Build rule all-supported completed successfully

instead of 
- [ ] Build rule all-supported completed successfully

or
- [x] Build rule all-supported completed successfully

**And for Issues**
It looks ugly if one forgets to add a space after the _

_Package Name:_sabnzbd
_Package Version:2.3.7
_NAS Model: Synology DS 418play

_instead of:_
_Package Name:_ sabnzbd
_Package Version:_ 2.3.7
_NAS Model:_ Synology DS 418play

Well, these are just templates that anyone might modify, but I think the default should have real checkboxes and should help to separate italic labels from entered text.

chickenandpork:
Thanks for the detailed explanation :)

ymartin59:
@hgy59 Please notice that your Git commit author is "root". May you please setup your Git client with "better" email address and author name. Thank you in advance.

hgy59:
@ymartin59 sorry, still learning some basic git

PR #3670, opened 2019-03-27
[WIP] Add Gitea

typeless:
This is still a work-in-progress

For https://github.com/go-gitea/gitea/issues/4669

I plan to download the tarball directly from, say, https://dl.gitea.io/gitea/1.8/
The tarball contains only the executable of Gitea, and it will create the required directories when launched at the same location as the executable.


typeless:
Given that Gitea requires three directories beside its sole executable for
1. a configuration file
2. a database file (sqlite)
3. Git repositories
4. log files

Are there recommended/conventional filesystem locations for them?

typeless:
It seems there is already prior art.
https://github.com/flipswitchingmonkey/gitea-spk

PR #3698, opened 2019-05-08
updated tmux to version 2.9a

MatthiasLohr:
_Motivation:_  Updated tmux to the last version
_Linked issues:_  n.a.

### Checklist


MatthiasLohr:
@ymartin59 Thanks for merging so quick! Can I somehow help with publishing?

ymartin59:
@MatthiasLohr I am trying to include utemper first: #3583

MatthiasLohr:
I would strongly suggest to work things seperately. tmux obviously works without it, so i would suggest to add it (now). If you add another library later which adds new features or other things, you can still do an update. Otherwise you create a really long backlog of unpublished things and people are getting demotivated in contributing here again. I already spoke to some people thinking about a fork of SynoCommunity because of the current politics and speed of the project.

For now, this tmux update is an isolated task. Adding/combining it with other (completely arbitrary) things is 1) hard to understand and 2) increasing more difficult to maintain.

Issue #2479, opened 2016-09-30
Tcl  stop with  error message


biloucat:
Package Name: tcl
Package Version: 1.0

NAS Model: DS412+ Intel Atom D2700 Architecture: cedarview (x64)
NAS Architecture:
DSM version: 5.2

### Expected behavior
tclsh should run

### Actual behavior
tclsh stop with  the message:
 error while loading shared libraries: libtcl8.6.so

it seems for me that the link  tcl -> /var/packages/tcl/target in /usr/local is missing

KaraokeStu:
@eelcohuininga Could you have a look into this please?

KaraokeStu:
I had a quick look and it appears that TCL needs to load shared libraries - these are linked to the correct location, but LD_LIBRARY_PATH is not set correctly.

Setting this via
   export LD_LIBRARY_PATH=/usr/local/lib

Makes it all work properly

biloucat:
i think that /usr/local/  is in the library path but that tcl is not in /usr/local because the link /usr/localtcl -> /var/packages/tcl/target is missing

biloucat:
if i remove  the link /usr/local/tcl -> /var/packages/tcl/target, without doing nothing else , tclsh break;
if i redo the link  /usr/local/tcl -> /var/packages/tcl/target, tclsh start again

biloucat:
perhaps the links in /usr/local/lib:
  libtcl.so -> /volume1/@appstore/tcl/lib/libtcl8.6.so 
  libtcl8.6.so -> /volume1/@appstore/tcl/lib/libtcl8.6.so
are useless but the link 
  tcl8.6 -> /volume1/@appstore/tcl/lib/tcl8.6/
is useful !

eelcohuininga:
Hi,
Can you check a couple of things for me?
1. Can you confirm that the symlink ```/usr/local/lib/libtcl8.6.so``` points to ```/volume1/@appstore/tcl/lib/libtcl8.6.so```?
2. Does ```tclsh``` work again if you de-install the package, and then re-install it?

eelcohuininga:
```tclsh``` needs the LD_LIBRARY_PATH to be set to the location of ```libtcl8.6.so```. This is set by the installer script, but I guess this is wrong, because if you reboot your Synology, the LD_LIBRARY_PATH never gets set. I think I'll have to move it to the start-stop-status script in order for it to get set every time the package is started.

eelcohuininga:
[...]

PR #3720, opened 2019-06-08
Updated Tcl to v8.6.9-3

eelcohn:
_Motivation:_  This update will fix several issues in the current Tcl 8.6.9-2 package:
_Linked issues:_  #2479

### Checklist

This update will fix several issues in the current Tcl 8.6.9-2 package:

* Added `DESTDIR` variable to `cross/tcl/Makefile` so that the `tclsh`-executable and `libtcl8.6.so`-library compile with the correct destination path. Several issues with the Tcl v8.6.9-1 package were reported by @biloucat in #2479, but they were never really fixed.
Adding this `make`-environment variable will fix them. See also https://github.com/tcltk/tcl/blob/master/unix/Makefile.in#L37
* Reverted cleanup of `spk/tcl/PLIST`. The `PLIST` file was cleaned up in the v8.6.9-2 package, but this actually broke some commands in Tcl because the `sqlite3_analyzer` and `*.tcl` files are required for Tcl to function properly.
The `sqlite3_analyzer` script is bundled by default with Tcl and developers will trust that it is avaiable when Tcl is installed. Some Tcl scripts will fail if it's not present. Removing this will therefore break several Tcl programs. See for more info this document: https://www.tcl.tk/community/tcl2017/assets/talk93/Paper.html
The `pkgIndex.tcl`-files and the accompanying `.tcl`-files are used to identify and load packages. By removing these bundled scripts we effectively break Tcl.
E.g. if you remove the `lib/tcl8.6/package.tcl` script and then try to do `package require itcl` from the `tclsh` command line, you'll get an error message saying `couldn't read file "/var/packages/tcl/target/lib/tcl8.6/package.tcl": no such file or directory`.
These files are mandatory for Tcl and without these the package loading won't work. See for more info the Tcl Wiki (https://wiki.tcl-lang.org/page/pkgIndex%2Etcl).
* Fixed 64-bit compiling for 64-bit CPU's
* Cleaned up non-functioning code from `cross/tcl/Makefile`


cross/tcl/Makefile
```diff
@@ -2,36 +2,31 @@ PKG_NAME = tcl
...
+LICENSE = http://tcl.tk/software/tcltk/license.html
 
 GNU_CONFIGURE = 1
 
-# Prevent re-declaration of fixstrtod.o
-# export ac_cv_func_strtod=yes
-# export tcl_cv_strtod_buggy=1
-
-CONFIGURE_ARGS = tcl_cv_strtod_buggy=1 --prefix=$(INSTALL_PREFIX)
+DESTDIR = /var/packages/$(PKG_NAME)/target
```
ymartin59: I guess this line is now useless as standard "install" target is no longer replaced by "tcl_custom_install". Please refer to https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.install.mk#L59

Issue #3638, opened 2019-02-24
Add open-zwave for HomeAssistant


ymartin59:
This is an endless long-running story since 2011:
- http://www.livingroomfun.info/blog/20111020/openzwave-running-on-synology-ds1010/
- https://www.nas-forum.com/forum/topic/32654-domotique-open-source-des-volontaires-pour-un-vrai-projet/
- issue about libudev dependency https://github.com/OpenZWave/open-zwave/issues/590
- openWRT includes a patch to disable libudev dependency
- upstream PR for same patch has been closed without merge: https://github.com/SynoCommunity/spksrc/pull/1769/files
- included in Domoticz from jadahl / jumbotroll but build receipts have not been published: http://www.domoticz.com/forum/viewtopic.php?f=56&t=2620&start=1560

Related subject, kernel modules for USB controllers:
- https://forum.synology.com/enu/viewtopic.php?f=155&t=82843&start=690
- http://www.domoticz.com/forum/viewtopic.php?f=47&t=7087&sid=b11752ab6ae3d7e4e9f25b21aa2fd9dd&start=420

I will give it another try myself...

ymartin59:
@chickenandpork You may be interested by this issue

ymartin59:
Buildling openzwave library is OK but then python-openzwave has it own library detection and compilation process which fails when cross-compiling...

PR #3731, opened 2019-07-02
Homeassistant 0.87.1 and Python 3.6.8 - fix cross-compilation issues

roleoroleo:
_Motivation:_   I'm trying to update Home Assistant and the last version deprecates Python 3.5
_Linked issues:_  https://github.com/SynoCommunity/spksrc/issues/3673

I cross-compiled them on x86_64 using a debian 9.9 with docker.
And I was able to test them on armada370.

Python seems to work correctly:
```
roleo@DiskStation:~$ /var/packages/python3/target/bin/python3   -c 'import sqlite3; import _ctypes; import lxml; import sys; print ("OK:", sys.platform, sys.byteorder, "endian", sys.version)'
OK: linux little endian 3.6.8 (default, Jul  2 2019, 12:36:20)
[GCC 4.9.3 20150311 (prerelease)]
roleo@DiskStation:~$ /var/packages/python3/target/bin/python3   -c 'import sqlite3; import _ctypes; import platform; import lxml; import sys; print ("OK:", sys.platform, sys.byteorder, "endian", sys.version, "machine:", platform.machine())'
OK: linux little endian 3.6.8 (default, Jul  2 2019, 12:36:20)
[GCC 4.9.3 20150311 (prerelease)] machine: armv7l
```

Also hass seems to work correctly, even if the first boot is very very slow.


schumi2004:
Could you compile a version for Alpine maybe since i'm not building it using Docker but with a Ubuntu Virtual box image and that 1 fails.
Maybe you have more luck using Docker ;)

roleoroleo:
Using docker it works.

python3_alpine-6.1_3.6.8-9.spk
https://drive.google.com/open?id=1ZVvH2nm8oEVGTZpIBoa_uzl9ofFhPjR1

homeassistant_alpine-6.1_0.87.1-3.spk
https://drive.google.com/open?id=1t_WE14R4lTW72v6Eyy5ansANSuvPcVQu

schumi2004:
Okay, so future compiling needs to be done using Docker it seems.
Great that it works, will test asap.

schumi2004:
@roleoroleo  It runs perfectly, great job!

schumi2004:
@roleoroleo Do you maybe know a way to install hassio to install plugins?

fredrike:
Why Home-Assistant 0.87.1 when 0.95 is the latest?

roleoroleo:
No particular reason. Only because the 0.87.1 was in the master.
We could probably try to compile also the 0.95.

schumi2004:
Did it work? Could you share the Alpine SPK for testing purpose? :D

roleoroleo:
I didn't try.

Safihre:
Could you give it a try and let me know? Then maybe I can assist in publishing. It seems @ymartin59 is busy or on vacation 🏖

roleoroleo:
I'd like to try to compile 0.95.
But I don't know how the requirements.txt file was created
Who decided which components to enable? Based on what logic?

fredrike:
I think that since Debian is using Python  3.7 it would be reasonable to build hass upon that too..

The different versions of hass are only dependent on a python version, there rest should be built via a venv so that hass can upgrade itself (or let the package dl the latest version on installation/upgrade) 


> but  I don't know how the requirements.txtwas created



It is created based on all available components. It is enough to build based on the `bin/setup` file 
https://developers.home-assistant.io/docs/en/development_environment.html#setting-up-virtual-environment

Or follow these instructions https://www.home-assistant.io/docs/installation/raspberry-pi/

fredrike:
What I mean is that the only wheel needed are the ones from the setup script not the whole requirements.txt.

roleoroleo:
Yes, but if you don't compile the wheels, who installs them in the virtualenv?

fredrike:
Hass installs the dependencies each of the component needs. So building and installing the home-assistant pip package should be enough.

chickenandpork:
roleoroleo all that’s needed are the critical dozen or so wheels; recall that this is running HA in a docker container that can be deployed with a consistent toolchain independent of host. Hass manages containers, including bolting-on containers for various modules.

My concern is how well it works with other containers in the docker — does it kill unknown containers (thinking it “owns” the docker) or leave them alone?

schumi2004:
And how about devices that can't run docker? I like the package how it is, it works and functions perfectly. Hass.io is just overkill imo.

chickenandpork:
The list of compatible architectures is fairly slim; I don't think this would means we don't need a native home-assistant package.

@schumi2004 > Hass.io is just overkill imo.

I can see that opinion; alternatively, I do like how the components Hass bolts-on are pre-configured and work.  just-add-unifi?  awesome.  Auto-git-pull a config?  way cool.  As above, I worry about how the rest of the ecosystem is affected -- getting Nginx or traefik to do the lets-encrypt dance for me, configuring forwarded ports, will it nuke my Jenkins and Gogs containers, etc.

I think this adds an option for some of the Syno community, and is a remarkable enhancement if it plays nicely.  I see @fredrike has been very detailed.  I'm a bit gun-shy to risk my setup to test without some assurance that @fredrike has other containers undisturbed in the same docker.

fredrike:
There are actually three usage modes for home-assistant.

1. Devices that can't run docker and this is where this package comes in handy (this package is using venv a vertual python environment).
2. Devices that can run docker and have a somewhat clueless user, this is where a hass.io package like #3732 comes in handy. Hass.io have support for managing related docker containers (and don't touch containers it have not created itself).
3. Devices that can run docker and have user that want full control. Here the user start its own homeassistant docker container and have to manually upgrade it.

Imho there should be two spk packages for use caseu one and two and for use case three there is quite good documentation at https://www.home-assistant.io/docs/installation/docker/#synology-nas

fredrike:
> roleoroleo all that’s needed are the critical dozen or so wheels; recall that this is running HA in a docker container that can be deployed with a consistent toolchain independent of host. Hass manages containers, including bolting-on containers for various modules.
> 
> My concern is how well it works with other containers in the docker — does it kill unknown containers (thinking it “owns” the docker) or leave them alone?

This package runs home-assistant in venv so no docker fuss. Are you referring to #3733?

chickenandpork:
> > My concern is how well it works with other containers in the docker — does it kill unknown containers (thinking it “owns” the docker) or leave them alone?

> This package runs home-assistant in venv so no docker fuss. Are you referring to #3733?

No, I’m referring to “and don't touch containers it have not created itself” which is good news.  I was a native-HA, change to a HA-in-docker, but I’d rather be lazier and go to a hass.io.  I’m not a “clueless user” just.. only have so much time.

fredrike:
> No, I’m referring to “and don't touch containers it have not created itself” which is good news. I was a native-HA, change to a HA-in-docker, but I’d rather be lazier and go to a hass.io. I’m not a “clueless user” just.. only have so much time.

I agree..  _only have so much time_. I think a good working hass.io would be beneficial for everyone but some users want the fuzz of fixing and trixing themselves.

roleoroleo:
I tried to build 0.95 but no luck.
Unfortunately the recent versions of various packages (bcrypt 3.1.7 for example) require pep 517 so I can't use the --no-use-pep517 option in "pip wheel".
Regarding the fact that the packages are installed directly from pip, I think this is not entirely true for platform dependent modules on arch than x86_64. Consider that I am trying to compile hass for armada370 and monaco.

ymartin59:
To notice that recent ARM architectures do not have Docker package either (even if docker may run on arm and i386 too...). For my own needs, I also tried to cross-compile python-openzwave wheel but fails because of rules logic there: #3638

mk/spksrc.wheel.mk
```diff
@@ -62,7 +62,9 @@ build_wheel_target: $(PRE_WHEEL_TARGET)
 		$(foreach e,$(shell cat $(WORK_DIR)/python-cc.mk),$(eval $(e))) \
 		if [ ! -z "$(CROSS_COMPILE_WHEELS)" ] ; then \
 			$(MSG) "Force cross-compile" ; \
+			cp $(WORK_DIR)/install/$(INSTALL_PREFIX)/$(PYTHON_LIB_DIR)/$(_PYTHON_SYSCONFIGDATA_NAME).py $(PYTHON_SITE_PACKAGES_NATIVE)/../ ; \
 			$(RUN) CFLAGS="$(CFLAGS) -I$(STAGING_INSTALL_PREFIX)/$(PYTHON_INC_DIR) $(WHEELS_CFLAGS)" LDFLAGS="$(LDFLAGS) $(WHEELS_LDFLAGS)" $(PIP_WHEEL) ; \
+			rm $(PYTHON_SITE_PACKAGES_NATIVE)/../$(_PYTHON_SYSCONFIGDATA_NAME).py ; \
```
ymartin59: Is this removal really required? This "specializes" `native/python3/work-native` content and prevent it to be re-used for any other architecture.
As a result, `make arch-x64-6.1 arch-aarch64-6.1` no longer works as expected (and cannot be built in parallel either...
Isn't possible to keep architecture specific files in `spk/python3/work-$ARCH-$TCVERSION` and leave `native` as-is?

ymartin59:
Python 3.6.8 is OK on my evansport with borgbackup package 1.1.9
But I expect not having to clean `native/python3/` before building each architecture... or else it means to compile python3 twice for each arch.

mk/spksrc.wheel.mk
roleoroleo: Your review is correct but I have not succeeded in any other way. I tried for a long time without success.
I think it's a pip problem: it compiles in a safe environment but doesn't known cross-compilation.
I known that this solution is dirty. As you write, the cp specializes the work-native, I added the line with rm just to delete the library after pip ran and restore the previous native.
But, in this way it cannot be compiled in parallel.

roleoroleo:
I found another solution. I thought I had already tried it but evidently ...
Instead of the 2 lines 'cp' and 'rm' I added a new line in the python-cc.mk file to add a path to the environment variable PYTHONPATH.
In this way the work-native is untouched.
I don't know how elegant it is but it seems to work.
```@echo 'ENV+=PYTHONPATH=$$(PYTHONPATH):$(PYTHON_LIB_CROSS)' >> $@```

I am very sorry, but as I told you in the other topic, I lost all my forked repo.
Is there another way to edit this PR? (I'm a github newbie)
If it's not possible I could refork the repo and create a new PR.

ymartin59:
Hello. I have a copy of your branch locally but it will not help you to update this PR. So yes, I think the easiest/fastest option is to re-fork, but create a dedicated branch first (to avoid to push on master) to host your commit - so that your master can be left clean and updated from "upstream": I found this tutorial about the workflow which should help you: https://gist.github.com/Chaser324/ce0505fbed06b947d962

mtonnie:
@roleoroleo and @ymartin59 
Any news why this PR has been closed without merging?
Is there any progress for python3 support somewhere else?

Thanks in advance

miklosandras:
Any update on this?

PR #3732, opened 2019-07-04
Add support for Hass.io

fredrike:
_Motivation:_  The Home Assistant package runs HA natively in Python, the preferred way is to run hass.io and let it manage HA via Docker.

### Checklist


chickenandpork:
FWIW I See you've taken great care to provide a lot of things to make this a well-behaved and functionality-complete config.  I worry that the supported architecture is very limited.

In hassio.sh@32 it might be worth noting that `172.30.32/24` is hard-coded in hass supervisor?  (is it?)  Same file, line 36, might be worth documenting the logic of the "attach to running image or attempt to start a new image to attach to" if I'm understanding correctly.

You've clearly looked closely at details, and I expect this is running on your person system?  It makes me want to try it on my avoton.

I notice you're commenting on #3731 as well.

fredrike:
`172.30.32/24` is hardcoded in hassio-supervisor. I've been in contact with the team of hassio for the changes in this and most of the hass.sh is adapted from the original `hassio-service` file. https://github.com/home-assistant/hassio-installer/blob/master/files/hassio-supervisor

fredrike:
@chickenandpork When can I expect this to be merged?

chickenandpork:
Often that depends on who else has time to test and verify.  The builds for Synology are really expensive in terms of time and test tiers, so they get scheduled when there are some ready and the personal projects are also complete.  My original python3.5.6 took so long, 3.6.8 was out.

I do think we need a way to automate some portion of testing so that we can check for compatibility a bit more autonomously.  I’m used to working at companies that do this automatically, trying to think on how we can move to something a bit more efficient.  I’ve got some work in that area to explore.

Personally, I’m curious to try your work in a week or so when I get back from travels.

fredrike:
I have 16 :heart: [here](https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/125559, ) not sure if that means users that have tested and verified but at least it's something :smile:

Tack your time and enjoy your travels @chickenandpork

jokerigno:
Well, if this can help, I use it on my xpenology NAS for 2 weeks. 
Very stable and reliable!

ymartin59:
@fredrike Thanks for this contribution.
I have a question about service start/stop, I find it strange to chain busybox start-stop-daemon, a custom script and then docker run. From my point of view (probably naive), it should be enough start_daemon method to invoke "docker run" and stop_daemon to invoke "docker stop". May you explain why implementation seems so "complicated"?

spk/hassio/src/hassio.sh
```diff
@@ -0,0 +1,42 @@
...
+}
+
+start_hassio () {
+    SUPERVISOR="$(jq --raw-output '.supervisor' ${CONFIG_FILE})"
+
+    HASSIO_IMAGE_ID=$(/usr/local/bin/docker inspect --format='{{.Id}}' ${SUPERVISOR})
+    HASSIO_CONTAINER_ID=$(/usr/local/bin/docker inspect --format='{{.Image}}' hassio_supervisor || echo "--")
+
+    # Fix routing
+    route -vn |grep 172.30.32.0 >/dev/null || \
```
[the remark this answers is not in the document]
fredrike: Hass.io creates a separate network `172.30.32.0/24`, there is a bug in the Synology Docker that does not automatically craters a route for that.

spk/hassio/src/hassio.sh
```diff
@@ -0,0 +1,42 @@
...
+        route add -net 172.30.32.0/23 gw 172.30.32.1
+
+    # Run supervisor
+    ([ "${HASSIO_IMAGE_ID}" = "${HASSIO_CONTAINER_ID}" ] && /usr/local/bin/docker start --attach hassio_supervisor) || runSupervisor
```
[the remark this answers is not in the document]
fredrike: Yes this is hass.io kung fu (copied from the original hassio.service scripts.

spk/hassio/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+HOMEPAGE   = https://www.home-assistant.io/hassio/
+LICENSE    = Apache License 2.0
+HELPURL    = https://www.home-assistant.io/docs/
+
+WIZARDS_DIR = src/wizard/
+
+SERVICE_SETUP = src/service-setup.sh
+SSS_SCRIPT = src/start-stop-status
+FWPORTS = src/${SPK_NAME}.sc
```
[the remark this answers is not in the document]
fredrike: I'll review that.

spk/hassio/Makefile
```diff
@@ -0,0 +1,48 @@
...
+include ../../mk/spksrc.spk.mk
+
+.PHONY: hassio_extra_install
+hassio_extra_install:
+	install -m 755 -d $(STAGING_DIR)/bin
+	install -m 755 src/hassio.sh $(STAGING_DIR)/bin/
+	install -m 755 -d $(STAGING_DIR)/var
+	install -m 644 src/hassio.json $(STAGING_DIR)/var/hassio.json
+	install -m 755 -d $(STAGING_DIR)/app
+	install -m 644 src/app/config $(STAGING_DIR)/app/config
```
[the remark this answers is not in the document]
fredrike: I'll review that.

spk/hassio/Makefile
```diff
@@ -0,0 +1,48 @@
...
+ENV += BUSYBOX_CONFIG="$(BUSYBOX_CONFIG)"
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: hassio_extra_install
+hassio_extra_install:
+	install -m 755 -d $(STAGING_DIR)/bin
+	install -m 755 src/hassio.sh $(STAGING_DIR)/bin/
+	install -m 755 -d $(STAGING_DIR)/var
+	install -m 644 src/hassio.json $(STAGING_DIR)/var/hassio.json
```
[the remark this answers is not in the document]
fredrike: What do you mean with `backup/restore`? This file is configured during installation and should be kept like that. I don't see how it can benifit from another location?

spk/hassio/src/hassio.sh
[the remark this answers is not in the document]
fredrike: https://github.com/home-assistant/hassio-installer/blob/cbae7bc68adcec26b2a1eafd8849d25c0354e2f5/files/hassio-supervisor#L42

fredrike:
@ymartin59 When a package is upgraded, will the files in the new package replace the old ones (in my case the `hassio.sh` file)?

What is the best way to ensure that the `hassio.json` file is moved to **etc** when upgrading? Will this do it 46e4571?

chickenandpork:
Had a `entrypoint.js` issue in your `d1b1997`; I'll rebuild and try your `7b9dbcc` soonest.  I'm using the simplest build command we have to ensure a clean environment:

```
docker run -it --rm \
  -v $(pwd):/spksrc -w /spksrc 
  synocommunity/spksrc:latest \
  /bin/bash   -c  'make clean  &&  make  spk-hassio-avoton-6.1'
```

spk/hassio/src/service-setup.sh
```diff
@@ -0,0 +1,55 @@
...
+    ln -s busybox ${SYNOPKG_PKGDEST}/bin/start-stop-daemon
+
+    # Fix install directory
+    mkdir -p "${DATA_DIR}"
+}
+
+service_preuninst() {
+    HOMEASSISTANT="$(jq --raw-output '.homeassistant' ${CONFIG_FILE})"
+    SUPERVISOR="$(jq --raw-output '.supervisor' ${CONFIG_FILE})"
+    HASSIO_DATA="$(jq --raw-output '.data // "/usr/share/hassio"' ${CONFIG_FILE})"
```
chickenandpork: `CONFIG_FILE` or `CFG_FILE` per lines 23-25 above?

spk/hassio/src/service-setup.sh
```diff
@@ -0,0 +1,55 @@
...
+    URL_VERSION="https://s3.amazonaws.com/hassio-version/stable.json"
+
+    HOMEASSISTANT_DOCKER="homeassistant/qemux86-64-homeassistant"
+    HASSIO_DOCKER="homeassistant/amd64-hassio-supervisor"
+
+    # Read infos from web
+    HASSIO_VERSION=$(curl -s $URL_VERSION | jq -e -r '.supervisor')
+
+    # Pull supervisor image
+    /usr/local/bin/docker pull "$HASSIO_DOCKER:$HASSIO_VERSION" >/dev/null
```
chickenandpork: This process of pulling, tagging the supervisor image, would it be difficult to move this logic to a separate file, and run it both here and when the service restarts?  This would allow it to handle things like a timeout blocking the pull, installation finishes, but the image is unusable -- by re-checking on restart, it would allow a user to retry by restarting.  It might also get around the issue of upgrading a `hassio` when the `homeassistant` isn't running causing no `homeassistant` to be running after re-installing.

fredrike: I guess we could. But that is not the way hassio normally works.

spk/hassio/src/service-setup.sh
fredrike: I'm not sure :)

fredrike: It should be `CFG_FILE`.

spk/hassio/src/service-setup.sh
fredrike: How should the test for this look? If i move it to `hassio.sh` I mean, hassio is responsible to update itself. But if this fails the first time we will have an issue. Wouldn't it be easier to just fail the installation?

fredrike:
> Had a `entrypoint.js` issue in your `d1b1997`; I'll rebuild and try your `7b9dbcc` soonest. I'm using the simplest build command we have to ensure a clean environment:
> 
> ```
> docker run -it --rm \
>   -v $(pwd):/spksrc -w /spksrc 
>   synocommunity/spksrc:latest \
>   /bin/bash   -c  'make clean  &&  make  spk-hassio-avoton-6.1'
> ```
> 
> _EDIT 2019-08-03 05:57Z_ `7b9dbcc` works much better. manually `docker rm` `homeassistant`/`hassio_scheduler` images and containers and installed. Second restart gets me a `hass.io` add-ons list. This is much better. Still configuring.

I'm not sure why we sometimes have the `entrypoint.js` issue. I can't really reproduce it anymore so I think it is gone.

schumi2004:
Does this mean we won't see any native HA package as seen and compiled here?
https://github.com/SynoCommunity/spksrc/pull/3732

fredrike:
> Does this mean we won't see any native HA package as seen and compiled here?
> #3732

No, the two packages complements each other. This hassio package is for devices that support Docker. The native home assistant package is for those that don't have such a device.

bruvv:
I tested it and when you install hass.io you cannot use VPN server package anymore. The network interface where you can select on which interface you want the VPN server to run at. Is empty and you cannot select anything. The error has to do with the hassio adapter that gets installed (after removing hassio it works again) other than that I cannot wait for this to be merged :)

lethalwallabies:
I'm testing on synology ds918+ but the USB fix is still not functioning, anything I can try to help figure this out?

fredrike:
> I tested it and when you install hass.io you cannot use VPN server package anymore. The network interface where you can select on which interface you want the VPN server to run at. Is empty and you cannot select anything. The error has to do with the hassio adapter that gets installed (after removing hassio it works again) other than that I cannot wait for this to be merged :)

That is really strange. I've been spending some time reading the source code for the VPN package and I can't figure out what made it break.

@lethalwallabies contact me via email (fredrik.e@gmail.com) and we'll try to get things working.

spk/hassio/src/service-setup.sh
chickenandpork: Hey @fredrike , what I was saying is "there is a tagging logic that pulls, tags, runs the supervisor image.  maybe also do it when the service restarts?" ; additionally: "this can moved to a separate script and called in two places to avoid maintaining copies of the same logic".  That is to say, if a script present in the SPK does the work of pulling, tagging an image, and is run from two places (current and in a restart) then easier to maintain.  I wasn't suggesting to have the script added to hassio itself and downloaded.

This could also be done as a shell-function, or as duplicate lines of code in two places.  Whichever.  Either way, because the reused function/behavior is present int hr SPK, a failed download would be remediable by restarting the SPK, which would effectively re-pull, re-tag, and start the supervisor.

chickenandpork:
Still poking with a `6dd0666c` built locally and installed; I think RPi hassio benefits from the notion that if the app stops, the system reboots: in Synology, the hassio docker container stopping doesn't restart the service, so the container doesn't restart until I remove and re-install the `hassio_supervisor` by uninstalling the SPK.  It's a bit of a nuisance.

I stopped it to try to get back the `add ons` side-menu bar -- it disappeared, acting like a non-hassio run of the container.  Not certain why.

Able to test an update if you have one to ensure I'm on the correct rev.

ymartin59:
Hello.
I really find the proposed "start-stop" sequence [ DSM -> start-stop-script -> busybox start-stop-daemon -> hassio script -> containers { hassio_supervisor, homeassistant } ] too complex.
And if it is already known not to work properly, I propose to start from scratch to get minimal working setup (so leaving RPi baseline alone).

Why not:
- write a generic simple start-stop-status script  to start/stop/query status of a single Docker container
- ask DSM to automatically restart package service if stopped - because of automated restart after configuration change, auto update or crash

bruvv:
whats the update on this?

fredrike:
> Hello.
> I really find the proposed "start-stop" sequence [ DSM -> start-stop-script -> busybox start-stop-daemon -> hassio script -> containers { hassio_supervisor, homeassistant } ] too complex.

It is borrowed from own cloud.. 
> And if it is already known not to work properly, I propose to start from scratch to get minimal working setup (so leaving RPi baseline alone).
> 
It is working really well. 
> Why not:
> 
> * write a generic simple start-stop-status script  to start/stop/query status of a single Docker container

That is what hassio.sh does today. The hassio-container handle start stop of the other containers. 
> * ask DSM to automatically restart package service if stopped - because of automated restart after configuration change, auto update or crash

How do I do that (can't find that setting anywhere). 

I could update the start-stop script but it is really convenient to have the hassio.sh script that does all the Synology stuff (we need to manually set routing, fix usb-device symlinks and so on). I'm actually thinking of having the start-stop script pull the latest hassio.sh from gist so eventual fixes can be propegated automatically.

chickenandpork:
>> ask DSM to automatically restart package service if stopped
> How do I do that

It's possible he may be referring to the setting of `--restart` in a docker command line: https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart

I use `docker-compose` and tend to always set a restart to `always` but since the supervisor controls the rest, I guess only supervisor would be always restarted.  I get times where the supervisor is running but `homeassistant` container is not, perhaps when it "restarts" (ie quits and assumes everything will restart?) when controlled to from within the GUI.

fredrike:
> > > ask DSM to automatically restart package service if stopped
> 
> > How do I do that
> 
> It's possible he may be referring to the setting of `--restart` in a docker command line: https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart
> 
> I use `docker-compose` and tend to always set a restart to `always` but since the supervisor controls the rest, I guess only supervisor would be always restarted. I get times where the supervisor is running but `homeassistant` container is not, perhaps when it "restarts" (ie quits and assumes everything will restart?) when controlled to from within the GUI.
> 
> I think this is really close to complete

No, package center should restart the package, not Docker restarting the container.

fredrike:
> > Why not:
> > 
> > * write a generic simple start-stop-status script  to start/stop/query status of a single Docker container
> 
> That is what hassio.sh does today. The hassio-container handle start stop of the other containers.
> 
> > * ask DSM to automatically restart package service if stopped - because of automated restart after configuration change, auto update or crash
> 
> How do I do that (can't find that setting anywhere).
> 
> I could update the start-stop script but it is really convenient to have the hassio.sh script that does all the Synology stuff (we need to manually set routing, fix usb-device symlinks and so on). I'm actually thinking of having the start-stop script pull the latest hassio.sh from gist so eventual fixes can be propegated automatically.

@ymartin59 What about using `SERVICE_SETUP        = src/service-setup.sh` and this:

```sh
service_prestart ()
{
    # Replace generic service startup, fork process in background
    echo "Starting hass.io at ${SYNOPKG_PKGDEST}" >> ${LOG_FILE}
    COMMAND="hassio.sh"

    cd ${SYNOPKG_PKGDEST};
    ${COMMAND} >> ${LOG_FILE} 2>&1 &
    echo "$!" > "${PID_FILE}"
}

service_poststop ()
{
    docker stop hassio_supervisor homeassistant
}
```

And remove `busybox` and the `start-stop-status` script? What about telling DSM to auto restart?

@hgy59, do you know how to tell DSM to auto restart a crashed package?

fredrike:
@ymartin59 2269c99 removes dependency of busybox.

spk/hassio/src/service-setup.sh
```diff
@@ -0,0 +1,93 @@
...
+}
+
+service_preuninst() {
+    if [ ${SYNOPKG_PKG_STATUS} == "UNINSTALL" ]; then
+        HOMEASSISTANT="$(jq --raw-output '.homeassistant' ${CFG_FILE})"
+        SUPERVISOR="$(jq --raw-output '.supervisor' ${CFG_FILE})"
+        HASSIO_DATA="$(jq --raw-output '.data // "/usr/share/hassio"' ${CFG_FILE})"
+
+        docker rm --force homeassistant hassio_supervisor hassio_dns
+        docker image rm ${HOMEASSISTANT} ${SUPERVISOR}
```
fredrike: should remove `hassio_dns` image too.

spk/hassio/src/service-setup.sh
```diff
@@ -0,0 +1,93 @@
...
+    fi
+}
+
+service_preuninst() {
+    if [ ${SYNOPKG_PKG_STATUS} == "UNINSTALL" ]; then
+        HOMEASSISTANT="$(jq --raw-output '.homeassistant' ${CFG_FILE})"
+        SUPERVISOR="$(jq --raw-output '.supervisor' ${CFG_FILE})"
+        HASSIO_DATA="$(jq --raw-output '.data // "/usr/share/hassio"' ${CFG_FILE})"
+
+        docker rm --force homeassistant hassio_supervisor hassio_dns
```
fredrike: Perhaps make sure that `homeassistant` really is based on `homeassistant/qemux86-64-homeassistant`

bruvv:
@fredrike are you reviewing your own stuff now :P ?

fredrike:
> @fredrike are you reviewing your own stuff now :P ?

Yes, no one else is. It is just an reminder I do not think this is crucial.

bruvv:
@ymartin59 why is this taking so long? it has been over 4 months now...

spk/hassio/Makefile
```diff
@@ -0,0 +1,45 @@
+SPK_NAME = hassio
+SPK_VERS = $(shell date +%Y%m%d)
```
ymartin59: Please create your own package versioning then. Because this way prevents "reproductable" build.

spk/hassio/Makefile
```diff
@@ -0,0 +1,45 @@
...
+SPK_REV = $(shell git rev-parse --short HEAD)
```
ymartin59: Revision must be incremental number. Please start at 1

spk/hassio/Makefile
```diff
@@ -0,0 +1,45 @@
...
+SPK_ICON = src/$(SPK_NAME).png
+SPK_DEPENDS = "Docker>=18.09.0-0506"
+DSM_UI_DIR = app
+
+DEPENDS =
+
+MAINTAINER = fredrike+hassio@gmail.com
+DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder."
+ADMIN_PORT = 8123
```
ymartin59: Redundant with `SERVICE_PORT`, please remove `ADMIN_PORT`

spk/hassio/Makefile
```diff
@@ -0,0 +1,45 @@
...
+HOMEPAGE   = https://www.home-assistant.io/hassio/
+LICENSE    = Apache License 2.0
+HELPURL    = https://www.home-assistant.io/docs/
+
+WIZARDS_DIR = src/wizard/
+
+SERVICE_SETUP = src/service-setup.sh
+
+#DSM UI integration
+SERVICE_PORT = 8123
```
ymartin59: This port is reserved for native python "Home Assistant" package: https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports
Please select another free port number.

chickenandpork: Can we configure that `homeassistant` and `hassio` (reminding that this installs a managed homeassistant) -- essentially two paths to the same result -- are incompatible, such that one cannot be installed when the other is present, and allow re-use of the `homeassistant` port for either instance of the same project?

spk/hassio/src/service-setup.sh
```diff
@@ -0,0 +1,93 @@
...
+        DATA_DIR="${wizard_share_path}/${wizard_folder_name}"
+        # Fix install directory
+        mkdir -p "${DATA_DIR}"
+
+        HOMEASSISTANT_DOCKER="homeassistant/qemux86-64-homeassistant"
+        HASSIO_DOCKER="homeassistant/amd64-hassio-supervisor"
+
+        # Read infos from web
+        URL_VERSION="https://s3.amazonaws.com/hassio-version/stable.json"
+        HASSIO_VERSION=$(curl -s $URL_VERSION | jq -e -r '.supervisor')
```
chickenandpork: If this config was pulled at SPK build time as a dependency of building the SPK, it could be interpreted for a constant version (see `spk/hassio/Makefile` line 2 discussion).  I'm not sure it's the best method, but it's more consistent than `$(date +...)` which sets a different version on the same thing depending on when it's done.

I'm not sure whether the package should display the Supervisor rev (currently 192) or the homeassistant rev (currently 0.102.3) based on what a user thinks they're seeing.

fredrike: The spk-package is just a template on building the hassio docker environment. So we should not bother so much with this.

spk/hassio/Makefile
fredrike: This is to create a "reproducablee build" ie tracking which build that is running.

I'll change it to 2019-224-1 to make every one happy. 🎅🏿

fxstein:
I have now installed and tested this package extensively on 2 different systems: RS818RP+ as well as a RS4017xs+. So far running without issues. This is the most comprehensible installation package of the homeassistant eco system. hassio is the path forward for the future of that project. 

I am also running the package through DSM Application Portal reverse proxy and and a custom domain/certificate from Let's Encrypt all managed and setup through DSM.

What needs to happen to get this package/pr merged and published? 

Happy to help in any way possible. Thank you!

spk/hassio/src/service-setup.sh
fxstein: Agreed the spk-package version should be an independent running number from supervisor or homeassistant as it pulls the latest versions of those as of install. Might be a good idea to give the user an idea what the latest versions of those where at time of spk release, but I'd consider that a documentation thing.

spk/hassio/Makefile
fxstein: Agreed. It would not make sense to install `homeassitant` and `hassio` as separate packages on the same machine. `hassio` supercedes a standalone `homeassistant` install. 
By now `homeassistant` itsef is just a fraction of the ecosystem that `hassio` manages for it. I would highly recommend against messing with its default port - the entire ecosystem defaults to it.
Given that the majority of the `homeassistant` community is migrating to `hassio` where possible, I'd even say that once the `hassio` package is available for synology, it would become the logical first choice of deployment.

spk/hassio/Makefile
fxstein: @fredrike Why not simply go with 

```
SPK_VERS = 1.0.0
SPK_REV = 1
```

and increment with every new build and PR from here on? I agree this should not be date dependent and if there is a numbering schema in place it should follow that. Seems like a very minor fix that could be applied and unblock this entire PR.

spk/hassio/Makefile
fredrike: I'll ignore this..

G3rry71:
What needs to happen to get this package/pr merged and published?

Happy to help in any way possible. Thank you!

spk/hassio/Makefile
```diff
@@ -0,0 +1,44 @@
+SPK_NAME = hassio
+SPK_VERS = 20191224
+SPK_REV = 1
```
chickenandpork: Testing, confirmed on `d67e38b` the trailing space on this line breaks build; instead of:
```
(cd /spksrc/spk/hassio/work-avoton-6.1 && tar cpf /spksrc/spk/hassio/../../packages/hassio_avoton-6.1_20191224-1.spk --group=root --owner=root package.tgz INFO scripts conf PACKAGE_ICON.PNG PACKAGE_ICON_256.PNG WIZARD_UIFILES)
```
... you have:
```
(cd /spksrc/spk/hassio/work && tar cpf /spksrc/spk/hassio/../../packages/hassio_noarch-all_20191224-1 --group=root --owner=root package.tgz INFO scripts conf PACKAGE_ICON.PNG PACKAGE_ICON_256.PNG WIZARD_UIFILES)
(cd /spksrc/spk/hassio/work && tar cpf .spk --group=root --owner=root package.tgz INFO scripts conf PACKAGE_ICON.PNG PACKAGE_ICON_256.PNG WIZARD_UIFILES)
```

Need to fix this

chickenandpork:
> What needs to happen to get this package/pr merged and published?

@ymartin59 has concerns
I raised one stopper

spk/hassio/Makefile
fxstein: Excuse my ignorance: When you say _markup_ - What do you mean? Put a comment in there or are you referring to extra code or settings being added somewhere? Is there a `conflicts` section that can be defined? 
Thank you!

fredrike: @chickenandpork how do we *markup* this?

spk/hassio/Makefile
fredrike: Perhaps the build process should handle spaces..

```suggestion
SPK_REV = 1
```

spk/hassio/src/service-setup.sh
fredrike: @chickenandpork no it is not satisfied, I'm not sure how to get that information from Docker..

spk/hassio/src/service-setup.sh
fredrike: No..

spk/hassio/src/service-setup.sh
```diff
@@ -0,0 +1,93 @@
...
+            sed -i -e "s|@homeassistant@|${HOMEASSISTANT_DOCKER}|g" ${CFG_FILE}
+            sed -i -e "s|@data_dir@|${DATA_DIR}|g" ${CFG_FILE}
+        fi
+    fi
+}
+
+service_preuninst() {
+    if [ ${SYNOPKG_PKG_STATUS} == "UNINSTALL" ]; then
+        HOMEASSISTANT="$(jq --raw-output '.homeassistant' ${CFG_FILE})"
+        SUPERVISOR="$(jq --raw-output '.supervisor' ${CFG_FILE})"
```
fredrike:
```suggestion
        SUPERVISOR="$(jq --raw-output '.supervisor' ${CFG_FILE})"
        HASSIO_DNS="$(docker inspect --format='{{.Image}}' HASSIO_DNS)"
```

spk/hassio/src/service-setup.sh
chickenandpork: seems good :)

spk/hassio/Makefile
chickenandpork: Sorry @fredrike if I used *markup* in an unusual way; I do mean flag, attribute, or configure.

@fxstein the `INFO` field would be `install_conflict_packages` based on setting a value for `SPK_CONFLICT` in the `spk/hassio/Makefile`

https://github.com/SynoCommunity/spksrc/blob/5a67e156fd8a6378d1a8e36366cf68adcaa0521f/mk/spksrc.spk.mk#L204

Seems the relevant Synology guide is over here now: https://originhelp.synology.com/developer-guide/synology_package/INFO_optional_fields.html (See 3.2.2)

Any time you feel this process is long -- and indeed, I've felt that with spksrc  -- consider that I suggested marking up the conflict a month ago, and wasn't answered.  I would blame GitHub, it's not the best conversation UI, so we all need to keep that in mind.  As well, builds are rate-limited and take a long time, so I can understand pushing back on possible reasons to rebuild by linting every PR to its limits.

fxstein: @chickenandpork Thank you very much! We are all learning how to do this right. This is very helpful and should be an easy fix.

So basically just add something like
```
SPK_CONFLICT = homeassistant
```
to the `hassio` makefile?

and vice versa  add
```
SPK_CONFLICT = hassio
```
to the `homeassistant` makefile?

Happy to submit a PR for `homeassistant` if that is required for this change.

PS: Not complaining about the length of the process, just want to help.

chickenandpork: I believe the change into the hassio would be sufficient, but marking in both directions seems pedantically correct ;)  This seems a more complete solution than my PR to @fredrike's branch (except the space, the one powerful extraneous space in the `SPK_VER`)

fxstein: I would recommend keeping it a separate PR but reference this PR in it. Keeps clean boundaries around the work but links the two together through GitHub.

If we agree on this, I'd be happy to make that change, test and open a separate PR for home assistant referencing this PR.

fredrike: > Any time you feel this process is long -- and indeed, I've felt that with spksrc -- consider that I suggested marking up the conflict a month ago, and wasn't answered. I would blame GitHub, it's not the best conversation UI, so we all need to keep that in mind. As well, builds are rate-limited and take a long time, so I can understand pushing back on possible reasons to rebuild by linting every PR to its limits.

I've actually missed the latest comments. It's indeed hard to keep track of comments that just comes occasionally. 

@chickenandpork what do you think think of my suggestions for uninstall? Wrote it on my phone while commuting and have limited access to a real computer in near future (only have a windows computer at work).

fredrike:
```suggestion
SERVICE_PORT = 8123
SPK_CONFLICT = homeassistant
```

fredrike:
> > What needs to happen to get this package/pr merged and published?
> 
> @ymartin59 has concerns
> I raised one stopper

I think everything is addressed now :)

fxstein:
@fredrike @chickenandpork Thank you both!

bruvv:
After 6 months are we finally going to see this merged :excited:

G3rry71:
> After 6 months are we finally going to see this merged 😳
> Thanks everyone involved especially @fredrike for your work! Can't wait to see this pop up in my app store

Congratulation team!

jokerigno:
Thank you all!

fredrike:
> Thank you all!

It is still not merged so I'm not super optimistic..

fxstein:
@chickenandpork @ymartin59 We all would very much appreciate your attention on getting this PR close out. Thank you very much!

chickenandpork:
`7abccb39` eliminates the concerns I raised:

I think the existing work still suffers from lacking the restart that happens on a hassio on a RaspPi: when HA stops in a Pi, IIRC the Pi restarts.  This means it doesn't *need* to get right the shutdown of addons and such.

Stop and Start are cleaner than I saw before; this is awesome.
* Stop still leaves `homeassistant/amd64-hassio-dns:1` running, plus my selected add-ons, but I'm OK accepting a later fix for that as a bugfix.  Removal stops `homeassistant/amd64-hassio-dns:1` but not my selected add-ons

"Updating" 20191224-1 to 20191224-1 seems to make no change.  Delete and install looks OK if I manually stop and delete my add-ons.  Unlike previous runs, this is good in the mode of "just reinstall and it'll do the right thing".  More people using this will clarify what circumstances images might still be running or any other issues.

I accept but I can't speak for @ymartin59 so I can't hit the "accept".

Be aware that accepted PRs become mergeable but there's a lag between that and building a release (based on space and cycles on a CI I think) but this one is a quick build so not as bad for this SPK.

fredrike:
> `7abccb39` eliminates the concerns I raised:
> 
> I think the existing work still suffers from lacking the restart that happens on a hassio on a RaspPi: when HA stops in a Pi, IIRC the Pi restarts. This means it doesn't _need_ to get right the shutdown of addons and such. I want to clarify that I don't see this as a failing: there's a previous comment about being similar to upstream, but upstream has some luxuries :) I really appreciate @fredrike 's work here.
> 
> Stop and Start are cleaner than I saw before; this is awesome.
> 
> * Stop still leaves `homeassistant/amd64-hassio-dns:1` running, plus my selected add-ons, but I'm OK accepting a later fix for that as a bugfix.  Removal stops `homeassistant/amd64-hassio-dns:1` but not my selected add-ons
> 

Uninstall of the package should remove the dns container. It would be fairly easy to make sure that all containers are stopped when stopping the package. Together with removing all related images. 

> "Updating" 20191224-1 to 20191224-1 seems to make no change. Delete and install looks OK if I manually stop and delete my add-ons. Unlike previous runs, this is good in the mode of "just reinstall and it'll do the right thing". More people using this will clarify what circumstances images might still be running or any other issues.
> 
> I accept but I can't speak for @ymartin59
> 
> Be aware that accepted PRs become mergeable but there's a lag between that and building a release (based on space and cycles on a CI I think) but this one is a quick build so not as bad for this SPK.


@fxstein, do you have an idea of how to detect the missing `entrypoint.js`  that happens once in a while (the fix is to restart the package)?

chickenandpork:
@ymartin59 I think this resolved the four issues you raised 2019-11-30:
* SPK_VER: @fredrike went with a static YYYYMMDD manually changed as a version
* SPK_REV: @fredrike change to 1
* ADMIN_PORT: @fredrike removed
* SERVICE_PORT 8123 collides with homeassistant::SERVICE_PORT: @fredrike marked the incompatibility in the SPK Makefile.

Looking for your OK with this work, and that any remaining issues can be pushed to a bug fix, so that @fredrike can get this merged.

spk/hassio/src/service-setup.sh
```diff
@@ -0,0 +1,97 @@
...
+
+        docker rm --force hassio_supervisor hassio_dns
+        if [ `docker inspect --format='{{.Config.Image}}' homeassistant|grep -q homeassistant/qemux86-64` ]; then
+            docker rm --force homeassistant && docker image rm ${HOMEASSISTANT}
+        fi 
+        docker image rm ${SUPERVISOR} ${HASSIO_DNS}
+        docker network rm hassio
+
+        if [ "${wizard_remove_addons}" == "true" ]; then
+            # Remove addons
```
fredrike: @chickenandpork, so you are saying that this doesn't work?

spk/hassio/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+DEPENDS =
+
+MAINTAINER = fredrike+hassio@gmail.com
+DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder."
+RELOAD_UI = yes
+STARTABLE = yes
+DISPLAY_NAME = Hass.io
+CHANGELOG = ""
+
```
ymartin59:
```suggestion
BETA = 1
```

fredrike:
> OK I publish for x64 architectures as beta for a first revision.

Great, thanks! Docker is only supported on x64. May i ask why beta and why you are so sceptical on this?

mp68:
Due to popular demand: When can we expect this to become available through the package manager?

spk/hassio/src/service-setup.sh
```diff
@@ -0,0 +1,97 @@
...
+    cd ${SYNOPKG_PKGDEST};
+    ${COMMAND} >> ${LOG_FILE} 2>&1 &
+    echo "$!" > "${PID_FILE}"
+
+    docker start homeassistant
+}
+
+service_poststop ()
+{
+    docker stop hassio_supervisor homeassistant
```
fredrike:
```
ADDONS=$(jq -r '.user| keys| map("addon_"+.)| join(" ")' $(jq -r '.data + "/addons.json"' ${CFG_FILE}))
docker stop $ADDONS
```

ymartin59:
At the moment, publish fails. A fix in spkrepo or in package is required: https://github.com/SynoCommunity/spkrepo/issues/39. Investigations in progress.

ymartin59:
@fredrike May you please register here with your github username: https://synocommunity.com/register so that I declare you as package owner.

dvisser:
is home assistant development still active? I see the package as beta and cant seem to find much development/progression? I could be looking in the wrong places though. Thanks.

fredrike:
> is home assistant development still active?

Well, I don't have any bugs to solve..

> I see the package as beta and cant seem to find much development/progression?

It is @ymartin59 that requested it to be BETA, https://github.com/SynoCommunity/spksrc/pull/3732#pullrequestreview-345028570. I would be happy to remove the BETA flag but have not been following the https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/125559 thread closely but it seems like @BeardedConti manages to get most configurations to run flawlessly.

bruvv:
> > is home assistant development still active?
> 
> Well, I don't have any bugs to solve..
> 
> > I see the package as beta and cant seem to find much development/progression?
> 
> It is @ymartin59 that requested it to be BETA, https://github.com/SynoCommunity/spksrc/pull/3732#pullrequestreview-345028570. I would be happy to remove the BETA flag but have not been following the https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/125559 thread closely but it seems like @BeardedConti manages to get most configurations to run flawlessly.
> 

Well when removing the package it doesn't remove all the created containers. Like hassio_dns and hassio_cli and some more I cannot remember atm

fredrike:
I've updated the package, but I don't see the BETA flag..

BeardedTinker:
I've checked few time and also don't see BETA flag.

fredrike:
But it is there in the community repository:

Screenshot 2020-09-11 at 8 00 45

ymartin59:
Your Package Center has not yet updated content: https://synocommunity.com/package/hassio

bruvv:
> 
> 
> Your Package Center has not yet updated content: https://synocommunity.com/package/hassio

I readded the community to my package manager but for me it is still showing up as beta as well ;)

PR #3733, opened 2019-07-04
Add ser2net and integrate to synocli-net

fredrike:
_Motivation:_  Adds ser2net.

### Checklist


spk/ser2net/PLIST
```diff
@@ -0,0 +1,29 @@
+bin:sbin/ser2net
```
hgy59: this whole file can be removed, since it is not used.
PLIST files are mostly used for cross builds only. They define the target files of a cross package to be copied into the staging folder when a spk target is built.

cross/ser2net/Makefile
```diff
@@ -0,0 +1,28 @@
+PKG_NAME = ser2net
+PKG_VERS = 4.0
+PKG_EXT = tar.gz
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+PKG_DIST_SITE =  http://downloads.sourceforge.net/project/$(PKG_NAME)/$(PKG_NAME)
+PKG_DIST_NAME = $(PKG_DIR).$(PKG_EXT)
+
+DEPENDS = cross/gensio cross/libyaml
+
+HOMEPAGE = https://sourceforge.net/projects/ser2net/
```
hgy59: this is not by ser2net
use https://github.com/cminyard/gensio instead

cross/ser2net/Makefile
```diff
@@ -0,0 +1,28 @@
...
+COMMENT  = Serial port to network proxy
+LICENSE  = GPLv2
+
+GNU_CONFIGURE = 1
+#PRE_CONFIGURE_TARGET = myPreConfigure
```
hgy59: please make a final version and remove commented code.

cross/gensio/Makefile
```diff
@@ -0,0 +1,29 @@
+PKG_NAME = gensio
+PKG_VERS = 1.0
+PKG_EXT = tar.gz
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+PKG_DIST_SITE = https://github.com/cminyard/$(PKG_NAME)/archive
+PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
+
+HOMEPAGE = https://sourceforge.net/projects/ser2net/
```
hgy59: this is not by ser2net
use https://github.com/cminyard/gensio instead

cross/gensio/Makefile
```diff
@@ -0,0 +1,29 @@
...
+COMMENT  = gensio - General Stream I/O
+LICENSE  = GPLv2
+
+DEPENDS = cross/openssl
+
+GNU_CONFIGURE = 1
+PRE_CONFIGURE_TARGET = myPreConfigure
```
hgy59: please name this label `gensio_pre_configure`
(and remove commented instructions)

cross/ser2net/Makefile
```diff
@@ -0,0 +1,28 @@
+PKG_NAME = ser2net
+PKG_VERS = 4.0
+PKG_EXT = tar.gz
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+PKG_DIST_SITE =  http://downloads.sourceforge.net/project/$(PKG_NAME)/$(PKG_NAME)
+PKG_DIST_NAME = $(PKG_DIR).$(PKG_EXT)
+
```
hgy59: for the sake of the order we prefere to define first the PKG_DIST_NAME than the PKG_DIST_SITE and both before PKG_DIR.
this will result in:
```
PKG_DIST_SITE = http://downloads.sourceforge.net/project/$(PKG_NAME)/$(PKG_NAME)
```

same for cross/gensio/Makefile

hgy59:
@fredrike did you succeed to build for archs ppc853x and qoriq?
when building for ppc853x-5.2 and qoriq-6.1 I get the following compile error for gensio:
```
make[4]: Entering directory '/spksrc/cross/gensio/work-ppc853x-5.2/gensio-1.0/lib'
  CC       gensio.lo
  CC       gensio_osops.lo
  CC       gensio_tcp.lo
  CC       gensio_udp.lo
  CC       gensio_stdio.lo
  CC       sergensio.lo
  CC       sergensio_telnet.lo
  CC       sergensio_serialdev.lo
sergensio_serialdev.c: In function 'process_rs485':
sergensio_serialdev.c:1447: error: 'struct serial_rs485' has no member named 'delay_rts_after_send'
sergensio_serialdev.c:1457: error: 'SER_RS485_RX_DURING_TX' undeclared (first use in this function)
sergensio_serialdev.c:1457: error: (Each undeclared identifier is reported only once
sergensio_serialdev.c:1457: error: for each function it appears in.)
Makefile:501: recipe for target 'sergensio_serialdev.lo' failed
```

hgy59:
@fredrike with this code I succeeded to compile gensio for QorIQ:
```

ifeq ($(findstring $(ARCH), $(PPC_ARCHES)),$(ARCH))
# archs/toolchains without compatible RS485 support (linux serial.h)
CONFIGURE_ARGS += ac_cv_have_decl_TIOCSRS485=no
endif
```

but ser2net did not compile for other ppc archs due too old gcc in toolchain (ppc853x-5.2 is still supported arch - and I have one of this)

so we have some unsupported archs that we can declare by adding `UNSUPPORTED_ARCHS` in ser2net Makefile for spk:
```
# qoriq is the only PPC arch that is supported
UNSUPPORTED_ARCHS = powerpc ppc824x ppc853x ppc854x
```

Maybe I will try an older version of ser2net to get running on my DS-210+ since the yaml config introduced in version 4 is the source of gcc incompatibility.

fredrike:
Thank you @hgy59 for your suggestions, I'll try to push a new version this weekend.

hgy59:
Ich hätte dazu noch ein Icon
(weil ich keines gefunden habe, musste ich selber eines zeichnen ;-)
ser2net

hgy59:
personally I would prefere a former version of ser2net.
Just the latest before 4.0 - Version 3.5.1 is much more crossplatform friendly. It does not depend on cross/yaml and cross/gensio and I succeeded to compile it for my oldest synology products like USB Station 2 (88f6180-3.2) and DS-107e (ppc824x-3.1) and all supported from ppc853x-5.2 to x64-6.1.
Just tested the build and not the function of ser2net. 
Is it worth to take the 4.x Versions when available for the platform?
If so, maybe we could first integrate and release version 3.5.1 and then an update to version 4.0 without support for ppx853x.

(I am not a friend of arch dependent versions as we find in the lirc spk, but this could be another solution)

chickenandpork:
@hgy59 ‘socat’ might still work for you.  I use it to proxy tcp but it does serial ports

fredrike:
> personally I would prefere a former version of ser2net.
> Just the latest before 4.0 - Version 3.5.1 is much more crossplatform friendly. It does not depend on cross/yaml and cross/gensio and I succeeded to compile it for my oldest synology products like USB Station 2 (88f6180-3.2) and DS-107e (ppc824x-3.1) and all supported from ppc853x-5.2 to x64-6.1.
> Just tested the build and not the function of ser2net.
> Is it worth to take the 4.x Versions when available for the platform?
> If so, maybe we could first integrate and release version 3.5.1 and then an update to version 4.0 without support for ppx853x.
> 
> (I am not a friend of arch dependent versions as we find in the lirc spk, but this could be another solution)

I don't have a preference here so if you like to change versions please do.

hgy59:
@ymartin59 shall we integrate into synocli-net?

hgy59:
@fredrike did you ever use ser2net on a Diskstation?
Can we use this without installation of the USB-Serial-Drivers package?

I just rebased this branch and it is now ready for merging into master.

Issue #3745, opened 2019-07-27
TVheadend build error - Libnettle 3.4 was not found


th0ma7:
### Setup
_Package Name:_ tvheadend
_Package Version:_ 4.2.8

_NAS Model:_ DS-918+
_NAS Architecture:_ apollolake
_DSM version:_ 6.2.2

### Expected behavior
Unable to rebuild package from source using Debian 10

### Actual behavior
Fails at the following:
```
  *** Libnettle 3.4 was not found.
```
### Steps to reproduce
_1._ Installed Debian 10 VM and followed well done documentation online at:
* https://github.com/SynoCommunity/spksrc
Needed small adjustment for dependencies:
```
$ sudo apt install build-essential debootstrap python-pip automake libgmp3-dev libltdl-dev libunistring-dev libffi-dev libcppunit-dev libncurses-dev imagemagick libssl-dev pkg-config zlib1g-dev gettext git curl subversion check intltool gperf flex bison xmlto php expect libgc-dev mercurial cython lzip cmake swig libc6-i386 libmount-dev libpcre3-dev libbz2-dev
```
_2._ Followed this doc, again well made:
* https://github.com/SynoCommunity/spksrc/wiki/Developers-HOW-TO
_3._ Invoked the build for tvheadend:
```
$ cd spksrc/spk/tvheadend/
$ make arch-apollolake
```

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
...
checking for threads.h... no
checking for getrandom... no
checking for getentropy... no
checking for NETTLE... no
configure: error: 

  *** Libnettle 3.4 was not found.

make[4]: *** [../../mk/spksrc.configure.mk:50: configure_target] Error 1
make[4]: Leaving directory '/home/th0ma7/spksrc/cross/gnutls'
make[3]: *** [../../mk/spksrc.depend.mk:44: depend_target] Error 2
make[3]: Leaving directory '/home/th0ma7/spksrc/cross/ffmpeg'
make[2]: *** [../../mk/spksrc.depend.mk:44: depend_target] Error 2
make[2]: Leaving directory '/home/th0ma7/spksrc/cross/tvheadend'
make[1]: *** [../../mk/spksrc.depend.mk:44: depend_target] Error 2
make[1]: Leaving directory '/home/th0ma7/spksrc/spk/tvheadend'
make: [../../mk/spksrc.spk.mk:453: arch-apollolake] Error 2 (ignored)
```

[...]

PR #3749, opened 2019-08-01
Debian10 fix

th0ma7:
_Motivation:_  Fix to get build functional on Debian 10
_Linked issues:_  #3745 

### Checklist

### DIFF
--- a/cross/nettle/Makefile
+++ b/cross/nettle/Makefile
@@ -32,7 +32,7 @@ endif
 ifeq ($(findstring $(ARCH), $(x64_ARCHES)),$(ARCH))
 CONFIGURE_ARGS = --target=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu
 endif
-CONFIGURE_ARGS += --prefix=$(INSTALL_PREFIX) --enable-shared
+CONFIGURE_ARGS += --prefix=$(INSTALL_PREFIX) --enable-shared --libdir=$(INSTALL_PREFIX)/lib
 
 # Fix for evansport builds to find gmp libraries
 CONFIGURE_ARGS += --with-include-path=$(STAGING_INSTALL_PREFIX)/include --with-lib-path=$(STAGING_INSTALL_PREFIX)/lib
--- a/README.rst
+++ b/README.rst
@@ -21,12 +21,12 @@ Docker
 
 Virtual machine
 ^^^^^^^^^^^^^^^
-A virtual machine based on an 64-bit version of Debian stable OS is recommended. Non-x86 architectures are not supported.
+A virtual machine based on an 64-bit version of Debian 10 stable OS is recommended. Non-x86 architectures are not supported.
 
 * Install the requirements::
 
     sudo dpkg --add-architecture i386 && sudo apt-get update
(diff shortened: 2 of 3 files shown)


th0ma7:
Using the following I was able to full test pulls #3750 and #3751.

cross/nettle/Makefile
```diff
@@ -32,7 +32,7 @@ endif
 ifeq ($(findstring $(ARCH), $(x64_ARCHES)),$(ARCH))
 CONFIGURE_ARGS = --target=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu
 endif
-CONFIGURE_ARGS += --prefix=$(INSTALL_PREFIX) --enable-shared
+CONFIGURE_ARGS += --prefix=$(INSTALL_PREFIX) --enable-shared --libdir=$(INSTALL_PREFIX)/lib
```
ymartin59: As far as I have tested with `arch-x64-6.1` on my Debian system, this option is not required at all as it is default behaviour. May you please check again or report if it is for only specific architecture.

th0ma7: My finding was that using specific 64-bit architectures (e.g. `make arch-apollolake`) would put some of the libraries under `lib64` instead of `lib`.  Thus other packages depending of it could not find them afterwards.  I would guess it is specific to theses sub-architectures?  For sure, this was mandatory for apollolake.

ymartin59: OK. Strange. To be honest, synocommunity packages are built with generic `arch-x64-6.1` target which generates a single binary package for all intel 64 bits architectures.

th0ma7: I'll try to find some time and confirm which of the 64-bit sub-arch are affected other than apollolake.  A `ifeq` statement could be added on a sub-arch basis type-thing...

ymartin59: I can already tell you this option is not required for `arch-x64-6.1` (generic intel 64 bits), `arch-aarch64-6.1` (ARM v8 64 bits), `arch-evansport-6.1` and `arch-apollolake-6.1`; neither in Docker context, nor in my "native" up-to-date Debian testing 64 bits with multi arch support (32 bits enabled). I really do not understand what may have triggered such a "failure" in your context/system.

th0ma7: That's rather interesting as I'm not getting the same results...
So lets trace back:
1) default minimal install of debian + setup `sudo`, ssh and `locales`
2) installations of the requirements (as per patch 01)
3) `sudo pip install -U setuptools pip wheel httpie`
4) then `spkrc`
```
$ git clone https://github.com/SynoCommunity/spksrc.git
$ cd spksrc
$ make setup
Setting default toolchain version to DSM-6.1
$ cd cross/nettle
$ make arch-apollolake
$ make arch-avoton
$ make arch-denverton
$ make arch-grantley
$ make arch-rtd1296
$ make arch-x64
```
Now checking the end-results:
```
$ ls -1ad work-*
work-apollolake-6.1
work-avoton-6.1
work-denverton-6.1
work-grantley-6.1
work-rtd1296-6.1
work-x64-6.1
```
And who used `lib64`:
```
$ ls -1ad work-*/install/usr/local/nettle/lib64/libnettle.so
work-apollolake-6.1/install/usr/local/nettle/lib64/libnettle.so
work-avoton-6.1/install/usr/local/nettle/lib64/libnettle.so
work-denverton-6.1/install/usr/local/nettle/lib64/libnettle.so
work-grantley-6.1/install/usr/local/nettle/lib64/libnettle.so
work-x64-6.1/install/usr/local/nettle/lib64/libnettle.so
```
Only `rtd1296` stayed into `lib` without any offending usage of `lib64`.

th0ma7: Was someone else able to reproduce?
Is there anything else I can do to get merged upstream?

Issue #3661, opened 2019-03-24
[Question] Medusa ugprade  requires python3


alexey-pimenov:
The medusa team have switched over to python3 in recent update.
But since medusa is based on `sickbeard-custom` in is not possible to upgrade because base spk and service relies on python 2 only.

Question is should be this issue resolved by creating separate spk package for medusa only or it's better to generify `sickbeard-custom`?

BenjV:
Medusa is running on both python 2.7 and python 3.5.
So sickbeard-custom can be used for running it on python 2.7.
Changing sickbeard-custom will prohibit others sickbeard forks which cannot run under python 3.5 to be used anymore.

Ading a new package for running Medusa under python 3.5 would be the best idea I think.
That way everybody can use his or hers own preference.

Jamyz:
@BenjV i create the package and you can download [here](https://mega.nz/#!XxYC3Ybb!cdKtWua9x51afjJDeUt1cHKzvy0FH3vlPIGTogk_nSI)
Thanks @alexey-pimenov

alexey-pimenov:
Closing because question was answered

Issue #3710, opened 2019-05-22
Need to Point Installed Sickbeard Custom to Python 3 on Synology NAS for Pymedusa (Now Python 3 Compatible)


wreid23:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_  sickbeard-custom
_Package Version:_ 

_NAS Model:_  DS3615xs
_NAS Architecture:_  XPENOLOGY INTEL
_DSM version:_  DSM 6.1.7-15284 Update 3

### Expected behavior
Tell us what should happen:

I  Need Package to start and use Python 3 package that is currently installed on NAS.
Pymedusa now supports python3: https://github.com/pymedusa/Medusa/wiki/Switch-to-Python-3
Need to point pymedusa to python3 module in: /volume1/@appstore/python3/bin and confirm that it will start and look for python3 going forward instead of 2.7 not sure how to do that .

### Actual behavior
Tell us what happens instead: 
![medusa](https://user-images.githubusercontent.com/24215762/58146377-cd34cd00-7c23-11e9-97b2-ac92f5a37f06.PNG)
Python 2.7 is used as it is currently the default

### Steps to reproduce
_1._  Start Pymedusa branch from Sickbeard-Custom
_2._  When in Pymedusa it will show running 2.7
_3._  Not sure what to change to use installed python3 packages

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
```
Insert the package log here
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here
```

ymartin59:
Problem is that "sickbeard-custom" is a single package to run multiple sickbeard-forks and switching to Python 3 will break any other fork...

BenjV:
It should be a relative simple change to the package to make this possible.
Just add a question to the package if python 2 or python 3 must be used and if  statement in the start-stop-status commandfile.
And remove the dependancy to python 2 in the INFO file.

Jamyz:
> It should be a relative simple change to the package to make this possible.
> Just add a question to the package if python 2 or python 3 must be used and if statement in the start-stop-status commandfile.
> And remove the dependancy to python 2 in the INFO file.

@BenjV  This package is available ?

Jamyz:
@BenjV i create the package and you can download here
Thanks @alexey-pimenov

BenjV:
Not until somebody change the package as I suggested.
[...]

PR #3756, opened 2019-08-11
Separate package for medusa with python 3

alexey-pimenov:
_Motivation:_  Medusa started to support python 3 some time ago and works way faster
_Linked issues:_  #3661 

### Checklist

I'm using this package for already 4 months with no issues


MrSparklle:
When this PR will be merged to master?

spk/medusa/Makefile
```diff
@@ -0,0 +1,39 @@
...
+STARTABLE = yes
+DISPLAY_NAME = Medusa
+CHANGELOG = "Separate service for Medusa"
+
+HOMEPAGE   = http://pymedusa.com
+LICENSE    = GPL
+
+SERVICE_USER = auto
+SERVICE_SETUP = src/service-setup.sh
+SERVICE_PORT = 8081
```
hgy59: this port is already used by synocommunity package sickbeard.
see wiki page https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports

please choose a different port, adjust the installation/configuration and document the port on the wiki page.

OR: define SPK_CONFLICT = sickbeard when parallel installations are not possible due concurrent functions (not concurrent ports alone).

p0psicles:
@alexey-pimenov could you pls update the port? In the not so distant future medusa is going to drop py2 support, and allot of users are going to panic when not know what to do. There already is a medusa specific PKG circulating. But we would very much prefer a synocomunity supported solution.

alexey-pimenov:
@p0psicles @hgy59 Sorry for such delayed reply - haven't had spare time recently.
I can look at this PR again this weekend.
But from comments in #3710  it wasn't clear for should I make a separate package or try to make existing package installable on top of python3

So I need help to decide if:
1. We just change the port or add SPK_CONFLICT, conflict seems more reasonable IMO
2. Make a package `sickbeard-custom-python3` which is an exact copy of the existing package but with different dependency
3. Try to make `sickbeard-custom` to work with both python 2 and 3 - which honestly I don't know how to do exactly

In all options, I don't see a way for the users to seamlessly upgrade because backup/restore and cache cleanup would be required.

p0psicles:
Preferred option would be 3. But for that you would need an extra prompt. Maybe @benjv or @mrsparklle can help with that?

For migrating data, users would need to create and recover a backup.

FireBladeNooT:
Hi, I propose to update the topic as Pymedusa last release manage only Python 3.
Do you think it's possible to have an update? Whatever the solution is used?

Sincerely

fredhen:
Hi @FireBladeNooT,

Have a look at the [Synology wiki section for Medusa](https://github.com/pymedusa/Medusa/wiki/Synology). There is a new [preferred package](https://github.com/BenjV/SYNO-packages) from [@BenjV](https://github.com/BenjV) that can now be used along with Python 3 support.

In my case (when moving from this package) I performed a backup, uninstalled this package, installed the new package and restored my backup. Just note that it uses port 8897 and cannot be changed.

Hopefully this helps!

FireBladeNooT:
Thanks, it works fine

PR #3767, opened 2019-09-09
monitoring-plugins: Add initial version

Gagi2k:
These plugins are useful when your NAS is part of bigger corporate
network where you use a nagios based system to monitor the health
of your machines.

Tested them to work on my DS214play


cross/monitoring-plugins/PLIST
```diff
@@ -0,0 +1,51 @@
...
+bin:bin/check_swap
+bin:bin/check_tcp
+bin:bin/check_time
+bin:bin/check_udp
+bin:bin/check_ups
+bin:bin/check_users
+bin:bin/check_wave
+bin:bin/negate
+bin:bin/urlize
+bin:bin/utils.pm
```
ymartin59: Only ELF-binary files must be declared as `bin`. So .pm and .sh are only `rsc`.
Refer to https://github.com/SynoCommunity/spksrc/wiki/PLIST-files

ymartin59:
Please also rebase over latest `master`

Gagi2k:
done

ymartin59:
Package is OK but I have found commands which does not work like:
```
$ /usr/local/monitoring-plugins/bin/check_apt 
'/usr/bin/apt-get -o 'Debug::NoLocking=true' -s -qq upgrade' exited with non-zero status.
APT UNKNOWN: 0 packages available for upgrade (0 critical updates).  errors detected.|available_upgrades=0;;;0 critical_updates=0;;;0

$ /usr/local/monitoring-plugins/bin/check_ifoperstatus 
Can't locate Net/SNMP.pm in @INC (you may need to install the Net::SNMP module) (@INC contains: /volume1/@appstore/monitoring-plugins/bin /usr/local/lib/perl5/site_perl /usr/local/share/perl5/site_perl /usr/local/lib/perl5/vendor_perl /usr/local/share/perl5/vendor_perl /usr/local/lib/perl5/core_perl /usr/local/share/perl5/core_perl .) at /usr/local/monitoring-plugins/bin/check_ifoperstatus line 42.
BEGIN failed--compilation aborted at /usr/local/monitoring-plugins/bin/check_ifoperstatus line 42.

$ /usr/local/monitoring-plugins/bin/check_ifstatus 
Can't locate Net/SNMP.pm in @INC (you may need to install the Net::SNMP module) (@INC contains: /volume1/@appstore/monitoring-plugins/bin /usr/local/lib/perl5/site_perl /usr/local/share/perl5/site_perl /usr/local/lib/perl5/vendor_perl /usr/local/share/perl5/vendor_perl /usr/local/lib/perl5/core_perl /usr/local/share/perl5/core_perl .) at /usr/local/monitoring-plugins/bin/check_ifstatus line 40.
BEGIN failed--compilation aborted at /usr/local/monitoring-plugins/bin/check_ifstatus line 40.
```

So I would propose to review each executable and either remove commands that are not expected to work on DMS, either fix them by including requirements in package (like perl Net::SNMP module).

Gagi2k:
Right, i also checked all commands now and found a few others which don't make sense or can't run because of missing binaries.

Regarding the missing perl module. Is there an easy way to install the missing module by adding a package dependency ? Or do i need to install it by using cpan ?

ymartin59:
Installing the Perl package may provide a more recent version and required modules but it is probably overkill for such a small feature.
Adding module from CPAN may be the wiser options, and easy as far as there is no native code to compile. So including the module in package and referring location with PERL5LIB may be enough.

ymartin59:
@Gagi2k May you please progress in removal of non relevant checks. We may look for Perl dependencies together later.

Gagi2k:
Hi,

sorry the very late response, end of the year is always a bit more demanding and i had less time to work on this.

I removed the apt check, as well as some other checks which doesn't make sense. For some others which should work out of the box i updated the configure line to use the correct location of these tools.

The perl module dependency is still a problem, i have no clue, how to do this. Do you have an example where something similar is already done, so i can adapt my package to do the same/similar ?

ymartin59:
@Gagi2k May you please progress in removal of non relevant checks. We may look for Perl dependencies together later.

Gagi2k:
I already removed all checks which i think are not relevant or their dependencies can't be satisfied easily. Do you have a particular check in mind which should be removed ?

ymartin59:
@Gagi2k Thanks for contributions. Package is now published.

Issue #3482, opened 2018-10-09
[Package Request] rmlint


1132719438:
Name: rmlint
Description:Extremely fast tool to remove duplicates and other lint from your filesystem
Website: http://rmlint.rtfd.org
Software documentation:
```
Features:

Finds…
   …Duplicate Files and duplicate directories.
   …Nonstripped binaries (i.e. binaries with debug symbols)
   …Broken symbolic links.
   …Empty files and directories.
   …Files with broken user or/and group ID.
Differences to other duplicate finders:
   Extremely fast (no exaggeration, we promise!).
   Paranoia mode for those who do not trust hashsums.
   Many output formats.
   No interactivity.
   Search for files only newer than a certain mtime.
...
```

Source Code: https://github.com/sahib/rmlint
License: GPLv3

ymartin59:
Published. Available from Package Center or https://synocommunity.com/package/synocli-file

PR #3776, opened 2019-09-22
synocli-file: add detox and rmlint

ymartin59:
Update Docker image to Debian buster for scons

_Motivation:_  Get synocli-file ready to publish
_Linked issues:_  #3482 #3250

### Checklist


ymartin59:
@hgy59 nano fails to find ncursew6-config in `install`, looks like pkg-config files are not generated properly
Was probably working before because Debian host version was the same as cross-compiled version... now I have updated to Debian duster and identify "ncurses" is not properly detected as compiled in `install`.

hgy59:
> 
> 
> @hgy59 nano fails to find ncursew6-config in `install`, looks like pkg-config files are not generated properly
> Was probably working before because Debian host version was the same as cross-compiled version... now I have updated to Debian buster and identify "ncurses" is not properly detected as compiled in `install`.

I can reproduce this, nano build does find neither ncurses nor ncursesw when built with a spksrc-docker-image based on debian:10-slim.

Tried with libncurses-dev installed and both (libncurses-dev  and ncurses-dev) installed with no success.

hgy59:
It's a reported bug in ncurses
[Debian Bug report logs - #933053](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=933053)
Therefore the build of synocli-file will fail on debian 10. Please use an build environment with debian 9 until this is fixed.

hgy59:
While testing with builds on debian 10 host, the version of 'file' haunted me, so I will push another fix with native version for file.
Does nano build with your changes here?

ymartin59:
I have fixed ncurses and nano compilation and increase `file` version too.

hgy59:
Looks good to me

ymartin59:
@hgy59 OK "Approval" would have been far enough - to notice I have not checked "all-supported" compilation yet...

Issue #3253, opened 2018-03-25
Design: new package "SynoCli dev"


ymartin59:
Create a package for small command line utilities dedicated to development tasks.
Revival of #2261
Tools candidates:
- ldd (provided by [synogear](https://www.synology-wiki.de/index.php/Synogear:_weitere_Tools))
- nice
- gcc
- gdb [cross package](https://github.com/SynoCommunity/spksrc/blob/master/cross/gdb/)  (provided by [synogear](https://www.synology-wiki.de/index.php/Synogear:_weitere_Tools))
- bc [available with #4037)
- binutils
- strace (provided by [synogear](https://www.synology-wiki.de/index.php/Synogear:_weitere_Tools))

Any other ideas?
Wanted contribution: an generic icon for "SynoCli" package group, with "dev" text addition for category

ymartin59:
@hgy59 Do you have ever started this package?

hgy59:
@ymartin59 I have started in january but didn't even create a branch. Only `bc` is integrated (into diyspk).
While trying to add `gdb` I left this aside...

michealespinola:
`bc` would be very useful so we could use diagnostic scripts like `dnsperftest`, etc.

cmencore:
+1 for bsdmainutils.. I'm trying to use [transcrypt](https://github.com/elasticdog/transcrypt) for git encryption, which requires the column command

hgy59:
> +1 for bsdmainutils.. I'm trying to use [transcrypt](https://github.com/elasticdog/transcrypt) for git encryption, which requires the column command

@cmencore what about the `column` command from **util-linux**? Would this solve your issue too?

hgy59:
closing this, as we now have syncli-devel (and synocli-misc).
The missing tool is gcc, but this will need to be delivered as non generic packages due to the dependency of kernel.

michealespinola:
Thank you very much for your continued efforts, and for both of these latest toolset packages. This helps further reduce my reliance on Entware.

hgy59:
> > closing this, as we now have syncli-devel (and synocli-misc). The missing tool is gcc, but this will need to be delivered as non generic packages due to the dependency of kernel.
> 
> Is `gcc` available anywhere for Synology?
> 
> (Sorry if this is not the place)

gcc is not and will not be included in synocli-devel as it must be different for each dedicated architecture.
And it is still in (slow) progress in #4469.

th0ma7:
[...]

Issue #1090, opened 2014-06-17
[Package Request] Prosody


tuxlagaffe:
Hello all,

I wanted to know if someone has tested prosody (as XMPP server) on thieir Synology ?
Because as I've read it seems better to configure it than ejjaberd and logs file are more human read.

Thanks in advanced for your return.

Best Regards.
##

mwild1:
I'm working on this :)

Safihre:
Any news?

PR #3780, opened 2019-10-06
Added new package: Lua v5.3.5

eelcohn:
_Motivation:_  This pull request will add the [Lua v5.3.5 package](https://www.lua.org/) to the SynoCommunity and supersedes pull request #3211. This new one has fixed the comments by ymartin59 and has updated Lua to v5.3.5. Please see the comments in #3211 for more info.
_Linked issues:_  N/A

### Checklist

**Build rule all-supported completed successfully**
Build rule is successfully completed for DSM6.1 on 88f6281, alpine, armada370, armada375, armada38x, armadaxp, comcerto2k, denverton, dockerx64, evansport, hi3535, kvmx64, monaco, ppc853x, qoriq, rtd1296 and x64 architectures.

**Package upgrade completed successfully**
N/A; this is the first package release for Lua.

**New installation of package completed successfully**
Tested on armadaxp and x64.


hgy59:
Lua is a candidate to be added to synocli-dev #3253

ymartin59:
@hgy59 From my point of view, Lua cannot be considered as a development tool but as an execution platform for other applications (like python, mono, java)... even if there is no package proposal for Lua application yet.

hgy59:
@ymartin59 lua is required for prosody as requested in  #1090 (I recently tried, but then I thought it would need a native/lua)

eelcohn:
Because some programs use Lua as a scripting language for macro's etc. and they call the shared library directly for that.

spk/lua/src/service-setup.sh
```diff
@@ -0,0 +1,16 @@
+service_postinst ()
+{
+	echo "Creating startup scripts for Lua executables" >> ${INST_LOG}
+	for executable in ${SYNOPKG_PKGDEST}/bin/lua*; do
+		ln -s "$executable" /usr/local/bin/$(basename "$executable")
+	done
```
ymartin59: @hgy59 Notice: I find it interesting... such logic may move to framework to reduce code in package, up to a point "service-setup.sh" may be removed for command line tools

eelcohn: Thanks. I created it this way so that the service-setup.sh file doesn't need to be updated when a version update of the package is published.

hgy59: I am not a fan of * here, because you can not see the affected names.

For the framework solution I suppose we can use the PLIST to get the names of the target files in bin and sbin folders (and maybe lib folder?) to create a "service-setup.sh" file that creates and removes links. 
Maybe we call the generated file "cli-setup.sh" and activate it with `MAKE_CLI_SETUP = 1`.
If this is based on the PLIST we get the right version anyhow.

PR #3788, opened 2019-10-16
bazarr: 0.9.0.8

### LINKED ISSUE #3535: [Package Request] Bazarr
Name: Bazarr
Description: Bazarr is a companion application to Sonarr and Radarr. It manages and downloads subtitles based on your requirements. You define your preferences by TV show or movies and Bazarr takes care of everything for you.

Website: https://github.com/morpheus65535/bazarr
Source code: https://github.com/morpheus65535/bazarr
License: GNU GPL v3

smaarn:
_Motivation:_  Creating initial bazarr spk
_Linked issues:_  Fixes #3535

### Checklist


smaarn:
This is a revival of PR #3784 .

As a reminder of what was said:
1. @hgy59 recommends not to proceed with the adding of another spk module which would be dependent upon python 2 (given its EOL coming soon)
2. There is a stream ongoing toward python 3 support for bazarr (as indicated by @morpheus65535)

cross/libicu/patches/0001-Fixed-added-missing-rule-for-generating-OUTDIR.patch
```diff
@@ -0,0 +1,15 @@
...
+ $(ICUDATA_SOURCE_ARCHIVE): $(ICUDATA_ARCHIVE) $(OUTDIR)
+ 	$(INVOKE) $(TOOLBINDIR)/icupkg -t$(ICUDATA_CHAR) $(ICUDATA_ARCHIVE) $(ICUDATA_SOURCE_ARCHIVE)
++	
++$(OUTDIR):
++	mkdir -p $(OUTDIR)
++
+ endif
+ else
+ ifneq ($(ENABLE_STATIC),YES)
+
```
smaarn: This has been proposed in PR on ICU project directly and is required for cross compilation purposes https://github.com/unicode-org/icu/pull/887.

spk/bazarr/src/service-setup.sh
```diff
@@ -0,0 +1,67 @@
...
+
+service_postupgrade ()
+{
+    if [ -d "${BACKUP_DIR}" ]
+    then
+        echo "Restoring data directory..." >> "${INST_LOG}"
+        mv -v "${BACKUP_DIR}/data" "${DATA_DIRECTORY}" >> "${INST_LOG}"
+        set_unix_permissions "${DATA_DIRECTORY}"
+    fi
+}
```
ymartin59: I would propose to configure bazarr to use "var/data" instead of "supposed read-only" package installation location - then there is no specific support to add for backup/restore

spk/bazarr/Makefile
```diff
@@ -0,0 +1,57 @@
...
+
+SPK_ICON = src/bazarr.png
+
+REQUIRED_DSM = 6.1
+
+BUILD_DEPENDS = cross/python cross/setuptools cross/pip cross/wheel
+
+BETA = 1
+
+DEPENDS = cross/busybox
```
ymartin59: There is no need to depend on busybox as python package already provides it

Dockerfile
```diff
@@ -39,13 +39,16 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
         libssl-dev \
         libtool \
         libunistring-dev \
+        libxml2-dev \
+        libxslt-dev \
```
ymartin59: Probably this is a mistake. There is no need to have these libraries header files installed as build framework will cross-compile its own versions as declared as application's dependencies

spk/bazarr/Makefile
```diff
@@ -0,0 +1,57 @@
...
+DEPENDS += cross/libxml2 
+DEPENDS += cross/libxslt
```
ymartin59: I would replace these dependencies by `cross/lxml` and update it if needed - and I guess it is then possible to remove `WHEELS` variable.

cross/libicu/Makefile
```diff
@@ -0,0 +1,35 @@
...
+include ../../mk/spksrc.cross-cc.mk
+
+.PHONY: pre_configure_icu
+pre_configure_icu:
+	if [ ! -d "$(ICU_WORK_DIR)" ]; then \
+		mkdir -p $(ICU_WORK_DIR); \
+		cd $(ICU_WORK_DIR) \
+			&& RANLIB= LDFLAGS= CXXFLAGS= CXX= CPPFLAGS= CPP= CFLAGS= CC= AR= ARFLAGS= $(WORK_DIR)/$(PKG_DIR)/configure $(SHARED_CONFIGURE_ARGS) \
+			&& make; \
+	fi;
```
ymartin59: I just guess icu requires a native compilation first to be able to cross-compile in a second phase. If so, please create a `native/icu` module and refers to it in a similar way as `cross/protobuf` and `native/protobuf`.

spk/bazarr/Makefile
```diff
@@ -0,0 +1,57 @@
...
+include ../../cross/bazarr/bazarr.mk
+
+SPK_VERS = $(BAZARR_VERSION)
+SPK_REV = 2
+
+SPK_ICON = src/bazarr.png
+
+REQUIRED_DSM = 6.1
+
+BUILD_DEPENDS = cross/python cross/setuptools cross/pip cross/wheel
```
ymartin59: I am almost sure this variable is useless - please compare how `cross/sabnzbd` and `spk/sabnzbd` are set up to generate wheels without building whole python stack.

spk/bazarr/src/service-setup.sh
smaarn: The only drawback here would be that we need to make sure that upon "actual" uninstallation the directory is not being removed, right ?

Dockerfile
smaarn: Well, it seemed to be required in the wheel generation, but then, if I upgrade cross/lxml package to 4.3.0 it may fix the issue. I'll check this.

- [ ] Upgrade cross/lxml to 4.3.0 (current is 4.2.5)
- [ ] Prefer using cross/lxml instead of wheels

spk/bazarr/Makefile
smaarn: Is it considered a good practice, though ? (Open question really)
Because, the day python removes busybox from its installation, this package would no longer be operational, right ?

spk/bazarr/Makefile
smaarn: I only managed to remove cross/pip from the build process (required for lxml actually)

spk/bazarr/Makefile
smaarn: Indeed, and it removed the requirement for the headers files !

hgy59:
@smaarn yes, this is a good solution (I was wondering why there is a mix of lxml in wheel and dependency of cross/lxml)

cross/libicu/Makefile
smaarn: Thanks for the tip, I had struggled with it !

- [ ] Create native/icu module

smaarn: (I might as well try removing the requirement for icu altogether... It was actually required to build the wheels)

spk/bazarr/src/service-setup.sh
smaarn: - [ ] Use `/var/data/bazarr` directory and clear it manually upon uninstallation (at worse create UI for uninstallation)

spk/bazarr/Makefile
smaarn: - [ ] Depend on busybox installation provided by python

cross/libicu/patches/0001-Fixed-added-missing-rule-for-generating-OUTDIR.patch
smaarn: ICU dependency has been identified as no longer useful (since switching to cross/lxml usage, basically).

cross/libicu/Makefile
smaarn: OK, turns out that ICU dependency was no longer required.
At least I learnt how it could be done :)
Thanks anyway for the help since, indeed, the native option was way simplier here !

smaarn:
OK so basically I reworked everything with the following changes:
- Use cross/lxml (upgraded from 4.2.5 to 4.3.0)
- Dropped icu thing
- Use /usr/local/bazarr/var/data directory (natively handled in the upgrade process)
- Use busybox binary from python installation

Validated on evansport infrastructure, running all-supported target while we speak.

spk/bazarr/src/dsm-control.sh
```diff
@@ -0,0 +1,83 @@
...
+    start-stop-daemon -S -q -m -b -N 10 -p "${PID_FILE}" -x "${SH_PATH}"  -- \
+        -c "source \"${INSTALL_DIR}/bin/settings.sh\" \
+        && pushd \"${BAZARR_DIR}\" \
+        && mkdir -p \"${BAZARR_DIR}/var\" \
+        && python bazarr.py >>\"${BAZARR_DIR}/var/bazarr.log\" 2>&1"
+}
+
+stop_daemon ()
+{
+    curl http://127.0.0.1:6767/shutdown || echo "Failed invoking smooth shutdown procedure";
```
smaarn: OK, I give you that this is pretty ugly and won't work if the port is changed which is ... bad.

An alternative would be to go with PPID mechanisms, WDYT ?

smaarn: Fixed to use PPID process lookup.

smaarn:
Rebased at head and squashed all commits.

smaarn:
Checklist

- [ ]  Build rule all-supported completed successfully
- [ ]  Package upgrade completed successfully

smaarn:
Upgraded to 0.8.3.3 and rebased @ head.

morpheus65535:
FYI: Bazarr 0.8.4 is out and fully supporting Python 3.6 or greater.

smaarn:
> FYI: Bazarr 0.8.4 is out and fully supporting Python 3.6 or greater. We'll drop support for 2.7.x by the end of January.

Great, I'll rewrite this to use python 3 then !

Thanks a lot

spk/bazarr/Makefile
```diff
@@ -0,0 +1,41 @@
+SPK_NAME = bazarr
+
+include ../../cross/bazarr/bazarr.mk
```
hgy59: I would prefere to hardcode the version here (as we are used to) and not to include bazarr.mk containing the version.
It would be more obvious when the version is visible here - and on updates when you add the version to the changelog you can verify it matches within this Makefile.
We should define this consistently within the framework.
And may be we will find a framework solution to define all cross-package-versions in skp/Makefile. Until now we need two cross packages to support two different versions of the same package (see lirc and berkeleydb)

spk/bazarr/Makefile
```diff
@@ -0,0 +1,41 @@
...
+REQUIRED_DSM = 6.1
+
+BUILD_DEPENDS = cross/python cross/setuptools cross/wheel
+
+BETA = 1
+
+DEPENDS = cross/$(SPK_NAME)
+
+WHEELS = 
+SPK_DEPENDS = "python>=2.7.13:git"
```
hgy59: Are you shure bazarr requires git spk package installed on the target system?
The installation instructions start with "installation of GIT" but IMHO git is used only to clone the bazarr repository.

morpheus65535:
Git is used by the updater. You can also add command line arguments when starting Bazarr:
`—no-update` to disable updater
`—release-update` to only fetch new release to master branch. 

https://github.com/morpheus65535/bazarr/blob/master/bazarr/get_args.py

hgy59:
1. Does this kind of update really work when the original installation was not done by cloning the git repository?
2. As for now the git spk does not but git into the path. Therefore bazarr must use the full path to call git (/var/packages/git/target/bin/git)
3. In your start_daemon function in dsm-control.sh there is none of the updater arguments. What kind of update does this mean?

I would propose to disable (automatic?) updates and to remove git from SPK_DEPENDS.
For advanced users a descriptions how to enable updates would be helpfully (modify the service start and provide a symlink /usr/local/bin/git)
The ultimative solution would be to add an option in the install-wizard whether to enable the update feature - and to modify the service start so that the update parameter is only added when selected in the wizard and when git is installed

spk/bazarr/Makefile
smaarn: > 1. Does this kind of update really work when the original installation was not done by cloning the git repository?
> 2. As for now the git spk does not but git into the path. Therefore bazarr must use the full path to call git (/var/packages/git/target/bin/git)
> 3. In your start_daemon function in dsm-control.sh there is none of the updater arguments. What kind of update does this mean?
> 
> I would propose to disable (automatic?) updates and to remove git from SPK_DEPENDS.
> For advanced users a description how to enable updates would be helpful (modify the service start and provide a symlink /usr/local/bin/git)
> The ultimative solution would be to add an option in the install-wizard whether to enable the update feature - and to modify the service start so that the update parameter is only added when selected in the wizard and when git is installed
> 
> But I would prefer just do disable the integrated updates.

Technically git is added to the path at the startup of this script. But, then, it doesn't mean necessarily it will work.
@morpheus65535: I will enforce the disabling of the "autoupdate through git".

smaarn:
- [ ] Disable integrated updates mechanism and remove git dependency

smaarn:
- [ ] Perform python3 migration (depends on #3833 to be validated on my side ... as I only have an evansport ... )

smaarn:
### Checklist (in progress)

smaarn:
New manually built artifacts: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4-1

JeeDeWee:
> New manually built artifacts: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4-1

Hi @smaarn, thanks for your great work on generating this package. I am trying to get this package onto my Synology for a long time but stopped work because you already nailed it.

I installed your latest build artifact but the package does not start because of the following error:

```
Bazarr` starting...`
File "bazarr/main.py", line 36, in <module>
  from init import *
  File "/volume1/@appstore/bazarr/share/bazarr/bazarr/init.py", line 65, in <module>
    fd = open(os.path.join(os.path.dirname(__file__), 'create_db.sql'), 'r')
FileNotFoundError: [Errno 2] No such file or directory: '/volume1/@appstore/bazarr/share/bazarr/bazarr/create_db.sql'
```

I installed fresh, i.e. no update. I believe this is caused by the fact that your `cross/bazarr/PLIST` file strips the `create_db.sqp` file. This file is needed for this package to work.

I am a novice on using GitHub, otherwise I would have prepared a pull-request.

smaarn:
> > New manually built artifacts: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4-1
> 
> Hi @smaarn, thanks for your great work on generating this package. I am trying to get this package onto my Synology for a long time but stopped work because you already nailed it.
> 
> I installed your latest build artifact but the package does not start because of the following error:
> 
> ```
> Bazarr` starting...`
> File "bazarr/main.py", line 36, in <module>
>   from init import *
>   File "/volume1/@appstore/bazarr/share/bazarr/bazarr/init.py", line 65, in <module>
>     fd = open(os.path.join(os.path.dirname(__file__), 'create_db.sql'), 'r')
> FileNotFoundError: [Errno 2] No such file or directory: '/volume1/@appstore/bazarr/share/bazarr/bazarr/create_db.sql'
> ```
> 
> I installed fresh, i.e. no update. I believe this is caused by the fact that your `cross/bazarr/PLIST` file strips the `bazarr/create_db.sql` file. This file is needed for this package to work.
> 
> I am a novice on using GitHub, otherwise I would have prepared a pull-request.

Let me fix this right away, thanks for the feedback, it turns out that I haven't been installing this guy for some time... ;)

smaarn:
Newly manually generated artifacts can be found here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4-2

pkmetski:
Thanks for the great work @smaarn 
Trying to install the package on my Synology, I get an error that it requires Python 3.6 or later.
As far as I see, the latest Python version available on the Synology is 3.5.1. Am I missing something?

smaarn:
> Thanks for the great work @smaarn
> Trying to install the package on my Synology, I get an error that it requires Python 3.6 or later.
> As far as I see, the latest Python version available on the Synology is 3.5.1. Am I missing something?

Hello,

Indeed this PR is relying on another PR ( #3833 ). I also generated corresponding binaries over there.

smaarn:
Following latest release of bazarr I updated the PR to take it into account.

CXgamer:
> If you're "happy" with it don't hesitate mentioning it on corresponding PR SynoCommunity#3788

I am happy.

WodKa78:
> Following latest release of bazarr I updated the PR to take it into account.
> 
> Binaries can be found here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.1

Hi @smaarn any chance you can compile a package for rt1296 devices, or is this not possible? 
Thanks in advance....

hgy59:
@WodKa78 for rtd1296 you have to use the aarch64 package. aarch64 includes rtd1296 and armada37xx.
You find this in the Generic Arch column on the wiki page https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures
Until now there are generic packages for x64 (x86_64) and aarch64. 
In the future it might be some generic armv7 packages too.

WodKa78:
> @WodKa78 for rtd1296 you have to use the aarch64 package. aarch64 includes rtd1296 and armada37xx.
> You find this in the Generic Arch column on the wiki page https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures
> Until now there are generic packages for x64 (x86_64) and aarch64.
> In the future it might be some generic armv7 packages too.

Perfect and thanks for the quick reply!

jasoncodes:
I’ve just installed the aarch64 package on my NAS and when I try to view my movie list I got the following while trying to stat a directory containing UTF-8 characters:

```
UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 37: ordinal not in range(128)
```

I added `export LANG=en_US.UTF-8` after `export PATH=…` in `/var/packages/bazarr/scripts/start-stop-status`(`spk/bazarr/src/dsm-control.sh` in this PR) and now it works.

smaarn:
> I’ve just installed the aarch64 package on my NAS and when I try to view my movie list I got the following while trying to stat a directory containing UTF-8 characters:
> 
> ```
> UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 37: ordinal not in range(128)
> ```
> 
> I added `export LANG=en_US.UTF-8` after `export PATH=…` in `/var/packages/bazarr/scripts/start-stop-status`(`spk/bazarr/src/dsm-control.sh` in this PR) and now it works.

Hello,

Playing with hardcoded encodings is usually a bad idea since it depends on user installation.

Your hack may be interesting but I must confess I would rather try identifying how this is configured in the NAS installation. For example, on my side, this environment variable is actually already configured that way.

What happens if you run `sudo -u sc-bazarr env` on your NAS ? 

On my side it gives this:
```
TERM=xterm-256color
LC_ALL=en_US.utf8
PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/syno/sbin:/usr/syno/bin:/usr/local/sbin:/usr/local/bin
LANG=en_US.utf8
MAIL=/var/mail/sc-bazarr
LOGNAME=sc-bazarr
USER=sc-bazarr
USERNAME=sc-bazarr
HOME=/var/packages/bazarr/target
SHELL=/sbin/nologin
SUDO_COMMAND=/bin/env
SUDO_USER=root
SUDO_UID=0
SUDO_GID=0
```

smaarn:
Just had a look and all of this seems to be configured in `/etc/profile` on my NAS. 
Could you check on your NAS what's the value ?

smaarn:
@jasoncodes it seems that there are other modules already using this kind of tricks (even though they seem to be more often using the `LC_ALL` environment variable.
I'll try to generate new binaries using that hack even though I'm not convinced with that path.

jasoncodes:
```
# grep -E 'LC_|LANG' /etc/profile
export LC_ALL=en_US.utf8
export LANG=en_US.utf8
# sudo -u sc-bazarr env | grep -E 'LC_|LANG'
LC_ALL=en_US.utf8
LANG=en_US.utf8
```

I don’t know for sure but I suspect services do not load `/etc/profile` into their environment. No other packages I have installed appear to hardcode these.

FWIW, Plex sets both `LANG` and `LC_ALL` to `en_US.utf8` directly in their startup script.

smaarn:
> ```
> # grep -E 'LC_|LANG' /etc/profile
> export LC_ALL=en_US.utf8
> export LANG=en_US.utf8
> # sudo -u sc-bazarr env | grep -E 'LC_|LANG'
> LC_ALL=en_US.utf8
> LANG=en_US.utf8
> ```
> 
> 
> I don’t know for sure but I suspect services do not load `/etc/profile` into their environment. No other packages I have installed appear to hardcode these.
> 
> 
> FWIW, Plex sets both `LANG` and `LC_ALL` to `en_US.utf8` directly in their startup script.

Just checked how it works and it seems that environment variables are actually provided by the invoking scripts.

By adding the below lines
```
locale >"${VAR_DIR}/bazarr-locale" 2>&1
```
In the script you mentioned earlier I got:
```
LANG=
LC_CTYPE="POSIX"
LC_NUMERIC="POSIX"
LC_TIME="POSIX"
LC_COLLATE="POSIX"
LC_MONETARY="POSIX"
LC_MESSAGES="POSIX"
LC_PAPER="POSIX"
LC_NAME="POSIX"
LC_ADDRESS="POSIX"
LC_TELEPHONE="POSIX"
LC_MEASUREMENT="POSIX"
LC_IDENTIFICATION="POSIX"
LC_ALL=
```

So, I'm guessing that it would be more efficient to have some kind of profiling system here to allow users to customize the environment variables they wish to use. I'm going, therefore, to create the support for a `profile` script which would allow users to customize their encoding if they don't want to use the pseudo-universal `en_US.utf8` one.
How does that sound to you ?

smaarn:
Squashed all commits.
Generating binaries while writing this. Should be ready by tomorrow morning CEST.

spk/bazarr/src/dsm-control.sh
```diff
@@ -0,0 +1,91 @@
...
+DNAME="Bazarr"
+
+# Others
+INSTALL_DIR="/usr/local/${PACKAGE}"
+PYTHON_DIR="/usr/local/python3"
+BAZARR_DIR="${INSTALL_DIR}/share/${PACKAGE}"
+VAR_DIR="${INSTALL_DIR}/var";
+
+export PATH="${INSTALL_DIR}/bin:${PYTHON_DIR}/bin:${PATH}"
+export LC_ALL=en_US.utf8
```
smaarn: This was made to enforce default locale settings as it seems to not really be handled in Synology mechanism to bootstrap package services. Unless someone knows how to read it... Personally I would rather go for the /etc/profile but then... It would introduce a heavy dependency here I believe so I went for the "shortest": default settings which can be overridden (along with other things) by defining a profile script under the var directory.

jasoncodes: I like the solution you came up with here. Default to `en_US.UTF-8` but one can easily override and/or provide additional environment variables via the optional `profile` file. 👍

smaarn:
New binaries can be found here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.1-2

smaarn:
Latest release (0.8.4.2) is now done with corresponding binaries here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.2

smaarn:
Checklist

 

- [ ]  Build rule all-supported completed successfully
- [ ]  Package upgrade completed successfully
- [ ]  New installation of package completed successfully

smaarn:
New release here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.2-2

spk/bazarr/Makefile
```diff
@@ -0,0 +1,41 @@
+SPK_NAME = bazarr
+
+BAZARR_VERSION = 0.8.4.2
+
+SPK_VERS = $(BAZARR_VERSION)
+SPK_REV = 2
+
+SPK_ICON = src/bazarr.png
+
+REQUIRED_DSM = 6.1
```
hgy59: what is the reason to define a minimal DSM Version?

cross/bazarr/PLIST
```diff
@@ -0,0 +1,10 @@
+rsc:share/bazarr/LICENSE
+rsc:share/bazarr/___init__.py
+bin:share/bazarr/bazarr.py
+rsc:share/bazarr/bazarr/*
+bin:share/bazarr/bin/Linux/**
+rsc:share/bazarr/libs/**
+rsc:share/bazarr/libs3/**
+rsc:share/bazarr/requirements.txt
+rsc:share/bazarr/static/**
+rsc:share/bazarr/views/**
```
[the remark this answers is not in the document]
smaarn: Ok for the single line, the size impact should ne negligible. There would be windows and macOsX binaries taken in that case, though (they are stored under (`bin`)[https://github.com/morpheus65535/bazarr/tree/master/bin]).
Regarding the equivalency between `**` and `*` I'm pretty sure it wasn't the case but I'll retry to check.

- [ ] Simplify into one line
- [ ] Replace `**` by `*` if it's equivalent
- [ ] Check if `bin`

cross/bazarr/Makefile
```diff
@@ -0,0 +1,26 @@
+PKG_NAME = bazarr
+
+BAZARR_VERSION = 0.8.4.2
+
+PKG_VERS = v$(BAZARR_VERSION)
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_VERS).$(PKG_EXT)
+PKG_DIST_FILE = $(PKG_NAME)-$(PKG_DIST_NAME)
+PKG_DIST_SITE = https://github.com/morpheus65535/bazarr/archive
+PKG_DIR = $(PKG_NAME)-$(BAZARR_VERSION)
```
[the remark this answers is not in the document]
smaarn: I'm not too sure about the choice I have regarding the `PKG_DIST_FILE` but I'll have a look.

spk/bazarr/Makefile
smaarn: None apart from the fact that since I was only testing it on that version.
I could put any version to be honest but then the packaging mechanism and the user logics would have to be checked, right ?
All the other packages I saw which were supporting older versions (than the 6, that is) seemed to have different logics to implement.

cross/bazarr/PLIST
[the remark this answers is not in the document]
smaarn: Just checked, if we simplify into one line we would be taking the following additional resources

- https://github.com/morpheus65535/bazarr/blob/master/bin/MacOSX/i386/UnRAR/unrar (319 KB)
- https://github.com/morpheus65535/bazarr/blob/master/bin/Windows/i386/UnRAR/UnRAR.exe (363 KB)
- https://github.com/morpheus65535/bazarr/blob/master/bin/Windows/i386/ffprobe/ffprobe.exe (43 MB)

So the size of the binaries would increase by more than 40MB. Would that be acceptable ?

I'll try checking about the strippable binary file status for the linux resources.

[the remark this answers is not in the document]
smaarn: Actually I may have a path through the usage of the tar command and exclusions.

cross/bazarr/Makefile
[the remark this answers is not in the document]
hgy59: As all source files are downloaded into the distrib folder, it is required to use PKG_DIST_NAME to rename the file if the filename (PKG_DIST_FILE) does not contain the package name (i.e. version only) to avoid conflicts with others.

cross/bazarr/PLIST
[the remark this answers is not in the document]
hgy59: we definitly should avoid to install Windows and MaxOS binaries...

spk/bazarr/Makefile
hgy59: IMHO we use REQUIRED_DSM when the spk explicitly depends on DSM features. For 6.1 (6.0 would be enought) this could be nginx, php configurations, minimal kernel, libc or TC compiler version etc.
As with `make all-supported` we officially support version 6.1 only (and arch-ppc853x-5.2) there is no need to use REQUIRED_DSM for other purpose.

cross/bazarr/Makefile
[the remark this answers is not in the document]
smaarn: Done.

cross/bazarr/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+DEPENDS = cross/lxml
+
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = nop
+INSTALL_TARGET = bazarrInstall
```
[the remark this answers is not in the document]
smaarn: Done.

cross/bazarr/PLIST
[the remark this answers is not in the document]
smaarn: Still some improvement which may be done I believe (we're shipping several versions of unrar & ffprobe in each package.

spk/bazarr/Makefile
```diff
@@ -0,0 +1,41 @@
+SPK_NAME = bazarr
+
+BAZARR_VERSION = 0.8.4.2
```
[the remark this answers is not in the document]
smaarn: Done.

spk/bazarr/Makefile
```diff
@@ -0,0 +1,41 @@
...
+STARTABLE = yes
+CHANGELOG = "Corrected shebang of python scripts."
+
+HOMEPAGE = https://www.bazarr.media
+LICENSE = GPLv3
+
+SERVICE_USER  = auto
+SERVICE_SETUP = src/service-setup.sh
+SSS_SCRIPT    = src/dsm-control.sh
+SERVICE_PORT  = 6767
```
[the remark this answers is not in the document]
smaarn: Done.

spk/bazarr/src/service-setup.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    ${INSTALL_DIR}/env/bin/pip install \
+        --no-deps --no-index -U --force-reinstall \
+        -f ${INSTALL_DIR}/share/wheelhouse ${INSTALL_DIR}/share/wheelhouse/*.whl >> "${INST_LOG}"
+
+    echo "Running busybox installation..."  >> "${INST_LOG}"
+
+    # Install busybox
+    ${PYTHON_DIR}/bin/busybox --install ${INSTALL_DIR}/bin
+}
```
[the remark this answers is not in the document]
smaarn: Done.

spk/bazarr/Makefile
smaarn: Done.

smaarn:
New binaries available here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.2-3

cross/bazarr/PLIST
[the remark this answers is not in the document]
smaarn: @hgy59 open question: I need unrar and ffprobe to be available in the PATH. While for the former there is no question about embedding it, regarding the latter (ffmpeg) I have three options:

1. Ship ffmpeg with bazarr (we're talking about a resulting package of 50 MB here)
2. Depend optionally on ffmpeg (e.g. auto-detect at runtime) (resulting package around 32 MB)
3. Explicitely depend on it

Knowing that, from what I see option 1. and 3. seem to be the existing options I've found (spk/chromaprint and spk/tvahead).

Is there any official preference about this ?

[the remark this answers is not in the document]
smaarn: I'm going to go, for now and unless instructed otherwise, for an "all in one" approach (e.g. ship ffmpeg with all its resources).

spk/bazarr/Makefile
```diff
@@ -0,0 +1,41 @@
...
+WHEELS = 
+SPK_DEPENDS = "python3>=3.6"
+
+MAINTAINER = SynoCommunity
+
+DESCRIPTION = "Bazarr is a companion application to Sonarr and Radarr. It manages and downloads subtitles based on your requirements. You define your preferences by TV show or movies and Bazarr takes care of everything for you."
+RELOAD_UI = no
+DISPLAY_NAME = Bazarr
+STARTABLE = yes
+CHANGELOG = "Corrected shebang of python scripts."
```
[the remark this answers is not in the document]
smaarn: Fair enough, I'll do this once all the rest is done or with next bazarr release.

Mickroz:
i would like to test this, but no package available for apollo lake.

smaarn:
> i would like to test this, but no package available for apollo lake.

@Mickroz I believe x64 packages should work no ? (I'm going to try an explicit one now but just in case it already works...)

smaarn:
By the way I took the liberty of publishing a new version which relies on cross compiled ffmpeg and unrar binaries: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.2-4
It takes a "little" bit more time to generated, but saves some space and ensures it runs appropriately on all "supported" archs.

Mickroz:
ah you're right, only need python 3.8 though, and sc version is on 3.5.6

ymartin59:
@smaarn I am OK that bazaar package embeds unrar, but "ffmpeg" is another beast (in term of package size and consumed bandwidth).
If bazaar invokes ffmpeg commands, you should depend on SynoCommunity ffmpeg package and include `/var/packages/ffmpeg/target/bin` in service PATH.

smaarn:
@ymartin59 following your recommendations I updated the spk to actually depend on ffmpeg package. Corresponding binaries can be found here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.2-5

edsenabr:
Hi, @smaarn ,

I've created a fully functional, pure python version of this package. I've tested it on 3 architectures, and it's working like a charm. How can I contribute this to the community ?

ymartin59:
@edsenabr I suppose you have used "spksrc" framework to build it. So create a dedicated branch for this contribution in your GitHub fork of "spksrc", create one commit with your changes and new files, push branch and trigger pull request creation so that it appears here: https://github.com/SynoCommunity/spksrc/pull

edsenabr:
Thanks @ymartin59 ; Since @smaarn was already working on it, I was thinking of adding to that initiave. On top of that, the latest build of bazarr depends now on python 3.7, instead of 3.6; I've build and installed mannually the python package from a fork/work he's also doing... 

Wanted a way of 'coordinating' all this effforts

smaarn:
> Thanks @ymartin59 ; Since @smaarn was already working on it, I was thinking of adding to that initiave. On top of that, the latest build of bazarr depends now on python 3.7, instead of 3.6; I've build and installed mannually the python package from a fork/work he's also doing...
> 
> Wanted a way of 'coordinating' all this effforts

@edsenabr thanks for joining !

Sorry if I didn't reply earlier as I thought @ymartin59 's input would actually be more meaningful than mine ;)

To be honest it really depends on how advanced you are since since I guess there are a couple (at least from what I see) of things which could need improving / checking. Among others: 

* the start / stop (at first it was done "by hand" because of the way children processes were to be detected and killed manually, now it's no longer needed so we could definitely try seeing how this can be simplified)
* ffmpeg link (because I'm relying on the spk installation but I'm not too sure on how to test this... since none of my cases actually use embedded subtitles)

At this stage it could be helpful to see how you did it on your side :) If you want to contribute I can definitely stop "rewriting" history on the smaarn:feat/add-bazarr-module branch (still not at head though).

Regarding python 3.7 thing the only "thing" which is blocking is me is basically the issue with wheels which are generated in a right manner but with invalid file names... 
It's being reported here https://github.com/SynoCommunity/spksrc/issues/3944

edsenabr:
I've been using my bazaar build for 3 days now, integrated with sonaar and it already downloaded half a dozen of subtitles, automatically. I'm going to integrate it with radarr to test the embedded things, as it way more common on the movies world. I'll let you know. Meanwhile, I'm going to follow @ymartin59 's recommendations for uploading the code.

smaarn:
Upgraded to latest version of bazarr.

Now relying on #3953 to be able to use Python 3.7+

smaarn:
Latest binaries available here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.3-2

smaarn:
Latest binaries available here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.8.4.4

smaarn:
Upgraded to 0.9.0.1 which now requires `numpy` to be available, which requires `cython` to build.
Rebased on top of #3953

cross/numpy/Makefile
```diff
@@ -0,0 +1,16 @@
...
+PKG_DIST_SITE = https://files.pythonhosted.org/packages/source/n/$(PKG_NAME)
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS =
+
+HOMEPAGE = https://numpy.org/
+COMMENT  = The fundamental package for scientific computing with Python
+LICENSE  = https://numpy.org/doc/stable/license.html
+
+ADDITIONAL_CFLAGS = -std=c99
```
smaarn, line 14: This is why a mere requirements file couldn't be used: compilation needed to be customized to pass.

cross/numpy/Makefile
```diff
@@ -0,0 +1,16 @@
+PKG_NAME = numpy
+PKG_VERS = 1.17.5
```
smaarn, line 2: 1.18.x and greater fail to compile on `aarch64` architecture (due to an internal compiler error...)
Since bazarr says `>=1.12` I took the latest version which builds on all platforms.

native/python3/Makefile
```diff
@@ -29,10 +29,12 @@ PIP_NATIVE = $(WORK_DIR)/../../../native/$(PKG_NAME)/work-native/install/usr/loc
...
 .PHONY: nativepython3_post_install
 nativepython3_post_install: $(WORK_DIR)/python-native.mk
-	@$(MSG) Installing setuptools, pip and cffi
-	@$(RUN) wget https://bootstrap.pypa.io/ez_setup.py -O - | $(PYTHON)
+	@$(MSG) Installing pip, setuptools, cffi and cross env
 	@$(RUN) wget https://bootstrap.pypa.io/get-pip.py -O - | $(PYTHON)
+	@$(PIP) install "setuptools==44.1.0"
 	@$(PIP) install "cffi==1.14.0"
+	@$(PIP) install "crossenv==0.7"
+	@$(PIP) install "cython==0.29.21"
```
smaarn, line 37: cython is required to compile some wheels (in my case: `numpy`).
As such I though it would be nice to actually ensure it gets integrated in the native environment.

smaarn:
Latest binaries (0.9.0.1) can be found here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.1-1.

At this stage I guess this PR could be re-reviewed since I believe most of the change requests were actually applied. So, unless I'm mistaken its "work in progress" status could be removed.

smaarn:
Latest version of this PR's outcome available here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.2-1

smaarn:
Hello @hgy59 and @ymartin59 

Sorry for bothering you but I was wondering if there was anything missing on this PR.
I believe there is nothing more which I should be doing but I may be totally missing some points.

Thanks

smaarn:
Updated the PR to take into account latest version of bazarr (https://github.com/morpheus65535/bazarr/releases/tag/v0.9.0.3) : https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.3-1

smaarn:
Rebased @ head

smaarn:
Rebased @ head

smaarn:
Rebased @ head

smaarn:
Reworked to take into account latest release 0.9.0.4 and to avoid polluting native / cross python3 based builds.

pkmetski:
Thanks for the great work, @smaarn!
I have a question - what is the plan with updates? Will every new Bazarr version require repackaging, or will the auto-updater be enabled when the final version of the package is released?

bobbels:
Hi @smaarn , i've installed your earlier build of Bazarr from file 'bazarr_alpine-6.1_0.8.4.4-1.spk'. It works perfect but i would like to update to your latest build. Can i request for this build via this comment? Or do i need to create a pull_request? I've never done this before so, not sure if this is the right way to ask. Many thank in advance for your answer!

hgy59:
> Hi @smaarn , i've installed your earlier build of Bazarr from file 'bazarr_alpine-6.1_0.8.4.4-1.spk'. It works perfect but i would like to update to your latest build. Can i request for this build via this comment? Or do i need to create a pull_request? I've never done this before so, not sure if this is the right way to ask. Many thank in advance for your answer!


The fastest way is to download the packages that are built automaticly with the above checks (github build action).
Your find the zip file with all packages [here](https://github.com/SynoCommunity/spksrc/suites/1282823311/artifacts/19910807) - the artifact package contains all the built packages and is nearly 400MB. For alpine arch you need the generic x64 arch from within packages.zip.

I cannot tell anything about the package update processing for bazarr - so it's better you make a copy of your configuration data first, and when the update fails you need to uninstall before installing the current package.

bobbels:
> > Hi @smaarn , i've installed your earlier build of Bazarr from file 'bazarr_alpine-6.1_0.8.4.4-1.spk'. It works perfect but i would like to update to your latest build. Can i request for this build via this comment? Or do i need to create a pull_request? I've never done this before so, not sure if this is the right way to ask. Many thank in advance for your answer!
> 
> The fastest way is to download the packages that are built automaticly with the checks below (github build action).
> Your find the zip file with all packages [here](https://github.com/SynoCommunity/spksrc/suites/1282823311/artifacts/19910807) - the artifact package contains all the built packages and is nearly 400MB. For alpine arch you need the generic x64 arch from within packages.zip.
> 
> I cannot tell anything about the package update processing for bazarr - so it's better you make a copy of your configuration data first, and when the update fails you need to uninstall before installing the current package.

Hi @hgy59 , I've managed to update Bazarr with the bazarr_armv7-6.1_0.9.0.4-1.spk.  build. The 'System > Status > About' section in Bazarr, showed me the OS on my Synology DS215+ is based on the armv7l chip. As this is 32bit, I went for the armv7 package instead of the x64 one and it worked flawlessly. I'll keep using the armv7 packages for future updates as well. Thanks for putting me in the right direction!

hgy59:
> Hi @hgy59 , I've managed to update Bazarr with the bazarr_armv7-6.1_0.9.0.4-1.spk. build. The 'System > Status > About' section in Bazarr, showed me the OS on my Synology DS215+ is based on the armv7l chip. As this is 32bit, I went for the armv7 package instead of the x64 one and it worked flawlessly. I'll keep using the armv7 packages for future updates as well. Thanks for putting me in the right direction!

@bobbels sorry for the confusion, my fault (I mixed up avoton and alpine)

smaarn:
> Thanks for the great work, @smaarn!
> I have a question - what is the plan with updates? Will every new Bazarr version require repackaging, or will the auto-updater be enabled when the final version of the package is released?

Hello @pkmetski , the plan is to rely on the spk updates as there is currently no way to make Synology NASes actually build native python wheels. For example when updating from 0.8.x to 0.9.x a new wheel (numpy) was required as a dependency which required native bindings. Using the auto updater would have failed on your NAS.

Once this initial PR is passed, updating would be a matter of new PRs.
While we speak I'm working on th 0.9.0.5 update (should be pretty quickly available).

hgy59 (approved):

cross/cython/Makefile
```diff
@@ -0,0 +1,15 @@
+PKG_NAME = cython
+PKG_VERS = 0.29.21
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_VERS).$(PKG_EXT)
+PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
```
smaarn, line 5: This has been inserted to ensure that locally the cache file is persisted with an explicit name.

cross/bazarr/Makefile
```diff
@@ -0,0 +1,27 @@
...
+PKG_DIST_SITE = https://github.com/morpheus65535/bazarr/archive
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = nop
+INSTALL_TARGET = bazarr_custom_install
+
+DEPENDS = cross/unrar
+DEPENDS += cross/numpy
+DEPENDS += cross/webrtcvad
```
smaarn, line 16: Required since 0.9.0.5 version

smaarn:
0.9.0.4-1 "old" version is available here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.4-1
0.9.0.5-1 version is available here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.5-1
Of course you can also take the one generated by the GitHub actions as mentioned by @hgy59 (you just click on the `Details` section basically and download the artifact).

hgy59:
@smaarn I am wondering why `cross/cython/PLIST` file is empty.
I no files of cython are needed in the install/staging folder, then this might be a build dependency only and it's better to move it from cross to native folder?

smaarn:
> @smaarn I am wondering why `cross/cython/PLIST` file is empty.
> I no files of cython are needed in the install/staging folder, then this might be a build dependency only and it's better to move it from cross to native folder?

It’s definitely a build only dependency but I didn’t want to pollute the native python3 installation (and therefore impact all python3 packages build). Of course I could move it inside native python3 installation if you believe it should be.

smaarn:
Rebased @ head.

smaarn:
Rebased @ head.

smaarn:
Rebased @ head.

@ymartin59 sorry for bothering you but do you believe there are still changes which would need being performed before this PR could be merged ?

@hgy59 regarding your point on `cross/cffi`, it's technically a python module. The options I currently have are:

1. Leave it as it is (under `cross` directory, only enabled for people requesting it)
2. Integrate it under `native` in its own directory (knowing that, at the end of the day, it needs to be integrated in python environment and can work both for python 2 and python 3)
3. Integrate it directly under python 3 (which would mean potential impacts on other builds... but hopefully it would be harmless)

smaarn:
Latest generated binaries (using Github actions) can be found directly here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.6-1

smaarn:
Rebasing @ head and attempt using the latest numpy version.

smaarn:
Depends on #4376

ymartin59:
@smaarn @hgy59 According to my understand, bazaar GitHub repository contains no sources to compile, so I would discard `cross/bazaar`. I export package to be provided with an explicit `requirements.txt` file listing all required wheels (even if all of them are only pure python). I propose to inspire from `beets` which is much simpler to maintain/update: https://github.com/SynoCommunity/spksrc/tree/master/spk/beets

smaarn:
> @smaarn @hgy59 According to my understanding, bazaar GitHub repository contains no sources to compile, so I would discard `cross/bazaar`. I expect package to be provided with an explicit `requirements.txt` file listing all required wheels (even if all of them are only pure python). See https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-packages for reference documentation.
> I propose to inspire from `beets` which is much simpler to maintain/update: https://github.com/SynoCommunity/spksrc/tree/master/spk/beets

@ymartin59 I'll try checking this, but bazarr source code comes with a bunch of binaries which are pretty useless given our usages (windows, mac os and linux...). I'll try checking if pip offers a way to filter them out but I have some doubts here.

smaarn:
> > @smaarn @hgy59 According to my understanding, bazaar GitHub repository contains no sources to compile, so I would discard `cross/bazaar`. I expect package to be provided with an explicit `requirements.txt` file listing all required wheels (even if all of them are only pure python). See https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-packages for reference documentation.
> > I propose to inspire from `beets` which is much simpler to maintain/update: https://github.com/SynoCommunity/spksrc/tree/master/spk/beets
> 
> @ymartin59 I'll try checking this, but bazarr source code comes with a bunch of binaries which are pretty useless given our usages (windows, mac os and linux...). I'll try checking if pip offers a way to filter them out but I have some doubts here.

@ymartin59 it doesn't seem to be a viable option for several reasons

1. bazarr doesn't come with a `setup.py` script (which is what `pip` seems to be relying upon, I'm not a python expert - far from it - so if there is any way to tell `pip` to not try to setup the package that could help :) )
2. bazarr comes with binaries for all the "supported" platforms: unrar and ffmpeg (minimum). And the PLIST mechanism doesn't seem to allow exclusion (unless I'm mistaken ?) so it wouldn't be that easy to maintain since we'd need to specify every single folder we wish to include

I must confess I would also have appreciated the option to "just use a requirements.txt file" but bazarr doesn't really come "as a wheel". Though it shouldn't be too hard to create a "bazarr" wheel I'm not too sure about the fact that other people than us (meaning SynoCommunity) would be interested in having such a thing available. (Whether on PyPi or not ;) )

smaarn:
@ymartin59 do you think having the spk Makefile actually handle the downloading part instead would be a viable option ?

I'll try that lead over the week-end if it seems indeed more relevant.

hgy59:
> 
> 
> @ymartin59 do you think having the spk Makefile actually handle the downloading part instead would be a viable option ?
> 
> I'll try that lead over the week-end if it seems indeed more relevant.

It is better to download sources in cross Makefiles. The caching (distrib folder) and validation (digests) is not supported in spk Makefiles.
IMHO this is for all sources, whether compiled or not.

smaarn:
I am doomed.
It seems that now the latest pip version available through the get-pip script no longer supports python 2 :(

th0ma7:
@smaarn You can probably use something like we do for meson `sudo pip3 install meson==0.56.0` ?

hgy59:
> 
> 
> I am doomed.
> It seems that now the latest pip version available through the get-pip script no longer supports python 2 :(

This is the challenge to migrate to python3!

smaarn:
> > I am doomed.
> > It seems that now the latest pip version available through the get-pip script no longer supports python 2 :(
> 
> This is the challenge to migrate to python3!
> 
> EDIT:
> oops, the error is for building spk/python (i.e. python 2). As long as we want to build python2 we need to pin pip for python2 as @th0ma7 proposed above.

Actually the issue issue comes from the pip script which we use for auto installation. It turns out that, indeed, it's unpinned. There may be a simple fix though by changing the URL (you can version it by python version, it seems).

spk/bazarr/src/service-setup.sh
```diff
@@ -0,0 +1,33 @@
...
+SVC_CWD="${SYNOPKG_PKGDEST}/share/${PACKAGE}"
+
+service_postinst ()
+{
+
+    # Create a Python virtualenv
+    ${VIRTUALENV} --system-site-packages ${SYNOPKG_PKGDEST}/env >> ${INST_LOG} 2>&1
+
+    # Install the wheels (using virtual env through PATH)
+    ${SYNOPKG_PKGDEST}/env/bin/pip install --no-deps --no-index -U --force-reinstall \
```
ymartin59, line 25: If wheels are too large, you may reuse generic alternate TMP location support from @hgy59

ymartin59 (approved):

cross/poetry/Makefile
```diff
@@ -5,21 +5,17 @@ PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
...
-BUILD_DEPENDS = cross/cffi
-DEPENDS =
+DEPENDS = cross/cffi
 
 HOMEPAGE = https://python-poetry.org/
 COMMENT  = Poetry helps you declare, manage and install dependencies of Python projects, ensuring you have the right stack everywhere.
 LICENSE  = MIT
 
-POST_INSTALL_TARGET = install_into_cross_env
+PRE_INSTALL_TARGET = pre_install_into_cross_env
```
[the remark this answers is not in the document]
smaarn: Unfortunately this was actually taken from another PR (#4376 ) :/

cross/cffi/Makefile
```diff
@@ -13,14 +13,15 @@ LICENSE  = MIT
 
 PLIST_TRANSFORM = sed -e 's%@PYTHON_SITE_PACKAGES@%$(PYTHON_LIB_DIR)/site-packages%' -e 's%@PYTHON_VERSION@%$(SPK_SHORT_VERS)%'
 
-POST_INSTALL_TARGET = install_into_cross_env
+PRE_INSTALL_TARGET = pre_install_into_cross_env
```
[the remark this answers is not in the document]
smaarn: Unfortunately this was actually taken from another PR (#4376 ) :/

smaarn:
Binaries have been published here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.8-1

tjapsorr:
> 
> 
> Binaries have been published here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.8-1

Nice, but no RTD1296?

smaarn:
> > Binaries have been published here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.8-1
> 
> Nice, but no RTD1296 supported package??

According to the doc https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures the `aarch64` packages should be just fine.

Note: those are **not** official binaries. They are for testing purposes. I personally use the `evansport` architecture and have been using it since binaries were published and everything works perfectly on my side. If you're looking for official packages this PR needs to be merged first.

@hgy59 @ymartin59 is there anything more needed on this PR in order for it to be merged ? Thanks

tjapsorr:
> > > Binaries have been published here: https://github.com/smaarn/spksrc/releases/tag/bazarr-0.9.0.8-1
> > 
> > 
> > Nice, but no RTD1296 supported package??
> 
> According to the doc https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures the `aarch64` packages should be just fine.
> 
> Note: those are **not** official binaries. They are for testing purposes. I personally use the `evansport` architecture and have been using it since binaries were published and everything works perfectly on my side. If you're looking for official packages this PR needs to be merged first.
> 
> @hgy59 @ymartin59 is there anything more needed on this PR in order for it to be merged ? Thanks

Thanks it worked with aarch64 package. I didn’t know.

smaarn:
Superseeded by #4453

nitantsoni:
Just wanted to express my gratitude for bringing Bazarr as a native package. I have been using 0.9.0.8.1 on my Syno since a few weeks and it seems to be working great!!

PR #3804, opened 2019-11-14
Bump dnscrypt-proxy

publicarray:


publicarray:
@ymartin59 do you think it's time to remove the beta flag?

PR #3809, opened 2019-11-22
Inotify tools

cyc1ingsir:
_Motivation:_  There is a very long and old thread in the MinimServer forum about a restart package which requires the inotify tools installed. (https://forum.minimserver.com/showthread.php?tid=426)
inotify tools -> https://github.com/rvoicilas/inotify-tools
In the past, there was a package available but it isn't any longer
This is PR adds a package for inotify tools
_Linked issues:_  No linked issues

### Checklist
- [X] Build rule `arch-x64-6.1` and `arch-aarch64-6.1` completed successfully
- Package upgrade completed successfully - no upgrade - initial vesion
- [X] uninstall of package completed successfully


spk/inotify-tools/src/service-setup.sh
```diff
@@ -0,0 +1,44 @@
...
+        if [ -e "${SYNOPKG_PKGDEST}/bin/$cmd" ]; then
+            ln -s "${SYNOPKG_PKGDEST}/bin/$cmd" "/usr/local/bin/$cmd"
+        fi
+    done
+
+    if [ -e "${SYNOPKG_PKGDEST}/lib/$LIBRARY_WITH_VERSION" ]; then
+        ln -s "${SYNOPKG_PKGDEST}/lib/$LIBRARY_WITH_VERSION" "/usr/local/lib/$LIBRARY_WITH_VERSION"
+        ln -s "${SYNOPKG_PKGDEST}/lib/$LIBRARY_WITH_VERSION" "/usr/local/lib/$LIBRARY"
+        ln -s "${SYNOPKG_PKGDEST}/lib/$LIBRARY_WITH_VERSION" "/usr/local/lib/${LIBRARY}.${LIBRARY_MAIN_VERSION}"
+    fi
```
ymartin59: When built properly with `prefix` option, executables are expected to search for library in `/var/packages/inotify-tools/target/lib/` thanks to ELF option RPATH.
If it is not correct, it is a best option to fix compilation chain to get it work as expected, instead of linking libraries in `/usr/local/lib` which may conflict with other packages or DSM executables.

ymartin59:
@cyc1ingsir Thanks for this contribution. I have requested some improvements in review. I may help with library loading if needed.

cyc1ingsir:
Thanks @ymartin59 for the time and your patience reviewing this PR.
Your suggestions are really helpful.

I pushed a commit addressing most of your suggestions. 
The reason behind most of the superfluous code was that I didn't know better.

Regarding the patches.

This comment and the lack of reaction to it seems to indicate that the project is not really maintained at the moment.
https://github.com/rvoicilas/inotify-tools/issues/68#issuecomment-383297054

I asked the author anyway (not expecting an answer - any time soon - though).
Using the latest git commit would be an option, this is true. However, one of the patches includes a security fix provided by a PR against the original project. What is the best practice here? Include the fix as patch or using the commit from the forked project?

What do you suggest?

P.S. I've seen the `PKG_DOWNLOAD_METHOD = git` at various other packages like ifuse.

ymartin59:
@cyc1ingsir Hello. Yes Makefile variables from ifuse is a good example: https://github.com/SynoCommunity/spksrc/blob/master/cross/ifuse/Makefile#L4-L8
And I proposed to set version as `3.20.1patch1`

cyc1ingsir:
Thanks for your suggestions @ymartin59 .
I applied the suggested changes.
Please have another look.

spk/inotify-tools/src/service-setup.sh
```diff
@@ -0,0 +1,27 @@
+
+BIN_COMMANDS="inotifywait inotifywatch"
+LIBRARY="libinotifytools.so"
+LIBRARY_MAIN_VERSION="0"
+LIBRARY_SUB_VERSION="4.1"
+LIBRARY_WITH_VERSION="${LIBRARY}.${LIBRARY_MAIN_VERSION}.${LIBRARY_SUB_VERSION}"
```
ymartin59: I do not understand what these variables are expected to operate here. From my point of view, should be removed.

cyc1ingsir:
inotify-tools has moved to an organisational account here at GH yesterday.  
A release is to follow soon.  
As soon, the release has been published, I'll update this PR.

ericcurtin:
3.20.2 release complete @cyc1ingsir 

https://github.com/inotify-tools/inotify-tools/releases

ericcurtin:
Maybe it's worth pulling https://github.com/inotify-tools/inotify-tools/releases/tag/3.20.2.2

I fixed a bug or two there. Apologies for the quick releases in just a few days.

hgy59:
@cyc1ingsir I pushed an update to the latest release of inotify-tools and used the new symlink creation support.

Maybe you have an icon with a higher resolution? 72x72 pixels is not optimal to scale up. I propose to use icons with at least 256x256 pixels.

spk/inotify-tools/src/service-setup.sh
hgy59: This is resolved by @cyc1ingsir and the whole service-setup.sh is removed by usage of generic symlink creation.

cyc1ingsir:
Thanks for updating this PR @hgy59 
I replaced the icon with one with 256x256.
Not pretty but hopefully conveys the purpose of the package.

hgy59:
@ymartin59 inotify-tools is provided to support MinimServer, so I do not propose to include it in any synocli-* package.

ymartin59:
Thanks for contribution. Package published.

Issue #3782, opened 2019-10-08
mpfr package needs update


flip111:
`https://www.mpfr.org/mpfr-current/mpfr-4.0.1.tar.xz` is no longer available

it should be `https://www.mpfr.org/mpfr-current/mpfr-4.0.2.tar.xz` now

hgy59:
With the follwing change you can still use the 4.0.1 version
`PKG_DIST_SITE = https://ftp.gnu.org/gnu/$(PKG_NAME)`
It is bad practice to use a dist site that contains the current version onyl. It will break with every new version.

I have a WIP for python 2 update to 2.7.16 where I am verifying the python3 packages too. Therefore I updated mpfr to version 4.0.2 and include the official patches last released on 20190602 to build homeassistant. But there are a lot of other issues on the homeassistant package ;-(

flip111:
Yes i have a workaround already for my local setup. I meant that it needs to be fixed in this repo

https://github.com/SynoCommunity/spksrc/blame/master/cross/mpfr/Makefile#L5

Actually i just removed the patch file, updated the digests and changed version number to 4.0.2
Seems good so far .. but i didn't compile anything with it yet, so a bit uncertain

PR #3811, opened 2019-12-01
Update mpfr (#3782)

hgy59:
- fix download location
- update mpfr from 4.0.1 to 4.0.2
- include patches up to 20190602


ymartin59:
OK but which package may be used to validate and test

hgy59:
It's for building homeassistant (I had it on my python 2.7.16/17 branch but think it is out of scope there, since ha is built with python3)

Issue #3815, opened 2019-12-11
[Package Request] TimescaleDB


SvenVandenbrande:
Name: TimeScaleDB

Description: TimescaleDB: An open-source database built for analyzing time-series data with the power and convenience of SQL — on premise, at the edge or in the cloud.

Website: https://www.timescale.com/
Software documentation: https://docs.timescale.com/latest/main
Source code: https://github.com/timescale/timescaledb
License: https://github.com/timescale/timescaledb/blob/master/LICENSE

Timescale is used by many people that do home automation.

SvenVandenbrande:
@publicarray Any new status for the Update on PostgresSQL + Timescale. :)

PR #3817, opened 2019-12-12
PostgreSQL v11.6

DigitalBox98:
_Motivation:_  Propose PostgreSQL to version 11.6

- In order to avoid any conflict with existing Postgresql 9 installed by default on Synology, port 5433 is used instead of 5432
- Default database in /var/packages/postgresql/target/var/data/ and config file is /var/packages/postgresql/target/var/data/postgresql.conf
- PostgreSQL admin user "pgadmin" added by default in order to define a user password
- Wizard UI at installation time in order to define PostgreSQL admin user/password
- To check that version use : /var/packages/postgresql/target/bin/psql -p 5433 -U pgadmin -d postgres

Testing successfully on braswell architecture

### Checklist
- [x] Build rule "all-supported" completed successfully.


ymartin59:
Thanks for contribution. But I wonder what application may require latest PostgreSQL version when version 9.x still does the job?

hgy59:
TimescaleDB (see #3815) needs PostgreSQL 9.6, 10 or 11
Version 9.3 (as found on my diskstation by synology) is not supported anymore by postgresql.org since Nov 2018
The latest version is 12.1

SvenVandenbrande:
Do you guys have any ETA on the PostgreSQL (+ TimescaleDB)?

DigitalBox98:
I can see several good reasons to have PostgreSQL 11.6 on Synology : 
- Version 9.3 no more officially supported by PostgreSQL, so no more patches will be available like security patches (and I wonder why Synology doesn't propose any update since)
- Some applications will rely on PostgreSQL 9.6, 10 or 11 (like TimescaleDB)
- Good alternative to MariaDB 10
Latest version is 12, but the idea is to propose version 11.6 because this version is more supported by other applications that version 12.

hgy59:
@DigitalBox98 please create a dedicated private branch for this PR.
See [this comment of @ymartin59](https://github.com/SynoCommunity/spksrc/pull/3824#issuecomment-566752272) on how to do it.

spk/postgresql/Makefile
```diff
@@ -0,0 +1,24 @@
...
+DISPLAY_NAME = PostgreSQL
+STARTABLE = yes
+
+HOMEPAGE   = http://www.postgresql.com
+LICENSE    = PostgreSQL Licence
+
+WIZARDS_DIR = WIZARD_UIFILES
+
+INSTALLER_SCRIPT = src/installer.sh
+SSS_SCRIPT = src/start-stop-status.sh
```
ymartin59: I propose to take benefits from generic installer support and service control methods described at: https://github.com/SynoCommunity/spksrc/wiki/Service-Support#installer

If you want to look at code, you can refer to:
- https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.installer
- https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.start-stop-status

I have no idea how long we may go on supporting DSM 5 but using these support allow package to run as-is there too.

ymartin59:
In my opinion, it is required to provide an uninstall wizard to generate a full database dump including configurations to shared folder, so that data does not wanish without notice.

It may also be interesting to propose a regular task to produce a full database dump (or per schema) as backup utility.

spk/postgresql/Makefile
DigitalBox98: My bad, it was OK for my previous Icecast submission, so I will mimick that one to be OK :)

DigitalBox98:
Good proposition, I will check how pgdump can be done before uninstall. Any proposal for an existing Synology folder for that ?
Regarding the regular task, no idea how to do it by script :)

DigitalBox98:
You can close this PR, I will create a new branch and a new PR :)

ymartin59:
I propose you to edit the PR and select the new source branch+repository. So that we keep review and comments.

DigitalBox98:
After a hard work, I got the uninstall wizard proposing the dump of all databases and config files :)
Installer & service should now comply with generic methods as proposed.

Sorry but I had to restart my fork to have a clean base, and so I have proposed a new PR (#3826)

One point I noticed during my tests  : with the new service method, the "daemon status" is not recognizing correctly the database server status compared to the old method. 
After looking at it, I didn't identify the reason (no PID file is generated) so at least that point to review for this new PR.

ymartin59:
@DigitalBox98 So I guess this PR can be closed

DigitalBox98:
Yes this one can be closed.

Issue #3206, opened 2018-03-06
FFMPEG has problems with more then 5.1 Channel audio in DTS


Denton22:
_Package Name:_ FFMPEG
_Package Version:_ 3.4.1-9

_NAS Model:_ DS918+
_NAS Architecture:_ Apollo Lake
_DSM version:_ DSM 6.1.5-15254 Update 1
--

### Expected behavior
That Videos with more then 5.1(side) load up and play.

### Actual behavior
The Video just doesent load at all. Everything works up to 5.1(side) in audio, but as soon as i load a video with 6Channel Audio or 7.1 Audio it will be stuck and "load" forever.

### Steps to reproduce
_1._ Load Any DTS File with 7.1 Audio in Video Station

#### Package log
 Sadly there is no log for FFMPEG on my device anyway.

Can anyone confirm this or is this already a known problem? 

Thanks already 👍

ymartin59:
Please report result of `/usr/local/ffmpeg/bin/ffprobe` with your media file.

Denton22:
Sure thing!
`ffprobe version 3.4.1 Copyright (c) 2007-2017 the FFmpeg developers
  built with gcc 4.9.3 (crosstool-NG 1.20.0) 20150311 (prerelease)
[...]

PR #3822, opened 2019-12-15
FFMPEG 4.2.1 fixes - hls_seek_time + x265 + fdk-acc encoder update + fdk-acc enable default

th0ma7:
_Motivation:_  Ported patches from previous pull where not taking effect due to missing define at compilation time.  Thus the `-hls_seek_time` option as well as other Synology optimizations where not enabled.  Also fdk-acc & x265 had updates available + the fdk-acc licensing as been formally reviewed by RedHat and confirmed to be free and GPL compatible.
_Linked issues:_  #2952 , #3206 

### Checklist

Functionality tested as part of bug entry #2952


m4tt075:
Great job on porting the Synology patches, @th0ma7 !

Just one comment on the licensing question: Activating the `non-free` compilation options indicates, that this is not considered (L)GPL compliant by `ffmpeg`, no matter what Redhat does. Please see #3564 and subsequent discussion. I'd not reintroduce it, until Redhat's understanding of the licensing situation has been "upstreamed" within `ffmpeg`. Once that's the case, `non-free` will no longer be required.

th0ma7:
Thnx!  But all the previous worked you did had paved to way and made this "easier" than it would have been otherwise (so thnx to you on that :) ).

On the FDK ACC front I don't agree but admiditly I need additional readings on this (and I'm not a specialist on that matter).  To me once the legal team of RedHat defines it as free and compatible the issue is basically gone.  I wonder what the `--non-free` notion of FFMPEG has in terms of legal value comparing to redhat take on it?  Also, as the code is free could I modify it so `--non-free` ain't needed with fdk-acc, would it still be OK?

But on the other end, I don't mind that much neither.  It's easy enough to keep as an option and not be activating it by default (that's why i've kept it apart).  And, it won't affect the end-product that much neither as there is a default built-in codec anyway.

Thnx for you time @m4tt075 , cheers!

th0ma7:
Added fix: in order to be using vaapi setuid must be set on the ffmpeg binary as otherwise it cannot access the dri device file `/dev/dri/renderD128` using the following:
```
$ sudo chmod u+s /var/packages/ffmpeg/target/bin/ffmpeg
```

ymartin59:
@th0ma7 Do you consider it as ready to merge or is there any fixes/improvements to include?

th0ma7:
I consider it ready.
There are still pre-existing bugs that will have to look into for subsequent releases of ffmpeg or in planning for dsm7 support.

th0ma7:
This should be my last change on the 4.2 branch for this upcoming release.
Next step is to look into possibility of better handling VAAPI on per arch basis (to be further investigated) and probably migrate on the 4.3.x release at that point in time.

th0ma7:
Lastest changes tested successfully as part of #2952 
Merge complete.
Ready to publish.
... and Merry Christmas ;)

Yod4z:
when it was ready i have to undo change made to work with you -16 version? 
like put old libsynovte.so and broke ln for ffmpeg and ffprobe? 
before installe the official package?

th0ma7:
When the official `ffmpeg` package from SynoCommunuity will come out you can simply install it and that should normally work.  No additional changes are needed.

Also I suggest you have a look at:
https://gist.github.com/BenjaminPoncet/bbef9edc1d0800528813e75c1669e57e

PR #3825, opened 2019-12-18
minio: updated to latest release

fgma:
_Motivation:_  Update minio to latest release

### Checklist


spk/minio/Makefile
```diff
@@ -12,7 +12,7 @@ MAINTAINER = fgma
 DESCRIPTION = The 100% Open Source, Enterprise-Grade, Amazon S3 Compatible Object Storage.
 RELOAD_UI = yes
 DISPLAY_NAME = MinIO
-CHANGELOG = "Update to 2019-10-12T01-39-57Z built with go 1.13.1"
+CHANGELOG = "Update to 2019-12-17T23-16-33Z built with go 1.13.1"
```
hgy59: Please add the kind of update (this one is a `Major Feature Release`) as the timestamp versioning is not very expressive.

And maybe its worth to add the changelog:
**Breaking Change** Lots of new features have been implemented, so rolling upgrade is not supported. Please upgrade all servers.

I would omit `built with go 1.13.1` (there is no change since spk rev 2)

hgy59:
Are there any restrictions for upgrading this package?
The changelog tells `rolling upgrade is not supported`

fgma:
Rolling upgrade probably means you should not upgrade half of your cluster but all of it at the same time. In a single node setup this is not relevant.

spk/minio/Makefile
fgma: Added in https://github.com/SynoCommunity/spksrc/pull/3825/commits/9e5ba6077374db6c32f82607e4ac6bd70b1352e0

fgma:
I've upgraded this PR to the latest minor release and added changelog information.

PR #3826, opened 2019-12-20
PostgreSQL 11.6 including wizard install/uninstall

DigitalBox98:
_Motivation:_  Propose PostgreSQL to version 11.6 including wizard 

- Wizard at installation time in order to define PostgreSQL admin user/password/port
- Wizard when uninstalling in order to propose dump of databases & config files
- To check that version use : /var/packages/postgresql/target/bin/psql -p 5433 -U pgadmin -d postgres

Testing successfully on braswell architecture. 

Note : the "daemon status" is not correctly recognized for an unknown reason, and the side effect is that the service stop via DSM GUI is not possible (only stopping the process manually works)
All other things work perfectly.

### Checklist


ymartin59:
@DigitalBox98 Great job. I have some remarks:

- `/tmp/` location may be too small for database dumps, I propose to request for dump location (volume storage) in uninstall wizard - you may inspire from old (not yet migrated to dsm 6) https://github.com/SynoCommunity/spksrc/blob/master/spk/owncloud/src/installer.sh

- when upgrading, package will probably need values for username and password. So a dedicated storage may be useful - see https://github.com/SynoCommunity/spksrc/pull/3763

Not sure if it is already relevant but some users move package from one device to another. In that case, install wizard should allow to restore setup and databases from dumps.

DigitalBox98:
Ok for dump location, I've made it configurable via the wizard and it's working fine.

Regarding the upgrade, looking at #3763, I'm not sure how to proceed : 
1) add a service_postinst() in order to store each variable in the $INST_VARIABLES file ?
and 
2) Add upgrade_uifile.sh in order to retrieve these variable in the update wizard file ?

Any clue regarding the "daemon status" which is not correctly recognized ? (it appears wrongly as "stopped" in the DSM GUI whereas the service is running fine)

I can propose a PR for the dump location, but I prefer to put this enhancement together with the upgrade enhancement once clarified :)

ymartin59:
@DigitalBox98 Your understanding is correct.

The "daemon status" issue comes from usage of
```
SVC_BACKGROUND=y
SVC_WRITE_PID=y
```
whereas the `pg_ctl` process is transient and stops as soon as forked as "postmaster". (typically this is not a "service command" but a "startup command"). You should remove these options and set `PID_FILE` to `postmaster.pid` location, according to logic of shell: https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.service.start-stop-status

By the way, I am not sure sending a kill signal to PID is the best way to stop PostgreSQL. Triggering stop from `pg_ctl` is probably better. In that case, a complete old-style "dsmcontrol.sh" provided as `SSS_SCRIPT` makefile variable is required. 
Here is an example of such a custom script from an old package not yet migrated to DSM 6 generic service concept:
- https://github.com/SynoCommunity/spksrc/blob/master/spk/ejabberd/Makefile
- https://github.com/SynoCommunity/spksrc/blob/master/spk/ejabberd/src/dsm-control.sh
but yours should be even lighter

Again many thanks for your contribution

DigitalBox98:
The various points should now be ok (dump location, upgrade wizard, deamon status)

spk/postgresql/src/installer.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/sh
+
+# Package
+PACKAGE="postgresql"
+DNAME="PostegreSQL"
```
ymartin59: These variables are never used (replaced by SYNOPKG_PKGNAME), correct? Is so please remove these lines.

spk/postgresql/src/installer.sh
```diff
@@ -0,0 +1,146 @@
...
+  su - ${DAEMON_USER} -s /bin/sh -c "${SYNOPKG_PKGDEST}/bin/pg_ctl -D ${DATABASE_DIR} -l ${DATABASE_DIR}/logfile start"
+
+  # Update existing role
+  su - ${DAEMON_USER} -s /bin/sh -c "${SYNOPKG_PKGDEST}/bin/psql -p ${PG_PORT} -d postgres -c \"ALTER ROLE \\\"${DAEMON_USER}\\\" WITH ENCRYPTED PASSWORD '${PG_PASSWORD}'; \""
+
+  # Create new role
+  su - ${DAEMON_USER} -s /bin/sh -c "${SYNOPKG_PKGDEST}/bin/psql -p ${PG_PORT} -d postgres -c \"CREATE ROLE ${PG_USERNAME} ENCRYPTED PASSWORD '${PG_PASSWORD}' SUPERUSER CREATEDB CREATEROLE INHERIT LOGIN REPLICATION BYPASSRLS; \""
+
+  # Stop server
+  su - ${DAEMON_USER} -s /bin/sh -c "${SYNOPKG_PKGDEST}/bin/pg_ctl -D ${DATABASE_DIR} stop"
```
ymartin59: I am quite worried about what may happen if any command fails. It may be wise to check return code and fail-fast (exit script) to avoid strange package internal state.
There is a common pattern invocation with `su - ${DAEMON_USER} -s /bin/sh -c`. I think it may be worth creating a function with a "command line parameter" handling "su" invocation, logging (like "Running xxxx" before call) and error checking.
What do you think about it?

spk/postgresql/src/start-stop-status.sh
```diff
@@ -0,0 +1,35 @@
...
+  
+DAEMON_USER="`echo ${SYNOPKG_PKGNAME} | awk {'print tolower($_)'}`"
+DAEMON_ID="${SYNOPKG_PKGNAME} daemon user"
+DAEMON_USER_SHORT=`echo ${DAEMON_USER} | cut -c 1-8`
+
+DATABASE_DIR="${SYNOPKG_PKGDEST}/share/data"
+
+daemon_status ()
+{
+    ps -efa | grep "postgresql" > /dev/null
```
ymartin59: Isn't there a way to test explicit process PID from "postmaster.pid" (you can reuse status check from framework script)? I wonder if such check may just always return true as DSM runs its own embedded postgresql instance.

spk/postgresql/src/start-stop-status.sh
```diff
@@ -0,0 +1,35 @@
...
+  status)
+    if daemon_status ; then
+      exit 0
+    else
+      exit 1
+    fi
+  ;;
+
+esac
+~
```
ymartin59: Is that character of any use?

spk/postgresql/Makefile
```diff
@@ -0,0 +1,24 @@
...
+RELOAD_UI = yes
+DISPLAY_NAME = PostgreSQL
+STARTABLE = yes
+
+HOMEPAGE   = http://www.postgresql.com
+LICENSE    = PostgreSQL Licence
+
+WIZARDS_DIR = WIZARD_UIFILES
+
+INSTALLER_SCRIPT = src/installer.sh
```
ymartin59: In my opinion you should keep `SERVICE_SETUP` variable and "service_preinst", "service_postinst" previous version of functions... Here package has lost many common framework support and it does not make sense to copy-paste them.
Please also take care to your scripts space indentation.

spk/postgresql/Makefile
```diff
@@ -0,0 +1,24 @@
...
+SSS_SCRIPT = src/start-stop-status.sh
```
ymartin59: That one is OK.

DigitalBox98:
I've put again the SERVICE_SETUP, but now I've an issue I can't identify : the start-stop-status.sh is not working at all.

Trying to debug, I have hardcoded the start-stop-status and I have the below : 

1) scripts/start-stop-status launched from DSM GUI : I can see in my log the call to pg_ctl is done but there's no postmaster.pid created, meaning the start process is failing (don't know why, very strange). Part of hardcoded script is as below : 

case $1 in
  start)
    echo "call started " >>/var/packages/postgresql/target/share/data/debug.log
    su - sc-postgresql -s /bin/sh -c "/var/packages/postgresql/target/bin/pg_ctl -D /var/packages/postgresql/target/share/data -l /var/packages/postgresql/target/share/data/logfile start"
    echo "call ended " >>/var/packages/postgresql/target/share/data/debug.log
    exit 0
  ;;

2) Launching manually via SSH (root) the hardcoded start-stop-status script : in that case it's working fine

./start-stop-status start
waiting for server to start.... done
server started

./start-stop-status status
pg_ctl: server is running (PID: 26350)
/volume1/@appstore/postgresql/bin/postgres "-D" "/var/packages/postgresql/target/share/data"


3) Launching manually the "pg_ctl" is working fine as well 
su - sc-postgresql -s /bin/sh -c "/var/packages/postgresql/target/bin/pg_ctl -D /var/packages/postgresql/target/share/data -l /var/packages/postgresql/target/share/data/logfile start"
pg_ctl: server is running (PID: 26350)
/volume1/@appstore/postgresql/bin/postgres "-D" "/var/packages/postgresql/target/share/data"

I don't see any good explanation why from DSM the start seems to fail. Well for the moment I stuck here. 
The current working version I have is when using both "old fashion method" (INSTALLER_SCRIPT & SSS_SCRIPT)

hgy59:
@DigitalBox98 did you look into /var/log/messages? When working on ntopng packet this gave me some additional information with similar problems

DigitalBox98:
Good point :)
I can notice this line when starting the postgresql server, but not sure what does it mean  : 
CloudStation synoscgi_SYNO.Core.Package_2_list[20798]: list.cpp:330 SYNO.Core.Package.list rollback to direct call

hgy59:
@DigitalBox98 I see such log entries too, but I have no idea what this about, but IMO it's not related to your package start.
I thought more of entries like "command not found" or "permission denied"

hgy59:
@DigitalBox98 I cannot reproduce you findings.
I cloned you repo, checked out your branch an built for x64.
I installed it on my DS218+ without auto running after installation.
The I manually started postgresql and it runs without any error. I can stop and run again all within the package center.

DigitalBox98:
Thanks for your tests.
What I mean is that : 
- Current PR is working fine as well on my NAS but is using INSTALLER_SCRIPT+SSS_SCRIPT instead of new SERVICE_SETUP 
- When trying to move to SERVICE_SETUP + INSTALLER_SCRIPT as propose by ymartin59, the start/stop/status is not working anymore from DSM : there's no PR currently for this, as I would like first to get it working before submitting the PR :)
I hope it clarifies
So currently this PR is working but with old methods.

ymartin59:
@DigitalBox98 Probably a limitation in framework code... I was convinced it is possible to mix both behavior but I may be wrong. I will have a look and try to fix, so that your package contains less lines to maintain.

hgy59:
Just a hint:
In PR #3866 I moved from INSTALLER_SCRIPT+SSS_SCRIPT to SERVICE_SETUP+SSS_SCRIPT and installation and service work for DSM>=6 (but not for older DSM yet)

DigitalBox98:
@hgy59 thanks for the hint. 
I had a look at it and comparing with my tests to use SERVICE_SETUP+SSS_SCRIPT I didn't identify what is wrong. The only thing I know is that postgresql might have some restrictions when running the server (like a specific username to use)

@ymartin59, sure let me know when you will have time if you succeed to run it with  SERVICE_SETUP. During my tests, I came to the conclusion that postgresql 9.3 from Synology is quite useable (even if this version is no more supported). The only changes to do is to setup pg_hba.conf or postgresql.conf, and to create a new admin user in order to define login/password when other packages need it. That's how I get nextcloud working with embedded Postgresql without using MariaDB :)

hgy59:
As the fork of this PR is gone, this work is resumed in #4267

Issue #3834, opened 2019-12-28
Midnight Commander: SFTP Support


chrisellsworth:
### Setup
_Package Name:_ Midnight Commander (mc)
_Package Version:_ 4.8.21-5

_NAS Model:_ DS1019+
_NAS Architecture:_ x64
_DSM version:_ 6.2.2-24922 Update 4

### Expected behavior
`SFTP link...` option exists in the PullDn menu

![Screen Shot 2019-12-27 at 21 24 40](https://user-images.githubusercontent.com/1240464/71539352-fde29080-28ef-11ea-95d4-0ad245fc62b8.png)
(screenshot from `mc` on macOS via homebrew)

### Actual behavior
`SFTP link...` option does not exist in the PullDn menu

![Screen Shot 2019-12-27 at 21 24 45](https://user-images.githubusercontent.com/1240464/71539353-fde29080-28ef-11ea-8d37-8d606c42e493.png)

### Steps to reproduce
_1._ Press F9
_2._ Examine menu options

If I were to guess, this is because the version of `mc` on my Mac was compiled with `'--enable-vfs-sftp'` (per `mc --configure-options`)

hgy59:
This is finally solved with #3837

chrisellsworth:
Thank you @hgy59 for working on this!

ymartin59:
Available with 1.1-2: https://synocommunity.com/package/synocli-file

Issue #3783, opened 2019-10-10
Reopen this - MC mouse is not working anymore #3258 - 


Merwenus:
Same issue as below
https://github.com/SynoCommunity/spksrc/issues/3258

Midnight Commander 4.8.21-5

When you start normally it works great, but when you open elevated:
sudo MC

You cant use the mouse, it writes random characters in the bottom, and clicks do not work.

hgy59:
Did some debugging now.
as elevated user (root, admin) on my diskstation the /bin/ash shell is used.
Then found that it is not an issue with the shell, but when missing the TERMINFO export the mouse does not work in mc.
Solved this with `export TERMINFO=/usr/share/terminfo` before running mc.
Hope this helps.

hgy59:
This is finally solved with #3837

Issue #3258, opened 2018-03-26
MC mouse is not working anymore


gielie:
### Setup
_Package Name:_midnight commander
_Package Version:_4.8.20

_NAS Model:_415play
_NAS Architecture:_evansport (x86)
_DSM version:_6.2

### Expected behavior
before the update i was able to select files with the mouse and go to directories

### Actual behavior
The mouse is not working at all

### Steps to reproduce
_1._reinstalled mc, no luck

gielie:
I tried both.
[code]
ash-4.3# echo $TERM
xterm-256color[/code]

after `export TERM=xterm`

[code]
ash-4.3# echo $TERM
xterm
[/code]

I have 2 different synology's and the one i updated version of mc doesn't have mouse support, the older version of mc works fine. I also tried putty on pc and  terminal on osx with the same result.

Where can i find the old version of mc?

ymartin59:
So you run `screen`, is that right? Have you without starting `screen`.
If `mc -x` works, trouble comes from your "terminal emulation" context: http://midnight-commander.org/wiki/doc/faq
Latest package runs properly with mouse support from a linux terminal emulator, so issue is not in package/application itself.

hgy59:
@gielie maybe this is related to the DSM version.
After update to DSM 6.1.6-15266 I can not reproduce this behavior. 
I remember to have the mouse problem too, i.e. when clicking the mouse, some ESC-sequences appeared on the commandline, and the `home` key had a delay of about 2 seconds.

- May be related to the DSM 6.2 beta version?
- Or may be the evansport arch? (works for me now with DSM 6.1.6 on apollolake and armada370)

hgy59:
just found that the mouse works find with regular admin user. But when running with `sudo mc` the mouse does not work and the home key has the delay as mentioned above.
But strange is, that when running  `sudo mc` I got a `$` prompt instead of a `#` prompt on the mc command line. (On other systems there is a  `#` prompt when running  `sudo mc`).
@gielie please verify whether this behavior is the same for yours.

ymartin59:
@gielie Please also compare with separate `sudo -s` and then use of `mc`.

gielie:
After the latest DSM update (DSM 6.1.6-15266 Update 1) my mc is working normal, i can start mc with sudo or with -s -c all variations work normally

PR #3837, opened 2019-12-30
Rework mc (midnight commander) of synocli-file

hgy59:
_Motivation:_  Handle reported issues and fix subshell for the synology way to define /bin/sh as link to /bin/bash
_Linked issues:_  #3834, #3783, #3258 

- add sftp support and hence add cross/libssh2
- fix mouse handling
- replace usage of ncurses by slang (makes the package smaller, fixes mouse handling, is default)
- add patch to fix subshell for /bin/sh shell linked to /bin/bash (subshell for /bin/sh is not supported)
- add new mc icon
- update **synocli-file** with reworked mc

### Checklist


spk/synocli-file/Makefile
```diff
@@ -1,37 +1,30 @@
 SPK_NAME = synocli-file
 SPK_VERS = 1.0
```
ymartin59: I propose to increase package version itself too
```suggestion
SPK_VERS = 1.1
```

spk/synocli-file/Makefile
```diff
@@ -1,37 +1,30 @@
...
-# to build rnm, GCC >= 4.8 is required, but not contained in older toolchains.
-DEPENDS += cross/rnm
-endif
 
 MAINTAINER = SynoCommunity
 DISPLAY_NAME = SynoCli File Tools
 ifneq ($(findstring $(ARCH), powerpc ppc824x ppc853x ppc854x $(ARM5_ARCHES)),$(ARCH))
+# to build rnm, GCC >= 4.8 is required, but not contained in older toolchains.
+DEPENDS += cross/rnm
 DESCRIPTION = "SynoCli File Tools provides a set of small command-line utilities: less, tree, ncdu, jdupes, rhash, mc \(midnight commander\), nano, file, detox, rmlint, rnm."
```
ymartin59: Just an idea to avoid duplicated DESCRIPTION: use a variable for only variable part content ", rnm" with list of tools according to condition...

spk/synocli-file/Makefile
hgy59: My intent is to update revision for updated components and update the version with adding/removing components.
But we could use major/minor package version for this.

ymartin59: I prefer the other way "major/minor" to get a reason to increase major from time to time...

hgy59:
sorry, just found that rmlint does not compile for ppc853x, so I removed the label "status/to-publish".

PR #3853, opened 2020-01-12
NextCloud 17.0.2 using Synology's embedded postgres

DigitalBox98:
_Motivation: Propose beta package for NextCloud 17.0.2 which will use Synology's embedded Postgres.

### Checklist

Tested OK on Synology DS716+II with the following setup : 
- Web Station
- Apache HTTP Server 2.4
- PHP 7.3 with all extensions enabled
- Embedded PostgreSQL 9.3.22

Installation process : 
During package installation, the user will define a new "pgadmin" user and password for the creation of the "nextcloud" database.
During nextcloud installation, the user will indicate a nextcloud "admin" and password user, plus PSQL database connection details ("pgadmin" user and password, database name "nextcloud" and host "127.0.0.1:5432").
Note : for an unknown reason, sometime an error is raised during the nextcloud installation. In order to bypass this issue, reconnect to nextcloud URL and redo the nextcloud installation with another admin user name.

nextcloud
nextcloud2

In order to completely remove "nextcloud" in cas of package installation, you will need to manually : 
- drop table nextcloud 
- drop role oc_admin


ymartin59:
@DigitalBox98 I had an idea: what about to use a wizard drop-down to select user preferred database setup (embedded postgresql, postgresql 9.6 or mariadb...) so that he may transfer existing mysql/mariadb database into DMS?

Please look at "generic service" support documentation to use "SERVICE_SETUP" single script instead of installer/dsm-control split: https://github.com/SynoCommunity/spksrc/wiki/Service-Support

Notice that DSM 6 comes with a "resource acquisition" support that ease database creation and connection settings: https://originhelp.synology.com/developer-guide/resource_acquisition/maria_db.html
I have no idea if it is worth to be used, as I do not see same section for embedded postgresql.

I am working on ffmpeg, then I will come back here and help

DigitalBox98:
For the wizard drop-down to select user preferred database, I'm not sure to get the idea. Currently nextcloud can use embedded postgresql 9.3 (ok), postgresql 11.6 (ok), MariaDB (not tested) or sqlite (ok). This choice will be done on the first connection to NextCloud URL (where new database is created, etc). The only difference I can see is that choosing embedded postgresql will imply a reboot of the NAS to handle changes done in the pg_hba.conf file. All other choices will run the same nextcloud installation.

Regarding the generic service, I'm just wondering what will be the SERVICE_COMMAND as nextcloud is launched via PHP pages which is using apache web server. Maybe I miss something here to provide a generic service ? 

Thank you for the "resource acquisition" information. I had a quick look at it, but currently I don't see any useful usage for postgresql :)

I'm currently looking how to use "onlyoffice" with nextcloud (there's a standard plugin for that, and onlyoffice document server is a really great "extension" to nextcloud). Cross compiling OnlyOffice seem quite a hard task, and I think a first approach is installing an OnlyOffice Document Server via Docker on the NAS : that's the way I'm investigating to have it working with latest nextcloud version (18).

DigitalBox98:
OnlyOffice Document Server is working fine with Docker on Synology and the NextCloud plugins recognize it :) The only drawback I can see, is that OnlyOffice Document Server will use an additional Postgresql server (inside docker)

hgy59:
Closed due inactivity and replaced by #6814

PR #3833, opened 2019-12-27
Python 3.6.10 + fix cross-compilation issues

smaarn:
_Motivation:_  I'm working on migrating bazarr (PR #3788 ) and it supports Python 3.6 ... But the one on spksrc doesn't work.
_Linked issues:_  Optionally, add links to existing issues or other PR's

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

This is a raw revival of PR #3731

PR #3854, opened 2020-01-13
Determine and build affected spks

chickenandpork:
_Motivation:_  It's difficult sometimes to know what SPKs to build to verify a change to a dependent `cross`.  This script is a way to determine the changes in the current PR, and for each affected `cross`, `native`, or `spk`, determine the dependencies.  Recurse up to 30 times to find all dependent SPKs.

Built in awk because awk is everywhere.  Like vi.

Originally started as #3459
_Linked issues:_  This is a tool I use, as follows:

### Setup
1. forked `SynoCommunity/spksrc` into my own space `chickenandpork/spksrc`
2. `git clone chickenandpork/spksrc`
3. `git add remote upstream https://github.com/SynoCommunity/spksrc.git`

You should now have something like the following, except with your own fork:
```
bash-4.3$ git remote -v
origin	git@github.com:chickenandpork/spksrc.git (fetch)
origin	git@github.com:chickenandpork/spksrc.git (push)
upstream	https://github.com/SynoCommunity/spksrc.git (fetch)
upstream	https://github.com/SynoCommunity/spksrc.git (push)
```

After you've made your change on your own development branch, run this:
```
    git diff --stat upstream/master...HEAD | \
        bash mk/build-target-determinator.bash | \
        sort
```

The results will show the SPKs to build; in my own work, I pipe this into a script to build each, and it takes a very long time, but for now, this will tell you what to build.  For example:

### Test: #3833
```
    bash-4.3$ git remote add smaarn https://github.com/smaarn/spksrc.git
    bash-4.3$ git fetch smaarn build/python3/fix-cross-compilation-issue
    bash-4.3$ git checkout build/python3/fix-cross-compilation-issue
    bash-4.3$ git diff --stat upstream/master...HEAD | bash mk/build-target-determinator.bash
    python3 borgbackup homeassistant 
```
This shows that @smaarn might need to test `borgbackup`, `homeassistant` might just be specific to my workspace.

### Test: #3648 (python-2.7.16)
```
bash-4.3$ git remote add hgy59 https://github.com/hgy59/spksrc.git
bash-4.3$ git fetch hgy59 python_2.7.16
remote: Enumerating objects: 42, done.
remote: Counting objects: 100% (42/42), done.
remote: Total 70 (delta 42), reused 42 (delta 42), pack-reused 28
Unpacking objects: 100% (70/70), done.
From https://github.com/hgy59/spksrc
 * branch              python_2.7.16 -> FETCH_HEAD
 * [new branch]        python_2.7.16 -> hgy59/python_2.7.16
bash-4.3$ git checkout python_2.7.16
Branch 'python_2.7.16' set up to track remote branch 'python_2.7.16' from 'hgy59'.
Switched to a new branch 'python_2.7.16'
bash-4.3$ git diff --stat upstream/master...HEAD | bash mk/build-target-determinator.bash
python3 sabnzbd rdiff-backup duplicity borgbackup python mercurial flexget pyload octoprint homeassistant ffsync deluge rutorrent 
```
I think this list is similar to what @hgy59 has been testing.

The longer-term intent is to enable a consistent logic for Developers to use to build-test, plus enable automating the architecture-specific build of updated packages in a TC or CI/CD

### Checklist
- [n/a] Build rule `all-supported` completed successfully

### DIFF
--- a/mk/build-target-determinator.bash
+++ b/mk/build-target-determinator.bash
@@ -1,0 +1,51 @@
+#!bash
+#
+# A fairly-reliable method of determining the SPKs that need to be rebuild based on the files that
+# changed in the last commit.  Only looks at cross/* native/* and spk/* because, well, mk/* might
+# be a bit harder, and toolchains/* would require a rebuild of the world.
+#
+# I did this as a script rather than inside a mk/spksrc.spk.mk target to ensure I'm not evaluating
+# this on every run, only when needed.
+
+# test via:  git diff --stat HEAD^^^ | bash mk/build-target-determinator.bash
+# expected output is a list of subdirs of spk dir
+awk '
+func getdeps(depdir) {
+	checked[depdir]++
+
+	# somewhat reliable shell-exec grep for Makefiles referring to what we have
+	command = "grep -l \"DEPENDS .*"depdir"\" {cross,native,spk}/*/Makefile"
+	while ((command | getline line) > 0) {
+		split(line,path,"/");
+		dep[path[1]"/"path[2]]++;
+	}
+	close(command)
+}
+
+BEGIN {
+    # This is effectively a typedef, avoids errors when check has no elements
+	checked[2]="bogus"; delete checked[2];
+}
+
+/^ (cross|native|spk)\// {
+	split($1,path,"/");
+	dep[path[1]"/"path[2]]++;
+}
+
+END {
+	max_cycles = 30  # avoid infinite loop
+	while ((length(checked) < length(dep)) && (max_cycles--)) {
+		# easier with re-entrant code :)
+		# but instead we loop until the checked == depends
+		for (d in dep) {
+			# simplified "if (d not in checked) { getdeps(d)}"
+			found=0;
+			for (c in checked) if (c == d) found++;
+			if (1 > found) {
+			   getdeps(d)
+			}
+		}
+	}
+	# print the collected dependents, spk subdirs only
+	for (d in dep) if (d ~ "^spk/") { gsub("^spk/","",d); printf "%s ",d; } printf "\n";
+}'


smaarn:
@chickenandpork great idea ! What about having a simple make rule which would list the affected packages ?

Example: `cd spk/python3 && make list-all-dependents` ?

By the way, are we sure that awk is also installed with the Dockerfile ? (Because the current Dockerfile image doesn't include vi for example ;) ) For Mac users this is kinda a minimum (I'm among them, see ;) )

chickenandpork:
> What about having a simple make rule which would list the affected packages ?

That's where I was originally going in the work leading to the original PR, but I'd like to get this merged first while improving the logic that can use it.
> 
> Example: `cd spk/python3 && make list-all-dependents` ?
> 
> By the way, are we sure that awk is also installed with the Dockerfile ? (Because the current Dockerfile image doesn't include vi for example ;) ) For Mac users this is kinda a minimum (I'm among them, see ;) )

chickenandpork:
I don't see how this got closed.

hgy59:
> @chickenandpork great idea ! What about having a simple make rule which would list the affected packages ?
> 
> Example: `cd spk/python3 && make list-all-dependents` ?
> 

I think `list-all-dependents` would easy lead to missunderstanding.
we already have  `cd spk/python3 && make dependency-tree`

I would propose a naming like `cd spk/python3 && make list-affected-spks`

hgy59:
until now my solution is to create a file with all the dependency trees:
```
for package in $(find spk/ -maxdepth 1 -type d | cut -c 5- | sort)
do
    pushd spk/$package > /dev/null
    make dependency-tree
    popd > /dev/null
done
```
and within the result I search for the package under modification to find the affected packages.

Finally my script to build all affected packages (for python2.7.16 branch)
[makeall_python2.sh](https://github.com/SynoCommunity/spksrc/files/4055149/makeall_python2.sh.txt)

As one can see python2 dependency is not as easy as your solution, since my tests include the packages with `SPK_DEPENDS="python>=2"` too.

chickenandpork:
> since my tests include the packages with SPK_DEPENDS="python>=2" too.

Good point.  My intention here is not to handle all possible cases, but to offer something that is better than "I built my own package and it was fine".  AWK can do a bit more to interpret the dependency, but that will take more development down the road.  I've built this sort of smarter dependency tracker (For a project called "Smallfoot" I built an interpreter of the `Kconfig` dependencies in the linux kernel)

We could also get ourselves in trouble with any conditionals, such as SPK_DEPENDS inside of `ifeq(` logic.

If I could extend this generic dependency tracker to get the same coverage as your `makeall_python2.sh`, that would be awesome, but clearly it currently falls short.

In the meantime, it would offer better than not knowing any things to buildtest.

hgy59:
It's no requirement to build packages with SPK_DEPENDS - it's more for testing such packages rather then building.

Alas another point: will the build stop on error or continue?
My makeall-scripts continue with the next arch/package on error. Such scripts run hours (or even days) and finally I have an overview of failing and succeeding package/arch compinations.

chickenandpork:
> Alas another point: will the build stop on error or continue?

`make -k` should control that, right?

I see the benefit of being able to blast through everything and see what fails; my goal here is a really brief way to say “run this and the right builds will be done” .. there’s work to enhance to handle more complicated dependencies, but my preference would be to merge something beneficial and make better and smarter rather than block on the broader use-case from the start.

publicarray:
I know it's rare only modify the cross part of a spk (usually the spk revision is increased too) but this returns nothing. Is than intentional? `echo '#nop' >> cross/openssl/Makefile works beautifully 

PS. Why not use `#!/usr/bin/env bash` as the shebang?

```sh
echo '#nop' >> cross/dnscrypt-proxy/Makefile
git add -A
git commit -m "nop"
wget https://raw.githubusercontent.com/SynoCommunity/spksrc/99a8e91726e2b9601bab1b31035188faeb5b8895/mk/build-target-determinator.bash
chmod +x build-target-determinator.bash
git diff --stat HEAD~
git diff --stat HEAD~ | bash build-target-determinator.bash

# cleanup
git reset HEAD~
git checkout cross/dnscrypt-proxy/Makefile
rm build-target-determinator.bash
```

Whats going on here?

```sh
❯ echo '#nop' >> native/go/Makefile

❯ git add -A

❯ git commit -m "nop"
[github-actions abf12779] nop
 2 files changed, 52 insertions(+)
 create mode 100755 build-target-determinator.bash

❯ git diff --stat HEAD~ 
 build-target-determinator.bash | 51 +++++++++++++++++++++++++++++++++++++++++++++++++++
 native/go/Makefile             |  1 +
 2 files changed, 52 insertions(+)

❯ git diff --stat HEAD~ | bash build-target-determinator.bash
synocli-net 
```

hgy59:
@chickenandpork such a dependency detection is now used in the github build action. This action builds dependent packages when you push changes to your local repo too (except you deactivated actions for your repo).

With the following script you can use the build-script outside the github action:

``` bash
#!/bin/bash
# run github build action locally on command line

# This script takes one single arch that may be overriden by env variable or script argument

if [ -z "${GH_ARCH}" ]; then
    if [ -n "${1}" ]; then
        # take command line parameter
        GH_ARCH="${1}"
    else
        # default
        GH_ARCH=x64-6.1
    fi
fi

GITHUB_WORKSPACE=$(realpath $(dirname $0))

# set parameters originally evaluated in .github/workflows/build.yml

export GH_ARCH=${GH_ARCH}
# evaluate changed files on local branch
export GH_FILES="$(git diff --name-only -r origin/master | xargs)"

# define response files
export BUILD_ERROR_FILE=$GITHUB_WORKSPACE/build_errors.txt
export BUILD_UNSUPPORTED_FILE=$GITHUB_WORKSPACE/build_unsupported.txt
export BUILD_SUCCESS_FILE=$GITHUB_WORKSPACE/build_success.txt
# remove response files
rm -f ${BUILD_ERROR_FILE}  &> /dev/null
rm -f ${BUILD_UNSUPPORTED_FILE}  &> /dev/null
rm -f ${BUILD_SUCCESS_FILE}  &> /dev/null

./.github/actions/build.sh
./.github/actions/build_status.sh
```

The used response files are introduced to collect succesful and failed builds and to ignore errors due to unsupported builds (unsupported archs or TC versions).

The dependency check is included in the framework with a new make target named `dependency-list`.
To work around conditional dependencies I introduced the variable `OPTIONAL_DEPENDS` that must define all optional dependencies in Makefiles.

example:
```bash
$ cd spk/minio
$ make dependency-list
minio: cross/busybox cross/minio native/go native/go-1.4
$
```

If you call `make dependency-list` in the root folder of the repository, you get a list of all spk-packages in the format above.

With such a list it was easy to use grep to parse for dependencies (unlike to parse the tree with awk).

_I created the script above to test builds locally with other root folder than `/spksrc`_.

chickenandpork:
Hi @hgy59 when I first did this work in January, it represented the running script I was using on my personal Jenkins build of spksrc -- I automate things as my day-job, and was doing the same for spksrc.

This work was based on a running config I was using at the time.  This was on a system that didn't have GitHub Actions.

> except you deactivated actions for your repo

My work was done when I don't think Actions existed, so it's a case of not enabling the new thing.

Indeed, my long-term goal was to enable random contributors who don't have the time to dive into the build process to at least offer the occasional build agent for a Jenkins build, to scale to the breadth of donated CPU time.  That constrained my deliverable somewhat.

I liked that the entire work was inside the Make file logic, and would work on other-than-github, but if this solution is working now, I can drop my stuff and maintain less.

The bigger problem is a common one in large organizations: we have no clean method of canarying new builds, so we don't know for example if a certain build will break an 11-yr-old DS209II+ until we ship the update and get complaints.

...but that's another thread.

PR #3880, opened 2020-02-12
syncthing: Update to 1.3.4

mattrobenolt:
Hi there, I'm not exactly sure how to test this or validate that things are working correctly, but kicking this off to see if this is good enough or if I can get some help. I'd like to contribute more and help maintain this package if I can get squared away.

Thanks!

### DIFF
--- a/cross/syncthing/Makefile
+++ b/cross/syncthing/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = syncthing
-PKG_VERS = 1.2.2
+PKG_VERS = 1.3.4
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-source-v$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/syncthing/syncthing/releases/download/v$(PKG_VERS)
--- a/cross/syncthing/digests
+++ b/cross/syncthing/digests
@@ -1,3 +1,3 @@
-syncthing-source-v1.2.2.tar.gz SHA1 2fb3525a8589878239917e4134dae5aa623a04d7
-syncthing-source-v1.2.2.tar.gz SHA256 7a6c6a427a84f33268f1e23ebf6e230abf140c01bb2476a7b34127c03b82b2f1
-syncthing-source-v1.2.2.tar.gz MD5 250c90695ddf0bdf1eefffc3048df630
+syncthing-source-v1.3.4.tar.gz SHA1 cf7efc6829f564e92c006c6ade1ec186a71a9441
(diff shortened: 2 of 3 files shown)


acolomb:
Looks good code-wise.

However, the general question is: What is the benefit of updating the SPK without any changes to the packaging? Syncthing will automatically update to the most recent release internally. I was planning to prepare a new SPK release myself when version 1.4.0 will be released in March, to avoid the faint possibility of database incompatibility problems when re-installing.

Note that there is now a parallel effort to package an official SPK for Syncthing which will be released in sync with the main program, see [this forum discussion](https://forum.syncthing.net/t/new-package-for-synology-nas/14396). Unfortunately it differs from our SynoCommunity SPK in some important aspects, which I hope to get resolved with the Syncthing maintainers to avoid compatibility issues when users switch their package source.

The new package will be released in SPK form, synced and automated, with every upstream Syncthing release. The internal update mechanism is disabled. Currently it lives only in a separate repository, but trying to get into the official Synology store. Such frequent SPK updates are problematic there however, just as they are for this project. This effort by @mattrobenolt is only the first step, but the spksrc maintainers need to publish each individual SPK release, which is a serious bottleneck.

So I agree that making the "distribution system" (SPK repo in this case) solely responsible for updates is a good idea in general and internal program updates should be disabled. However, with a lively project like Syncthing with a frequent release schedule, it needs to be fully automated to work out, and the upstream maintainers should also support the package when something breaks.

SynoCommunity, what are your opinions on the broader matter?
* Should we keep publishing SPKs that will always lag behind, although the software itself updates internally?
* Can we reach a more frequent release cycle to stay up-to-date with upstream?
* Or should all efforts be focused on supporting upstream to produce their own SPKs?

mattrobenolt:
> What is the benefit of updating the SPK without any changes to the packaging? Syncthing will automatically update to the most recent release internally.

I wasn't aware this was the case, if so, that's a bit counter intuitive for me as a user. I will also admit that I'm relatively new to the Synology ecosystem, and I would not expect this behavior.

I personally try not to mix major versions in my syncthing clusters when I can, and when I installed the package here I got 1.2.2 so thought I'd go down the path of seeing how to upgrade it into 1.3.x. Maybe it would have updated if I left it running longer?

mattrobenolt:
Oh, I reinstalled the 1.2.2 package and you're right, I got prompted to upgrade. Yeah, so not sure if there's much value in updating the package here. I personally don't like this behavior, but it seems to be the way things are. :)

ymartin59:
Thanks for your contribution. Included in @hgy59 refactor of Go compilation. PR #3884

Issue #3386, opened 2018-07-16
Improve golang support


ymartin59:
Current support for golang has following drawbacks:
- download same `dep` each time for each arch
- `dep ensure` takes ages without progress bar - is it possible to re-use and prevents its run for each arch?
- more support for dependencies has to be provided, to ease application compilation process. See #2533 for Gogs as example

```
===>  Compiling for dnscrypt-proxy
mkdir -p /spksrc/spk/dnscrypt-proxy/work-rtd1296-6.1/bin
curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  5110  100  5110    0     0    852      0  0:00:05  0:00:05 --:--:--  1140
ARCH = amd64
OS = linux
Will install into /spksrc/spk/dnscrypt-proxy/work-rtd1296-6.1/bin
Fetching https://github.com/golang/dep/releases/latest..
Release Tag = v0.4.1
Fetching https://github.com/golang/dep/releases/tag/v0.4.1..
Fetching https://github.com/golang/dep/releases/download/v0.4.1/dep-linux-amd64..
Setting executable permissions.
Moving executable to /spksrc/spk/dnscrypt-proxy/work-rtd1296-6.1/bin/dep
cd /spksrc/spk/dnscrypt-proxy/work-rtd1296-6.1/src/github.com/dnscrypt-proxy/dnscrypt-proxy-2.0.15 && dep ensure
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  github.com/agl/ed25519

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies,

cd /spksrc/spk/dnscrypt-proxy/work-rtd1296-6.1/src/github.com/dnscrypt-proxy/dnscrypt-proxy-2.0.15/dnscrypt-proxy && go build -ldflags="-s -w"
```

hgy59:
[...]

PR #3883, opened 2020-02-14
synocli-file: add ripgrep, initial rust support

Hylen:
This also adds support for building rust packages. 

This solves issue #3799 (closes #3799)

I followed the Developers-HOW-TO and tested this using:
`make arch-x64-6.1 arch-aarch64-6.1 arch-armada38x-6.1`
I only tested installing `arch-armada38x-6.1` as that's the device I own.

Tasks done:
- [x] create `mk/spksrc.rust.mk` top build rust projects
- [x] define RUST_TRIPLE based on ARCH
- [x] download rust targets on demand
- [x] define CARGO_HOME
- validate `all-supported` targets (first rust build needs enhanced validation)
  - [x] x64
  - [x] armada370, armada38x
  - [x] aarch64
  - [ ] qoriq
  - [ ] evansport
  - [ ] ARMv5 (88f621)

Archs that are not supported (with `make all-supported`)
- hi3535 
- ppc853x

PR #3884, opened 2020-02-15
Add generic arch for ARMv7 and update go packages

hgy59:
_Motivation:_  As go build depends on host go only, it is possible to build generic targets for armv7 archs
_Linked issues:_  #3880, #3386

### Checklist

### Remarks
- add arch-armv7 based on armada38x
- update common Makefile with GENERIC_ARCHS to optionally create generic armv7 targets
- update go to version 1.13.8
- update syncthing to version 1.3.4 (includes and closes #3880)
- update minio to latests version
- add dedicated diyspk for fritzctl
- add go build Makefile to implement common compile and install targets
- add go-env Makefile to configure go build (to be used from cross and spk modules)
- use generic armv7 targets in go based packages: fritzctl, minio, syncthing, dnscrypt-proxy, node-exporter


hgy59:
A similar generic approach may be used for rust builds (#3883)

cross/dnscrypt-proxy/Makefile
```diff
@@ -1,64 +1,29 @@
...
+DEPENDS = native/go
+UNSUPPORTED_ARCHS = $(PPC_ARCHES)
+GENERIC_ARCHS = ARM7
 
 HOMEPAGE = https://github.com/jedisct1/dnscrypt-proxy
 COMMENT  = A flexible DNS proxy, with support for modern encrypted DNS protocols such as DNSCrypt v2 and DNS-over-HTTP/2.
 LICENSE  = ISC
 
-DEPENDS = native/go
+POST_INSTALL_TARGET = dnscrypt_extra_install
```
ymartin59: I find no "extra" here... prefer to name it `dnscrypt_post_install`

mk/spksrc.cross-go.mk
```diff
@@ -0,0 +1,162 @@
...
+	@for depend in $(DEPENDS) ; \
+	do                          \
+	  $(MAKE) WORK_DIR=$(WORK_DIR) --no-print-directory -C ../../$$depend cat_PLIST ; \
+	done
+	@if [ -f PLIST ] ; \
+	then \
+	  $(PLIST_TRANSFORM) PLIST ; \
+	else \
+	  $(MSG) "No PLIST for $(NAME)" >&2; \
+	fi
```
ymartin59: Sounds to me that section has been copy-paste... Probably a spksrc.plist.mk to be included would be relevant

mk/spksrc.cross-go.mk
```diff
@@ -0,0 +1,162 @@
...
+	do \
+	  $(MAKE) --no-print-directory -C ../../$$depend dependency-tree ; \
+	done
+
+.PHONY: all-archs
+all-archs: $(addprefix arch-,$(AVAILABLE_ARCHS))
+
+arch-%:
+	@$(MSG) Building package for arch $*
+	-@MAKEFLAGS= $(MAKE) ARCH=$(basename $(subst -,.,$(basename $(subst .,,$*)))) TCVERSION=$(if $(findstring $*,$(basename $(subst -,.,$(basename $(subst .,,$*))))),$(DEFAULT_TC),$(notdir $(subst -,/,$*)))
```
ymartin59: Same remark here... common code should be collected in dedicated makefile if possible and include when needed.

native/go/Makefile
```diff
@@ -12,14 +12,14 @@ COMMENT    = Go is an open source programming language that makes it easy to bui
 LICENSE    =
 
 CONFIGURE_TARGET = nop
-COMPILE_TARGET = myComp
+COMPILE_TARGET = compile-native-go
```
ymartin59: Please replace dashes by underscores for consistency

hgy59:
back to WIP, node-exporter doesn't build anymore...

mk/spksrc.cross-go.mk
hgy59: Finilly I generalized dependency-tree and digests rules

ymartin59:
Hello @hgy59 May you please rebase to adapt to latest minio version (I merged PR without noticing impact here)

hgy59:
@ymartin59 I resolved the conflicts, but need to rebuild and test.

publicarray:
I just compiled dnscrypt-proxy, works great!

Just some feedback:
Why not upgrade to go 1.14.1?

toolchains/syno-armv7-6.1/Makefile
```diff
@@ -0,0 +1,29 @@
+TC_NAME = syno-armv7
+
+# generic toolchain for arm7hf
+TC_ARCH = armada370 armada375 armada38x alpine armadaxp comcerto2k monaco
```
ymartin59, line 4: @hgy59  alpine4k is lacking from my point of view:
- https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model
- https://github.com/SynoCommunity/spksrc/blob/master/toolchains/syno-alpine-6.1/Makefile#L3

ymartin59:
@hgy59 Marvelous. Build is fast and number of published packages have been reduced. I love it !

acolomb:
Why didn't you go for the current Syncthing v1.4.0?

That one would be kind of important since it changes internal database formats and could lead to trouble when Syncthing auto-updates from this SPK version, then gets reinstalled for example, so v1.3.4 sees the database that was already migrated to v1.4 format.  I was about to file a PR for v1.4.0 myself, but after seeing your comment in #3880, I assumed it's already being worked on here.

Unfortunately there wasn't any feedback about [my questions](https://github.com/SynoCommunity/spksrc/pull/3880#issuecomment-585092391) regarding the (incompatible) upstream Syncthing SPK either.  This might become a recurring situation as things currently stand.

acolomb:
Scratch that, there was already a separate PR for Syncthing v1.4.0: #3931 

Please consider upgrading soon.

Otiel:
> Why didn't you go for the current Syncthing v1.4.0?
> 
> That one would be kind of important since it changes internal database formats and could lead to trouble when Syncthing auto-updates from this SPK version, then gets reinstalled for example, so v1.3.4 sees the database that was already migrated to v1.4 format.

Yep. 😞 
I went a bit too fast and upgraded my SynoCommunity-Syncthing package without first checking the commit details - didn't see that it would revert Syncthing to v1.3.4, while my Syncthing was already to v1.4.0.

Now the Synology package seems to crash at start. How can I upgrade Syncthing to v1.4.0 again?

acolomb:
See also #3941.

Kaelber:
I doesnt see this discussion, so I repeat. I´m a user of different Synologys and in the last years I use your Syncthing package. Up to now your packages running very good.

The current version of Syncthing is v1.4.0 and tomorrow v1.4.1 will official release. Until v1.4.0 a new database layout is implemented and the database needs a migration from versions <1.4.0, so also from v1.2.2 or v1.3.4. The new database layout could be a confusion, if a user use the inside updater function and update the system files from the old v1.2.2 to the newest one, today v1.4.1-rc.3. I also doe that.

Now you update your package from v1.2.2 to v1.3.4 and messed with the above facts, because a re-migration of the database is not possible and the package crashes. Next is, why you did not go straight to v1.4.0 is also not clear to me, because of course many of the updater are surely now at least on v1.4.0 or even on v1.4.1-rc.3. The database schema no longer fits. Next is now, the whole package no longer runs after the update.

In any case, you not paying attention. I must therefore deleted the SC package.

acolomb:
I can understand your frustration, this is pretty much the worst possible outcome.

@ymartin59 please upgrade to at least version v1.4.0 as soon as possible, and withdraw the current package from downloads in the meantime.

acolomb:
I've rebased #3931 onto master and am currently making test builds. Will make them available ASAP as a stop-gap solution.

@Kaelber don't touch your package data files in the meantime, it should work again without intervention when you install an SPK of version v1.4.0.

Kaelber:
I urgently needed a solution, so I installed the version of Kastelo (Syncthing-SPK from Syncthing for Synology). However, I will test when the time comes. 

For you Andre, I´m "Andy" in the Syncthing Forum, but it was not possible to choose this name here.

ymartin59:
@Kaelber @acolomb This PR is closed and not related to syncthing. Right? So please report a specific issue in a dedicated issue as a "new issue" in spksrc/issues...
Do you know how many mail notifications from GitHub I receive per day? You are really lucky I catch that one!

acolomb:
@Kaelber Luckily the Kastelo and spksrc versions don't conflict in this case, just be careful in case you are trying to share the config and DB among the two installations.

Unfortunately my test builds are not going well, hitting some unexpected errors. Still investigating. Let's continue discussion in your issue #3941.

PR #3873, opened 2020-01-28
FFMPEG 4.2.2: Update, fixes and cleanup

th0ma7:
_Motivations:_  Previous patchset contained many unneded features & limitations now fixed by breaking into individual patches every synology changesets.  As well, porting some of the other options used by Synology in their default build.

_Linked issues:_   #3716, #3861, #3862, #3868 
  * ffmpeg conversion from FLAC to AAC not working #3861 - Fixed & Closed
  * Cannot convert .mp4 to .jpg #3862 - Fixed & Closed
  * dts wav failed due to `-write_info_tag` option #3716 - Presumed fixed & Closed
  * TVHeadend/Surveillance Station/DSCam #3868 - Fixed & Closed
  * TVheadend 4.3 for cedarview #3863 - **NOT Related**, Fixed in PR #3876
  * Comskip - codec problems #3878 - Fixed & Closed

### Changes
* Major clean-up of Synology patches against previous versions to cherry-pick interesting bits
* Update to latest stable
* Activation of neon acceleration in libswscale (should "boosts" transcoding performance on Alpine platform)
* Groundwork for `_smd` acceleration enablement in the future for `evansport` arch
* evansport specific platform compile time cpu arch setting
* Enablement of opencore-amr codec
* Update libva-utils to v2.6 + fix https://github.com/intel/libva-utils/pull/185
* `-O3` compile-time optimization
* Update libvpx from v1.6.1 to v1.8.2 and enable high bit depth

_Related Work:_
* `apparmor` limitations: #3847
* Automated VideoStation fixes: https://github.com/SynoCommunity/spksrc/pull/3828

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

### IMPORTANT NOTES
* Due to the amount of patches this PR requires #3896 as dependancy
* Publishing of this updated package will require publishing new tvheadend packages as well in order to fix #3868 and other potential limitations due to previous version of the synology patchset applied.
* Contains groundwork in preparation for `evansport` native hardware acceleration.  The enablement of this feature is not yet in a readiness state to be applied.  Thus this pull request contains notes for future work related to `evansport` native hardware acceleration using built-in PowerVR GPU.

PR #3896, opened 2020-02-24
Allow arch-specific level patch to mk build

th0ma7:
_Motivation:_  Allow arch specific patches to be applied automatically.
_Linked issues:_  Relates to multi-patches on ffmpeg where some of them are arch specific

### Checklist
@hgy59 : To be sure about existing packages I propose to test the following:
- [x] Buildrule make all-supported completed successfully for `mono` (has patch in native module)
- [x] Buildrule make all-supported completed successfully for `redis` (regular patches)
- [x] Buildrule make all-supported completed successfully for `minio` (without any patch file)
- [x] Buildrule make all-supported completed successfully for `ffmpeg` #3873 (regular + subarch patches)
- [x] Buildrule make all-supported completed successfully for `tvheadend` on top of #3873

### DIFF
--- a/mk/spksrc.patch.mk
+++ b/mk/spksrc.patch.mk
@@ -17,17 +17,17 @@ endif
 
 ifeq ($(strip $(PATCHES)),)
 ifeq ($(findstring $(ARCH),$(ARM5_ARCHES)),$(ARCH))
-PATCHES = $(sort $(wildcard patches/*.patch patches/[aA][rR][mM][vV]?5/*.patch))
+PATCHES = $(sort $(wildcard patches/*.patch patches/[aA][rR][mM]5/*.patch patches/[aA][rR][mM][vV]5/*.patch))
 else ifeq ($(findstring $(ARCH),$(ARM7_ARCHES)),$(ARCH))
-PATCHES = $(sort $(wildcard patches/*.patch patches/[aA][rR][mM][vV]?7/*.patch))
+PATCHES = $(sort $(wildcard patches/*.patch patches/[aA][rR][mM]7/*.patch patches/[aA][rR][mM][vV]7/*.patch))
 else ifeq ($(findstring $(ARCH),$(ARM8_ARCHES)),$(ARCH))
-PATCHES = $(sort $(wildcard patches/*.patch patches/[aA][rR][mM][vV]?8/*.patch))
+PATCHES = $(sort $(wildcard patches/*.patch patches/[aA][rR][mM]8/*.patch patches/[aA][rR][mM][vV]8/*.patch))
 else ifeq ($(findstring $(ARCH),$(PPC_ARCHES)),$(ARCH))
 PATCHES = $(sort $(wildcard patches/*.patch patches/[pP][pP][cC]/*.patch))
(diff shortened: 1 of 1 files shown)


th0ma7:
I need to rework my regex for arm & x64 as no shell compliant.

hgy59:
Just wonder, why use regex? As we have x64_ARCHES etc. we could define the specific patch folders must match exactly the prefix of `_ARCHES` (i.e. x64, x86, ARM5, ARM7, ARM8, PPC). Or do I miss the background?

th0ma7:
@hgy59 agreed and coffee must have been missing this morning.  I guess it's because everytime I look for ARM specific searches it always refers to ARMvX and same for x86_64.  On the other hand, the simpler the better.

Additional suggestion I just pushed is to dig this a little further in order to have another level for sub-arch specific patches.  I currently have a case where I have a patch to apply only on `alpine`, another on `qoriq` and a last one that I really absolutely only want applied on `evansport`.  I guess a for loop against each $(ARCH) would be more efficient than the quick patch proposal I just uploaded, but this is for discussion...

hgy59:
Is it worth to add such specific patching to the framework?
You always have custom PATCH_TARGET (incl. PRE_ and POST_) available in the package Makefile.
This way you can override the whole patch process or add specific steps.

And you have a variable `PATCHES` that can be used to define specific patchfiles to apply (e.g. depending on arch...)

th0ma7:
I tried the `PATCHES` variable but with already a few patches to deal with it really become cumbersome to add them all individually for a single extra.
I haven't looked into `PATCH_TARGET`, any doc on it so I have a quick look at that option?  I've looked into the code and haven't quite understood that.
Still, something automated eliminate the extra oops and special cases, additional code and would simplify things.

th0ma7:
Found a much simpler approach using nested `foreach`... 
Does all I needed by sorting all patches from:
* `patches/*.patch`
* `patches/arch/*.patch`
* `patches/subarch/*.patch`

th0ma7:
@hgy59 thnx!  And you are absolutely right!

th0ma7:
@hgy59 as this is framework related does it require additional approvals before merging or can I proceed as is?

hgy59:
@th0ma7 I suppose you have tested this with the tvheadend package and the ppc specific patches. 

To be shure about existing packages I propose to test the following:
- [ ] Buildrule make all-supported completed successfully for tvheadend 
- [ ] Buildrule make all-supported completed successfully for mono (has patch in native module)
- [ ] Buildrule make all-supported completed successfully for e.g. redis (regular patches)
- [ ] Buildrule make all-supported completed successfully for e.g. minio (without any patch file)

Feel free to add this list to the PR Checklist

th0ma7:
@hgy59 , theses are good set of tests.
I've added the new PR for ffmpeg as there are subarch patches in it.

It allowed me to find an issue with `x64` arch where there are multiple matches due to other subarches containing `x64` string.  The usage of `filter` instead of `findstring` allowed to solve that.
Test-builds undergoing.

th0ma7:
@hgy59 I believe it's now ready for merging.

PR #3908, opened 2020-03-04
framework: drop ppc853x from supported arch

th0ma7:
_Motivation:_  `ppc853x` is not supported by DSM 6.x and should be dropped off from supported list.
_Linked issues:_  Getting harder to provide proper support or even functional packages for this specific arch & DSM version due to unsupported compiler.

### Checklist

This proposal is for discussion.

### DIFF
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -27,7 +27,7 @@ AVAILABLE_TCS = $(notdir $(wildcard ../../toolchains/syno-*))
 AVAILABLE_ARCHS = $(notdir $(subst syno-,/,$(AVAILABLE_TCS)))
 
 # Toolchain filters
-SUPPORTED_ARCHS = $(sort $(filter-out powerpc% ppc824% ppc854x%, $(AVAILABLE_ARCHS)))
+SUPPORTED_ARCHS = $(sort $(filter-out powerpc% ppc824% ppc853x% ppc854x%, $(AVAILABLE_ARCHS)))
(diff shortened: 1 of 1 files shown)


hgy59:
@th0ma7 I like the ppx853x and still have one in production and get security updates from time to time (and my DS210+ has more power than the DS115j even it has less memory).
But I propose to drop the support for TC-Version < 6 on the master branch in the near future.

My suggestions:
- make a branch of the current master and name it like `dsm5` so nothing gets lost that is available today.
- cleanup the master by removing all toolchains < 6
- on the master remove all dsm 5 -> 6 migration code (installer, ...)
- may be we will need migration code for dsm 6 -> 7 updates (and nobody should be forced to care about dsm 5 anymore)
- on the master, packages built for `noarch` may still be available for dsm 5 (required DSM in the INFO files of the packages should be kept as low as possible) (noarch: pure python or php)
- the existing dsm 5 packages in the spk repo should stay there (or are removed on security issues).
- for special needs, dsm 5 packages can still be updated and built on the dsm5 branch and published to the spk repo on demand.
- on the master, we should remove packages that never got dsm 6 support.
- unfortunatly we have no further information about the coming dsm 7. May be that some archs can be removed on the master branch too.


_and I would like to drop python 2.7 on the master, but this is OT here_

th0ma7:
Then I would add to the list (additional food for thoughts)
- update master branch to latest 6.2.2
- create a dsm6 branch and keep migrate upstream to dsm 7.x dev. (when ready)

PR #3928, opened 2020-03-21
fix: python3 virtualenv dependencies

fastcat:
_Motivation:_  `virtualenv` in `python3` is not working due to missing dependencies
_Linked issues:_
* Needed for #3916, probably other python3 packages too such as `borgbackup`

### Checklist


yuroller:
I confirm that this pull request is required for borgbackup to work with the updated python3 package or else it fails the during installation (see logs in /tmp).

PR #3939, opened 2020-04-04
Compile and distribute immutables along with Python3

nbarrientos:
Hi,

Not sure I'll be able to convince you that this is a good idea but it's worth a shot. Here I go.

I install [Flexget ](https://flexget.com/) on my Synology using a Python virtual environment. Unfortunately for me, recent versions of Flexget require [`immutables`](https://github.com/MagicStack/immutables) (via `contextvars`) which has C extensions. As I don't have any toolchain installed on the Synology and there are no precompiled wheels for `aaarch64` when it's time to install this dependency via `pip install` everything fails for obvious reasons.

Having this library distributed by Python3 would make my life much easier. Creating a symlink from the Python virtual environment to `@appstore/python3/lib/python3.6/site-packages/immutables` is enough to make `pip` happy and get the latest Flexget installed. Also, this might be useful is somebody gives a push to the Synocommunity Flexget package in the future.

I'd totally understand if you didn't want this here but please don't delete the PR, just close it. It might be useful for my future me as a recipe for whenever I have to build Python3 again. I guess it's the price I have to pay for not having bought an Intel Synology where I could just run my stuff in Docker containers and forget about all this 😁.

The patch has been tested with a build for `arch-rtd1296`.

https://cern.ch/nacho/srcpkg/python3/

Thanks for your time 🤞

### DIFF
--- a/spk/python3/src/requirements.txt
+++ b/spk/python3/src/requirements.txt
@@ -5,6 +5,7 @@
 #Pillow==5.3.0
 #pycurl==7.43.0.2
 pyyaml==5.3
+immutables==0.11
 
 ## Cross-compiled via spksrc.wheel.mk
 markupsafe==1.1.1
--- a/spk/python3/src/requirements.txt
+++ b/spk/python3/src/requirements.txt
@@ -4,8 +4,8 @@
 #pycrypto==2.6.1
 #Pillow==5.3.0
 #pycurl==7.43.0.2
-pyyaml==5.3
-immutables==0.11
+#pyyaml==5.3
+#immutables==0.11
 
 ## Cross-compiled via spksrc.wheel.mk
 markupsafe==1.1.1
--- a/cross/immutables/Makefile
+++ b/cross/immutables/Makefile
@@ -9,4 +9,4 @@ HOMEPAGE = https://github.com/MagicStack/immutables
 COMMENT  = A high-performance immutable mapping type for Python.
 LICENSE  = Apache-2.0
 
-include ../../mk/spksrc.python-module.mk
+include ../../mk/spksrc.python-wheel.mk
--- a/cross/immutables/Makefile
+++ b/cross/immutables/Makefile
@@ -1,0 +1,12 @@
+PKG_NAME = immutables
+PKG_VERS = 0.11
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://files.pythonhosted.org/packages/source/i/$(PKG_NAME)
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+HOMEPAGE = https://github.com/MagicStack/immutables
+COMMENT  = A high-performance immutable mapping type for Python.
+LICENSE  = Apache-2.0
+
+include ../../mk/spksrc.python-module.mk
--- a/cross/immutables/PLIST
+++ b/cross/immutables/PLIST
@@ -1,0 +1,1 @@
+lib:@PYTHON_SITE_PACKAGES@/immutables/_map*.so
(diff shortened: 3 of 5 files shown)


ymartin59:
Your proposal sounds like a good idea. Synocommunity Flexget package requires an update itself, and so this dependency is probably a requirement for update too.

spk/python3/src/requirements.txt
```diff
@@ -5,6 +5,7 @@
 #Pillow==5.3.0
 #pycurl==7.43.0.2
 pyyaml==5.3
+immutables==0.11
```
hgy59: immutables (and pyyaml) should be deactivated with # here, as these are included by compiled wheels.
as written on line 2 these are documented here for referency only

hgy59:
@ymartin59 I fixed this to integrate immutables as compliled wheel.
But I suppose there is a general issue with compiled wheels on python3.

I analyzed the compiled wheels of all-supported archs on the PR and all arch specific compiled shared objects (*.so) have x86_64 in the name. Even these files have different size (and I assume the correct arch) they have a wrong name (IMO).

example:
the file `share/wheelhouse/immutables-0.11-cp36-none-any.whl` containes `/immutables/_map.cpython-36m-x86_64-linux-gnu.so` for all compiled archs.

I can analyze self compiled packages only (synocommunity.com is not reachable) but as far as I see this is the same with the working version of Python3 (3.6.8-9).

So please @nbarrientos and @ymartin59 (or any one else) can you test whether Python3 and immutable of this PR work with non intel archs (other than x64 and x86)?

hgy59:
@ymartin59 as Python 3.6.10 is not yet published I didn't increment SPK_REV.

nbarrientos:
Thanks for the review.

> So please @nbarrientos and @ymartin59 (or any one else) can you test whether Python3 and immutable of this PR work with non intel archs (other than x64 and x86)?

Will generate a build with your patch for `rtd1296` an report back.

nbarrientos:
> Will generate a build with your patch for rtd1296 an report back.

WFM.

```
foo@bar:/somewhere/@appstore/python3$ uname -m
aarch64
foo@bar:/somewhere/@appstore/python3$ ./bin/python3
Python 3.6.10 (default, Apr  5 2020, 20:07:11) 
[GCC 4.9.4 20150629 (prerelease)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import immutables
>>> 
```

Indeed the wrong arch is in the name of shared object but it seems to be loaded just fine.

```
# file immutables/_map.cpython-36m-x86_64-linux-gnu.so 
immutables/_map.cpython-36m-x86_64-linux-gnu.so: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

spk/python3/src/requirements.txt
nbarrientos: Sure, was not sure about this :)

ymartin59:
@nbarrientos Thanks for testing. This invalid file naming for shared objects is a concerned as far as we have reports wheels do not load properly when trying to run application packages over it.

nbarrientos:
Thanks for merging this! Just a follow-up question for my own education. When can we expect to have an official build of Python 3.6.10-11 for `rtd1296` (`aarch64`)? I'm currently using my own build but I'd like to align my system to the packages in the official Synocommunity repos. Thanks again!

ymartin59:
@nbarrientos What applications are you running over Python 3.6.10. For details please refer #3944

nbarrientos:
I see, thanks for the info @ymartin59 :)

I only run Flexget but installed standalone using a Python venv, not via Synocommunity packages. Flexget is supported on 3.6 and 3.7 and it's "meant to work" using 3.8. Happy help testing in the future if needed. Just ping me.

Issue #3949, opened 2020-04-11
Add 6.2.2 toolchains


ymartin59:
Including "purley" architecture as x64 alias

ymartin59:
As far as I understood, you have already done part of the job

PR #3952, opened 2020-04-12
Add initial purley TC 6.2.2

hgy59:
_Motivation:_  The intel purley arch is not supported yet (introduced by model FS6400)
_Linked issues:_  #3949

### Remarks
- add purley to generic x64 archs
- reorder ARCHS_DUPES similar to x64 TC_ARCH
- add firmware to INFO too, when TC_OS_MIN_VER is specified


hgy59:
@ymartin59 I suppose you expect to add purley TC 6.2.2 only and not the toolchains of other archs (to stay on TC 6.1 as default for `make all-supported`)

Issue #3944, opened 2020-04-10
Fix Python 3 wheels cross-compilation


ymartin59:
At the moment application packages publication are stuck because Python 3.8 cross-compiled wheels are badly generated with "x64" architecture in their shared objects (so) file names, which prevent their loading at runtime.

Example: file `share/wheelhouse/immutables-0.11-cp36-none-any.whl` containes `/immutables/_map.cpython-36m-x86_64-linux-gnu.so` for all compiled archs.

```
# file immutables/_map.cpython-36m-x86_64-linux-gnu.so 
immutables/_map.cpython-36m-x86_64-linux-gnu.so: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

Some application installation just fail because of this invalid file naming. It is expected wheels file to match target runtime architecture like `_map.cpython-36m-aarch64-linux-gnueabi.so` according to toolchain `TC_TARGET` value.

Pending packages because of this:
- borgbackup #3934 #3644 #3334 
- homeassistant #3740 #3643 #3573 #3584
- rdiff-backup #3916
- deluge (move to python3)
- flexget (move to python3) 
- duplicity (move to python3) #3823

And many other Python applications to come...

@roleoroleo and @smaarn  did progress with #3726 #3731 #3739 #3833
But there is still trouble according 

Related issues: #3737

Objectives: fix current 3.6.10 to be published, and prepare for next version, either 3.7 or directly 3.8

smaarn:
@ymartin59 open question: is there any specific reason to skipping the intermediary 3.7 version ?
I mean, in the "master" branch only 3.6 version seems to be present. On my side I'm working on having a 3.7 version running (turns out that bazarr is upgrading its requirements to 3.7 ...) and I was wondering if there was a specific reason which would imply the dismissal of that one.

ymartin59:
@smaarn Simply because it is to much work. Publishing 3.6.10 is still in pending/failed state because of that issue. As far as I know, all applications' package list is expected to run on Python 3.8

smaarn:
@ymartin59 ok I was misled by the title of this issue to be honest !

I'm having a look to see how this works.

smaarn:
Ok after a "little" of digging:

- The file name format is actually determined by the `EXT_FILE` parameter computed when building the python executing the build. So here it's the native guy which, obviously, depends on the ABIFLAGS.
[...]

PR #3953, opened 2020-04-12
[python3] Upgrade to 3.7.7 + homeassistant build fixes

### LINKED ISSUE #3737: [Package Request] Python 3.7+
Name: Python 3.7

Description: Python 3.7.0 is the newest major release of the Python language, and it contains many new features and optimizations.

Website: https://www.python.org/

Software documentation: https://www.python.org/doc/

Source code: https://www.python.org/downloads/source/

License: https://docs.python.org/3/license.html

Many new modules were added with python 3.7+. Releasing an updated version would benefit for the entire community.

smaarn:
_Motivation:_  Latest version of bazarr (0.8.4.3) now requires Python 3.7 at least
_Linked issues:_  Closes #3737 and #3944 (hopefully) 

### Checklist


smaarn:
Still an issue with the `_ctypes` extension.

spk/python3/Makefile
```diff
@@ -1,7 +1,7 @@
 SPK_NAME = python3
-SPK_SHORT_VERS = 3.6
-SPK_VERS = $(SPK_SHORT_VERS).10
-SPK_REV = 11
+SPK_SHORT_VERS = 3.7
+SPK_VERS = $(SPK_SHORT_VERS).7
+SPK_REV = 1
```
hgy59, line 4: SPK_NAME is still python3, so we must increment and use SPK_REV = 12

cross/libffi/Makefile
```diff
@@ -14,6 +14,10 @@ LICENSE  = libffi BSD-like
 GNU_CONFIGURE = 1
 PRE_CONFIGURE_TARGET = libffi_pre_configure
 
+# Otherwise the generated pkg-config resources would basically contain invalid information
+# On file locations
+CONFIGURE_ARGS = --prefix=$(INSTALL_PREFIX)
+
```
ymartin59, line 20: In my opinion, this line is redundant after `GNU_CONFIGURE=1` if you look at configure Makefile targets: https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.configure.mk#L40

spk/python3/Makefile
smaarn: Thanks for the tip. I thought the revision was "on top" of the short version not parallel.

cross/libffi/Makefile
smaarn: I'll check what $(TC_CONFIGURE_ARGS) are, just to be sure.

smaarn: Hum that's odd. If I remove this line I get as contents the actual deployed value (e.g. for python3 build `/var/packages/python3/target`).
Here my goal was to have `<path to spksrc>/spk/python3/work-<arch>/var/packages/python3/target`.
Maybe it would be better for me to fix this in the cross/python3 build then ?

ymartin59: GNU configure prefix is used to locate binary and share objects at runtime. It cannot be compilation destination directory. If python build chain is not able to locate library, it is required to provide additional options at build so that it finds it.

smaarn: OK I'll remove this and try to fix the issue I had with libffi another way.

cross/libffi/patches/includesdir.patch
```diff
@@ -1,10 +1,25 @@
-diff -ur include-old/Makefile.am include/Makefile.am
---- include-old/Makefile.am	2011-08-22 20:38:12.000000000 +0200
-+++ include/Makefile.am	2011-12-16 23:15:45.000000000 +0100
-@@ -5,5 +5,5 @@
- DISTCLEANFILES=ffitarget.h
- EXTRA_DIST=ffi.h.in ffi_common.h
+From 982b89c01aca99c7bc229914fc1521f96930919b Mon Sep 17 00:00:00 2001
```
smaarn, line 1: To be honest I could move this outside of this PR since it turns out to be unrelated and not required.

smaarn:
I generated an initial version here: https://github.com/smaarn/spksrc/releases/tag/python-3.7.7-12-rc1

The "only" thing which remains is this thing about customizing (when cross compiling obviously) the extension enforced at global python level...

smaarn:
New "release" available here: https://github.com/smaarn/spksrc/releases/tag/python-3.7.7-12-rc2

smaarn:
Latest binaries generated here: https://github.com/smaarn/spksrc/releases/tag/python-3.7.7-12-rc4

cross/python3/Makefile
```diff
@@ -51,6 +55,8 @@ PATH := $(WORK_DIR)/../../../native/$(PKG_NAME)/work-native/install/usr/local/bi
 LD_LIBRARY_PATH := $(WORK_DIR)/../../../native/$(PKG_NAME)/work-native/install/usr/local/lib:$(LD_LIBRARY_PATH)
 export PATH LD_LIBRARY_PATH
 
+ENV += LIBFFI_INCLUDEDIR=$(STAGING_INSTALL_PREFIX)/include/
```
smaarn, line 58: I ensured by patching the configure mechanism of the python3 cross compiled package that the LIBFFI_INCLUDEDIR variable can actually be overridden explicitely using environment variable.
And here I ensure that the environment variable is appropriately defined.

cross/python3/patches/005-configure.ac-libffi.patch
```diff
@@ -0,0 +1,37 @@
+Ensures that the LIBFFI_INCLUDEDIR environment variable can be forced
+
+*** configure.ac.orig	2020-04-13 14:50:17.476803111 +0000
+--- configure.ac	2020-04-13 14:52:29.859184113 +0000
+***************
+*** 3011,3022 ****
+      with_system_ffi="yes"
+  fi
+  
+! if test "$with_system_ffi" = "yes" && test -n "$PKG_CONFIG"; then
```
smaarn, line 10: The default python mechanism is to use pkg configuration to locale libffi binaries.
The only issue here is that libffi binaries are installed with information which actually corresponds to the target installation not the build installation.
So I basically needed to override it using environment variable.

Safihre:
Maybe a stupid question, but why not directly go for 3.8?

smaarn:
> Maybe a stupid question, but why not directly go for 3.8?

Not a stupid question. It's more a matter of lack of personal expertise to do an upgrade from a non operational 3.6 to 3.8 ;)

Safihre:
Might be worth a try, you might be very good at it 💯👍

Safihre:
Is there an update for this one? :)

thezoggy:
additional comment to try and push this along as well.. 
it's widely known that python 2.7 was faster than python3 and it wasn't until python 3.6 before they really started to get improvements in to put it back on par with 2.7. Then py3.7+ finally we saw it starting to be fundamentally faster due to all the optimizations. So with a lot of syno being cpu limited.. wouldn't it just make sense to move to 'faster' python3.7 (or 3.8) if there is no harm? :)

Additionally there is benefit in this day and age in keeping up with releases to have security fixes.. which 3.5 does not get.

smaarn:
Rebased at head.

Safihre:
@hgy59 Hmm, shouldn't a pull request also trigger a full build?

Safihre:
I am not sure I can give meaningful feedback on the PR's content, maybe @ymartin59 can?
Can't wait for this update 😃

ymartin59:
I have checkout PR branch and I have planned to validate result with applications (beets, borgbackup) on x64 and evansport

hgy59:
> @hgy59 Hmm, shouldn't a pull request also trigger a full build?

Did you push your changes? The last I can see here are from April 15th.

smaarn:
> @Safihre
> 
> > @hgy59 Hmm, shouldn't a pull request also trigger a full build?
> 
> May be @smaarn did not push the changes? The last commit I can see here is from April 15th.

Actually there were issues with the github actions (just did a PR to fix it)

smaarn:
There seems to be issue with homeassistant build (yaml header file it can't find... I'm trying a fix in a separate branch to see if I can - at least - make it build)

ekanters71:
Hi, would it be possible to add the avoton arch as I have a DS415+?

smaarn:
homeassistant should now be building. Needed to rework the way python modules were built + reworked a little bit homeassistant packaging.

native/python3/Makefile
```diff
@@ -29,10 +29,11 @@ PIP_NATIVE = $(WORK_DIR)/../../../native/$(PKG_NAME)/work-native/install/usr/loc
 
 .PHONY: nativepython3_post_install
 nativepython3_post_install: $(WORK_DIR)/python-native.mk
-	@$(MSG) Installing setuptools, pip and cffi
-	@$(RUN) wget https://bootstrap.pypa.io/ez_setup.py -O - | $(PYTHON)
+	@$(MSG) Installing pip, setuptools, cffi and cross env
 	@$(RUN) wget https://bootstrap.pypa.io/get-pip.py -O - | $(PYTHON)
+	@$(PIP) install "setuptools==44.0.0"
```
smaarn, line 34: I aligned it with the version under `cross/setuptools` instead of using `ez_setup` which is now deprecated

spk/homeassistant/src/requirements.txt
```diff
@@ -466,7 +466,8 @@ gTTS-token==1.1.1
 # homeassistant.components.frontend
 home-assistant-frontend==20180903.0
 user_agents		# discovered during first run
-ua_parser		# discovered during first run
+# discovered during first run
+ua_parser==0.8.0
```
smaarn, line 470: I was facing weird issues with the "latest" version (which is `0.10.0`).

smaarn:
last failure encountered related to building of psutil wheel under ppc853x which uses a switch the corresponding compiler doesn't support.

Safihre:
In python 2 package we include `cryptography` by default. Shouldn't we also do that for python 3? @ymartin59
https://github.com/SynoCommunity/spksrc/blob/master/spk/python/Makefile#L10

smaarn:
> In python 2 package we include `cryptography` by default. Shouldn't we also do that for python 3? @ymartin59 / @smaarn
> https://github.com/SynoCommunity/spksrc/blob/master/spk/python/Makefile#L10

Indeed that would seem fair enough. At this stage it shouldn't be too complicated to change in this PR but I'd say it's a judgement call here ! Just tell me.

Safihre:
@smaarn How did you get Homeassistant to build? 
It also has this `@PYTHON_SITE_PACKAGES@` in the `Makefile`, but this variable isn't defined anywhere as far as I can see.. This is also causing the build of SABnzbd to fail.

smaarn:
> @smaarn How did you get Homeassistant to build?
> It also has this `@PYTHON_SITE_PACKAGES@` in the `Makefile`, but this variable isn't defined anywhere as far as I can see.. This is also causing the build of SABnzbd to fail.

There were one issue related to missing yaml headers => cross/libyaml was added.
There were issues building bcrypt within the pip process => I added cross/bcrypt and ensured the corresponding resources were properly installed when setting up the virtual environment

Same was for cryptography basically.

cross/psutil/Makefile
```diff
@@ -0,0 +1,12 @@
+PKG_NAME = psutil
```
smaarn, line 1: I had to introduce the cross compiled version of psutil.
Indeed there was an issue which made the pip process fail building the corresponding wheel.

smaarn:
Latest binaries (computed by github actions) can be found here: https://github.com/smaarn/spksrc/releases/tag/python-3.7.7-12-rc5

Safihre:
I used your branch to build SABnzbd, and it worked! But only after I added the saw like that Homeassistant has:
```
PLIST_TRANSFORM = sed -e 's%@PYTHON_SITE_PACKAGES@%$(PYTHON_LIB_DIR)/site-packages%' -e 's%@PYTHON_VERSION@%$(SPK_SHORT_VERS)%'
```
Otherwise the `cffi` and `bcrypt` won't work. Isn't that strange? If this is required, why isn't it part of the `python3` package (either cross or spk) by default?

smaarn:
> Otherwise the `cffi` and `bcrypt` won't work. Isn't that strange? If this is required, why isn't it part of the `python3` package (either cross or spk) by default?

Sounds odd indeed. What do you mean by `won't work` ? You mean the wheels won't build ? Or you mean the packages which are available as cross compiled packages won't build ?

Safihre:
The package that lists `cffi` or `bcrypt` as `DEPENDS` will fail to compile in the last steps when it combines all the plist entries. Unless the package includes the`PLIST_TRANSFORM` that for example Homeassistant has.
It's not too relevant for this PR, was just hoping for the insight of @ymartin59 if he maybe knows how this `PLIST_TRANSFORM` started :)

smaarn:
> The package that lists `cffi` or `bcrypt` as `DEPENDS` will fail to compile in the last steps when it combines all the plist entries. Unless the package includes the`PLIST_TRANSFORM` that for example Homeassistant has.
> It's not too relevant for this PR, was just hoping for the insight of @ymartin59 if he maybe knows how this `PLIST_TRANSFORM` started :)

Actually it's something done in several other projects.

It can be fixed, I believe, by putting this declaration at the level of the cross compiled project if I'm not mistaken. That's what's being done by other cross compiled dependencies at least. Maybe that should be the way to go.

smaarn:
ok so now what remains are ffsync build failures (impacted because of sqlite build fixes...)

cross/python3/patches/006-configure-fix-autodetection-for-unused-result.patch
```diff
@@ -0,0 +1,14 @@
+Fixes the mechanism used to detect if the no-unused-result flag is supported
+
+--- configure.ac.orig
++++ configure.ac
+@@ -1630,7 +1630,7 @@ yes)
+     *)
+     AC_MSG_CHECKING(if we can turn off $CC unused result warning)
+      ac_save_cc="$CC"
+-     CC="$CC -Wunused-result -Werror"
++     CC="$CC -Wno-unused-result -Werror"
```
smaarn, line 10: For some reason the compiler used for ppc853x platform did support the `-Wunused-result` flag but not the `-Wno-unused-result` one. The thing is it was trying, here, to identify whether the latter was supported.

cross/bcrypt/Makefile
```diff
@@ -11,4 +11,6 @@ HOMEPAGE = https://github.com/pyca/bcrypt
 COMMENT  = Modern password hashing for your software and your servers
 LICENSE  = Apache 2.0
 
+PLIST_TRANSFORM = sed -e 's%@PYTHON_SITE_PACKAGES@%$(PYTHON_LIB_DIR)/site-packages%' -e 's%@PYTHON_VERSION@%$(SPK_SHORT_VERS)%'
```
Safihre, line 14: Thinking of this again I wonder if it shouldn't just be part of `cross/python3`?

smaarn: Actually, if anything, it should be part of some .mk import files I believe.

smaarn:
Corrected a typo which prevented the build from passing on ppc archs... :(

cross/bcrypt/Makefile
smaarn: Just checked and it cannot "just" be put in the `cross/python3` logic. At this level it "could" have been inserted in the auto-generated `python-cc.mk` file but this file is imported **after** the one which defines the default `PLIST_TRANSFORM` value (which is `mk/spksrc.plist.mk` from what I see).

If we are to go towards the `cross/python3` approach I guess the ordering of  https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.python-module.mk#L17-L20 and https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.python-wheel.mk#L17-L20 lines need to be inverted.

spk/python3/src/requirements.txt
```diff
@@ -9,7 +9,7 @@
 
 ## Cross-compiled via spksrc.wheel.mk
 markupsafe==1.1.1
-psutil==5.7.0
+#psutil==5.7.0
```
Safihre, line 12: Got my old DS112j up and running again. I installed your build and it seems to work.
First thing I noticed is that we include all the requirements for `cryptograpghy` in the `requirements.txt` (few lines below), but that we don't actually include it in the `spk/python3`. So let's add that back in, just like we do for `spk/python`.

Safihre: @smaarn Me again 😄 I wanted to share some work I have been doing: 
So, while trying for hours to get `spk/sabnzbd` to run, it kept failing on `cryptography` not being available. Inspecting the `package.tgz` it turns out that the `package.tgz\lib\python3.7\site-packages` did not contain any of the `cryptography` files. Even though the build log shows they are being created.
After digging I found that is because the `PLIST` for `cross/cryptography` [is empty](https://github.com/SynoCommunity/spksrc/blob/master/cross/cryptography/PLIST). This is not a problem if it is part of `python3` package, because there the `PLIST` contains just, well, _everything_. 
So either we update the `PLIST` of `cross/cryptography` or we just include it in `pyton3`.

smaarn: Working on it while also fixing the ppc853x architecture build issues.

hgy59: > Working on it while also fixing the ppc853x architecture build issues.

I just picked this up and solved it by moving the wheel build of rdiff-backup from spk/rdiff-backup/Makefile to cross/rdiff-backup (including all dependencies except BUILD_DEPENDS)
My intent was to add a patch for ppx853x and hoped to solve this with cross/rdiff-backup. But luckily the build succeeded with this approach without a patch (no unknown compiler option -Wno-unused-result).

hgy59:
@smaarn I fixed rdiff-backup cross compilation
I think something is wrong with cross compilation of wheels included via requirements.txt in spk/{package}/Makefile.
The package `rdiff-backup_armv7-6.1_2.0.3-3.spk` build with github action on 2020/06/03 18:45 there is a problem with the wheel rdiff_backup-2.0.3-cp37-none-any.whl. The tag in the WHEEL file is `Tag: cp37-cp37m-arm_unknown_linux_gnueabi` but the shared libraries in the wheel are named `C.cpython-37m-x86_64-linux-gnu.so` and `_librsync.cpython-37m-x86_64-linux-gnu.so`. 
I figured out, that only the name of the shared libraries containe the wrong arch, but the content is ELF for ARM.

The move of rdiff-backup wheel to cross/rdiff-backup/Makefile fixed this. The wheels built for cross/pyxattr and cross/pylibacl already have the corrent names in it. Therefore I suppose this is a general issue with python3 wheels. The pragmatic solution would be to build cross-modules for all cross compiled wheels.

smaarn:
@hgy59 I actually found a way to fix this in another wip branch. Unfortunately I’m away from my computer and will only be able to push the whole thing tonight (Paris time).

cross/python3/Makefile
```diff
@@ -82,12 +88,18 @@ python3_custom_install:
 python3_post_install: $(WORK_DIR)/python-cc.mk
 	mkdir -p $(PYTHON_LIB_CROSS)
 	cp -R $(HOSTPYTHON_LIB_NATIVE) $(PYTHON_LIB_CROSS)/../
+	@$(RUN) $(PYTHON_NATIVE) -m crossenv $(STAGING_INSTALL_PREFIX)/bin/python$(PKG_VERS_MAJOR).$(PKG_VERS_MINOR) $(WORK_DIR)/crossenv/
+	. $(WORK_DIR)/crossenv/bin/activate && $(RUN) wget https://bootstrap.pypa.io/get-pip.py -O - | build-python
```
smaarn, line 92: Aligned all of those to be in line with the native python build libraries

mk/spksrc.common.mk
```diff
@@ -17,7 +17,8 @@ RUN = cd $(WORK_DIR)/$(PKG_DIR) && env $(ENV)
 PIP ?= pip
 # Why ask for the same thing twice? Always cache downloads
 PIP_CACHE_OPT ?= --cache-dir $(PIP_DIR)
-PIP_WHEEL = $(PIP) wheel --no-binary :all: $(PIP_CACHE_OPT) --no-deps --requirement $(WORK_DIR)/wheelhouse/requirements.txt --wheel-dir $(WORK_DIR)/wheelhouse --build-dir $(WORK_DIR)/wheelbuild
+PIP_WHEEL_ARGS = wheel --no-binary :all: $(PIP_CACHE_OPT) --no-deps --requirement $(WORK_DIR)/wheelhouse/requirements.txt --wheel-dir $(WORK_DIR)/wheelhouse -b $(WORK_DIR)/wheelbuild
```
smaarn, line 20: It turns out that `--build-dir` is not a documented option of wheel (from what I saw in the cli documentation), switched to using the official `-b` one.

mk/spksrc.wheel.mk
```diff
@@ -62,7 +62,11 @@ build_wheel_target: $(PRE_WHEEL_TARGET)
 		$(foreach e,$(shell cat $(WORK_DIR)/python-cc.mk),$(eval $(e))) \
 		if [ ! -z "$(CROSS_COMPILE_WHEELS)" ] ; then \
 			$(MSG) "Force cross-compile" ; \
-			$(RUN) CFLAGS="$(CFLAGS) -I$(STAGING_INSTALL_PREFIX)/$(PYTHON_INC_DIR) $(WHEELS_CFLAGS)" LDFLAGS="$(LDFLAGS) $(WHEELS_LDFLAGS)" $(PIP_WHEEL) ; \
+			if [ -z "$(CROSSENV)" ]; then \
+				$(RUN) _PYTHON_HOST_PLATFORM="$(TC_TARGET)" CFLAGS="$(CFLAGS) -I$(STAGING_INSTALL_PREFIX)/$(PYTHON_INC_DIR) $(WHEELS_CFLAGS)" LDFLAGS="$(LDFLAGS) $(WHEELS_LDFLAGS)" $(PIP_WHEEL) ; \
+			else \
+				. $(CROSSENV) && $(RUN) _PYTHON_HOST_PLATFORM="$(TC_TARGET)" CFLAGS="$(CFLAGS) -I$(STAGING_INSTALL_PREFIX)/$(PYTHON_INC_DIR) $(WHEELS_CFLAGS)" LDFLAGS="$(LDFLAGS) $(WHEELS_LDFLAGS)" pip $(PIP_WHEEL_ARGS) --no-build-isolation ; \
```
smaarn, line 68: With python 3.7.7 build isolation is enforced which leads to invalid rebuild and reinstallations of `setuptools`, `wheel` and others when it shouldn't actually be done (because we already provide them).

spk/rdiff-backup/src/requirements.txt
```diff
@@ -1,3 +1,3 @@
 rdiff-backup==2.0.3
-pylibacl==0.5.4
-pyxattr==0.7.1
+pylibacl==0.5.4
```
smaarn, line 2: ok I can revert this change altogether ;)

spk/homeassistant/Makefile
```diff
@@ -1,14 +1,15 @@
...
-SPK_REV = 4
+SPK_REV = 5
 SPK_ICON = src/${SPK_NAME}.png
 
 BUILD_DEPENDS = cross/python3 cross/setuptools cross/pip cross/wheel
 DEPENDS = cross/cffi cross/nmap  # for python-libnmap
 DEPENDS += cross/gmp cross/mpfr cross/mpc  # for gmpy2
-DEPENDS += cross/bcrypt
+DEPENDS += cross/libyaml cross/bcrypt
+DEPENDS += cross/pip
```
smaarn, line 10: Could actually be reverted

smaarn:
I released binaries generated by pull_request job here: https://github.com/smaarn/spksrc/releases/tag/python-3.7.7-12-rc6

Safihre:
@ymartin59 / @smaarn It works! Tested on `88f6281` architecture.
I can start and import `cryptography` + it's sub-modules and other C-compiled stuff.
I was also able to run SABnzbd with it (been testing in a seperate branch). 💯

smaarn:
Rebased at head.

For what it's worth I don't know how to remove the tags "work in progress" and don't know if the "do-not-merge" flag is still relevant.

ymartin59:
@smaarn According to GitHub Actions, build process is in trouble

smaarn:
> @smaarn According to GitHub Actions, build process is in trouble

@ymartin59 Just saw this. There seems to be issue in building `cross/portaudio` which is used in the `spk/python` build. I'm trying to see if there is a simple workaround here (the issue is that it cannot fetch the source code for the library...)

Safihre:
@smaarn try a simple restart of the build, often that works :)

smaarn:
> @smaarn try a simple restart of the build, often that works :)

I'd be glad to be the underlying side (http://www.portaudio.com/) is still down so there would be no real point...

cross/python3/patches/006-configure-fix-autodetection-for-unused-result.patch
smaarn: No, actually this was not a trial.
The point of this process was to detect automatically detect if the `-Wno-unused-result` switch was supported.
It turns out, that, for unknown reason, the ppc853x architecture compiled did support the `-Wunused-result` switch but not the `-Wno-unused-result` one.
So here I changed the logic to use the `-Wno-unused-result` switch to detect if it was supported.
Obviously I could change it only for the given architecture, but I thought that it was kinda weird.

Obviously, tell me if you'd rather go for the architecture switch. (I saw that, indeed, you could create a specific folder for architecture specific options).

smaarn:
> > @smaarn try a simple restart of the build, often that works :)
> 
> I'd be glad to be the underlying side (http://www.portaudio.com/) is still down so there would be no real point...

ok the website is back online. Rebased the PR and it worked.

cross/python3/patches/006-configure-fix-autodetection-for-unused-result.patch
th0ma7: With a previous change I made to the toolchain you can then simply create a directory `patches/ppc853x` and drop the patch that is only specific to that `arch`.

smaarn:
@ymartin59 updated PR to use platform specific patches.

ymartin59:
borg is OK but borgmatic requires some attention
```
borgmatic --version
  File "/usr/local/bin/borgmatic", line 8, in <module>
    sys.exit(main())
  File "/volume1/@appstore/borgbackup/env/lib/python3.7/site-packages/borgmatic/commands/borgmatic.py", line 380, in main
    print(pkg_resources.require('borgmatic')[0].version)
  File "/volume1/@appstore/borgbackup/env/lib/python3.7/site-packages/pkg_resources/__init__.py", line 901, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/volume1/@appstore/borgbackup/env/lib/python3.7/site-packages/pkg_resources/__init__.py", line 792, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (ruamel.yaml 0.16.10 (/volume1/@appstore/python3/lib/python3.7/site-packages), Requirement.parse('ruamel.yaml<0.16.0,>0.15.0'), {'borgmatic'})
```

PR #3982, opened 2020-05-01
Add Github Actions

publicarray:
Automate Lint & Build on push to catch errors quickly. Other advantage is that people that fork the repo can easily create their own builds.

Thanks to @hgy59 for the dependency tree script: https://github.com/SynoCommunity/spksrc/pull/3854#issuecomment-573813380

### DIFF
--- a/.github/actions/build-push-pr.sh
+++ b/.github/actions/build-push-pr.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+cat <<EOF > local.mk
+PUBLISH_URL=https://api.synocommunity.com
+PUBLISH_API_KEY=
+MAINTAINER?=
+MAINTAINER_URL=
+DISTRIBUTOR=SynoCommunity
+DISTRIBUTOR_URL=https://synocommunity.com/
+REPORT_URL=https://github.com/SynoCommunity/spksrc/issues
+SUPPORT_URL=https://github.com/SynoCommunity/spksrc/issues
+DEFAULT_TC=6.1
+PARALLEL_MAKE=4
+EOF
+
+# echo "FILES: $FILES"
+
+# filter for changes made in the cross and spk directories
+GH_FILES=$(echo "$GH_FILES" | grep -oE "(spk.*)|(cross.*)")
+
+# create array of potential packages where files have changed
+GH_PACKAGES_ARR=()
+for GH_FILE in $GH_FILES
+do
+    # remove leading spk and cross from string
+    GH_FILE=${GH_FILE#spk/}
+    if [[ "$GH_FILE" == cross/* ]]; then
+        make dependency-tree
+        GH_FILE=${GH_FILE#cross/}
+    fi
+
+    # get package name / folder name
+    GH_PACKAGE=$(echo "$GH_FILE" | grep -oE "^[^\/]*")
+    # echo "PACKAGE: $PACKAGE"
+    GH_PACKAGES_ARR+=("$GH_PACKAGE")
+done
+
+# de-duplicate packages
+GH_PACKAGES=$(printf %s "${GH_PACKAGES_ARR[*]}" | tr ' ' '\n' | sort -u)
+
+if [ -z "$GH_PACKAGES" ]; then
+    echo "no package built. Empty PACKAGES var"
+    exit 0
+fi
+
+echo "===> PACKAGES to Build: $GH_PACKAGES"
+
+for GH_PACKAGE in $GH_PACKAGES
+do
+    # make sure that the package exists
+    if [ -d "spk/$GH_PACKAGE" ]; then
+        cd spk/"$GH_PACKAGE" && make "$GH_ARCH"
+    else
+        # must be from cross/
+        echo "$GH_PACKAGE is not a spk PACKAGE" # TODO: maybe find depended packages
+        make dependency-tree
+    fi
+done
--- a/.github/actions/build-push-pr.sh
+++ b/.github/actions/build-push-pr.sh
@@ -13,46 +13,76 @@ DEFAULT_TC=6.1
 PARALLEL_MAKE=4
 EOF
 
-# echo "FILES: $FILES"
+
+# get dependency tree
+DEPENDENCY_TREE=
+
+echo "Building dependency tree..."
+for package in $(find spk/ -maxdepth 1 -type d | cut -c 5- | sort)
+do
+    pushd "spk/$package" > /dev/null || exit
+    DEPENDENCY_TREE+=$(make dependency-tree)$'\n'
+    popd > /dev/null || exit
+done
+
 
 # filter for changes made in the cross and spk directories
 GH_FILES=$(echo "$GH_FILES" | grep -oE "(spk.*)|(cross.*)")
 
+echo "FILES: $GH_FILES"
+
 # create array of potential packages where files have changed
 GH_PACKAGES_ARR=()
-for GH_FILE in $GH_FILES
+for file in $GH_FILES
 do
-    # remove leading spk and cross from string
-    GH_FILE=${GH_FILE#spk/}
-    if [[ "$GH_FILE" == cross/* ]]; then
-        make dependency-tree
-        GH_FILE=${GH_FILE#cross/}
+    # Add spk
+    # remove leading spk from string
+    if [[ "$file" == spk/* ]]; then
+        package=${file#spk/}
+        # get package name / folder name
+        package=$(echo "$package" | grep -oE "^[^\/]*")
+        GH_PACKAGES_ARR+=("$package")
     fi
 
+    # Find dependencies
+    # remove leading spk/cross/native from string
+    file=${file#spk/}
+    file=${file#cross/}
+    file=${file#native/}
     # get package name / folder name
-    GH_PACKAGE=$(echo "$GH_FILE" | grep -oE "^[^\/]*")
-    # echo "PACKAGE: $PACKAGE"
-    GH_PACKAGES_ARR+=("$GH_PACKAGE")
+    package=$(echo "$file" | grep -oE "^[^\/]*")
+    echo "package: $package"
+    packages=$(echo "$DEPENDENCY_TREE" \
+        | awk -v package="$package" \
+        'NF == 2 {x=$2} $2 == package {print x}' \
+         | sort -u)
+
+    echo "awk: $packages"
+
+    for package in $packages
+    do
+        GH_PACKAGES_ARR+=("$package")
+    done
+
 done
 
 # de-duplicate packages
-GH_PACKAGES=$(printf %s "${GH_PACKAGES_ARR[*]}" | tr ' ' '\n' | sort -u)
+packages=$(printf %s "${GH_PACKAGES_ARR[*]}" | tr ' ' '\n' | sort -u)
 
-if [ -z "$GH_PACKAGES" ]; then
+if [ -z "$packages" ]; then
     echo "no package built. Empty PACKAGES var"
     exit 0
 fi
 
-echo "===> PACKAGES to Build: $GH_PACKAGES"
+echo "===> PACKAGES to Build: $packages"
 
-for GH_PACKAGE in $GH_PACKAGES
+# Build
+for package in $packages
 do
     # make sure that the package exists
-    if [ -d "spk/$GH_PACKAGE" ]; then
-        cd spk/"$GH_PACKAGE" && make "$GH_ARCH"
+    if [ -d "spk/$package" ]; then
+        cd spk/"$package" && make "$GH_ARCH"
     else
-        # must be from cross/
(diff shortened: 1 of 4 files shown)


.github/actions/build-push-pr.sh
```diff
@@ -0,0 +1,73 @@
...
+    file=${file#spk/}
+    file=${file#cross/}
+    file=${file#native/}
+    # get package name / folder name
+    package=$(echo "$file" | grep -oE "^[^\/]*")
+    echo "===> Searching for dependent package: $package"
+    packages=$(echo "$DEPENDENCY_TREE" \
+        | awk -v package="$package" \
+        'NF == 2 {x=$2} $2 == package {print x}' \
+         | sort -u)
```
publicarray, line 44: This awk command is a bit fragile as it depends that the output columns from `make dependency-tree` does not change, if someone with more awk wizardry can come up with something better please share 👍

.github/actions/build-push-pr.sh
```diff
@@ -0,0 +1,73 @@
...
+MAINTAINER?=
+MAINTAINER_URL=
+DISTRIBUTOR=SynoCommunity
+DISTRIBUTOR_URL=https://synocommunity.com/
+REPORT_URL=https://github.com/SynoCommunity/spksrc/issues
+SUPPORT_URL=https://github.com/SynoCommunity/spksrc/issues
+DEFAULT_TC=6.1
+PARALLEL_MAKE=4
+EOF
+
```
hgy59, line 15: I suppose it would be better to call `make setup` instead.
With `make setup` only `DEFAULT_TC=6.1` is overwritten when the file `local.mk` already exists. All other settings will not be changed in an existing `local.mk`
`make setup` is based on two other make targets:
- `local.mk` to create a default local.mk when the file does not exist.
- `dsm-%` to set DEFAULT_TC to the given value

Is is definitly bad when this script overwrites manually added settings like `PUBLISH_API_KEY`

publicarray: Thanks good call!

PUBLISH_API_KEY should not be committed to the git repository...🧐 Github secrets should be used instead.

publicarray: To actually publish one needs to uncomment the API_KEY and add publish- to the GH_ARCH environment variable.

https://github.com/SynoCommunity/spksrc/blob/5a6bae603e4dd3bd37c09acee9c4ca09bf0f07d4/.github/workflows/build.yml#L67-L70

hgy59:
I am trying to make a different dependency rule that does not include the `+` sign and the version number, but includes the spk/cross/native folder. 
I have a simple problem that I cannot solve. I need to concatenate the folder and the package name and cannot avoid a space in between. Any hint or good reference for shell commands in makefile rules would be appreciated.

.github/actions/build-push-pr.sh
ymartin59: Does it make sense to use #3854 code?

ymartin59:
I failed to do clever things with CircleCI like running parallel jobs, one job per package-architecture, to prevent a "all-or-nothing failure"...
A tricky point is to detect UNSUPPORTED_ARCH in package to avoid to report a failure if architecture artifact is lacking because of that flags. Probably additional support in Makefiles is required for such detection/validation, thanks to better make exit code generation.

hgy59:
> A tricky point is to detect UNSUPPORTED_ARCH in package to avoid to report a failure if architecture artifact is lacking because of that flags. Probably additional support in Makefiles is required for such detection/validation, thanks to better make exit code generation.

Just a hint: calling `make $(arch)-$(tc_version)` from bash script always returns exit code 0. Only `make ARCH=$(arch) TCVERSION=$(tc_version)` gives a exit code 2 on errors. My last attempt to generate a different exit code for unsupported archs didn't succeed.

hgy59:
What about the integration of dependabot into github actions?
At least for modified packages we should scan for security vulnarabilities.
- the existing dependabot repository scaner looks for python dependencies in requirements.txt only, but does not scan for WHEELS in Makefiles.
- May be other dependency scans could be integrated like PHP, JS, GO

.github/actions/build-push-pr.sh
publicarray: I've tried it, but it failed to return any packages when I modified a single line in a package in cross. I don't know how it works so I assume the best way is to walk the dependency tree.

ymartin59:
I find it a good idea. My opinion is that we should discard "WHEELS" variable to enforce a `requirements.txt` file usage for packages.

hgy59:
@ymartin59 I can make such a cleanup in #3648  (I was not aware that this is equivalent wheels definition).

publicarray:
>  scan for security vulnerabilities

I also think its a good idea, but there are a few challenges:

To scan the dependencies the source code needs to be downloaded and extracted. If this is done when the package is initially committed it provides very limited value since up-to-date packages rarely have a vuln on release. It's usually sometime after that they are found. So we need to download all packages & scan at a regular interval to be effective.

.github/actions/build-push-pr.sh
chickenandpork: @publicarray you're saying that there's an error in #3854.  Let's fix it.  What's your repro of the error?

ymartin59:
@publicarray @hgy59 @chickenandpork @th0ma7 @Safihre Do you all agree to merge and activate these GitHub Actions on project?

Safihre:
Can't judge the quality and would like to see it work, but this would be great if it's working as intended!

publicarray:
I should probably document that when a tag like `ffmpeg-4.2.2_0` or `dnscrypt-proxy-2.0.42`  is pushed a GH release is made, IIRC just like how the travis builds work. Note however that there can't be any dashes `-` in the version string.

th0ma7:
@ymartin59 and all, this looks like a pretty interesting idea.  If I get this right, when new releases gets published on github by various projects a "bot" will test-build associated `cross/package` using `spksrc` toolkit?  Thus resulting in detection of conflicts or errors and successes.

Will this generate end-result packages as well based on dependency policies?  Such as, a new zlib release would invoke a build of XXX package builds?  Also, I noticed over time that some of our cross configurations uses the default project web site but often that project also uses github releases.  Should we change the source over time to further enable the "action" feature?

Many questions and sorry for that.  But to answer you simply @ymartin59 yes I'm all in.  Only caveat I have is that with current DSM6 the gcc is so old that complex projects like ffmpeg may become interesting to follow.

publicarray:
@th0ma7 test builds will be created on any relevant file changes on a push or pull request. For example if  `cross/zlib/Makefile` is modified the action will build zlib and it's dependents.

As for "default project web site" it doesn't matter for the actions. we just use the regular makefile

th0ma7:
> test builds will be created on any relevant file changes on a push or pull request. For example if cross/zlib/Makefile is modified the action will build zlib and it's dependents.

Got it! Thnx for the explanation, I thought it was the other way around: when zlib project pushes a new release then github actions invokes a "staging/experimental-update" of relevant `cross/package` with a test-build (i.e auto-modifying `digest` and `Makefile`).  That would have been awesome :)

Question then @publicarray , if after a push it builds the package with its dependencies, will it also build packages that rely upon it as well to cover the entire spectrum up to the `spk/*` directory tree?

publicarray:
> auto-modifying digest and Makefile

Now that would be awesome! We are not there yet but it could be combined with vulnerability scanning. Maybe after I've settled on Mars 🚀 /jk

> it also build packages that rely upon it

Yes precisely 😃

th0ma7:
Then this is really nice (and by Mars it might become awesome!)

ymartin59:
@publicarray I wonder why actions run on "ubuntu-latest" when our Docker image and documentation is for Debian. Is it possible to use "debian-latest" instead?
Then I would like to have some details how to generate and setup token to publish release.

publicarray:
@ymartin59 [supported-runners-and-hardware-resources](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources) does not list Debian as an option. 🤷‍♂️

For the Github releases the [token is automatically created and removed](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#about-the-github_token-secret)
If you want to publish to the spkrepo you need to add your API Key (secret) to Github: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets and uncomment `API_KEY: ${{ secrets.PUBLISH_API_KEY }}` (I have not tested this bit) see: https://github.com/SynoCommunity/spksrc/pull/3982#discussion_r418918122

Diaoul:
I just wanted to say good work on this PR for the CI, seeing all those green arrows fills me with joy every time :heart:

publicarray:
Thank you! Just a small hiccup as it doesn't work as promised yet. The git command to get the changes doesn't work as intended: `Run echo "::set-output name=files::$(git diff-tree --no-commit-id --name-only -r 74d8a1eb3ca9da97596eba31257750f61d265c10 | xargs)`

running the command locally does what is says on the tin
```sh
git diff-tree --no-commit-id --name-only -r 74d8a1eb3ca9da97596eba31257750f61d265c10 | xargs
cross/ffmpeg/Makefile cross/ffmpeg/digests cross/ffmpeg/patches/006-Add-ability-for-ffmpeg-to-run-svt-hevc.patch cross/ffmpeg/patches/007-Add-ability-for-ffmpeg-to-run-svt-av1.patch cross/freetype/Makefile cross/freetype/PLIST cross/freetype/digests cross/fribidi/Makefile cross/fribidi/digests cross/gmmlib/Makefile cross/gmmlib/digests cross/intel-media-driver/Makefile cross/intel-media-driver/digests cross/intel-media-driver/patches/001-gcc493-fix.patch cross/intel-media-driver/patches/001-std_c++11.patch cross/intel-media-driver/patches/002-fix-error_array_used_as_initializer-PerfUtility.patch.DISABLE cross/intel-media-driver/patches/003-fix-error_MOS_Delete_void.patch cross/intel-media-sdk/Makefile cross/intel-media-sdk/PLIST cross/intel-media-sdk/digests cross/intel-media-sdk/patches/001-gcc493-fix-TEBMap-TEBIt-protected-to-public.patch cross/libaom/Makefile cross/libaom/PLIST cross/libaom/digests cross/libbluray/Makefile cross/libbluray/PLIST cross/libbluray/digests cross/libogg/Makefile cross/libogg/PLIST cross/libogg/digests cross/libva-utils/Makefile cross/libva-utils/digests cross/libva-utils/patches/001-ifdev-va_x11-in-videoprocess-vppscaling_n_out_usrptr.patch cross/libva/Makefile cross/libva/PLIST cross/libva/digests cross/svt-av1/Makefile cross/svt-av1/PLIST cross/svt-av1/digests cross/svt-hevc/Makefile cross/svt-hevc/PLIST cross/svt-hevc/digests cross/util-linux/Makefile cross/util-linux/PLIST cross/util-linux/digests cross/x264/Makefile cross/x264/PLIST cross/x264/digests cross/x265/Makefile cross/x265/PLIST cross/x265/digests spk/ffmpeg/Makefile spk/ffmpeg/src/service-setup.sh
```

but is empty here https://github.com/SynoCommunity/spksrc/runs/719290946

chickenandpork:
I used awk for its consistency in certain functions.  And for doing things in some binary rather than two.

You might be running into an oddity of the default command in xargs being an “echo”: maybe the command is not there to shell out, maybe it’s reacting differently to different env.  Maybe xargs has differences (BSD-vs-AT&T lineage).  You may be exhausting max shell line lengths (some are very short)

Try being explicit with the command: “xargs echo” (echo -n ?).  Try a different explicit app to print the lines.  Maybe try to limit the number of items xargs will send to each iteration of “echo”

Piping through python3 is always an option if you have that on the build env.

publicarray:
Thanks @chickenandpork for you help

The environment is known and I don't particularly like awk for complex stuff. But that is only my opinion.

I used https://github.com/nektos/act/ for local testing with some small modifications (use a single arch in the matrix and remove the cache step) 
Run with `act -b -P ubuntu-latest=node:lts-buster -j build` and the commands work fine.

So something is different in the environments. But what? I think it's something to with the state git is in not with xargs

```
Run echo "--"
  echo "--"
  git diff-tree --no-commit-id --name-only -r 7360d2c8566527f979633357f735270707e1095e
  echo "--"
  FILES=$(git diff-tree --no-commit-id --name-only -r 7360d2c8566527f979633357f735270707e1095e | xargs)
  echo "::set-output name=files::$FILES"
  shell: /bin/bash -e {0}
--
--
```

th0ma7:
> I just wanted to say good work on this PR for the CI, seeing all those green arrows fills me with joy every time heart

To make mileage from @Diaoul , thumbs up guys, really awesome work on this!  Loving it!

PR #4003, opened 2020-05-30
Fix gh actions

publicarray:
By default actions/checkout only retrieves the latest commit. To calculate the diff we need the previous commit. I failed to catch this problem earlier.

_Linked issues:_  #3982

Here is a run with ffmpeg: https://github.com/publicarray/spksrc/actions/runs/119815811


publicarray:
@ymartin59 can I just hit merge on this?

ymartin59:
@publicarray You may, but I often do not use the "standard" merge. Typically here, I have used "squash merge" to create a single commit and "normalize" commit summary/first line. I should document this "way of contributing" in wiki...

PR #4016, opened 2020-06-07
Add make dependency-list and update github-actions

hgy59:
_Motivation:_  Provide dependency-list as it is difficult to parse the dependency-tree in github actions

- [x] use dependency-list in github-actions
- [x] list optional dependencies
- [x] include folder to distict e.g. native/python3 and cross/python3
- [x] fix github build actions

### Remarks
Until now I didn't see builds in github actions. This might be a result of wrong includes in cross/sabnzbd/Makefile. 
Anyway this PR not only fixes the cross/sabnzbd/Makefile but adds the dependency-tree and dependency-list targets to the global Makefile.

To resolve optional dependencies I introduced the variable `OPTIONAL_DEPENDS` that must be used in Makefiles with dependencies that are not added by default (e.g. arch specific dependencies).

Unfortunately dependency-tree does not eliminate duplicate dependencies (but dependency-list does).


hgy59:
@publicarray I cancelled the github build actions after 20 minutes for the commit 9178991b1af1250032c3aa32f5bb7c3a67a2c25a.
I suppose this is due to changing Makefiles that trigger ffmpeg and other long lasting builds. 
The following commit did not build anything.

This leads me to the follwing questions:
- does it make sense to build all affected packages for 7 different archs on each commit?
- the packages to build must not be evaluated by the changes of one single commit but for all changes on the individual branches (or the changes related to the current master).
- I wanted to cancel all but the x64 builds, but the respective button cancelled all build tasks. Therefore I dont know how long the the actions would have taken (I don't know the limits for github actions but I doubt that hours of cpu for each commit will be free).

hgy59:
@publicarray I just fixed the github build actions to generate errors when make fails.
But now we need to skip the generic arch builds (x64, aarch64, armv7) for packages with `REQ_KERNEL` set.
How can we achieve this?

And we should handle `noarch` packages to build without ARCH parameter.

hgy59:
@publicarray 
may be this error is related to failure that not all affected packages are built

there is an error in the step `Get changed files PR`
```
Run git diff-tree --no-commit-id --name-only -r master d7eb04d0e1cf43958397ff10c7a730581e16b8f1 | xargs
fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

hgy59:
@publicarray
I fixed the build step "Get changed files PR" and now all changed files on the brach are used to trigger builds.
As documented https://github.com/features/actions the github actions are free for public repos. Therefore I will not cancel the checks now.

Since we can not skip errors for `KERNEL_REQ` or `UNSUPPORTED_ARCH` for now, I propose to use `continue-on-error: true` in `Build Package (file changes)`.

hgy59:
@publicarray 
I found the github action limits here: [Usage Limits](https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits)

So the first limit we may reach is that a job may not exceed 6 hours.

publicarray:
Thanks @hgy59 Some good improvements here! There is also a global limit of [2,000 hours/month](https://help.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions#about-billing-for-github-actions) 
Since ppc853x-5.2 uses an old DSM version I did not include it in the build but I do the the value of having it here.

> does it make sense to build all affected packages for 7 different archs on each commit?

I think so but there could be a problem if a commit modifies a lot of packages at once. We my hit the 6 hour limit. However I did modify ffmpeg and it "only" took 4h 20m 13s for ffmpeg and all dependent packages

> the packages to build must not be evaluated by the changes of one single commit but for all changes on the individual branches (or the changes related to the current master).

Yes I made a change to do that for pull requests (diff from master .. HEAD)

> skip the generic arch builds (x64, aarch64, armv7) for packages with REQ_KERNEL set.
How can we achieve this?

Maybe we need to grep the Makefile for each potential build?: `grep 'REQ_KERNEL' spk/sabnzbd/Makefile` or `grep -E 'REQ_KERNEL\s?=\s?1' cross/lirc-0.9.0/Makefile`

> And we should handle noarch packages to build without ARCH parameter.

This one if more difficult maybe we can address this in a future PR. We need to get the noarch packages before we use the per arch matrix.

> `unknown revision or path not in the working tree.`

Thanks I missed it 🤦‍♂️

hgy59:
@publicarray 
With `continue-on-error: true` the last commit produced some artifacts.
Some builds failed due to certificate issues. Created issue #4017 for this.

Do we really need to double the artifacs size by creating zip files for each arch and one zip containing the packages of all archs?

publicarray:
The one containing all of them is for the tagged build, they get downloaded and pushed as a release. The others are for convenience

hgy59:
> There is also a global limit of [2,000 hours/month]

This limit is for private repositories only. There is no limit for `Public repositories`
(taken from https://github.com/features/actions)

publicarray:
Ah ok, 2,000 would have been enough anyway that's 65 per day / 8 arches = 8h

hgy59:
To avoid another build that needs more than 5 h 20 min to complete, I propose to merge this PR now and to remove the obsolete artificats (the arch specific archives) later.
For this PR the artifact packages.zip is 775MB (but the download menu displays 106 MB, the same size as the x64 package).

publicarray:
You don't need to remove the artifacts they are only temporary and will be gone in a few days. They are mostly used for other workflows. e.g To create a Github Release. https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts#about-workflow-artifacts

publicarray:
Thank you for your work on this!

PR #4023, opened 2020-06-11
Build/fix GitHub actions

smaarn:
_Motivation:_  Correcting execute flag on script used in github actions


smaarn:
> @smaarn Thanks for this!

well thank you for setting up github actions ! I was getting mad with twisted Circle CI logics to avoid messing up the SynoCommunity repository... now things will be much simplier ;)

hgy59:
@smaarn The setup up of gh actions was done by @publicarray. I just simplified the dependency evaluation, and now I am improving some details. I am not used with this stuff for github (at work I use gitlab CI/CD).

PR #4024, opened 2020-06-11
[github actions] Update build_status.sh to print out the proper error output file contents

smaarn:
_Motivation:_  When errors are being spotted the contents of the error file is not printed out


hgy59:
oops

Issue #3941, opened 2020-04-06
Syncthing package v1.3.4


Kaelber:
The current version of Syncthing is v1.4.0 and tomorrow v1.4.1 will official release. Until v1.4.0 a new database layout is implemented and the database needs a migration from versions <1.4.0, so also from v1.2.2 or v1.3.4.

The new database layout could be a confusion, if a user use the inside updater function and update the system files from the old v1.2.2 to the newest one, today v1.4.1-rc.3. I also doe that.

Now you update your packkage from v1.2.2 to v1.3.4 and messed with the above facts, because a re-migration is not possible and the package crashes. Next is, why you did not go straight to v1.4.0 is also not clear to me, because of course many of the updater are surely now at least on v1.4.0 or even on v1.4.1-rc.3. The database schema no longer fits. Next is now, the whole package no longer runs after the update.

In any case, you not paying attention. I must therefore deleted the SC package.

acolomb:
There is already an ongoing discussion in #3884.

ymartin59:
@Safihre We already had such a discussion - but here is a perfect scenario to motivate my willing to disable "in application live update" (moreover it is probably not functional for some architectures)...

acolomb:
@ymartin59 I see you found the right issue about an unfortunate Syncthing version.

Disabling in-app upgrades would be fine and is what many people would expect from a "managed" package anyway. It requires looking through DSM notifications for upgrades more often, but I think people can live with that.

If we go down that route, there should be a process though to ensure more timely SPK updates. I'd volunteer to push Syncthing releases more often, but it doesn't help if the limiting factor is your time to review PRs and sift through e-mails. Which I totally understand, no offense intended.

Please review my thoughts about the broader matter in [this comment](https://github.com/SynoCommunity/spksrc/pull/3880#issuecomment-585092391), which apparently went under your radar as well.

acolomb:
@ymartin59 While trying to rebase PR #3931 onto current master, I hit some problems regarding `PKG_DIST_FILE` in `cross/syncthing/Makefile`:

```
make[3]: Entering directory '/spksrc/cross/syncthing'
===>  Downloading files for syncthing
===>    File syncthing-1.4.0.tar.gz already downloaded
===>  Verifying files for syncthing
===>    Downloaded file syncthing-1.4.0.tar.gz not in digests file
../../mk/spksrc.checksum.mk:42: recipe for target 'checksum_target' failed
```

[...]

PR #4027, opened 2020-06-12
syncthing: Incorporate upstream release v1.7.1.

### LINKED ISSUE #4026: No GUI after updating to Syncthing 1.4.0-19

Syncthing
1.4.0-19

DS213A and DS213J
6.2.3-25426

I updated both machines from 1.3.something to 1.4.0-19. On both machines Syncthing doesn't start, hitting the run button in the package center gives very briefly a running state and then again stopped state.

 
### Expected behavior
Tell us what should happen

### Actual behavior
Tell us what happens instead

### Steps to reproduce
_1._ 
_2._ 
_3._ 

#### Package log
```
Insert the package log here
```
#### Other logs
```
Insert log here
```

acolomb:
Only package version and hashes updated.  Built successfully for
all-supported.

Note that the previous package version should have already
self-updated an installed Syncthing binary when the service was
running long enough.

_Motivation:_  Syncthing updates itself beyond what was packaged in the SPK. Users upgrading from some older SPK version (< v1.4.0 currently) will already have the Syncthing core binary running at a newer version (currently v1.6.1).  That newer version will have changed the database schema, causing the "upgrade" to v1.4.0 to choke on the existing database.

_Linked issues:_  Previously witnessed in #3941. Fixes #4026 hopefully.

### Checklist


acolomb:
Note that the same problem will pop up repeatedly, next occasion will be July 7th when version v1.7.0 will be released which includes another database schema change.

Unfortunately, the SPK update can only be prepared after the release since we need the final file hashes. So releasing shortly after Syncthing itself will be a tedious task and still error prone for this issue.  I'm actually quite happy with the internal updater because it requires zero attention. So maybe we should take precautions in the packaging to avoid this problem?

acolomb:
Ping?

Syncthing version 1.7.0 will be released in just over a week.  If there is no reaction on this PR until then, I will keep it updated for the next version.

acolomb:
Syncthing v1.7.0 has been released. Should I update this PR or will you rather stay on v1.6.6?

acolomb:
I went ahead and pushed the new version. Feel free to revert if that was not desired.

Version v1.6.1 was tested by a couple of users already. This v1.7.0 has only seen a successful build on `all-supported` so far.

acolomb:
According to https://github.com/SynoCommunity/spksrc/issues/4026#issuecomment-664367690, the v1.7.0 version was also tested successfully.  However there is now already v1.7.1 available.

I pushed that update as well to minimize the potential for upgrade-related errors here. @ymartin59 Any news on when this can finally be published?

I assume there will regularly be some users affected by #4026, which this PR addresses. By now, Syncthing already has better handling in place to try an upgrade before bailing out on a too recent database schema. But not before this PR.

PR #4030, opened 2020-06-14
Upgrade to vim 8.2 and build with huge features + dynamic python3

lwfitzgerald:
_Motivation:_  Upgrade `vim` to 8.2 and compile with huge features and dynamic `python3` support. This supports recent versions of plugins and those that rely on Python 3 support (if the `python3` SPK is installed).

Configuring `vim` with dynamic Python 3 support means it will attempt to
load `libpython` using `dlopen`, allowing it to gracefully fail if it is
unavailable. This removes the need for a dependency on the `python3`
SPK and avoids shipping of a redundent set of Python 3 bins/libs within
the `vim` SPK package.

### Checklist


spk/vim/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = vim
-SPK_VERS = 7.4
-SPK_REV = 2
+SPK_VERS = 8.2
+SPK_REV = 1
```
hgy59, line 3: you need to increment revision (SPK_REV) to 3
it must be ascending over all packages with the same name (SPK_NAME) independent of SPK_VERS.
see https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables

hgy59:
Does this version of vim need python3 to compile or only at runtime?

When python3 is required at runtime only, it must not depend on cross/python3, but something like `SPK_DEPENDS = "python3>=3.5.6-8"` to depend on python3 package installation on the device.

lwfitzgerald:
> Does this version of vim need python3 to compile or only at runtime?

Prior to the update I just pushed, `python3` was required for both building and running. A set of Python 3 bins/libs were also included in the SPK.

Now `vim` is built with "dynamic" Python 3 support. This means it will attempt to
load `libpython` using `dlopen`, allowing it to gracefully fail if it is unavailable. This removes the need for a dependency on the python3 SPK and avoids shipping of a redundent set of Python 3 bins/libs within the `vim` SPK package.

ymartin59:
@lwfitzgerald Is it really required to (re-)build Python 3 to compile this support in Vim? Wouldn't be possible to "fake" Python 3 support using the system python version - even if not exactly same version as SPK package?

cross/vim/Makefile
```diff
@@ -1,25 +1,46 @@
...
-CONFIGURE_ARGS += vim_cv_stat_ignores_slash=yes vim_cv_memmove_handles_overlap=yes
+CONFIGURE_ARGS = --with-features=huge vim_cv_toupper_broken=no --with-tlib=ncurses vim_cv_terminfo=yes vim_cv_tgetent=zero vim_cv_tty_group=world
+CONFIGURE_ARGS += vim_cv_getcwd_broken=yes vim_cv_stat_ignores_slash=yes vim_cv_memmove_handles_overlap=yes
+CONFIGURE_ARGS += --enable-python3interp=dynamic --with-python3-command=$(HOSTPYTHON) vi_cv_path_python3_conf=$(PYTHON_CFG_PATH) vi_cv_path_python3_pfx=$(PYTHON_PREFIX) vi_cv_path_python3_epfx=$(PYTHON_EXEC_PREFIX) vi_cv_path_python3path=$(PYTHONPATH)
 
+# Add the Python 3 SPK's installed lib dir to vim's rpath to allow loading of libpython dynamically if it's available
+ADDITIONAL_LDFLAGS = -Wl,--rpath,/var/packages/python3/target/lib
+
+CONFIGURE_TARGET = myConfigure
 INSTALL_TARGET = myInstall
```
ymartin59, line 25: We have changed target naming policy. Please rename as vim_install

cross/vim/Makefile
```diff
@@ -1,25 +1,46 @@
...
+endif
+
+PYTHONPATH = $(PYTHON_SITE_PACKAGES_NATIVE):$(PYTHON_LIB_NATIVE):$(INSTALL_DIR)$(INSTALL_PREFIX)/$(PYTHON_LIB_DIR)/site-packages/
+PYTHON_PREFIX = $(STAGING_INSTALL_PREFIX)
+PYTHON_EXEC_PREFIX = $(STAGING_INSTALL_PREFIX)
+PYTHON_CFG_PATH = $(shell $(STAGING_INSTALL_PREFIX)/bin/python3-config --configdir)
+
+.PHONY: myConfigure
+myConfigure:
+	$(RUN) PYTHONPATH=$(PYTHONPATH) ./configure $(REAL_CONFIGURE_ARGS)
```
ymartin59, line 42: Do not replace original configure target but append PYTHONPATH to dedicated variable `ENV`

cross/vim/Makefile
```diff
@@ -1,25 +1,46 @@
...
+# Only trigger building of dependencies (for the Python config) for targets where it's used
+ifeq ($(strip $(filter-out configure compile install plist all,$(MAKECMDGOALS))),)
+$(WORK_DIR)/python-cc.mk: depend ;
+include $(WORK_DIR)/python-cc.mk
+endif
+
+PYTHONPATH = $(PYTHON_SITE_PACKAGES_NATIVE):$(PYTHON_LIB_NATIVE):$(INSTALL_DIR)$(INSTALL_PREFIX)/$(PYTHON_LIB_DIR)/site-packages/
+PYTHON_PREFIX = $(STAGING_INSTALL_PREFIX)
+PYTHON_EXEC_PREFIX = $(STAGING_INSTALL_PREFIX)
+PYTHON_CFG_PATH = $(shell $(STAGING_INSTALL_PREFIX)/bin/python3-config --configdir)
```
ymartin59, line 38: Why not hard-code configdir location the same way as rpath?

spk/vim/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = vim
-SPK_VERS = 7.4
-SPK_REV = 2
+SPK_VERS = 8.2
+SPK_REV = 3
```
ymartin59, line 3: Please also update package CHANGELOG

lwfitzgerald:
> Wouldn't be possible to "fake" Python 3 support using the system python version - even if not exactly same version as SPK package?

I don't think this is possible - The `vim` build includes a header from the Python config dir and uses this to configure the functions/symbols`vim` tries to find in the shared lib. There's an example of this here - https://github.com/vim/vim/blob/v8.2.0985/src/if_python3.c#L490-L493. This fixes the minimum supported version of Python to one that has those functions available in the shared lib. You can see a failure occurs if a symbol is not available here - https://github.com/vim/vim/blob/v8.2.0985/src/if_python3.c#L676-L684.

The build environment only has access to the build configuration for the latest version of the Python SPK, so it can only use the Python config generated by building that (as a build-only dep).

The other option would be to declare `cross/python3` as a `DEPENDS` dependency and do a non-dynamic build, leading to a separate (and potentially redundant) set of Python bins/libs getting bundled in the the `vim` SPK. @hgy59 wanted to avoid this and use an SPK dependency though.

cross/vim/Makefile
lwfitzgerald: The config directory has the architecture+ABI mangled into the name (e.g `config-3.6m-arm-linux-gnueabihf`) so this differs between the builds.

Issue #4028, opened 2020-06-13
ntopng multi-arch build failure on master


th0ma7:
### Setup
_Package Name:_ ntopng
_Package Version:_ 3.8.1 (master).  While at it also note that a new v4.0 just got out in march.

_Problem Description:_ While working on cmake mk framework PR #3998 I ended getting a `ntopng` build failure on multiple arches as part of github actions.  Problem is currently present on master `spksrc`.

_NAS Architecture:_ `x64`, `evansport`, `aarch64`, `armv7-6.1`, `armv7-1.2`, `88f6281`, `qoriq`, `ppc853x-5.2`

#### Build log
```
[...]

PR #4031, opened 2020-06-16
ntopng: build fix when using non /spksrc root dir

th0ma7:
_Motivation:_  Building `ntopng` only works from `/spksrc` root dir.  This fix addresses this.
_Linked issues:_  #4028

### Checklist

### Notes
- Previous builds doesn't show support for PPC arches except `qoriq` - now forced as such.
- Makefile default `REQUIRED_DSM = 6.1` resulting in `armv7-1.2` to no be built


th0ma7:
@hgy59 interesting case here to test out false-negative when an `arch` isn't supported.

PR #4039, opened 2020-06-21
[github-action] cleanup disk space in build action

hgy59:
_Motivation:_  The build action for python fails with `System.IO.IOException: No space left on device`.

The space of the build environment in github actions is currently limited to 14GB.
On my private build system the size of the work folder of a sucessful python build is about 1.5GB.
When the build action builds more than one package depending on cross/python the limit is reached soon. As each package uses it's individual work folder, we can cleanup this folder after each package build.

- use make clean after each build to limit used space
- build ffmpeg before dependents and do not clean ffmpeg
- add annotations for failed builds
- add additional structure to build log
- rename main build script to build.sh


hgy59:
@th0ma7 do we have to handle this cleaning differently for ffmpeg and it's dependents?

th0ma7:
Good question.  `ffmpeg` takes a huge amount of space.  But it is important to have `ffmpeg` being built (and kept) prior to `chromaprint` and `tvheadend` as it allows much faster build time.  Although once theses two are completed `ffmpeg` can be cleaned-up as well.

As such, an exception for `ffmpeg` is probably best or a way to state such as a new variable "`BUILD_DEPENDENCY`" to create such linkage?

hgy59:
> As such, an exception for `ffmpeg` is probably best or a way to state such as a new variable "`BUILD_DEPENDENCY`" to create such linkage?

Until now the build dependency check in the gh build action does not handle any build order of dependent packages.
We already have `BUILD_DEPENDS` variable, but this is used to build in the work-folder and to avoid to include such dependents into the spk.
May be a better naming would be `DEPENDS_ON_BUILD_OF`  or `DEPENDS_ON_SPK_BUILD` or `USE_CACHED_SPK_BUILD`.

For a generic approach I think we need both ends of the dependency. I.e. ffmpeg defines `PROVIDE_CACHED_BUILD = 1` and the dependents define `USE_CACHED_BUILD = spk/ffmpeg`.
Or we simply hard code this for ffmpeg in the build action script.

Anyway the build behaviour would be like this:
1. if only one package with such a dependency to ffmpeg needs to be built, there is nothing to keep (not even to build spk/ffmpeg)
2. if only ffmpeg needs to be built, there is nothing to keep too.
3. if more than one package with ffmpeg dependency (or one package and ffmpeg) needs to be built, we first must build spk/ffmpeg and than all ffmpeg dependents before cleaning spk/ffmpeg.

A try to make this even simpler:
All packages that have `PROVIDE_CACHED_BUILD = 1` that are triggered to build by its own or by a  `USE_CACHED_BUILD` dependent, are built before all other packages and always skip the cleaning step.

PS: until now the packages are built in alphabetical order.

hgy59:
@th0ma7 I worked out a solution for ffmpeg without modification of any Makefile.
This is specific to ffmpeg an may be mirgrated to a more generic approach when needed by other packages.
Dependents are identified by `export FFMPEG_DIR` in the spk Makefile.

th0ma7:
@hgy59 got another idea... what if instead of referring to `cross/ffmpeg` in `chromaprint` and `tvheadend` we refer to `spk/ffmpeg` directly?  

github-actions could use a rule of thumb that any package found as part of all `DEPENDS` as `spk/?` is not being cleaned-up until the end of build for that specific arch?  Also, alphabetical order would not have any impact anymore as it would end-up being built anyway when needed using `spk/ffmpeg` and kept until end of build for the arch...

hgy59:
> @hgy59 got another idea... what if instead of referring to `cross/ffmpeg` in `chromaprint` and `tvheadend` we refer to `spk/ffmpeg` directly?
> 

This would be a good approach. But I am afraid that this would have great impact to the framework.
Until now dependencies are built in the work folder of the current spk Makefile.

May be a solution would be to use a pre-target to build ffmpeg in it's own folder like:
`make -C ../ffmpeg`

If make was called with ARCH and TCVERSION defined in the environment this would work, but the arch-{ARCH}-{TCVERSION} parameter may not work with this.

Anyway, the github build action must avoid to clean the work-folder of ffmpeg.

hgy59:
@th0ma7 as this is very specific to ffmpeg I will leave this for now and merge as is.

Please have your 👀 on the build action the next time you commit ffmpeg or one of its dependent.

th0ma7:
> @th0ma7 as this is very specific to ffmpeg I will leave this for now and merge as is.
> 
> Please have your eyes on the build action the next time you commit ffmpeg or one of its dependent.

Will certainly do!

Still I believe using `DEPENDS = spk/ffmpeg` would allow killing two birds with one stone...  And potentially reuse that on other similar dependencies we have.

PR #4061, opened 2020-07-06
Update Sonarr version to v0.2.0.1504

mike10010100:
_Motivation:_  Updating Sonarr version to v0.2.0.1504

In the process of building all-supported, not sure how to test package upgrade and new installation, please advise.

### Checklist

### DIFF
--- a/cross/radarr/Makefile
+++ b/cross/radarr/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = Radarr
-PKG_VERS = 0.2.0.1293
+PKG_VERS = 0.2.0.1504
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME).develop.$(PKG_VERS).linux.$(PKG_EXT)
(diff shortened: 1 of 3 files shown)


Safihre:
What is the point of this? It can just auto-update itself?

mike10010100:
I was under the impression that Syno packages required updates to be made through the synology package manager. I've had an instance in the past where self-updating with Sonarr/Radarr ended up with them being in a bad state. If this is not the case, then I'll go ahead and close this.

Safihre:
No, the packages support auto-updating so we only provide the base here and then it updates from there.
If something is wrong with the base, we can update. But so far for Sonar v2 it has never been needed.

mike10010100:
Oh man, my bad, this is updating radarr's version, but I'll give the self-update a try. It just seemed like Sonarr had been updated to a newer package version in the synocommunity repo, so I figured the same needed to be done with radarr.

PR #4066, opened 2020-07-12
update curl to fix usage of ca certificates

hgy59:
_Motivation:_  As fixed for git with #4012 this is a general issue with cross/curl
_Linked issues:_  closes #4032, fixes #4084

### Checklist
_for checked packages: see list below Remarks._
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

### Remarks
The target of this PR is to update all related packages to use the DSM bundled ca certificates.
- [x] domoticz
- [x] git
- [x] icecast
- [ ] jackett
- [x] lidarr
- [x] ntopng
- [x] nzbdrone (sonarr)
- [x] python
- [x] python3
- [x] radarr
- [ ] rutorrent (configures cert.pem in .rtorrent.rc)
- [x] transmission

PR #4067, opened 2020-07-12
force tls 1.2 for downloads with wget

hgy59:
_Motivation:_  github build actions fail often to download sources
_Linked Issues:_ fixes build of #4066 

### Remarks
1. As SSL protocols <= TLSv1.1 are not supported by most sites nowadays and wget defaults to TLSv1.0 we need to force the use of TLSv1.2 for source downloads ([see reference](https://www.gnu.org/software/wget/manual/html_node/HTTPS-_0028SSL_002fTLS_0029-Options.html)).
This option is ignored for downloads with http protocol. Anyway a lot of download sites support https and should be updated in cross and native Makefiles.

2. The option to define `DOWNLOAD_ARGS` is removed, as it was used only to disable certificate check in cross/libsodium/Makefile with `DOWNLOAD_ARGS = --no-check-certificate`, and this is fixed with TLS 1.2.

### DIFF
--- a/mk/spksrc.download.mk
+++ b/mk/spksrc.download.mk
@@ -128,7 +128,7 @@ download_target: $(PRE_DOWNLOAD_TARGET)
 	        rm -f $${localFile}.part ; \
 	        url=`echo $${url} | sed -e '#^\(http://sourceforge\.net/.*\)$#\1?use_mirror=autoselect#'` ; \
 	        echo "wget $${url}" ; \
-	        wget $(DOWNLOAD_ARGS) -nv -O $${localFile}.part $${url} ; \
+	        wget --secure-protocol=TLSv1_2 -nv -O $${localFile}.part $${url} ; \
 	        mv $${localFile}.part $${localFile} ; \
 	      else \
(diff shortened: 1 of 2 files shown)


mk/spksrc.download.mk
```diff
@@ -128,7 +128,7 @@ download_target: $(PRE_DOWNLOAD_TARGET)
 	        rm -f $${localFile}.part ; \
 	        url=`echo $${url} | sed -e '#^\(http://sourceforge\.net/.*\)$#\1?use_mirror=autoselect#'` ; \
 	        echo "wget $${url}" ; \
-	        wget $(DOWNLOAD_ARGS) -nv -O $${localFile}.part $${url} ; \
+	        wget --secure-protocol=TLSv1_2 -nv -O $${localFile}.part $${url} ; \
```
ymartin59, line 131: I wonder if it may be better to replace wget by curl here

hgy59: I tried to use curl, but curl fails to download from ftp:// sites (wget has no problems with ftp://)

PR #4094, opened 2020-08-02
borgbackup: downgraded ruamel.yaml to 0.15.78

smaarn:
_Motivation:_  ruamel.yaml version shipped with python3 is now "too recent"
_Linked issues:_  Fixes partially #3944 

### Checklist

### DIFF
--- a/spk/borgbackup/src/requirements.txt
+++ b/spk/borgbackup/src/requirements.txt
@@ -1,7 +1,7 @@
 # Cross compiled packages dependencies
 #borgbackup==1.1.9
 #msgpack-python==0.5.6
-#ruamel.yaml==0.15.78
+ruamel.yaml==0.15.78
 
 # Downloaded dependencies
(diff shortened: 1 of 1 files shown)


spk/borgbackup/src/requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Cross compiled packages dependencies
 #borgbackup==1.1.9
 #msgpack-python==0.5.6
-#ruamel.yaml==0.15.78
+ruamel.yaml==0.15.78
```
smaarn, line 4: It turns out that the cross compiled version shipped with python 3 is too recent here as compared to the implicit requirements (which basically enforce that it should be lower than 0.16.0

ymartin59:
@smaarn In fact I expect to publish a borgbackup update to latest version...

smaarn:
Closing the PR since the target is more the other way around.

Issue #4075, opened 2020-07-23
Nano not available for DS720+ DSM 6.2.3-25426 Update 2


mshipek:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ Nano
_Package Version:_ 2.9.8-5

_NAS Model:_ DS720+
_NAS Architecture:_ geminilake
_DSM version:_ 6.2.3-25426 Update 2

### Expected behavior
Tell us what should happen - Package should be available in the community area like it is for my other 5 Synology NAS boxes.  It's not visible in the list.  Other NAS boxes are RS814, RS815+, etc.. If this package hasn't been built for this Arch yet, would I request a new package?  Thank you,

### Actual behavior
Tell us what happens instead - Nano is missing from the list on this Synology Model

### Steps to reproduce
_1._ 
_2._ 
_3._ 

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
```
Insert the package log here
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here
```

ymartin59:
You can find nano in synocli-file package: https://synocommunity.com/package/synocli-file

Issue #4070, opened 2020-07-15
Synology DS720+ with Syncthing


rdebruin:
_For new Package Requests, see the guidelines_  

### Setup
Syncthing

Synology DS720+
Gemini lake
DSM: latest release 

### Expected behavior
Maybe releated to a previous entry, but I would like to see Geminilake is supported. I am trying to install Syncthing, but the package cannot be installed.

ymartin59:
Published

Issue #4065, opened 2020-07-09
Synology DS 420 + http://packages.synocommunity.com 


Herodes1:
Hello i get no response from http://packages.synocommunity.com via DSM and website https://synocommunity.com/ seems to have cloudfare problems, and it have been like this for some days now.
Anyone knows when the site will be back up.

On my DS 213j it is working.
On my DS 420+ isnt it.

ymartin59:
You should register https://packages.synocommunity.com in DSM as documented at https://synocommunity.com/
DS420+ is geminilake arch and no arch specific package has been published for this arch yet - https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures

Herodes1:
Hallo Yves, I did this but nothing is happening. please take look on the ScreenshotsAm 09.07.2020 22:08 schrieb Yves Martin <notifications@github.com>:
You should register https://packages.synocommunity.com in DSM as documented at https://synocommunity.com/
DS420+ is geminilake arch and no arch specific package has been published for this arch yet - https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures

icegunner:
Given that the geminilake arch is a still an x86/x64 arch, shouldn't packages that are built for archs like avoton, apollolake, etc. be supported? Wouldn't most of those packages also be compatible/tagged and therefore show up in Package Manager?
If this isn't the case, isn't there a way to manually install packages even they don't appear in the Package Manager for those brave enough to give it a try? I'd be up for that.

ymartin59:
@icegunner No because package contains exact list of architectures it is expected to support. At worst, it may be possible to edit INFO file in spk tar archive - you may try on your own to use `tar tvf package-version.spk`, extract INFO, edit and update INFO in archive.
Until this trouble https://github.com/SynoCommunity/spkrepo/issues/47 is not fixed, I cannot publish any update

icegunner:
@ymartin59 Now that [SynoCommunity/spkrepo#47](https://github.com/SynoCommunity/spkrepo/issues/47) is fixed, what does the timeline pushing updates look like? Would love to migrate all my packages to my 920+. No rush, just very interested. Thank again for all you do.

THX723:
@ymartin59 I tried building the package on my private server and there appears to be some checksum related issue with x264.
```
make[3]: Entering directory '/spksrc/cross/x264'
[...]

Issue #4063, opened 2020-07-07
[FFmpeg] Request for update: Gemini Lake arch


THX723:
I see that `geminilake` architecture is finally included in Master. Would like to request for an update/rebuild of the existing FFmpeg package so it is compatible with the newer DSx20 series on Gemini Lake CPU.

I do understand this build will only be leveraging the generic x64 sets. The Gemini Lake specifics will have to wait until Synology release and updated toolchain.

th0ma7:
`geminilake` package of `ffmpeg` is on its way (version 4.2.4).  Leave me a day or two and it should be available.

th0ma7:
Package are now published, closing.

PR #4105, opened 2020-08-05
arch-geminilake: Add DSM 6.2.2 toolchain and kernel

th0ma7:
_Motivation:_  Follow-up on a request to Synology to make accessible the `geminilake` arch toolchain & kernel sources.

Also, following a discussion with @ymartin59 it now also includes:
* all DSM 6.2 and 6.2.2 toolchain & kernels for all compatible arches
* fake DSM 6.2 arches for `purley` and `geminilake` to circumvent limitation of the publishing api (max 3 chars. for version)

Lastly, due to URL change for toolchain on sourceforge.net:
* Update & tested all DSM 5.2, 6.1, 6.2, 6.2.2
* Bulk changed <= DSM 5.1

_Linked issues:_  #4075, #4070, #4065, #4063, #4044, #3999
_Associated PR:_ #3787

### Checklist

**NOTE:** Currently pending `geminilake` kernel sources, discussion with Synology on-going.


th0ma7:
@ymartin59 next steps are adding/completing the DSM-6.2 for API compatibility (not just `purley` and `geminilake`) and bulk change all other pre-DSM-5.2 toolchain configs.  Overall toolchain & kernel for DSM-5.2, 6.1 and 6.2.2 are now fully tested.  Also, I've added a TC_KERNEL variable for future kernel related work and fixed `mk/spksrc.kernel-modules.mk` so the built kernel version really matches Synology's version (e.g. through `mk/spksrc.kernel-modules.mk` add '`+`' sign to `EXTRAVERSION` variable).

Any additional thoughts before I move along with next changes?

th0ma7:
@ymartin59 it is now ready for a good review as this is trivial but rather large.  I manually tried many many of them, and **all** of DSM 5.2+ up to 6.2.2.

Synology's reshuffling of directories online makes me believe DSM 6.2.3 will be made available shortly to fill in the gap for missing kernel files for `geminilake` as I requested, or better yet, a DSM7-beta, who knows?  If that's the case then it could end-up going into another PR.  At least this one if fixing the current issue.

th0ma7:
> It would have been possible to collect many lines of variable settings and "myFix" target in per architecture and/or per major version common files (for instance `toolchains/common-x64.mk`) to discard copy-paste pattern and ease future maintenance.

@ymartin59 actually perhaps a change directly in the `mk/spksrc.tc.mk` would be more appropriate as most of them do have a `myFix`.  At least the most common one could be there such as the `chmod`.  I'll have a look at it in the weeks ahead and come back with a proposal.

On that note, what's the intent of the `find` command in the `myFix`?  I ask because `purley` didn't had any and I only included the `chmod` as it was needed but left aside the `find` command.  Lack of having it would it create build or runtime issues afterwards (e.g. thinking of `purley` `ffmpeg` package I've uploaded yesterday that I can't really test)?

On a somewhat similar note I also intent to move `cross/ds.evansport` to a new directory structure called `spksrc/toolkit` with probably a `mk/spksrc.tk.mk` helper... What do you think of that?

ymartin59:
As far as I understand "myFix", it fixes ownership, group and access rights on file extracted from toolchain .tar.xz where (probably) Unix permissions are (sometimes/often) mangled if created from a non-Unix system and quickly published without control...
My opinion is whatever toolchain is correct or not, we should apply fix-logic - notice that when run in our docker on macOS is in trouble according many reports - that "fix" would also make sense there.
My idea of `toolchains/common-x64.mk` is to collect common variables for `x64` and avoid some many replicated lines.

PR #4108, opened 2020-08-06
Increase package revision for publish into repository

ymartin59:
Required to publish non-updated packages for geminilake

### DIFF
--- a/spk/gateone/Makefile
+++ b/spk/gateone/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = gateone
 SPK_VERS = 20171125
-SPK_REV = 8
+SPK_REV = 9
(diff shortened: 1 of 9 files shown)


ymartin59:
@hgy59 In this specific case, checks are useless. Is there a way to disable them for a single PR?

publicarray:
@ymartin59 not by default but we could add something like this to the workflow

```yml
jobs:
  build:
    if: "! contains(toJSON(github.event.commits.*.message), 'skip-ci')"
```

https://github.community/t/github-actions-does-not-respect-skip-ci/17325/8

ymartin59:
@Safihre for geminilake support, nzbget was published again.

PR #4120, opened 2020-08-11
Make nv12 the default filter. This fixes hvec->h264 VAAPI transcoding.

jonozzz:
_Motivation:_  VideoStation fails to play hevc videos.

### Checklist

VideoStation's default ffmpeg args include:
`-vcodec h264_vaapi -vf format=nv12|vaapi,hwupload,setsar=sar=1,scale_vaapi=w=1920:h=1072`

which doesn't work in ffmpeg 4.1+ without adding `format=nv12` to the scaling specs:
`-vcodec h264_vaapi -vf format=nv12|vaapi,hwupload,setsar=sar=1,scale_vaapi=w=1920:h=1072:format=nv12`

VideoStation would have to change the way it constructs the ffmpeg args for transcoding, but we can't fix the code in VideoStation. Instead I set nv12 to be the default scaling format.

Disclaimer: I only tested it on DS918+ (apollo lake, x86_64)

Example of the error in ffmpeg 4.2.4:
[...]
Stream mapping:
  Stream #0:0 -> #0:0 (hevc (native) -> h264 (h264_vaapi))
  Stream #0:1 -> #0:1 (eac3 (native) -> mp3 (libmp3lame))
Press [q] to stop, [?] for help
Incompatible pixel format 'yuv420p' for codec 'h264_vaapi', auto-selecting format 'vaapi_vld'
[h264_vaapi @ 0xabf340] No usable encoding profile found.
Error initializing output stream 0:0 -- Error while opening encoder for output stream #0:0 - maybe incorrect parameters such as bit_rate, rate, width or height
[libmp3lame @ 0xac1580] 4 frames left in the queue on closing
Conversion failed!

### DIFF
--- a/cross/ffmpeg/patches/040-nv12-default-format.patch
+++ b/cross/ffmpeg/patches/040-nv12-default-format.patch
@@ -1,0 +1,11 @@
+--- a/libavfilter/vf_scale_vaapi.c
++++ ./libavfilter/vf_scale_vaapi.c
+@@ -206,7 +206,7 @@ static const AVOption scale_vaapi_options[] = {
+     { "h", "Output video height",
+       OFFSET(h_expr), AV_OPT_TYPE_STRING, {.str = "ih"}, .flags = FLAGS },
+     { "format", "Output video format (software format of hardware frames)",
+-      OFFSET(output_format_string), AV_OPT_TYPE_STRING, .flags = FLAGS },
++      OFFSET(output_format_string), AV_OPT_TYPE_STRING, {.str = "nv12"}, .flags = FLAGS },
+     { "mode", "Scaling mode",
+       OFFSET(mode), AV_OPT_TYPE_INT, { .i64 = VA_FILTER_SCALING_HQ },
+       0, VA_FILTER_SCALING_NL_ANAMORPHIC, FLAGS, "mode" },


ymartin59:
@th0ma7 What is your opinion about this proposal?

th0ma7:
I think this it's a rather interesting approach.  Just wonder what the side effects may end-up being: would this only affect vaapi hevc compatible platforms?

Question is, should the patch potentially apply only for intel x64 such as `cross/ffmpeg/patches/x64_ARCHES/040-nv12-default-format.patch` (not sure if my previous changes to `spksrc.patch.mk` allows that).

@jonozzz are you aware of this thread: https://gist.github.com/BenjaminPoncet/bbef9edc1d0800528813e75c1669e57e ?  Perhaps a package could be made available to get feedbacks from people who may have encountered the issue and get confirmation of it working or not on their platform?  Would you be up to do that?

jonozzz:
The truth is I don't know what other platforms are affected.

I wasn't aware of that thread, but I knew about the ffmpeg wrapper script trick. I tried to avoid that but that's another valid option.
I see that he's rewriting the `-vf` argument completely if `vaapi` is used and adding that `nv12` format which is required in ffmpeg 4.1+. I haven't tried that script but I bet it'll fix the transcoding just by looking at the code.

```
	-vf)
		shift
		if [ "$hwaccel" = "vaapi" ] && [ "$vcodecprofile" = "Main 10" ]; then
			scale_w=`echo "${1}" | sed -e 's/.*=w=//g' | sed -e 's/:h=.*//g'`
			scale_h=`echo "${1}" | sed -e 's/.*:h=//g'`
			if let ${scale_w} AND let ${scale_h}; then
				argsnew+=("-vf" "scale_vaapi=w=${scale_w}:h=${scale_h}:format=nv12,hwupload,setsar=sar=1")
			else
				argsnew+=("-vf" "scale_vaapi=format=nv12,hwupload,setsar=sar=1")
			fi
```

I thought I could share my quick workaround patch for others who noticed this issue as well. I don't have the resources to fully test it on other platforms.

th0ma7:
Thnx @jonozzz, indeed, I think that @BenjaminPoncet script is providing exactly what you where trying to address.  With that in mind I rather not be further modifying `ffmpeg` to try alleviate VideoStation.  Also thinking that DSM-7 is on the way which I believe might address that as well.

Again thnx for your contribution but for now I think it is addressed elsewhere.  Feel free to contribute again at a later time, cheers!

PR #4135, opened 2020-08-16
python3: fixed endianness detection for ppc853x architecture

smaarn:
_Motivation:_  Fixing detection of endianness for cross compiled python in case of big endian architecture
_Linked issues:_  #3944 

This actually addresses two concerns:

* the `sys.byteorder` attribute seemed to be invalid (it turns out that this information is actually not valid, even when using `crossenv` since that attribute is actually hardcoded in the python executable itself)
* borgbackup was actually trying to do something clever depending on that attribute to pass flags which were actually already appropriately defined and provided by python headers

### Checklist

### DIFF
--- a/cross/borgbackup/patches/001-remove-useless-endianness-logic.patch
+++ b/cross/borgbackup/patches/001-remove-useless-endianness-logic.patch
@@ -1,0 +1,25 @@
+Disables pseudo logic made to enforce setting of macros which are actually already defined in corresponding headers.
+
+--- setup.py
++++ setup.py
+@@ -804,20 +804,16 @@ if not on_rtd:
+                                                system_prefix=libb2_prefix, system=libb2_system,
+                                                **crypto_ext_kwargs)
+ 
+-    msgpack_endian = '__BIG_ENDIAN__' if (sys.byteorder == 'big') else '__LITTLE_ENDIAN__'
+-    msgpack_macros = [(msgpack_endian, '1')]
+     msgpack_packer_ext_kwargs = dict(
+         sources=[msgpack_packer_source],
+         include_dirs=include_dirs,
+         library_dirs=library_dirs,
+-        define_macros=msgpack_macros,
+         language='c++',
+     )
+     msgpack_unpacker_ext_kwargs = dict(
+         sources=[msgpack_unpacker_source],
+         include_dirs=include_dirs,
+         library_dirs=library_dirs,
+-        define_macros=msgpack_macros,
+         language='c++',
+     )
+ 


cross/python3/patches/ppc853x/007-fix-endianness-detection.patch
```diff
@@ -0,0 +1,12 @@
+Fixes the endianness detection by forcing it to be big endian
+--- configure.ac
++++ configure.ac
+@@ -4734,7 +4734,7 @@ else
+ fi
+ 
+ # check for endianness
+-AC_C_BIGENDIAN
```
smaarn, line 8: Here the issue is that this macro actually *runs* compiled code to see the output.
Now, since we're running (in docker case at least) in a little endian environment it says... "little endian".

The test to validate the outcome is to run `python -c "import sys; print(sys.byteorder)`

cross/python3/Makefile
```diff
@@ -24,6 +24,13 @@ ADDITIONAL_CFLAGS = -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -L $(STAGING_IN
 CONFIGURE_ARGS  = --enable-shared --enable-ipv6 --without-ensurepip --enable-loadable-sqlite-extensions PGEN_FOR_BUILD=$(HOSTPGEN)
 CONFIGURE_ARGS += ac_cv_buggy_getaddrinfo=no ac_cv_file__dev_ptmx=no ac_cv_file__dev_ptc=no ac_cv_have_long_long_format=yes
 
+ifeq ($(findstring $(ARCH), qoriq),$(ARCH))
```
smaarn, line 27: At this stage the point of this was to, at least, fix qoriq version.

I would need a little bit more digging to understand which platforms are actually big endian and, maybe, at this stage it would be wise to explicitely define this hack at the level of the TC since, then, it would fix the usage of the standard `AC_C_BIGENDIAN` macro of autoconf which seems to not work in our cases :( But then the root cause may also be an invalid toolchain configuration.

cross/borgbackup/patches/001-remove-useless-endianness-logic.patch
```diff
@@ -0,0 +1,25 @@
+Disables pseudo logic made to enforce setting of macros which are actually already defined in corresponding headers.
+
+--- setup.py
++++ setup.py
+@@ -804,20 +804,16 @@ if not on_rtd:
+                                                system_prefix=libb2_prefix, system=libb2_system,
+                                                **crypto_ext_kwargs)
+ 
+-    msgpack_endian = '__BIG_ENDIAN__' if (sys.byteorder == 'big') else '__LITTLE_ENDIAN__'
```
smaarn, line 9: It turns out that all this logic is actually useless and conflicting with the headers already provided by the cross compiling environment. Thus it should be safe to remove it.

Additionally the `sys.byteorder` information is actually stored in native resources and, as such, are not relevant when cross compiling.

cross/python3/patches/006-allow-forced-endianness-detection.patch
```diff
@@ -0,0 +1,19 @@
...
+@@ -4734,7 +4734,13 @@ else
+ fi
+ 
+ # check for endianness
+-AC_C_BIGENDIAN
++if test "$ac_cv_c_bigendian" = yes
++then
++   AC_DEFINE(WORDS_BIGENDIAN, 1, [Define this if your platform is using big endian byte ordering])
++else
++   AC_C_BIGENDIAN
```
smaarn, line 14: It could be worth checking if this is still relevant.

PR #4142, opened 2020-08-21
Toolchain code duplication removal

th0ma7:
_Motivation:_  Almost every toolchain has a `myFix` function to adjust `libdir=` variable in `*.la` files and adjust permissions.  With newly added toolchains for DSM-6.2 and DSM-6.2.2 and forthcoming DSM-7+ the idea here is to limit duplication of code to the bare minimum.
_Linked issues:_  N/A

### Checklist
Tried out a few toolchains and confirmed the following afterwards:
```
$ cd spksrc/toolchains/syno-armada38x-5.2
$ make
$ find . -type f -name '*.la' -exec strings {} \; | grep -i ^libdir=
```
or test the resulting directory or symlink:
```
$ find . -type f -name '*.la' -exec strings {} \; | grep -i ^libdir= | awk -F "'" '{print $2}' | while read libdir; do [ -f $libdir -o -s $libdir ] && echo "$libdir - OK" || echo "$libdir - ERROR"; done
```
update test to confirm `include` & `lib` path with added `work/tc_vars.mk`
```
$ find . -type f -name '*.la' -exec strings {} \; | grep -i ^libdir= | awk -F "'" '{print $2}' | while read libdir; do [ -f $libdir -o -s $libdir ] && echo "$libdir - OK" || echo "$libdir - ERROR"; done; [ -s $(grep ^CPPFLAGS work/tc_vars.mk | awk -F'-I' '{print $2}' | cut -f1 -d' ')/linux/version.h ]; echo "linux/version.h - OK" || echo "linux/version.h - ERROR"; [ -s $(grep ^LDFLAGS work/tc_vars.mk | awk -F'-I' '{print $2}' | cut -f1 -d' ')libc.so.[1-9] ]; echo "lib/libc.so - OK" || echo "lib/libc.so - ERROR";
```
and to test all toolchains (in progress):
```
$ cd spksrc/toolchains/
$ for dir in $(ls -1d syno-*); do cd $dir; make clean; make; find . -type f -name '*.la' -exec strings {} \; | grep -i ^libdir= | awk -F "'" '{print $2}' | while read libdir; do [ -f $libdir -o -s $libdir ] && echo "$libdir - OK" || echo "$libdir - ERROR"; done; cd -; done
```
updated test against all toolchains including confirmation of `include` & `lib` path with added `work/tc_vars.mk`
```
$ cd spksrc/toolchains/
$ for dir in $(ls -1d syno-*); do cd $dir; make clean; make; find . -type f -name '*.la' -exec strings {} \; | grep -i ^libdir= | awk -F "'" '{print $2}' | while read libdir; do [ -f $libdir -o -s $libdir ] && echo "$libdir - OK" || echo "$libdir - ERROR"; done; [ -s $(grep ^CPPFLAGS work/tc_vars.mk | awk -F'-I' '{print $2}' | cut -f1 -d' ')/linux/version.h ]; echo "linux/version.h - OK" || echo "linux/version.h - ERROR"; [ -s $(grep ^LDFLAGS work/tc_vars.mk | awk -F'-I' '{print $2}' | cut -f1 -d' ')libc.so.[1-9] ]; echo "lib/libc.so - OK" || echo "lib/libc.so - ERROR"; cd -; done
```


hgy59:
@th0ma7 sorry for my late response, I was very busy for some weeks.

I did similar redesign of toolchain builds. I did not only remove the myFix target, but also removed the definitions of TC_PREFIX, TC_TARGET, TC_CFLAGS. TC_CPPFLAGS, TC_LDFLAGS too and used only TC_BASE_DIR and two new variables TC_LIBDIR (with default of $(TC_BASE_DIR)/sys-root/lib) and TC_INCDIR (with default of $(TC_BASE_DIR)/sys-root/usr/include)
I did this redesign 4 months ago, but never commited it 😞 

You may have a look, as I just rectified this as new branch [hgy59:redesign_tc_builds](https://github.com/hgy59/spksrc/tree/redesign_tc_builds)

The relevant part of `spksrc.tc-fix.mk` looks like this:
```

fix_msg:
	@$(MSG) "Fixing access rights and libtool files for $(NAME)"

pre_fix_target: fix_msg

fix_target:  $(PRE_FIX_TARGET)
	@chmod -R u+w $(WORK_DIR)
	@if [ ! -d "$(WORK_DIR)/$(TC_BASE_DIR)/$(TC_LIBDIR)" ]; then \
	  echo "libdir does not exist: $(WORK_DIR)/$(TC_BASE_DIR)/$(TC_LIBDIR)" ; \
	  exit 1; \
	fi
	@find $(WORK_DIR)/$(TC_BASE_DIR) -type f -name '*.la' -exec sed -i -e "s|^libdir=.*$$|libdir='$(WORK_DIR)/$(TC_BASE_DIR)/$(TC_LIBDIR)'|" {} \;
```

th0ma7:
> @th0ma7 sorry for my late response, I was very busy for some weeks.

@hgy59 No issue at all, this isn't a pressing PR anyway :)

> I did similar redesign of toolchain builds. I did not only remove the myFix target, but also removed the definitions of TC_PREFIX, TC_TARGET, TC_CFLAGS. TC_CPPFLAGS, TC_LDFLAGS too and used only TC_BASE_DIR and two new variables TC_LIBDIR (with default of $(TC_BASE_DIR)/sys-root/lib) and TC_INCDIR (with default of $(TC_BASE_DIR)/sys-root/usr/include)

That's rather interesting... *sigh*, I should have thought of that, its obvious, good catch!

> You may have a look, as I just rectified this as new branch [hgy59:redesign_tc_builds](https://github.com/hgy59/spksrc/tree/redesign_tc_builds)
> 
> The relevant part of `spksrc.tc-fix.mk` looks like this:
> 
> ```
> fix_msg:
> 	@$(MSG) "Fixing access rights and libtool files for $(NAME)"
> 
> pre_fix_target: fix_msg
> 
> fix_target:  $(PRE_FIX_TARGET)
> 	@chmod -R u+w $(WORK_DIR)
> 	@if [ ! -d "$(WORK_DIR)/$(TC_BASE_DIR)/$(TC_LIBDIR)" ]; then \
> 	  echo "libdir does not exist: $(WORK_DIR)/$(TC_BASE_DIR)/$(TC_LIBDIR)" ; \
> 	  exit 1; \
> 	fi
> 	@find $(WORK_DIR)/$(TC_BASE_DIR) -type f -name '*.la' -exec sed -i -e "s|^libdir=.*$$|libdir='$(WORK_DIR)/$(TC_BASE_DIR)/$(TC_LIBDIR)'|" {} \;
> ```

There is a major issue with `$(WORK_DIR)/$(TC_BASE_DIR)/$(TC_LIBDIR)` as it will not work on some DSM / arch versions which is the main reason I decided to rather opt for a new variable called `TC_SYSROOT` to manage all cases:
```
TC_SYSROOT = $(TC_BASE_DIR)
TC_SYSROOT = $(TC_BASE_DIR)/libc
TC_SYSROOT = $(TC_BASE_DIR)/sys-root
TC_SYSROOT = $(TC_BASE_DIR)/sysroot
```
Where `spksrc.tc-fix.mk` uses the following to change the `libdir`:
 ```
@find $(WORK_DIR)/$(TC_BASE_DIR) -type f -name '*.la' -exec sed -i -e "s|^libdir=.*$$|libdir='$(WORK_DIR)/$(TC_BASE_DIR)/$(TC_SYSROOT)/lib'|" {} \;
```
But now, what's nice is that the conjunction of yours + mines makes this really lean and clean and quite bullet-proof.  We really need to merge our code together.

Would it be possible for you to apply your changes on-top of mine while re-using my `TC_SYSROOT` to build your `TC_LIBDIR` variable leading to your massive cleanup of `TC_PREFIX`, `TC_TARGET`, `TC_CFLAGS`, `TC_CPPFLAGS`, `TC_LDFLAGS` and usage of only `TC_BASE_DIR` ?

**Comment by th0ma7** [comment]
@hgy59 so we don't duplicated the work, I may have a few cycles available so I'll give it a shot at including your changes on-top of mine within this PR.

**Comment by th0ma7** [comment]
@ymartin59 and @hgy59 the updated patches makes this way more intrusive and thus much longer to test-build but it does make the `toolchains` section really much nicer to work with.

@hgy59 I've made much mileage from your changeset although I did used a slightly different approach reducing even further the artifacts in each `syno-arch-*` directories.  Thnx again for your proposals, much appreciated and comments welcomed.

I'll be investigating the remaining bits but my guess is that there is not much that can be done to fix them.  For instance anything depending to `erlang` such as `ejabberd` is broken by default since way before this changeset.

Please have a look in terms of review.  Bulk of the changes are :
- `toolchains/syno*` 
- `mk/spksrc/tc*.mk`
- Domino effect due to keeping only the `TC_PREFIX` variable and removing the trailing `-` from the `tc_vars.mk` environment leading to changing a few `cross/*` packages to re-add the `-` when needed.
- Now also generates a `tc_vars.mk` in `toolchains/syno-*/work` directories helping quite much to investigate issues

**Comment by th0ma7** [comment]
BTW, sourceforge is, again down leading to additional build failures.

**Comment by th0ma7** [comment]
From the failures:
- ejabberd: (x64-6.1) FAILED - Does not build as `earlang` is broken upstream (needs manual intervention to work)
- ffsync: (x64-6.1) FAILED
- rutorrent: (x64-6.1) FAILED
- squidguard: (x64-6.1) FAILED - Does not build on master, version 6.10 of `squidclamav` is not available anymore
- tt-rss: (x64-6.1) FAILED - Does not build on master, fails to download `ERROR 404: Not Found.`

**Comment by th0ma7** [comment]
Last but not least update related to the `TC_CFLAGS` variable: As it now contains the "extra" CFLAGS such as `-mcpu=8548 -mhard-float -mfloat-gprs=double` I found it more appropriate to be called accordingly:
1. Due to its content I renamed the variable such as `TC_EXTRA_CFLAGS` and thus discarding `TC_CFLAGS` entirely.
2. Packages `c-ares` and `nmap` are making a reference to them.  After testing it seems they need exactly that e.g. `TC_EXTRA_CFLAGS` so I adjusted `Makefile` accordingly.

**Comment by ymartin59** [comment]
@th0ma7 Great job. All these renaming are relevant but... so many diff/files I have to review now...

**Comment by publicarray** [comment]
Hi, I appreciate that I was invited for review but some of the makefiles are pure magic to me 😄. Reducing duplicate code & variables is awesome and much appreciated! 👍 From a glance these changes look sane and well thought out.

**Comment by th0ma7** [comment]
Thnx @publicarray , I know it isn't an easy one to review.

Interestingly yesterday I heard back from synology and `geminilake` kernel sources and other related source code got release under the 6.2.2 umbrella.  As such we now have a full toolchain & kernel for native `geminilake` builds.

@ymartin59 I know this is quite a bite to go through but let me know once it is reviewed or if there anything I can do to help.  I prefer making sure my pending PR are all merged before moving onto other things.

File: cross/ffmpeg/Makefile [hunk 1]
```diff
@@ -24,7 +24,7 @@ LICENSE  = GPLv2
 
 NASM_PATH=$(WORK_DIR)/../../../native/nasm/work-native/install/usr/local/bin
 
-CONFIGURE_ARGS = --target-os=linux --cross-prefix=$(TC_PATH)$(TC_PREFIX) --prefix=$(INSTALL_PREFIX)
+CONFIGURE_ARGS = --target-os=linux --cross-prefix=$(TC_PATH)$(TC_PREFIX)- --prefix=$(INSTALL_PREFIX)
```
ymartin59, line 27: I would avoid these TC_PREFIX "meaning" changes... why not keep dash in variable value.

mk/spksrc.tc.mk
```diff
@@ -38,51 +39,44 @@ include ../../mk/spksrc.extract.mk
...
 	@echo TC_ENV += CPPFLAGS=\"$(CPPFLAGS) $$\(ADDITIONAL_CPPFLAGS\)\"
 	@echo TC_ENV += CXXFLAGS=\"$(CXXFLAGS) $$\(ADDITIONAL_CXXFLAGS\)\"
 	@echo TC_ENV += LDFLAGS=\"$(LDFLAGS) $$\(ADDITIONAL_LDFLAGS\)\"
-	@echo TC_CONFIGURE_ARGS := --host=$(TC_TARGET) --build=i686-pc-linux
-	@echo TC_TARGET := $(TC_TARGET)
-	@echo TC_PREFIX := $(TC_PREFIX)-
-	@echo TC_PATH := $(WORK_DIR)/$(TC_BASE_DIR)/bin/
+	@echo TC_CONFIGURE_ARGS := --host=$(TC_PREFIX) --build=i686-pc-linux
+	@echo TC_PREFIX := $(TC_PREFIX)
+	@echo TC_PATH := $(WORK_DIR)/$(TC_PREFIX)/bin/
```
ymartin59, line 72: Does TC_PREFIX changes "meaning" from that point?

mk/spksrc.tc.mk
```diff
@@ -8,23 +8,24 @@ include ../../mk/spksrc.directories.mk
...
-DIST_EXT      = $(TC_EXT)
+DISTRIB_DIR      = $(TOOLCHAINS_DIR)/$(TC_VERS)
+DIST_FILE        = $(DISTRIB_DIR)/$(LOCAL_FILE)
+DIST_EXT         = $(TC_EXT)
+TC_LOCAL_VARS_MK = $(WORK_DIR)/tc_vars.mk
 
 #####
 
-RUN = cd $(WORK_DIR)/$(TC_BASE_DIR) && env $(ENV)
+RUN = cd $(WORK_DIR)/$(TC_PREFIX) && env $(ENV)
```
ymartin59, line 28: Is TC_PREFIX really a folder?

ymartin59 (changes requested):
I had difficulties to understand "what is what" in variables.
From my point of view, it is more explicit to keep previous variable names, even if they are built from single TC_TARGET "by default" - allowing to override for some "exceptional" toolchain.
```
TC_TARGET = x86_64-pc-linux-gnu
TC_PREFIX = "$(TC_TARGET)-"
TC_BASE_DIR = $(TC_TARGET)
```
So $(TC_TARGET) is parameter for compiler (a target), $(TC_PREFIX)gcc is executable name (from a prefix) and $(TC_BASE_DIR)/lib points to compiler libraries (from a directory)

From my point of view, objective is to avoid useless copy-pasted lines in each toolchain, but I do not mind if there are many variables set by framework in a single place. 
What do you think about it?

cross/ffmpeg/Makefile
th0ma7: Excellent catch.  I'll come up with another approach.

mk/spksrc.tc.mk
th0ma7: That the whole point... I tried to reduce the amount of variable duplication to the maximum and might have stretched `TC_PREFIX` a bit too much.  I'll readjust.

th0ma7:
@ymartin59 based on your comment on PR #4143 I manage to remove additional variable from most `Makefile`.
Other changes you request above to come shortly.

cross/ffmpeg/Makefile
th0ma7: As always your comments are valuable, thnx @ymartin59 .
On this, I've reverted to the previous behavior and re-declared `TC_TARGET`.
If `TC_PREFIX` is not set it becomes `TC_PREFIX = $(TC_TARGET)-` ... 
Note that it is never manually set in any of the `toolchains/*/Makefile`
And this avoids all the small changes to reflect the missing `-` reverting back multiple files to its original state.

mk/spksrc.tc.mk
th0ma7: As mentioned on comment above: re-declared `TC_TARGET` and reverted `TC_PREFIX` to original intended behaviour.

mk/spksrc.tc.mk
th0ma7: Same as above, but now using `TC_TARGET` for directory purpose as `TC_BASE_DIR` was never different anyway.

th0ma7:
@ymartin59 I've made all the changes you requested and reduced amount of files from 190+ down to 177 by re-introducing the `TC_TARGET` variable and restoring `TC_PREFIX` originally intended behavior with the trailing `-`.

I'll let github-action run overnight and do a few tests of my own to confirm all is still good for a merge shortly.
Thnx again for reviewing.

th0ma7:
So migrating the download URL is the last stretch I was able to migrate as common variable.
Should also be able in case of URL changes without having to edit 100+ files.
This make the `Makefile` really neat and hopefully really easy to maintain in the long run (and to incorporate DSM7 shortly).

Here's an example:
```
$ more syno-apollolake-6.2.2/Makefile 
TC_ARCH = apollolake
TC_VERS = 6.2.2
TC_KERNEL = 4.4.59+

TC_DIST = apollolake-gcc493_glibc220_linaro_x86_64-GPL
TC_DIST_SITE_PATH = Intel%20x86%20Linux%204.4.59%20%28Apollolake%29

TC_TARGET = x86_64-pc-linux-gnu
TC_SYSROOT = $(TC_TARGET)/sys-root

include ../../mk/spksrc.tc.mk
```

th0ma7:
@ymartin59 as this PR is already quite invasive, I could do the following changes right away to better align with toolkit PR #4143:
- move `toolchains` directory to `toolchain`
- rename all `TC_` variables to `TOOLCHAIN_`

Let me know if you see value in doing so, and right away or into a subsequent PR or not at all.
Otherwise I went through all my testing and PR looks good for a merge.
Thnx in advance.

ymartin59 (approved):
Great. I would not do more refactoring at the moment, to focus on package delivery

toolchains/syno-armada375-5.1/Makefile
```diff
@@ -1,21 +1,20 @@
...
-TC_CXXFLAGS = -I$(WORK_DIR)/$(TC_BASE_DIR)/$(TC_PREFIX)/libc/usr/include -mhard-float -mfpu=vfpv3
-TC_LDFLAGS = -L$(WORK_DIR)/$(TC_BASE_DIR)/$(TC_PREFIX)/libc/usr/lib -mhard-float -mfpu=vfpv3
+PRE_FIX_TARGET = myFix
 
 include ../../mk/spksrc.tc.mk
+
+.PHONY: myFix
+myFix:
+	@$(MSG) "Fixing libtool files for $(NAME)"
+	@mv $(WORK_DIR)/armv7-marvell-linux-gnueabi-hard $(WORK_DIR)/$(TC_TARGET)
```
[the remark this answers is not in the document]
th0ma7: Literally this is the only cruft left-over from the refactoring that was outstanding that I decided to leave as is considering this is against DSM-5.1... and you found it!  just wow!

So if I understand this correctly you would like to see the `myFix` renamed to `syno_armada375_fix`?  I can certainly do that it's just that I'm not sure that this is exactly what you meant.

th0ma7:
I'll do another full round of testing to confirm all is good.  While looking in detail I've found from last run was with the download URL centralization:
1. centralizing the URL allowed catching a few files from wrong DSM versions being download in a some cases.  Update `digests` files are now in place.
2. the need to manage the `SRM` vs `DSM` toolchain type (new var `TC_TYPE`) for download URL adjustment
3. new arches only available from 6.2.2+ that download URL are backported to fake 6.2 that needed exceptions

All theses last bits should now be solved but will do another in-depth re-download + testing to confirm all is good.

toolchains/syno-armada375-5.1/Makefile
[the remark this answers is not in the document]
ymartin59: Exactly. I always fear Makefile target collision when using so many includes

th0ma7:
Confirmed, re-processed it entirely and no more errors, now merging.

Issue #4104, opened 2020-08-05
[Package Request] Update znc 1.7.5 -> 1.8.1


worstje:
### Package to update

_Package Name:_ znc
_Package Version:_ 1.8.1

### Motivation

This package improves on several SSL related things, fixes a few (crashing) bugs, improves Python support and perhaps most relevant includes a change that prevents passwords from leaking into unsecure logging facilities.

[Changelog znc 1.8.0](https://wiki.znc.in/ChangeLog/1.8.0)
[Changelog znc 1.8.1](https://wiki.znc.in/ChangeLog/1.8.1)

### Addendum

I just noticed while editing user settings in 1.7.5 that it says...

> ZNC is compiled without encodings support. ICU is required for it.
> * Don't ensure any encoding at all (legacy mode, not recommended)

as well as

> ZNC is compiled without i18n support

If znc ends up being updated, it would be nice to have this support added in, too.

PR #4148, opened 2020-08-24
znc: update to 1.8.1

worstje:
*Note: I wanted to apply the __status/help-wanted__ label to this PR as the contributing guidelines recommend, but I don't think I have the access for that. Either that or I am blind and failing at finding the spot where to apply the label!*

_Motivation:_  I wanted the latest ZNC 1.8.1 with Python support on my device. So I worked on making it happen.
_Linked issues:_ My original request can be found at issue #4104.

### Checklist
- [🤷] Build rule `all-supported` completed ~~successfully~~ **partially**.

```
root@synocommunity-spksrc1:/spksrc/spksrc/spk/znc# ls ../../packages/ -lh
total 245M
-rw-r--r-- 1 root root 18M Aug 24 16:05 znc_aarch64-6.1_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:05 znc_alpine-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:05 znc_armada370-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:05 znc_armada375-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:05 znc_armada38x-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:06 znc_armadaxp-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 17M Aug 24 16:06 znc_comcerto2k-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:06 znc_dakota-1.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:06 znc_evansport-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:06 znc_ipq806x-1.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:06 znc_monaco-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 18M Aug 24 16:06 znc_northstarplus-1.2_1.8.1-1.spk
-rw-r--r-- 1 root root 20M Aug 24 16:06 znc_qoriq-6.2.2_1.8.1-1.spk
-rw-r--r-- 1 root root 19M Aug 24 16:06 znc_x64-6.2.2_1.8.1-1.spk
```

It took me a long time to compile these (the above list is after I emptied the packages directory and re-ran the command to make sure it reflected `make all-supported`, hence the similar timestamps.). I think I saw a few errors along the way, but I was too slow in figuring out which or where. I am not sure which packages are necessary to be supported, or how to fix the missing packages. This time is best spent by someone who actually knows what they are doing, as opposed to an idiot novice like me researching and shoddily reinventing the wheel.

- [✔️] Package upgrade completed successfully

I have done this twice (once with an earlier build, and once more recently with a fresh config made for this purpose), and it worked fine. I'd have done more tests if I didn't accidentally wipe my config at one point and started to care considerably less.

- [✔️] New installation of package completed successfully

Wiped my config by accident at one point at which point I stopped caring about preserving it, so numerous fresh installations are confirmed to work.

### Other questions that might come up

**Why ICU 67 and not ICU56 that is on your own device?**

Because I had no clue how to build against it or any certainty that would work for other devices! And also because it covers Unicode 8.0 with the most recent being Unicode 13.0, I felt that pulling in this dependency and giving other packages that want to benefit from worldly progress the opportunity to do so a worthwhile investment.

I expect that the biggest issues with missing cross-compilation targets can be blamed on this new package. I have tried including some targeting stuff I saw in other packages to improve it, but I am completely out of my depth on all this stuff, so I can only ask the experts to take care of the final mile here.

**Should SPK_DEPENDS be set for ZNCs dependency on python3?**

Since modpython is an optional global module to load for ZNC users, I think this could probably be dropped with a 'soft' dependency instead. But because I don't know nearly enough of anything, I decided to just copy what I saw done elsewhere and include the SPK_DEPENDS for now.

**Why not remove the useless modules (like sample and samplewebapi) from the PLIST?**

I considered it. They are useless for direct-use as far I can tell. But they were there already and my aim is to upgrade the versions, not to apply personal preference on what builtin modules are good enough to ship. The slope can easily become quite slippery.

I can remove pointless modules easily if that is the consensus.

**Has it been tested?**

Yes, if you consider my audience of one DS418play. I have run various builds of this package on my DS418play for the past two weeks and even developed a custom Python module along the way to make sure it all worked. (And I also need that module eventually...) I ran into a bug at one point, but that was quite specific, several years old Python bug, and thankfully not something to do with the way Python or Znc are built.

**Does anything else need doing?**

Kind of?

For future versions of ZNC, the project has removed the GNU configure script and fully switched to CMake. Unfortunately, I could not get everything properly integrated using CMake in this version, so future versions will still need some elbow grease. But for znc 1.8.1 the legacy script is still present and works completely fine, so it should not be a blocker for this particular PR.

Additionally, there is one more optional i18n dependency for ZNC, but at this point, I simply couldn't be bothered trying to include support for that as well. Plain english works just fine for me.

**How much more effort will you put into this PR?**

I won't put in any more time figuring out how to make things build on targets it is not building for. I am new to __all of this__, and I think the energy I put in preparing this PR over the past two weeks already speaks for itself. That said, if any of you experts know what needs to change to make it build for things it fails to build for right now, I am happy to include those changes. (Hence allowing edits by maintainers!)

Simply put: I make this PR because I want to give back where I have benefited in the past, but I lack the expertise (and at this point the energy and will) that would allow me to solve any remaining issues with the PR. I see myself cooperating to make this PR happen, but not as a leader who will make it happen. 😄


native/libicu/Makefile
```diff
@@ -0,0 +1,26 @@
...
+GNU_CONFIGURE = 1
+
+#POST_INSTALL_TARGET = myPostInstall
+
+include ../../mk/spksrc.native-cc.mk
+
+#.PHONY: myPostInstall
+#myPostInstall:
+#	sed -i -e 's#/usr/local#/usr/local/mono#g' $(WORK_DIR)/install/usr/local/bin/*
+#	sed -i -e 's#/usr/local#/usr/local/mono#g' $(WORK_DIR)/install/usr/local/etc/mono/config
```
ymartin59, line 26: Please discard commented lines

spk/znc/Makefile
```diff
@@ -1,18 +1,19 @@
 SPK_NAME = znc
-SPK_VERS = 1.7.5
-SPK_REV = 13
+SPK_VERS = 1.8.1
+SPK_REV = 1
```
ymartin59, line 3: Revision 14 is required.

ymartin59 (changes requested):
Thanks for this contribution
I have minor improvements to propose

ymartin59:
@worstje Great job. About ppc compilation failure, it probably comes from too-old compiler for source code - in that case `UNSUPPORTED_ARCHS` variable is added in spk Makefile: https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables

worstje:
@ymartin59 I'll apply the suggested changes later today when I have the time for it. Good catch on the commented lines I missed out on while cleaning up, and apologies about misinterpreting the meaning of `SPK_REV`. I had interpreted it as an extra build number to allow new spk's for the same version.

**What do you reckon is the best way to handle the new Python3 dependency?** I don't understand the build/dependency system well enough, but I think that it might be useful as a 'soft' spk dependency in case people don't want the 50mb+ required package they won't use? From ZNC's runtime perspective, Python is a completely optional module. If minimalism for security reasons is something users want to keep in mind, it is worth considering. But at the same time I am not sure how that would interfere with upgrades and how many problems that could create in the future as Python and/or ZNC are updated further and them not being in proper sync.

ymartin59:
@worstje About "soft dependency", it depends on code capability itself. I propose you comment `SPK_DEPENDS` from package Makefile and check how ZNC behaves if Python 3 package is lacking on system (and of course module is disabled in configuration)

worstje:
Rebased to merge clean. While doing so, I (finally) acted on issues mentioned above...

**Leftover commented lines**: Removed. ✔️ 

**SPK_VERSION**: Updated to revision 15 since current master already bumped it once. ✔️ 

**Build failure for platform ppc853x**: It was recommended I add the failing platform to the **UNSUPPORTED_ARCHS** option in the spk Makefile, but it is already present in `spk/znc/Makefile`. As such, I am not sure how I can fix this buildbot issue. ❌ 

**SPK_DEPENDS on Python3**: Left it in, except updated to the most recent revision.

This is because I am not sure why my manually-built znc refuses to run for me since I updated Python3, but this is a clear indicator that this build is very fragile. If znc quits hard over limping along with a module that gracefully failed to load, it is worth taking the extra precaution to make sure that the Python dependency is as clearly defined as possible, regardless of whether znc itself actually loads the python module. (Since the update to Python3 was to its openssl, it could merely be that the problem isn't Python itself but just the two openssl builds clashing... but it is still too harsh a punishment to an end-user who ends up with a broken service!) ✔️ 

Besides, disk space doesn't really come at a premium nowadays, so I think it is fine to demand one extra dependency even for users who don't need it. 😄

native/libicu/Makefile
worstje: Done.

spk/znc/Makefile
worstje: By the time I updated, it had to be revision 15. Done.

ymartin59 (approved):

PR #4151, opened 2020-08-25
Rev bump hassio

andyfowler:
I have no idea what i'm doing, but it seems like this package is compatible with geminilake arch and simply needs to be rebuilt. Running this to let github actions do the work, and I'll give the compiled version a test run

_Motivation:_  To see if @fredrike's hassio package can run on a DS920+ with a rebuild.

### Checklist


fredrike:
Their should be a requirement for DSM 6.2 too.

andyfowler:
Hi @fredrike! Thanks for putting this package together! So far, a test installation of the package as compiled here is working on my DS920+.

When you say a requirement for DSM 6.2, do you mean this would be an additional field in the `Makefile`? I couldn't find examples of other packages doing this. Happy to help where I can to get this merged.

fredrike:
I think it is `REQUIRED_DSM = 6.2` https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables

justinas-b:
Hey Gents, I have added `REQUIRED_DSM = 6.2` under https://github.com/SynoCommunity/spksrc/pull/4175

hgy59:
> 
> 
> I think it is `REQUIRED_DSM = 6.2` https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables

No this is not the case.

`REQUIRED_DSM` must be used only, when DSM Version is required for all archs. We should avoid the usage of REQUIRED_DSM anyway as it conflicts with REQUIRED_SRM (for routers).
As armv7 is used for DSM and SRM we have unresolvable conflicts when building for arch-armv7-6.1 and arch-armv7-1.2.

If we explicitly want to build for geminilake (or purley) we have to build with arch-geminilake-6.2 or arch-purley-6.2 but we do not need, as the generic arch-x64-6.1 can be installed on the x20 DSM series.

fredrike:
> > I think it is `REQUIRED_DSM = 6.2` https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables
> 
> No this is not the case.
> 
> `REQUIRED_DSM` must be used only, when DSM Version is required for all archs. We should avoid the usage of REQUIRED_DSM anyway as it conflicts with REQUIRED_SRM (for routers).
> As armv7 is used for DSM and SRM we have unresolvable conflicts when building for arch-armv7-6.1 and arch-armv7-1.2.
> 
> If we explicitly want to build for geminilake (or purley) we have to build with arch-geminilake-6.2 or arch-purley-6.2 but we do not need, as the generic arch-x64-6.1 can be installed on the x20 DSM series.

I don't quite follow. The hassio package does not really work if installed on anything less than 6.2

fredrike:
I think this PR should be closed as https://github.com/SynoCommunity/spksrc/pull/4173 super seeds it.

PR #4173, opened 2020-09-10
hassio: Remove support containers and images, require DSM 6.2

fredrike:
_Motivation:_  Attempt to get it out of BETA.
_Linked issues:_  https://github.com/SynoCommunity/spksrc/pull/3732#issuecomment-690119165

### Checklist


spk/hassio/Makefile
```diff
@@ -1,10 +1,12 @@
 SPK_NAME = hassio
-SPK_VERS = 20191224
+SPK_VERS = 20200910
 SPK_REV = 1
 SPK_ICON = src/$(SPK_NAME).png
 SPK_DEPENDS = "Docker>=18.09.0-0506"
 DSM_UI_DIR = app
 
+REQUIRED_DSM = 6.2
```
fredrike, line 8: @hgy59 Are you saying that this should be:
```suggestion
ARCH = dockerx64-6.2
```

ymartin59 (approved):

spk/hassio/Makefile
ymartin59: In fact, no. The "perfect way" would be to list all architecture without Docker available in `UNSUPPORTED_ARCHS` but from my point of view `SPK_DEPENDS` is far enough to prevent any mistake/trouble.

hgy59: ARCH=dockerx64
means that DSM is running in a docker container (aka DDSM)

fredrike:
Thanks for the merge!

Does this mean that the beta tag is removed and that it only is supported on dsm <=6.2

Issue #3400, opened 2018-07-26
Wallabag 403 Error After Installation


adan89lion:
### Setup
_Package Name:_ Wallabag
_Package Version:_ 2.2.3-2

_NAS Model:_ DS216play
_NAS Architecture:_ monaco x16 series (STM Monaco STiH412)
_DSM version:_ DSM 6.2-23739 Update 2

### Expected behavior
Open the package and show the login screen at yournas.com/wallabag/app

### Actual behavior
Shows 403 error at the webpage

### Steps to reproduce
_1._ Enables beta version opt-in and installs Wallabag package via SynoCommunity repository
_2._ Enters MariaDB 5 root password and password for the Wallabag user
_3._ Runs the install script by the package centre
_4._ Shows successful however cannot enter the package

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
```
No log produced at /usr/local/wallabag/share/var
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
2018/07/22 19:57:58     start radarr 20180303-6 End start-stop-status start ret$
2018/07/22 19:58:00     start radarr: start version 20180303-6 successfully, re$
2018/07/24 16:22:24     install wallabag 2.2.3-2 Begin preinst
2018/07/24 16:22:24     install wallabag 2.2.3-2 End preinst ret=[1]
2018/07/24 16:22:24     install wallabag 2.2.3-2 Begin /bin/rm -rf /volume2/@tm$
2018/07/24 16:22:36     install wallabag 2.2.3-2 End /bin/rm -rf /volume2/@tmp/$
2018/07/24 16:22:36     install wallabag 2.2.3-2 failed
2018/07/24 16:26:16     install wallabag 2.2.3-2 Begin preinst
2018/07/24 16:26:16     install wallabag 2.2.3-2 End preinst ret=[1]
2018/07/24 16:26:16     install wallabag 2.2.3-2 Begin /bin/rm -rf /volume2/@tm$
2018/07/24 16:26:35     install wallabag 2.2.3-2 End /bin/rm -rf /volume2/@tmp/$
2018/07/24 16:26:35     install wallabag 2.2.3-2 failed
2018/07/24 16:28:16     install wallabag 2.2.3-2 Begin preinst
2018/07/24 16:28:17     install wallabag 2.2.3-2 End preinst ret=[1]
[...]

Issue #3113, opened 2018-01-18
Wallabag : mySQL root password issue


pfml:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ Wallabag
_Package Version:_ 2.2.3-2

_NAS Model:_ DS411+
_NAS Architecture:_ Intel Atom D510
_DSM version:_ DSM 6.1.4-15217 Update 5

### Expected behavior
Try to install package wallabag on my Syno (php 5.6 & 7.0, MariaD B10 and WebStation enabled)

### Actual behavior
Got an error msg : "failed to install "wallabag". Incorrect MySQL root password"
Using phpMyAdmin and MariaDB root password, I'm able to create/delete account

### Steps to reproduce
_1._
_2._
_3._

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_

found no log in the mentioned places

#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_

in /var/log/messages :
```
2018-01-18T14:55:07+01:00 cezoffnas02 synoscgi_SYNO.Core.Package.Server_1_list[12263]: pkgcurltool.cpp:456 http://packages.missilehugger.com, Failed to request packages, httpResponseCode=404
2018-01-18T15:07:02+01:00 cezoffnas02 synoscgi_SYNO.Core.Package.Installation_1_install[2283]: pkginstall.cpp:466                                                             Failed to run preinst script for wallabag, 1

```
in /var/log/synopkg.log 
````
2018/01/18 15:07:02     install wallabag 2.2.3-2 Begin preinst
2018/01/18 15:07:02     install wallabag 2.2.3-2 End preinst ret=[1]
2018/01/18 15:07:02     install wallabag 2.2.3-2 Begin /bin/rm -rf /volume1/@tmp/pkginstall
2018/01/18 15:07:28     install wallabag 2.2.3-2 End /bin/rm -rf /volume1/@tmp/pkginstall ret=[0]
2018/01/18 15:07:28     install wallabag 2.2.3-2 failed
````

**Comment by MysteryWoMan** [comment]
Have you tried this? :
Log in on your Synology device using telnet or SSH (with your admin account).

```
mkdir /usr/syno/mysql
mkdir /usr/syno/mysql/bin 
ln -s /usr/bin/mysql /usr/syno/mysql/bin/mysql
```
or
```
mkdir /usr/syno/mysql
mkdir /usr/syno/mysql/bin
[...]

PR #3054, opened 2017-12-20
Updated wallabag package to 2.3.0

nicosomb:
_Motivation:_ wallabag 2.3.0 was released.

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

I don't know how Synology users can handle the update of wallabag: there are some database migrations to execute to upgrade wallabag from 2.2.x to 2.3.0. Any idea?

PR #4179, opened 2020-09-14
Wallabag: Update Wallabag to 2.3.8 and various fixes

publicarray:
* update Wallabag to 2.3.8
* incorrect mysql password fix
* various fixes
* migrated to generic service script

_Linked issues:_  #3113, #3400, #3054

*Build:* https://github.com/publicarray/spksrc/releases/tag/wallabag-2.3.8

### Checklist


ymartin59:
@publicarray Thanks for contribution. In fact I understand it is important to keep MySQL support because of previous version installation upgrade, but I would find more resource-effective to deliver package with PostgreSQL support by default, connected to DSM embedded version like for nextcloud PR #3853

In fact, I dream about a generic wizard process (shell script based) where user may select its database from a list (generated from available services in DSM): embedded postgresql, mariadb, sqlite, own package postgresql 11... with backup/restore support when uninstall/install. It would make sense for many applications...

spk/wallabag/Makefile
```diff
@@ -13,16 +13,14 @@ ADMIN_URL = /wallabag/web/
...
 WIZARDS_DIR = src/wizard/
-
-INSTALLER_SCRIPT = src/installer.sh
-SSS_SCRIPT       = src/dsm-control.sh
-
+SERVICE_SETUP = src/service-setup.sh
+STARTABLE = no
 
 INSTALL_DEP_SERVICES = apache-web mysql
 START_DEP_SERVICES = apache-web mysql
```
ymartin59, line 26: I would not enforce mysql dependency for users expecting to connect to embedded postgresql

publicarray:
Yes that is a good idea, How about we merge this PR first to fix the package (currently it's broken) and later I can work postgresql in a separate PR?

publicarray:
FYI [SQLite support is has been dropped](https://doc.wallabag.org/en/user/faq.html#why-has-sqlite-support-been-dropped)

ymartin59:
In fact my idea would have a "generic database support" in framework (like for service) that a package may re-use to avoid copy/paste... and sqlite may be useful for other application like ffsync

publicarray:
@ymartin59 I understand, but how about we do this one step at a time 😅 
1. Fix this package as is
2. add ability to select postgress, sqlite and redis (for parrallel importing suppot) for this package only
3. generate the nessesary files from the framework - also checkout https://help.synology.com/developer-guide/resource_acquisition/maria_db.html

ymartin59:
Yes I agree. I will do some testing on this PR and publish first.
I am aware about "resource acquisition" and I guess a shell-based wizard should be able to generate require file according to user's answer "but" it would be probably easier to do all operations (connection and DB creation) in a single place in pre_install step after wizard - when we expect to use our on PostgreSQL package version or SQLite when available...

ymartin59:
@publicarray I just discover GH action does not build non-arch package and package fails to build on my system... I still do not know why

ymartin59:
Found: ARCH should not be override in cross/wallabag

spk/wallabag/src/parameters.yml
```diff
@@ -1,34 +1,64 @@
 parameters:
     database_driver: pdo_mysql
+    database_driver_class: ~
     database_host: 127.0.0.1
     database_port: 3306
```
ymartin59, line 5: Bad luck. On my test system, I have installed both MariaDB 5 and 10... so port is 3307 now

ymartin59:
I have published package

emo1509:
I get following error when installing the package on my DS2018+ : Failed to install wallabag. See errors at /var/services/web/wallabag/install.log
[wallabag install.log](https://github.com/SynoCommunity/spksrc/files/5542052/wallabag.install.log)

publicarray:
From the log you need to check your database configuration. Do you have MariaDB installed? maybe you need to reset the password and try again?

```
Checked                  Status   Recommendation
 ------------------------ -------- ----------------------------------------------------------------------------------------------------------
  PDO Driver (pdo_mysql)   OK!
  Database connection      ERROR!   Can't connect to the database: An exception occured in driver: SQLSTATE[HY000] [2002] Connection refused
  Database version         OK!
  curl_exec                OK!
  curl_multi_init          OK!
```

emo1509:
I'm running MariaDB10 on Synology (port 3307). Do I need to create user & database first or will it be created by the package ?

publicarray:
Change your port to `3306` (this is the default Mariadb port but because Synology has 2 Mariadb versions they changed the port number so they can be run side by side.)

And the wallabag install script will create the database and user for you.

emo1509:
I have now also installed MariaDB 5 on my NAS which uses by default port 3306. I also have MariaDB 10 running because already created several databases in there for other applications. If both are running and I install Wallabag package , he creates user wallabag and database name (not yet the structure) wallabag in MariaDB 10 (which runs on port 3307). If I stop MariaDB 10 and try to install the package again, I get error message "Failed to install Wallabag. Incorrect MySQL root password." I would have expected that in the second case he would try to create user and database in the MariaDB 5.

jamallonon:
I am having the same plain HTML issue described here: https://github.com/wallabag/wallabag/issues/4061 .  What do you guys have configured for SYMFONY__ENV__DOMAIN_NAME?

Issue #3742, opened 2019-07-24
[Package Request] Updates for SynoCli Network Tools


jonathanweinberg:
Just a heads up about a couple packages within SynoCli Network Tools that are out of date.

Further, as of 2019/07/23, the GNU Screen version being distributed also has a fairly severe CVE that's been since been fixed.

Details below!

If someone wants to do a bit of hand holding with me, I'd be happy to help out in implementing / testing changes needed.

### GNU Screen
GNU Screen | ---
------------ | -------------
[Version in use](https://github.com/SynoCommunity/spksrc/blob/master/cross/screen/Makefile) | 4.0.3
[Current version](http://ftp.gnu.org/gnu/screen/) | 4.6.2
Specific Escalation CVE | [CVE-2017-5618](https://www.cvedetails.com/vulnerability-list/vendor_id-72/product_id-1860/year-2017/opgpriv-1/GNU-Screen.html)
General CVE's for Product | [All CVE's](https://www.cvedetails.com/product/1860/GNU-Screen.html?vendor_id=72)

### sshfs 
sshfs | ---
------------ | -------------
[Version in use](https://github.com/SynoCommunity/spksrc/blob/master/cross/sshfs/Makefile) | 2.10
[Current version](https://github.com/libfuse/sshfs/releases/tag/sshfs-3.5.2) | 3.5.2
Patch Note Urging 3.x | [From 2.10 release](https://github.com/libfuse/sshfs/releases/tag/sshfs-2.10)

### socat 
socat | ---
------------ | -------------
[Version in use](https://github.com/SynoCommunity/spksrc/blob/master/cross/socat/Makefile) | 1.7.3.2
[Current version](http://www.dest-unreach.org/socat/doc/CHANGES) | 1.7.3.3
Note | Bugfix update.

### tmux
tmux | ---
------------ | -------------
[Version in use](https://github.com/SynoCommunity/spksrc/blob/master/cross/tmux/Makefile) | 2.9a 
[Current version](https://github.com/tmux/tmux/releases/tag/2.9a) | 2.9a
Note | There is an RC 3.0rc3 available.

### mosh 
mosh | ---
------------ | -------------
[Version in use](https://github.com/SynoCommunity/spksrc/blob/master/cross/mosh/Makefile) | 1.3.2
[Current version](https://github.com/mobile-shell/mosh/releases/tag/mosh-1.3.2) | 1.3.2

### nmap
nmap | ---
------------ | -------------
[...]

PR #3820, opened 2019-12-14
WIP mosh dedicated package

ymartin59:
sshfs/fuse now builds with meson/ninja
fuse requires udev to build

Updates
- fuse 3.8.0
- glib 2.62.2
- libevent 2.1.11-stable
- libffi 3.3
- socat 1.7.3.3
- sshfs 3.6.0
- tmux 3.0a

_Motivation:_  tools updates
_Linked issues:_  #3742 

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

Issue #3582, opened 2019-01-16
Mosh Detached session message


sn00pster:
mosh-server has the ability to display a message telling you of any detached sessions that are connected to the server.

This support is based on utmp and libutempter.  The current build of the most server does not link against libutempter, therefore the functionality is not available.

I have done some very hacky exploratory work, basically fudging into the build a Makefile for utempter which requires a lot of interaction while building to get it to use and link against utempter.

However, the message is still not displayed, so there is something else at fault here.  I will try to discover what it is, but I know nothing about the correct way of getting utempter to automatically build and use a configure script, so I cannot share my work on this as it requires me to manually edit files during building and after failures in order to get the library built.

Any pointers on how I can get libutempter to download and configure/build properly would be helpful and would certainly advance figuring out what is going on.

m4tt075:
I'll be happy to help, if you are willing to do all the testing. Just PR your WIP and we'll take it from there...  ;-)

sn00pster:
Let me try and figure something out, I can’t do a PR on the hacky bit as I’m fudging stuff inside the work-x64 folder as it builds....  I can do a PR on the libutempter “module” under cross and the additional dependency that mosh now has on it.

libutempter does not use a configure script, it’s just a Makefile, I used autoconf to generate a configure script and then once it started building I then replaced the generated Makefile with a modified version of the original makefile from the source package so that the libraries get written to the right place so that when mosh builds its configure script picks up that libutempter is available.....told you it was a hack! Lol

I’m an embedded software engineer by trade and my only PC coding these days is limited to Qt, so I never get involved with makefiles which is why i’ve Struggled to this point.

m4tt075:
To get you started with our build system:
- You will need to create a cross-package for `utempter` in `/cross/utempter`. Then, you will have to add that as a dependency in the `cross/mosh/Makefile` , make it compile and put the libraries in the right place.
- Start creating the `cross/utempter` package first
- You can copy the `Makefile` from `/cross/mosh` for instance to get you started
- If there is no configure script, you can override configuration by adding `CONFIGURE_TARGET = nop" in the `Makefile` and the step will be skipped. I don't think `autoconf` will be needed.
- Pathes can be a bit tricky. Is there an option to define an `INSTALL_PREFIX`?

P.S.: I'm working in marketing these days. If I can do it, you can...  :-)

sn00pster:
Thanks! I spent ages looking for how to disable configure!!

I’ve done cross/utempter
[...]

PR #4195, opened 2020-09-28
Update synocli net

### LINKED ISSUE #3519: Arp-scan for synology?
There is a good utility in Linux called arp-scan. It shows computers in networks (ip, hostname, MAC). Is it possible to build spk with arp-scan? It might be very helpful.

hgy59:
_Motivation:_  Revise outdated synocli-net packages and add arp-scan
_Linked issues:_  #3742, #4196, closes #3519

### Checklist

### Remarks
- add diyspk for some packages of synocli-net to verify updates before building the collection
- ~unfortunately with newer toolchains added, the rule `all-supported` does not use the TC 6.1 anymore but takes the newest TC available. Therefore I created the rule `all-general` to build TC 6.1 and ppc853x-5.2 version.~
- most of the contained packages depend on cross/openssl and my intention is to merge this to master (and not to publish) before the update of openssl to version 1.1.x.
- dependency of `Perl` is removed and `mosh` is moved to a dedicated package
- add `arp-scan` (includes three perl scripts that are not mandatory and therefore `SPK_DEPENDS=Perl` is omitted
- add `links` and remove dedicated package (moved to diyspk)


hgy59:
rutorrent still fails to build (depending on updated screen) but I didn't find a fix so far.
So I propose to merge this PR to go further with openssl update.

th0ma7:
Hi @hgy59 , a few questions for you:
> unfortunately with newer toolchains added, the rule all-supported does not use the TC 6.1 anymore but takes the newest TC available. Therefore I created the rule all-general to build TC 6.1 and ppc853x-5.2 version.

- Doesn't a `make setup` forces to use DSM6.1 thus building package as before?  Or should the `make all-supported` rule set be reviewed as well?
- Indeed the `openssl.cnf` is not installed anymore with the usage of `/etc/ssl` by default.  What was the reason to revert the change that you first did?  And should it apply to newer 1.1.1 version?
- For my own understanding what does the `diyspk` stands for?

hgy59:
> * Doesn't a `make setup` forces to use DSM6.1 thus building package as before?  Or should the `make all-supported` rule set be reviewed as well?

`make setup` sets the default TC to DSM 6.1, but this is only for make arch-{arch} without specification of the TC version.
`make all-supported` does not regard the DSM version set with make setup - it takes the highest version available for each arch. I don't know why the 6.2 and 6.2.2 toolchains where added. We would need the exact version only for packages that require the kernel sources (afaik). Otherwise we (should) publish with the lowest version compatible with the actual version.

May be it is worth to remove my `all-general` target and adjust the `all-supported` target to work like this. ❓ 

_Another approach would be to regard the default TC version set with `make setup` - but with auto resolving the version when the exact version is not available. i.e. take the highest TC version of the arch if it is less than the default version - or - take the lowest TC version if versions greater than the default are available._

Other considerations:
- `make all-legacy` builds still for TC 4.3
- `make all-supported` should include the build for ppc853x-5.2 as this is a still supported version by synology.com.
- I suppose that with DSM 7 we will have to build for 6.1 and 7.0 in parallel, as some models will not get the update to DSM 7 (as DS210+ did not for DSM 6)

*****

> * Indeed the `openssl.cnf` is not installed anymore with the usage of `/etc/ssl` by default.  What was the reason to revert the change that you first did?  And should it apply to newer 1.1.1 version?

I removed the openssl.cnf because it is not used anymore and gives a warning due to missing the file. And I reverted it as this is not related to synocli-net and forces all the openssl dependent packages to be built by the github build action. (it sould be removed from the PLIST in a PR related to openssl)

*****
> 
> * For my own understanding what does the `diyspk` stands for?

This stands for do it your self packages, i.e. packages that can be built by developpers themself and that are not published (any more) to the SynoCommunity package repo. It was introduced with the creation of synocli-* package collections. Since the synocli collections contain a lot of packages it is very useful to have the single package diyspk versions to build and test updated packages that are part of synocli-*.

th0ma7:
@hgy59 thnx for the answers, additional comments on it:
1. Agreed, now that I better understand the behaviour of `all-supported` I believe this really this should map to what we **do** support which should range from ppc853x-5.2 up to DSM-6.1 (optionnally new 6.2 only arches such as purley, geminilake, etc). This will need to be readjust later anyway with the arrival of DSM7.  With a similar mindset, `all-legacy` should be revisited as well but its not urgent for the moment (perhaps with DSM7). And yes 6.2 & 6.2.2 where added due to kernel related requirements.  Sadly 6.2.3 was never published by Synology which contains all the intel cpu security fixes that where backported.  This leads to module & kernel crashes at loading time when loading a 6.2.2 compiled module under a 6.2.3 running kernel... 
2. Good, I will ensure its there in the `openssl` PR (lack of time this week to look into it but I believe I already removed it)
3. Really interesting I didn't know... there may be some interest for sub `ffmpeg` packages like intel drivers for hardware acceleration for instance.  I'll keep that to mind.

th0ma7 (commented):
Why removing sshfs? ... just re-read and figure you are not removing sshfs but rather moving it as diyspk instead of standalone spk but also keeing in into `synocli-net`, I'm all good witht hat.

But shouldn't the `make all-general` be superseeded with #4200 by now?

cross/arp-scan/patches/001-update-get-iab.patch
```diff
@@ -0,0 +1,23 @@
+# update url for iab.txt
+# fix wide character output
```
ymartin59, line 2: Please do not "hesitate" to submit such patch upstream to ease future maintenance/update

ymartin59:
@hgy59 @th0ma7 I would like to avoid to create "some many" diyspk/ packages because it increases "maintenance cost" when there is "no need" or no obvious tips they will be used by anyone sooner-or-later.
First "diyspk" packages were introduced to keep previous "standalone package" for small applications still available (not to enforce users to change if able to build on their own), when introducing aggregation in "synocli" packages...

Notice that in previous discussions, I agree to extract "mosh" out of "synocli-net" as a standalone package spk/mosh because it is quite heavy and introduces dependency to Python. May you please move diyspk/mosh to spk/mosh and cut dependencies to make synocli-net far lighter.

mk/spksrc.common.mk
```diff
@@ -33,6 +33,10 @@ LEGACY_ARCHS = $(sort $(filter-out $(SUPPORTED_ARCHS), $(AVAILABLE_ARCHS)))
 # SRM - Synology Router Manager
 SRM_ARCHS = northstarplus ipq806x dakota
 
+# arch-tc to build 
+# make all-general is now provided to build the general packages - same as built by github-build-action
+ALL_GENERAL_ARCH_TC = 88f6281-6.1 armv7-6.1 armv7-1.2 x64-6.1 aarch64-6.1 evansport-6.1 ppc853x-5.2 qoriq-6.1
+
```
ymartin59, line 39: IMO Conflicting with #4200. Please discard

mk/spksrc.spk.mk
```diff
@@ -333,6 +333,36 @@ publish-all-archs: $(addprefix publish-arch-,$(AVAILABLE_ARCHS))
...
+	  done \
+	else \
+	  for arch in $(sort $(basename $(subst -,.,$(basename $(subst .,,$(ARCHS_NO_KRNLSUPP)))))) ; \
+	  do \
+	    $(MAKE) publish-arch-$$arch ; \
+	  done \
+	fi
+
+####
+
```
ymartin59, line 365: IMO Conflicting with #4200. Please discard

ymartin59 (changes requested):
I would also move spk/sshfs to diyspk/sshfs for "compatibility" only

ymartin59:
According to this diyspk "migration" policy, it would make sense to remove them "one day". But again I would not create diyspk package for so small new tools.

hgy59:
> According to this diyspk "migration" policy, it would make sense to remove them "one day". But again I would not create diyspk package for so small new tools.

I find it rather usefull to have such diyspk for the synocli-* collections and would like to keep them, as you can build and test single package updates without the need to builld and install the whole collection.

ymartin59:
@hgy59 OK. For sure if they are used for testing, there is less chance to forget to update them. We will see... until then go on that way.
About mosh own package, would it be possible for you to take over what I started at #3820 (of course forget about Meson support which will be addressed later)
There is also a trouble to investigate with mosh (and probably tmux) concerned detached session: #3582 #3583
Again many thanks for your work here.

cross/arp-scan/patches/001-update-get-iab.patch
hgy59: @ymartin59 just found that this works with the current release

hgy59:
I propose to rebase and merge only after #4155 is merged into master.

hgy59:
@ymartin59 build fails due to error in openssl build (1.0.1u).
So I definitly wait for the merge of #4155 before finalizing this.

The error comes form installing openssl parts into /var/packages instead of using the install folder ../install/var/packages/synocli-net/target/...
I propose to forget openssl 1.0.1u and focus on 1.1.1.

ymartin59:
@hgy59 The fix quite simple. I propose you grab it from my PR #4195 to confirm https://github.com/SynoCommunity/spksrc/blob/cbda607efcccafc1781cf4a80a14316012426e21/cross/openssl/Makefile#L59

hgy59:
> 
> 
> @hgy59 The fix quite simple. I propose you grab it from my PR #4195 to confirm
> 
> https://github.com/SynoCommunity/spksrc/blob/cbda607efcccafc1781cf4a80a14316012426e21/cross/openssl/Makefile#L59

Yes this is the fix, but commiting would lead to merge conflicts - therefore I proposed to forget about openssl 1.0.1u.

ymartin59:
@hgy59 No there will be no merge conflict with Git if change is stricly the exact same line, that is why it is worth trying.

PR #4213, opened 2020-10-10
Add 6.2.3 toolchains

### LINKED ISSUE #3949: Add 6.2.2 toolchains

hgy59:
_Motivation:_  Add the toolchains published on 2020-10-08  by synology.com
_Linked issues:_  closes #3949

### Remarks
- add all toolchains for DSM 6.2.3
- add new arch v1000 and include in generic x64 (this looks like a new arch that will be used for future Diskstations).
- remove inexistant 6.2 versions (denverton, geminilake, hi3535, purley)
- thanks to the great work of @th0ma7 with #4142 this was reduced to minimal effort for new toolchains.


th0ma7:
@hgy59 glad it made it easy to add 6.2.3, the hope was to pave the way for DSM7 eventually.  #4200 will need to be merged before this one.  Also note that I noticed a few generic arch (i.e. armv7 and others) that I had forgotten that I added in that PR.  I haven't look if you have added them or not but just a heads-up.

Also, it's sad that they haven't published updated kernel which is really the part I am interested the most.

Again, great work!

hgy59:
@th0ma7 I hope you have no troubles with #4200 for the remove of the following toolchains:
- denverton-6.2
- geminilake-6.2
- hi3535-6.2
- purley-6.2

These toolchains are not provided by synology. The deleted makefiles used version 6.2.2 instead.

th0ma7:
> @th0ma7 I hope you have no troubles with #4200 for the remove of the following toolchains:
>     * denverton-6.2
>     * geminilake-6.2
>     * hi3535-6.2
>     * purley-6.2
> 
> These toolchains are not provided by synology. The deleted makefiles used version 6.2.2 instead.

Actually I do as there is a "limitation" with our SynoCommunity web engine where the version has to be two digits (e.g. 6.1, 6.2) which makes it to not accepting 6.2.2 or 6.2.3 sub-versions.  This is why I "faked" a 6.2 for theses arches using their closest version available.

Can you re-add them up please?  And while at it create a V1000?  Once done I'll rebase PR #4200 and add V1000 where needed  in the "spksrc.common.mk" file.

Lastly, could you give it a shot at #4200 once rebased and confirm it builds all-right?

hgy59:
@th0ma7 please avoid fake versions. Either a version is available or we must not publish it.
Therefore for denverton, and hi3535 version 6.1 is available for arch specific packages and for geminilake and purley is none available as long as three part versions are not supported in the repo.

For most packages we use the generic x64 arch that is available for version 6.2. Only for the rare arch hi3535 a version 6.2 is missing.

----

**Please consider:**

I suppose there was a mistake in TC cleanup. 
IMHO the value of os_min_ver is always **X.Y-Z** (DSM major number, DSM minor number, DSM build number).
**So for the former `TC_OS_MIN_VER = 6.2.2-24922` in 6.2.2 toolchain makefiles was invalid.**
This means that for DSM 6.2.2 and DSM 6.2.3 there will still be an os_min_ver of **6.2-xxxxx**. I suppose this is what the synocommunity repo was designed for and we must fix this in the publish and build process and not in the data model of the repo.

I don't know why the os_min_ver was removed from the toolchain makefiles but without this information we cannot solve this.
And while fixing this: we should use os_min_ver only, and avoid to reintroduce TC_FIRMWARE variable i.e firmware=$(TC_FIRMWARE) in the INFO file.
Current INFO files should and future INFO files must define `os_min_ver=...` and avoid `firmware=...`.

hgy59:
@th0ma7 just found the definitions in spksrc.tc-vers.mk
This should be easy to fix.

and please use TC_OS_MIN_VER and avoid TC_FIRMWARE (everywhere).

th0ma7:
> denverton, and hi3535 version 6.1 

Both of them do have a 6.1 & 6.2.2 but no 6.2.  Indeed that 6.2 needs to be removed.

> geminilake and purley 

Well, indeed there is none but redirecting to `x64` which uses an old toolchain doesn't allow the full experience for vaapi hardware acceleration.  Further more with newer version where a kernel version validation will be needed before long.  Creation of a "fake" 6.2 allows this, something that dsm7 will supersede entirely hopefully.  I believe it's a low cost to maintain theses two but ok at following the lead if needed.  So a fix on the online API would fix that or eventually migrating to dsm7.  In the meantime this was the easy fix.

> @th0ma7 just found the definitions in spksrc.tc-vers.mk
> This should be easy to fix.
> and please use TC_OS_MIN_VER and avoid TC_FIRMWARE (everywhere).

This was a total mess... It varied from one toolchain to another without any clear direction.  I'm totally OK at forcing one for all as anyway it looks like it becomes mandatory with dsm7.  I'll add that to #4200 .

th0ma7:
So while reviewing the `spksrc.tc-vers.mk` code, it was always setting up `TC_FIRMWARE` and if `TC_OS_MIN_VER` wasn't set it would TC_OS_MIN_VER = $(TC_FIRMWARE).  Following your comment above it now does the reverse.  If you believe `TC_FIRMWARE` should not be set then it's really easy to simply remove the 3x offending lines on top.

Also, for now I've re-included `geminilake-6.2` and `purley-6.2`.  For vappi acceleration purpose I've added those two "fake" archs into the `all-supported` builds.  If this needed to be remove then `SUPPORTED_ARCHS` variable in `spksrc.common.mk` also needs to be adjusted.

hgy59:
> Also, it's sad that they haven't published updated kernel which is really the part I am interested the most.

@th0ma7 the 6.2.3 kernels are available on sf.net since about 12 hours.
see https://sourceforge.net/projects/dsgpl/files/Synology%20NAS%20GPL%20Source/25426branch/

th0ma7:
This is great news! Thnx!  Finally!
Will update the kernels & all accordingly...

Issue #4211, opened 2020-10-09
Fall 2020 security package update


th0ma7:
Issue for tracking all packages needing publishing following major updates to `openssl`, `glib`, `gnutls` & `libgcrypt` adressing multiple security issues namely:

* https://www.cvedetails.com/vulnerability-list/vendor_id-217/product_id-383/version_id-194420/Openssl-Openssl-1.0.2g.html
* https://www.cvedetails.com/vulnerability-list/vendor_id-283/product_id-16275/version_id-260024/Gnome-Glib-2.56.1.html
* https://www.cvedetails.com/vulnerability-list/vendor_id-72/product_id-4433/GNU-Gnutls.html
* https://www.cvedetails.com/vulnerability-list/vendor_id-4711/product_id-25777/Gnupg-Libgcrypt.html

Related PR: 
* ~#4155: Update openssl~ (merged)
* ~#4010: Update glib, gnutls & libgcrypt~ (merged)
* ~#4209: Update ruTorrent~ (merged, see Package updates below)
* ~#4195: Update synocli-net~ (merged)
* ~#4218 Update to tvheadend~ (merged)
* ~Update mosquitto (todo: create new PR after merge of #4115)~

Indirectly related PR:
* ~#4200: Update to `make all-supported` with parallel build to ease publishing~ (merged)
* ~#4208: Remove bitlbee~ (merged)

### Package updates
- [ ] boxbackup-client        0.12-2
- [x] ffmpeg                    4.3.1
- [x] fossil-scm              2.12.1-2 (~BUG: error while loading shared libraries: libssl.so.1.1: cannot open shared object file~ fixed with #4241)
- [x] git                     2.28.0-16
- [x] gnupg                     2.2.23-3 (#4010, #4233)
- [ ] haproxy                 2.2.4-22 (#4010)
- [x] links                   2.21-2 (moved to synocli-net with #4195)
- [x] mediainfo          20.08-11 (consolidate with #4209)
- [x] ntopng          4.2-3 (consolidate with #4255 + download `libmysqlclient` package fix #4245 + libzmq duplicate removal #4254)
- [x] sslh            1.21c-5  (#4010 + fix with  #4246 + libcap #4248)
- [x] synocli-file            1.2-4  (update midnight commander to 4.8.25)
- [x] rutorrent                    3.10  (BUG: service does not start - consolidate with #4209 and #4234)
[...]

PR #4221, opened 2020-10-15
plist.mk: Cheap fix for PLIST madness found with Issue #4211

th0ma7:
_Motivation:_  Fix `PLIST` black-magic issue found
_Linked issues:_  #4211

### Checklist

### DIFF
--- a/mk/spksrc.plist.mk
+++ b/mk/spksrc.plist.mk
@@ -9,14 +9,13 @@ endif
 
 .PHONY: cat_PLIST
 cat_PLIST:
-	@for depend in $(DEPENDS) ; \
+	@echo Processing PLIST dependencies 1>&2
+	@for depend in `$(MAKE) dependency-list` ; \
 	do                          \
-	  $(MAKE) WORK_DIR=$(WORK_DIR) --no-print-directory -C ../../$$depend cat_PLIST ; \
-	done
-	@if [ -f PLIST ] ; \
-	then \
-	  $(PLIST_TRANSFORM) PLIST ; \
-	else \
-	  $(MSG) "No PLIST for $(NAME)" >&2; \
-	fi
-
+	  if [ "$${depend%/*}" = "cross" ] && [ -s ../../$${depend}/PLIST ]; then \
+	    echo "$${depend}" 1>&2 ; \
+	    cat ../../$${depend}/PLIST 1>&2 ; \
+	    $(PLIST_TRANSFORM) ../../$${depend}/PLIST ; \
+	  fi ; \
+	done ; \
+	$(PLIST_TRANSFORM) $(WORK_DIR)/../PLIST


th0ma7:
@hgy59 and @ymartin59 and all, here is a temporary fix proposal for the `PLIST` issue as documented in #4211.
Preliminary test shows that is works against `fossil-scm`.
Requires more testing...

th0ma7:
Solution ain't sufficient, needs additional thinking or complete rewrite of the `PLIST` code.
This quick fix proposal doesn't handle per arch sub-dependencies which are then considered mandatory for all arches.
Complex package such as `ffmpeg` is a good one to test with.

mk/spksrc.plist.mk
```diff
@@ -9,14 +9,13 @@ endif
 
 .PHONY: cat_PLIST
 cat_PLIST:
-	@for depend in $(DEPENDS) ; \
+	@echo Processing PLIST dependencies 1>&2
+	@for depend in `$(MAKE) dependency-list` ; \
```
hgy59, line 13: If you take dependency-list instead of `$(DEPENDS)` you get the dependencies of  `$(BUILD_DEPENDS) $(DEPENDS) $(OPTIONAL_DEPENDS)`
but this will fail because all files in PLIST must exist (and the optional are dependencies that exist not for all archs).
And BUILD_DEPENDS is for packages that are used only in the build environment and not needed at runtime (i.e. dependency of another spk like python).
If dependency-list fixes the PLIST file for some packages, those packages may have used BUILD_DEPENDS but should have used DEPENDS.
As for fossil-scm if the error you meant in #4222 is missing PLIST files of openssl or zlib, than the real error is that fossil-scm has defined BUILD_DEPENDS instead of DEPENDS for cross/zlib and cross/openssl.

hgy59:
Just used grep and found false usage of BUILD_DEPENDS in
- cross/autossh
- cross/he853
- cross/tcl
- spk/borgbackup
- spk/duplicity
- spk/ffsync
- spk/mercurial

th0ma7:
> Just used grep and found false usage of BUILD_DEPENDS in

How do you do to you determine if theses are false usages of `BUILD_DEPENDS` ?

hgy59:
> 
> 
> > Just used grep and found false usage of BUILD_DEPENDS in
> 
> How do you do to you determine if theses are false usages of `BUILD_DEPENDS` ?

if a package has BUILD_DEPENDS on it's own it must fail, as it is not (properly) packaged into the spk.
- may be I am wrong here (cross/duplicity and cross/borgbackup are wheels) 

cross/tcl:
- this package links to zlib (I cannot see any options that zlib is statically linked), and must define DEPENDS = cross/zlib

cross/autossh:
- this package links to openssh and must define DEPENDS = cross/openssh

cross/he853:
- this package links to cross/libusb and must define DEPENDS = cross/libusb

spk/ffsync:
- here I am not sure, as cross/gevent cross/greenlet may be used for building wheels only. But I guess these two libraries are needed at runtime and not only at build time.

spk/mercurial:
- another wheel package, where I don't know whether the wheel includes all the dependencies or not.

So my statement was hasty - not regarding the dependencies of wheels -  but for non python wheels following packages must use DEPENDS instead of BUILD_DEPENDS:

- cross/autossh
- cross/he853
- cross/tcl

and 
- cross/fossil-scm

ymartin59:
I am curious to know what "code changes" has introduced bug to better understand how to get it fixed. Does it come from moving targets to dedicated "plist.mk" file?

th0ma7:
> I am curious to know what "code changes" has introduced bug to better understand how to get it fixed. Does it come from moving targets to dedicated "plist.mk" file?

There was no code change from the framework point of view.  Also the changes I recently did had no effect on the `PLIST` part of the framework.

If there was a change it may end-up being in a dependency being build only with shared libs or a change in configuration option that make a binary to not link against the expected static lib but rather needing the shared lib instead.

Thus, from @hgy59 analysis and comment above, some of the packages must use `DEPENDS` instead of `BUILD_DEPENDS`.  No one yet took action AFAIK on that.  Part of the proposed list figures also `fossil-scm`, package which me starting this whole thread originating from https://github.com/SynoCommunity/spksrc/issues/4211#issuecomment-708379978 and proposed approach in this "tentative" PR.

Lastly, note that code in this PR is irrelevant (thnx to @hgy59)... although discussions may lead at finding a better approach.

hgy59:
> I am curious to know what "code changes" has introduced bug to better understand how to get it fixed. Does it come from moving targets to dedicated "plist.mk" file?

The code change for fossil was with commit 26ad9005a9e50d3014b958d5af601497977c4674 on `cross/fossil-scm/Makefile`
https://github.com/SynoCommunity/spksrc/commit/26ad9005a9e50d3014b958d5af601497977c4674#diff-8a946aa031b07b52cf3ae47688cfe3536e3dba5ad14d011e6fa45c4c0c4659e4

by removing `--static` from `CONFIGURE_ARGS` without changing BUILD_DEPENDS to DEPENDS.

ymartin59:
As a local mistake in fossil-scm I propose to close without merge

th0ma7:
Agreed.
@hgy59 Before closing, is there any action left to do considering your previous comment https://github.com/SynoCommunity/spksrc/pull/4221#issuecomment-711067242 ?  Should another issue be opened to track it?

hgy59:
> 
> 
> Agreed.
> @hgy59 Before closing, is there any action left to do considering your previous comment [#4221 (comment)](https://github.com/SynoCommunity/spksrc/pull/4221#issuecomment-711067242) ? Should another issue be opened to track it?

@th0ma7 PR #4257 created

PR #4236, opened 2020-10-26
framework: fix make publish-all-* in spk.mk

th0ma7:
_Motivation:_  `$(PUBLISH)` variable is not being honored in last recipe call thus needs a different approach to properly handle publishing flag.
_Linked issues:_  N/A

### Checklist
- [x] Build rule `publish-all-supported` completed successfully
- [x] Build rule `all-latest` completed successfully
- [x] Build rule `publish-all-latest` completed successfully

### DIFF
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -341,7 +341,7 @@ ALL_ACTION = $(sort $(basename $(subst -,.,$(basename $(subst .,,$(DEFAULT_ARCHS
 endif
 
 ifeq (publish,$(subst -all-latest,,$(subst -all-supported,,$(firstword $(MAKECMDGOALS)))))
-PUBLISH = publish
+PUBLISH = publish-
 .NOTPARALLEL:
 endif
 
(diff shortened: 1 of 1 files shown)


th0ma7:
Noticed the following issue while publishing `mediainfo`.  Could probably be addressed slightly differently if someone comes up with a better idea for handling `$(PUBLISH)` flag.

hgy59:
@ymartin59 as this PR concerns the publish process I have the following proposal

The current command used by the publish rule is:
`http --verify=no --auth $(PUBLISH_API_KEY): POST $(PUBLISH_URL)/packages @$(SPK_FILE_NAME)`

There is a colon (:) after the api key and I have to put my user name just after the colon to get successful authorization.
As this is a personal (local) setting I propose to introduce the variable `PUBLISH_USER` that may be set in `local.mk` and change the command to:
`http --verify=no --auth $(PUBLISH_API_KEY):$(PUBLISH_USER) POST $(PUBLISH_URL)/packages @$(SPK_FILE_NAME)`

If the auth user is mandatory, a validation like for the api key should be added too.

hgy59:
@th0ma7 I still wonder, why the target `all-supported` builds for `apollolake-6.2.3 geminilake-6.2.3 purley-6.2.3`.
The generic `arch-x64-6.1` can be used for apollolake, geminilake, purley and other x86_64 cpu.
If there are dedicated packages that need the latest (or exact?) DSM version, we should introduce a variable similar to KERNEL_REQUIRED (may be we call it LATEST_REQUIRED) that must be defined for such packages.

Importent note:
There is usually no problem to install packages built for arch-x64-6.1 on Devices with geminilake or purley that never supported DSM 6.1.

th0ma7:
> @ymartin59 as this PR concerns the publish process I have the following proposal
> 
> The current command used by the publish rule is:
> `http --verify=no --auth $(PUBLISH_API_KEY): POST $(PUBLISH_URL)/packages @$(SPK_FILE_NAME)`
> 
> There is a colon (:) after the api key and I have to put my user name just after the colon to get successful authorization.

I never had that behaviour... On the other hand I never used the `make publish-all-supported` neither (but intend of this PR is to switch to that).  I'm using the following `local.mk` :
```
PUBLISH_URL=https://api.synocommunity.com
PUBLISH_API_KEY=<MYKEY>
MAINTAINER?=
MAINTAINER_URL=
DISTRIBUTOR=SynoCommunity
DISTRIBUTOR_URL=https://synocommunity.com/
REPORT_URL=https://github.com/SynoCommunity/spksrc/issues
DEFAULT_TC=6.1
```
> As this is a personal (local) setting I propose to introduce the variable `PUBLISH_USER` that may be set in `local.mk` and change the command to:
> `http --verify=no --auth $(PUBLISH_API_KEY):$(PUBLISH_USER) POST $(PUBLISH_URL)/packages @$(SPK_FILE_NAME)`
> 
> If the auth user is mandatory, a validation like for the api key should be added too.

Aaahhh I think I understand... I use my API key, thus I do not need my username.  Although I guess that functionality could easily be added if there really is a reason for it.

th0ma7:
> @th0ma7 I still wonder, why the target `all-supported` builds for `apollolake-6.2.3 geminilake-6.2.3 purley-6.2.3`.
> The generic `arch-x64-6.1` can be used for apollolake, geminilake, purley and other x86_64 cpu.
> If there are dedicated packages that need the latest (or exact?) DSM version, we should introduce a variable similar to KERNEL_REQUIRED (may be we call it LATEST_REQUIRED) that must be defined for such packages.

There are multiple views on this.  The issue I found is that `ffmpeg` produces a slightly different output depending of the toolchain used.  In comparison the default x64 toolchain does not produce the exact same output as there are additional options being activated starting with `apollolake` toolchain for `ffmpeg`.

Also the hardware acceleration for Intel drivers (older and newer) has a direct relationship with the underlying kernel.  This starting with 4.4 there is better support (actually older kernels only support the older driver).  That is something I'd like to better support in the future for DSM7 so appropriate drivers gets build depending of the kernel version.  Initial groundwork for this was layed down with the addition of `TC_KERNEL` variable.

Lastly, what I would like is to add is something similar to `ARCH_REQUIRED` so any additional archs gets build with the all-supported and/or all-latest.  That in conjunction with `TC_KERNEL` would lead in custom builds for specific arches as needed.  Considering the upcoming AMD V1000 arch it may well be further required.

> Importent note:
> There is usually no problem to install packages built for arch-x64-6.1 on Devices with geminilake or purley that never supported DSM 6.1.

Yeah, that is the other side of things... As the two arches never got supported in the first place shouldn't we provide arch-specific builds?  But again, is it really needed always?  Again, `ARCH_REQUIRED` would allow producing them only where really needed and getting rid of them from the default `all-supported`.

th0ma7:
Lastly, second view on the actual code change would be appreciated.
While functional I don't believe it is optimal (and may not need to neither).

hgy59:
> Aaahhh I think I understand... I use my API key, thus I do not need my username. Although I guess that functionality could easily be added if there really is a reason for it.

@th0ma7 forget my comments. It works now without username. I can't reproduce the issue I had...

ymartin59 (approved):

PR #4234, opened 2020-10-25
spk/rutorrent: fix startup issues

smaarn:
rtorrent upgrade followup: fixed invalid configuration clause for certificate path
Added: missing pgrep commandline tool (used by _tasks plugin)
Added: missing sox commandline tool
Added: php workers configuration (for webstation integration)
Fixed: automatic detection of binaries was broken because of rights on php/test.sh script
Added: if python 3 installation is present also install cloudscraper wheel (required for _cloudflare plugin)
Fixed: former configuration file was always lost upon upgrade
Fixed: .htaccess file defined at root level wasn't backed up
Fixed: permissions were lost on the tmp directory (which needed to be accessible to both web and rutorrent user)

_Motivation:_  Having rutorrent back operational
_Linked issues:_  Should tackle https://github.com/SynoCommunity/spksrc/issues/4227

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

### Pending issues
- [x] Fix auto-download of new torrent: `Download event action failed: Bad return code.`
- [x] Deleting a torrent error:  `Event 'event.download.erased' failed: Bad return code.`
- [x] Fails to download torrent from `https` URL: `Failed to add torrent. Can't retrieve URL.`
- [x] Screen session msg every 5s `Scheduled command failed: watch_directory: Command "load_start" does not exist.`
- [x] Icon missing in DSM menu

### Optional
- [ ] ~Add built-in support for `autodl-rutorrent`~ - issue #2221

PR #4239, opened 2020-10-28
service.mk: Fix integer comparison with null string

th0ma7:
_Motivation:_  Found a bug related to shell integer comparison in default startup scripts.
_Linked issues:_  N/A

### Checklist
- [ ] N/A

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -80,7 +80,7 @@ ifneq ($(strip $(SPK_USER)),)
 	@echo USER=\"$(SPK_USER)\" >> $@
 	@echo "PRIV_PREFIX=sc-" >> $@
 	@echo "SYNOUSER_PREFIX=svc-" >> $@
-	@echo 'if [ -n "$${SYNOPKG_DSM_VERSION_MAJOR}" -a "$${SYNOPKG_DSM_VERSION_MAJOR}" -lt 6 ]; then EFF_USER="$${SYNOUSER_PREFIX}$${USER}"; else EFF_USER="$${PRIV_PREFIX}$${USER}"; fi' >> $@
+	@echo 'if [ -n "$${SYNOPKG_DSM_VERSION_MAJOR}" ] && [ "$${SYNOPKG_DSM_VERSION_MAJOR}" -lt 6 ]; then EFF_USER="$${SYNOUSER_PREFIX}$${USER}"; else EFF_USER="$${PRIV_PREFIX}$${USER}"; fi' >> $@
 endif
 ifneq ($(strip $(SERVICE_WIZARD_GROUP)),)
 	@echo "# Group name from UI if provided" >> $@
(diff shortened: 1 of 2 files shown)


th0ma7:
Hit the following bug while playing with faisibility to convert `rutorrent` to generic installer (indirectly relates PR #4234).

While doing the following:
```
# ./start-stop-status start
./service-setup: line 6: [: : integer expression expected
Starting  ...
./start-stop-status: line 29: [: : integer expression expected
```
It appears that the code checks if value is not null **and** lower than 6 (for prior to DSM6 version) such as:
```
[ -n "${SYNOPKG_DSM_VERSION_MAJOR}" -a "$SYNOPKG_DSM_VERSION_MAJOR" -lt 6 ]
```
The issue is that the entire condition `[ ]` gets evaluated as a whole thus the second condition will generate an error if length of string is 0.  The following attempts to fix this as the first condition will not pass thus second will not be checked:
```
[ -n "${SYNOPKG_DSM_VERSION_MAJOR}" ] && [ "$SYNOPKG_DSM_VERSION_MAJOR" -lt 6 ]
```

ymartin59:
This is not a bug but a "feature". Please review doc about command line usage: https://github.com/SynoCommunity/spksrc/wiki/Frequently-Asked-Questions#how-to-query-package-status-or-start-it-from-command-line
"&&" is only Bash but not Sh. "-a" is used so that script runs on both DSM 5 (sh probably dash) and DSM 6 (bash).

th0ma7:
Then it really is broken as both tests cannot be evaluated within a single `[ ]` : if the first one fails, the second will still get evaluated and thus generate an error.

Another view on things is to process them as strings using case which would solve that.  I'll try to propose another approach then.

th0ma7:
New approach tries to alleviate that using `case` which also opens the door to future changes due to DSM7 if any.
Also in `spksrc.service.start-stop-status` a `-n` test was done against `${USER}` while it isn't being used it the code below.

Converting:
```
if [ -n "$${SYNOPKG_DSM_VERSION_MAJOR}" -a "$${SYNOPKG_DSM_VERSION_MAJOR}" -lt 6 ]; then ...; else ... ;
```
to:
```
case $${SYNOPKG_DSM_VERSION_MAJOR} in [12345] ) ...  ;; * ) ... ;; esac
```

ymartin59:
Again, script is designed to run from Package Center or `synopkg` context only  https://github.com/SynoCommunity/spksrc/wiki/Frequently-Asked-Questions#how-to-query-package-status-or-start-it-from-command-line
I find no  "relevant" way of "fixing" it as it may impact on package files (logs, permissions)... except if we officially discard DSM 5 support from framework

th0ma7:
@ymartin59 honestly I'm OK at simply closing the issue and fully agree on the context how the code should be called.

Although, independently from how the script is being invoked, the `if` statement is fundamentally broken.  The reason it works is that when being properly called the `${SYNOPKG_DSM_VERSION_MAJOR}` variable is never `null` thus it passes the test.

PR #4247, opened 2020-10-31
tt-rss: added troubleshooting logs

smaarn:
_Motivation:_  Some twisted logics may actually fail or log errors which would need to be logged for troubleshooting + migration of installation procedure to use standard service-setup
_Linked issues:_  Without solving it it could help identifying the issue faced in https://github.com/SynoCommunity/spksrc/issues/4231

### Checklist


smaarn:
Made this PR into a draft until I have validated all upgrade and installation procedures

spk/tt-rss/src/service-setup.sh
```diff
@@ -77,14 +71,16 @@ postinst ()
     fi
 
     # Fix permissions
-    chown ${USER} ${WEB_DIR}/${PACKAGE}/lock
-    chown ${USER} ${WEB_DIR}/${PACKAGE}/feed-icons
-    chown -R ${USER} ${WEB_DIR}/${PACKAGE}/cache
+    set_syno_permissions "${INSTALL_DIR}/var/logs" "${USER}"
```
smaarn, line 74: The logs directory is populated with files which may be written by `root` user (during startup) or by upgrade script (run by the daemon as `${USER}` user).

spk/tt-rss/src/service-setup.sh
```diff
@@ -77,14 +71,16 @@ postinst ()
...
+
+    set_unix_permissions "${WEB_DIR}/${PACKAGE}/lock"
```
smaarn, line 76: It seems php 5.6 does not *really* support ACLs when using the `is_writable` method and therefore causes Tiny Tiny RSS to complain about those files not being appropriately accessible.

spk/tt-rss/src/conf/resource
```diff
@@ -10,17 +10,5 @@
 			"relpath": "etc/php/conf.d/com.synocommunity.tt-rss.ini",
 			"destpath": "/usr/local/etc/php70/conf.d"
 		}]
-    },
-    "php72": {
```
smaarn, line 14: Those php workers actually don't exist. Therefore the ttrss configuration need to be manually configured at the level of the PHP profile :(

spk/tt-rss/src/dsm-control.sh
```diff
@@ -18,21 +18,21 @@ HT_ACCESS_SECTION_DELIMITER="#Synology PHP"
 
 compute_generated_htaccess_section()
 {
-    local php_fpm="$1"
+    local php_profile_name="$1"
```
smaarn, line 21: php-fpm version based handlers was actually a backward compatible hack enforced by the webstation default configuration. 
For 7.3+ versions you actually need to rely on the handlers defined by the current php profile.

spk/tt-rss/src/service-setup.sh
```diff
@@ -45,24 +44,19 @@ preinst ()
     exit 0
```
ymartin59, line 44: Please remove "exit" statements to keep framework installer expected processing

smaarn:
Superseeded by #4274

PR #4286, opened 2020-12-05
Hassio: Fix for newer supervisor

fredrike:
_Motivation:_  Explain here what the reason for the pull request is.

### Checklist

### DIFF
--- a/spk/hassio/Makefile
+++ b/spk/hassio/Makefile
@@ -10,7 +10,7 @@ REQUIRED_DSM = 6.2
 DEPENDS =
 
 MAINTAINER = fredrike+hassio@gmail.com
-DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder. This is an unsupported installation method and is not maintained by the core developers group. For support please see this forum thread: https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/."
+DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder. This is an unsupported installation method and is not maintained by the Home Assistant developers. For support please see this forum thread: https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/."
 RELOAD_UI = yes
 STARTABLE = yes
 DISPLAY_NAME = Hass.io (unofficial)
--- a/spk/hassio/Makefile
+++ b/spk/hassio/Makefile
@@ -10,7 +10,7 @@ REQUIRED_DSM = 6.2
 DEPENDS =
 
 MAINTAINER = fredrike+hassio@gmail.com
-DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. <br/>Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder. <br/>This is an unsupported installation method and is not maintained by the core developers group. For support please see this forum thread: <a href=\"https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/\">https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/</a>."
+DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder. This is an unsupported installation method and is not maintained by the core developers group. For support please see this forum thread: https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/."
 RELOAD_UI = yes
 STARTABLE = yes
 DISPLAY_NAME = Hass.io (unofficial)
--- a/spk/hassio/Makefile
+++ b/spk/hassio/Makefile
@@ -10,10 +10,10 @@ REQUIRED_DSM = 6.2
 DEPENDS =
 
 MAINTAINER = fredrike+hassio@gmail.com
-DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder. This is an unsupported installation method and is not maintained by the core developers group. For support please see this forum thread: https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/"
+DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. <br/>Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder. <br/>This is an unsupported installation method and is not maintained by the core developers group. For support please see this forum thread: <a href=\"https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/\">https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/</a>."
 RELOAD_UI = yes
 STARTABLE = yes
-DISPLAY_NAME = Hass.io
+DISPLAY_NAME = Hass.io (unofficial)
 CHANGELOG = ""
 
 UNSUPPORTED_ARCHS = $(ARM_ARCHES) $(PPC_ARCHES) $(x86_ARCHES)
--- a/spk/hassio/Makefile
+++ b/spk/hassio/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = hassio
-SPK_VERS = 20200910
-SPK_REV = 2
+SPK_VERS = 20201206
+SPK_REV = 3
 SPK_ICON = src/$(SPK_NAME).png
 SPK_DEPENDS = "Docker>=18.09.0-0506"
 DSM_UI_DIR = app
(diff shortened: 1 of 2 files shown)


balloob:
I suggest we don't add this PR.

This file you're adding disables all checks that help supervised installations stay up to date and make sure the supervisor won't break the host OS. 

Synology OS is not supported by Hass.io. Users should instead run HassOS in a VM if supported by their Synology or run Home Assistant Core. 

Even making it a tiny bit easier to do this hack means you're sending people in the wrong direction, filled with broken things that then end up reflecting bad at Home Assistant.

bruvv:
I suggest we do add this PR.

This fixes the weird implementation that is being added. I understand that home assistant wants to see what is causing issues. But there should be a less intrusive way of disabling just that check.

Yes I've watchtower running, but I've set it up correctly
Yes I've an older docker version, but it's very stable and working just fine

So far for me and many others, hassio has been working just perfect on synology docker. This is nothing but a hack in my eyes. This is a fix that is needed because home assistant made a hack to show false positives. 

So please merge this PR otherwise a lot of people will start to complain that they cannot run hassio anymore, which is the opposite of what we al want isn't it?

fredrike:
I've added a note in the description: https://github.com/SynoCommunity/spksrc/pull/4286/commits/a7bc4a032c281b0b6e56dd16275c3dd21e51e49a#diff-e5fe16e0d41defdd807c0ed7bf51b290a3641492f0bc5ed25e7c7dd3df2c9b05R13

balloob:
> a lot of people will start to complain that they cannot run hassio anymore, which is the opposite of what we al want isn't it?

In a perfect world where we have all the time in the world and unlimited resources, sure we would want people to run Home Asssistant Supervisor on every system in the world. However, we don't have all the time. If you know what you're doing, fine that's great. But don't steer people with less experience down a road that with challenges they are not capable of solving on their own. 

When (not if) they encounter a problem that they can't solve, they think Home Assistant is broken. Doesn't matter if we tell them they run an unsupported system. We get the complaints.

We have two alternatives for Synology users that work great and that we support.

We all want Home Assistant to evolve, get better, support more things. That also takes time. With our limited time we need to choose how we can have the biggest impact and serve the most people. Trying to make things work with outdated and unsupported Linux kernels does not fit.

bruvv:
> In a perfect world where we have all the time in the world and unlimited resources,

In a perfect world, we all would contribute and make something great, together :)

> Trying to make things work with outdated and unsupported Linux kernels does not fit.

100% agree. But a less destructive way, will be a much better approach.
Just a huge banner explaining what the issue is exactly (running watchtower container for example, running old docker version etc). This way people will know that support will be limited and people that know what they're doing should have the option to turn this big warning message off.

fredrike:
> We have two alternatives for Synology users that work great and that we support.

Well, if you refer to the VM alternative there is a very limited number of devices that actually support it.

Just running the Docker version of hass-core might be too limiting.

BeardedTinker:
> I understand that home assistant wants to see what is causing issues.

It is already known what's the issue - Docker version on Synology - it's 18.x and it needs to be at least 19.x (if I'm not mistaken)
And we will not see official update from Synology - not as we have DSM 7.0 approaching. And I'm not sure what version of Docker is on 7.0. or other potential troubles when we get there. 

Not going to write here as I've already written in community post my thoughts but I support decision of the development (and support) team of HA and I really do understand why they've added this "unsupported" flag, but also I support approval of fredrike PR after this addition disclaimer.

hgy59:
> It is already known what's the issue - Docker version on Synology - it's 18.x and it needs to be at least 19.x (if I'm not mistaken)
> And we will not see official update from Synology - not as we have DSM 7.0 approaching. And I'm not sure what version of Docker is on 7.0. or other potential troubles when we get there.

The DSM 7.0 Preview has:
`Docker version 18.09.8, build bfed4f5`

BeardedTinker:
> 
> The DSM 7.0 Preview has:
> `Docker version 18.09.8, build bfed4f5`
> 

That's, no good. To say lightly.

balloob:
> In a perfect world, we all would contribute and make something great, together :)

When Home Assistant decides to support something, we make sure it works before we ship it. If we decide to support everything, we end up either a) having to wait for every platform to be updated or b) ship things that break setups. Neither are acceptable. 

@fredrike can you please update the display name and add `(Unofficial)` to indicate that this is not an official Home Assistant release.

spk/hassio/Makefile
```diff
@@ -10,10 +10,10 @@ REQUIRED_DSM = 6.2
 DEPENDS =
 
 MAINTAINER = fredrike+hassio@gmail.com
-DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder."
+DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder. This is an unsupported installation method and is not maintained by the core developers group. For support please see this forum thread: https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/."
```
balloob, line 13: This is a little more clear:
```suggestion
DESCRIPTION = "Open source home automation that puts local control and privacy first. Powered by a worldwide community of tinkerers and DIY enthusiasts. Note: identified USB devices will be added to the file usb_devices.txt located in the hass.io folder. This is an unsupported installation method and is not maintained by the Home Assistant developers. For support please see this forum thread: https://community.home-assistant.io/t/hass-io-on-synology-dsm-native-package/."
```

rngtng (approved):
Nice!

BradleyFord:
The key here is that the decision to add this blocking should not have occured withour an architecture discussion, and it should have been labeled as a breaking change.
Neither of those things occured, but that is what everyone outside the core team is force to do.

ymartin59 (approved):

fredrike:
@ymartin59, could we please see a package build in the near future?

balloob:
@BradleyFord Running our supervisor on a Synology system is not supported. There have been 5 ADRs discussed in the architecture repository in the last year defining supported installation methods. 

Our supervisor does a lot, and so when we see that a system is not healthy, we need to treat with caution. You can choose between a running system with no updates versus a system that will break with an update. In that case, we rather keep a system up and running than getting it bricked. Either way we are made the bad guys because it's our software (even if we don't support how it's running).

Promoting an unsupported install to users with a single click package install is cruel as eventually it will require maintenance and the users are not prepared for that. But then… you guys don't care because we get the hate, the issues, the down votes, not you. Thanks 😞

BradleyFord:
Paulus,
Sorry I have not only down voted without sharing the comments of why; and I
really appreciated the awesome software that has been created (Also why I
am chipping into NabuCasa even though I am not fully using it).

I cannot call myself a developer anymore (unless you want me to break out
some COBOL!!); but I am trying to support the project where I can, such as
getting the product logos.

I just think we needed some improvement on the communications side:
- I subscribe to the architecture repository so I can see everything that
is submitted, I remember reading through these discussions and agreeing
with the direction as it is needed for a maturing project; and it was a
good and reasonable discussion/outcome.
- My concerns are more around that I don't remember seeing this specific
change in the breaking changes, and I would have expected that there would
be some method for override at your own risk.

I fully agree we should not promote this as supported because it definitely
should not be.

cjamieso:
I haven't previously added my comments here, but I do support the objections that Paulus has to making this change.  I apologize that I hadn't done this sooner and it's contributed to the perception that users of the package don't care about everyone else.

I don't think there's any way that a synology installation can be supported, since the user has no control over the underlying docker implementation.  My only hope is that it's permitted to continue to exist as an unsupported method.  I did explore the two other methods that were presented this past weekend, but each comes with a drawback of its own (happy to explain further) and after weighing each, I've elected to continue with the package method.  I anticipate returning to those tradeoffs in the future as things change and that's OK.

Paulus, I appreciate all of the work that you put into making home assistant successful.  I think that your philosophy and reasoning are a big part of that success, because there is a clear vision guiding the project.  I experienced a bug on the google cast side earlier this year (entirely unrelated to synology) and I was impressed by the speed at which it was fixed and the communication throughout.  I don't think this was accidental, but rather something that your leadership enables.

The only bit that I object to is the characterization that users of the package don't care.  I care about the hate, issues, and downvotes because home assistant has been such a joy to use and I want to see it continue.  I also think that love of home assistant is why this package was created.  That being said, I'm happy to acknowledge that I should do my part in expressing support, so that you'll see something other than hate or downvotes.  My hope is that there is some middle ground here where the package can continue to exist, but with language around what a user can expect.

chickenandpork:
> you guys don't care because we get the hate, the issues, the down votes, not you. Thanks

This "thanks" sounds like sarcasm, and I'm not sure this is the best form of communication.  If we didn't have issues to resolve, Fredrike wouldn't be spending cycles to make a package, and have interested installations.  It's easy to toss things back and forth: you can claim we don't care about the burden it causes, we could claim that we, as developers who might like yaml and gitops, are increasingly left behind, but these arguments merely emphasize the differences or goals and opinions.

Neither of us wants the other to have more effort, but we also see and share a demand for a thing.  In common, we both are passionate about home automation, and we likely all view the recent hardware release that suggests why HA architecture has gone in a certain direction.  We applaud the work even if it's not our chosen implementation.

Do you know how many installs there are of each release type?

What does Supervisor need that necessitates Docker-18.x ?

Is there a way that you can work with Fredrike as a member of your team to steer the DSM release in a way that works well with your longer-term plan, understands there reasons for Supervisor evolving, and can bring us closer to compatibility?

I think that would be better than the sarcasm and "we get the hate not you".  Let's avoid all of the hate, even the issues that Fredrike's volunteer work and attention to details has worked to alleviate.

fredrike:
> Is there a way that you can work with Fredrike as a member of your team to steer the DSM release in a way that works well with your longer-term plan, understands there reasons for Supervisor evolving, and can bring us closer to compatibility?

I would be happy to help to keep this package stable (which I think it already is). 

@chickenandpork, would it be possible to get qemu to run on all x86/x64 based Synology devices? I think I saw that it lacks kernel support but I might be mistaken. This would be an alternative approach to keep Home Assistant (!core) running and supported on Synology.

fredrike:
Due to various issues but mostly due to Home Assistant Supervised requires newer Docker (and kernel) this should not be merged.

PR #4288, opened 2020-12-06
spk.mk: Add a new spkclean option for package specific build files

th0ma7:
_Motivation:_  When debugging package compilation it is really useful to be able to only remove spk package creation files.
_Linked issues:_  None

### Checklist
- [x] N/A

### DIFF
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -319,6 +319,21 @@ endif
 clean:
 	rm -fr work work-*
 
+spkclean:
+	rm -fr work-*/.copy_done \
+	       work-*/.depend_done \
+	       work-*/.icon_done \
+	       work-*/.strip_done \
+	       work-*/.wheel_done \
+	       work-*/conf \
+	       work-*/scripts \
+	       work-*/staging \
+	       work-*/package.tgz \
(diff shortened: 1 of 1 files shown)


publicarray:
@th0ma7 Is this so that after I run `make spkclean` I don't need to recompile but the package files are regenerated on the next `make arch-*-*`?

th0ma7:
> @th0ma7 Is this so that after I run `make spkclean` I don't need to recompile but the package files are regenerated on the next `make arch-*-*`? If yes than that would reduce the time it takes to debug scrips. But aren't startup/install scripts regenerated every time anyway? So it would be a huge benefit other generated assets?

It is exactly for that... all the cases where you: play with scripts, have forgottent a library version update in a PLIST somewhere and also when you want to rebuild only a specific sub-dependency (`make spkclean && rm -fr work*/subdep* work*/.subdep*`).  Using that simple new function allows you to play with your existing build as will.

publicarray:
Thanks @th0ma7,  Good work! Should make debugging things easier.

publicarray (approved):

publicarray:
Ups had my phone in my pocket

Issue #4250, opened 2020-11-01
autossh fails siliently


andrewchenshx:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ SynoCli Network Tools
_Package Version:_ 1.5-5

_NAS Model:_ DS720+
_NAS Architecture:_ geminilake  x86_64
_DSM version:_ DSM 6.2.3-25426 Update 2

### Expected behavior
(1)command line output
(2)ssh tunnel established

### Actual behavior
the autossh command exited siliently without any message, and ssh tunnel was not established.  

### Steps to reproduce
_1._ install the SynoCli Network Tools
_2._ ssh login to nas (ssh 192.168.1.9 -p 222)
_3._ sudo -i
_4._ autossh (output the help as expected)
_5._ autossh -M 0 -o "ServerAliveInterval 30" -o "ServerAliveCountMax 3" -i /volume1/data/_rsa -R 0.0.0.0:6690:127.0.0.1:6690 -p 2222 nas_tunnel@remote_ip
this command finished without any error massage, but no tunnel was established
#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
```
Insert the package log here
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here
```

hgy59:
@andrewchenshx can you please test this issue with the updated synocli-net v1.5-6?

Xiut:
Unfortunately I have the same problem. After a long time of trial and error maybe I got a hint for the error with AUTOSSH_DEBUG=7:

On my DS218play
Realtek RTD1296 SoC
DSM 6.2.3-25426 Update 2
```
autossh[12214]: port set to 0, monitoring disabled
autossh[12214]: checking for grace period, tries = 0
autossh[12214]: starting ssh (count 1)
autossh[12214]: ssh child pid is 12215
autossh[12214]: check on child 12215
autossh[12214]: clear alarm timer (0 secs left)
autossh[12214]: set alarm for 600 secs
autossh[12215]: child of 12214 execing /spksrc/spk/synocli-net/work-aarch64-6.1/install/var/packages/synocli-net/target/bin
autossh[12215]: /spksrc/spk/synocli-net/work-aarch64-6.1/install/var/packages/synocli-net/target/bin: No such file or directory
autossh[12214]: received signal to exit (15)
autossh[12214]: clear alarm timer (600 secs left)
```


On my DS412+
[...]

PR #4289, opened 2020-12-06
Autossh fix ac_cv_path_ssh

th0ma7:
_Motivation:_  Output binary uses the `$(STAGING_INSTALL_PREFIX)` instead of $(INSTALL_PREFIX) leading to an non-existing path once installed when setting the `ac_cv_path_ssh` build parameter.
_Linked issues:_  #4250

### Checklist

### DIFF
--- a/cross/autossh/Makefile
+++ b/cross/autossh/Makefile
@@ -12,6 +12,6 @@ COMMENT  = Automatically restart SSH sessions and tunnels.
 
 GNU_CONFIGURE = 1
 CONFIGURE_ARGS = ac_cv_func_realloc_0_nonnull=yes ac_cv_func_malloc_0_nonnull=yes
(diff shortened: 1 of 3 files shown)


hgy59:
As of the error text (and analyzing the configure script) I suppose it wants to execute a directory, so you must declare
`ac_cv_path_ssh=$(STAGING_INSTALL_PREFIX)/bin/ssh`

th0ma7:
It's exactly what my new build does, and was waiting for the user to confirm it solves the issue before pushing the change... but now committed.

hgy59:
@th0ma7 I don't know whether autossh uses the slogin command of openssh. I have fixed this, as it was linking to staging path too.

th0ma7:
@hgy59 thnx, good catch!

Issue #4295, opened 2020-12-09
ruTorrent errors (3.10-9)


defaultsecurity:
### Actual behavior
ruTorrent launches and runs, but produces errors in its own log.

### Setup
_Package Name:_  ruTorrent
_Package Version:_ 3.10-9
_Package url:_ https://packages.synocommunity.com/rutorrent/9/rutorrent.v9.f15047[apollolake-avoton-braswell-broadwell-broadwellnk-bromolow-cedarview-denverton-dockerx64-geminilake-grantley-purley-kvmx64-v1000-x86-x86_64].spk
_NAS Model:_ DS918+
_NAS Architecture:_ apollolake
_DSM version:_ DSM 6.2.3-25426 Update 2

### Steps to reproduce
_1._ I installed the package without disabling open_baseddir during install.
_2._ ruTorrent launched and it produced the errors in its own log.
_3._ I reinstalled the package, this time disabling open_basedir via the installer.
_4._ ruTorrent didn't launch anymore.
_5._ I reinstalled the package a couple more times.
_6._ Now ruTorrent launches again, but still produces the same errors as before.
_7._ I have the php 5.3, 7.0 and 7.3 packages installed. ruTorrent doesn't launch without 5.3.
_8._ I have changed the setting _"customize open_basedir"_ in web station to `:/` for all php versions, but didn't help.

#### ruTorrent log
```
[09.12.2020 14:15:10] WebUI started.
[09.12.2020 14:15:10] _cloudflare: Plugin will not work. Webserver user can't access external program (python).
[09.12.2020 14:15:10] unpack: Plugin will not work. rTorrent user can't access external program (php).
[09.12.2020 14:15:10] loginmgr: Some functionality will be unavailable. rTorrent user can't access external program (php).
[09.12.2020 14:15:10] ratio: Some functionality will be unavailable. rTorrent user can't access external program (php).
[09.12.2020 14:15:10] autotools: Plugin will not work. rTorrent user can't access external program (php).
[09.12.2020 14:15:10] rss: Plugin will not work. rTorrent user can't access external program (php).
[09.12.2020 14:15:10] datadir: Plugin will not work. rTorrent user can't access external program (php).
[09.12.2020 14:15:10] scheduler: Plugin will not work. rTorrent user can't access external program (php).
[09.12.2020 14:15:10] _task: Plugin will not work. rTorrent user can't access external program (php).
[...]

PR #4299, opened 2020-12-10
Rutorrent fix

th0ma7:
_Motivation:_  Fix issues found in latest `rutorrent` release
_Linked issues:_  #4295

### Checklist


spk/rutorrent/src/service-setup.sh
```diff
@@ -49,13 +49,13 @@ fix_shared_folders_rights()
     synoacltool -add "${folder}" "user:${EFF_USER}:allow:rwxpdDaARWc--:fd" >> "${INST_LOG}" 2>&1
     synoacltool -add "${folder}" "group:${USER}:allow:rwxpdDaARWc--:fd" >> "${INST_LOG}" 2>&1
     synoacltool -add "${folder}" "user:${APACHE_USER}:allow:rwxp-D------:fd" >> "${INST_LOG}" 2>&1
-    synoacltool -add "${folder}" "group:${APACHE_USER}:allow:rwxp-D------:fd--" >> "${INST_LOG}" 2>&1
+    synoacltool -add "${folder}" "group:${APACHE_USER}:allow:rwxp-D------:fd" >> "${INST_LOG}" 2>&1
```
smaarn, line 52: For what it's worth I believe you can skip the dashes.

spk/rutorrent/src/service-setup.sh
```diff
@@ -181,12 +190,29 @@ service_restore ()
 {
     # Restore the configuration file
     mv -f "${TMP_DIR}/config.php" "${WEB_DIR}/${PACKAGE}/conf/" >>"${INST_LOG}" 2>&1
+    set_unix_permissions "${WEB_DIR}/${PACKAGE}/conf/config.php"
```
smaarn, line 193: Shouldn't this be done after the configuration file is updated?

spk/rutorrent/src/service-setup.sh
```diff
@@ -154,7 +164,6 @@ service_save ()
     if [ -f "${WEB_DIR}/${PACKAGE}/.htaccess" ]; then
         mv "${WEB_DIR}/${PACKAGE}/.htaccess" "${TMP_DIR}/" >>"${INST_LOG}" 2>&1
     fi
-    cp -pr ${WEB_DIR}/${PACKAGE}/share/ ${TMP_DIR}/ >>"${INST_LOG}" 2>&1
```
smaarn, line 157: Maybe a stupid question but I believe this would mean some manually installed resources would get lost. Besides, I believe the files which were "saved" here are copied back later on. No other risks foreseen?

spk/rutorrent/Makefile
```diff
@@ -44,6 +44,8 @@ ENV += BUSYBOX_CONFIG="$(BUSYBOX_CONFIG)"
 
 include ../../mk/spksrc.spk.mk
 
+SPK_DEPENDS = "python3>=3.7.7:php5.6"
```
smaarn, line 47: Shouldn't we, instead, be checking for the presence of python before starting / installing and disabling the plugin if missing ?

smaarn (commented):
Mostly open questions here and one concern regarding the change on the share folder backup mechanism.

spk/rutorrent/src/service-setup.sh
th0ma7: Indeed, that is really only cosmetic as all previous didn't had the trailing `--`.

spk/rutorrent/src/service-setup.sh
th0ma7: That would mean that.  I noticed that an issue where the share directory ended-up being recpvered to a wrong location over to `/var/package/rutorrent/target/var` for an unknown reason.  And really to me, playing with that also means being in a non-supported environment.  I believe as is all regular users are covered by default.

spk/rutorrent/src/service-setup.sh
th0ma7: Well this is overwriting the default provided configuration file.  I believe (but may be wrong) the `service_postint` is not being invoked when "updating" a package, thus the configuration file is default until being overwritten by the recovery from the previous version.

If you where specifically referring to the `set_unix_permissions`, modifications done afterwards by appending the `define_external_program` for missing parameters when porting over an old configuration but that doesn't affect its permissions.  Similar to:
```
$ sudo touch toto
$ sudo chown sc-rutorrent:rutorrent toto
$ echo "AAA" | sudo tee -a toto
$ ll toto
-rw-r--r-- 1 sc-rutorrent rutorrent 4 Dec 10 10:46 toto
```

spk/rutorrent/Makefile
th0ma7: I thought it was much more simple to add `SPK_DEPENDS` which forces the installation of prerequisites packages in the first place, thus ensuring the plugin will work?  This way, in theory, there is no need to check anything and we can assume the plugin to work.

smaarn (approved):
As far as I'm concerned the only bit which worried me was the support for share location.
If people don't really care and if we don't want to leave the pption opened then go for it!

th0ma7:
> If people don't really care and if we don't want to leave the pption opened then go for it!

I'm not sure of the actual value but I may be missing the point?  Recovering that wouldn't it overwrite the new `rutorrent` web site with the older version, thus contradicting the benefit?

spk/rutorrent/src/service-setup.sh
smaarn: Actually the service_postinst is supposed to be invoked but some of its steps are only executed when not upgrading. Indeed one of the impacts is that the default configuration file gets overwritten by the one "in place". 
I was actually referring to the authorizations logics but indeed if ordering has no impact then my point was "moot" ;)

smaarn:
> > If people don't really care and if we don't want to leave the pption opened then go for it!
> 
> I'm not sure of the actual value but I may be missing the point? Recovering that wouldn't it overwrite the new `rutorrent` web site with the older version, thus contradicting the benefit?

Actually it wasn't overwritten if I'm not mistaken (from what I recall the `cp -pr` doesn't overwrite), thus the default values being brought by the core installation weren't overwritten. The impact of the changes here is that existing custom plugin installations would get wiped out. Maybe it would be worth documenting it.

spk/rutorrent/src/service-setup.sh
```diff
@@ -118,8 +118,7 @@ service_postinst ()
         fi
     fi
 
-
-    #If python3 is available setup a virtual environment with cloudscraper
+    # If python3 is available setup a virtual environment with cloudscraper
     if [ -f "${PYTHON_DIR}/bin/python3" ]; then
```
smaarn, line 122: My guess is, we could remove this condition and execute the then part whatever happens, since we're relying on python3 to be present.

th0ma7: good point, done.

th0ma7:
> Actually it wasn't overwritten if I'm not mistaken (from what I recall the `cp -pr` doesn't overwrite), thus the default values being brought by the core installation weren't overwritten. The impact of the changes here is that existing custom plugin installations would get wiped out. Maybe it would be worth documenting it.

As this is ran as `root` it does overwrite and preserve old permissions.  Thus migration from DSM5 to DSM6 will pose problem with that and it would overwrite the newly updated site.  What would be needed to do is fine a way to only copy over custom plugins... but practically speaking that wont be sustainable or require a lot of efforts for good chances of failures.

Although, if replacing `-pr` with `-nr` that would not be replacing the destination files and might end-up doing what you are looking for?

spk/rutorrent/src/service-setup.sh
th0ma7: You are totally right about `service_postint`... and re-reading the code makes it now obvious, thnx for the explanations :)

smaarn:
> > Actually it wasn't overwritten if I'm not mistaken (from what I recall the `cp -pr` doesn't overwrite), thus the default values being brought by the core installation weren't overwritten. The impact of the changes here is that existing custom plugin installations would get wiped out. Maybe it would be worth documenting it.
> 
> 
> 
> As this is ran as `root` it does overwrite and preserve old permissions.  Thus migration from DSM5 to DSM6 will pose problem with that and it would overwrite the newly updated site.  What would be needed to do is fine a way to only copy over custom plugins... but practically speaking that wont be sustainable or require a lot of efforts for good chances of failures.
> 
> 
> 
> Although, if replacing `-pr` with `-nr` that would not be replacing the destination files and might end-up doing what you are looking for?
> 
> 
> 
> ```
> 
> -n, --no-clobber
> 
>       do not overwrite an existing file (overrides a previous -i option)
> 
> ```

How about -npr then ? We still need rightd to be preserved as much as possible, don't we?

th0ma7:
> How about -npr then ? We still need rightd to be preserved as much as possible, don't we?

Currently looking into it... staging some code for this currently.

th0ma7:
@smaarn I think I solved it... I had an issue with file permissions being re-added when restoring backup files... now solved.
Would you mind having a shot at it and see how things flies: upgrade, new install, downloads... etc?

smaarn (approved):
Tested and validated:

- [x] update 9 -> 10 + downloads
- [x] install from scratch + downloads
- [x] changing configuration file directly and restarting

Regarding the usability of the configuration file feature, I'm only not convinced by the location. My point being that, as a random user I don't have any idea on where to look for the configuration file through the DSM UI interface. Of course, I can launch an ssh session and do that, but otherwise I must confess I have no idea on how to find the corresponding file through the DSM interface. Any hint ?

spk/rutorrent/src/bin/rtorrent-daemon
```diff
@@ -0,0 +1,22 @@
...
+
+export HOME=${INSTALL_DIR}/var
+export PATH="${INSTALL_DIR}/env/bin:${INSTALL_DIR}/bin:${INSTALL_DIR}/usr/bin:${PATH}"
+export LD_LIBRARY_PATH="${INSTALL_DIR}/lib:${LD_LIBRARY_PATH}"
+RTORRENT="${INSTALL_DIR}/bin/rtorrent"
+
+# If default config ($HOME/.rtorrent.rc) is a symlink
+# and package default location file exists, use parameters
+# to force using that config file
+if [ -L ${HOME}/.rtorrent.rc -a -f ${RTORRENT_RC} ]; then
```
smaarn, line 18: Maybe a stupid question, but why having this `.rtorrent.rc` symlink file hanging around anyway ? If you can use the configuration file directly ? It would simplify a lot the logic here, wouldn't it ?

smaarn: I mean we could just get rid of that file, right ?

th0ma7:
@smaarn Thnx a lot for the testing, much appreciated!

>  but otherwise I must confess I have no idea on how to find the corresponding file through the DSM interface. Any hint ?

Using FileStation under the web volume such as:
rutorrent-filestation

spk/rutorrent/src/bin/rtorrent-daemon
th0ma7: Indeed we could... and ain't sure what to do about it.

By default `rtorrent` looks into the `$HOME` directory.  I thought one might prefer having another copy instead of using the package default... I noticed that it does provide a sort of failsafe when things turns sideways.  Worth keeping?  honestly no clue.  Worth the effort to modify, again?

th0ma7: Honestly, I tend to like the current behavior... and I am too lazy to do the change (as I think we spent to many hours on this package).  I'd rather just move on and go ahead with the merge & distributing, and perhaps, revisit some other time.  But feel free to adjust as you see fit :)

smaarn:
> @smaarn Thnx a lot for the testing, much appreciated!
> 
> > but otherwise I must confess I have no idea on how to find the corresponding file through the DSM interface. Any hint ?
> 
> Using FileStation under the web volume such as:
> rutorrent-filestation

I must confess the folder is not available under my FileStation view. I'll check this then ;)

Well I guess it's all good on my end. And one last thing: regarding the symlink I believe it's actually a good idea to keep it since former users would still be able to find it.

th0ma7:
> I must confess the folder is not available under my FileStation view. I'll check this then ;)

There was another user from the bug request who was also missing that folder... actually, the symlink from `/var/sevices/*` towards /volume1/*` was missing.  Pehaps something to check?

> Well I guess it's all good on my end. And one last thing: regarding the symlink I believe it's actually a good idea to keep it since former users would still be able to find it.

Good then, lets move forward and merge & distribute.  Again, thnx for your time on this :)

PR #4153, opened 2020-08-30
deluge 2 upgrade to python 3.10 fix for DSM 7

Strangelovian:
_Motivation:_  Have the deluge 2.0 upgrade functional with python 3.7
_Linked issues:_  #4134, #4158, closes #4701, hopefully closes #5177

Update as deluge 2.0.5 was officially released in Dec 2021
- [ ] Update Deluge to v 2.0.5
- [ ] Update to Python 3.10
- [ ] Update for DSM 7
- [ ] No need for prerelease by @Strangelovian (tanks for investigation anyway).
- [ ] Finally libtorrent was updated to v1.12.10 (further updates, or updates to libtorrent v2 are out of scope here, and probably not compatible with the synology toolchains).

Notes:
- Cross compilation works for at least: x64, aarch64, denverton
- x64 and denverton packages are confirmed working on DS1819+: install OK, download torrent OK, uninstall OK
- the core.conf was changed to have initially a single port configured, else deluge 2 will not managed to bind any port. Deluge 2 manages only one listen port. The core.conf still accepts a list of listen ports in the core.conf for compatibility reasons with deluge 1.x.
- I updated the package icon to new deluge 2 icon
- boost upgrade is needed from v1.63 to v1.67, else Python3.Boost fails to compile
- cross compilation fails with boost v1.74 for aarch64. Fortunately v1.67 works with all architectures

**CAVEAT PLEASE READ:**
deluge daemon v2.0.3 won't start because of this issue: https://dev.deluge-torrent.org/ticket/3278
The fix is already done here https://github.com/deluge-torrent/deluge/commit/1b4ac88ce72515bcf36684b942e3998900f13cea
But no release was made since 2.0.3 so I forked it to have the last update: https://github.com/Strangelovian/deluge/tree/2.0.4-str0
Now, I will ask the deluge team to build an official 2.0.4  to avoid using my fork. https://forum.deluge-torrent.org/viewtopic.php?f=8&t=55767
Normally this should be quick to do, even though the deluge developments have really slowed down recently.

### Checklist
- [X] Build rule `all-supported` completed successfully
- [X] Package upgrade completed successfully
- [X] New installation of package completed successfully

upgrade to boost 1.67

upgrade to libtorrent 1.2.9

build from deluge bleeding edge 2.0.4

update deluge icon

adjust core.config with a single port, as required by deluge version 2.x

PR #4096, opened 2020-08-02
update octoprint to v1.7.2 and Python38

hgy59:
_Motivation:_  Octoprint spk is really outdated and does not work
_Linked issues:_  closes #3128, closes #3966, may close #2676

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade is not supported due to versioning scheme change, all older versions deleted from repository, further upgrade will be supported
- [x] New installation of package completed successfully

### Remarks
package upgrade is not provided as the versioning scheme changed (you cannot upgrade from 20160321-4 to 1.4.0-5 as this is a downgrade)
I propose to retire the old packages (remove from the repository) as they did not really work and were beta anyway.

PR #3849, opened 2020-01-10
ffsync: update to 1.8.0

ymartin59:
_Motivation:_  Update, DMS 6 support, fix uninstall mysql password issue
_Linked issues:_  #3124

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

PR #3823, opened 2019-12-16
Update duplicity with oauth and Python 3

heralight:
_Motivation:_  Update duplicity to 0.8.18 to support onedrive with their new api.

~Remarks: others backends doesn't compile anymore, synocommunity docker image is to outdated.~

### Checklist
- [X] Build rule `all-supported` completed successfully
- [X] Package upgrade completed successfully
- [x] New installation of package completed successfully

### Remarks
The current version of duply does not work with cross/bash. As DSM >= 6.0 has /bin/bash available and this version works with duply the cross/bash dependency is removed with this PR. For DSM < 6.0 only duplycity is supported, the duply script is not.

PR #4303, opened 2020-12-12
Deactivate python2

hgy59:
_Motivation:_  Python 2 is EOL since 01.01.2020 and should be deactivated
_Linked issues:_  #3633 

Deactivate all automated builds for Python2 and related packages.

### Tasks
- [x] Develop environment: install pip < 21 as pip 21.0 will drop python2 support
- [ ] introduce cross/pip2 for python2
- [ ] Deactivate CI (github build action) for python2 and related packages

### Related Packages
Packages to deactivate or migrate to python3:
  - [ ] python2
  - [ ] deluge (#4153)
  - [x] duplicity (#3823)
  - [ ] ffsync (#3849 but still python2)
  - [x] mercurial (#4591)
  - [x] octoprint (#4096)
  - [x] tvheadend (#4394)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,7 +64,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 RUN wget https://bootstrap.pypa.io/get-pip.py -O - | python3
 
 # Install setuptools, pip, virtualenv, wheel and httpie for Python2
-RUN wget https://bootstrap.pypa.io/get-pip.py -O - | python
+# Remarks: pip >= 21 has removed support for python2
+RUN wget https://bootstrap.pypa.io/get-pip.py && python get-pip.py "pip<21" && rm -f get-pip.py
 RUN pip install virtualenv httpie
 
 # Volume pointing to spksrc sources


ymartin59:
Synology delivers a Python 2.7.18 package with DSM 7... so we may port packages for applications without Python 3 support to use Synology version

hgy59:
> 
> 
> Synology delivers a Python 2.7.18 package with DSM 7... so we may port packages for applications without Python 3 support to use Synology version

Sure we may implement such ports for DSM7 but we will still support DSM6 for models not supporting DSM7.
And January 2021 comes fast and will break all python2 packages due to pip >= 21.

th0ma7:
I'll look into `tvheadend`.  There where recent changes upstream related to python3 (https://github.com/tvheadend/tvheadend/pull/1372) which might be sufficient to cover it all...

hgy59:
> Synology delivers a Python 2.7.18 package with DSM 7... so we may port packages for applications without Python 3 support to use Synology version

Unfortunately the installable Python2 (2.7.18) that comes from Synology for DSM 7 does not contain neigther `pip` nor `virtualenv`, so it is not an option to migrate to the synology version of Python2.
But this will only be required for `ffsync` package (#3849), as all other packages can be migrated to Python 3 (see #4998).

hgy59:
Closed as not planned.

PR #4207, opened 2020-10-06
Jackett "Standalone" aka build using .NET

publicarray:
_Motivation:_  https://github.com/SynoCommunity/spksrc/issues/3697#issuecomment-698116461
_Linked issues:_ #3697

CC @m4tt075

**Caveats:**
* Only `x64`, `ARM 8` and `ARM 7` are supported, ~~so I've moved it to a new package: `jackett-dotnet`~~
~~* Currently the .NET framework keeps data in `~/.dotnet`, so I used `docker run --name spksrc --rm -it -v $HOME/git/spksrc:/spksrc:delegated -v :/root/.dotnet synocommunity/spksrc:latest /bin/bash` to run docker~~

Build: https://github.com/publicarray/spksrc/releases/tag/jackett-dotnet-0.16.1583

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4306, opened 2020-12-13
dotnet-runtime 3.1.10

publicarray:
_Motivation:_  In the future having a single runtime allows us to update it independently of the dotnet apps. Reduces overall system file size.
_Linked issues:_  #4207 

The installation method is not my preferred method to do this, it just runs a script that downloads the runtime from Microsoft and extracts the binary during install time. yuck.

An alternative is to compile it using these instructions https://github.com/dotnet/runtime/blob/master/docs/workflow/README.md Just note the runtime repo is over 300mb while the script mentioned above only downloads a ~30mb binary.

### Checklist
- [x] Install works on DSM7 when compiled on the dsm7 branch
[dotnet-runtime_x64-7.0_3.1.10-1.zip](https://github.com/SynoCommunity/spksrc/files/5684012/dotnet-runtime_x64-7.0_3.1.10-1.zip)

Proof:

```sh
seby@NAS:~$ /volume1/@appstore/dotnet-runtime/dotnet --list-runtimes
Microsoft.NETCore.App 3.1.10 [/volume1/@appstore/dotnet-runtime/shared/Microsoft.NETCore.App]
```

Packages that depend on this runtime must set `DOTNET_ROOT=/volume1/@appstore/dotnet-runtime`

Can someone confirm this also works on dsm6 ?


DigitalBox98:
I got this package build as well on my repo and I have noticed the "sponge" utility was required in the docker image.
So I have to "apt-get install moreutils" before having the SPK build sucessfully

hgy59:
> 
> 
> I got this package build as well on my repo and I have noticed the "sponge" utility was required in the docker image.
> So I have to "apt-get install moreutils" before having the SPK build sucessfully

moreutils was added for sponge to the development evironment and to the Dockerfile.
current docker image is located in the github project registry: `ghcr.io/synocommunity/spksrc:latest`

publicarray:
@DigitalBox98 if you are interested in .net you might like the SDK https://github.com/publicarray/spksrc/releases/tag/dotnet-fix

hgy59:
@publicarray, I think you can close this, it's outdated.
Meanwile we do have dotnet runtime packages for version 6, 8 and 9.

And IMHO it does not make sense to create dotnet-sdk packages.

Additionally:
- For DSM < 7, the packages must include an updated libstdc++ and dotnet must be patched to use it, so the download at installation time wouldn't work
- It is not possible to install runtime and sdk in parallel (only one dotnet is found in the path) and we can't have a common root path for different versions or runtime and sdk into the same `DOTNET_ROOT` path (or we would need such a path outside /var/packages/...)

publicarray:
Thanks @hgy59

PR #4309, opened 2020-12-14
Update Radarr to v3.0.1.4259

houndtt:
_Motivation:_  To incorporate latest Radarr release code. Includes a fix to backup v0.2 DB as part of upgrade if it exists.
_Linked issues:_  N/A

### Checklist

### DIFF
--- a/cross/radarr/Makefile
+++ b/cross/radarr/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = Radarr
-PKG_VERS = 3.0.0.4204
+PKG_VERS = 3.0.1.4259
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME).master.$(PKG_VERS).linux.$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/Radarr/Radarr/releases/download/v$(PKG_VERS)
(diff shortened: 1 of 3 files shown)


Safihre:
This shouldn't be needed, Radarr internal update process should do the updates. Why is a seperate update of the package required?

houndtt:
> This shouldn't be needed, Radarr internal update process should do the updates. Why is a seperate update of the package required?

Yes, you are correct that this should not be needed. This was mostly to assist users who may be upgrading from the older v0.2 database as it adds some backup features in this build. Additionally, it can aid newbie users who may be presented with a health warning regarding upgrading to the .NET Core version of Radarr which they seemed to have wisely removed from this version.

bakerboy448:
> This shouldn't be needed, Radarr internal update process should do the updates. Why is a seperate update of the package required?

It's needed given how Syno is now running apps as dockers; thus updating is blocked as radarr blocks updating containers

bakerboy448:
just tested a fresh install 8 year old NAS (DS412+ w/ DSM 6.2.3-25426 Update 3) with SynoCommunity package

Though mine did install mono 5.2 as a prereq

updated to 3.0.2.4369 no issue; no docker error

I still suggest releasing the base package as at least 3.0.1 as it backs up v0.2 DBs on install

PR #4307, opened 2020-12-13
Add meson + ninja support

th0ma7:
_Motivation:_  Some newer package version mandatory require meson + ninja native build environment.
_Linked issues:_  N/A

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

### GLIB Specific
- Known build issue starting with 2.63.4
- Last known version to work on `hi3535` (`armv7l`): 2.63.3 (unstable)
- Issue created at:  https://gitlab.gnome.org/GNOME/glib/-/issues/2274

### Migrated `cross/package`
- [x] `libdri`
- [x] `libsigc++`
- [x] `glib`
- [x] `glibmm`
- [x] `pixman`
- [x] `zstd`

PR #4316, opened 2020-12-17
meson: Add meson to pip3 install

th0ma7:
_Motivation:_  Add meson to default image in preparation for mk meson+ninja
_Linked issues:_  #4307

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,6 +62,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 # Install setuptools, wheel and pip for Python3
 RUN wget https://bootstrap.pypa.io/get-pip.py -O - | python3
+RUN pip3 install meson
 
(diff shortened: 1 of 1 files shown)


th0ma7:
Tested on my `lxc` environment against Debian 10.  In theory should also work on Debian 9...

hgy59 (approved):

Dockerfile
```diff
@@ -62,6 +62,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 # Install setuptools, wheel and pip for Python3
 RUN wget https://bootstrap.pypa.io/get-pip.py -O - | python3
+RUN pip3 install meson
```
hgy59, line 65: may be it is better to pin the version (as we do in requirement files) to prevent unpredictable behaviour.
```suggestion
RUN pip3 install meson==0.56.0
```

PR #4143, opened 2020-08-22
DSM toolkit mk helper

th0ma7:
_Motivation:_  In order to get `evansport` architecture hardware acceleration we require to link binaries to the DSM toolkit.  I created a set of mk helpers to allow managing the download and extraction of platform toolkits independently of it's flavor.
_Linked issues:_  https://github.com/SynoCommunity/spksrc/pull/3873#issuecomment-596095210

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Build rule `arch-evansport-5.2` completed successfully
- [x] Build rule `arch-evansport-6.1` completed successfully
- [x] Build rule `arch-evansport-6.2` completed successfully
- [x] Build rule `arch-evansport-6.2.2` completed successfully

PR #4320, opened 2020-12-19
Update framework archs

hgy59:
_Motivation:_  Improve the ARCH-TC handling and make it more readable
_Linked issues:_  #4200, #4143

### Checklist
      - `make all-supported` restored to behave as before #4200
- [x] ~Remove the x86 toolchains as never used (was just a copy of x86_64 / pineview)~
- [x]  Prepare for DSM7 (when syno-x64-7.0 is available `make all-supported` with `DEFAULT_TC=6.1` should build x64-6.1 and not x64-7.0)
- [x] rename spksrc/toolchains to spksrc/toolchain (open issue of #4143)
- [x] add TC_GCC and TC_GLIBC definitions to toolchain Makefiles (for future _UNSUPPORTED_ package validation).
- [x] move ppc853x to deprecated archs and remove ppc853x-5.2 from github build action

### Remarks
- outsource the ARCH definitions to `mk/spksrc.archs.mk` so Makefiles need to include this file for arch specific rules only (before `spksrc.common.mk` had to be used)
- *_ARCHES definitions are renamed to *_ARCHS (overall consistent naming)
- rename x86_ARCHS to i686_ARCHS for x86 32-bit archs (evansport only)
- introduce ARMv7L_ARCHS for armv7 without hard float (hi3535 only)
- OLD_PPC_ARCHS is defined as often used for UNSUPPORTED_ARCHS
- remove the GENERIC_ARCHS handling as we have armv7 generic arch (it was used for cross-go before generic armv7 was fully established)
- Include some cleanup of #4200
  - exclude generic archs from all-legacy targets
- use LEGACY_ARCHS now for builds with kernel source required
- add patch subfolder handling for arch-group with TCVERSION

### Wiki variables for review:
- ARCHS_NO_KRNLSUPP to ARCHS_WITH_KERNEL_SUPPORT

### Results

- make all-supported
├── work-88f6281-6.1
├── work-aarch64-6.1
├── work-armv7-6.1
├── work-evansport-6.1
├── work-hi3535-6.1
├── work-qoriq-6.1
└── work-x64-6.1

- make all-legacy
├── work-88f6281-5.2
├── work-alpine-5.2
├── work-armada370-5.2
├── work-armada375-5.2
├── work-armada38x-5.2
├── work-armadaxp-5.2
├── work-avoton-5.2
├── work-braswell-5.2
├── work-bromolow-5.2
├── work-cedarview-5.2
├── work-comcerto2k-5.2
├── work-dakota-1.2
├── work-evansport-5.2
├── work-ipq806x-1.2
├── work-monaco-5.2
├── work-northstarplus-1.2
├── work-ppc853x-5.2
├── work-qoriq-5.2
└── work-x86-5.2

- make all-latest
├── work-88f6281-6.2.3
├── work-aarch64-6.2.3
├── work-armv7-6.2.3
├── work-evansport-6.2.3
├── work-hi3535-6.2.3
├── work-qoriq-6.2.3
└── work-x64-6.2.3


diyspk/dar/Makefile
```diff
@@ -11,7 +11,7 @@ DISPLAY_NAME = DAR
 STARTABLE = no
 
 # archs without C++11 support are not supported (libthreadar):
-UNSUPPORTED_ARCHS = $(ARM5_ARCHES) hi3535 powerpc ppc824x ppc853x ppc854x
+UNSUPPORTED_ARCHS = $(ALL_ARM5_ARCHS) $(ALL_ARM7L_ARCHS) $(OLD_PPC_ARCHS)
```
th0ma7, line 14: First I notice this, `ALL_` variables?

mk/spksrc.archs.mk
```diff
@@ -0,0 +1,36 @@
...
+x64_ARCHS = $(GENERIC_x64_ARCH) apollolake avoton braswell broadwell broadwellnk bromolow cedarview denverton dockerx64 geminilake grantley purley kvmx64 v1000 x86_64
+
+# Arch groups
+ALL_ARCHS = $(x64_ARCHS) $(i686_ARCHS) $(PPC_ARCHS) $(ARM_ARCHS)
+ARCHS_WITH_GENERIC_SUPPORT = $(filter-out $(GENERIC_ARCHS), $(ARM7_ARCHS) $(ARM8_ARCHS) $(x64_ARCHS))
+# PPC_ARCHS except qoriq
+OLD_PPC_ARCHS = powerpc ppc824x ppc853x ppc854x
+
+# outdated unsupported archs
+OBSOLETE_ARCHS = powerpc ppc824x ppc854x x86
```
th0ma7, line 36: I suggest renaming `OBSOLETE_ARCHS` to `DEPRICATED_ARCHS`

mk/spksrc.archs.mk
```diff
@@ -0,0 +1,36 @@
+# Common arch definitions
+
+# All available CPU architectures
+
+# Distinct SRM and DSM archs to allow handling of different TCVERSION ranges.
+# SRM - Synology Router Manager
+SRM_ARM7_ARCHS = northstarplus ipq806x dakota
+# DSM - all ARMv7 except SRM specific archs
+DSM_ARM7_ARCHS = alpine armada370 armada375 armada38x armadaxp comcerto2k monaco
```
th0ma7, line 9: What is the use-case of theses two variables?
Could instead this be `SRM_ARCHS` and `DSM_ARCHS` more broadly?

th0ma7 (commented):
This is looking good.
Going through the code why not drop entirely the `S` ?
With that in mind, `toolchains` is still something that would need renaming to `toolchain` ... 
Overall a lot of comments for consideration.  Not that they all need to be actionned (some may be for discussion or for a later PR)

mk/spksrc.archs.mk
```diff
@@ -0,0 +1,36 @@
...
+GENERIC_ARM64_ARCH = aarch64
+GENERIC_x64_ARCH = x64
+GENERIC_ARCHS = $(GENERIC_ARMv7_ARCH) $(GENERIC_ARM64_ARCH) $(GENERIC_x64_ARCH)
+
+ARM5_ARCHS = 88f6281
+ARM7_ARCHS = $(GENERIC_ARMv7_ARCH) $(DSM_ARM7_ARCHS) $(SRM_ARM7_ARCHS)
+# hi3535 is not supported by generic ARMv7 arch
+ARM7L_ARCHS = hi3535
+ARM8_ARCHS = $(GENERIC_ARM64_ARCH) rtd1296 armada37xx
+ARM_ARCHS = $(ARM5_ARCHS) $(ARM7_ARCHS) $(ARM7L_ARCHS) $(ARM8_ARCHS)
```
th0ma7, line 22: I'm having a little trouble here with the ARM naming overall.
ARM7 family and ARMv7 architecture does not have the same meaning at all but it becomes really hard to follow.
Does this needs to be addressed now or within another PR?  Duno.
But at least it should be much easier as now separated into its own file which is a great idea.
See: https://en.wikipedia.org/wiki/List_of_ARM_microarchitectures

mk/spksrc.common.mk
```diff
@@ -23,52 +23,41 @@ PIP_WHEEL = $(PIP) $(PIP_WHEEL_ARGS)
...
-# Available Arches
-ARM5_ARCHES = 88f6281
-ARM7_ARCHES = armv7 alpine armada370 armada375 armada38x armadaxp comcerto2k monaco hi3535 ipq806x northstarplus dakota
-ARM8_ARCHES = aarch64 rtd1296 armada37xx
-ARM_ARCHES = $(ARM5_ARCHES) $(ARM7_ARCHES) $(ARM8_ARCHES)
-PPC_ARCHES = powerpc ppc824x ppc853x ppc854x qoriq
-x86_ARCHES = evansport
-x64_ARCHES = x64 apollolake avoton braswell broadwell broadwellnk bromolow cedarview denverton dockerx64 geminilake grantley purley kvmx64 v1000 x86 x86_64
+# Global arch definitions
+include ../../mk/spksrc.archs.mk
```
th0ma7, line 31: if your new `archs.mk` is included within `common.mk` there may not be any changes needed in other Makefiles?  no?
Unless you wanted to reduce the scope within cross/package definitions?

mk/spksrc.common.mk
```diff
@@ -23,52 +23,41 @@ PIP_WHEEL = $(PIP) $(PIP_WHEEL_ARGS)
 # Available languages
 LANGUAGES = chs cht csy dan enu fre ger hun ita jpn krn nld nor plk ptb ptg rus spn sve trk
 
-# Toolchains
-AVAILABLE_TCS = $(notdir $(wildcard ../../toolchains/syno-*))
-AVAILABLE_ARCHS = $(notdir $(subst syno-,/,$(AVAILABLE_TCS)))
-
-# Toolchain filters
-SUPPORTED_ARCHS = $(sort $(filter-out $(ARCHS_DUPES), $(AVAILABLE_ARCHS)) apollolake-6.2.3 geminilake-6.2.3 purley-6.2.3)
```
th0ma7, line 31: My intent here with the trailing `apollolake-6.2.3 geminilake-6.2.3 purley-6.2.3` was to change this into `FORCED_ARCH` or something similar so when I compile lets say `ffmpeg` I could force some specific archs at `all-supported` time to create various x86_64 packages so `libdrm` would enable either Intel OR AMD and depending of   `TC_KERNEL` version I can then either only compile the old intel driver or both on newer 4.4+ platforms.

So below at line 54 when defining `SUPPORTED_ARCHS` I would include room for a added variable such as `FORCED_ARCHS` to be appended to the resulting value.

mk/spksrc.cross-cc.mk
```diff
@@ -73,7 +73,7 @@ include ../../mk/spksrc.generate-digests.mk
 include ../../mk/spksrc.dependency-tree.mk
 
 .PHONY: all-archs
-all-archs: $(addprefix arch-,$(AVAILABLE_ARCHS))
+all-archs: $(addprefix arch-,$(AVAILABLE_TOOLCHAINS))
```
th0ma7, line 76: can we use singular such as `AVAILABLE_TOOLCHAIN` ?
and while at it migrate `spksrc/toolchains` to `spksrc/toolchain`

mk/spksrc.patch.mk
```diff
@@ -17,14 +17,16 @@ endif
...
+#    patches/$(arch)/*.patch
+#    patches/$(arch)-$(TCVERSION)/*.patch
+# supported groups: arm5, arm7, arm7l, arm8, ppc, i686, x64
 ifeq ($(strip $(PATCHES)),)
-PATCHES = $(foreach arch,ARM5_ARCHES ARM7_ARCHES ARM8_ARCHES PPC_ARCHES x86_ARCHES x64_ARCHES, \
-	$(foreach subarch,$($(arch)), \
-	$(if $(filter $(ARCH),$(subarch)),$(sort $(wildcard patches/*.patch patches/DSM-$(TCVERSION)/*.patch patches/$(shell echo ${arch} | cut -f1 -d'_'| tr '[:upper:]' '[:lower:]')/*.patch patches/$(subarch)/*.patch patches/$(subarch)-$(TCVERSION)/*.patch)),)))
+PATCHES = $(foreach group,ARM5_ARCHS ARM7_ARCHS ARM7L_ARCHS ARM8_ARCHS PPC_ARCHS i686_ARCHS x64_ARCHS, \
+	$(foreach arch,$($(group)), \
+	$(if $(filter $(ARCH),$(arch)),$(sort $(wildcard patches/*.patch patches/DSM-$(TCVERSION)/*.patch patches/$(shell echo ${group} | cut -f1 -d'_'| tr '[:upper:]' '[:lower:]')/*.patch  patches/$(shell echo ${group} | cut -f1 -d'_'| tr '[:upper:]' '[:lower:]')-$(TCVERSION)/*.patch patches/$(arch)/*.patch patches/$(arch)-$(TCVERSION)/*.patch)),)))
```
th0ma7, line 29: I remember having a hard time creating this 3x line code for patches...
besides the variable renaming (subarch,arch to arch,group), what is the effect of the change downstream?
If I get this right, this will allow managing patches at even a lower level such as `qoriq-6.1/*.patch`
Based on my previous comment related to ARM, perhaps then this could be renamed using `family` instead of `group` which would be more aligned.

mk/spksrc.spk.mk
```diff
@@ -347,7 +347,7 @@ endif
 
 KERNEL_REQUIRED = $(MAKE) kernel-required
 ifeq ($(strip $(KERNEL_REQUIRED)),)
-ALL_ACTION = $(sort $(basename $(subst -,.,$(basename $(subst .,,$(ARCHS_NO_KRNLSUPP))))))
+ALL_ACTION = $(sort $(basename $(subst -,.,$(basename $(subst .,,$(ARCHS_WITH_KERNEL_SUPPORT))))))
```
th0ma7, line 350: I suggest once the patchset is in a readiness state that we go over the documentation for all the new variables that needs to be adjusted and/or removed/added.

mk/spksrc.spk.mk
```diff
@@ -378,34 +378,34 @@ publish-supported-arch-%:
...
 	done \
 
-publish-all-toolchain-%:
+publish-legacy-toolchain-%:
 	@$(MSG) Built packages for toolchain $*
-	@for arch in $(sort $(basename $(subst -,.,$(basename $(subst .,,$(filter %$*, $(AVAILABLE_ARCHS))))))) ; \
+	@for arch in $(sort $(basename $(subst -,.,$(basename $(subst .,,$(filter %$*, $(LEGACY_ARCHS))))))) ; \
 	do \
 	  $(MAKE) publish-arch-$$arch-$* ; \
 	done \
```
th0ma7, line 411: Is it possible to provide the output of what this creates in terms of directory tree in the PR description?
BTW, I found a really easy way to do that by forcing `wget` through a `.wgetrc` file to use a proxy and simply turn off your proxy.  It ends-up failing at download time **BUT** creating the directory structure.

mk/spksrc.archs.mk
hgy59: SRM_ARCHS are archs of Synology Routers and DSM_ARCHS are for Diskstations.
With the generic armv7 arch we have the problem that it is not possible to distinct the different TC-Version range (1.0-1.2 for SRM, 4.3 - 7.x for DSM)
Here I separated to be prepared for TCVERSION handing, but it has no priority for me in this PR.

mk/spksrc.archs.mk
hgy59: `armv7` (all lowercase) is the generic arch for ARM7 architectures in the same way as `x64` is for x86_64 and `aarch64` for ARM8.
These generic archs are pseudonyms to distinguish from real archs.
Under the hood there are no dedicated toolchains for these. But the toolchains of the generic archs contain all the dedicated archs that are compatible (variable TC_ARCH) that will define the "arch" variable in the INFO file of the generict spk packages.
The mapping used for generic toolchains is
- x64 -> pineview
- armv7 -> armada38x
- aarch64 -> rtd1296

I hope this helps. I really wanted to clarify such definitions with this PR.

mk/spksrc.common.mk
hgy59: And my intention is that `make all-supported` shall build the default TC version (or the highest TC version of the arch if lower) of all supported archs.
`make all-supported` is the minimal amount of packages every contributor must build to validate a pull-request.

mk/spksrc.cross-cc.mk
hgy59: No, AVAILABLE_TOOLCHAINS is a list of all available "arch-tc" toolchains.
But good to remember to rename the folder to spksrc/toolchain.

mk/spksrc.patch.mk
hgy59: `subarch` was confusing me as it really means the arch. I use group to distinguish PPC, ARM5, ARM7, ARM8, x64 etc.. Family is already used to distiguish CPU families like intel core, intel atom etc. 
You already defined the lowest level like `qoriq-6.1/*.patch`, I only introduced a intermediate level like `ppc-5.2/*.patch`.

mk/spksrc.spk.mk
hgy59: yes indeed. First I renamed ARCHS_NO_KRNLSUPP to ARCHS_WITH_KERNEL_SUPPORT to give it a correct naming.

cross/bison/Makefile
```diff
@@ -12,7 +12,10 @@ COMMENT  = General-purpose parser generator
 LICENSE  = GPLv3
 
 GNU_CONFIGURE = 1
-ifeq ($(findstring $(ARCH), powerpc ppc824x ppc853x ppc854x),$(ARCH))
+
+include ../../mk/spksrc.archs.mk
+
+ifeq ($(findstring $(ARCH),$(OLD_PCC_ARCHS)),$(ARCH))
```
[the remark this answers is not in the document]
hgy59: We should clarify this.
Until now I used `legacy` in this context for `non generic archs` and not for _old_ ones.

cross/bison/Makefile
```diff
@@ -12,7 +12,10 @@ COMMENT  = General-purpose parser generator
 LICENSE  = GPLv3
 
 GNU_CONFIGURE = 1
-ifeq ($(findstring $(ARCH), powerpc ppc824x ppc853x ppc854x),$(ARCH))
+
+include ../../mk/spksrc.archs.mk
```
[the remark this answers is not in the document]
hgy59: `spksrc.archs.mk` *is included* in `spksrc.cross-cc.mk` and other mk files (all that include `spksrc.common.mk`).
But sometimes you need to have the ARCHS definitions available before the include of `spksrc.cross-cc.mk` and for such situations I prefer to include the new `spksrc.archs.mk` instead of `spksrc.common.mk`. And from the filename you can see that such includes are for arch definitions only.

diyspk/dar/Makefile
hgy59: Sorry, my mistake. Forgot to remove this ...

cross/bison/Makefile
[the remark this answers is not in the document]
th0ma7: Agreed.  I see `DEPRECATED` and `LEGACY` as being of interest... one being now unsupported and the other being "old"... We could add `VINTAGE` for the fun of it :)

mk/spksrc.common.mk
hgy59: I it is my request to reduce the scope and to have a single file that contains the available archs.

mk/spksrc.archs.mk
th0ma7: I have no issue with that beside that in reality `ARM7` family actually includes arm v3-4-5 architectures.
So the usage of ARM with no `V` isn't appropriate overall as it describe something else entirely.
Personally, I'd rather rename all `ARMX` to using a **V** such as `ARMV7_ARCHS`

mk/spksrc.common.mk
th0ma7: Besides the `apollolake-6.2.3 geminilake-6.2.3 purley-6.2.3` my previous changes the intent was that `all-supported` output be closer to github-action.  Now that would be different and move back to the previous behavior if I understand correctly?

As for the `FORCED_ARCH` I'll look into that later once this PR is merged.

mk/spksrc.patch.mk
th0ma7: But that might become quite useful for DSM7 migration later on.  I think it's a really good addition.

mk/spksrc.spk.mk
hgy59: done
_(I just built diyspk/tree that has no dependencies and is very fast...)_

mk/spksrc.archs.mk
hgy59: done

mk/spksrc.archs.mk
hgy59: Finally renamed to match your preference 😃

mk/spksrc.spk.mk
th0ma7: Thnx for the tip, good to know :)

th0ma7:
Wow this is massive: close to 500 file changed... (most of it due to the `toolchain` renaming most probably).
Allow me a couple of days to fully review that.  At first glance this is looking good.

th0ma7:
It seems somehow my intent was to propose a change but it got pushed in... no wonder why I prefer command lines to web interface... *sigh*.  Sorry for that.

I think you'll get what the intent is.  As we are often using the exact same combination of "LEGACY" archs that a flagged as unsupported might be worth flagging them as legacy?

cross/haproxy/Makefile
```diff
@@ -20,20 +20,20 @@ BUILD_OPTIONS = USE_TPROXY=1 USE_PTHREAD_PSHARED=1 USE_OPENSSL=1 USE_ZLIB=1
...
 BUILD_OPTIONS += USE_LUA=1 LUA_INC=$(INSTALL_DIR)$(INSTALL_PREFIX)/include LUA_LIB=$(INSTALL_DIR)$(INSTALL_PREFIX)/lib
 
-include ../../mk/spksrc.common.mk
+include ../../mk/spksrc.archs.mk
 
 ARCH_OPTIONS =
 
-ifeq ($(findstring $(ARCH), $(ARM5_ARCHES) powerpc ppc824x ppc853x ppc854x),$(ARCH))
-# ARM5_ARCHES and PPC_ARCHES (except QorIQ) have older kernels and need rt library for clock_gettime()
+ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS) $(OLD_PPC_ARCHS)),$(ARCH))
```
th0ma7, line 27: Again, similar, use `LEGACY_ARCHS`

cross/libthreadar/Makefile
```diff
@@ -8,7 +8,7 @@ PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 DEPENDS =
 
 # archs without C++11 support are not supported:
-UNSUPPORTED_ARCHS = $(ARM5_ARCHES) hi3535 powerpc ppc824x ppc853x ppc854x
+UNSUPPORTED_ARCHS = $(ARMv5_ARCHS) $(ARMv7L_ARCHS) $(OLD_PPC_ARCHS)
```
th0ma7, line 11: I've been seing ARMv7L_ARCHS often included with ARMv5 + PPC_LEGACY... could ARMv7L be simply added to the generic LEGACY_ARCHS based on my proposed commit?

mk/spksrc.cross-cmake-env.mk
```diff
@@ -41,25 +41,25 @@ ifeq ($(strip $(CMAKE_BUILD_DIR)),)
...
 endif
-ifeq ($(findstring $(ARCH),$(ARM7_ARCHES)),$(ARCH))
+ifeq ($(findstring $(ARCH),$(ARM7_ARCHS)),$(ARCH))
   CMAKE_ARGS += -DCMAKE_CXX_FLAGS=-fPIC -DCROSS_COMPILE_ARM=ON
   CMAKE_ARGS += -DCMAKE_SYSTEM_PROCESSOR=armv7
 endif
-ifeq ($(findstring $(ARCH),$(ARM8_ARCHES)),$(ARCH))
+ifeq ($(findstring $(ARCH),$(ARM8_ARCHS)),$(ARCH))
   CMAKE_ARGS += -DCMAKE_CXX_FLAGS=-fPIC -DCROSS_COMPILE_ARM=ON
   CMAKE_ARGS += -DCMAKE_SYSTEM_PROCESSOR=aarch64
```
th0ma7, line 54: Shouldn't they use the `v` to now refer to `ARMvX` ?

mk/spksrc.cross-go-env.mk
```diff
@@ -1,8 +1,7 @@
 # Configuration for go build
 # 
 
-GENERIC_ARCHS = ARM7
```
th0ma7, line 4: Be careful  with the `GENERIC_ARCHS` as I had found it had a direct impact on the build process where sub-arm7 targets gets built instead of the generic arm7.  Does your change keeps this or is it addressed differently?

spk/syncthing/Makefile
```diff
@@ -6,8 +6,7 @@ DSM_UI_DIR = app
 
 DEPENDS = cross/busybox cross/$(SPK_NAME)
 
-UNSUPPORTED_ARCHS = $(PPC_ARCHES)
-GENERIC_ARCHS = ARM7
```
th0ma7, line 10: Again, this GENERIC_ARCHS which had an implication towards GO arm builds.

th0ma7 (changes requested):
So last bit of comments:
- Overall looking quite good
- Really much like the reworking of the archs.mk, should be more flexible and easier to maintain
- Much like the new patch/arch-DSMVER.patch funcionality.  Should help for DSM7 migration.

Proposals:
- Rename OLD_PPC_ARCHS into LEGACY_PPC_ARCHS
- Add a LEGACY_ARCHS that might include ARM5 + potentially also hi3535
- Document the GENERIC_ARCHS which is I recall had an implication over GO related builds

th0ma7:
Quick question, could the github-action be modified so hi3535 builds are done?  perhaps instead of armv7-1.2 for instance thus aligning the `all-supported` further with the github-action.

hgy59:
> 
> 
> Quick question, could the github-action be modified so hi3535 builds are done? perhaps instead of armv7-1.2 for instance thus aligning the `all-supported` further with the github-action.

I wanted to suggest that too 👍

mk/spksrc.archs.mk
```diff
@@ -33,4 +33,4 @@ ARCHS_WITH_GENERIC_SUPPORT = $(filter-out $(GENERIC_ARCHS), $(ARM7_ARCHS) $(ARM8
 OLD_PPC_ARCHS = powerpc ppc824x ppc853x ppc854x
 
 # outdated unsupported archs
-OBSOLETE_ARCHS = powerpc ppc824x ppc854x x86
+DEPRECATED_ARCHS = powerpc ppc824x ppc854x x86
```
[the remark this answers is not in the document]
hgy59: I propose to solve this in a different manner.
The real cause for unsupported archs is based on old kernel version (e.g. <= 2.6) or toolchains with too old glibc or compiler.

For this we can use or introduce TC-variables like
- TC_KERNEL = 3.2.40+
- TC_GLIBC = 2.20
- TC_GCC = 4.9.3

we still need arch based requirements like

and we could then implement requirements like
SUPPORTED_GCC_MIN = 4.9.6
SUPPORTED_GLIBC_MIN = 2.22


we could implement such rules in a future PR, but as this PR touches all the toolchain files we could implement it here. 
I propose to add new TC variables in this PR and implement the algorithms in a future PR.

cross/lame/Makefile
```diff
@@ -18,7 +18,7 @@ NASM_PATH = $(WORK_DIR)/../../../native/nasm/work-native/install/usr/local/bin
 
 include ../../mk/spksrc.cross-cc.mk
 
-ifeq ($(findstring $(ARCH), $(x86_ARCHES)),$(ARCH))
```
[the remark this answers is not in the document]
hgy59: @th0ma7 you must have an outdated copy. Actual source is:
```ifeq ($(findstring $(ARCH),$(x64_ARCHS) $(i686_ARCHS)),$(ARCH))```

File: mk/spksrc.archs.mk [hunk 20]
```diff
@@ -29,8 +29,12 @@ x64_ARCHS = $(GENERIC_x64_ARCH) apollolake avoton braswell broadwell broadwellnk
...
 # PPC_ARCHS except qoriq
-OLD_PPC_ARCHS = powerpc ppc824x ppc853x ppc854x
+LEGACY_PPC_ARCHS = $(DEPRECATED_PPC_ARCHS) ppc853x
+LEGACY_ARCHS = $(LEGACY_PPC_ARCHS) $(ARMv5_ARCHS)
 
 # outdated unsupported archs
-DEPRECATED_ARCHS = powerpc ppc824x ppc854x x86
+DEPRECATED_PPC_ARCHS = powerpc ppc824x ppc854x
+DEPRECATED_i686_ARCHS = x86
+DEPRECATED_ARCHS = $(DEPRECATED_PPC_ARCHS) $(DEPRECATED_i686_ARCHS)
```
hgy59: Sorry, had to revert these changes to avoid conflicts when I brought back the x86 arch.

mk/spksrc.spk.mk
hgy59: This variable was framework internal and not documented in any wiki page.

th0ma7 (changes requested):
Why not rename x86 to pineview so it is consistent?

mk/spksrc.archs.mk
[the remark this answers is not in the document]
th0ma7: I really much like the idea.
Added to a mutual TODO list I guess, but only after this PR.
Just a thought, would probably require its own `spksrc.requirements.mk` where we could review the SRM|DSM min requirement variable so we can use that `.mk` directly when needed...

mk/spksrc.spk.mk
th0ma7: And could potentially be reworked with the TC_* requirement makefile...

mk/spksrc.wheel.mk
```diff
@@ -59,7 +59,7 @@ pre_wheel_target: wheel_msg_target
 
 build_wheel_target: $(PRE_WHEEL_TARGET)
 	@if [ ! -z "$(WHEELS)" ] ; then \
-		$(foreach e,$(shell cat $(WORK_DIR)/python-cc.mk),$(eval $(e))) \
+		$(foreach e,$(shell cat $(WORK_DIR)/python-cc.mk 2> /dev/null),$(eval $(e))) \
```
th0ma7, line 62: Shouldn't be echoed somewhere somehow?  Even perhaps in `$(WORK_DIR)/python-cc.mk.error.log`.  Otherwise redirecting it to `/dev/null` will complexity debugging.

spk/rutorrent/Makefile
```diff
@@ -5,12 +5,12 @@ SPK_ICON = src/rutorrent.png
 DSM_UI_DIR = app
 
 # some dependents require C++11 support
```
th0ma7, line 7: Perhaps something to add in the requirement.mk file (or what ever the name becomes).
Or manage using GCC_MIN_VERSION... food for thoughts.

th0ma7 (changes requested):
So really much like the idea of managing at TC_* level the GLIB, GCC and KERNEL which should in theory remove all the UNSUPPORTED and the likes.  Proposal's left in my review are:
- GENERIC_ARCHS vs GO build support - is it handled with your changes?
- python error redirection to /dev/null
- renaming x86 to pineview and include it in x64 I guess?

hgy59:
@th0ma7 as you introduced the `TC_KERNEL` variable, can you tell me whether it is mandatory to have the `+` appended to the version?
As the toolchain files have no `+` in the name I would like to ommit this in `TC_KERNEL`.

hgy59:
> Why not rename x86 to pineview so it is consistent?

No, please not.
The official arch name of synology.com ist `X86`/`x86` (look for DS411+ and other atom processors in the arch lists of [synology wiki](https://www.synology.com/en-us/knowledgebase/DSM/tutorial/Compatibility_Peripherals/What_kind_of_CPU_does_my_NAS_have) and [synocommunity wiki](https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures).)

The name pineview appears only in the toolchain file name. And the `pineview` toolchain is used for the generix x64 arch too.

mk/spksrc.cross-cmake-env.mk
hgy59: oops

mk/spksrc.cross-go-env.mk
hgy59: I cannot see any sideeffect by the removal of `GENERIC_ARCHS`.
IMHO it is obsolete since the introduction of generic armv7 arch (or at least since all-supported builds armv7 generic arch).

mk/spksrc.wheel.mk
hgy59: It is boring me for a long time, that it looks like an error for packages without python-wheels and regularly no python-cc.mk file.
```
===>  Processing wheels of {package-name}
cat: /spksrc/spk/{package-name}/work-{arch}-6.1/python-cc.mk: No such file or directory
```

The only error output that `cat` might produce is `No such file or directory`.


The real problem might be that `$(WHEELS)` is not empty. But anyway this should be moved to a dedicated PR.

mk/spksrc.archs.mk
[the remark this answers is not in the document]
hgy59: Such a requirements file is a good idea, particularly as the github action has to recognize and ignore all kind of unsupported errors.

cross/monit/Makefile
```diff
@@ -18,9 +18,9 @@ CONFIGURE_ARGS = --without-pam
 #CONFIGURE_ARGS += --enable-optimized
 CONFIGURE_ARGS += libmonit_cv_setjmp_available=yes libmonit_cv_vsnprintf_c99_conformant=yes 
 
-include ../../mk/spksrc.common.mk
+include ../../mk/spksrc.archs.mk
 
-ifeq ($(findstring $(ARCH),$(PPC_ARCHES)),$(ARCH))
+ifeq ($(findstring $(ARCH),$(PPC_ARCHS)),$(ARCH))
 CONFIGURE_ARGS += ac_cv_c_bigendian=yes
```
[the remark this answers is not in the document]
hgy59: AFAIR the configure of monit was not able to evaluate the correct endianess. So this is a workaround for it.

mk/spksrc.cross-go-env.mk
th0ma7: Perhaps someone knows where this came from originally...  From the history it looks like this came from @ymartin59 , any clue on this, is this still a requirement?

mk/spksrc.wheel.mk
th0ma7: indeed that is the only side effect but at least we know when it is happening.  Why not simply redirect it to stdout (`2>&1`)?

th0ma7:
> @th0ma7 as you introduced the `TC_KERNEL` variable, can you tell me whether it is mandatory to have the `+` appended to the version?
> As the toolchain files have no `+` in the name I would like to omit this in `TC_KERNEL`.
> 
> Edit:
> @th0ma7 sorry, I didn't await your feedback on this and removed the `+` in existing TC_KERNEL definitions.

This will have to be revisited.  In order for the modules to properly load it requires a perfect `uname` match and Synology is adding a `+` flag at the end of its version such as:
```
$ uname -r
4.4.59+
```
I'll check if there is a way to compile the kernel and passing a `EXTRAVERSION` through a variable instead.

th0ma7:
I was hoping we would aim towards automating the kernel, gcc and glibc version by pulling it out from the source... But I know that kernel pre-3.something is not easy to reference (e.g. `make version`).  Thnx for going through all the manual work of finding it.

th0ma7 (changes requested):
You forgot to modify `spksrc.tc.mk` in order to echo the new variables:
`TC_KERNEL`
`TC_GCC`
`TC_GLIBC`

hgy59:
> This will have to be revisited. In order for the modules to properly load it requires a perfect `uname` match and Synology is adding a `+` flag at the end of its version such as:
> 
> ```
> $ uname -r
> 4.4.59+
> ```
> 
> I'll check if there is a way to compile the kernel and passing a `EXTRAVERSION` through a variable instead.

Of all my diskstations only the 4.x kernels have the `+`  with `uname -r`.

th0ma7 (commented):
Good catch!

mk/spksrc.cross-go-env.mk
hgy59: That was me that introduced it with commit 27445308dfa27b2892b1ddfb9af721b348ffbbdb 😄

th0ma7: Even better.. :1st_place_medal:

th0ma7 (approved):

hgy59:
@ymartin59 @publicarray please have a look at this PR. As all toolchain makefiles are modified you may have additional ideas for future requirements, particularly related to DSM7.

th0ma7:
@hgy59 I would suggest one last change: dropping `ppc853x-5.2` entirely e.g. removing from the default build process on both github-action & `make all-supported`.

publicarray (commented):
Regarding DSM7 stuff I can't think of anything specific that would relate to this PR at this point.

Awesome cleanup though!

hgy59:
> @hgy59 I would suggest one last change: dropping `ppc853x-5.2` entirely e.g. removing from the default build process on both github-action & `make all-supported`.

Done ✔️ 
Regarding the [Synology Product Update Status](https://www.synology.com/en-global/products/status) the x10 models are now End of Life. When all models get EOL 10 years after models postfix, the next deprecated arch will be 88f6281 in 2023.

_but I still will build packages for ppc853x-5.2 as I have a DS-210+ that is far more performant than my DS-115j :smile:_

.github/workflows/build.yml
```diff
@@ -23,17 +23,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # x64=x86_64, evansport=x86, aarch64=armv8, armv7, 88f6281=armv5, ppc853x/qoriq=ppc
+        # x64=x86_64, evansport=i686, aarch64=armv8, armv7, hi3535=armv7l, 88f6281=armv5, ppc853x/qoriq=ppc
```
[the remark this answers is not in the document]
hgy59: Done ✔️ 
As there is no x86 generic arch and there will be no i686, there is no problem with updated information before the merge.

publicarray:
Just in case it's not obvious: to migrate to the singular toolchain and without redownloading them
just execute `mv  /spksrc/distrib/toolchains /spksrc/distrib/toolchain` in docker.

PR #4321, opened 2020-12-20
spksrc.native-cmake.mk

th0ma7:
_Motivation:_  Some native package requires `cmake` to be built.  New Makefile definitions helpers allows building them easily.  Included in this PR is an example to build `ninja` using `spksrc.native-cmake*.mk` helper files.
_Linked issues:_  #4307

### Checklist

### Important note
The usage of a specific `native/cmake` version is not mandatory as default Debian image already provides one.  In the example provided `ninja` needed an earlier version thus `spksrc.native-cmake-env.mk` helper code make usage of the following:
```
# Use native cmake
DEPENDS += native/cmake
CMAKE_PATH = $(WORK_DIR)/../../../native/cmake/work-native/install/usr/local/bin
ENV += PATH=$(CMAKE_PATH):$$PATH
```
If this PR is being merge, the usage of `native/cmake` within `spksrc.native-cmake*.mk` could in theory be discarded.


hgy59:
@th0ma7 is there a reason not to remove the installed cmake completely (from Dockerfile) and only use native/cmake?
This would keep the *.mk files clean (avoid spksrc.native-cmake*.mk files)

th0ma7:
@hgy59 it all started with the need of having `native/ninja` which in turns needed a `spksrc.native-cmake.mk` so I could build native package using `cmake`.  It turned out that `ninja` also needed a newer version of `cmake` which led to the addition of `native/cmake` as well.

In the end the default Debian `ninja` works great so there is no need for neither `native/ninja` or `native/cmake`.  Althgouh the `spksrc.native-cmake*.mk` can still be of interest.  Question is: what to do about them?

Personally I would (and had done previously):
- remove `native/ninja` and `native/cmake`
- remove references to `native/cmake` in `spksrc.native-cmake*.mk`
- keep the `spksrc.native-cmake*.mk`

But then again, without any examples or usages, is it really worth it?  @ymartin59 was suggesting to just be keeping the PR alive in case we ever need it.

hgy59:
> @ymartin59 was suggesting to just be keeping the PR alive in case we ever need it.

you can create a branch like `native-cmake` in this repo (similar as we have dsm7 branch) and create and merge a PR with the spksrc.native-cmake*.mk files there, so these are not lost.

Issue #4334, opened 2020-12-27
update sshfs of synocli-net


hgy59:
This is a followup of #2661

update sshfs to version 3.x
- depends on cross/libfuse version 3.x
- cross/libfuse version 3.x needs meson build (#4307)

th0ma7:
I have an early prototype where I was able to compile `udev` from systemd-204 (same version as DSM-6) using the toolkit framework with a few modifications.  It applies on top of #4307.

I'll submit a [WIP] PR later today/tomorrow, please consider that code to be of prototype level at this early stage.

PR #4335, opened 2020-12-27
Update to libfuse3

th0ma7:
_Motivation:_  With the upcoming support of meson+ninja it's now possible to update `libfuse` & `sshfs`.
_Linked issues:_  #4334 

### Checklist

### Note
This is in early dev/prototype.  Not to be used as-is.


th0ma7:
@hgy59 so basically what does this do:
1. It download the relevant toolkit (for now I only added `apollolake` 6.x on top of the existing `evansport`)
2. Add compiler & linker flags to add the toolkit headers and libraries (really not satisfied of the code as I'm running from the `cross/` directory where some variables are not available, more of a proof of concept stage)
3. Re-added a `native/glib` in order to have access to `bin/glib-genmarshal`.  So I introduced a `spksrc.native-meson.mk` along with `native/glib` to #4307 (DONE).
3. Compiled `udev` from `systemd`. To do so I have to fully compile `systemd` and that was quite a pain to do...  Probably needs more optimizing and perhaps rearranging the directory structure so it is similar to the one on DSM.  At least, its a proof of concept that it can be compiled.
4. That allows to compile `libfuse` using meson+ninja and linking to `systemd/udev`.
5. Now ready to compile `sshfs` using all of the above points... Currently fails at finding `glib` headers, probably not that an issue (either needs using `ENV` variables or adjusting toolkit variables to add proper includes).

And that is considering that you merge #4307 on top of this.
Builds successfully using `make arch-apollolake-6.2.3` from `cross/fuse` directory.
Currently fails from `cross/sshfs`.

hgy59:
@th0ma7 I propose to keep cross/fuse untouched and to create new cross/fuse3 with libfuse >= 3.x.
This will allow usage of libfuse 2.x for packages that do not support libfuse 3.x (or do not support meson build system)

th0ma7:
@hgy59 I'm having an issue on managing `TOOLKIT*` variables where you might be able to help.

Basically I'm trying to add as build environment-wide accessible variables various TOOLKIT* variables but I just can't figure how to achieve this.  That is why I ended-up "forcing" the tc environment to include tk FLAGS into the `tc_vars.mk` in the first place.

What I've been trying to achieve (ref: https://github.com/SynoCommunity/spksrc/pull/4335/commits/6e51315e5c74e1be8ae5e94c44974b04d3be77c8) is to handle multiple sub-toolkit include path that can be passed as parameters.  Current code-path fails totally as none of the variable gets passed to the toolkit makefiles.

This would allow compiling some of the tools against the toolkit where we see fit instead of rebuilding from the grounds-up.

Help on this would be much appreciated.

cross/systemd/Makefile
```diff
@@ -0,0 +1,59 @@
...
+	# gperf >= 3.1 now uses 'size_t' rather than 'unsigned'
+	$(RUN) sed -i.bak -e "/journald_gperf_lookup/s/unsigned/size_t/" $(WORK_DIR)/$(PKG_DIR)/src/journal/journald-server.h
+	# Modify line after: definition spreads over 2x lines
+	$(RUN) sed -i.bak ":1;/lookup_syscall/{n;s/unsigned int/size_t/;b1}" $(WORK_DIR)/$(PKG_DIR)/src/core/syscall-list.c
+	$(RUN) sed -i.bak -e "/load_fragment_gperf_lookup/s/unsigned/size_t/" $(WORK_DIR)/$(PKG_DIR)/src/core/load-fragment.h
+	$(RUN) sed -i.bak -e "/logind_gperf_lookup/s/unsigned/size_t/" $(WORK_DIR)/$(PKG_DIR)/src/login/logind.h
+	$(RUN) sed -i.bak -e "s/SG_FLAG_LUN_INHIBIT/SG_FLAG_UNUSED_LUN_INHIBIT/g" $(WORK_DIR)/$(PKG_DIR)/src/udev/cdrom_id/cdrom_id.c
+	$(RUN) sed -i.bak -e "/lookup_key/s/unsigned int/size_t/" $(WORK_DIR)/$(PKG_DIR)/src/udev/keymap/keymap.c
+	# include <stdint.h> for missing types for error: unknown type name 'uint32_t' and 'uint8_t'
+	$(RUN) sed -i.bak -e "/mtd-user.h/a\$\\#include <stdint.h>" $(WORK_DIR)/$(PKG_DIR)/src/udev/mtd_probe/mtd_probe.h
```
ymartin59, line 54: Please use framework patch support for easier maintenance and readability: more comments, compare before/after in context, detect upstream changes that may break replacement...

cross/systemd/Makefile
```diff
@@ -0,0 +1,59 @@
+PKG_NAME = systemd
```
ymartin59, line 1: I guess you do not expect to provide "systemd" on its whole on DSM 6 (or 7) but only grab "udev" support as far as I understand. If so, I propose to name it as `cross/systemd-udev` or other relevant postfix to identify concerned subsystem.

ymartin59 (approved):
Great job. This sshfs update is a long waiting delivery and you did progress there. I have some minor improvements request (as usual)

hgy59:
As `libfuse3` and thus `sshfs` package depends on arch-tc specific toolkit, it does not suit in the **generic** synocli-net package.
As we don't want to publish synocli-net for each arch/tc combination, sshfs must be moved from synocli-net into a dedicated package.

cross/systemd/Makefile
th0ma7: This is early work to get something to compile.
This code gets recreated dynamically... I'll see what I can do when I come back on this PR.

cross/systemd/Makefile
th0ma7: agreed, make sense.

th0ma7:
@hgy59 I'm not even sure this is the right approach... Indeed currently it is arch specific using toolkit.  There may be other options, duno yet.  I'll have to put more time on this to confirm a path forward.

th0ma7:
Note to self: it is possible to build fuse3 without udev by passing something like:
```
  - -Dudevrulesdir=/app/etc/udev/rules.d
```
Other potential candidate to look at (or not) `  - -Duseroot=false`.
Info from https://github.com/libfuse/libfuse/issues/668

cross/systemd/Makefile
hgy59: The patched source files are generated with `gperf` and gperf >= 3.1 generates different types than the version in 2013 (we need such an old version of systemd for compatibility).
Therefore the framework patch support is not applicable here.

hgy59:
@th0ma7 I have already rebased this PR to current master and am about to fix some cross compilation issues before pushing.

Will use this (merged) PR for an update of synocli-net.

th0ma7:
@hgy59 this PR is relatively heavy on framework changes.  Are you still using the toolkit part or not anymore as if I recall many changes where on that side of the fence which would then require including all archs toolkit into the loop?

Good thing you where able to rebase it against master :)

hgy59:
> Are you still using the toolkit part

no, the toolkit part is gone. 
I haven't a clue what this was for.
```
REQUIRE_TOOLKIT = 1
TOOLKIT = glib

...

# Force indlude path to toolkit dbus & glib
#ENV += "DBUS_CFLAGS=-I$(TOOLKIT_ROOT)/include/dbus-1.0 -I$(TOOLKIT_ROOT)/lib/dbus-1.0/include"
#ENV += "GLIB_CFLAGS=-I$(TOOLKIT_ROOT)/include/glib-2.0 -I$(TOOLKIT_ROOT)/lib/glib-2.0/include"
```

sshfs3 now uses cross/glib.

hgy59:
@th0ma7 do you have a use case, so we can test that fuse3 works as expected?

th0ma7:
> no, the toolkit part is gone. I haven't a clue what this was for.

This was a failed attempt at using the toolkit glib headers in trying to get libfuse3 to build.  I recall going deep in that rabbit hole then looking into using systemd-udev either by building it or my using the headers from the toolkit.

> sshfs3 now uses cross/glib.

If I recall correctly, the issue was glib being crossed-compiled using [cmake without a toolchain file](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html).  This resulted in bulding the entire glib for the said target arch whereas normally some bits mandatory need to get built for the host arch.  As a result it wasn't possible to build libfuse3 against it and I ended-up trying to build it against the toolkit header files (wich failed as well).

I believe I ended-up fixing that entirely when adding the cmake toolchain file #5280 which in turns now allows building libfuse3 against that said glib which wasn't feasible previously.  That would then mean that all toolkit parts of this PR can be removed entirely.

> do you have a use case, so we can test that fuse3 works as expected?

We could proably find something meaningful using recent sshfs linked against libfuse3?  Although I don't have one that comes to mind.

hgy59:
> That would then mean that all toolkit parts of this PR can be removed entirely.

So it is ok to remove changes in `mk/spksrc.toolkit-flags.mk` (from `TOOLKIT_CFLAGS = -I$(TOOLKIT_ROOT)/include` to EOF)?

This is the only toolkit related change.

_I propose to keep the toolkits added for apollolake_

hgy59:
Successfully tested sshfs 3.7.3 on DS218 (aarch64) with **DSM 6.2.4-25556 Update 6** and on DS218+ (x64) with **DSM 7.1.1-42962 Update 2**.

th0ma7:
> > That would then mean that all toolkit parts of this PR can be removed entirely.
> 
> So it is ok to remove changes in `mk/spksrc.toolkit-flags.mk` (from `TOOLKIT_CFLAGS = -I$(TOOLKIT_ROOT)/include` to EOF)?
> 
> This is the only toolkit related change.

Yes, good to go.  I'll double check through an in-depth review but I believe it is.

> _I propose to keep the toolkits added for apollolake_

good idea.

> Successfully tested sshfs 3.7.3 on DS218 (aarch64) with DSM 6.2.4-25556 Update 6 and on DS218+ (x64) with DSM 7.1.1-42962 Update 2.

This really is awesome!  Woot!  Honestly it's great finding that older pending work can finally becoming to fruition and relevant.  Congrats on this one that took over two year to finally work out  :rocket: 

Proposal:  that you write down how you tested this in the wiki for future reference?

cross/libudev/Makefile
```diff
@@ -0,0 +1,97 @@
...
+CONFIGURE_ARGS += --disable-timedated
+CONFIGURE_ARGS += --disable-localed
+CONFIGURE_ARGS += --disable-coredump
+CONFIGURE_ARGS += --disable-polkit
+CONFIGURE_ARGS += --disable-efi
+CONFIGURE_ARGS += --disable-myhostname
+CONFIGURE_ARGS += --disable-keymap
+CONFIGURE_ARGS += --with-dbuspolicydir=$(INSTALL_PREFIX)/share/dbus-1/system.d
+CONFIGURE_ARGS += --with-dbussessionservicedir=$(INSTALL_PREFIX)/share/dbus-1/services
+CONFIGURE_ARGS += --with-dbussystemservicedir=$(INSTALL_PREFIX)/share/dbus-1/system-services
```
th0ma7, line 69: Opened question: should theses aligne with actual DSM install path and therefore rely on them?

DSM6 (And what about the non-existing ones, probably due to older version in use?) :
```
/usr/share/dbus-1/services
/usr/share/dbus-1/system-services
```
DSM7 (still missing `interfaces` directory):
```
/usr/share/dbus-1/services
/usr/share/dbus-1/system-services
/usr/share/dbus-1/system.d
```

cross/libudev/Makefile
```diff
@@ -0,0 +1,97 @@
+PKG_NAME = libudev
+PKG_REAL_NAME = systemd
+PKG_VERS = 204
```
th0ma7, line 3: Trying to remember why I selected release 204?
1. because this was the minimum version needed for fuse3?
2. because that is the default version for DSM7?
```
# udevadm --version
219
```
3. because that is the default version for DSM6?  `<--- probably it`
```
# udevadm --version
204
```

Question ends-up being, should we align `udev` with DSM releases?  I would believe it would be best?

mk/spksrc.tc.mk
```diff
@@ -77,21 +77,25 @@ else ifeq ($(findstring $(ARCH),$(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
...
-	echo "set(CMAKE_OBJDUMP $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)objdump)" ; \
-	echo "set(CMAKE_RANLIB $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)ranlib)" ; \
-	echo "set(CMAKE_READELF $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)readelf)" ; \
-	echo "set(CMAKE_STRIP $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)strip)" ; \
+	@echo "# define compilers and tools to use" ; \
+	for tool in $(TOOLS) ; \
+	do \
+	  target=$$(echo $${tool} | sed 's/\(.*\):\(.*\)/\1/' | tr [:lower:] [:upper:] ) ; \
+	  source=$$(echo $${tool} | sed 's/\(.*\):\(.*\)/\2/' ) ; \
+	  if [ "$${target}" = "CC" ] ; then \
```
th0ma7, line 89: Good idea to re-use code from `tc_vars:`

th0ma7 (commented):
Besides the udev versions for DSM6 vs DSM7 this looks really good to me.

th0ma7:
@hgy59 related to the build errors, I've double check and from the `evansport` toolkit I can confirm it is also using systemd-219.  So I downloaded the evansport systemd sources from https://archive.synology.com/download/ToolChain/Synology%20NAS%20GPL%20Source/7.0-41890 and there is a `config.h.in` in there that may provide additional options being passed to disable other features.

Further looking around, there is a `SynoBuildConf` directory which includes a `build` file with the following:

```
env \
	CC="${CC}" \
	CXX="${CXX}" \
	LD="${LD}" \
	CFLAGS="${CFLAGS} ${SDKGPL_INCS} -fPIC" \
	LDFLAGS="${LDFLAGS} ${SDKGPL_LIBS} -lrt -lcap -lattr" \
	AR=${AR} \
	STRIP=${STRIP} \
	RANLIB=${RANLIB} \
	NM=${NM} \
	ac_cv_func_malloc_0_nonnull=yes \
	${LD_GC_SECTION_BROKEN} \
	./configure ${ConfigOpt} \
	--enable-largefile \
	--disable-nls \
	--disable-address-sanitizer \
	--disable-undefined-sanitizer \
	--disable-python-devel \
	--disable-dbus \
	--disable-utmp \
	--enable-kmod \
	--disable-xkbcommon \
	--enable-blkid \
	--disable-seccomp \
	--disable-ima \
	--disable-chkconfig \
	--disable-selinux \
	--disable-apparmor \
	--enable-xz \
	--disable-zlib \
	--disable-bzip2 \
	--disable-lz4 \
	--enable-pam \
	--disable-acl \
	--disable-smack \
	--disable-gcrypt \
	--disable-audit \
	--disable-elfutils \
	--disable-libcryptsetup \
	--disable-qrencode \
	--disable-microhttpd \
	--disable-gnutls \
	--disable-libcurl \
	--disable-libidn \
	--disable-libiptc \
	--disable-binfmt \
	--disable-vconsole \
	--disable-bootchart \
	--disable-quotacheck \
	--enable-tmpfiles \
	--disable-sysusers \
	--disable-firstboot \
	--disable-randomseed \
	--disable-backlight \
	--disable-rfkill \
	--enable-logind \
	--disable-machined \
	--disable-importd \
	--disable-hostnamed \
	--disable-timedated \
	--disable-timesyncd \
	--disable-localed \
	--disable-coredump \
	--disable-polkit \
	--disable-resolved \
	--disable-networkd \
	--disable-efi \
	--disable-terminal \
	--disable-kdbus \
	--disable-myhostname \
	--disable-gudev \
	--disable-hwdb \
	--disable-manpages \
	--disable-hibernate \
	--disable-ldconfig \
	--libexecdir=/usr/lib \
	--sysconfdir=/etc \

make ${MAKE_FLAGS}
```
I thought there may be some usefull stuff in there to try out.

cross/libudev/Makefile
hgy59: And version 204 is also on my DS210+ with DSM 5.2

cross/libudev/Makefile
hgy59: > EDIT: Further investigating I believe `interfaces` is restricted to desktop components (e.g. `org.freedesktop.*` , `org.gnome.Shell.*`, `org.kde.*`) and is probably not needed at all.

interfaces directory is not supported/configurable in systemd-219 anymore.

th0ma7: Question really is, currently we are making theses directories to point to the package path, but should it instead point to the system-wide default dbus-1 directories?  While I do not have a definitive answer on this I'd recommend using the system-wide path, similar to openssl certificates.

cross/libudev/Makefile
hgy59: comcerto2k-7.0 builds now with libudev_219 too, but I don't know, wheter systemd 219 or 204 is installed.

th0ma7: I've asked on discord... Might get something from there on our support channel...

PR #4344, opened 2020-12-31
Added dsm7 preview toolchain for syno-armada38x-7.0

zyrill:
_Motivation:_  Added dsm7 preview toolchain for syno-armada38x-7.0 to build packages for dsm7

### Checklist
- Build of package (transmission) succeeded
- Needs more work (elsewhere?) though, as the correct setting for TC_VERS in the Makefile is "7.0%20preview" (else the tc download fails), but the correct setting for os_min_ver in the INFO file is "7.0-40000" (else the package won't install in the dsm GUI). No idea where to fix this, seems to be a double-usage of a variable that doesn't cover this edge case


publicarray:
Thanks for your contribution! and for willing to test the dsm7. But due to having #4320 merged recently I think it is best that I rebase the dsm7 branch with master first before adding more architectures. I think it will be easier that way.

publicarray:
I set the `TC_DIST_SITE_URL = https://sourceforge.net/projects/dsgpl/files/Tool%20Chain/DSM%207.0%20preview%20Tool%20Chains/` variable (we can remove it later) when/if the final versions are releases

toolchains/syno-armada38x-7.0/Makefile
```diff
@@ -0,0 +1,11 @@
+TC_ARCH = armada38x
+TC_VERS = 7.0%20preview
+TC_KERNEL = 3.10.108+
```
publicarray, line 3: For the DSM7 I assume the kernel version is higher than `4.4.xxx+`

publicarray:
```
$ uname -a
Linux NAS 4.4.180+ #41222 SMP Fri Dec 4 19:02:12 CST 2020 x86_64 GNU/Linux synology_apollolake_418play
```

publicarray: NVM They did not update the kernel for this arch :shrug:  https://sourceforge.net/projects/dsgpl/files/Tool%20Chain/DSM%207.0%20preview%20Tool%20Chains/Marvell%20Armada%2038x%20Linux%203.10.108/

publicarray:
Thanks, I've cherry picked your commit.

zyrill:
> 
> 
> Thanks for your contribution! and for willing to test the dsm7. But due to having #4320 merged recently I think it is best that I rebase the dsm7 branch with master first before adding more architectures. I think it will be easier that way.

Thank you for your help, for the PR in general and for the TC_DIST_SITE_URL issue in particular, @publicarray - much appreciated!

PR #4358, opened 2021-01-03
syno-magnet: Add support for DSM 7

neykov:
Make the package compatible with DSM7.

Build with "make noarch-6.1" & "make noarch-7.0" to get a compatible package.

~Hold off on merging before a package that's compatible with both DSM7 & DSM6 can be created. The changes here unblock manual install of the package for people that are already on DSM7.~

Fixes https://github.com/neykov/syno-magnet/issues/3.

### DIFF
--- a/spk/syno-magnet/Makefile
+++ b/spk/syno-magnet/Makefile
@@ -8,13 +8,13 @@ PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 SPK_NAME = $(PKG_NAME)
 SPK_VERS = $(PKG_VERS)
-SPK_REV = 2
+SPK_REV = 3
 SPK_ICON = src/magnet.png
 
 MAINTAINER = neykov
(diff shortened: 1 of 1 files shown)


hgy59:
@neykov as @publicarray has enhanced the framework for noarch-TCVERSION build it is now possible to build such packages for both DSM6 and DSM7 on the dsm7 branch.
see c502ef65a8fc48f7ddb4847b8bb65a104b706d6a

neykov:
Thanks @hgy59 . Updated the PR to make use of the recent changes.

publicarray:
The privilege file is auto generated by the framework. It also includes the username and group. Unless you don't want that then no change is required when you build from the DSM7 branch. (Besides a revision number bump I suppose)

neykov:
Oh missed that somehow. That's nice!
Should I close this PR then?

publicarray:
@neykov Thanks! We are trying hard to make minimal changes to the indivual packages and your package is one of the lucky ones :)
IMHO you can go ahead and close this. However I don't think we are ready publish DSM7 packages for a little while yet :)

PR #4364, opened 2021-01-06
[WIP] Update transmission dependencies

zyrill:
_Motivation:_  updates transmission deps curl to 7.74.0 and openssl to 1.1.1i which fixes [CVE-2020-1971](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-1971).

Is busybox still required? The service-setup.sh states in the comment preceding the busybox commands that they are needed to "Discard legacy obsolete busybox user account[s]" only and transmission itself doesn't need it to build. So if I understand this correctly, the busybox dependency could be removed. Could somebody with a better understanding of how the packages work comment? Thanks!

### Checklist
- [x] Built deps and the transmission package successfully (on armada-38x-7.0)
- [X] Checked that package installs and works correctly (on DS216j)


hgy59:
@zyrill thanks for contributing to synocommunity.
This PR has some major problems and is not recommended to merge.
Please:
- do not mix different topics into one PR (e.g. update only openssl and curl)
- cross/openssl and cross/curl are widely used dependencies so `transmission` is only one of them
- use your github account to commit (and not `root`)
- create different PRs to merge on master and dsm7 branches
- do not modify .github/workflows/build.yml like you did (and for armada38x-7.0 we should introduce armv7-7.0 as generic arch)
- do not modify unrelated packages (ejabberd, ffsync, monit, etc.), some of them are WIP on other PRs

zyrill:
@hgy59 thanks for your feedback, I shall structure the commit(s) according to your suggestions and clean up my mess. Thanks for your patience and taking the time to comment!

I have a few questions if you can spare another moment:

- I used the automatic builds as a litmus test to check whether my original changes (the version bumps) affect any other packages or architectures (not just transmission on armada38x-7.0 that I built and tested manually) - is there a better way to build/test packages dependent on a commit? Ideally, I'd only test those packages affected by a commit, not everything.
- Am I safe in assuming that build errors in the automatic checks are not necessarily my fault? I ask because I was under the impression that all builds have to pass else a commit would automatically disqualify for a merge - hence my modifications on other unrelated packages. If I understand your comment correctly, I may disregard build errors probably not stemming from my changes.
- Can you help me with your next to last suggestion? I don't know where and how to introduce armv7-7.0 as a generic arch.

zyrill:
It's probably easiest if I simply create multiple new PRs, so I'm closing this one.

hgy59:
>     * Can you help me with your next to last suggestion? I don't know where and how to introduce armv7-7.0 as a generic arch.

As per now we support three generic archs:
- x64 (a.k.a x86_64)
- aarch64 (a.k.a ARMv8)
- armv7

As those generic archs are not provided by synology with dedicated toolchains, we use a _best fit_ toolchain. As sole difference we list all supported archs in the toolchain/Makefile variable `TC_ARCH`

| Generic 	| based on	| used for	|
|--------------- |------ |------ |
| x64 | x86 (pineview) | apollolake avoton braswell broadwell broadwellnk bromolow cedarview denverton dockerx64 geminilake grantley purley kvmx64 v1000 x86 x86_64 |
| aarch64 |rtd1296 | rtd1296 armada37xx aarch64 |
| armv7 | armada38x | armada370 armada375 armada38x alpine alpine4k armadaxp comcerto2k monaco |

So a package built for a generic arch can be installed on all diskstation models with one of the arch listed in the _used for_ column. Generic archs must not be used for packages that depend on kernel-sources or for some packages that benefit of cpu specific features (like ffmpeg and related packages).

AFAIK the generic archs are already defined form DSM 7.0

zyrill:
Ok I think I've got it now... I'd really appreciate it if you took another look. :)

zyrill:
Are the new PRs acceptable regarding coding standards, structuring and so on, @hgy59 ?

PR #4369, opened 2021-01-07
Bump openssl version to 1.1.1i

zyrill:
_Motivation:_  Bump openssl version to 1.1.1i.
_Linked issues:_  Follow-up to (https://github.com/SynoCommunity/spksrc/pull/4364)

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4371, opened 2021-01-07
Bump libewf version to experimental-20201230 fixing synocli-disk build

zyrill:
_Motivation:_  While working on https://github.com/SynoCommunity/spksrc/pull/4369 I noticed that the synocli-disk build was failing even without changes so I investigated and noticed that the libewf devs had removed their old releases (which were also all marked experimental). They did however provide new releases so I picked the most up to date release (no stable releases exist at all, so I had to take another experimental release) and bumped the libewf version to experimental-20201230 in order to fix the synocli-disk build
_Linked issues:_  Follow-up to (https://github.com/SynoCommunity/spksrc/pull/4364)

### Checklist

### DIFF
--- a/cross/libewf/Makefile
+++ b/cross/libewf/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = libewf
-PKG_VERS = 20171104
+PKG_VERS = 20201230
 PKG_EXT = tar.gz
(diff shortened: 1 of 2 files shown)


zyrill:
I'd appreciate it if you could take a few minutes to review this PR, @publicarray.

publicarray:
Thanks for your Contribution @zyrill I suspect this will happen again until they have a proper release.

PR #4370, opened 2021-01-07
Bump curl version to 7.74.0

zyrill:
_Motivation:_  Bump curl version to 7.74.0.
_Linked issues:_  Follow-up to (https://github.com/SynoCommunity/spksrc/pull/4364)

### Checklist
- [x] Build rule `all-supported` completed successfully

Issue #4362, opened 2021-01-06
Transmission DSM7: 403: Forbidden


publicarray:
403: Forbidden
Unauthorized IP Address.

Either disable the IP address whitelist or add your address to it.
If you're editing settings.json, see the 'rpc-whitelist' and 'rpc-whitelist-enabled' entries
If you're still using ACLs, use a whitelist instead. See the transmission-daemon manpage for details.

_Originally posted by @tianchaomaker in https://github.com/SynoCommunity/spksrc/issues/4215#issuecomment-753692984_

publicarray:
@tianchaomaker

I moved our discussion to a new issue.

The install script only updates the config file if  `/var/packages/transmission/var/` is empty. The previous version did not uninstall all files correctly. Can you uninstall transmission again and let me know if there are any file present in that locations before installing? Thanks

publicarray:
@tianchaomaker how did it go? any updates?

publicarray:
No response and I can't reproduce with the latest changes. Closing

Issue #4313, opened 2020-12-16
Can‘t start transmission on DSM7.0


jinfeng1985:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ transmission
_Package Version:_ 3.0

_NAS Model:_ synology
_NAS Architecture:_ 920+
_DSM version:_ 7.0

### Expected behavior
No root permission for installation and running
Will the software support 7.0?

publicarray:
In DSM7 packages can not run as a root user anymore and so The scripts need updating. Please see issue #4215 and the dsm7 branch for current progress. Pull requests are welcome.

lbwstco:
@hgy59 would you help fix this problem soon?

zyrill:
transmission also needs additional "resources" in order to access the files in the download folder. I've managed to get it running manually (with root in the terminal), but starting from the GUI ends up in transmission not being able to access the files or read/save its state.

publicarray:
I've added added the ability to add shared folders through "resources". So if you want you should be able to build and use transmission from the dsm7 branch without root. Please let me know if there are any problems.

Running as root may damage your permissions, I don't recommend doing it. You have to manually reset the permissions later back the original owners of the files. It may also need to reinstall transmission.

zyrill:
@publicarray seems to work beautifully now, thanks a lot. I noticed that conf/priviledge has "groupname" set to "sc-transmission" - shouldn't it be "sc-download"? Probably my mistake, sorry about that...

jinfeng1985:
> @publicarray seems to work beautifully now, thanks a lot. I noticed that conf/priviledge has "groupname" set to "sc-transmission" - shouldn't it be "sc-download"? Probably my mistake, sorry about that...

how to install the transmission?

zyrill:
Depends on what kind of model you want to run it, but in general, you need to check out the latest dev version and compile the package yourself, then manually install it.

Check the guide here: https://github.com/SynoCommunity/spksrc#docker
Then you want to switch to the dsm7 branch: `git checkout dsm7`
Then you want to follow this guide: https://github.com/SynoCommunity/spksrc/wiki/Developers-HOW-TO#build-a-package

[...]

Issue #4215, opened 2020-10-12
DSM 7 support framework design


ymartin59:
I propose to collaboratively edit this issue description (with comments for discussions about points) so that to prepare how framework has to evolve.

Objectives:
- if possible, package should run seamlessly on both DSM 6 and DSM 7
  - this is not possible, packages for DSM 6 and DSM 7 are not compatible
  - packages for DSM 7 need at least `os_min_ver="7.0-40000"`
- reduce as much as possible package maintenance: implement support in framework make files and generated/embedded scripts

Design proposals:
- :heavy_check_mark: At upgrade "var" content is restored as "home" in both DSM 6 and DSM 7
- Migrate to the new resource/permission file/specification which help implement common uses cases that otherwise required root Privileges (may not be required see [comment](#issuecomment-707620864), but is something we are working towards)
  - :heavy_check_mark: SPK_COMMANDS and SPK_LINKS have to generate resource file for links
  - :heavy_check_mark:  installer has to generate database resource file from wizard for concerned application
  - :x:  Use privilege file to add package to a group e.g. `sc-media`
  - :heavy_check_mark: Use privilege file to add create shared folder

To investigate
- Privileges: what should be done?
- :x: Verify that adding package to a group e.g. `sc-media` still works from the install script.

Minimal requirements for DSM7
- `os_min_ver=` must be present and the value at least `"7.0-40000"`
- A `conf/privilege` file must exist (and have  `"defaults": { "run-as": "package" }` [other values not validated yet])

publicarray:
I've played around a little in my own fork, just to see what breaks and I think I got the minimal requirements done so packages can be "installed" but probably won't work. And I converted binary packages to be linked to `/usr/local` using Synology's resource file.

th0ma7:
OPTIONAL: There are a few changes I'd like to complete on the framework before jumping on the dsm7 bang-wagon:
1. Review the DSM vs SRM management by using the new `TC_TYPE` variable now in place in the toolchain
2. Review toolchain**s** naming to be singular and use long naming for `mk/spksrc.tc*.mk` for `mk/spksrc.toolchain*.mk`
[...]

PR #4373, opened 2021-01-07
Remove busybox from transmission deps

zyrill:
_Motivation:_  busybox shouldn't be required by transmission anymore as it was only used in dsm6 to get rid of legacy accounts and groups.
_Linked issues:_  Follow-up to (https://github.com/SynoCommunity/spksrc/pull/4364)

### Checklist

### DIFF
--- a/spk/transmission/Makefile
+++ b/spk/transmission/Makefile
@@ -30,12 +30,7 @@ ADMIN_PORT = $(SERVICE_PORT)
 
 POST_STRIP_TARGET = transmission_extra_install
 
-BUSYBOX_CONFIG = usrmng
-ENV += BUSYBOX_CONFIG="$(BUSYBOX_CONFIG)"
-
-
 include ../../mk/spksrc.spk.mk
-
 
 .PHONY: transmission_extra_install
 transmission_extra_install:
--- a/spk/transmission/Makefile
+++ b/spk/transmission/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = transmission
 SPK_VERS = 3.00
-SPK_REV = 20
+SPK_REV = 19
 SPK_ICON = src/transmission.png
 
 DEPENDS = cross/$(SPK_NAME)
@@ -12,7 +12,7 @@ DESCRIPTION_SPN = Transmission es un cliente BitTorrent simple y rápido. Puedes
 RELOAD_UI = yes
 STARTABLE = yes
 DISPLAY_NAME = Transmission
-CHANGELOG = "Remove busybox dependency"
+CHANGELOG = "Update openssl to 1.1."
 
 HOMEPAGE = https://transmissionbt.com
 LICENSE  = GPLv2/GPLv3
--- a/spk/transmission/Makefile
+++ b/spk/transmission/Makefile
@@ -1,9 +1,9 @@
 SPK_NAME = transmission
 SPK_VERS = 3.00
-SPK_REV = 19
+SPK_REV = 20
 SPK_ICON = src/transmission.png
 
-DEPENDS = cross/busybox cross/$(SPK_NAME)
+DEPENDS = cross/$(SPK_NAME)
 
 MAINTAINER = SynoCommunity
 DESCRIPTION = Transmission is an easy and fast BitTorrent client. You can control it remotely with its web interface or dedicated applications.
@@ -12,7 +12,7 @@ DESCRIPTION_SPN = Transmission es un cliente BitTorrent simple y rápido. Puedes
 RELOAD_UI = yes
 STARTABLE = yes
 DISPLAY_NAME = Transmission
-CHANGELOG = "Update openssl to 1.1."
+CHANGELOG = "Remove busybox dependency"
 
 HOMEPAGE = https://transmissionbt.com
 LICENSE  = GPLv2/GPLv3
--- a/spk/transmission/src/service-setup.sh
+++ b/spk/transmission/src/service-setup.sh
@@ -5,8 +5,6 @@ PATH="${SYNOPKG_PKGDEST}/bin:${PYTHON_DIR}/bin:${PATH}"
 CFG_FILE="${SYNOPKG_PKGVAR}/settings.json"
 TRANSMISSION="${SYNOPKG_PKGDEST}/bin/transmission-daemon"
 
-GROUP="sc-transmission"
-
 SERVICE_COMMAND="${TRANSMISSION} -g ${SYNOPKG_PKGVAR} -x ${PID_FILE} -e ${LOG_FILE}"
 
 service_preinst ()
--- a/spk/transmission/src/service-setup.sh
+++ b/spk/transmission/src/service-setup.sh
@@ -5,6 +5,8 @@ PATH="${SYNOPKG_PKGDEST}/bin:${PYTHON_DIR}/bin:${PATH}"
 CFG_FILE="${SYNOPKG_PKGVAR}/settings.json"
 TRANSMISSION="${SYNOPKG_PKGDEST}/bin/transmission-daemon"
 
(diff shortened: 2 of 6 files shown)


spk/transmission/src/service-setup.sh
```diff
@@ -5,7 +5,7 @@ PATH="${SYNOPKG_PKGDEST}/bin:${PYTHON_DIR}/bin:${PATH}"
 CFG_FILE="${SYNOPKG_PKGVAR}/settings.json"
 TRANSMISSION="${SYNOPKG_PKGDEST}/bin/transmission-daemon"
 
-GROUP="sc-download"
+GROUP="sc-transmission"
```
publicarray, line 8: You can remove this line the default is `sc-{packagename}`

zyrill:
No idea why pyload builds should fail if I remove a dep of a different package. If somebody more knowledgeable could have a look at the pyload build errors and whether they are connected to this PR I'd be grateful. Thanks in advance!

spk/transmission/src/service-setup.sh
publicarray: Actually can you leave it as `sc-download` for backwards compatibility with DSM 6?

publicarray:
Hm, I haven't seen that before, weird.

zyrill:
Anything else I can do to improve the PR, @publicarray or do I leave it like this?

publicarray:
Thank you for your work on this @zyrill.

I'm just thinking that we should leave the SPK_REV and CHANGELOG alone until this branch is merged to reduce merge conflicts. Also if we remove busybox here it would make sense to remove it from other packages that no longer require it aswell. Just my 2 cents.

@hgy59 what do you think? Do you remember how the migration to DSM6 done?
Besides a review, I'm thinking the Transmission package could be one of the first to be published for DSM7.

zyrill:
> 
> 
> Thank you for your work on this @zyrill.
> 
> I'm just thinking that we should leave the SPK_REV and CHANGELOG alone until this branch is merged to reduce merge conflicts. Also if we remove busybox here it would make sense to remove it from other packages that no longer require it as well. Just my 2 cents.
> 
> @hgy59 what do you think? Do you remember how the migration to DSM6 was done?
> Besides a review, I'm thinking the Transmission package could be one of the first to be published for DSM7.

I'm just a newbie but removing busybox everywhere it isn't needed anymore sounds like a good idea.

hgy59:
> @hgy59 what do you think? Do you remember how the migration to DSM6 was done?
> Besides a review, I'm thinking the Transmission package could be one of the first to be published for DSM7.

@publicarray I can't remember the dsm6 migration, as I was not member of SynoCommunity then.
But I can see that some packages are still not updated for dsm 6 😠 

I propose to merge the dsm 7 branch back to master as soon as the first package is ready to publish for DSM7 - but not as long as stable packages fail to build for DSM < 7.0.

Preferably we shall not publish packages before the official DSM 7 is released. 
-- OR --
we mark all DSM 7 packages as BETA. (_But only the DSM 7 packages should become BETA - this might be managed generally in the INFO file creation code_)

zyrill:
Excuse the heretical question, but why are packages that fail to build considered "stable"? :D Failing packages seem rather _unstable_ to me... And if there is no maintainer for those, shouldn't they rather be moved to some kind of staging section?

publicarray:
@zyrill Thanks for your contribution!

 Regarding the build errors, I think this was recently fixed on the master branch. For this PR I'm just going to ignore them.
 
 @hgy59 Can I assume we don't need to code to migrate from DSM5->DSM6 anymore? Otherwise, we need to find a way include  busybox when we compile DSM6 packages only.

@zyrill 👆 this is the only reason I haven't merged this yet.

zyrill:
Thanks for the explanation and it was my pleasure to be able to help. Thanks for being so forthcoming, really helped me a lot making a contribution - I've had other experiences with FOSS projects in the past - IMHO you're doing an excellent job.

publicarray:
Thanks @zyrill, much appreciated! Although I would thank @Diaoul, @ymartin59, @hgy59, @th0ma7 and [others](https://github.com/SynoCommunity/spksrc/graphs/contributors) I think they are the real hereos here :1st_place_medal: 

@hgy59 sounds like a good plan :+1:

zyrill:
@publicarray @hgy59 is there a current overview of which packages are still failing to build in order for us to be able to determine what tasks still need to be completed in order to merge dsm7 and related commits to master? The openssl PR https://github.com/SynoCommunity/spksrc/pull/4369 showed a few failing builds but those should be mostly fixed if I understood your comments correctly. I'm willing to try to help out some more in order to get "closure" on this PR. ;)

spk/transmission/Makefile
```diff
@@ -3,7 +3,7 @@ SPK_VERS = 3.00
 SPK_REV = 19
 SPK_ICON = src/transmission.png
 
-DEPENDS = cross/busybox cross/$(SPK_NAME)
+DEPENDS = cross/$(SPK_NAME)
```
hgy59, line 6: When removing cross/busybox the lines below should be removed too.

``` makefile
BUSYBOX_CONFIG = usrmng
ENV += BUSYBOX_CONFIG="$(BUSYBOX_CONFIG)"
```

hgy59:
> @hgy59 Can I assume we don't need to code to migrate from DSM5->DSM6 anymore? Otherwise, we need to find a way include busybox when we compile DSM6 packages only.

Agreed. IMHO the support of DSM5 is for models only, that do not support an update to DSM6. So it is admissible to remove the the DSM5 -> DSM6 migration.

hgy59:
> is there a current overview of which packages are still failing to build in order for us to be able to determine what tasks still need to be completed in order to merge dsm7 and related commits to master? 

As this repository contains more than 100 packages it is not possible to have a stable state where all packages for all supported archs can be built without failures. It is a kind of agile project. Packages that are requested and contributors are insterested to spend time for, are created or updated. Others lay around for months and even years before an update or initial release is created.

The best is to focus on a single package to contribute. It is always painful to update more than one package in one Pull-Request. A successful build is only the first step. You should test a fresh installation and a package update before a PR is merged into master and the related package can be published. 
***What we are missing here are test instructions and use cases or even test scripts to validate a package works as expected.***

spk/transmission/Makefile
zyrill: Makes perfect sense.

zyrill:
> The best is to focus on a single package to contribute. It is always painful to update more than one package in one Pull-Request. A successful build is only the first step. You should test a fresh installation and a package update before a PR is merged into master and the related package can be published.
> _**What we are missing here are test instructions and use cases or even test scripts to validate a package works as expected.**_

Understood, working on just one package is obviously easier - the confusion arose from the comment above where it was stated that the code would be merged "but not as long as stable packages fail to build for DSM < 7.0" - I now understand that to be a more general comment though. Thanks for the clarification.

Regarding tests: I built the transmission package and tested it on both my DS1515+ and my DS216j in both fresh installs and upgrades. I understood that among others (pls see the issues in https://github.com/SynoCommunity/spksrc/issues/4313 and https://github.com/SynoCommunity/spksrc/issues/4215#issuecomment-753692984 and https://github.com/SynoCommunity/spksrc/issues/4362 ) @publicarray has also tested the package and it seems to be working fine.

publicarray:
@hgy59 We do have *some* instructions here: https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#testing

zyrill:
Thanks a lot for the merge, @publicarray - could you also take a look at the related PRs, please? Those would be:
- https://github.com/SynoCommunity/spksrc/pull/4370
- https://github.com/SynoCommunity/spksrc/pull/4369

PR #4375, opened 2021-01-08
cross/libjpeg: archive contents were fixed

smaarn:
_Motivation:_  Build of spks depending on cross/libjpeg would no longer be passing

Basically the archives have changed.

Manually checked the difference between the two archived contents and it's XML resources which have been changed (fixed, really) inside.
[jpegsrc.v9c.tar.gz (old one)](https://github.com/SynoCommunity/spksrc/files/5789596/jpegsrc.v9c.tar.gz) vs 
[jpegsrc.v9c.tar.gz (new one)](https://github.com/SynoCommunity/spksrc/files/5789599/jpegsrc.v9c.tar.gz)

To be fair, another option could have been to ask was this file changed. But I must confess I don't know any of the contributors nor any of the owners so...

### Checklist

### DIFF
--- a/cross/libjpeg/digests
+++ b/cross/libjpeg/digests
@@ -1,3 +1,3 @@
-jpegsrc.v9c.tar.gz SHA1 2ce111c8c0ac828a44b13ad28c265e954a342d07
-jpegsrc.v9c.tar.gz SHA256 650250979303a649e21f87b5ccd02672af1ea6954b911342ea491f351ceb7122
-jpegsrc.v9c.tar.gz MD5 93c62597eeef81a84d988bccbda1e990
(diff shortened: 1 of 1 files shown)


hgy59:
> To be fair, another option could have been to ask was this file changed. But I must confess I don't know any of the contributors nor any of the owners so...

The comparision of the content tells me that only some MS VisualStudio 2003 Project (?) Files have changed and the change is only the removement of 3 leading bytes in the utf-8 xml files [removed bytes are 0x8B 0xAF 0xA8]
So it absolutly has no side effect for our cross compilation...

hgy59 (approved):

PR #4380, opened 2021-01-09
glib: Update to version 2.66.4 and fix hi3535 build

th0ma7:
_Motivation:_  Build failure on `hi3535` block upgrading `glib` to latest version.  A patch was provided by `glib` developers (https://gitlab.gnome.org/GNOME/glib/-/issues/2274 - will not be maintained upstream) that allow fixing the issue.  We will have to maintain the patch internally within SynoCommunity.
_Linked issues:_  #4307

### Checklist


cross/glib/patches/hi3535/001-fix-build-SYS_sched_getattr.patch
```diff
@@ -0,0 +1,20 @@
+--- meson.build-old	2021-01-09 14:43:01.994870325 +0000
```
hgy59, line 1: I would like to include such comment in the patch file (_the `-old => .orig` change is not relevant, but I had to change the existing line to make a suggestion_)

```suggestion
# see https://gitlab.gnome.org/GNOME/glib/-/issues/2274
# as this will not be maintained upstream we will have to maintain this patch
--- meson.build.orig	2021-01-09 14:43:01.994870325 +0000
```

hgy59 (approved):

PR #4292, opened 2020-12-08
[WIP] Added new spk for activebackup-prometheus-exporter

fleaz:
_Motivation:_
I work at a company where we are responsible for the infrastructure in some office buildings and we use Disk- and Rackstations together with the backup tools from the ActiveBackup suite to backup virtual machines, physical hosts and Google Drives. Because this was a blank spot in our monitoring I started to build [activebackup-prometheus-exporter](https://github.com/codemonauts/activebackup-prometheus-exporter) which is an Prometheus exporter which presents the data scraped from the internal sqlite databases generated by the ActiveBackup tools to get informations about the backup jobs. So if you have a running Prometheus server, you can just add your Synology NAS as a target and scrape the data about the backup jobs after installing this tool and create alerts when e.g. the last backup is older than 24 hours or something like that.

I'm new to this whole Synocommunity thing and this is my first spk package I build. So if I did something wrong, please let my know and I will fix it :)

Due to the similarities I used the spk and cross paccakge  from the normal Prometheus node-exporter as a starting point and changed it to my needs. The packages was tested on my DS218+ (apollolake) and on a RS1619xs+ (broadwellnk) both running DSM 6.2.

The only thing that is currently not working is that it's not recognized as "running" after an installation, even though it's running (checked with `ps` and I can curl it). Maybe someone more experienced can give me a hint on how to fix this?

Best regards,
Felix 

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4383, opened 2021-01-12
native/go: Fix build issue when first invoke through cross-build

th0ma7:
_Motivation:_  `native/go` fails to build when first invoked from a cross-build (e.g. non-x86_64).  Forcing the build environment for amd64 along with setting the default `gcc` and `g++` to use the docker built-in compiler solves this.
_Linked issues:_   #4292

### Checklist


native/go/Makefile
```diff
@@ -23,4 +23,4 @@ include ../../mk/spksrc.native-cc.mk
 go_native_compile:
 	@$(MSG) "Building Go for host system"
 	# GO111MODULE disable go.mod per https://github.com/golang/go/issues/26996
```
hgy59, line 25: as GO111MODULE=off is dropped, this comment may be dropped too.

th0ma7: nope, it ain't removed but rather moved on that command line further to the left :)

th0ma7:
@hgy59 and @ymartin59 do you think this one is good for a merge?  I believe so but rather have a second opinion first.

ymartin59 (approved):

PR #4388, opened 2021-01-13
Mark znc as unsupported on hi3535

zyrill:
_Motivation:_   znc fails to build on hi3535 which is why I'm marking it unsupported for now (builds fine on other archs).
_Linked issues:_   See e.g. build log here https://github.com/SynoCommunity/spksrc/pull/4369/checks?check_run_id=1694292733

### DIFF
--- a/spk/znc/Makefile
+++ b/spk/znc/Makefile
@@ -20,8 +20,7 @@ HOMEPAGE = https://wiki.znc.in/
 LICENSE  = Apache License 2.0
 
 # Ensure C++11 compatibility
-UNSUPPORTED_ARCHS = $(ARMv5_ARCHS) $(OLD_PCC_ARCHS)
-
+UNSUPPORTED_ARCHS = $(ARMv5_ARCHS) $(OLD_PCC_ARCHS) $(ARMv7L_ARCHS)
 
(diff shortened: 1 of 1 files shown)


hgy59:
this might be obsolete by the fix with #4377

zyrill:
Is all that's missing with https://github.com/SynoCommunity/spksrc/pull/4377 the fresh install/upgrade testing? Then this I could close this PR here as it is indeed superfluous...

smaarn:
> Is all that's missing with https://github.com/SynoCommunity/spksrc/pull/4377 the fresh install/upgrade testing? Then this I could close this PR here as it is indeed superfluous...

To tell you the truth I must confess I haven't had the time yet to test the change *and* I don't have an hi3535 NAS on my side. Should be able to test over the week-end but I must confess I don't even know what this spk does...

zyrill:
znc is an IRC (ooooold chat protocol) bouncer. No idea if anybody still uses stuff like that anymore. Not me certainly and I also don't have a hi3535 device so I sadly can't help test either. If you have access over the weekend, that would probably be perfect. Go hang out on the Freenode service in #warez a bit - and say hello to the 80ies for me. ;)

PR #4389, opened 2021-01-15
fix line 8: [: : integer expression expected

publicarray:
-a executes both left and right hand expressions.
When $SYNOPKG_DSM_VERSION_MAJOR is empty it shows a warning since "" is not an integer and evaluates to false.

Test case (as root):

```
root@dsm6 ~# /bin/sh /var/packages/transmission/scripts/service-setup log
/var/packages/transmission/scripts/service-setup: line 7: [: : integer expression expected
# vs
root@dsm6 ~# synopkg log transmission
The log file is /volume1/@appstore/transmission/var/transmission_temp.log
```

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -75,6 +75,10 @@ SERVICE_FILES =
 $(DSM_SCRIPTS_DIR)/service-setup:
 	$(create_target_dir)
 	@echo "### Package specific variables and functions" > $@
+	@echo 'if [ -z "$${SYNOPKG_PKGNAME}" ] || [ -z "$${SYNOPKG_PKGDEST}" ] || [ -z "$${SYNOPKG_DSM_VERSION_MAJOR}" ]; then' >> $@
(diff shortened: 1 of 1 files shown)


ymartin59:
@publicarray From my point of view, such a fix is useless as many application scripts will fail "later". Command-line stop/start must be done thanks to `synopkg`. Refer to https://github.com/SynoCommunity/spksrc/wiki/Frequently-Asked-Questions#how-to-query-package-status-or-start-it-from-command-line

ymartin59:
In fact a "relevant" fix would be to prepend checks so that script fails with a proper error message mentioning `synopkg`

publicarray:
I've updated the script:

```sh
$ /var/packages/transmission/scripts/start-stop-status status
Error: Environment variables are not set.
Please run me using "synopkg start [packagename]" instead.

$ synopkg status transmission
transmission package is started
```

ymartin59 (approved):

ymartin59:
@publicarray Perfect! Thanks

PR #4395, opened 2021-01-17
Add DSM 7 Support

publicarray:
I think we are getting close enough that packages can be build and run, things left to do are documenting the new permissions/wizards, database creation from Synology resource worker, maybe more testing that there are no regressions for DSM 6 packages (I think I've found them all but one can't be too careful). Also, there is a Docker resource worker, but that will need individual packages to migrate to it.

Note: I frequently rebase with master to get the latest fixes and to avoid a large merge conflict. 
This means there will be force pushes!


hgy59:
@publicarray sorry, had to fixup the gh build action for noarch-7.0

publicarray:
@hgy59 no need to apologize, If you follow the git log I've made many silly mistakes. Also debugging GitHub actions is hard since you can't run them locally (There is https://github.com/nektos/act/ but It's a little harder to setup with GH events)

hgy59:
Now both the noarch-7.0 (dsm7) and noarch (all) are built...

toolchain/syno-aarch64-7.0/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+TC_ARCH = rtd1296 armada37xx aarch64
+TC_VERS = 7.0
+TC_KERNEL = 4.4.180
+TC_GCC = 7.3.0
+TC_GLIBC = 2.26
+
+TC_DIST = rtd1296-gcc730_glibc226_armv8-GPL
+TC_DIST_SITE_PATH = Realtek%20RTD129x%20Linux%204.4.180
+TC_DIST_SITE_URL = https://sourceforge.net/projects/dsgpl/files/Tool%20Chain/DSM%207.0%20preview%20Tool%20Chains/
```
th0ma7, line 11: An option here could be to change `spksrc.tc-flags.mk` such as:
```
ifeq ($(strip $(TC_DIST_SITE_URL)),)
ifeq ($(TC_VERS),7.0)
TC_DIST_SITE_URL = https://sourceforge.net/projects/dsgpl/files/Tool%20Chain/$(TC_TYPE)%20$(TC_VERS)%20preview%20Tool%20Chains/
endif
TC_DIST_SITE_URL = https://sourceforge.net/projects/dsgpl/files/Tool%20Chain/$(TC_TYPE)%20$(TC_VERS)%20Tool%20Chains/
endif
```
And therefore get rid of all the unnecessary duplicate line from each toolchain definitions.

.github/workflows/build.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       matrix:
         # x64=x86_64, evansport=i686, aarch64=armv8, armv7, hi3535=armv7l, 88f6281=armv5, qoriq=ppc
         # https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures
-        arch: [noarch, x64-6.1, evansport-6.1, aarch64-6.1, armv7-6.1, hi3535-6.1, 88f6281-6.1, qoriq-6.1]
+        arch: [noarch, noarch-7.0, x64-6.1, x64-7.0, evansport-6.1, aarch64-6.1, armv7-6.1, hi3535-6.1, 88f6281-6.1, qoriq-6.1]
```
th0ma7, line 28: From the wiki page shouldn't all the DSM7 arches be added?
```
arch-x64-7.0 arch-armv7-7.0 arch-aarch64-7.0 arch-evansport-7.0
```

cross/libaom/Makefile
```diff
@@ -51,3 +51,5 @@ endif
 ifeq ($(findstring $(ARCH),$(i686_ARCHS)),$(ARCH))
 CMAKE_ARGS += -DAOM_TARGET_CPU=x86
 endif
+
+ENV += CXXFLAGS=-D_GLIBCXX_USE_C99_MATH
```
th0ma7, line 55: That's a hell of a good catch!
Does this impact DSM<7.0 ?  Or should a DSM version test be done?
Or even better:
```
ifeq ($(TC_GCC),7.3.0)
ENV += CXXFLAGS=-D_GLIBCXX_USE_C99_MATH
endif
```

mk/spksrc.service.start-stop-status
```diff
@@ -35,11 +35,17 @@ start_daemon ()
             else
                 SVC_CD="cd ${SVC_CWD}; "
             fi
+            if [ -n "${SYNOPKG_DSM_VERSION_MAJOR}" ] && [ "$SYNOPKG_DSM_VERSION_MAJOR" -gt 7 ]; then
```
th0ma7, line 38: Shouldn't this be `-ge 7` ?

spk/nzbget-testing/src/service-setup.sh
```diff
@@ -1,9 +1,9 @@
 PATH="${SYNOPKG_PKGDEST}/bin:${PATH}"
 NZBGET="${SYNOPKG_PKGDEST}/bin/nzbget"
-CFG_FILE="${SYNOPKG_PKGDEST}/var/nzbget.conf"
+CFG_FILE="${SYNOPKG_PKGVAR}/nzbget.conf"
```
th0ma7, line 3: Reviewing this, I must say I really much like the `${SYNOPKG_PKGVAR}`.
But while at it, why not add a `${SYNOPKG_PKGBIN}` and `${SYNOPKG_PKGSHARE}`?

spk/transmission/src/wizard/upgrade_uifile_fre
```diff
@@ -1,9 +1,9 @@
...
     "step_title": "Permissions DSM",
     "items": [{
-        "desc": "Les permissions de toutes les applications de SynoCommunity liés au téléchargement sont gérées par le groupe <b>'sc-download'</b> dans DSM.<br>Le groupe 'users' n'est plus utilisé depuis DSM 6.<br>L'utilisateur spécifique à l'application n'apparaît plus dans la plupart des interfaces de configuration.<br>Merci de lire <a target=\"_blank\" href=\"https://github.com/SynoCommunity/spksrc/wiki/Permission-Management\">Permission Management</a> pour plus de détails."
+        "desc": "Merci de lire <a target=\"_blank\" href=\"https://github.com/SynoCommunity/spksrc/wiki/Permission-Management\">Permission Management</a> pour plus de détails."
     },
     {
-        "desc": "<strong style='color:red'>NOTE:</strong> La mise à jour de l'application va tenter de corriger les permissions des répertoires. En cas d'erreur de permission dans Transmission, donner les droits de Lecture/Ecriture au groupe 'sc-download' sur les répertoires mentionnés depuis File Station."
+        "desc": "En cas d'erreur de permission dans Transmission, donner les droits de Lecture/Ecriture sur les repertoires mentionnès depuis File Station."
```
th0ma7, line 7: `mentionnès` requiert un accent aigu tel que `mentionnés`

toolchain/syno-alpine-7.0/digests
```diff
@@ -0,0 +1,2 @@
+alpine4k-gcc730_glibc226_hard-GPL.txz SHA1 a6082db9cc84066ba93da32968fc604e8525f0e9
+alpine4k-gcc730_glibc226_hard-GPL.txz MD5 a259283312b3489085fb658dbf768f8f
```
th0ma7, line 2: Missing SHA256, no?

toolchain/syno-apollolake-7.0/digests
```diff
@@ -0,0 +1,2 @@
+apollolake-gcc730_glibc226_x86_64-GPL.txz SHA1 3d6a03823ef3465881354d1e68552a9546f9b6c8
+apollolake-gcc730_glibc226_x86_64-GPL.txz MD5 d3e7cb99168a2ed98d5a6c1d5f49d8fc
```
th0ma7, line 2: Missing SHA256, again, no?

toolchain/syno-armada370-7.0/digests
```diff
@@ -0,0 +1,2 @@
+armada370-gcc730_glibc226_hard-GPL.txz SHA1 8f0973be4c5f61ae919971cc9f1c1b9cfa76b96c
+armada370-gcc730_glibc226_hard-GPL.txz MD5 ca60fb63a0ed204f523398051dab171b
```
th0ma7, line 2: same, SHA256

toolchain/syno-armada375-7.0/digests
```diff
@@ -0,0 +1,2 @@
+aarmada375-gcc730_glibc226_hard-GPL.txz SHA1 1e3253f06bf24f303134e9030cce73ffdc2d50ec
+aarmada375-gcc730_glibc226_hard-GPL.txz MD5 ae1bf54e0b16955a43fd7b5991d35c14
```
th0ma7, line 2: same, SHA256

toolchain/syno-armadaxp-7.0/digests
```diff
@@ -0,0 +1,2 @@
+armadaxp-gcc730_glibc226_hard-GPL.txz SHA1 d0331bdbff1ca21e5d3b0580228c81dbc6892b80
+armadaxp-gcc730_glibc226_hard-GPL.txz MD5 8fe37c76c43e9777b55bf8d0c5c469b7
```
th0ma7, line 2: same, SHA256

toolchain/syno-avoton-7.0/digests
```diff
@@ -0,0 +1,2 @@
+avoton-gcc730_glibc226_x86_64-GPL.txz SHA1 6ada3b4d7065fca46c5aa3f64f4b4a17049e8056
+avoton-gcc730_glibc226_x86_64-GPL.txz MD5 5be42d301d16afdb2d17a85334724a10
```
th0ma7, line 2: same, SHA256

th0ma7 (changes requested):
This is looking quite good.
I noticed quite a few missing SHA256 in the toolchains.
Beyond that I'm much excited to give this a road test...

toolchain/syno-aarch64-7.0/Makefile
publicarray: That's a good point. I would just wait until Synology releases the toolchains for the released version. Until then I don't see an issue having them there. It's in beta after all

spk/nzbget-testing/src/service-setup.sh
publicarray: This is a newly set variable from DSM7 and points to a different folder:

```
SYNOPKG_PKGVAR: /var/packages/{package}/var
SYNOPKG_PKGDEST/var: /var/packages/{package}/target/var
```
On pre DSM7, `SYNOPKG_PKGVAR` gets set to `SYNOPKG_PKGDEST/var` to keep backwards compatibility and make migrating packages easier. Adding our own I would use a different prefix: `SC_PKG_PKGSHARE` what use cases do you have for those variables? Or what should they be set to?

In any case I think this is out of scope at the moment.

Why do any of it? On dsm7 `SYNOPKG_PKGDEST` gets overridden during package upgrades. So `SYNOPKG_PKGDEST/var` can't store settings anymore, see the wiki for the new folder structure.

mk/spksrc.service.start-stop-status
publicarray: Good catch. Now I wonder how any of this worked....

To make it consistent with the other ifs I reversed it:

```sh
if [ -n "${SYNOPKG_DSM_VERSION_MAJOR}" ] && [ "$SYNOPKG_DSM_VERSION_MAJOR" -lt 7 ]; then
    SU="su ${EFF_USER} -s" 
else
    SU=""
fi
```

cross/libaom/Makefile
publicarray: I don't see why it should be impacted negatively.

```
gcc 4.9.3 in DSM 6
The default, if no C++ language dialect options are given, is -std=gnu++98. 
https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Standards.html

gcc 7.3 in DSM 7
The default, if no C++ language dialect options are given, is -std=gnu++14. 
https://gcc.gnu.org/onlinedocs/gcc-7.3.0/gcc/Standards.html
```

Seems to work fine on DSM 6. But I haven't done any C stuff in ages.

.github/workflows/build.yml
publicarray: Yes but I think we should keep the list small, builds can take a long time and there is limited amount of free hours, it's a large amount but I don't remember how large.

toolchain/syno-aarch64-7.0/Makefile
th0ma7: Indeed but migration path afterwards become less cumbersome by changing a single file instead of +++

spk/nzbget-testing/src/service-setup.sh
th0ma7: While looking at the code `SYNOPKG_PKGDEST/bin` and `SYNOPKG_PKGDEST/share` came up quite often where you have modified the `var` ones.  I though that for consistency it may be worth while.  Out of scope, indeed, but best window of opportunity for such a change is probably now :)

mk/spksrc.spk.mk
```diff
@@ -14,8 +14,14 @@ TC = syno$(ARCH_SUFFIX)
...
 SPK_ARCH = noarch
 SPK_NAME_ARCH = noarch
+ifeq ($(shell expr "$(TCVERSION)" \>= 7.0),1)
+SPK_TCVERS = dsm7
+OS_MIN_VER = 7.0-4000
+else
 SPK_TCVERS = all
 OS_MIN_VER = 3.1-1594
+endif
+ARCH_SUFFIX = -$(SPK_TCVERS)
```
th0ma7, line 24: The ARCH_SUFFIX, I don't understand what it's for?

mk/spksrc.tc-vers.mk
```diff
@@ -57,3 +57,7 @@ endif
 ifeq ($(TC_VERS),6.2.3)
 TC_BUILD = 25423
 endif
+
+ifeq ($(TC_VERS),7.0)
+TC_BUILD = 4000
```
th0ma7, line 62: 4000 (four-thousand) is definitively wrong, must be over fourthly-thousand.  I saw a build number of `41222` yesterday.

toolkit/syno-evansport-7.0/digests
```diff
@@ -0,0 +1,2 @@
+ds.evansport-7.0.dev.txz SHA1 be47d9a3147d62bff0827b07d590f428483703a0
+ds.evansport-7.0.dev.txz MD5 9d1c7f8695f5b7abdb352919344d35ec
```
th0ma7, line 2: left over lacking SHA256

toolchain/syno-geminilake-7.0/digests
```diff
@@ -0,0 +1,2 @@
+geminilake-gcc730_glibc226_x86_64-GPL.txz SHA1 6872bddab01e63d07568be8ad5e96dee15aa897e
+geminilake-gcc730_glibc226_x86_64-GPL.txz MD5 710dd6b43262a9de8f5cfd5341b5a314
```
th0ma7, line 2: another left over still lacking SHA256

th0ma7 (changes requested):

mk/spksrc.spk.mk
hgy59: This is for the noarch packages i.e. all (and new dsm7)
<package>-noarch-all-*.spk
<package>-noarch-dsm7-*.spk

mk/spksrc.tc-vers.mk
hgy59: Yes it is documented that for DSM 7 the min. os version is 7.0-40000 (but it was no problem to install adminer with  7.0-4000).

toolchain/syno-geminilake-7.0/digests
hgy59: I will fix this.
Is is as easy as `make clean && make digests` in the syno-*7.0 toolchain folders...

mk/spksrc.spk.mk
hgy59: and ARCH_SUFFIX is used to define the WORK_DIR in spksrc.directories.mk
```
ifndef WORK_DIR
WORK_DIR = $(PWD)/work$(ARCH_SUFFIX)
endif
```

and it is mandatory to have different work dir for the two noarch builds.

hgy59:
@publicarray finally the adminer package for DSM7 is installable through the package center.

toolchain/syno-geminilake-7.0/digests
publicarray: Thanks @hgy59 :)

publicarray:
@hgy59 Awesome, Thank you!

mk/spksrc.spk.mk
```diff
@@ -14,8 +14,14 @@ TC = syno$(ARCH_SUFFIX)
 else
 SPK_ARCH = noarch
 SPK_NAME_ARCH = noarch
+ifeq ($(shell expr "$(TCVERSION)" \>= 7.0),1)
```
th0ma7, line 17: `expr` doesn't really support float.
Another method could be using the `sort -V` for versioning like I did for the kernel side of things:
```
# olddefconfig is not available <= 3.2
ifeq ($(shell printf '%s\n' "$(TC_KERNEL)" "3.3" | sort -V | head -1),$(TC_KERNEL))
        @$(MSG) "oldconfig OLD style... $(TC_KERNEL) <= 3.2"
        $(RUN) yes "" | $(MAKE) oldconfig
else
        $(RUN) $(MAKE) olddefconfig
endif
```

mk/spksrc.tc-vers.mk
th0ma7: Still, this needs to be readjusted as once we publish online it will require a proper build number.

th0ma7 (changes requested):

mk/spksrc.spk.mk
publicarray: Hm, It works as it should and is easy to read. If I look at your version I have to take longer to understand it and uses 3 tools.

I've read the man page and skimmed `info expr` I don't see anywhere where floats are unsupported... where 
 
If we need to change can I propose `bc` instead?

```sh
docker run -it --rm --name spksrc -v $HOME/git/spksrc:/spksrc synocommunity/spksrc
root@04c82828bdd1:/spksrc# expr "6.9" \>= 7.0
0
root@04c82828bdd1:/spksrc# expr "7.1" \>= 7.0
1

root@04c82828bdd1:/spksrc# echo "6.9 >= 7.0" | bc -l
0
root@04c82828bdd1:/spksrc# echo "7.1 >= 7.0" | bc -l
1

root@04c82828bdd1:/spksrc# expr "7.1234" \> 7.123
1
root@04c82828bdd1:/spksrc# expr "7.0234" \> 7.123
0
```

publicarray: Interesting a snipit from `info expr`
```
‘< <= = == != >= >’
     Compare the arguments and return 1 if the relation is true, 0
     otherwise.  ‘==’ is a synonym for ‘=’.  ‘expr’ first tries to
     convert both arguments to integers and do a numeric comparison; if
     either conversion fails, it does a lexicographic comparison using
     the character collating sequence specified by the ‘LC_COLLATE’
     locale.
```

publicarray: So where would a lexicographic comparison fail as apposed to a float? If anything it eliminates float rounding errors.

publicarray: Ok I found a rare case:

$ expr 7 \>= 7
1

$ expr 7  \>= 7.0
0

th0ma7: I hit this rather interesting issue when comparing kernel version 3.10 with 3.2 ...

publicarray: Ah ok the problem isn't with floats (mathematically 3.10 is less than 3.2, float or no float will help). Maybe we can wrap your code into a function? Will that work in make files? e.g. `version_lt 3.10 3.2`?  https://stackoverflow.com/questions/16989598/comparing-php-version-numbers-using-bash/24067243

th0ma7: Perhaps something as simple as (totally untested):
```
version_comparison:
ifeq ($(shell printf '%s\n' "$(VERSION)" "$(VERSION_TARGET)" | sort -V | head -1),$(VERSION_TARGET))
        echo 1
endif

...

VERSION=a.b
VERSION_TARGET=b.c

ifeq ($(strip $(version_comparison)),1)
        @$(MSG) "$(VERSION) is >= to $(VERSION_TARGET)"
else
        @$(MSG) "$(VERSION) is <= to $(VERSION_TARGET)"
fi
```

hgy59:
@publicarray I have added the python script to create the resource file (as [discussed](https://github.com/SynoCommunity/spksrc/issues/4400#issuecomment-766680894))

The integration in `spksrc.service.mk` would look like:
```
$(DSM_CONF_DIR)/resource:
	$(create_target_dir)
	@echo '{}' > $@
ifneq ($(strip $(SPK_COMMANDS) $(SERVICE_WIZARD_SHARE)),)
	@../../mk/get_resources.py SPK_COMMANDS "$(SPK_COMMANDS)" SERVICE_WIZARD_SHARE "$(SERVICE_WIZARD_SHARE)"  >$@
endif
SERVICE_FILES += $(DSM_CONF_DIR)/resource
# STARTABLE needs to be yes, the resource linking and unlinking works on start and stop
# see spsrc.spk.mk
```

mk/spksrc.spk.mk
publicarray: With the help of https://stackoverflow.com/a/60021707
```MAKE
# version_lt = $(shell if printf '%s\n%s\n' "$(1)" "$(2)" | sort --version-sort --check=quiet ; then echo 1; fi)
# version_gt = $(shell if printf '%s\n%s\n' "$(2)" "$(1)" | sort --version-sort --check=quiet ; then echo 1; fi)
version_le = $(shell if printf '%s\n' "$(1)" "$(2)" | sort -V -C ; then echo 1; fi)
version_ge = $(shell if printf '%s\n' "$(2)" "$(1)" | sort -V -C ; then echo 1; fi)

$(info *********)
VERSION=7.0
ifeq ($(call version_le, ${VERSION}, 6.0),1)
$(info "$(VERSION) is <= 6.0")
else
$(info "$(VERSION) is > 6.0")
endif
$(info *********)

ifeq ($(call version_ge, ${VERSION}, 6.0),1)
$(info "$(VERSION) is >= 6.0")
else
$(info "$(VERSION) is < 6.0")
endif
$(info *********)
```

Results
```
VERSION=3.10
*********
"3.10 is > 3.2"
*********
"3.10 is >= 3.2"
*********


VERSION=4.0
*********
"4.0 is <= 6.0"
*********
"4.0 is < 6.0"
*********

VERSION=6.0
*********
"6.0 is <= 6.0"
*********
"6.0 is >= 6.0"
*********
```

publicarray: Hm, which of these easier to read?

```make
ifneq ($(call version_ge, ${TCVERSION}, 7.0),1)
# old code
endif
ifeq ($(call version_le, ${TCVERSION}, 6.2.3),1)
# old code
endif
```

hgy59: maybe we need an additional function `version_lt` (less than) to implement like this
``` makefile
ifeq ($(call version_lt, ${TCVERSION}, 7.0),1)
# old code
endif
```

This will evaluate as expected, even when version 6.2.4 or other would be released.
For this reason I would prefer the first solution of yours.
_And it is better than the use of any magic version like 6.99999)_

publicarray: @hgy59 Thanks that was my concern as well, that is why I've already used first version. I just noticed there is a lot of "if not equal to" in the Makefiles. Not a big issue but I have been caught out on it on occasion, It's easy to miss the little `n`. 
From your feedback I've added the `version_lt` function as well.

spk/nzbget-testing/src/service-setup.sh
publicarray: Thanks, I understand, but  since we agree it's out of scope I'll leave it for now. Personally I think keeping `SYNOPKG_PKGDEST/bin` makes the intent clearer than hiding it behind another variable. I actually hope some packages can make use of the new variables set by synology. e.g : `SYNOPKG_PKGHOME`, `SYNOPKG_PKGDEST_VOL` `SYNOPKG_TEMP_UPGRADE_FOLDER`

th0ma7:
@publicarray how far you think we are to getting this merge?
I haven't been involved much myself on this but let me know how I can help.

publicarray:
I have been waiting for a reply from Synology regarding the creation of databases. It's been a few days... but after that I think we can merge this. There is always a need to build more packages and find more broken ones. Maybe you can find out why chromaprint fails to build?

th0ma7:
@publicarray sure, I can definitively have a look at it.
Can you rebase as I just merged to master the latest ffmpeg changes which will have an impact on chromaprint (which built all-right on dsm6).

publicarray:
Sure. It's done.

th0ma7:
@publicarray I have found hints as of why it fails on `chromaprint`.  I'll probably be able to come up with a fix (fingers-crossed).
Would you have a good example I can follow in order to remove the `chown`/`chmod`/`set_unix_permissions`/`set_syno_permissions` from tvheadend?

hgy59:
@publicarray, @th0ma7 please let us avoid package specific changes on this dsm7 branch.
Otherwise it will be very hard or maybe impossible to merge this branch back to master without breaking the DSM 6 builds.

Let us start with the implementation of `${SYNOPKG_PKGVAR}` on the master.

Other topics to be moved to the master branch:
- remove cross/busybox for user migration
- create even dsm7 specific service-setup code on the master branch, if independent of spksrc.service.installer.dsm7
- avoid package updates on the dsm7 branch only (i.e. wallabag), implement this on the master first.

IMHO the additional effort to integrate as much as possible on the master and then rebase the dsm7 branch is worth it.

th0ma7 (approved):
Looks good to me, thnx!

th0ma7:
@hgy59 to be honest (and I may change my mind) I have no intention to release another `ffmpeg` or any of its dependency to DSM6 anymore.  My plan is to focus solely on DSM7 once the framework is ready.  Reasoning is that `ffmpeg` is much too complex to be maintained in multi-streams.

Indeed https://github.com/SynoCommunity/spksrc/pull/4395/commits/71c0ccb0ffe4f80b4ac53a77fb8ead5e2f82dbe9 can be merged upstream instead of here.

But thnx, I'll keep good note of that and limit the changes when ever I can.

.github/ISSUE_TEMPLATE.md
```diff
@@ -21,6 +21,8 @@ _3._
 
 #### Package log
 _Check Package Center or `/usr/local/{package}/var/`_
+
+_DSM7: `/var/log/packages/{package}.log`_
 ```
```
hgy59, line 26: I have integrated this in the official issue templates under in the community profile under `Insights - Community - issue templates`.
The ISSUE_TEMPLATE.md is now obsolete (on master branch).

I propose to incorporate any other DSM7 related changes of the templates on the master branch.

th0ma7:
@publicarray after spending hours at trying to figure out things, no luck.  I can't understand why the behavior is different in GCC 7.3 from 4.9 related to the `error: ‘avcodec_alloc_frame’ was not declared in this scope`.

@ymartin59 as you worked on `chromaprint` previously, perhaps you have an idea on this on?

Error:
```
In file included from /home/spksrc/spksrc-dsm7/spksrc/spk/chromaprint/work-apollolake-7.0/chromaprint-v1.5.0/src/cmd/fpcalc.cpp:7:0:
/home/spksrc/spksrc-dsm7/spksrc/spk/chromaprint/work-apollolake-7.0/chromaprint-v1.5.0/src/audio/ffmpeg_audio_reader.h:31:24: error: ‘avcodec_alloc_frame’ was not declared in this scope
 #define av_frame_alloc avcodec_alloc_frame
                        ^
/home/spksrc/spksrc-dsm7/spksrc/spk/chromaprint/work-apollolake-7.0/chromaprint-v1.5.0/src/audio/ffmpeg_audio_reader.h:194:12: note: in expansion of macro ‘av_frame_alloc’
  m_frame = av_frame_alloc();
            ^~~~~~~~~~~~~~
/home/spksrc/spksrc-dsm7/spksrc/spk/chromaprint/work-apollolake-7.0/chromaprint-v1.5.0/src/audio/ffmpeg_audio_reader.h:31:24: note: suggested alternative: ‘avcodec_send_frame’
 #define av_frame_alloc avcodec_alloc_frame
                        ^
/home/spksrc/spksrc-dsm7/spksrc/spk/chromaprint/work-apollolake-7.0/chromaprint-v1.5.0/src/audio/ffmpeg_audio_reader.h:194:12: note: in expansion of macro ‘av_frame_alloc’
  m_frame = av_frame_alloc();
            ^~~~~~~~~~~~~~
/home/spksrc/spksrc-dsm7/spksrc/spk/chromaprint/work-apollolake-7.0/chromaprint-v1.5.0/src/audio/ffmpeg_audio_reader.h: In member function ‘void chromaprint::FFmpegAudioReader::Close()’:
/home/spksrc/spksrc-dsm7/spksrc/spk/chromaprint/work-apollolake-7.0/chromaprint-v1.5.0/src/audio/ffmpeg_audio_reader.h:35:23: error: ‘avcodec_free_frame’ was not declared in this scope
 #define av_frame_free avcodec_free_frame
                       ^
```

publicarray:
Thanks @th0ma7 for giving it your best shot. 

Maybe it's time to open a dialog with them: https://github.com/acoustid/chromaprint/issues?

.github/ISSUE_TEMPLATE.md
publicarray: Thanks I've removed the commit

publicarray:
@hgy59 

> please let us avoid package specific changes on this dsm7 branch.

Agreed, however transmission/wallabag/adminer are good test cases and I've made every effort to maintain backwards compatibility. transmission (shared folder worker), wallabag (webservice and DB worker), adminer (webservice and php worker)

> Otherwise it will be very hard or maybe impossible to merge this branch back to master without breaking the DSM 6 builds.

I think this branch can be merged as is. From builds that I've tested on this branch they work on both DSM6 and DSM7

> Let us start with the implementation of ${SYNOPKG_PKGVAR} on the master.

I think it's a waste of time maybe we can have 2 Pull requests, package specific stuff and framework stuff.

> Other topics to be moved to the master branch:

>    remove cross/busybox for user migration

I think this can be independently done when the package specific install/service scripts need to be modified (next point)

>    create even dsm7 specific service-setup code on the master branch, if independent of spksrc.service.installer.dsm7

This will be a slow process since it's different for every package.

>   avoid package updates on the dsm7 branch only (i.e. wallabag), implement this on the master first.

Hard to do since It relies on code on this branch e.g. uninstaller But I propose to separate package specific stuff to a separate PR but keep this one and merge the framework stuff ASAP.

> IMHO the additional effort to integrate as much as possible on the master and then rebase the dsm7 branch is worth it.

Sorry but, I humbly disagree. I think the opposite is better, upgrade the framework first with this PR once the package specific stuff is moved out. And then add the packages. My reasoning for not doing this your way is that it makes development for DSM7 much harder.

hgy59:
> > avoid package updates on the dsm7 branch only (i.e. wallabag), implement this on the master first.
> 
> Hard to do since It relies on code on this branch e.g. uninstaller But I propose to separate package specific stuff to a separate PR but keep this one and merge the framework stuff ASAP.
> 
> > IMHO the additional effort to integrate as much as possible on the master and then rebase the dsm7 branch is worth it.
> 
> Sorry but, I humbly disagree. I think the opposite is better, upgrade the framework first with this PR once the package specific stuff is moved out. And then add the packages. My reasoning for not doing this your way is that it makes development for DSM7 much harder.

Indeed, the separation of the package specific stuff is all I wanted.
I was afraid that by merging this branch as is, we would trigger the github build action to build too many packages to complete within the limit of 5h. This would make it difficult to identify the packages that do not compile anymore.
But when we merge the framework specific stuff first, we get the necessary control.

I am just wondering, how you separate the package specific stuff to a separate branch without much effort...

publicarray:
Thanks @hgy59 

It's done: 
* [dsm7](https://github.com/SynoCommunity/spksrc/tree/dsm7)-framework
* [dsm7-packages](https://github.com/SynoCommunity/spksrc/tree/dsm7-packages) 
* [dsm7-combined](https://github.com/publicarray/spksrc/tree/dsm7-combined) (probably won't be kept in sync, but good for quick building)

> I am just wondering, how you separate the package specific stuff to a separate branch without much effort...

I used a GUI tool Sublime Merge. It makes it easy to look at individual commits and the files that changed, then cherry-pick and rebase...

mk/spksrc.spk.mk
```diff
@@ -275,14 +302,17 @@ info-checksum:
...
 	@mkdir -p $(DSM_WIZARDS_DIR)
+ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
+	@find $(SPKSRC_MK)wizard -maxdepth 1 -type f -and \( -name "uninstall_uifile" -or -name "uninstall_uifile_???" \) -print -exec cp -f {} $(DSM_WIZARDS_DIR) \;
+endif
+ifneq ($(strip $(WIZARDS_DIR)),)
 	@find $${SPKSRC_WIZARDS_DIR} -maxdepth 1 -type f -and \( -name "install_uifile" -or -name "install_uifile_???" -or -name "install_uifile.sh" -or -name "install_uifile_???.sh" -or -name "upgrade_uifile" -or -name "upgrade_uifile_???" -or -name "upgrade_uifile.sh" -or -name "upgrade_uifile_???.sh" -or -name "uninstall_uifile" -or -name "uninstall_uifile_???" -or -name "uninstall_uifile.sh" -or -name "uninstall_uifile_???.sh" \) -print -exec cp -f {} $(DSM_WIZARDS_DIR) \;
+endif
 	@find $(DSM_WIZARDS_DIR) -maxdepth 1 -type f -not -name "*.sh" -print -exec chmod 0644 {} \;
 	@find $(DSM_WIZARDS_DIR) -maxdepth 1 -type f -name "*.sh" -print -exec chmod 0755 {} \;
 	$(eval SPK_CONTENT += WIZARD_UIFILES)
```
hgy59, line 315: we should avoid the creation of $(DSM_WIZARDS_DIR) for DSM < 7 without wizards.
```suggestion
ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
	@$(MSG) "Create default DSM7 uninstall wizard"
	@mkdir -p $(DSM_WIZARDS_DIR)
	@find $(SPKSRC_MK)wizard -maxdepth 1 -type f -and \( -name "uninstall_uifile" -or -name "uninstall_uifile_???" \) -print -exec cp -f {} $(DSM_WIZARDS_DIR) \;
endif
ifneq ($(strip $(WIZARDS_DIR)),)
	@$(MSG) "Create DSM Wizards"
	@mkdir -p $(DSM_WIZARDS_DIR)
	@find $${SPKSRC_WIZARDS_DIR} -maxdepth 1 -type f -and \( -name "install_uifile" -or -name "install_uifile_???" -or -name "install_uifile.sh" -or -name "install_uifile_???.sh" -or -name "upgrade_uifile" -or -name "upgrade_uifile_???" -or -name "upgrade_uifile.sh" -or -name "upgrade_uifile_???.sh" -or -name "uninstall_uifile" -or -name "uninstall_uifile_???" -or -name "uninstall_uifile.sh" -or -name "uninstall_uifile_???.sh" \) -print -exec cp -f {} $(DSM_WIZARDS_DIR) \;
endif
	@find $(DSM_WIZARDS_DIR) -maxdepth 1 -type f -not -name "*.sh" -print -exec chmod 0644 {} \;
	@find $(DSM_WIZARDS_DIR) -maxdepth 1 -type f -name "*.sh" -print -exec chmod 0755 {} \;
	$(eval SPK_CONTENT += WIZARD_UIFILES)
```

hgy59 (commented):
just two minor suggestions...

publicarray:
@hgy59 do you think this is ready to be merged?

hgy59 (approved):
LGTM

publicarray:
Ok I'm hitting the big green button...

th0ma7:
woot!  Awesome work, big big kudos for this!

Casuallynoted:
Not super familiar with a lot of this stuff, does this mean that DSM 7 packages will begin going live on the repo?

hgy59:
Only the base (i.e. framework) part is merged yet. All the work done for packages to support dsm7 is detached on the [dsm7-packages](https://github.com/SynoCommunity/spksrc/tree/dsm7-packages) branch.

- AFAIC see most of the CLI packages are close to be ready for dsm7 (synocli*, ruby, python*, git, gnupg, ...)
- Packages that cannot run without root permissions are almost impossible to support dsm7 (ntopng, ...) other might be redesigned to run without root.
- Some packages depend on mysql db and must be redesigned to use mariadb-10, as mysql/mariadb-5 is not available on dsm7 (fengoffice, cops, horde, mantisbt, owncloud, roundcube, selfoss, ttrss)

One breaking feature is the logging of package installation.
All packages that use `>> ${INST_LOG}` for logging fail as the log-file variable is not defined (and cannot be solved by defining the variable for the DSM7 log file).

PR #4144, opened 2020-08-22
Linuxtv out-of-tree kernel modules

th0ma7:
_Motivation:_  Long-term goal of providing media_tree kernel modules for various DVB dongles.
_Linked issues:_  Redo of previous PR #4080

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

### IMPORTANT
1. <s>This PR also include changes in `spksrc.checksum.mk` code in order to allow multiple package target checksums.  See change at commit https://github.com/SynoCommunity/spksrc/pull/4144/commits/560e379d9eb7134a14af0e72e72013683372958f  This is being used to manage different git hash downloads based on kernel compatibility for LinuxTV snapshots.</s>  **TODO:** Remove this patch and rebase on master following merge of #4566
2. This PR also produces two different package versions: one for 3.10 kernels using a specific git hash, and the other for 4.4 kernel using latest git hash available.

### Compatible kernels and DSM versions
- [x] 4.4.59+ (DSM-6.2.3)
- [x] 4.4.15+ (DSM-6.1)
- [x] 3.10.105 (DSM-6.2.3)

### Incompatible kernels (<3.10)
- [ ] 3.6.11 (SRM-1.2) - `northstarplus`
- [ ] 3.4.35 (DSM-6.2.3) - armv7l -  `hi3535`
- [ ] 3.2.40 (DSM-6.2.3) - `armada370`, `armada375`, `armadaxp`, `comcerto2k`, `evansport`
- [ ] 2.6.32 (DSM-6.2.3) - `88f6281`, `qoriq`

### Unsuccessful builds
- [ ] `alpine`  (DSM-6.2.3 - 3.10.105 - armv7)
- [ ] `armada38x` (DSM-6.2.3 - 3.10.105 - armv7)
- [ ] `monaco` (DSM-6.2.3 - 3.10.105 - armv7)
- [ ] `armada37xx` (DSM-6.2.3 - 4.4.59 - armv8)
- [ ] `rtd1296` (DSM-6.2.3 - 4.4.59 - armv8)

### Successful builds
```
make -j2 arch-dakota-1.2 arch-ipq806x-1.2
```
- [x] `dakota` (SRM-1.2 - 4.4.60)
- [x] `ipq806x` (SRM-1.2 - 4.4.60)
---
```
[...]

PR #3787, opened 2019-10-14
Add usbserial kernel modules

fredrike:
_Motivation:_  Synology DSM is lacking usb-serial kernel support, this fixes that.

### Checklist
- [x] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4399, opened 2021-01-20
Add kernel sources for DSM 6.2.3

th0ma7:
_Motivation:_  Been waiting for kernel sources for months for DSM-6.2.3 in order to get functional modules to run.
_Linked issues:_  #4144

### Checklist

### Changes
- Now only does a `modules_prepare` instead of building all the modules by default
- Only add `+` to EXTRAVERSION if kernel is > 4.x


kernel/syno-armada375-6.2.3/Makefile
```diff
@@ -0,0 +1,22 @@
...
+PKG_DIST_NAME = linux-3.x.$(PKG_EXT)
+PKG_DIST_SITE = https://downloads.sourceforge.net/project/dsgpl/Synology%20NAS%20GPL%20Source/$(PKG_BRANCH)branch/$(PKG_ARCH)-source
+PKG_DIST_FILE = $(PKG_NAME)-$(PKG_DIST_NAME)
+PKG_EXTRACT = linux-3.x
+SYNO_CONFIG = synoconfigs/$(PKG_ARCH)
+BASE_ARCH = arm
+
+HOMEPAGE = http://www.synology.com/
+
+PRE_COMPILE_TARGET = myPreCompile
```
ymartin59, line 16: I propose a more explicit target name like for instance `kernel_armada375_platv2`

ymartin59 (approved):
Just a minor improvement proposal.

kernel/syno-armada375-6.2.3/Makefile
th0ma7: I'm not sure I follow?
BTW, kernel is the only target I haven't "simplified" the Makefiles yet (vs `toolchain` and `toolkit`)
On my todo list for a later time.

hgy59: @th0ma7 please do this little housekeeping now, the `myPreCompile` naming is an influenza that still has its traces every where.
my proposal is to name the target `kernel_armada375_pre_compile`

th0ma7:
@hgy59 and @ymartin59 I decided to go all-in and make all the in-dept changes to the kernel building process as there where way too much churn in the kernel Makefiles ...

Testing still under going to confirm it builds successfully on all 5.x+ DSM.

th0ma7:
@ymartin59 and @hgy59 May I suggest we remove entirely all kernels < DSM-6.1.
File size are just ridiculous, 1.5-1.6gb where a fraction actually is the kernel sources.
And honestly, I doubt they will ever be used again...

th0ma7:
Honestly this was brutal... but it ends-up now working quite well.
There is only the obscure case of DSM-4.2 for `pcc824` arch which fails to engage on `make oldconfig`.
Beyond that, I tested every single one of them (and have a feeling of having downloaded the entire sourceforge.net)

Now is pretty much the time to adjust variable names or any other change you see fit.
Otherwise, I think its good for a merge.

ymartin59 (approved):
Great job and big effort, really much more maintainable code !

ymartin59:
@th0ma7 I am interested in building some USB modules for my model, how should I process?
I searched in wiki but with no luck.
Notice there are old references to "REQ_KERNEL" in https://github.com/SynoCommunity/spksrc/wiki/Compile-and-build-rules
May you please write some detail about kernel module compilation?

th0ma7:
@ymartin59 I'll look into adjusting the wiki pages to reflect some of the change.
Honestly this PR was overdue for kernel dev to actually work.
There are two approach that can be taken: outside from kernel tree or in-tree.
My PR #4144 is using a mechanism that is out-of-tree which requires the kernel to be prepared first than new code is built against it.

Personally I'm more used to the in-tree by simply adapting the configuration to enable the needed device then building the needed pieces at the directory level or down to the driver specific level such as:
```
$ make modules_prepare
$ make modules M=drivers/i2c
$ make modules M=drivers/media/i2c
$ make modules M=drivers/media/dvb-frontends -j4
$ make modules M=drivers/media/tuners -j4
$ make modules M=drivers/media/usb/em28xx
$ make modules M=drivers/media -j4
```
After this PR we should be in good shape to enable this functionality through a wrapper of some sort... more work in perspective.

th0ma7:
BTW, I guess the configuration adjustment helper used `busybox` could in a form or another be reused for kernel configs... just a thought.

th0ma7:
@ymartin59 forgot to mention, as the kernels are now built within the package directory tree (e.g. `spksrc/spk/<package>/work*`) this allows to play with the default kernel configuration without affecting the default kernel tree (`spksrc/kernel/syno*`).  That should give much more flexibility to compile specific modules for various DSM kernels & versions.

ymartin59:
@th0ma7 Ok, my interest is to achieve the idea of simplifying PR #3787 to deliver first a single package with all USB serial drivers, and eventually a custom do-it-yourself one driver package (for advanced users)

PR #4402, opened 2021-01-20
Adminer: adjust dependencies

hgy59:
- incorporate changes to reflect the latest published release
- set current spk dependencies
- remove obsolete backslashes from description


publicarray:
Why is this in master? The package won't work properly with Apache2.4 it's still on Apache2.2 see
5c6a4a562aaa34cb9c8a2cb933dd5d5e90f5e1b7

Some commands in the htaccess file changed.

Maybe I should have made that change in a separate PR...

PR #4413, opened 2021-01-29
python38: initial version

### LINKED ISSUE #4284: Upgrade request for Python 3.7 to 3.8

### Setup
_Package Name:_ Python3
_Package Version:_ Python3.7

_NAS Model: all
_NAS Architecture: all
_DSM version:_>6.0

Python 3.7 is the current release of the SynoCommunity, but this is already end of life as of june 2020.
Python 3.9 is already released and the current mainstream is Python 3.8.

Would it be possible to upgrade the python3 package to at least python 3.8 or maybe 3.9?

smaarn:
_Motivation:_  New python 3.8 package containing python version 3.8 (since it's now the oldest "stable" version)
_Linked issues:_  closes #4284 

### Checklist


spk/python38/Makefile
```diff
@@ -0,0 +1,49 @@
+SPK_NAME = python38
+SPK_SHORT_VERS = 3.8
+SPK_VERS = $(SPK_SHORT_VERS).6
+SPK_REV = 1
+SPK_ICON = src/python3.png
+
+DEPENDS  = cross/busybox cross/$(SPK_NAME)
```
smaarn, line 7: Maybe that should be the right time to remove the busybox binary from this installation since it's kinda incompatible with DSM 7.0 securization stream, should it ?

ymartin59: I agree. It was mainly delivered here to provide these generic service/tools to application packages (and reuse instead of rebuild)

ymartin59 (approved):

cross/python38/Makefile
```diff
@@ -0,0 +1,109 @@
+PKG_NAME = python38
+PKG_VERS = 3.8.6
```
hgy59, line 2: as of December 21 2020 the release 3.8.7 is officially available, so I propose to update to the current 3.8 version

hgy59:
🥇 for the dedicated python38 package name. This allows rolling updates of python 3.7 dependent packages.
The same does synology with different php 7 versions.

spk/python38/Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: python3_extra_install
+python3_extra_install:
+	rm -f $(STAGING_DIR)/$(PYTHON_LIB_DIR)/config/libpython*.a
+	rm -fr $(STAGING_DIR)/$(PYTHON_LIB_DIR)/test/
+	find $(STAGING_DIR)/$(PYTHON_LIB_DIR) -type f -regex '.*\.py[co]' | xargs rm -f
+	mkdir -p $(STAGING_DIR)/usr/bin
+	ln -sf /usr/local/$(SPK_NAME)/bin/python3 $(STAGING_DIR)/usr/bin/python3
```
smaarn, line 45: I actually wonder if this command shouldn't be avoided (same goes for the one below).

cross/python38/Makefile
smaarn: Done.

cross/python38/Makefile
```diff
@@ -0,0 +1,109 @@
...
+GNU_CONFIGURE = 1
+ADDITIONAL_CFLAGS = -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -L $(STAGING_INSTALL_PREFIX)/lib -I $(STAGING_INSTALL_PREFIX)/include
+CONFIGURE_ARGS  = --enable-shared --enable-ipv6 --without-ensurepip --enable-loadable-sqlite-extensions
+CONFIGURE_ARGS += ac_cv_buggy_getaddrinfo=no ac_cv_file__dev_ptmx=no ac_cv_file__dev_ptc=no ac_cv_have_long_long_format=yes
+
+POST_PATCH_TARGET = python3_post_patch
+PRE_CONFIGURE_TARGET = python3_pre_configure
+COMPILE_TARGET = python3_custom_compile
+INSTALL_TARGET = python3_custom_install
+POST_INSTALL_TARGET = python3_post_install
```
hgy59, line 29: probably we should prefix all targets with `python38_`

spk/python38/Makefile
```diff
@@ -0,0 +1,46 @@
...
+WHEELS = src/requirements.txt
+
+MAINTAINER = Allan Clark
+DESCRIPTION = Python Programming Language.
+DESCRIPTION_FRE = Langage de programmation Python.
+DESCRIPTION_SPN = Lenguaje de programación Python.
+RELOAD_UI = yes
+STARTABLE = no
+DISPLAY_NAME = Python 3.8
+CHANGELOG = "Initial Python 3.8.7 release"
```
hgy59, line 24: as the package is python38:
```suggestion
CHANGELOG = "Initial Python 3.8 release"
```

hgy59:
I am awainting for this, as I have a homeassistent update ready that depends on Python 3.8.

spk/python38/Makefile
smaarn: (Just realized you had proposed a change... I did it manually instead :/ )

hgy59 (approved):
LGTM

Safihre:
I am confused, why a seperate package for 3.8?! I know some packages required <3.6 but why would a package run on 3.7 but not on 3.8?
Now when I update SABnzbd I would have to force users to download an additional python? Probably messing things up with the "original" python3?

smaarn:
> I am confused, why a seperate package for 3.8?! I know some packages required <3.6 but why would a package run on 3.7 but not on 3.8?
> 
> Now when I update SABnzbd I would have to force users to download an additional python? Probably messing things up with the "original" python3? 

There may be issues with other spks depending on python: wheels are not compiled for "any" python version. We could upgrade the current python version but then it would be in "big bang" mode, meaning that users would need to update all depending spks.
To be honest I don't have any strong opinion but at least, this way, we get to have spks tested progressively and by interested parties.

Safihre:
Alright, I understand the big bang problem indeed. That's a good reason to split them up.
When just need to keep track that Python packages get updated and not just left on the old Python. 
But how do you avoid the dual Python problem? So each package uses the correct one? 
I know for SAB we just [use] (https://github.com/SynoCommunity/spksrc/blob/master/spk/sabnzbd/src/service-setup.sh#L1) `/usr/bin/python3`. Just like many other packages..

Forge36:
Should there be a project to track python 3.8 migration completion? (Python 38 full version?)

hgy59:
> I know for SAB we just [use](https://github.com/SynoCommunity/spksrc/blob/master/spk/sabnzbd/src/service-setup.sh#L1) `/usr/bin/python3`. Just like many other packages..

Indeed, me must care that python38 does not create a link `/usr/bin/python3` and dependent packages must no rely on this.
I think most of the python3 dependent packages do not use such a link, but add /var/packages/python3/target/bin to the package specific path (at least startable services).

rs1gg:
As far as I understand the state of this PR, the community agrees to have separate packages for python minor versions, because it allows to separate the update lifecycles for different applications (some applications need >=3.8, not all applications can go from 3.7 to 3.8 at the same point in time).

For the symlinks (as hgy59 points out) I think it would be ok if:
- /usr/local/bin/python3 still points to 3.7 (from the python3 package) in order to not break existing dependencies
- /usr/local/bin/python38 points to current 3.8.x (from the python38 package) 

which is already what happens right now.

Applications that explicitly need 3.8.x (such as current homeassistant core) can use /usr/local/bin/python38. I already have a first prototype based on smaarn:feat/python38 with HA core 2021.2.2 compiling and running on XPEnology (arch-bromolow).

What does it take for this request to be merged? I don't want to rush, I just want to understand the procedures.

Safihre:
@rs1gg you seem exactly right.
@smaarn ready to merge? Does it need more testing?

smaarn:
> @rs1gg you seem exactly right.
> 
> @smaarn ready to merge? Does it need more testing? 

All good on my end in terms of tests. Ready to merge on my end.

spk/python38/Makefile
hgy59: Yes, I have removed this obsolete code.

spk/python38/src/service-setup.sh
```diff
@@ -0,0 +1,17 @@
...
+
+    # Log installation informations
+    ${SYNOPKG_PKGDEST}/bin/python3 --version >> ${INST_LOG} 2>&1
+    echo -e "\nSystem installed modules:" >> ${INST_LOG}
+    ${SYNOPKG_PKGDEST}/bin/pip freeze >> ${INST_LOG}
+
+    # Byte-compile in background
+    PYTHON_SHORT_VER=$(${SYNOPKG_PKGDEST}/bin/python3 -c 'import sys; print("{0}.{1}".format(*sys.version_info[:2]))')
+    ${SYNOPKG_PKGDEST}/bin/python3 -m compileall -q -f ${SYNOPKG_PKGDEST}/lib/python${PYTHON_SHORT_VER} >> ${INST_LOG} &
+    ${SYNOPKG_PKGDEST}/bin/python3 -OO -m compileall -q -f ${SYNOPKG_PKGDEST}/lib/python${PYTHON_SHORT_VER} >> ${INST_LOG} &
```
Safihre, line 16: @hgy59 I was testing this package on my NAS which is ancient, and I saw that the `compileall` is run twice. 
Do you know why? 
Or maybe @smaarn knows?

smaarn: To be honest this was a brutal copy paste without any questioning on the purpose.
I'll be checking for next package release (a 3.8.8 version was recently released).

smaarn: It seems that both steps are generating different kind of output and are dedicated to generating python compiled files which will be used depending on the "optimization" requested at runtime (see https://docs.python.org/3/using/cmdline.html#cmdoption-o ).
I'm guessing both run could actually be useful, but since I'm no python expert I should rather keep it quiet here ;)

hgy59: The background compilation was always done twice, even in the python 2.7 package of December 2015

``` bash
    # Byte-compile in background
    ${INSTALL_DIR}/bin/python -m compileall -q -f ${INSTALL_DIR}/lib/python2.7 > /dev/null &
    ${INSTALL_DIR}/bin/python -OO -m compileall -q -f ${INSTALL_DIR}/lib/python2.7 > /dev/null &
```

So I suppose this is done by intention.
The first line creates `*.pyc` files and the second one `*.pyo`. But I don't know whether this has any benefit.

smaarn: @Safihre @hgy59 As a matter of fact, if we **really** wanted to keep some kind of logic we should even re-run a third time it with a single `-O` optimization :(.

At this stage I wonder if we shouldn't extract this logics into separate dedicated scripts ? From what I see most projects are actually running python without any optimization specified so only the first command is useful to them.
The second command would only be useful to code being run with `-OO` switch. 

I see the following leads:

* Remove the second run.
  * according to the documentation, only the **initial** loading time of projects using that switch would be impacted
* Remove the second run and add a script which could be run to force the compileall to be run with optimization parameters being provided as arguments
  * Would need being run at install time by projects requiring it (but then it would need to cache the various optimization requirements asked to rerun those upon upgrade)
  * Would need being run at startup time by projects requiring it

WDYT ?

rs1gg: In my understanding (not 100% sure), .pyo are "drop-in" replacements for .pyc, with assertions and __debug__ stripped (-O) and additionally docstrings stripped (-OO). Precompiling them reduces first load time for runtimes that make use of them. In order to use a .pyo, the runtime must be started with -O/-OO (please confirm/falsify).

So separate runs for -O and -OO would just overwrite each other, since both generate .pyo.

If compile execution time is the issue here, the -OO run might be removed at potential initial loading time cost for some usages of this python SPK. If compile execution time is not a big deal, leaving things unchanged will be the safe option.

Safihre: Now I understand. We at SABnzbd do run with `-OO` so it is usefull. I didn't know it created seperate files on each run.
Thanks for letting me know :)

rs1gg: @Safihre Thanks, I was close to ask for known usages of -OO runtimes in my last reply. But since this is a generic python package, we would never know ;-)

smaarn: Actually, since 3.8 they don't "overlap" each other. Hence the fact that in prior versions only two runs were "necessary". (see https://www.python.org/dev/peps/pep-0488/ for more details)

hgy59: Indeed, the python38 package does not create `*.pyo` files. It creates `*.pyc` and `*.opt-2.pyc` files as by `PEP 488`.

As long as none of the packages runs python with `-O` we don't need to create respective bytecode at installation.

rs1gg: Just to prevent future misunderstandings: pep-0488 has been implemented already in python 3.5, not 3.8 (https://bugs.python.org/issue23731). So this is not an issue regarding 3.7->3.8 upgrade.

Issue #4403, opened 2021-01-21
Can't connect to Ikea Tradfri gateway


reapp3r:
### Setup
_Package Name:_ Home Assistant Core / Tradfri Integration
_Package Version:_ 0.118.5-11

_NAS Model:_ DS218+
_NAS Architecture:_ INTEL Celeron J3355
_DSM version:_ DSM 6.2.3-25426 Update 3

### Expected behavior
Connect from NAS to IKEA Tradfri Gateway

### Actual behavior
Connect dialog box shows "Unknown error occurred" when I press submit
The following is added to the logs:

```
Logger: aiohttp.server
Source: components/tradfri/config_flow.py:168 
First occurred: 13 January 2021, 14:41:39 (10 occurrences) 
Last logged: 7:22:20

Error handling request
Traceback (most recent call last):
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/aiohttp/web_protocol.py", line 422, in _handle_request
    resp = await self._request_handler(request)
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/aiohttp/web_app.py", line 499, in _handle
    resp = await handler(request)
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/aiohttp/web_middlewares.py", line 118, in impl
    return await handler(request)
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/homeassistant/components/http/request_context.py", line 18, in request_context_middleware
    return await handler(request)
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/homeassistant/components/http/ban.py", line 72, in ban_middleware
    return await handler(request)
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/homeassistant/components/http/auth.py", line 127, in auth_middleware
    return await handler(request)
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/homeassistant/components/http/view.py", line 129, in handle
    result = await result
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/homeassistant/components/config/config_entries.py", line 169, in post
    return await super().post(request, flow_id)
  File "/volume1/@appstore/homeassistant/env/lib/python3.7/site-packages/homeassistant/components/http/data_validator.py", line 60, in wrapper
    result = await method(view, request, *args, **kwargs)
[...]

Issue #4382, opened 2021-01-10
Home Assistant Core for Synology NAS and Python3 version issue


MarkusSacramento:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name: Home Assistant Core
_Package Version: 0.118.5-11

_NAS Model: Synoogy NAS DS218Play
_NAS Architecture: ARM8 (Realtek RTD1296 SoC)
_DSM version: DSM 6.2.3-25426 Update 3
--

The PAckage requires Python3 to work.
If not installed it installs it during the installation of HomeAssistant Core package.
The problem is that the package requires the Python3 package create by Allan Clark which is only updated to 3.7.7-15.
It is also possible to install Python3 package created by Python Software Foundation which is updated to 3.8.2-0150 but this isn't supported by the Home Assistant Core package.

My suggestion is to let th Home Assistant Core package verify if python is installed and if as long as the version is at minimum what Home Assistant Core requires a third party package should not be needed.

 Allan Clarks python3 version is to old to run newer Home Assistant updates.

hgy59:
All the synocommunity packages that depend on python3 need the synocommunity python3. It is not possible to support both python3 packages as dependency. The internal package names differ (python3: synocommunity, Python3: synology).
The synocommunity python3 is currently 3.7.7-15 (Maintainer is Allan Clark.) 
The homeassistant package 0.118.5-11 works well with python3 3.7.7-15.

The next homeassistant package needs python3 3.8.x when homeassistant is updated to a version > 0.118.5.
I cannot see the real issue. You didn't tell us about error messages or installation failures.
The only issue I see in my installation is a warning, that further homeassistant updates will not be supported for Python 3.7.

iandarbey:
The ikea integration fails for me using HA core (with python errors). Is an update to python not currently WIP anyway?

MarkusSacramento:
I don't have an issue more than it complains about Python version is out of date and that HA won't work if updated.
Just thought it would be easier working with relationship to the Internal package since it gets updated more frequently. Just a tip and a Heads Up.

hgy59:
@MarkusSacramento thanks for claryfying this 👍 
Out experiance is, that synology is not so fast with updates, but I must confess that with python3 we are delayed due to the cross compiling issues we had with python 3.6...

knag1:
Thank you for your great work! I run ha on a DS120j but now I miss the GOODWE integration and some other.
[...]

PR #4421, opened 2021-01-31
Home Assistant Core: update to 2021.1.5 and python 3.8

hgy59:
_Motivation:_  Home Assistant Core updates after 0.118.5 require python 3.8.
_Linked issues:_  requires #4413, #4382, consider #4403

### Checklist


Forge36:
I'm new to github CI I'm hoping to help with development if possible, or help test new builds. I've been able to clone your branch, and compile on my local docker instance.

I'm at a slight loss where the log files for the check failures end up to investigate the failures.
If I've missed a wiki please let me know which one.

From running a local compile i saw output for python 3.7
Does that mean I need to include changes from #4413 in my local branch? Or that something is wrong with my setup?

Apologies if this isn't the right commination channel, I didn't have any luck finding a discord/slack/forum to communicate on.

hgy59:
> From running a local compile i saw output for python 3.7
> Does that mean I need to include changes from #4413 in my local branch? Or that something is wrong with my setup?

I have pushed this to keep my work, knowing that it is not ment to bild now.
If you want to build it before  #4413 is merged and my branch is rebased, you can clone the repo of @smaarn, checkout the branch `smaarn:feat/python38` and copy the changed files of this PR into your local branch (I recommend to work this way, as homeassistant update has only two files to update). 
_Warning: this is for your local build only and not ment to push back to upstream._

hopefully the python38 goes productive soon.

publicarray:
Thanks @hgy59

This also addresses the security vulnerability: https://www.home-assistant.io/blog/2021/01/23/security-disclosure2/

cuki:
Be sure to also update the `PYTHON_DIR` in [spk/homeassistant/src/service-setup.sh](https://github.com/hgy59/spksrc/blob/14da6d8142f2e32ddc3103b296b23f101fe2d9d7/spk/homeassistant/src/service-setup.sh#L5) to utilize the correct path for specified python38 dependency. Otherwise installation would not complete, unless python3 is (also) installed.

hgy59:
> Be sure to also update the `PYTHON_DIR` in [spk/homeassistant/src/service-setup.sh](https://github.com/hgy59/spksrc/blob/14da6d8142f2e32ddc3103b296b23f101fe2d9d7/spk/homeassistant/src/service-setup.sh#L5) to utilize the correct path for specified python38 dependency. Otherwise installation would not complete, unless python3 is (also) installed.

@cuki thanks for this hint.

spk/homeassistant/src/service-setup.sh
```diff
@@ -2,7 +2,7 @@
 # - installer
 # - start-stop-status
 
-PYTHON_DIR="/usr/local/python3"
+PYTHON_DIR="/var/packages/python38/target"
```
cuki, line 5: Just out of curiosity, what is the difference between **/usr/local/python38** and **/var/packages/python38/target**, they both seem to be a symbolic link which reference to **/volume1/@appstore/python38**.

hgy59: `/usr/local/{package}` is created by the generic installer, but it will not be supported by DSM 7+, therefore I start to use `/var/packages/{package}/target` instead. 
Such links are independent on matter whether the packages are installed on `/volume1`, `/volume2`, etc.

cuki: @hgy59 Thanks for the explanation, much appreciated!

PR #4333, opened 2020-12-27
update ejabberd to  v21.01

hgy59:
_Motivation:_  Ejabberd does not build since some time
_Linked issues:_  #4214. closes #3770, closes #2397, closes #2260, closes #2095

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

### Remarks
- update erlang to v23.2.5
- update ejabberd to v21.01
- ppc853x is not supported anymore
- use generic service setup
- activate install wizard

PR #4436, opened 2021-02-09
Erlang v23.2

DigitalBox98:
_Motivation:_  Update Erlang to version 23.2 

Proposal to have Erlang as a dedicated package and update to latest version. 
It was not possible to test it on my NAS because I now switch to DSM 7.0 to test future package (a similar package has been created & tested fine for DSM 7.0) 

### Checklist


hgy59:
There is alread #4333 to update erlang to 22.3.4.13.
What is the motivation to have a dedicated erlang package?

ymartin59:
@hgy59 Proposed CFLAGS may be interesting to include in your PR. I would approve creation of a "erlang" package as an interactive shell is available - indeed the same way we have python, mono and lua "pure language" packages.

digitalbox94:
The idea behind is that Erlang can be used by other softwares (like RabbitMQ).

I'm trying to propose a RabbitMQ package, so the first step is completed : dependency with Erlang is now working on my NAS (working and tested on DSM7 branch, because I'm now only on DSM7).

PR #4459, opened 2021-02-27
Update SABnzbd to 3.2.0 and use Python 3.8 SPK

Safihre:
### Checklist
- [x] Package upgrade completed successfully, tested on 88f6281
- [x] New installation of package completed successfully, tested on 88f6281

Will require users to manually (at least on my DSM 6.2) install Python 3.8 from #4413 when updating.


Safihre:
@smaarn / @hgy59 Can you review this and publish this? Or is @ymartin59 still the only one that publishes stuff?
I have not been involved in SynoCommunity recently so not sure who is doing what nowadays 😅

Safihre:
@publicarray maybe you can help. Me get this merged and published?

spk/sabnzbd/src/service-setup.sh
```diff
@@ -8,7 +8,7 @@ LANGUAGE="env LANG=en_US.UTF-8"
 
 GROUP="sc-download"
 
-SERVICE_COMMAND="${LANGUAGE} ${PYTHON} ${SABNZBD} -f ${CFG_FILE} --pidfile ${PID_FILE} -d"
+SERVICE_COMMAND="${LANGUAGE} ${PYTHON} -OO ${SABNZBD} -f ${CFG_FILE} --pidfile ${PID_FILE} -d"
```
publicarray, line 11: @Safihre LGTM but what does `-OO` do?

publicarray (approved):

spk/sabnzbd/src/service-setup.sh
Safihre: It runs in "optimized" mode, so potentially a tiny bit faster.
https://docs.python.org/3/using/cmdline.html#cmdoption-o

Safihre:
@publicarray Thank you! Just wondering, when publishing, you still build the packages on your local machine and then push them or is there already automation (Github Actions) in place for it?

publicarray:
Thanks for the update! Yea locally at the moment. From memory there was a concern that by automating there could be more accidental releases (possible error in the automation) that can't be undone easily on the server.

Safihre:
Can't we really automate this? It has bugged me so much ever since working on this project back 3 years ago that building and releasing is still a manual and longg process. The Github Actions pipelines are just so pretty 😍
Why not include something specific in the commit message to specify what packages to build `PUBLISH=sabnzbd`, or `PUBLISH=sabnzbd,nzbget`.

publicarray:
The code is mostly there only thing missing is `make publish-*` if the PUBLISH_API_KEY is set: https://github.com/SynoCommunity/spksrc/blob/master/.github/actions/build-tag.sh

So on a git tag  package-name-version the action would build a git release and publish to synocommunity

publicarray:
While a special commit name is not a bad idea. I think git tags are better suited.

Safihre:
Sounds good to me 🚀

PR #4300, opened 2020-12-10
[WIP] Update Radarr to v3 (.NET Core version)

houndtt:
_Motivation:_  To update Radarr to v3 release (.NET Core version) and correct service startup on Synology.
_Linked issues:_  Fixes issue #3987 and issue #4285.

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

PR #3932, opened 2020-03-25
Add Jellyfin (BETA)

publicarray:
_Motivation:_  I wanted to have a go and try Jellyfin for myself. I probably stay on Plex for a little while longer but Jellyfin is really promising. ~~This build creates a single 148mb binary that includes the required parts of the .NET framework. This way there are not hundreds of small dll files and is in-depended from other .NET framework versions. I'm not a .NET developer so I ask that the people over at jellyfin can have a look at some of the warnings in the build log.~~

~~Jellyfin works but adding plugins likely crashes the app~~

_Linked issues:_  #3660 https://github.com/jellyfin/jellyfin/issues/293

### Checklist
- [x] Build rule `all-supported` [completed successfully](https://github.com/publicarray/spksrc/releases/tag/jellyfin-10.5.2)
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

Here is a compiled version:
~~[jellyfin_x64-6.1_10.5.2-0.spk](https://seby.io/download/jellyfin_x64-6.1_10.5.2-0.spk)~~

**Updated:** https://github.com/publicarray/spksrc/releases/tag/jellyfin-10.7.2

If there are errors please attach the log file:
```sh
# DSM 6 jellyfin log
cat /var/packages/jellyfin/target/var/jellyfin.log
# DSM 7 jellyfin log
cat /var/packages/jellyfin/var/jellyfin.log
# install log
cat /var/log/packages/jellyfin.log
```

PR #4464, opened 2021-03-01
Use .NET core builds for Radarr

### LINKED ISSUE #4398: Radarr V3 package (includes removal of .Net warning)
Title:

[Package Request] Radarr V3

Content: Radarr program that removes the .NET warning. For intel CPU.

Name: Radarr V3
Description: Film and Movie manager. Can be used to import files into other software.
Website: https://radarr.video/#home
Software documentation: https://github.com/Radarr/Radarr
Build/Installation documentation: https://github.com/Radarr/Radarr
Source code: https://github.com/Radarr/Radarr
License: https://github.com/Radarr/Radarr

This would be a really useful package for those who do not run docker. Please consider. 

Thanks for your time.

### LINKED ISSUE #4444: [Bug] Radarr Search Movie does not work
### Setup
_Package Name:_ Radarr
_Package Version:_ 3.0.0.4204

_NAS Model:_ DS418
_NAS Architecture:_ 64bits ???
_DSM version:_ 6.2.3-25426

### Expected behavior
Normally, when we do a search, it should display the movie we search for.

### Actual behavior
If I search for a movie, I got this error: 
> Failed to load search results, please try again.
> Search for 'wonder' failed. Invalid response received from TMDb.
> 

### Steps to reproduce
_1._ Lunch Radarr
_2._ Go to Movie Section
_3._ Search a movie

Insert log here
```
2021-2-19 14:53:28.7|Warn|SkyHookProxy|Error: SecureChannelFailure (Authentication failed, see inner exception.): 'https://radarrapi.servarr.com/v1/search?q=vgv&year='

[v3.0.0.4204] System.Net.WebException: Error: SecureChannelFailure (Authentication failed, see inner exception.): 'https://radarrapi.servarr.com/v1/search?q=vgv&year=' ---> System.Net.WebException: Error: SecureChannelFailure (Authentication failed, see inner exception.) ---> System.Security.Authentication.AuthenticationException: Authentication failed, see inner exception. ---> Mono.Btls.MonoBtlsException: Ssl error:0f000078:elliptic curve routines:OPENSSL_internal:POINT_IS_NOT_ON_CURVE
  at /spksrc/spk/mono/work-aarch64-6.1/mono-5.20.1.34/external/boringssl/crypto/ec/ec.c:835
  at Mono.Btls.MonoBtlsContext.ProcessHandshake () [0x00047] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/Mono.Btls/MonoBtlsContext.cs:211 
  at Mono.Net.Security.MobileAuthenticatedStream.ProcessHandshake (Mono.Net.Security.AsyncOperationStatus status, System.Boolean renegotiate) [0x000a1] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs:829 
  at (wrapper remoting-invoke-with-check) Mono.Net.Security.MobileAuthenticatedStream.ProcessHandshake(Mono.Net.Security.AsyncOperationStatus,bool)
  at Mono.Net.Security.AsyncHandshakeRequest.Run (Mono.Net.Security.AsyncOperationStatus status) [0x00000] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/Mono.Net.Security/AsyncProtocolRequest.cs:283 
  at Mono.Net.Security.AsyncProtocolRequest.ProcessOperation (System.Threading.CancellationToken cancellationToken) [0x000ff] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/Mono.Net.Security/AsyncProtocolRequest.cs:221 
  at Mono.Net.Security.AsyncProtocolRequest.StartOperation (System.Threading.CancellationToken cancellationToken) [0x00046] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/Mono.Net.Security/AsyncProtocolRequest.cs:187 
   --- End of inner exception stack trace ---
  at Mono.Net.Security.MobileAuthenticatedStream.ProcessAuthentication (System.Boolean runSynchronously, Mono.Net.Security.MonoSslAuthenticationOptions options, System.Threading.CancellationToken cancellationToken) [0x0024c] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs:404 
  at Mono.Net.Security.MonoTlsStream.CreateStream (System.Net.WebConnectionTunnel tunnel, System.Threading.CancellationToken cancellationToken) [0x00126] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs:129 
  at System.Net.WebConnection.CreateStream (System.Net.WebOperation operation, System.Boolean reused, System.Threading.CancellationToken cancellationToken) [0x00170] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/System.Net/WebConnection.cs:222 
   --- End of inner exception stack trace ---
  at System.Net.WebConnection.CreateStream (System.Net.WebOperation operation, System.Boolean reused, System.Threading.CancellationToken cancellationToken) [0x00208] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/System.Net/WebConnection.cs:234 
  at System.Net.WebConnection.InitConnection (System.Net.WebOperation operation, System.Threading.CancellationToken cancellationToken) [0x000f7] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/System.Net/WebConnection.cs:263 
  at System.Net.WebOperation.Run () [0x00052] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/System.Net/WebOperation.cs:268 
  at System.Net.WebCompletionSource`1[T].WaitForCompletion () [0x0008e] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/System.Net/WebCompletionSource.cs:111 
  at System.Net.HttpWebRequest.RunWithTimeoutWorker[T] (System.Threading.Tasks.Task`1[TResult] workerTask, System.Int32 timeout, System.Action abort, System.Func`1[TResult] aborted, System.Threading.CancellationTokenSource cts) [0x000e8] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/System.Net/HttpWebRequest.cs:956 
  at System.Net.HttpWebRequest.GetResponse () [0x0000f] in /spksrc/native/mono/work-native/mono-5.20.1.34/mcs/class/System/System.Net/HttpWebRequest.cs:1218 
  at NzbDrone.Common.Http.Dispatchers.ManagedHttpDispatcher.GetResponse (NzbDrone.Common.Http.HttpRequest request, System.Net.CookieContainer cookies) [0x00125] in <c3304113e0934557b92848ed63d0f8d1>:0 
   --- End of inner exception stack trace ---
  at NzbDrone.Common.Http.Dispatchers.ManagedHttpDispatcher.GetResponse (NzbDrone.Common.Http.HttpRequest request, System.Net.CookieContainer cookies) [0x001ee] in <c3304113e0934557b92848ed63d0f8d1>:0 
  at NzbDrone.Common.Http.HttpClient.ExecuteRequest (NzbDrone.Common.Http.HttpRequest request, System.Net.CookieContainer cookieContainer) [0x00080] in <c3304113e0934557b92848ed63d0f8d1>:0 
  at NzbDrone.Common.Http.HttpClient.Execute (NzbDrone.Common.Http.HttpRequest request) [0x00008] in <c3304113e0934557b92848ed63d0f8d1>:0 
  at NzbDrone.Common.Http.HttpClient.Get (NzbDrone.Common.Http.HttpRequest request) [0x00007] in <c3304113e0934557b92848ed63d0f8d1>:0 
  at NzbDrone.Common.Http.HttpClient.Get[T] (NzbDrone.Common.Http.HttpRequest request) [0x00000] in <c3304113e0934557b92848ed63d0f8d1>:0 
  at NzbDrone.Core.MetadataSource.SkyHook.SkyHookProxy.SearchForNewMovie (System.String title) [0x00331] in <d190c5eb47524ebc878d54daaf20567d>:0 

```

Thank you very much

ta264:
_Motivation:_  Radarr will soon be dropping mono compatible builds so we need to swap to .NET core builds
_Linked issues:_  Fixes #4398, #4300, Fixes #4444

### Checklist

The previous attempt at a .NET core build for jackett failed because `libstdc++.so.6` was too old.  This version provides `libstdc++.so.6.0.22` from Debian stretch in the package `lib` directory.  Tested by @Wuszek on a DS218.


Wuszek:
Just tested upgrading from previous Radarr to new, NETcore one using aarch64 spk from this test run build https://github.com/SynoCommunity/spksrc/actions/runs/611623265 and it work pefectly fine. 

Health status reports no issues, finally.
[screenshot](https://imgur.com/2uRzzVA)

JeremieBergeron:
@Wuszek I would like to test it, but I can not see spksrc in the community package.
image
How can I test it?
Thank you

JeremieBergeron:
Hi,
I test it and I do not see any problem.

houndtt:
So I tested this in my VM which was an upgrade from the existing version. It's mostly a clean-install so not much configurations present. While the upgrade completed and says it was running, I could not connect. I then manually stopped and started the app and it seemed to work well. I attempted to restart the app from within the interface but the app only stopped but failed to start again. Manually starting it in the Package Center was successful. Based on these anomalies and the many errors in the log I don't think I'll be using this in my production NAS even though it gives a good health status.

Some logs below:
https://pastebin.com/dTy1G9m2

My setup:
The host NAS is a DS916+ with 8 GB RAM. The details of the VM are as follows:
```
Model name:            VirtualDSM
CPU:                   INTEL Pentium N3710
CPU clock rate:        1.6 GHz
CPU cores:             2
Total physical memory: 2048 MB
DSM version:           DSM 6.2.3-25426 Update 3
```

Status screen:
Screenshot 2021-03-01 at 9 37 59 PM

bakerboy448:
@houndtt something didn’t work right. Do you have the installer log?  You grabbed the test files from the GitHub action outright right? 

Based on your logs it is still launching mono and the wrong, out of date version, [v3.0.1.4259] and spewing kestrel errors everywhere where

This test build is for 3.0.2.

ta264:
>  I attempted to restart the app from within the interface but the app only stopped but failed to start again. 
This is a known bug in 3.0.2 - the restart button in Radarr's UI is broken under net core. Is fixed in 3.1.

houndtt:
> @houndtt something didn’t work right. Do you have the installer log? You grabbed the test files from the GitHub action for this PR right?

Here's the installer log: https://pastebin.com/1gNn0iMi and last app run log: https://pastebin.com/3HQJWbnV
Yes, I grabbed the package build from the [URL](https://github.com/SynoCommunity/spksrc/actions/runs/611623265) above which linked to these [packages](https://pipelines.actions.githubusercontent.com/Iq2h6h9IXoMUQgTLIOxO5M4OA8gQOidpS1ET9TUAvwpP264Y5H/_apis/pipelines/1/runs/2887/signedartifactscontent?artifactName=packages&urlExpires=2021-03-02T01%3A06%3A08.6274557Z&urlSigningMethod=HMACV1&urlSignature=6OYdkD%2FZYe%2FlrPvjps0S%2BqQqoA0foJByEQK0EIeGh74%3D).

> Based on your logs it is still launching mono and the wrong, out of date version, [v3.0.1.4259] and spewing kestrel errors everywhere where. So I don’t think those provided logs correlate at all to this build.
> 
> This test build is for 3.0.2.

The build is 3.0.2 as shown in the screen shot above. I took a log snippet from just after the upgrade. Here's the latest section: https://pastebin.com/nQ9CeWWc

Looking at the tail of the log, it seems to have started running the correct version. I don't know if this was after the reboot I did or not but it seems to be no longer throwing out strange log entries. Not sure why the upgrade created them in the first place though.

bakerboy448:
@houndtt based on those logs; Kestrel errors are only on 3.0.1 and not in 3.0.2;  3.0.2 appears fine?

publicarray:
@ta264 generally we compile the libs per architecture rather than download them. This avoids having different digests per architecture and ensures they are compatible. But considering that it is hard to compile `gcc` and by extension `libstdc++` in this environment I'm leaning towards allowing this as an exception. Thoughts @th0ma7 @hgy59?

publicarray:
I put the status to `do-not-merge` until the above discussion is resolved.

houndtt:
> @houndtt based on those logs; Kestrel errors are only on 3.0.1 and not in 3.0.2; 3.0.2 appears fine?

Yes, but the problem I am asserting is that the upgrade appears to have caused an in-between state which resulted in the pages and pages of Kestrel errors. From the timeline in the logs I see the following:
```
2021-3-1 21:12:16.6 - Successful startup of 3.0.1.4259 reported in logs
Mon Mar  1 21:18:45 AST 2021 - Update started - starting pre-update tasks
Mon Mar  1 21:18:51 AST 2021 - Update complete - starting post-update tasks
2021-3-1 21:19:05.0 - First Kestrel error in log
2021-3-1 21:21:45.1 - Last Kestrel error (after pages and pages of logs across 12 log files)
2021-3-1 21:22:23.0 - First instance of version 3.0.2.4552 reported in logs
2021-3-1 21:22:29.0 - Successful startup of 3.0.2.4552 reported in logs
2021-3-1 21:24:13.6 - Manual restart of Radarr requested
Mon Mar  1 21:25:14 AST 2021 - Manual start of Radarr triggered (likely from Package Center)
```

Based on the above sequence, it appears that the upgrade completed via the Package Center caused the app to go into this in-between state which resulted in the loads of Kestrel errors. During this time I suspect is when I could not connect to the application even though it showed it was running in the Package Center. After this, I believe I manually stopped it in the Package Center and started it again. This is where I believe the logs stabilised. Following this is where I believe I requested a restart from within the app which was not successful based on the bug reported above by @ta264.

I will try to re-create this upgrade scenario later today in the VM to see if I can re-create the above interpretation of events.

ta264:
> @ta264 generally we compile the libs per architecture rather than download them. This avoids having different digests per architecture and ensures they are compatible. But considering that it is hard to compile `gcc` and by extension `libstdc++` in this environment I'm leaning towards allowing this as an exception. Thoughts @th0ma7 @hgy59?

I agree it's a bit of a hack but I think it's probably better than nothing.  The mono builds in Radarr will be switched off in the not too distant future (in case you haven't realised, I'm a Radarr dev) - we're planning to release 3.1 but then I think the next release will be 4.0 which won't support mono.  We'd love to not leave the non-docker-compatible synology users behind, but we're not willing to maintain mono builds to make that happen I'm afraid.

Happy to try to assist to make the hack more palatable.  I guess the other option is to wait for DSM7, and any users that can't / won't upgrade will be stuck on a legacy Radarr build.

ta264:
> Based on the above sequence, it appears that the upgrade completed via the Package Center caused the app to go into this in-between state which resulted in the loads of Kestrel errors. During this time I suspect is when I could not connect to the application even though it showed it was running in the Package Center. After this, I believe I manually stopped it in the Package Center and started it again. This is where I believe the logs stabilised. Following this is where I believe I requested a restart from within the app which was not successful based on the bug reported above by @ta264.
> 
> I will try to re-create this upgrade scenario later today in the VM to see if I can re-create the above interpretation of events.

@houndtt In the light of morning, I'm guessing that somehow the change to net core meant that the service program didn't manage to shut down the old Radarr version before beginning the upgrade.  I can take a look into this in more detail.  @publicarray If you have any hints as to how the existing service gets stopped that would be handy.

publicarray:
Thanks @ta264 for working on this. I appreciate your help. 
I've also attempted to compile gcc but failed: https://github.com/publicarray/spksrc/commit/5038ab9891373965985e627768e630fe0762aaf2 I've more progress locally but still not much better. Since this the first package that I'm aware of that downloads a lib from Debian I need to wait for feedback from the others I've CCed.

> Happy to try to assist to make the hack more palatable.
Thanks, the framework can compile dotnet apps. I suggest you look at #4207 and use `include ../../mk/spksrc.cross-dotnet.mk` instead of `include ../../mk/spksrc.cross-cc.mk`. It should simplify the makefile. Please let me know if you need help.

ta264:
> Thanks, the framework can compile dotnet apps. I suggest you look at #4207 and use `include ../../mk/spksrc.cross-dotnet.mk` instead of `include ../../mk/spksrc.cross-cc.mk`. It should simplify the makefile. Please let me know if you need help.

Thanks, but we don't support third party compilations of the Radarr binaries.  If you want to package Radarr and have users able to access our third party support you'll need to use the downloaded Radarr binaries.  These are what will be grabbed on the first upgrade via the built in mechanism anyway.  You're already downloading the mono binaries so this isn't a new exception.

Downloading the libstdc++ is a hack (I don't disagree on that at all!) but compiling gcc is a tricky thing!

ta264:
@houndtt Can I confirm what version of the synology package you were on before?  I guess you were using the updated v3 package since I think you did the work for that, but would be good to confirm.  One case we do perhaps need to guard against is a user going from v0.2 to v3, where the name of the PID file changes, but I'm guessing that isn't the case for you.

I'm a little surprised I don't see something corresponding to 
```
echo "Stopping ${DNAME} ..."
```
In the update logs, but this is the first time I've looked at synology packages in any detail.  If there is a problem stopping radarr before the upgrade I think I'm going to need some assistance from the experts to fix it.

publicarray:
> Thanks, but we don't support third party compilations of the Radarr binaries, If you want to package Radarr and have users able to access our upsteam support you'll need to use the downloaded Radarr binaries

@ta264 Sorry but I disagree on your `upsteam support` idea. We have always provided support to users that use our packages here. IMHO The main difference compiling would be that we only need to download the source code once and not a different binary per architecture (4x). Speeding up releases. It also simplifies the code. I looked at how raddar does its releases and the makefiles should be able to compile almost identical binaries (difference being the hash): https://github.com/SynoCommunity/spksrc/blob/master/mk/spksrc.cross-dotnet-env.mk. 
In the future we could also add a .NET runtime package, reducing on disk file size if you install multiple .NET packages and allow us to update the runtime independently (not sure if this will happen but the possibility is there). Lastly any package repository that I've come across would like to compile the binaries themselves so this is not new.

>  You're already downloading the mono binaries so this isn't a new exception.

I see where you're coming from but .NET is different, the framework is embedded into the binary. Previously we compiled mono per arch and the mono binaries worked across arches.

ta264:
I'm afraid we'll have to agree to disagree.

Compiling the backend is only one aspect - there's also the frontend which requires yarn.  I would council that this is a futile effort, because the first upgrade from the built in updater will remove your compiled binaries and replace them with ones from our CI, which are the full standalone build.

We have actively decided against shipping framework dependent builds precisely so that we are in control of which framework Radarr is running on, which will avoid many of the issues we faced with mono.  We have no plans to ever produce framework independent builds.

You may be providing some support but I can assure you *a lot* of Synology package users come by our discord asking for help.  To be blunt, A synocommunity compiled Radarr will be on you to support.

publicarray:
> We have actively decided against shipping framework dependent builds precisely so that we are in control of which framework Radarr is running on

That's a good point, If you require `full standalone builds` than that is fine. 

However, personally I still prefer if the package was build. Regarding yarn, I've made a Jellyfin package that uses nodejs and npm: #3932 It's not merged yet but that should address the front-end issue.

ta264:
> I've made a Jellyfin package that uses nodejs and npm: #3932 It's not merged yet but that should address the front-end issue.

You'd need to add yarn addionally, nodejs and npm isn't sufficient for Radarr / Sonarr.

> However, personally I still prefer if the package was build.

Of course that choice is up to you.  I've suggested a way that we as Radarr devs would be happy to help support.  If you wish to go in a different direction then I wish you well and I'll let you get on with it.

publicarray:
@ta264 Sorry to be clear it's not only up to me. The other @SynoCommunity/developers will write what they think first.

publicarray:
Here is a wild proposal: Maybe we do what [AUR](https://aur.archlinux.org/packages/) does and postfix binary package as `-bin` and git packages as `-git`? Then they are out of our (SynoCommunity) control and all support links must point to the package maintainer. Would this be acceptable @ta264 @SynoCommunity/developers?

ta264:
I think that would be ideal from our point of view.  If I understand it you're proposing relabelling the existing `radarr` as `radarr-bin`, and then we can carry on along the lines proposed in this PR.  We would be happy (I assume, thought must check with my fellow Radarr devs) to be the main point of contact for support for this package.  We may need some assistance in terms of ironing out any bugs in the service control scripts etc, but hopefully that should be fairly straightforward.  And once the mono dependency is gone, hopefully synology specific bugs are rarer and users get the big speed bump .NET core brings.

As an aside, my view is that it's inconsistent to require that spks in this repository are compiled from source but to allow self-updating packages.  If you want to insist on compiling from source, you should also disable the built in updaters (which are downloading binaries).  But I'm not sure that you want to commit to having to update your packages for every release of Radarr?

A big plus side from our POV of having a self-updating package is we have reasonable number of users on the nightly branch (i.e. an update for every commit to the upstream Radarr repository).  This means that issues are normally picked up pretty quickly and fixed.  If you disable self-update for Radarr, you are locking synology users into only using the builds we *think* are stable, but we won't have had chance to see any synology specific issues, so I think you dramatically increase the chance of us pushing out a "stable" build that doesn't work properly on synology, with no easy way of testing fixes.  As mentioned, if you are willing to allow a self-updating package, I don't think insisting that spks are built from source is a principle worth falling out over.

We're pretty reasonable guys, we'd love to find a solution to keep the optimal experience for Synology users.  My views in this thread are born out of the practicalities of contributing to a complex app like radarr rather than just a desire to be in control or any particular principle.  I believe self-updating packages are a win for your users, and sticking to the binaries we provides makes it much easier for us to identify where issues are arising.

th0ma7:
> If I understand it you're proposing relabelling the existing `radarr` as `radarr-bin`

Personally I think this is a reasonable approach.  I guess a `radarr-nightly-bin` could also be useful or a method to choose between the two streams at install time where post-install at first update it would fetch `nightly` when applicable (see `src/wizard` `.ini` + checkbox example in PR #4144 that could be useful)

Note that I am a big fan of always rebuilding from scratch.  But in this case, creating a decent `libstdc++` dependency is not a short term goal (although now added to my personal TODO list... but I expect months before tires can hit the road).  One could argue to support only DSM-7 with the newer compiler and that may work but even there, today, we don't have such solution available just yet.

> As an aside, my view is that it's inconsistent to require that spks in this repository are compiled from source but to allow self-updating packages. If you want to insist on compiling from source, you should also disable the built in updaters (which are downloading binaries). But I'm not sure that you want to commit to having to update your packages for every release of Radarr?

This is where the `-bin` package are probably easier to get self-updated.  Otherwise would this mean having a `stdc++` package available as well as dependency?
 
> We're pretty reasonable guys, we'd love to find a solution to keep the optimal experience for Synology users. My views in this thread are born out of the practicalities of contributing to a complex app like radarr rather than just a desire to be in control or any particular principle. I believe self-updating packages are a win for your users, and sticking to the binaries we provides makes it much easier for us to identify where issues are arising.

I believe that having inter-project collaborations is more than welcomed :)

houndtt:
> @houndtt Can I confirm what version of the synology package you were on before? I guess you were using the updated v3 package since I think you did the work for that, but would be good to confirm. One case we do perhaps need to guard against is a user going from v0.2 to v3, where the name of the PID file changes, but I'm guessing that isn't the case for you.
> 
> I'm a little surprised I don't see something corresponding to
> 
> ```
> echo "Stopping ${DNAME} ..."
> ```
> 
> In the update logs, but this is the first time I've looked at synology packages in any detail. If there is a problem stopping radarr before the upgrade I think I'm going to need some assistance from the experts to fix it.

@ta264, you are correct. I was already using a v3 package prior to executing the update. Unfortunately, when I tried to replicate the issue two different ways this morning I was unable to do so.

In my original setup I was running the last build I did which was merged running v3.0.1.4259 (https://github.com/SynoCommunity/spksrc/actions/runs/420620824). When I did a clean install with this same version and then an upgrade from the Package Center the logs look nice and clean: https://pastebin.com/kFWXAAZ6

Furthermore, if I install from the current SynoCommunity published version which was the previous one I built running v3.0.0.4204 (https://github.com/SynoCommunity/spksrc/actions/runs/411699059) this too was clean: https://pastebin.com/8BLEa3Bc

I cannot explain why the original upgrade I did failed in the way it did since I can't replicate it on the same setup unfortunately.

ta264:
> This is where the `-bin` package are probably easier to get self-updated. Otherwise would this mean having a `stdc++` package available as well as dependency?

Yes, if you want to compile from source and support Radarr past 3.1 (once we drop mono) you'll need to have a `stdc++` package, or wait for DSM7 where the library synology ship is new enough.  I would also point out that you're not compiling the dotnet sdk itself, so I'm not quite sure why that's OK as a binary, but libstdc++ or Radarr are not.

My only concern with renaming `radarr` to `radarr-bin` is whether the package would auto migrate for users?  From a user's point of view, being told they have to swap from one package (which just downloaded binaries) to another package (which also just downloads binaries) is rather clunky.

ta264:
> I cannot explain why the original upgrade I did failed in the way it did since I can't replicate it on the same setup unfortunately.

Thanks a lot for testing!  If you do manage to reproduce anything let me know.

publicarray:
> My only concern with renaming `radarr` to `radarr-bin`

@ta264 thanks I should have caught that error too. 

I just want to point out that SynoCommunity is more than a distribution platform. Since you have a good technical reason (tie a release to a specific .NET) I think the package is Ok as is. My reason for updating the support links is because we can no longer "control" the code that is run. IMHO not providing support because it's not compiled through your CI is against what I thought OpenSource stands for. On the other hand I understand the difficulty with providing support when the dependencies/.NET don't match. But I think we could have kept the same versions in sync (maybe deviate a little from .NET patch version).

To be clear if support requests land here we can still provide support, but we I don't think we can be the main support channel. (not sure if we ever where if support is mostly on discord now 🤷)

> not compiling the dotnet sdk itself

There was talk to do it but it requires a complex toolchain setup. I think @hgy59 was working on doing it. If we do get it the aim is to add x64_32bit support too.

angrycuban13:
> IMHO not providing support because it's not compiled through your CI is against what I thought OpenSource stands for. 

I thought open source stood for something else but that might just be me...

publicarray:
I often compile from source when filing bug reports. Also tools like git bisect come in handy

ta264:
It seems like there's a common ground we can aim for! :)

> Since you have a good technical reason (tie a release to a specific .NET) I think the package is Ok as is. My reason for updating the support links is because we can no longer "control" the code that is run.

I'm happy with that.  I couldn't see any support links in the `Makefile` but if you point me to them I can update them.  Within the app, the support links already point to our discord server / reddit / GitHub repository.

I propose the following:

1. We use the package along the current lines (with binary Radarr / libstdc++) to establish whether it fixes the problem you found with Jackett in the wild.
2. If it all goes well and you still feel very strongly about compiling from source, I'm willing to try to assist you in setting up a source build that is in line with our CI process and so produce a source compiled package we are willing to support.  It's not totally straightforward, hence my reluctance, but I can respect your principles here.  My condition for this is that you use *exactly* the .NET framework version (down to the same patch) we used to build the corresponding build in our CI.  It's going to be a fair bit of work I fear so not something I'm willing to take on myself.
3. I have some experience compiling dotnet 5 sdk for the FreeBSD builds we produce.  The process could probably be adapted to add linux-x86 support and I'm happy to provide some assistance with that.  It would also require source code changes in Radarr / Jackett and a custom upstream nuget package to produce a linux-x86 build but we could probably figure it out together.

Safihre:
Just to add my opinion: seeing that so far we have multiple packages where we don't compile everything (Sonarr, current Radarr, NZBGet), I don't see it as a bad thing at all to just use the pre-compiled ones 👍 
I have seen many mono related problems over the years, so I think the benefit of solving those outweighs the use of pre-compiled binaries.

ta264:
The latest build cleans up a debug line I'd added to the extract makefile, and adds a `package_info` file so that the radarr status screen should show something like:

```
Version 3.0.2.4552
Package Version radarr_aarch64-6.1_20210303-15 by Team Radarr (https://github.com/Radarr/Radarr)
```

spk/radarr/src/service-setup.sh
```diff
@@ -35,12 +27,6 @@ service_postinst ()
     # If nessecary, add user also to the old group before removing it
     syno_user_add_to_legacy_group "${EFF_USER}" "${USER}" "${LEGACY_GROUP}"
     syno_user_add_to_legacy_group "${EFF_USER}" "${USER}" "users"
-
-    # Discard legacy obsolete busybox user account
-    BIN=${SYNOPKG_PKGDEST}/bin
-    $BIN/busybox --install $BIN >> ${INST_LOG}
-    $BIN/delgroup "${USER}" "users" >> ${INST_LOG}
-    $BIN/deluser "${USER}" >> ${INST_LOG}
```
ta264, line 43: Is this safe to remove? It's very old and radarr doesn't depend on busybox itself

spk/radarr/src/service-setup.sh
```diff
@@ -55,39 +41,17 @@ service_preupgrade ()
...
-    . ${CONFIG_DIR}/KEEP_VAR
-    if [ "$KEEP_CUR" == "yes" ]; then
-        echo "Restoring Radarr version from before upgrade" >> ${INST_LOG}
-        rm -fr ${SYNOPKG_PKGDEST}/share >> ${INST_LOG} 2>&1
-        mv ${INST_VAR}/share ${SYNOPKG_PKGDEST}/ >> ${INST_LOG} 2>&1
-        set_unix_permissions "${SYNOPKG_PKGDEST}/share"
-    fi
+    # Make Radarr do an update check on start to avoid possible Radarr
+    # downgrade when synocommunity package is updated
+    touch ${CONFIG_DIR}/update_required
```
ta264, line 50: creating this file should prompt radarr to do an update check on startup, negating the need for manually preserving the binaries

spk/radarr/Makefile
```diff
@@ -1,22 +1,21 @@
 SPK_NAME = radarr
 SPK_VERS = $(shell date +%Y%m%d)
-SPK_REV = 14
+SPK_REV = 15
 SPK_ICON = src/radarr.png
 
 REQUIRED_DSM = 5.0
```
ta264, line 6: Should this be increased?

Wuszek:
Updated using latest build. It works ok, and it looks like this when upgrading.
radarr-update1
radarr-update2

ta264:
The latest change means this shows up in radarr about screen now:
image

ta264:
@Wuszek confirms that upgrading from 3.0.2.4552 in the spk to 3.1.0.4462 works correctly (and bumps .NET to 5.0.3 in the process) 
image

ta264:
@Wuszek has gone above and beyond and we have successfully tested out the "automatic update on package upgrade" functionality, so that if a user is on a newer build of Radarr, and installs an SPK with bundling an older version, radarr will upgrade itself back to the correct version instead of failing.

From my point of view, I think this is good to go now unless you have any other comments.  I updated the maintainer to `Team Radarr` - not sure if this was what you meant by the support links.  @Wuszek mentioned that this `Team Radarr` text still links to SynoCommunity, I'm not sure how to change that.

ta264:
OK `helpurl` updated to https://wiki.servarr.com/Radarr and `support_url` to https://radarr.video/#support

Please let me know if you have any other comments

ta264:
PS I looked into building net 5 for x86 and ran into this issue, which is a bit beyond my ability to understand... https://github.com/dotnet/runtime/issues/40003

houndtt:
> OK `helpurl` updated to https://wiki.servarr.com/Radarr and `support_url` to https://radarr.video/#support
> 
> Please let me know if you have any other comments

I downloaded the new [source](https://github.com/SynoCommunity/spksrc/actions/runs/618668964) packages and tried an upgrade using this path:
1. Clean-install current version in SynoCommunity (v3.0.0.4204)
2. Upgraded in-app to current release (v3.0.2.4552)
3. Upgraded package using new source (at same v3.0.2.4552)

The upgrade process seemed to work well. Post-upgrade the Package Center looks like this:

Screenshot 2021-03-04 at 4 59 44 AM

And the application status page looks like this:

Screenshot 2021-03-04 at 5 00 12 AM

From what I see, the log files also continue to look clean: https://pastebin.com/rT9SZHb4. Hope this helps.

Wuszek:
Lgtm, later I will try to upgrade yestarday's package with new one. I'm still on nightly, so we will see how it works on this branch.

Safihre:
Maybe a stupid question, but can the other *arr programs also use this method instead of Mono?

ta264:
Yes for Lidarr / Readarr.  Sonarr is still .NET Framework (and so mono) until v4.

miigotu:
This might be a strange question, but why is it so difficult for the creators of software to get their software into a community project that extends the available applications to syno users?

If the binaries are provided by radarr, and they have control over the package that provides their software, it's understood that it is them who provides the user support, not synocommunity.

We have been trying to get a proper SickChill package in for years, and a proper SickRage package in for years (sickrage wasn't allowed to have our own package until echelon took the name, we had to modify sickbeard-custom which is no longer compatible with SickChill). Every time we make the changes they request, it's more changes requested undoing or completely changing what they asked the last go round, and it never ends. 

It's very frustrating, because this org is supposed to be for users, for the community.

azz22:
Hi @houndtt, i'm sorry but not as technical as you guys. Ive got as far as 2. in your comment. How do i do 3. and where do i get it from? I'd really like to finish this as the .NET issue is a bug bare.

Wuszek:
@azz22 hi, download package from here https://github.com/SynoCommunity/spksrc/suites/2171641228/artifacts/44632009
choose appropriate package for your system, and install it manually from package center

azz22:
@Wuszek, really appreciate your help. installed on a DS716+, no issues. Restored my settings and DB.

Screenshot 2021-03-09 at 11 10 58

ta264:
Just for the record, if you have an x64 synology then docker is by far the better way to run Radarr.  These .NET core packages are really a hack to support those devices that don't support docker and don't have a working mono build.

Wuszek:
> maybe a placebo effect but thought i would mention.
@azz22 not really a placebo, just .NET core :+1:

publicarray:
@ta264 I've added a few commits with some minor cleanup and fixed the URL issue (All current contributors need to update their `local.mk` to allow the MAINTAINER_URL to be overridden, just add a `?` before the `=`). Are you ok with these changes?

I can confirm the correct libstdc++ is loaded (at least for the .NET bin):

<details>
<summary>ldd output (ldd copied from manjaro)</summary>
<pre>
root@dsm6 /t/ss# ./ldd /var/packages/radarr/target/share/Radarr/bin/Radarr 
/var/packages/radarr/target/share/Radarr/bin/Radarr: /lib/libstdc++.so.6: no version information available (required by /var/packages/radarr/target/share/Radarr/bin/Radarr)
/var/packages/radarr/target/share/Radarr/bin/Radarr: /lib/libstdc++.so.6: no version information available (required by /var/packages/radarr/target/share/Radarr/bin/Radarr)
/var/packages/radarr/target/share/Radarr/bin/Radarr: /lib/libstdc++.so.6: no version information available (required by /var/packages/radarr/target/share/Radarr/bin/Radarr)
/var/packages/radarr/target/share/Radarr/bin/Radarr: /lib/libstdc++.so.6: no version information available (required by /var/packages/radarr/target/share/Radarr/bin/Radarr)
/var/packages/radarr/target/share/Radarr/bin/Radarr: /lib/libstdc++.so.6: no version information available (required by /var/packages/radarr/target/share/Radarr/bin/Radarr)
	linux-vdso.so.1 (0x00007ffe6951f000)
	libpthread.so.0 => /lib/libpthread.so.0 (0x00007f9b0e2ec000)
	libdl.so.2 => /lib/libdl.so.2 (0x00007f9b0e0e8000)
	libstdc++.so.6 => /lib/libstdc++.so.6 (0x00007f9b0ddf6000)
	libm.so.6 => /lib/libm.so.6 (0x00007f9b0daf1000)
	libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x00007f9b0d8e0000)
	libc.so.6 => /lib/libc.so.6 (0x00007f9b0d53a000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x000055c31ff00000)
root@dsm6 /t/ss# LD_LIBRARY_PATH=/var/packages/radarr/target/lib ./ldd /var/packages/radarr/target/share/Radarr/bin/Radarr
	linux-vdso.so.1 (0x00007ffe13355000)
	libpthread.so.0 => /lib/libpthread.so.0 (0x00007fd14bc1a000)
	libdl.so.2 => /lib/libdl.so.2 (0x00007fd14ba15000)
	libstdc++.so.6 => /var/packages/radarr/target/lib/libstdc++.so.6 (0x00007fd14b693000)
	libm.so.6 => /lib/libm.so.6 (0x00007fd14b38e000)
	libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x00007fd14b17d000)
	libc.so.6 => /lib/libc.so.6 (0x00007fd14add7000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x0000559fc48ae000)
</pre>
</details>

Shame about the build date being the version number... too late to change it now without breaking updates.

spk/radarr/src/service-setup.sh
publicarray: Yes, we are removing them as we update packages

spk/radarr/Makefile
publicarray: I don't think it's necessary. I can't test on DSM 5.0, but we release mainly on DSM 6.x anyway. Upping it would just block anyone from building it on lower DSM versions locally. IMHO it could just be removed.

ta264:
Your changes look fine to me, thanks for cleaning it up.

ta264:
Thanks!

publicarray:
The package is [published](https://synocommunity.com/package/radarr) but will take around 48 hours for our CDN to distribute it to the package center.

houndtt:
Okay, the published version appeared on my NAS and I was able to upgrade successfully. The upgrade process took quite some time to complete but seemed to do so without major error. Below are the logs:

_radarr_install.log_
```
Fri Mar 12 20:18:37 -04 2021
===> Step preupgrade. USER=radarr GROUP=sc-download SHARE_PATH=
Invoke service_preupgrade
mkdir: cannot create directory '/volume1/@appstore/radarr/var/.config': File exists
Fri Mar 12 20:21:09 -04 2021
===> Step preuninst. USER=radarr GROUP=sc-download SHARE_PATH=
Fri Mar 12 20:21:15 -04 2021
===> Step postuninst. USER=radarr GROUP=sc-download SHARE_PATH=
Fri Mar 12 20:21:16 -04 2021
===> Step preinst. USER=radarr GROUP=sc-download SHARE_PATH=
Fri Mar 12 20:21:20 -04 2021
===> Step postinst. USER=radarr GROUP=sc-download SHARE_PATH=
Installing service configuration /var/packages/radarr/conf/radarr.sc
Invoke service_postinst
Granting 'sc-radarr' unix ownership on /volume1/@appstore/radarr/var/.config
Granting 'sc-radarr' unix ownership on /volume1/@appstore/radarr/var
Fri Mar 12 20:21:21 -04 2021
===> Step postupgrade. USER=radarr GROUP=sc-download SHARE_PATH=
Granting 'sc-radarr' unix ownership on /volume1/@appstore/radarr/var
Invoke service_postupgrade
Granting 'sc-radarr' unix ownership on /volume1/@appstore/radarr/var/.config
Granting 'sc-radarr' unix ownership on /tmp/radarr_backup
Granting 'sc-radarr' unix ownership on /tmp/radarr_update
chown: cannot access '/tmp/radarr_update': No such file or directory
```

_radarr.txt_
```
2021-3-12 20:22:33.9|Info|Bootstrap|Starting Radarr - /volume1/@appstore/radarr/share/Radarr/bin/Radarr.dll - Version 3.0.2.4552
2021-3-12 20:22:35.3|Warn|SingleInstancePolicy|Failed to check for multiple instances of Radarr.

[v3.0.2.4552] System.UnauthorizedAccessException: Access to the path '/proc/897/maps' is denied.
 ---> System.IO.IOException: Permission denied
   --- End of inner exception stack trace ---
   at System.IO.FileStream.CheckFileCall(Int64 result, Boolean ignoreNotSupported)
   at System.IO.FileStream.ReadNative(Span`1 buffer)
   at System.IO.FileStream.ReadSpan(Span`1 destination)
   at System.IO.FileStream.Read(Byte[] array, Int32 offset, Int32 count)
   at System.IO.StreamReader.ReadBuffer()
   at System.IO.StreamReader.ReadLine()
   at System.IO.ReadLinesIterator.MoveNext()
   at Interop.procfs.ParseMapsModulesCore(IEnumerable`1 lines)+MoveNext()
   at System.Diagnostics.ProcessManager.GetModules(Int32 processId)
   at System.Diagnostics.Process.get_Modules()
   at NzbDrone.Common.Processes.ProcessProvider.<>c__DisplayClass21_0.<GetProcessesByName>b__0(Process process) in D:\a\1\s\src\NzbDrone.Common\Processes\ProcessProvider.cs:line 344
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Set`1.UnionWith(IEnumerable`1 other)
   at System.Linq.Enumerable.UnionIterator`1.FillSet()
   at System.Linq.Enumerable.UnionIterator`1.ToList()
   at NzbDrone.Common.Processes.ProcessProvider.GetProcessesByName(String name)
   at NzbDrone.Common.Processes.ProcessProvider.FindProcessByName(String name) in D:\a\1\s\src\NzbDrone.Common\Processes\ProcessProvider.cs:line 91
   at Radarr.Host.SingleInstancePolicy.GetOtherNzbDroneProcessIds() in D:\a\1\s\src\NzbDrone.Host\SingleInstancePolicy.cs:line 68


2021-3-12 20:22:35.4|Info|Router|Application mode: Interactive
2021-3-12 20:22:35.7|Info|MigrationController|*** Migrating data source=/volume1/@appstore/radarr/var/.config/Radarr/radarr.db;cache size=-20000;datetimekind=Utc;journal mode=Wal;pooling=True;version=3 ***
2021-3-12 20:22:36.5|Info|MigrationController|*** Migrating data source=/volume1/@appstore/radarr/var/.config/Radarr/logs.db;cache size=-20000;datetimekind=Utc;journal mode=Wal;pooling=True;version=3 ***
2021-3-12 20:22:40.8|Info|WebHostController|Listening on the following URLs:
2021-3-12 20:22:40.8|Info|WebHostController|  http://*:8310
2021-3-12 20:22:41.3|Info|RadarrBootstrapper|Starting Web Server
2021-3-12 20:22:46.4|Info|CommandExecutor|Starting 2 threads for tasks.
```

Looking at the above it's strange that it is trying to access the /proc directory. At first I thought it was a fluke so I triggered a restart from within the app. This process didn't complete because it seemed to be doing the same thing as shown below:

```
2021-3-12 20:33:38.2|Info|LifecycleService|Restart requested.
2021-3-12 20:36:18.0|Info|Bootstrap|Starting Radarr - /volume1/@appstore/radarr/share/Radarr/bin/Radarr.dll - Version 3.0.2.4552
2021-3-12 20:36:19.2|Warn|SingleInstancePolicy|Failed to check for multiple instances of Radarr.

[v3.0.2.4552] System.UnauthorizedAccessException: Access to the path '/proc/897/maps' is denied.
 ---> System.IO.IOException: Permission denied
   --- End of inner exception stack trace ---
   at System.IO.FileStream.CheckFileCall(Int64 result, Boolean ignoreNotSupported)
   at System.IO.FileStream.ReadNative(Span`1 buffer)
   at System.IO.FileStream.ReadSpan(Span`1 destination)
   at System.IO.FileStream.Read(Byte[] array, Int32 offset, Int32 count)
   at System.IO.StreamReader.ReadBuffer()
   at System.IO.StreamReader.ReadLine()
   at System.IO.ReadLinesIterator.MoveNext()
   at Interop.procfs.ParseMapsModulesCore(IEnumerable`1 lines)+MoveNext()
   at System.Diagnostics.ProcessManager.GetModules(Int32 processId)
   at System.Diagnostics.Process.get_Modules()
   at NzbDrone.Common.Processes.ProcessProvider.<>c__DisplayClass21_0.<GetProcessesByName>b__0(Process process) in D:\a\1\s\src\NzbDrone.Common\Processes\ProcessProvider.cs:line 344
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Set`1.UnionWith(IEnumerable`1 other)
   at System.Linq.Enumerable.UnionIterator`1.FillSet()
   at System.Linq.Enumerable.UnionIterator`1.ToList()
   at NzbDrone.Common.Processes.ProcessProvider.GetProcessesByName(String name)
   at NzbDrone.Common.Processes.ProcessProvider.FindProcessByName(String name) in D:\a\1\s\src\NzbDrone.Common\Processes\ProcessProvider.cs:line 91
   at Radarr.Host.SingleInstancePolicy.GetOtherNzbDroneProcessIds() in D:\a\1\s\src\NzbDrone.Host\SingleInstancePolicy.cs:line 68


2021-3-12 20:36:19.3|Info|Router|Application mode: Interactive
2021-3-12 20:36:19.6|Info|MigrationController|*** Migrating data source=/volume1/@appstore/radarr/var/.config/Radarr/radarr.db;cache size=-20000;datetimekind=Utc;journal mode=Wal;pooling=True;version=3 ***
2021-3-12 20:36:20.2|Info|MigrationController|*** Migrating data source=/volume1/@appstore/radarr/var/.config/Radarr/logs.db;cache size=-20000;datetimekind=Utc;journal mode=Wal;pooling=True;version=3 ***
2021-3-12 20:36:21.3|Info|WebHostController|Listening on the following URLs:
2021-3-12 20:36:21.3|Info|WebHostController|  http://*:8310
2021-3-12 20:36:22.0|Info|RadarrBootstrapper|Starting Web Server
2021-3-12 20:36:27.5|Info|CommandExecutor|Starting 2 threads for tasks.
```

If this is a bug then a new issue would need to be opened. Anyone else experiencing this?

publicarray:
> This might be a strange question, but why is it so difficult for the creators of software to get their software into a community project that extends the available applications to syno users?

Limited time/resources usually. From my (limited) experience here there are on average 0~3 people with the ability to publish packages for a given week. There over a hundred of packages here and Synology is not a typical Linux environment. Testing is done manually per package. And I do get it wrong a lot because we have very limited architectures/units to test on.

> If the binaries are provided by radarr, and they have control over the package that provides their software, it's understood that it is them who provides the user support, not synocommunity.

I don't disagree, but the reality is mixed. To begin with the package was maintained by synocommunity, I think it's the first time a radarr developer submitted a PR for radarr: https://github.com/SynoCommunity/spksrc/commits/955a8537929aa4151353cca1355ea40d40cbb316/spk/radarr/Makefile

> We have been trying to get a proper SickChill package in for years, and a proper SickRage package in for years (sickrage wasn't allowed to have our own package until echelon took the name, we had to modify sickbeard-custom which is no longer compatible with SickChill). Every time we make the changes they request, it's more changes requested undoing or completely changing what they asked the last go round, and it never ends.
>
> It's very frustrating, because this org is supposed to be for users, for the community.

I emphasise and It saddens me too. I don't have any python knowledge especially how it works with Synology. Otherwise, I would have helped review your package. 

Also, I've tried submission a package to Synology directly. Trust me when I say they have more restrictions than we do. (e.g. all packages must me self-contained, installable offline, updates through their package center only and with no errors in any logfiles (not even mkdir file exists errors which I want to remove here as well since it makes debugging harder and logs more unreadable)

spk/radarr/Makefile
```diff
@@ -1,22 +1,24 @@
...
-# Mono not supported on ppc platforms. C.f. cross/mono/Makefile and references therein for details.
-UNSUPPORTED_ARCHS = $(PPC_ARCHS)
 
-DEPENDS = cross/busybox cross/curl cross/mediainfo cross/sqlite cross/radarr
+# .NET is not supported on PPC, ARM5 or x86
+UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARM5_ARCHES) $(x86_ARCHES)
```
ymartin59, line 9: @publicarray Why not use compilation with Mono implementation for these architectures? (the same way as for Jackett if I remember well)

publicarray: I mean we can add it for one more version:

@ta264 

> I agree it's a bit of a hack but I think it's probably better than nothing. The mono builds in Radarr will be switched off in the not too distant future (in case you haven't realised, I'm a Radarr dev) - we're planning to release 3.1 but then I think the next release will be 4.0 which won't support mono. We'd love to not leave the non-docker-compatible synology users behind, but we're not willing to maintain mono builds to make that happen I'm afraid.

publicarray: Actually there shouldn't be any need since the package auto updates until it won't be supported anymore.

houndtt:
> If this is a bug then a new issue would need to be opened. Anyone else experiencing this?

As this continues to occur for me, I opened a new bug report here: https://github.com/Radarr/Radarr/issues/6068

ta264:
@publicarray not sure of the best place to mention this, apologies if this is not the best place.  Just run through debugging with a user on a 213j where it doesn't work at all - it fails with 
```
Illegal instruction (core dumped)
```
I think we might need to mark armv7l as unsupported too.  Does this sound plausible or do you think there's something else going on?

publicarray:
Yes you are correct, sorry must have missed it in my review. I'll disable the offending architecture

ta264:
Thanks. I think the 213j is an Armada370 which is listed as ARMv7 here: https://github.com/SynoCommunity/spksrc/blob/2f2428cc7f58694bdbd34f7fef72dceb766d5062/mk/spksrc.archs.mk#L9

But the user reported this as the output of `uname -a`:
```
Linux DISKSTATION 3.2.40 #25426 Wed Jul 8 03:15:48 CST 2020 armv7l GNU/Linux synology_armada370_213j
```

Is it misfiled as arm7 and should be armv7l? Or am I reading too much into the uname output?

miigotu:
> Thanks. I think the 213j is an Armada370 which is listed as ARMv7 here:
> 
> https://github.com/SynoCommunity/spksrc/blob/2f2428cc7f58694bdbd34f7fef72dceb766d5062/mk/spksrc.archs.mk#L9
> 
> But the user reported this as the output of `uname -a`:
> 
> ```
> Linux DISKSTATION 3.2.40 #25426 Wed Jul 8 03:15:48 CST 2020 armv7l GNU/Linux synology_armada370_213j
> ```
> 
> Is it misfiled as arm7 and should be armv7l? Or am I reading too much into the uname output?

You would want to check something more like `gcc -dumpmachine | sed "s/-/-$(uname -p)-/"`, uname is unreliable when some archs are compatible with more than one processor class. I just had this issue with ARMv6 (detects as arm7 on uname) and i386/i686 which get uname -a of x86_64 in docker builds at least (have not tested outside docker)

https://github.com/rust-lang/rustup/issues/2700

publicarray:
Going by dev wiki it's ARMv7 https://deviwiki.com/wiki/List_of_NAS_Devices

ta264:
This link suggests its as much about how stuff is compiled as the SoC

https://stackoverflow.com/questions/29166619/differences-between-arm-versions-armv7-only

publicarray:
Oh I learned something new today.

ta264:
I think that might have been a red herring.

Awaiting confirmation from the user, but it my guess is that the armada 370 cpu is armv7 but with an FPU that doesn't support net core.  The requirements are listed [here](https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/cross-building.md#cross-compiling-coreclr-for-other-vfp-configurations).

Looking at a core dump in gdb, it's falling over trying to access registers 16-31, so I guess this cpu will have the `vfpv3d16` feature as mentioned [here](https://github.com/dotnet/runtime/issues/9969).  If confirmed, then armada370 is not supported for net core even though it's an arm7 architecture.

ta264:
Ok mystery solved I think. This is the output of cpuinfo on a ds213j:
```
Processor       : Marvell PJ4Bv7 Processor rev 1 (v7l)
BogoMIPS        : 1196.85
Features        : swp half thumb fastmult vfp edsp vfpv3 vfpv3d16 tls
CPU implementer : 0x56
CPU architecture: 7
CPU variant     : 0x1
CPU part        : 0x581
CPU revision    : 1

Hardware        : Marvell Armada-370
Revision        : 0000
Serial          : 0000000000000000
```
CPUs with vfpv3d16 are not supported by the MS published SDK. It's possible to compile a version that would support it I think but that may be a little too involved...

For now I think we need to mark armada370 as not supported for any net core spks. I don't know if the same would affect the other armada variants.

miigotu:
> Ok mystery solved I think. This is the output of cpuinfo on a ds213j:
> 
> ```
> Processor       : Marvell PJ4Bv7 Processor rev 1 (v7l)
> BogoMIPS        : 1196.85
> Features        : swp half thumb fastmult vfp edsp vfpv3 vfpv3d16 tls
> CPU implementer : 0x56
> CPU architecture: 7
> CPU variant     : 0x1
> CPU part        : 0x581
> CPU revision    : 1
> 
> Hardware        : Marvell Armada-370
> Revision        : 0000
> Serial          : 0000000000000000
> ```
> 
> CPUs with vfpv3d16 are not supported by the MS published SDK. It's possible to compile a version that would support it I think but that may be a little too involved...
> 
> For now I think we need to mark armada370 as not supported for any net core spks. I don't know if the same would affect the other armada variants.

Looks like it might vbe arm v7 without neon? armv6 binary would run on it then afaik

ta264:
> Looks like it might vbe arm v7 without neon? armv6 binary would run on it then afaik

.NET Core doesn't support armv6.  So essentially Radarr is dropping support for this architecture after the release of 3.1 (hopefully next few weeks).  3.1 will run under mono on this architecture.

azz22:
Hi, sorry if this is not the right place to post but as its related thought i would.

I tried to upgrade in app tonight from 3.0.2.4552 to 3.2.0.5048 and it didn't work. The app went through the motions, backup, restarted but didn't update. Is this because its .NET and not mono? I've attached my log file for you to peruse.

appreciate your help in advance.
[radarr.txt](https://github.com/SynoCommunity/spksrc/files/6541923/radarr.txt)

bakerboy448:
Info logs are typically useful for debugging.

for this: stop sonarr, update radarr, start sonarr has typically been doing it.  Synology has something weird going on with DSM

PR #4465, opened 2021-03-01
Remove dependency of sqlite on speex (an audio codec)

ta264:
_Motivation:_  Pretty sure sqlite shouldn't depend on speex.  Speeds up build of packages depending on sqlite a little.
_Linked issues:_  

### Checklist


publicarray:
Thanks, We should probably document out CI. It builds all packages where files where touched and packages that depend on it. It's  also why it takes so long. We had errors show up in when a dependency was updated that would make other packages fail (python etc.) Here is's only ffsync that failed. It probably failed before we had the GitHub Action and hasn't been maintained for some time, so we can safely ignore it. If you look at the logs you can see all other packages are fine.

publicarray (approved):

publicarray:
Good catch! It's been there for 9 years!

PR #4489, opened 2021-03-11
fix builds for DSM<7 without wizard files

hgy59:
_Motivation:_  Build of spk packages without wizard fail for DSM < 7
_Linked issues:_  Followup of #4395


publicarray:
Thanks!

hgy59:
@publicarray unfortunately this introduced another error.
The wizard files are not packed into the spk now. The wizard files are added to the spk file if you run make two or more times.

The expresson
`ifneq ($(wildcard $(DSM_WIZARDS_DIR)/.*),)`

evaluates to false at the first time.
I tried to get this evaluation fixed, but no success for now.
Any hint on conditionally execution of this block would be helpful.

hgy59:
⚠️ this has broken all packages with wizards on the master.

PR #4491, opened 2021-03-12
GitHub Action: Added the ability to publish to synocommunity

publicarray:
_Motivation:_  Open the conversation again to enable the packages to be published via GitHub actions
_Linked issues:_  https://github.com/SynoCommunity/spksrc/pull/4459#issuecomment-797304855. There was another discussion somewhere, but I can't find it.

This would only apply on a GitHub tag like `git tag -a -s -m jackett-0.17.668 jackett-0.17.668` feel free to come up with a better way to initiate a publish action.

~~I disabled the GitHub release when the API key is set because I think it's unnecessary, but I don't feel indifferent about it if there is a use case for having them hosted on GitHub also.~~

### DIFF
--- a/.github/actions/build-tag.sh
+++ b/.github/actions/build-tag.sh
@@ -14,6 +14,6 @@ echo "$GH_PACKAGE"
 make TCVERSION=${GH_ARCH##*-} ARCH=${GH_ARCH%%-*} -C spk/${GH_PACKAGE}
 
 # publish to synocommunity.com when the API key is set
-if [-n $API_KEY ]; then
+if [ -n "${API_KEY}" ]; then
     make TCVERSION=${GH_ARCH##*-} ARCH=${GH_ARCH%%-*} publish -C spk/${GH_PACKAGE}
 fi
--- a/.github/actions/build-tag.sh
+++ b/.github/actions/build-tag.sh
@@ -12,3 +12,8 @@ echo "$GH_PACKAGE"
 
 # use TCVERSION and ARCH parameters to get real exit code.
 make TCVERSION=${GH_ARCH##*-} ARCH=${GH_ARCH%%-*} -C spk/${GH_PACKAGE}
+
+# publish to synocommunity.com when the API key is set
+if [-n $API_KEY ]; then
+    make TCVERSION=${GH_ARCH##*-} ARCH=${GH_ARCH%%-*} publish -C spk/${GH_PACKAGE}
+fi
--- a/.github/actions/build-tag.sh
+++ b/.github/actions/build-tag.sh
@@ -2,7 +2,7 @@
 
 make setup-synocommunity
 sed -i -e "s|#PARALLEL_MAKE=.*|PARALLEL_MAKE=4|" \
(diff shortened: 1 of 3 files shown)


publicarray:
One thing to consider is that people that have github push permissions can now also publish

ymartin59:
@publicarray I would like to stress that people with "publish API keys" are supposed to be responsible for testing and support after publish... spkrepo also keep track about "who" has published package... And they are also supposed to do some regular cleanup to prevent storage leak and "out of disk space" errors... typically discarding package versions that are known to be broken and replaced with bug fixes or of course out-dated.

May you please document in wiki how a maintainer is supposed to setup his API key as secret?

hgy59:
> May you please document in wiki how a maintainer is supposed to setup his API key as secret?

And publishing is not enough. To appear in the DSM package center the packages must be activated after upload.

ymartin59:
@hgy59 I agree. But as they already appear as versions in package view, it may help to deliver preview and request feedback from interested users... but at least it will make deliver more fluent.

ymartin59:
In my opinion we should document a "publisher chart" in addition to a "contribution chart" to list expectations to achieve some quality goals.

publicarray:
Hi @ymartin59 thanks for the feedback.

> @publicarray I would like to stress that people with "publish API keys" are supposed to be responsible for testing and support after publish... spkrepo also keep track about "who" has published package...

Hm true, maybe this should only be used on a contributor's fork then. (not directly here)

> And they are also supposed to do some regular cleanup to prevent storage leak and "out of disk space" errors... typically discarding package versions that are known to be broken and replaced with bug fixes or of course out-dated.

How is that currently possible, I don't think that action is available in the admin interface?

The benefit here is that it for packages that take long time to build where or where I want to use a dev machines for other tasks. Sure we still need to build one arch & test locally and activate a package, But the cross build process can be offloaded (ffmpeg comes to mind). I think it's also good from a security perspective. It's more verifiable that the code that is build is the one on github, no local modification on a dev machine is possible (not that anyone has done that)

Sure, I'll add instructions it to the wiki, I think https://github.com/SynoCommunity/spksrc/wiki/Update-Policy has been my goto for the publish process. But yea :+1:  on updating the wiki

publicarray:
I've released radarr using this (found and fixed a bug), and I've added the necessary documentation: https://github.com/SynoCommunity/spksrc/wiki/Publish-a-package-from-Github

publicarray:
@ymartin59 how do i remove a package version?

ymartin59:
@publicarray Have you been able to use "dustbin" icon in Version view? or else you are lacking permissions. If so, please provide me a list of package+version to clean up and I will do.

ymartin59 (approved):
I am OK with this, please provide some usage documentation in wiki in a developer specific page

publicarray:
With Githubs fairly new [`workflow_dispatch`](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) and merging of #4810 this PR is obsolete. Tags are no longer  required, just the `gh` cli or github.com website.

 How to setup your own fork: https://github.com/SynoCommunity/spksrc/wiki/Publish-a-package-from-Github

PR #4495, opened 2021-03-14
Firefox Sync fails to build. Move to DIY until it's fixed

publicarray:
Currently, fails with 
```
===>  Compiling for gevent
[...]
configure: CFLAGS error: CFLAGS may only be used to specify C compiler flags, not include directories. Use CPPFLAGS for: -I/spksrc/spk/ffsync/work-x64-6.1/install//usr/local/ffsync/include
```

I tired setting various variables, with no avail. 

Even if the package compiles it needs an overhaul anyway.

_Linked issues:_  #4465 https://github.com/SynoCommunity/spksrc/issues?q=firefox+sync


hgy59:
as this is a long work in progress you might not belive it, but just yesterday I reverted the python dependency to python2 on #3849 and it compiles again (not yet pushed).
By now the mysql/mariadb initialization still not works, but compilation is file.

I fear that moving this package on the master will result in merge conflicts for mergin and will make it hard to rebase #3849 with updated master.

I know you would like to see ✔️ on build actions.
Maybe there other solutions to skip failing builds there like specific Makefile variable.

publicarray:
Hey thanks, I should have looked at open PRs first...

Agreed let's close this. I wasn't aware that there was anyone working on it.

> Maybe there other solutions to skip failing builds there like specific Makefile variable.

Hm I'm not a fan of this idea. If there are failing builds it means there is work todo but once they are ignored we might forget about them. I rather have a red :x: than blissfully ignoring broken packages IMHO.

I like :heavy_check_mark: too, but for newcomers they are confused why their PR fails for no apparent reason. Eventually I like to propose that PRs need to pass CI before merging.

PR #4496, opened 2021-03-14
fix wizard creation

hgy59:
_Motivation:_  Finally fix wizard creation broken by #4395
_Linked issues:_  fixes #4489

### Remarks
- Fix missing wizards in packages introduced with #4489
- This is a working solution for packages without wizards as expected by #4489
- The solution of #4489 did not work, as the copied files where not detected with `ifneq ($(wildcard $(DSM_WIZARDS_DIR)/.*),)` and therefore not included in the package. (but a second build without `make clean` created the package with the wizard files - so it was hard to detect the error).

The solution uses duplicated statements as there is no solution found for logical OR conditions.


publicarray:
My attempt also has publication #4498

PR #4497, opened 2021-03-14
Add db creation to demoservice

hgy59:
_Motivation:_ Demonstrate the creation of MySQL Database with resource configuration.
_Linked issues:_  

### Checklist

### Remarks
- add demo of creating a mysql db by resource configuration
- add uninstall wizard to optionally drop the database and/or the db user


ymartin59:
@hgy59 Would it be possible to also produce a database dump from a wizard checkbox before deletion? probably requesting for a dump destination folder, using an obvious file name like "mysql_schema_timestamp.dump.gz" for instance.

hgy59:
> 
> 
> @hgy59 Would it be possible to also produce a database dump from a wizard checkbox before deletion? probably requesting for a dump destination folder, using an obvious file name like "mysql_schema_timestamp.dump.gz" for instance.

As I am currently working on ffsync this would be a logical extension. 
Another task is to test this with DSM7.

DigitalBox98:
For DSM7 if you need testers let me know. 
I'm currently moving/testing some packages to this new DSM.

spk/demoservice/src/conf/resource
```diff
@@ -0,0 +1,17 @@
+{
+  "mysql-db": {
```
publicarray, line 2:
```suggestion
  "mysql-db": {
  "mariadb10-db" {
```

I contacted Synology a while back, and they gave me updated documentation. I assumed it was already updated online as well. [developer-guide-ex.pdf](https://github.com/SynoCommunity/spksrc/files/6164396/developer-guide-ex.pdf)

Also see 

https://github.com/SynoCommunity/spksrc/blob/dsm7-packages/spk/wallabag/src/conf/resource

https://github.com/SynoCommunity/spksrc/wiki/DSM-7.0-Beta#package-migration

publicarray (changes requested):

spk/demoservice/src/conf/resource
```diff
@@ -0,0 +1,17 @@
...
+    "info": {
+      "db-name": "demoservice"
+    },
+    "root-pw": "{{wizard_mysql_password_root}}",
+    "create-db" : {
+      "db-collision": "skip"
+    },
+    "grant-user": {
+      "user-name" : "demoservice",
+      "rand-pw" : true
```
publicarray, line 12: This is not implemented unfortunately

publicarray: Did this work in DSM 6 ?

hgy59: > 
> 
> Did this work in DSM 6 ?

Yes it works, and the random password is written into `demoservice.ini` in this demo package at installation time.

publicarray:
Hm so we have to create different json for DSM6 or DSM7...

You might like this bit, for removing the db only when the user wants to: https://github.com/SynoCommunity/spksrc/blob/eea72307f2c0bdaf0217e5ecba90b94476ae6eb1/spk/wallabag/src/installer.sh#L66-L75

spk/demoservice/src/conf/resource
publicarray: 👍

hgy59:
> You might like this bit, for removing the db only when the user wants to:

The current code has already implemented this behaviour, but does not need any code like you propose. The variables are already in the wizard.
And you cannot only drop the db, there is a separate option (and wizard check box) to drop the user.

And AFAICS this is compatible with DSM 7.

publicarray:
Thanks @hgy59. From memory it did not work correctly on DSM7 (hence my workaround) but maybe it will be fixed before DSM7 is released. Woud be possable to have an example that works on both DSM6 and DSM7? Otherwise good job.

publicarray (approved):

publicarray:
The additional changes can be added in a follow-up PR

hgy59:
This pull request is too far from the current master branch; a new attempt is required.

Issue #2484, opened 2016-10-08
[Package Request] Music Player Daemon (MPD)


clauseggers:
Name: Music Player Daemon (MPD)
Description: Server-side network controlled music player daemon.
Website: https://www.musicpd.org
Software documentation: https://www.musicpd.org/doc/user/
Build/Installation documentation: https://www.musicpd.org/doc/user/install_source.html
Source code: http://git.musicpd.org/cgit/master/mpd.git/ alt. mirror https://github.com/MaxKellermann/MPD 
License: GNU General Public License

vkostas:
FYI, I am currently working on an amended version of MPD.

hgy59:
@vkostas are you still working on this?
@Safihre is about to remove the cross/mpd with #3173

vkostas:
@hgy59  I had stopped like one year ago (still have the packages) due to many unsolved dependencies + DSM 6 user issue (haven't checked yet if definitely solved). Added a msg to #3173.

hgy59:
initial mpd package released on 2022-06-06

PR #4502, opened 2021-03-16
Propose MPD as a package and fix audiofile for latest toolchains

DigitalBox98:
_Motivation:_  Propose the Music Player Daemon as a standalone package

### Checklist

A version for DSM 7 was tested successfully (ie daemon launched) once the mpd.conf was setup correctly and ".mpd" folder created.
Possible enhancements : move to latest MPD version and provide a start/stop via DSM


cross/mpd/PLIST
```diff
@@ -1 +1,3 @@
 bin:bin/mpd
+rsc:share/doc/mpd/mpdconf.example
```
publicarray, line 2: If you want to include a configuration file I propose to create a working config file instead, if possible. We don't expect most users to use the command line, if there are options that need to be selected by a user you can ask the user at install time by using the wizard variables. You can use the transmission package as an example: https://github.com/SynoCommunity/spksrc/tree/master/spk/transmission/src/wizard

cross/mpd/PLIST
```diff
@@ -1 +1,3 @@
...
+rsc:share/icons/hicolor/scalable/apps/mpd.svg
```
publicarray, line 3: Why include the logo?

publicarray:
fixes #2484

DigitalBox98:
Ok I have fixed the point raised and checking on DSM 7, MPD is running 100% fine ! (on braswell)
For DSM 6, it needs to be confirmed but the spk creation is fine.

Some remarks or points to improve : 
- For DSM 7, I was able to compile with ffmpeg allowing to recognize additional formats. Dor DSM 6 I was unable to cross compile MPD adding the cross/ffmpeg dependency so I have removed it (to be investigated).
- This is not the last version but a more recent version that the one on SPKSRC :) I wasn't able to compile the last version (to be investigated as well)
- In the current setup (ie mpd.conf), MPD is launching its internal HTTP server. But it's possible also to let IceCast deliver the audio flow (testing OK on DSM 7) : ie audio_output type "cast" in the configuration file.
- The audio flow will be effective only if an MPD client is used with MPD server (see www.musicpd.org for details on MPD clients)
- The status of the MPD server was not recognized well in DSM 7 (I had to do some manual change in the start/stop script), so to be checked if on DSM 6 it's the same
- The user "sc-mpd" must have "read" writes on the music folder defined during installation process, else MPD server will not be able to read the audio file

hgy59:
>  * This is not the last version but a more recent version that the one on SPKSRC :) I wasn't able to compile the last version (to be investigated as well)

- MPD 20.x is the latest version that is compatible with DSM6
- MPD 21.x requires GCC 6 and might be built for DSM7, as DSM7 toolchains come with gcc 7.30
- MPD 22.x is not compatible with any DSM toolchain as it requires GCC 8

cross/libaom/Makefile
```diff
@@ -24,6 +24,12 @@ PKG_WORK_DIR = $(WORK_DIR)/$(PKG_NAME)_$(PKG_VERS)-build
 CMAKE_USE_NASM = 1
 CMAKE_USE_NINJA = 1
 
+include ../../mk/spksrc.common.mk
+ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
+# fix compilation with newer compilers
+CMAKE_ARGS += -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_C99_MATH"
+endif
```
th0ma7, line 31: This is not needed as already in the current code on master
Ref line 35+ https://github.com/SynoCommunity/spksrc/blob/master/cross/libaom/Makefile

hgy59: @th0ma7 thanks for the hint, this saves me the time of digging into this...

PR #4514, opened 2021-03-22
Update adminer

hgy59:
_Motivation:_  This is taken from dsm7_packages branch with the latest update
_Linked issues:_  

### Checklist

### Remarks
- this is now a noarch build for three different DSM versions (DSM <6.0, DSM >=6.0;<7.0; DSM >=7.0)


spk/adminer/src/conf/resource
```diff
@@ -0,0 +1,48 @@
...
+        "display_name": "Adminer",
+        "type": "apache_php",
+        "root": "adminer",
+        "backend": 2,
+        "icon": "app/images/adminer-{0}.png",
+        "php": {
+          "profile_name": "adminer Profile",
+          "profile_desc": "PHP Profile for adminer",
+          "backend": 8,
+          "open_basedir": "/var/services/web_packages/adminer:/tmp:/var/services/tmp",
```
publicarray, line 15: I highly recommend removing `open_basedir`. I might found a bug in the DSM7 where the first package that sets the open_basedir will also wrongly apply for any package installed after it. I'm in contact with Synology but their communication is slow... Since the DSM7 is further locked down I don't see a security issue not having it. but would be nice if it worked correctly.

spk/adminer/Makefile
```diff
@@ -21,22 +20,37 @@ INSTALLER_SCRIPT = src/installer.sh
...
+## make && make TCVERSION=6.0 && make TCVERSION=7.0 
+##
+
+ifeq ($(call version_ge, ${TCVERSION}, 6.0),1)
 SPK_DEPENDS = "WebStation>=2.1.0:PHP7.4>=7.4.0:Apache2.4>=2.4.0"
+CONF_DIR = src/conf
+APACHE_VERSION = 2.4
+else
+APACHE_VERSION = 2.2
+endif
```
publicarray, line 37: Doesn't `SPK_DEPENDS` apply for the else statement for `Apache2.2`? or was it not implemented in DSM5?

publicarray (changes requested):

spk/adminer/src/conf/resource
publicarray: <pre>
AH01071: Got error 'PHP message: PHP Warning: Unknown: open_basedir** restriction in effect. 
File(/volume1/web_packages/<b>wallabag</b>/web/app.php) is not within the allowed path(s): 
(/var/services/web_packages/<b>adminer</b>:/tmp:/var/services/tmp) in Unknown on line 0
PHP message: PHP Warning: Unknown: failed to open stream: Operation not permitted in Unknown on line 0
Unable to open primary script: /var/services/web_packages/wallabag/web/app.php (Operation not permitted)'
</pre>

spk/adminer/Makefile
hgy59: DSM5 has apache integrated into DSM and not as installable packages.

spk/adminer/src/conf/resource
hgy59: @publicarray thanks 👍

spk/adminer/Makefile
publicarray: Ok Thanks,

publicarray (approved):

spk/adminer/Makefile
```diff
@@ -21,22 +20,37 @@ INSTALLER_SCRIPT = src/installer.sh
 SSS_SCRIPT = src/dsm-control.sh
 WIZARDS_DIR = src/wizard/
 
+include ../../mk/spksrc.common.mk
+
+##
+## This package can be built for DSM <6.0, DSM >=6.0;<7.0 and DSM >=7.0
+## To bild all versions call:
```
publicarray, line 27:
```suggestion
## To build all versions call:
```

Issue #4482, opened 2021-03-10
[Package Request]  Generic WIFI driver


JohnAllijn:
Name: Wifi driver 
Description: The currrent DSM version does not support wifi anymore. Sometimes wifi is very useful. For example when you’re setting up a temporary file storage or a demo. 
Website: n/a
Software documentation: n/a
Build/Installation documentation:
https://www.reddit.com/r/synology/comments/jdvmh2/making_a_wifi_usb_dongle_work_on_dsm_62/
And:
https://help.synology.com/developer-guide/create_package/compile_kernel_module.html

Source Code:
License: open source (inux)

th0ma7:
There could be a relatively easy build process using changes to the spksrc framework pending the merge of my PR #4420 and #4144 are merged.  I'll add this to my todo list if you're both patient enough and willing to test bed a few packages.

JohnAllijn:
Hi, That would be great! And of course I'll test :)
Just let me know how I can help.

th0ma7:
@JohnAllijn btw, note that Synology Linux kernel sources for 6.2.4 and 7.0 are not yet released.
I can definitively build modules for 6.2.3 DSM version but nothing after that until Synology releases the source code on SourceForge.net.  So heads-up, if you're interested in this you'll have to stick on 6.2.3 for now (and if already upgrade then you'll have to wait a bit longer).

Also, reading about it seems to be quite easy to do.  Will look into this shortly.

JohnAllijn:
Thanks! I’ll turn off auto updates for now.

Verstuurd vanaf mijn iPad

th0ma7:
Hi @JohnAllijn got to a point of an experimental driver package.
What NAS model, arch and exact version are you using?
Also, what is the USB wireless device you want to use?

JohnAllijn:
Hi vincent
I have a DS218+. 
i updated it to the latest version of dsm a few days ago. We are remoddeling our livingroom and do not have internet until wednesday. I”ll check the version then. 
What do you mean with Arch?

I plan on using this adapter because it should be supported by most OS’s:

https://a.aliexpress.com/_v4fNQf

Https://www.realtek.com/en/component/zoo/category/rtl8811cu-software
Feature:
Dual Band: Alternatieve 11AC(5.8G) en 11N(2.4G) aansluitingen voor TV shows, HD films, streaming en online gaming te voorkomen draadloze interferentie.
[...]

PR #4420, opened 2021-01-31
synocli-kernel: tools + usbserial + usbwifi kernel modules

th0ma7:
_Motivation:_  Design a method to easily build and package kernel modules using `usbserial` as a proof of concept
_Linked issues:_  #3787

### Checklist
- [x] Build rule all archs that has a supported kernel - completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

### TODO
- [x] use "true" or "false" values from the module loading scripts
- [x] find out why it fails to build correctly on a few platforms, errors varies (found, related to `LDFLAGS`)
- [x] using lsof manage to confirm the kernel module device file descriptor to check if someone is using it before unloading. If someone is, stop that process (for USB DVB dongles with tvheadend usage) - THIS IS BUILT INTO #4144 specifically for LinuxTV instead of being generic to synocli-kernel package.

## synokernel-usbwifi
Only compatible with 4.4.59+ kernels.
```
make -j6 arch-apollolake-6.2.3 arch-broadwellnk-6.2.3 arch-denverton-6.2.3 arch-geminilake-6.2.3 arch-purley-6.2.3 arch-v1000-6.2.3
```
- [x] `apollolake` (DSM-6.2.3 - 4.4.59)
- [x] `broadwellnk` (DSM-6.2.3 - 4.4.59)
- [x] `denverton` (DSM-6.2.3 - 4.4.59)
- [x] `geminilake` (DSM-6.2.3 - 4.4.59)
- [x] `purley` (DSM-6.2.3 - 4.4.59)
- [x] `v1000` (DSM-6.2.3 - 4.4.59)
---
```
make -j2 arch-armada37xx-6.2.3 arch-rtd1296-6.2.3
```
- [x] `armada37xx` (DSM-6.2.3 - 4.4.59)
- [x] `rtd1296` (DSM-6.2.3 - 4.4.59)

### Incompatible kernels & archs
- [ ] 2.6.32 (DSM-6.2.3) - 88f6281, qoriq

### Successful builds
```
make -j3 arch-dakota-1.2 arch-ipq806x-1.2 arch-northstarplus-1.2
```
- [x] `dakota` (SRM-1.2 - 4.4.60)
- [x] `ipq806x` (SRM-1.2 - 4.4.60)
- [x] `northstarplus` (SRM-1.2 - 3.6.11)
---
```
[...]

PR #4515, opened 2021-03-23
toolchain: Add DSM-6.2.4 toolchains

th0ma7:
_Motivation:_  Adding latest toolchain.  Will be of use for upcoming kernel drivers packages
_Linked issues:_  #4482, #4420 and #4144

### Checklist

### Notes
Not manually verified but expecting no changes to:
```
TC_KERNEL = 2.6.32
TC_GCC = 4.6.4
TC_GLIBC = 2.15
```


ymartin59 (approved):

hgy59 (changes requested):
As the toolchain makefiles do not contain the build number, you have to update `mk/spksrc.tc-vers.mk` too.
AFAIK the build number for 6.2.4 is 25556.

So you have to add the following code there:
```

ifeq ($(TC_VERS),6.2.4)
TC_BUILD = 25556
endif
```

th0ma7:
> As the toolchain makefiles do not contain the build number, you have to update `mk/spksrc.tc-vers.mk` too.
> AFAIK the build number for 6.2.4 is 25556.
> 
> So you have to add the following code there:
> 
> ```
> 
> ifeq ($(TC_VERS),6.2.4)
> TC_BUILD = 25556
> endif
> ```

Thnx for the reminder... as I should have known as it's I who added that previously, shrug..

PR #4522, opened 2021-03-26
memcached: upgrade 1.6.9 + DSM7 support

smaarn:
_Motivation:_  Being able to use memcache and (while we're at it) use the latest version
_Linked issues:_  None that I know of

### Checklist


smaarn:
Binaries can be found here: https://github.com/smaarn/spksrc/releases/tag/memcache-1-6-9

ymartin59 (approved):

ymartin59:
@hgy59 Hello. May you please help with merge and publish this update?

spk/memcached/src/service-setup.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
+
+# Package
+PACKAGE="memcached"
+
+# Others
+INSTALL_DIR="/var/packages/${PACKAGE}/target/"
```
hgy59, line 7: please use system provided `SYNOPKG_PKGDEST` variable in this script (`INSTALL_DIR` is the very old way we did it in install.sh scripts).

dito for `PACKAGE` variable. If you really need the plain package name, please  use `SYNOPKG_PKGNAME` instead.

hgy59:
@smaarn can you please remove cross/busybox dependency.
This was the old DSM 5 way in former installer.sh scripts to create system users (and to delete those when the generic service installer was introduced).

I don't know whether this package will be running on DSM 7 (as the root user is not available anymore to install and update packages or start/stop services).
- the current `service_postinst` function will fail on DSM 7.
- at least for DSM 7 we need to define a resource worker for the installation of a web service (you find one example in spk/adminer/src/conf/resource)

smaarn:
Flagging this PR as draft until DSM7 related logics are implemented

hgy59:
for `cross/memcached/Makefile` I am pretty sure that 
```
CONFIGURE_ARGS = ac_cv_c_endian=little
```

must be different for `PPC_ARCHS` as those are big endian cpu architecture.

hgy59 (changes requested):
The following commands in `service-setup.sh` are not allowed on DSM 7  (and variable `CONFIG_FILE` must be assigned before)
```
      cp -pv "${CONFIG_DIR}/Memcache.sample.php" "${CONFIG_FILE}"
      chgrp http "${CONFIG_FILE}"
      chmod g+w "${CONFIG_FILE}"
```

and those  are the errors in the installation log:
```
2022/05/03 20:18:58     Begin service_postinst
2022/05/03 20:18:58     cp: failed to access '/var/services/web/phpMemcachedAdmin/Config/Memcache.php': Permission denied
2022/05/03 20:18:58     chgrp: cannot access '': No such file or directory
2022/05/03 20:18:58     chmod: cannot access '': No such file or directory
2022/05/03 20:18:58     End service_postinst
```

smaarn:
> The following commands in `service-setup.sh` are not allowed on DSM 7 (and variable `CONFIG_FILE` must be assigned before)
> 
> ```
>       cp -pv "${CONFIG_DIR}/Memcache.sample.php" "${CONFIG_FILE}"
>       chgrp http "${CONFIG_FILE}"
>       chmod g+w "${CONFIG_FILE}"
> ```
> 
> and those are the errors in the installation log:
> 
> ```
> 2022/05/03 20:18:58     Begin service_postinst
> 2022/05/03 20:18:58     cp: failed to access '/var/services/web/phpMemcachedAdmin/Config/Memcache.php': Permission denied
> 2022/05/03 20:18:58     chgrp: cannot access '': No such file or directory
> 2022/05/03 20:18:58     chmod: cannot access '': No such file or directory
> 2022/05/03 20:18:58     End service_postinst
> ```

And I was wondering why upgrades weren't working... Thanks for pointing it out.

smaarn:
Rebased at head.

All reported issues are now fixed AFAICT.

hgy59:
@smaarn my review ended up in da8abee6303dea48cc5595993b039c92de96efab

The remaining issue is that the memcached icon in DSM UI links to the service port instead of phpMemcachedAdmin.

mk/spksrc.service.mk
```diff
@@ -7,7 +7,7 @@
 #   conf/SPK_NAME.sc       if SERVICE_PORT and DSM<7
 #   conf/resource          if SERVICE_CERT or DSM7
 #   app/SPK_NAME.sc        if SERVICE_PORT and DSM7
-#   app/config             if DSM_UI_DIR
+#   app/config             if DSM_UI_DIR (may be overwritten by DSM_UI_CONFIG)
```
hgy59, line 10: @ymartin59, @publicarray some time ago we added the generation of `app/config` files.
For this package we do not want the generated `app/config` for the service but one for the included phpMemcachedAdmin web interface.

The app/config file is used with DSM < 7 only, as for DSM 7 the web UI is defined in the resources file and app/config seems to be ignored.

So a lot of packages still have a `src/app/config` file in the spk folder, and we will have to validate, whether those match the generated one. Matching config files could be removed and for different files we have to apply `DSM_UI_CONFIG` too.

hgy59 (approved):

hgy59:
Now my testing was successfuly with `aarch64-6.1` and `x64-7.0`.
But I used the default configuration only, and could not verify whether package updates keep the configuration.

hgy59:
@smaarn Did you have a change to test this package? _(IMO it is good to merge.)_

smaarn:
> @smaarn Did you have a change to test this package? _(IMO it is good to merge.)_
> 
> 
> 
> 

Will be having a look today. Seems all good to me TBH.

smaarn:
Seems that the upgrade process looses the phpMemcached configuration. Since it's not the memcached configuration in itself I'm guessing we can assume it's fine and good to merge.

I will try having a look at the php configuration to have it remain persistent next week.

hgy59:
> I will try having a look at the php configuration to have it remain persistent next week.

@smaarn did you find this issue with DSM 6 or DSM 7 (or both)?.

smaarn:
> > I will try having a look at the php configuration to have it remain persistent next week.
> 
> 
> 
> @smaarn did you find this issue with DSM 6 or DSM 7 (or both)?. 

DSM 7 only. TBH it may be a very simple thing to do since the logic is there for DSM 6

cross/phpmemcachedadmin/Makefile
```diff
@@ -1,24 +1,29 @@
...
+include ../../mk/spksrc.install-resources.mk
 
 .PHONY: phpmemcachedadmin_install
+# make a copy of the sample config as on DSM7 it is not possible to create this at installation time.
 phpmemcachedadmin_install:
-	mkdir -p $(STAGING_INSTALL_PREFIX)/share/phpMemcachedAdmin
-	tar -cf - -C $(WORK_DIR)/$(PKG_DIR) . | tar -xf - -C $(STAGING_INSTALL_PREFIX)/share/phpMemcachedAdmin
+	@mkdir -p $(STAGING_INSTALL_PREFIX)/share/$(PKG_NAME)
+	@tar -cf - -C $(WORK_DIR)/$(PKG_DIR) . | tar -xf - -C $(STAGING_INSTALL_PREFIX)/share/$(PKG_NAME)
+	@cp -f $(STAGING_INSTALL_PREFIX)/share/$(PKG_NAME)/Config/Memcache.sample.php $(STAGING_INSTALL_PREFIX)/share/$(PKG_NAME)/Config/Memcache.php
```
hgy59, line 29: @smaarn  This **is** the problem.
we **must** create `Memcached.php` on demand at installation time. Otherwise the package always uses the default config.

hgy59:
@smaarn it was not so easy to create the config on demand for DSM7 and add save/restore on package updates.
It actually required a fix in the generic installer for DSM 6.

Issue #4533, opened 2021-03-31
Change in behaviour - latest framework snapshot with DSM7 support


th0ma7:
### Setup
_Package Name:_ `synokernel-usbserial`
_Package Version:_ PR #4420 

_NAS Model:_ DS918+
_NAS Architecture:_ `apollolake`
_DSM version:_ 6.2.3 Update 3

### Expected behavior
Since I've rebased against master which now includes DSM7 preliminary support, my development packages behave differently and fail to start.  I've started investigating and didn't noticed much difference besides the permissions now being assigned to `sc-synokernel-usbserial` on files which wasn't the case before (probably expected on DSM7 but wasn't on DSM6) and not necessarely a blocker neither in theory.
```
$ ll -d /volume1/@appstore/synokernel-usbserial
drwxr-xr-x 1 sc-synokernel-usbserial synokernel-usbserial 38 Mar 30 20:08 /volume1/@appstore/synokernel-usbserial
```
Which makes me wonder if it isn't trying to run the service script as user  `sc-synokernel-usbserial` which will necessarily fail (loading kernel modules require `root` access).

#### Other logs
New install log:
```
Tue Mar 30 20:07:36 EDT 2021
===> Step preuninst. USER= GROUP= SHARE_PATH=
Tue Mar 30 20:07:36 EDT 2021
===> Step postuninst. USER= GROUP= SHARE_PATH=
Invoke service_postuninst
Tue Mar 30 20:08:06 EDT 2021
===> Step preinst. USER=synokernel-usbserial GROUP= SHARE_PATH=
Tue Mar 30 20:08:08 EDT 2021
===> Step postinst. USER=synokernel-usbserial GROUP= SHARE_PATH=
Invoke service_postinst
Granting 'sc-synokernel-usbserial' unix ownership on /volume1/@appstore/synokernel-usbserial/var
```
Old install log:
```
Tue Mar 30 20:06:06 EDT 2021
===> Step preinst. USER= GROUP= SHARE_PATH=
Tue Mar 30 20:06:07 EDT 2021
===> Step postinst. USER= GROUP= SHARE_PATH=
Invoke service_postinst
```

BenjV:
If you need root for this package you could add "run-as' root to the start action:
Example of a privilege file who runs all scripts as root:
```
{
	"defaults": {
		"run-as": "package"
	},
	"username": "sc-synokernel-usbserial",
	"ctrl-script": [{
		"action": "preinst",
[...]

PR #4535, opened 2021-03-31
DSM6: Keep previous behaviour and assume no file by default

th0ma7:
_Motivation:_  On DSM6 with latest update for DSM7 it creates, by default, a `conf/privilege` file while previous behavior did not.  Proposed patch revert back to this behavior on DSM <= 7.
_Linked issues:_  #4533 4533

### Checklist

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -228,24 +228,8 @@ $(DSM_CONF_DIR)/privilege:
 	@jq -n '."defaults" = {"run-as": "package"}' > $@
 	@$(MSG) "Creating $@"
 	@$(MSG) '(privilege) run-as: package'
-# DSM <= 6 and SERVICE_USER defined
-else ifneq ($(strip $(SPK_USER)),)
-ifeq ($(strip $(SERVICE_EXE)),)
-$(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege-installasroot
-	@$(dsm_script_copy)
-	@$(MSG) "(privilege) spksrc.service.privilege-installasroot"
-else
-$(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege-startasroot
-	@$(dsm_script_copy)
-	@$(MSG) "(privilege) spksrc.service.privilege-startasroot"
-endif
-else
-$(DSM_CONF_DIR)/privilege:
-	@$(MSG) "Creating $@"
-	@$(MSG) "(privilege) DSM <= 6 and SERVICE_USER undefined"
-endif
+	@$(MSG) "(privilege) DSM >= 7 $(DSM_CONF_DIR)/privilege"
 # Apply variables to privilege file
-ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 ifneq ($(strip $(GROUP)),)
 # Creates group but is different from the groups the user can create, they are invisible in the UI an are only usefull to access another packages permissions (ffmpeg comes to mind)
 # For DSM7 I recommend setting permissions for individual packages (System Internal User)
@@ -263,9 +247,21 @@ endif
 ifneq ($(strip $(SPK_GROUP)),)
 	@jq '."groupname" = "$(SPK_GROUP)"' $@ 1<>$@
 endif
+ifneq ($(findstring conf,$(SPK_CONTENT)),conf)
+SPK_CONTENT += conf
+endif
 
-# Less then DSM 7
+# DSM <= 6 and SERVICE_USER defined
+else ifneq ($(strip $(SPK_USER)),)
+ifeq ($(strip $(SERVICE_EXE)),)
+$(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege-installasroot
+	@$(dsm_script_copy)
+	@$(MSG) "(privilege) spksrc.service.privilege-installasroot"
 else
+$(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege-startasroot
+	@$(dsm_script_copy)
+	@$(MSG) "(privilege) spksrc.service.privilege-startasroot"
+endif
 ifneq ($(strip $(SYSTEM_GROUP)),)
 # options: http, system
 	@jq '."join-groupname" = "$(SYSTEM_GROUP)"' $@ 1<>$@
@@ -276,12 +272,15 @@ endif
 ifneq ($(strip $(SPK_GROUP)),)
 	@jq '."groupname" = "$(SPK_GROUP)"' $@ 1<>$@
 endif
-endif
-
-ifneq ("$(wildcard $(DSM_CONF_DIR))","")
 ifneq ($(findstring conf,$(SPK_CONTENT)),conf)
 SPK_CONTENT += conf
 endif
+
+# DSM <= 6 and SERVICE_USER is NOT defined
+else
+$(DSM_CONF_DIR)/privilege:
+	@$(MSG) "NOT creating $@"
+	@$(MSG) "(privilege) DSM <= 6 and SERVICE_USER undefined"
 endif
 
 # Call $(DSM_CONF_DIR)/privilege:
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -226,14 +226,16 @@ ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 $(DSM_CONF_DIR)/privilege:
 	$(create_target_dir)
 	@jq -n '."defaults" = {"run-as": "package"}' > $@
-else ifeq ($(strip $(SERVICE_EXE)),)
+# If DSM <= 6 and privilege file exists
+else ifneq ("$(wildcard $(DSM_CONF_DIR)/privilege)","")
+ifeq ($(strip $(SERVICE_EXE)),)
 $(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege-installasroot
 	@$(dsm_script_copy)
 else
 $(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege-startasroot
 	@$(dsm_script_copy)
 endif
-
+endif
 
 # Apply variables to privilege file
 ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
@@ -244,6 +246,9 @@ ifneq ($(strip $(GROUP)),)
 	@jq --arg packagename $(GROUP) '."join-pkg-groupnames" += [{$$packagename}]' $@ 1<>$@
 endif
 endif
+
+# If privilege file exists
+ifneq ("$(wildcard $(DSM_CONF_DIR)/privilege)","")
 ifneq ($(strip $(SYSTEM_GROUP)),)
 # options: http, system
 	@jq '."join-groupname" = "$(SYSTEM_GROUP)"' $@ 1<>$@
@@ -262,7 +267,7 @@ ifneq ($(findstring conf,$(SPK_CONTENT)),conf)
 SPK_CONTENT += conf
 endif
 SERVICE_FILES += $(DSM_CONF_DIR)/privilege
-
+endif
 
 # Generate service configuration for admin port
 ifeq ($(strip $(FWPORTS)),)
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -278,7 +278,7 @@ ifneq ($(strip $(SPK_GROUP)),)
 endif
 endif
 
-ifneq ("$(wildcard $(DSM_CONF_DIR)/privilege)","")
+ifneq ("$(wildcard $(DSM_CONF_DIR))","")
 ifneq ($(findstring conf,$(SPK_CONTENT)),conf)
 SPK_CONTENT += conf
 endif
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -257,6 +257,8 @@ endif
 ifneq ($(strip $(SPK_GROUP)),)
 	@jq '."groupname" = "$(SPK_GROUP)"' $@ 1<>$@
 endif
+
+# Less then DSM 7
 else
 ifneq ($(strip $(SYSTEM_GROUP)),)
 # options: http, system
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -53,10 +53,10 @@ pre_service_target: service_msg_target
 # auto uses SPK_NAME
 ifeq ($(SERVICE_USER),auto)
 SPK_USER = $(SPK_NAME)
-else
+else ifneq ($(strip $(SERVICE_USER)),)
 SPK_USER = $(SERVICE_USER)
-endif
-ifeq ($(strip $(SPK_USER)),)
+# else always enforce setting SPK_USER on DSM >= 7
+else ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 SPK_USER = $(SPK_NAME)
 endif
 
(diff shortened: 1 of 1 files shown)


th0ma7:
@BenjV this should fix the issue on DSM6.  Worth checking if it doesn't break anything on DSM7.

publicarray:
Hi Thanks, unfortunately this won't work for the DSM 7.0. There is no privilege file generated when I compile adminer `make TCVERSION=7.0`

publicarray (changes requested):
Maybe a better option is to change the default for DSM 6? Or make sure that the privilege file is still auto generated for DSM 7

th0ma7:
@publicarray it sounded like too easy of a change as it worked off the bat for my use case :)  I'll have to further look at the code (now that I know how to reproduce on DSM7).

Feel free to chim-in and propose a different approach.  Otherwise I'll try to find time over the week-end to make this work in a way or another.  I'm thinking of doing what's specific to DSM7 first (leading to generation of the `privilege` file), then apply what's common to all DSM if `privilege` file exists.

publicarray:
Did you have a look at the `mk/spksrc.service.privilege-startasroot` file will that do what you need?

publicarray:
@BenjV has a different approach: https://github.com/SynoCommunity/spksrc/issues/4533#issuecomment-811730282

th0ma7:
I saw that, great info btw, but it doesn't solve the issue.
My point is (and I'll happily align with the group decision on this), from a framework perspective this is a **regression**.

The behavior changed over existing packages DSM <= 6 leading to inability to rebuild to identical the packages (for those who do not already have a `privilege` file).  I believe prior behavior must be kept for pre-DSM7 packages to avoid unforeseen problems down the road.

I currently have an eye on:
1. package upgrades switching from `root` to `appuser` where backup file permissions are mistakenly following as root (need to migrate `mv` to `cp` & `rm /tmp/backupFile` or something else was already thought of?).
2. `/tmp/log*` files owned by root leading to package startup breakage as log output cannot be written to as `appuser`.

My 2 cents.

BenjV:
@th0ma7 
Did you read my last post?

With solution 2 nothing has to change and there is no regression from DMS 6
Only for the packages that needs root you have to change the service setup script and remove te su command, the it will runas root.

th0ma7:
@ BenjV 
> Only for the packages that needs root you have to change the service setup script and remove the su command, then it will run as root.

Incidentally, if you don't do that change to the packages, and publish an update for DSM6, this will lead to issues for our user base.  I may be blunt but I would call that a **regression**.

Unless I don't get this right, this mean that no DSM6 packages can be published without additional checks as its behavior **is** impacted if the package needs `root`.

BenjV:
@th0ma7 
I think you are wrong.

If the start option in the privilege file has the "run as root option" then all packages will be run as root just like the current DSM 6 situation without a privilege file.

I agree that is not the best option (package should be "run as package"), but to stay compatible with the current situation it is for sure the best option.

th0ma7:
@BenjV and @hgy59 here is an updated proposal which resume previous behavior in the following conditions:
- DSM <= 7
- `SERVICE_USER` is undefined
- `privilege` file non-exitant

Seems to be working out just fine but feel free to disregard if you disagree with proposed approach.  This should allow rebuilding existing packages to identical during the framework "migration" period...

A completely different approach (which I would propose we do after this PR is merged) is the following:
1. remove `SERVICE_USER = auto` from every packages
2. add `SERVICE_USER = root` for packages which does not currently have one by default (where we see fit)
3. adapt the `spksrc.service.mk` code accordingly (which removes a few DSM7 version checks)

publicarray:
Thanks @th0ma7 that sounds like a reasonable approach.

mk/spksrc.service.mk
```diff
@@ -239,24 +242,37 @@ ifneq ($(strip $(GROUP)),)
...
 endif
 ifneq ($(strip $(SPK_USER)),)
 	@jq '."username" = "sc-$(SPK_USER)"' $@ 1<>$@
-ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 	@jq '."groupname" = "sc-$(SPK_USER)"' $@ 1<>$@
 endif
+ifneq ($(strip $(SPK_GROUP)),)
+	@jq '."groupname" = "$(SPK_GROUP)"' $@ 1<>$@
+endif
+else
```
publicarray, line 256: I suggest making the separation clearer:
```suggestion
# Less then DSM 7
else
```

publicarray:
<details>
<summary>41 Packages without SERVICE_USER</summary>
<pre>
$ rg -g 'spk/**/Makefile' --files-without-match SERVICE_USER --stats
spk/mosh/Makefile
spk/ruby/Makefile
spk/python38/Makefile
spk/python2/Makefile
spk/erlang/Makefile
spk/ffsync/Makefile
spk/zsh-static/Makefile
spk/stockfish/Makefile
spk/monitoring-plugins/Makefile
spk/lua/Makefile
spk/node-exporter/Makefile
spk/nodejs/Makefile
spk/hassio/Makefile
spk/fish/Makefile
spk/inotify-tools/Makefile
spk/mantisbt/Makefile
spk/haproxy/Makefile
spk/tt-rss/Makefile
spk/ejabberd/Makefile
spk/git/Makefile
spk/gnupg/Makefile
spk/squidguard/Makefile
spk/imagemagick/Makefile
spk/plowshare/Makefile
spk/zsh/Makefile
spk/curaengine/Makefile
spk/darkstat/Makefile
spk/synocli-monitor/Makefile
spk/plexivity/Makefile
spk/mercurial/Makefile
spk/synocli-file/Makefile
spk/jupp/Makefile
spk/htpcmanager/Makefile
spk/he853/Makefile
spk/monit/Makefile
spk/sslh/Makefile
spk/mutt/Makefile
spk/maraschino/Makefile
spk/fishnet/Makefile
spk/debian-chroot/Makefile
spk/shairport-sync/Makefile
spk/mtproxy/Makefile
spk/irssi/Makefile
spk/rdiff-backup/Makefile
spk/domoticz/Makefile
spk/mono/Makefile
spk/dante-sockd/Makefile
spk/octoprint/Makefile
spk/salt-master/Makefile
spk/tcl/Makefile
spk/memcached/Makefile
spk/comskip/Makefile
spk/owncloud/Makefile
spk/full-text-rss/Makefile
spk/syno-magnet/Makefile
spk/beets/Makefile
spk/git-server/Makefile
spk/rsnapshot/Makefile
spk/roundcube/Makefile
spk/horde/Makefile
spk/bicbucstriim/Makefile
spk/cops/Makefile
spk/synocli-disk/Makefile
spk/selfoss/Makefile
spk/saltpad/Makefile
spk/synocli-net/Makefile
spk/boxbackup-client/Makefile
spk/chromaprint/Makefile
spk/python3/Makefile
spk/fossil-scm/Makefile
spk/lirc/Makefile
spk/adminer/Makefile
spk/jappix/Makefile
spk/wallabag/Makefile
spk/gentoo-chroot/Makefile
spk/ffmpeg/Makefile
spk/borgbackup/Makefile
spk/duplicity/Makefile
spk/fengoffice/Makefile
spk/subliminal/Makefile
spk/mediainfo/Makefile
spk/vim/Makefile
spk/nzbmegasearch/Makefile

41 matches
41 matched lines
41 files contained matches
124 files searched
0 bytes printed
163091 bytes searched
0.000700 seconds spent searching
0.010392 seconds
</pre>
</details>

<details>
<summary>41 Packages with SERVICE_USER </summary>
<pre>
rg -g 'spk/**/Makefile' SERVICE_USER --no-heading --stats
spk/ntopng/Makefile:23:SERVICE_USER   = auto
spk/rabbitmq/Makefile:18:SERVICE_USER   = auto
spk/stunnel/Makefile:19:SERVICE_USER = auto
spk/salt-minion/Makefile:32:SERVICE_USER = auto
spk/minio/Makefile:22:SERVICE_USER = auto
spk/jackett/Makefile:19:SERVICE_USER = auto
spk/znc/Makefile:30:SERVICE_USER = auto
spk/sonarr/Makefile:25:SERVICE_USER = auto
spk/flexget/Makefile:24:SERVICE_USER = auto
spk/couchpotatoserver-custom/Makefile:22:SERVICE_USER = auto
spk/nzbget/Makefile:22:SERVICE_USER = auto
spk/sickbeard-custom/Makefile:20:SERVICE_USER = auto
spk/tvheadend/Makefile:35:SERVICE_USER = auto
spk/lidarr/Makefile:25:SERVICE_USER = auto
spk/transmission/Makefile:22:SERVICE_USER = auto
spk/icecast/Makefile:23:SERVICE_USER = auto
spk/ps3netsrv/Makefile:24:SERVICE_USER = auto
spk/sabnzbd/Makefile:28:SERVICE_USER = auto
spk/mylar/Makefile:20:SERVICE_USER = auto
spk/pyload/Makefile:22:SERVICE_USER   = auto
spk/radarr/Makefile:26:SERVICE_USER = auto
spk/itools/Makefile:27:# omit SERVICE_USER (this would overwrite the privilege file with a generic one).
spk/headphones/Makefile:21:SERVICE_USER = auto
spk/umurmur/Makefile:18:SERVICE_USER = auto
spk/couchpotatoserver/Makefile:23:SERVICE_USER = auto
spk/demoservice/Makefile:19:SERVICE_USER         = auto
spk/gateone/Makefile:22:SERVICE_USER = auto
spk/deluge/Makefile:32:SERVICE_USER = auto
spk/lazylibrarian/Makefile:19:SERVICE_USER = auto
spk/nzbget-testing/Makefile:24:SERVICE_USER = auto
spk/sickrage/Makefile:18:SERVICE_USER = auto
spk/sickchill/Makefile:22:SERVICE_USER = auto
spk/headphones-custom/Makefile:22:SERVICE_USER = auto
spk/redis/Makefile:24:SERVICE_USER = auto
spk/homeassistant/Makefile:36:SERVICE_USER = auto
spk/dnscrypt-proxy/Makefile:23:SERVICE_USER = auto
spk/nzbhydra/Makefile:20:SERVICE_USER = auto
spk/plexpy-custom/Makefile:22:SERVICE_USER = auto
spk/rutorrent/Makefile:26:SERVICE_USER = auto
spk/syncthing/Makefile:23:SERVICE_USER = auto
spk/mosquitto/Makefile:19:SERVICE_USER = auto

41 matches
41 matched lines
41 files contained matches
124 files searched
1971 bytes printed
163091 bytes searched
0.000884 seconds spent searching
0.012781 seconds

</pre>
</details>

th0ma7:
> Thanks @th0ma7 that sounds like a reasonable approach.

Thnx.

> Edit: Since `join-groupname` is duplicated maybe it can go after the if else?

I thought of that as well but I found it much simpler to keep totally distinct blocks for now until a subsequent PR to clean-up things further.

th0ma7:
@publicarray we really should consider adding `ripgrep` to the default image...

th0ma7:
I've adjusted output messages for clarity so we always know in what state we're in.
I believe its ready for a merge if ok with this.

publicarray:
Thanks, For some reason the `conf` folder is not copied to the DSM7 version of the package. :eyes:

mk/spksrc.service.mk
```diff
@@ -239,24 +248,39 @@ ifneq ($(strip $(GROUP)),)
...
+endif
+ifneq ($(strip $(SPK_USER)),)
+	@jq '."username" = "sc-$(SPK_USER)"' $@ 1<>$@
 endif
 ifneq ($(strip $(SPK_GROUP)),)
 	@jq '."groupname" = "$(SPK_GROUP)"' $@ 1<>$@
 endif
+endif
 
+ifneq ("$(wildcard $(DSM_CONF_DIR)/privilege)","")
```
publicarray, line 277: This line courses the `conf` folder to be missed inside the spk

publicarray (changes requested):

mk/spksrc.service.mk
publicarray: Would `ifneq (,$(wildcard $(DSM_CONF_DIR)/privilege))` still work for you?

th0ma7: I believe it would.
On the other hand, the `wildcard` may need to include `$(DSM_CONF_DIR)/*` or just `$(DSM_CONF_DIR)` so it checks if the directory exists or if there are other files in there and therefore add the `conf` directory if it is the case.  Testing only on the `privilege` file omits the existence of other files.  But enforcing to include the `conf` leads into build failures when it doesn't exists.

th0ma7: @publicarray I've changed this to only look for the `conf` directory, and if exists, set `SPK_CONTENT += conf`.
Let me know if that's ok with you.
From the tests I did on my end all looks ok but you might find another corner-case...

publicarray: Sorry but that did not work:

```sh
spksrc/packages/adminer_noarch-dsm7_4.7.6-3 on  dsm6-privilege-fix [📦🤷‍] 
❯ ls -l
.rw-r--r-- 1.2k seb 12 Apr  0:00 INFO
.rw-r--r-- 362k seb 12 Apr  0:00 package.tgz
.rw-r--r-- 4.0k seb 12 Apr  0:00 PACKAGE_ICON.PNG
.rw-r--r--  19k seb 12 Apr  0:00 PACKAGE_ICON_256.PNG
drwxr-xr-x    - seb 12 Apr  0:00 scripts
drwxr-xr-x    - seb 12 Apr  0:00 WIZARD_UIFILES

```

th0ma7: @publicarray You are absolutely right,  output led me to think all was good but it wasn't.
I ended-up reworking the code to group per use-case which led to removing `wildcard` call entirely.
Code is now much clearer than I had expected.
Let me know if that looks ok to you.

mk/spksrc.service.mk
```diff
@@ -222,41 +222,64 @@ ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
...
+	@$(MSG) "Creating $@"
+	@$(MSG) '(privilege) run-as: package'
+	@$(MSG) "(privilege) DSM >= 7 $(DSM_CONF_DIR)/privilege"
 # Apply variables to privilege file
-ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 ifneq ($(strip $(GROUP)),)
 # Creates group but is different from the groups the user can create, they are invisible in the UI an are only usefull to access another packages permissions (ffmpeg comes to mind)
 # For DSM7 I recommend setting permissions for individual packages (System Internal User)
 # or use the shared folder resource worker to add permissions, ask user from wizard see transmission package for an example
 	@jq --arg packagename $(GROUP) '."join-pkg-groupnames" += [{$$packagename}]' $@ 1<>$@
```
publicarray, line 233: No package is using it, yet maybe we can rename this?  (actually I don't know 100% what it is useful for) afaik the package can add another package here to allow it to access its files (I just don't remember which package needs to define it). DSM7 only.

publicarray: Actually I don't see it documented any more. Maybe it was only in the DSM7 preview?

publicarray:
This works nicely! Would you mind rebasing to master? Otherwise, I don't have any complains maybe just the possibility to remove some duplicate code. But I understand we also want to keep it clean, without 100 nested ifs.

mk/spksrc.service.mk
th0ma7: If I get this right, a good use-case that comes to mind would be for my `spksrc-kernel` toolkit where a `sc-kernel` group could be created and `synokernel-usbserial`, `synokernel-usbwifi`, `synokernel-linuxtv` could then make use of it?

If that is the case, I guess we should elaborate a little more and allow it on DSM6 somehow if possible to facilitate future migrations as I would use that right away.

th0ma7:
@publicarray now rebased.

mk/spksrc.service.mk
publicarray: It is implemented by using package names. Not groups. I know I fell for that one too.

th0ma7:
@publicarray Is this good for a merge?

publicarray:
Yea 👍

PR #4161, opened 2020-09-06
sickchill: initial version v2020.11.24-1

Nyaran:
_Motivation:_  Add the package SickChill. The latest version of SickChill is running with Python3 and also change the startup file, so the package "SickBeard Custom" is not valid to install SickChill

Build: [sickchill_noarch-all_20210313-1.spk](https://drive.google.com/file/d/1z4sWGcjL4AYL7sP90W6XR-RyT7se557j/view?usp=sharing) (Don't browse the package, use the download option on top-bar)

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4538, opened 2021-04-02
sickchill package updates

miigotu:
remove cross https://github.com/SynoCommunity/spksrc/pull/4161#issuecomment-809753885
set update_frequency to 24 hours https://github.com/SynoCommunity/spksrc/pull/4161#issuecomment-809768213
fix upgrade, force port to 8081 https://github.com/SickChill/SickChill/issues/7070

~~Removed the pure python forcing, because lxml, greenleaf, sqlalchemy, tornado, mako, and cryptography have c extensions
We can build without c extensions for most of them, except cryptography, but they all make the package run faster and with less resources if compiled.~~

@publicarray @ymartin59 Please let me know of any requested changes as soon as possible. Users of the existing package are having update issues because of the port. 
Please do not merge until tested.

Signed-off-by: miigotu <miigotu@gmail.com>

### Checklist


miigotu:
Please do not merge until this is tested.

publicarray (changes requested):
Ok when we wanted cross/sickchill removed it was because I was under the assumption that this was a pure python package. So we probably want to revert part of this change here (sorry)

Please make sure the source code for your package is included otherwise the package will fail to install.

Since @BenjV has a noarch package: https://github.com/BenjV/SYNO-packages/blob/master/SickChill%201.2.Python%203%20noarch.spk I'd like he's feedback. Is it worth cross compiling for a potential speedup? Has anyone noticed a performance issue with the noarch version? I'm asking because cross compiling takes significantly more time for me to publish with seemingly little benefit.

You can see what I mean here: https://github.com/publicarray/spksrc/commits/sickchill-cross

https://github.com/publicarray/spksrc/commit/d20c8eeb786c6573a4151c7df01a68482202154d.patch

spk/sickchill/src/service-setup.sh
```diff
@@ -12,32 +12,43 @@ SC_CFG_FILE="${SC_DATA_DIR}/config.ini"
...
 EOF
     fi
 }
 
+service_preinst ()
+{
+    # Avoid ssl errors in git
+    SSL_CRT="/etc/ssl/certs/ca-certificates.crt"
+    if [ -f ${SSL_CRT} ]; then
+        ${GIT} config --system http.sslCAInfo ${SSL_CRT} > /dev/null 2>&1
```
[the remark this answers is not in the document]
miigotu: That was from an old package, pretty sure it can be removed now that it is in default gitconfig.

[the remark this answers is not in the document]
miigotu: benjv's package is a mess, SC used to be no-arch when you could build python cryptography and lxml without a c compiler and that is no longer the case. Technically any python application that uses python-requests with ssl cert verification requires a compiler to install unless it is on a platform that manylinux wheels download for.

Recently, cryptography also requires a rust compiler :(

I am unsure about the cross compiler being needed, it was only used before to hook in and create a tar archive? We can just grab the spk from build artifacts.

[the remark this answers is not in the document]
miigotu: I've added your commit, thanks for that. I'm sure you know better about how the cc works than me lol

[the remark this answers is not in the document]
publicarray: Thanks for explaining. So cross arch it is.

I like rust but yea sounds odd that python packages requires more different compilers than other languages.

Yes it's just used to download the tar archive at the moment (talking about my branch) the benefit is that the package can be installed in an offline environment and the sha265 hash is checked in. So I'm happy to accept the PR with my changes included. If you want to download the tar from spk/makefile that is another option, but it won't speedup the build and is not recommended. Doing it in the service-startup.sh means a network connection is required on installation (not that bad but means the package is unpredictable might work for one user might fail for another, harder to offer support).

spk/sickchill/src/service-setup.sh
```diff
@@ -12,28 +12,29 @@ SC_CFG_FILE="${SC_DATA_DIR}/config.ini"
...
-        sed -i "/^\s*branch\s*=/s/\s*=\s*.*/ = ${SC_CONFIG_GIT_BRANCH}/" ${SC_CFG_FILE}
-        sed -i "/^\s*cur_commit_hash\s*=/s/\s*=\s*.*/ = ${SC_CONFIG_GIT_COMMIT_HASH}/" ${SC_CFG_FILE}
-        sed -i "/^\s*cur_commit_branch\s*=/s/\s*=\s*.*/ = ${SC_CONFIG_GIT_COMMIT_BRANCH}/" ${SC_CFG_FILE}
+        # postupgrade we don't want to change their branch or commit, just remove them as the updater should repopulate these
+        sed -i "/^branch/d" ${SC_CFG_FILE}
+        sed -i "/^cur_commit_hash/d" ${SC_CFG_FILE}
+        sed -i "/^cur_commit_branch/d" ${SC_CFG_FILE}
+
+        # Make sure we always update the port in their restored config
+        sed -i "s/\(web_port *= *\).*/\18081/" ${SC_CFG_FILE}
```
publicarray, line 29: Should the user configuration for `update_frequency` also be restored?

publicarray: But the port should be `8081` not `18081`?

publicarray: `\1` here is the capture group 🤦

miigotu:
Thanks, I'm sure there are a lot more changes that will need made while the package is young. Hopefully this works.

cross/sickchill/Makefile
```diff
@@ -1,14 +1,10 @@
 PKG_NAME = SickChill
-PKG_VERS = 2021.03.10-1
+PKG_VERS = 2021.03.28-3
 PKG_EXT = tar.gz
 PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = https://github.com/SickChill/SickChill/archive
+PKG_DIST_SITE = https://github.com/SickChill/SickChill/archive/refs/tags/
```
miigotu, line 5: @publicarray is this line wrong?

cross/sickchill/Makefile
```diff
@@ -1,14 +1,10 @@
 PKG_NAME = SickChill
-PKG_VERS = 2021.03.10-1
+PKG_VERS = 2021.03.28-3
```
miigotu, line 2: Is this supposed to match the PKG version?

cross/sickchill/Makefile
publicarray: GitHub differentiates between releases and tags. At the moment your releases are just tags: https://github.com/SickChill/SickChill/releases. (hover over tar.gz to see the url)

The last proper GitHub release was https://github.com/SickChill/SickChill/releases/tag/v2020.07.09-1 but wanted a more recent one.

cross/sickchill/Makefile
publicarray: Ideally yes but the other way around? https://github.com/SynoCommunity/spksrc/pull/4161#discussion_r593827730

publicarray: This is also used to download the tarball

cross/sickchill/Makefile
publicarray: Actually now I don't know both use the `/refs/tags`

cross/sickchill/Makefile
miigotu: The shell date that was there had no dots, so when I tried changing version to be set with the dots properly I could not update in syno gui because it thought I was downgrading. I would rather have the dots.

publicarray: Sorry it's too late. 2021 is simply smaller than 20210329 there is nothing we can do now. My hint/warning could have been clearer. You probably wanted `date +%Y.%m.%d`

publicarray (approved):

miigotu:
!DNM

miigotu:
@publicarray I do not think cross compiling is proper here unless we can just cross-compile the python requirements and then clone SC. We need to retain git history and cross currently would download a tarball which breaks SC. I am not sure git is always available in the package build environment, so cloning inside cross/* might not always work. As an upside, we can now allow users to select the git branch at install. The purpose of that is so that users can test changes on synology to see if their issues are fixed without pushing a new update to synocommunity.

miigotu:
Anyone know why the service is created with no privileges? 

package.tgz/app/config has the proper info generated afaik:
```
{ ".url": { 
  "com.synocommunity.packages.sickchill": {
    "title": "SickChill",
    "desc": "Automatic Video Library Manager for TV Shows. It watches for new episodes of your favorite shows, and when they are posted it does its magic.",
    "icon": "images/sickchill-{0}.png",
    "type": "url",
    "protocol": "http",
    "port": "8081",
    "url": "/",
    "allUsers": true,
    "grantPrivilege": "all",
    "advanceGrantPrivilege": true
} } }
```

ymartin59 (approved):

BKSteve:
Is this update frozen?

miigotu:
> Is this update frozen?

This is just the packaging source, changes to SC are applied when you update SC inside after installing.

BKSteve:
It's just no update on Community since 'Initial version' Nyaran. 
Is this update going to also look for the Synology Python3 package rather than 3.7.10?

miigotu:
Both pythons packages are acceptable, we just require python3 >3.6

miigotu:
Once the package is set proper on community, it never needs updated again unless requirements change.

spk/sickchill/Makefile
```diff
@@ -1,20 +1,19 @@
 SPK_NAME = sickchill
-SPK_VERS = $(shell date +%Y%m%d)
+SPK_VERS = 20210419
 SPK_REV = 1
 SPK_ICON = src/sickchill.png
 
-DEPENDS = cross/$(SPK_NAME)
 SPK_DEPENDS = "python3>=3.6:git"
```
miigotu, line 6: @publicarray can we do `python3>=3.6||python>=3.6:git` here? Or is there some other reason Synology official python won't satisfy the dependency?

hgy59: No we can't. We must decide for one or the other (or create two different packages like sickchill/Sickchill (or sc-sickchill/synology-sickchill), but as git package is by synocommunity I would prefere to keep it like `"python3;git"`

miigotu: Yeah it wasnt the git package I was worried about, it is the confusion by some people who already have python 3 installed from syno official and then are asked to install python3 from sc when they install sickchill. One is named python, the other is named python3, even though both are python3.6+

hgy59: As a rule of thumb the (internal) name of packages from SynoCommunity are all lower case and the packages from synology start with uppercase letter (synocommunity: mono, synology: Mono)

BenjV:
@hgy59 
Because DSM 7 has python 3 standard installed you could remove the dependency to python3 from the INFO file.

If you add this to the start-stop-status script is would use the python from the SynoCommunity is installed, else the standard python3 from DSM.
```
if [ -x "/var/packages/python38/target/bin/python3.8" ]
then
      PYTHON="/var/packages/python38/target/bin/python3.8"
else
      PYTHON="python3"
fi
```

miigotu:
Im reworking this package for dsm7 (are the changes necessary usually backwards compatible to dsm6?)
Do you guys have a discord/telegram/irc or something to discuss easier?

BenjV:
Not backwards compatible, DSM 7 needs this in the INFO file:

> os_min_ver="7.0-40000"

This will prevent it from being able to install it on DSM versions lower then 7
Also on DSM 7 you cannot run packages nor install script as root, so that has to be removed from the privilege file.
This file must now look like this:
```
{
  "defaults": {
    "run-as": "package"
  },
  "username": "sc-medusa",
  "groupname": "synocommunity"
}
```
So there must be two differnet packages, one for DSM 7 and higher and one for DSM 6 and lower.

hgy59:
> Im reworking this package for dsm7 (are the changes necessary usually backwards compatible to dsm6?)

The intention is to create packages for dsm6 and dsm7 with a single package definition (i.e. files below `spk/{package-name}`. Most of the compatibility is done in the development environment (i.e. the spksrc `framework`). The so called `generic installer` uses different installer generation for DSM7, DSM6 (and DSM5).
For special cases, it is possible to execute DSM version specific code for a package in service-setup.sh by evaluation of the `SYNOPKG_DSM_VERSION_MAJOR` variable.

miigotu:
Trying to fix and update in #4743 - This is outdated but I will come back to this branch if the other approach does not work out

boomvalk:
After updating all packages on my Synology, including Sickchill: SickChill nor Sickrage will startup again "Failed to run the package service."I have DSM 6.2, Sickchill210329-01, Python 3, Pytohn 3.8, Python all installed, mono 5.2.1.34-16, SABnzbbd 3.2.1-48,

PR #4537, opened 2021-04-01
Update SABnzbd to 3.2.1

Safihre:
### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4527, opened 2021-03-28
syncthing: Upgrade to v1.15.1 and ignore folder permissions by default.

acolomb:
_Motivation:_  Upstream v1.14.0 was released on 2021-03-02.  Some option defaults are changed in this PR, which only applies for new installations, because `config.xml` is (and must be) kept during updates.

The new version introduces a "folder defaults" section in the configuration, which is used to preset the `ignorePerms` option to `true`.  This will not touch any existing folders, but set the option when configuring a new folder to be synced.  Specifically on DSM, the `ignorePerms` option should definitely be used with every Syncthing folder.  Otherwise, it will possibly try to override the UNIX permissions on synced files, which will reset the ACL-based permissions.  Possibly making them unusable for the rest of the DSM services.

Change the default for the `cacheIgnoredFiles` option to `false` (which is the upstream default). Many DiskStations are rather tight on RAM, so deviating from Syncthing's default here is probably a bad idea for most users.  It only matters when many (or very complex) ignore patterns are used in a folder, so usually for power users.  Those will certainly be able to change the option manually.

Drop the obsolete `symlinksEnabled` and `weakHashSelectionMethod` entries.  Switch from `minHomeDiskFreePct` to the new syntax `minHomeDiskFree unit="%"`.

Note that the previous package version should have already self-updated an installed Syncthing binary when the service was running long enough.  So this PR should not really change anything for users updating the SPK.

_Linked issues:_  
Configurable folder defaults: https://github.com/syncthing/syncthing/pull/7131
`symlinksEnabled`: https://github.com/syncthing/docs/commit/392dbd0b
`weakHashSelectionMethod`: https://github.com/syncthing/syncthing/pull/4767

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

(Sorry, I have no DSM installation to spare where Syncthing is not used in production, so testing is limited. Should be low impact though as mentioned above.)

Issue #4524, opened 2021-03-27
Meta: DSM7 package status


hgy59:
Overview of the status of packages for DSM7

| PACKAGE | Build | Install | Run | Published |COMMENT |
|-----------|------|--------|---------|------------|-------------|
| adminer | ✔️ | ✔️ | ✔️ |   | noarch |
| beets | | | | | |
| bicbucstriim | ✔️ | | | | |
| borgbackup | | | | | |
| boxbackup-client | ✔️| | | | |
| chromaprint | | | | | |
| comskip | | | | | |
| cops | ✔️| | | | |
| couchpotatoserver | ✔️| | | | |
| couchpotatoserver-custom | ✔️| | | | |
| curaengine | ✔️| | | | |
| dante-sockd | | | | | |
| darkstat | ✔️| | | | |
| debian-chroot | | | | | |
| deluge | | | | | |
| demoservice | ✔️| | | | |
| dnscrypt-proxy | ✔️| | | | |
| domoticz | | | | | |
| duplicity | ✔️| | | | |
| ejabberd | | | | | |
| erlang | ✔️| | | | |
| fengoffice | ✔️| | | | |
| ffmpeg | | | | | |
| ffsync | | | | | |
| fish | | | | | |
| fishnet | ✔️| | | | |
| flexget | ✔️| | | | |
| fossil-scm | ✔️| | | | |
| full-text-rss | ✔️| | | | |
| gateone | ✔️| | | | |
| gentoo-chroot | | | | | |
| git | ✔️| | | | |
| git-server | ✔️| | | | |
| gnupg | ✔️| | | | |
| haproxy | ✔️ | | | | |
| hassio | ✔️| | | | |
| he853 | ✔️| | | | |
| headphones | ✔️| | | | |
| headphones-custom | ✔️| | | | |
| homeassistant | ✔️| | | | |
| horde | ✔️| | | | |
| htpcmanager | ✔️| | | | |
| icecast | ✔️| | | | |
| imagemagick | ✔️| | | | |
| inotify-tools | ✔️| | | | |
| irssi | ✔️| | | | |
| itools | ✔️| | | | |
| jackett | | | | | |
| jappix | ✔️| | | | |
| jupp | ✔️| | | | |
| lazylibrarian | ✔️| | | | |
| lidarr | ✔️| | | | |
| lirc | | | | | |
| lua | ✔️| | | | |
| mantisbt | ✔️| | | | |
| maraschino | ✔️| | | | |
[...]

PR #4494, opened 2021-03-14
Fishshell: Fix DSM7 build

publicarray:
_Motivation:_  Start support for DSM7. On armv7/aarch64 cmake can't find pthreads.h
_Linked issues:_  https://github.com/SynoCommunity/spksrc/pull/4463#issuecomment-798780125

### Checklist
- [x] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4539, opened 2021-04-02
redesign installer logs

hgy59:
_Motivation:_  Installation scripts using `>> ${INST_LOG}` do not work with DSM7
_Linked issues:_ #4524, #4494, #4527, #4537, #4545, #4153

### Checklist

Documentation that needs update:
- [ ] [CONTRIBUTING.md](https://github.com/SynoCommunity/spksrc/blob/master/CONTRIBUTING.md)
- [ ] [generic service support](https://github.com/SynoCommunity/spksrc/wiki/Service-Support#installer)

### Remarks
It is time to redesign the package installation logs, as appending output with  `>> ${INST_LOG}` is not supported with DSM7.
In DSM7 write access to the new DSM installation log file `/var/log/packages/{package-name.log}` is not allowed. 
With DSM7 all stderr output during installation is written to the logfile by the system and output to stdout is presented in a modal dialog to notify the user.

The purpose of this PR is to realize installation logging independently on the DSM version.

Finally the new function `install_log` is used internally by the generic installer only.
For all packages with custom installer functions (service_*) the only change is to remove the output redirections `>> ${INST_LOG}` and `>> ${INST_LOG} 2>&1`.
All output in the package specific installer functions will be written to the DSM dependent installer log file.

Packages that use `exit 1` and/or user notifications in the installation process by writing to stdout, must use the new validate-functions to do so. User notification is not supported for DSM < 6. 

The new validate functions are:
- validate_preinst
- validate_preunint
- validate_preupgrade

The benefit of negative validation (exit1) is, that the user can navigate in the wizard to fix the installation parameters and try again. This might be `mysql administration password`, `shared folder that must exist`, etc.

_For the rare case a package needs to download or install code before the validation, a validate function must be used and the install_log function can be used there for regular installation logs._

### Installer log file
The install log with `INST_LOG` was introduced for pre DSM 6 packages, where DSM system did not create log file for package installation scripts.
Finally the `/var/log/packages/{package-name}.log` is not new for DSM7 but is provided since DSM6. So this PR modifies all DSM6 packages to use the `/var/log/packages/{package-name}.log` and do not create `/var/package/{package-name}/target/var/{package-name}_install.log` anymore.
The standard for installer log is now `/var/log/packages/{package-name}.log`.
Only for DSM < 6 the `/var/package/{package-name}/target/var/{package-name}_install.log` is still in use.

### Affected Packages
most of the Comments below are for building arch-x64-7.0

✔️: validation of installer log fixed for DSM7 
✔️✔️: validation of installer log fixed for DSM7 and DSM6

fixed:
- remove `>> ${INST_LOG}`
- firewall rules for DSM7 (@publicarray)
- replace `/usr/local/{package-name}`by `/var/packages/{package-name}/target`

not fixed in this PR:
- remove busybox and deprecated user management at installation (DSM 5 to DSM 6 migration is not supported anymore!)
- not all packages use the new var folder with DSM7, some use it only for log and pid of service)
- package warnings like `has not been updated to DSM7 yet` as function ... is no longer supported.
- `SERVICE_WIZARD_SHARE` implementation needs update vor DSM7

| Package | WIP | Comment |
| - | - | - |
| ✔️✔️ beets| | noarch |
| ✔️✔️ borgbackup| |
| ✔️✔️ couchpotatoserver| | noarch |
| ✔️✔️ couchpotatoserver-custom| | noarch |
| ✔️✔️ deluge| #4153 | compile error: expected primary-expression before ‘float’|
| ✔️✔️ demoservice| | noarch, created link to @installer.log does not work on DSM6, but works on DSM7 |
| ✔️✔️ dnscrypt-proxy| | not installable, requires root permissions, DSM6 has logs about missing `localized string` for several languages  |
| ✔️✔️ fish| #4494 | fails to compile for DSM7,  error: ‘__builtin_isnan’ is not a member of ‘std’ |
| ✔️✔️ fishnet| | no wizard to enter `fishnet_key` |
| ✔️✔️ flexget| | service log has all entries doubled |
| ✔️✔️ gateone| | service log has all entries doubled, DSM7: certificate cannot be installed, start fails |
| ✔️✔️ headphones| | noarch |
| ✔️✔️ headphones-custom| | noarch |
| ✔️✔️ homeassistant| |
| ✔️✔️ itools| | not installable, requires root permissions, depends on python module |
| ✔️✔️ lazylibrarian| | noarch |
| ✔️✔️ lidarr| |
| ✔️✔️ minio| |DSM7: migrate SERVICE_EXE (start-stop-daemon) to SERVICE_COMMAND|
| ✔️✔️ mono| |
| ✔️✔️ mosquitto| | mosquitto.conf and *.pid in wrong var folder |
| ✔️✔️ mutt| |
| ✔️✔️ mylar| | datadir in wrong var folder, fails to start on DSM6 |
| ✔️✔️ nzbget| | with multistage validation and and print logfile name to stdout, must not use `su` anymore |
| ✔️✔️ nzbget-testing| | with multistage validation and print logfile name to stdout, must not use `su` anymore, package is obsolete, as `stable / testing (beta)` is provided by the wizard of nzbget package |
| ✔️✔️ nzbhydra| | noarch |
| ✔️✔️ plexpy-custom| noarch |
| ✔️ pyload| |
| ✔️✔️ python3| |
| ✔️✔️ python38| |
| ✔️✔️ rabbitmq| |
| ✔️✔️ radarr| [dsm7-packages](https://github.com/SynoCommunity/spksrc/compare/dsm7-packages) |
| ✔️✔️ rdiff-backup| | usr/local links do not work |
| ✔️✔️ rsnapshot| | noarch, configure: error: rsync not found (DSM 6 and 7)|
| ✔️✔️ rutorrent| |libtorrent:  error: ‘__builtin_isnan’ is not a member of ‘std’|
| ✔️✔️ sabnzbd| #4537 | installation fails, /var/log/messages: failed to alloc _share_, failed to get defualt status for _share_, failed to create _share_, failed to acquire resource after install |
| ✔️✔️ salt-minion| | installation stays `installing...` |
| ✔️✔️ sickbeard-custom| | noarch |
| ✔️✔️ sickrage| | noarch, fails to install: missing apispec==4.0.0 |
| ✔️✔️ sonarr (nzbdrone) | |
| ✔️✔️ stockfish| | web app must not access `/var/services/web/{package-name}` |
| ✔️✔️ syncthing| #4527 | DSM7: migrate SERVICE_EXE (start-stop-daemon) to SERVICE_COMMAND|
| ✔️✔️ synocli-monitor| |
| ✔️✔️ transmission|  [dsm7-packages](https://github.com/SynoCommunity/spksrc/compare/dsm7-packages)  | installation fails, /var/log/messages: failed to alloc _share_, failed to get defualt status for _share_, failed to create _share_, failed to acquire resource after install |
| ✔️✔️ tvheadend| #4545,  [dsm7-packages](https://github.com/SynoCommunity/spksrc/compare/dsm7-packages)  | configure ERROR: ffmpeg development support not found, ffmpeg:  error: ‘__builtin_isnan’ is not a member of ‘std’|
| ✔️✔️ umurmur| | certificate generation failed, /var/packages/umurmur/target/openssl.cnf is missing |


mk/spksrc.service.installer
```diff
@@ -20,6 +23,21 @@ else
...
+install_log ()
+{
+  local _msg_="$@"
+  if [ -z "${_msg_}" ]; then
+      # read multiline from stdin
+      while IFS=$'\n' read -r line; do
+        install_log "${line}"
+      done
+  else
+    echo $(date +'%Y/%m/%d %H:%M:%S')$'\t'${_msg_} >> ${INST_LOG} 2>&1
```
publicarray, line 36:
```suggestion
    echo $(date +'%Y/%m/%d %H:%M:%S')$'\t'${_msg_} >> ${INST_LOG} 1>&2
```

publicarray (changes requested):
Nice work! It was on my low priority todo list :+1: 

> With DSM7 all stdout during installation is written to the logfile by the system and output to stderror is presented to the user in a modal dialog.

From my testing it's the other way around.  `stderr` is logged and `stdout` is presented to the user.

`1>&2` redirects stdout to stderr

mk/spksrc.service.create_links
```diff
@@ -5,9 +5,9 @@ service_create_links ()
     for cmd in ${SPK_COMMANDS}
     do
         if [ -e "${SYNOPKG_PKGDEST}/${cmd}" ]; then
-            mkdir -p "$(dirname /usr/local/${cmd})"  >> "${INST_LOG}" 2>&1
-            echo "create link: /usr/local/${cmd} -> ${SYNOPKG_PKGDEST}/${cmd}"  >> "${INST_LOG}"
-            ln -s "${SYNOPKG_PKGDEST}/${cmd}" "/usr/local/${cmd}"  >> "${INST_LOG}" 2>&1
+            mkdir -p "$(dirname /usr/local/${cmd})" 2>&1 | install_log
```
publicarray, line 8: Can we remove the duplicate redirections trough out?

hgy59:
> Nice work! It was on my low priority todo list

And for me it is on the high priority list, as a couple of packages fail to install (ambigous redirect) when I tried to build and install all packages for arch-x64-7.0 (see #4524)
This is high priority for packages that others depend on (python2, python3, python38, and maybe mono).

publicarray:
Oh, I don't remember single package to fail with that error. Thanks for fixing it. Before my NAS died, I had my own compiled python3 and mono installed.

mk/spksrc.service.create_links
hgy59: > Maybe we can use install_log to encapsulate all package scripts but allow a switch or function to disable it in case a package wants to notify the user. IMHO logs should always be written and not something a package author should worry about. 

I like the idea to always write to the installer log file and it is easy to implement.
But I did not find any solution on how to disable this and write to stdout while all output is captured for the installer log (I tried to collect notification output in a global variable or to use file descriptor 3 - nothing worked).

The only solution I can imagine for user notification (i.e. writing to stdout) is to implement additional service functions that are not redirected to the installer log like `validate_preinst` or `validate_preupgrade` and that must be used for such validations.
This would work at least for validation of wizard variables, but not for complex installers that have multiple installation steps with extra validation (if we have any).
The real benefit of user notification is for `preinst`, `preuninst` and `preupgrade` anyway, as the user can navigate back in the wizard and fix the configuration before starting the installation/uninstallation.
Further a `validate_postinst` could notify when additional manual configuration might be required.

th0ma7:
This is good work indeed.
Is the intent to also adapt all `src/service-setup.sh` scripts as well right away to migrate them to the new logging facility?

And, out of curiosity, I wonder if it could be possible to alter default `syslog` configuration through a generic default `synocommunity.spk` to add a new logging facility channel to redirect synoCommunity package logs to our own logs files, thus having the ability to capture the stderr?

Just a thought.

hgy59:
> Is the intent to also adapt all `src/service-setup.sh` scripts as well right away to migrate them to the new logging facility?

Yes, this is my intention to update all packages, in spite of those that are WIP on branches (or - maybe it is better not to touch such packages and advise the maintainers after this one is merged).

@th0ma7 regarding sc specific syslog configuration I propose to work on a different branch. After my redesign here there will be no need to modify each package for such a feature....

publicarray:
Unfortunately using the `|` pipe or `call_func` negates the exit status. I can't find a way to manually `exit` on an error.


```sh
service_preinst ()
{
    echo "****************"
    exit 1
}
``` 

I tried to grab it and pass it on with `local _exit_code_="$?"` `local _exit_code_="${PIPESTATUS[0]}"` but no success so far

hgy59:
> 
> 
> Unfortunately using the `|` pipe or `call_func` negates the exit status. I can't find a way to manually `exit` on an error.
> 
> ```shell
> service_preinst ()
> {
>     echo "****************"
>     exit 1
> }
> ```
> 
> I tried to grab it and pass it on with `local _exit_code_="$?"` `local _exit_code_="${PIPESTATUS[0]}"` but no success so far

This works with the provided validate_ functions:
```shell
validate_preinst ()
{
    echo "****************"
    exit 1
}
```

So far I only found one package with user notifications that cannot be easyly migrated to validate-functions (nzbget/nzbget-testing).
Others can use this validation straight foreward (headphones/headphones-custom, transmission, ...)

publicarray:
Sorry my point it that if a command fails that the script will continue to run instead of exiting.

hgy59:
> 
> 
> Sorry my point is that if a command fails then the script will continue to run instead of exiting.

IMHO this was always the case. You must use `exit` to break a function. 
If you want to exit with user notification, you must now implement a validate-function.

publicarray:
@hgy59 It seems you are right. I just found it surprising that `exit 1` no longer stops the installation in the service functions.

mk/spksrc.service.installer
hgy59: in the pre dsm7 installer, we want to redirect everyting (stdout and stderr) to the log file (not to stderr).

th0ma7:
> @th0ma7 regarding sc specific syslog configuration I propose to work on a different branch. After my redesign here there will be no need to modify each package for such a feature....

Page 115 of the dev guide, part of the resource list there are hooks available to add `syslog-ng` configurations.  Definitively something worth investigating (but my cycles are rather limited currently)

publicarray:
`INST_VAR` is only used in 4 packages, so I'm also happy to remove it :+1: 

```sh
$ rg -g '**/*.sh' INST_VAR -l
spk/minio/src/service-setup.sh
spk/minio/src/wizard/upgrade_uifile.sh
spk/sonarr/src/service-setup.sh
spk/lidarr/src/service-setup.sh
spk/lidarr/src/service-setup-mono.sh
spk/radarr/src/service-setup.sh
```

I noticed another oversight on my part, the firewall rules on the DSM7 are also defined in the resource file. I'll happily send a PR
 once the 2 current ones are merged.

spk.service.mk
```make
ifeq ($(strip $(FWPORTS)),)
	@jq '."port-config"."protocol-file" = "$(SPK_NAME)"' $@ 1<>$@
endif
```

spk.service.installer.dsm7
```diff
- # Add firewall config
-  if [ -r "${FWPORTS_FILE}" ]; then
-        install_log "Installing service configuration ${FWPORTS_FILE}"
-        servicetool --install-configure-file --package "${FWPORTS_FILE}" 2>&1 | install_log
-  fi

- # Remove firewall config
- if [ -r "${FWPORTS_FILE}" ]; then
-        install_log "Removing service configuration ${SYNOPKG_PKGNAME}.sc"
-        servicetool --remove-configure-file --package "${SYNOPKG_PKGNAME}.sc" 2>&1 | install_log
- fi
```

publicarray (approved):

hgy59:
> I noticed another oversight on my part, the firewall rules on the DSM7 are also defined in the resource file. I'll happily send a PR
> once the 2 current ones are merged.

Yes, I noticed install errors like
 `install configue file fail! (/var/packages/rabbitmq/conf/rabbitmq.sc)`

So shall I commit all the package fixes and then we merge it - or can I cherry pick the DSM7 firwall rules fix somewhere?

For now I am holding back, as I have to test the installation of 61 packages (12 noarch/49 arch specific) on DSM 7 and DSM 6.1 and  evtl. DSM 5.2.

hgy59:
> ```shell
> $ rg -g '**/*.sh' INST_VAR -l
> spk/minio/src/service-setup.sh
> spk/minio/src/wizard/upgrade_uifile.sh
> spk/sonarr/src/service-setup.sh
> spk/lidarr/src/service-setup.sh
> spk/lidarr/src/service-setup-mono.sh
> spk/radarr/src/service-setup.sh
> ```

A search for `INST_VAR`  as whole word omits minio, as it contains `INST_VARIABLES` and not `INST_VAR`.
`$ rg -g '**/*.sh' INST_VAR -lw`

publicarray:
The firewall rule is a bit more complicated than I thought. Using the resource worker it expects the firewall file to reside in `/var/package/{$package}/target/` so I've moved it to `/var/packages/${SYNOPKG_PKGNAME}/target/app/${SYNOPKG_PKGNAME}.sc` feel free to change it. I've tested the `SERVICE_PORT` option but have not tested the `FWPORTS` option yet.

publicarray:
Sorry I had these locally and missed pushing them earlier.

mk/spksrc.service.mk
```diff
@@ -196,10 +196,16 @@ ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 $(DSM_SCRIPTS_DIR)/installer: $(SPKSRC_MK)spksrc.service.installer.dsm7
 	@$(dsm_script_copy)
 else
+ifeq ($(call version_lt, ${TCVERSION}, 6.0),1)
```
publicarray, line 199: I think this could be written more consistently e.g. as greater than or equal to 7.0; dsm7 stuff, else greater than or equal to 6.0; dsm6 stuff else dsm5 stuff. Or the reverse. But I dislike this version.

hgy59: @publicarray i have a question regarding your comment in the servive.mk for DSM7
`# STARTABLE needs to be yes, Resource linking and unlinking works on start and stop`

I wanted to fix this, as only `SPK_COMMANDS` was tested here and `SPK_USR_LOCAL_LINKS` was ignored.
But my findings are, that it is not needed to omit `ctl_stop=no` (i.e. use default `ctl_stop=yes`) for resource linker to work.
The packages with `ctl_stop=no` are started by the system at installation time (and on every restart) and are stopped at uninstall.
This would simplify the code as it will depend on DSM >= 6.1 only to use `ctl_stop=no` or `startable=no`.

hgy59: @publicarray the implementation of `ctl_stop=no` is in `spksrc.spk.mk` and not in `spksrc.service.mk` but your comment is in both.

hgy59:
@publicarray now I still only problems with demoservice on DSM7.
The installation log does not show any error (or ret != 0) but package fails to install. Only /var/log/message shows the following error (json formatted for better reading).
```json
Failed to install package, pkg=[demoservice] 
result = [{
        "action": "install",
        "beta": false,
        "betaIncoming": false,
        "error": {
            "code": 276,
            "description": "failed to acquire postinst worker",
            "worker_msg": []
        },
        "finished": true,
        "installReboot": false,
        "installing": true,
        "language": "ger",
        "last_stage": "postreplace",
        "package": "demoservice",
        "packageName": "DemoService",
        "pid": 18788,
        "scripts": [{
                "code": 0,
                "message": "",
                "type": "preinst"
            }, {
                "code": 0,
                "message": "",
                "type": "postinst"
            }
        ],
        "spk": "/volume1/@tmp/upload_tmp.180930",
        "stage": "install_failed",
        "status": "stop",
        "success": false,
        "username": "xxxxxxxx"
    }
]
```

Did you ever get the error `"failed to acquire postinst worker"` ?

hgy59:
I have added a note for packages using `SERVICE_WIZARD_SHARE` in #4524.
@publicarray: as `transmission` uses this, you might have solved this in the dsm7-packages branch?

publicarray:
> I have added a note for packages using `SERVICE_WIZARD_SHARE` in #4524.
> @publicarray: as `transmission` uses this, you might have solved this in the dsm7-packages branch?

Indeed, It's in 54c56c620ad2abce87321ca5f6dd85115c9bd913

publicarray:
Ah, I found the error the implemented version required a single folder name, not the full path, maybe there could be backwards compatibility shim by splitting the path `/` alternatively I suggest upgrading the wizard files (transmission example). IIRC They work in both dsm6/dsm7

spk/dnscrypt-proxy/src/service-setup.sh
```diff
@@ -117,25 +117,25 @@ service_prestart () {
...
-    echo "Running service_postinst script" >> "${INST_LOG}"
-    mkdir -p "${SYNOPKG_PKGDEST}"/var >> "${INST_LOG}" 2>&1
+    echo "Running service_postinst script"
+    mkdir -p "${SYNOPKG_PKGDEST}"/var
     if [ ! -e "${CFG_FILE}" ]; then
         # shellcheck disable=SC2086
-        cp -f ${EXAMPLE_FILES} "${SYNOPKG_PKGDEST}/var/" >> "${INST_LOG}" 2>&1
-        cp -f "${SYNOPKG_PKGDEST}"/offline-cache/* "${SYNOPKG_PKGDEST}/var/" >> "${INST_LOG}" 2>&1
-        cp -f "${SYNOPKG_PKGDEST}"/blocklist/* "${SYNOPKG_PKGDEST}/var/" >> "${INST_LOG}" 2>&1
+        cp -f "${EXAMPLE_FILES}" "${SYNOPKG_PKGDEST}/var/"
```
publicarray, line 130: I'm not sure this works. `${EXAMPLE_FILES}` includes a glob: `example-*`

spk/borgbackup/Makefile
```diff
@@ -27,6 +27,12 @@ HOMEPAGE = https://borgbackup.readthedocs.io
 LICENSE  = BSD-3-Clause
 
 SERVICE_SETUP = src/service-setup.sh
+
+include ../../mk/spksrc.common.mk
+ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
```
publicarray, line 32: Since `usr-local-linker` was introduced since `6.0-5941` why not use for the DSM6 as well?

publicarray:
I was under the impression that we wanted to address this in a PR per package basis. But I can't complain, this is big step forward, thanks for your hard work! Just FYI this will course many merge conflicts.

mk/spksrc.service.mk
publicarray: Thanks for the optimisation. The downside is that all CLI apps have `STARTABLE=no` set. I just tested it again and CLI packages (I use `zsh` for testing) that have `ctl_stop=no` will still work but the `start/stop` button is obviously missing. IMHO we should override and remove the `ctl_stop=no` if `SPK_COMMANDS` is set. `SPK_USR_LOCAL_LINKS` doesn't guarantee an executable but yes it should also apply then.

hgy59:
Originally this PR should only address the installer log redesign. But now also the migration from `/usr/local/{package-name}` to `/var/packages/{package-name}/target` is included.
Only packages that are not WIP in another PR are updated now to avoid merge conflict.
I propose to merge as it is now.

spk/borgbackup/Makefile
hgy59: Yes, lets do it in another PR.

hgy59:
merge conflicts are expected for demoservice (#4497) but this is my own workplace.

hgy59:
About the github build action failed DSM 6 builds 
- rsnapshot: I have a fix ready to pull after this PR is merged
- mosquitto: the server mosquitto.org is currently offline
- pyload: Cannot reproduce, successfully compiles, installs and runs for me, even on DSM7

spk/dnscrypt-proxy/src/service-setup.sh
publicarray:
```suggestion
        cp -f ${EXAMPLE_FILES} "${SYNOPKG_PKGDEST}/var/"
```

This is a regression, otherwise I'm happy to merge as is

hgy59:
@publicarray there is now some rework on the dsm7-packages branch required.

for ***radarr*** in service-setup.sh in the function service_preupgrade:
- replace INST_VAR by SYNOPKG_PKGVAR
- remove `>> ${INSTLOG}`

TODO for ***transmission*** in service-setup.sh
- rename service_preinst to validate_preinst
- the ` >> ${INST_LOG}`  have to be removed (service_postint)
- and for DSM7 compatibility use: PYTHON_DIR="/var/packages/python/target"

I think by #4545 that ***tvheadend*** is outdated on the dsm7-packages branch (I posted a note in #4545).

PR #4541, opened 2021-04-04
Bump OpenSSL to 1.1.1k and Python to 3.8.9

Safihre:
OpenSSL 1.1.1k fixes 2 high-severity problems for servers, so relevant for DMS users: CVE-2021-3449 and CVE-2021-3450.
I think the pipeline will try to build like 15 packages, so not sure if it fails if it has anything to do with the change.

### DIFF
--- a/cross/openssl/Makefile
+++ b/cross/openssl/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = openssl
-PKG_VERS = 1.1.1j
+PKG_VERS = 1.1.1k
 PKG_EXT = tar.gz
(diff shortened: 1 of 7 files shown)


Safihre:
Oh, I realize that if we want to really do this correctly all the packages depending on `cross/openssl` need to be bumped, that's a lot..

Safihre:
Guess this PR would involve too many packages.

zyrill:
Why would you close this PR just because a number of packages are affected? It needs to be fixed eventually, might as well do it now or what's the alternative?

hgy59:
I propose tp split the update of openssl and python38 to different PRs.

zyrill:
Yeah ok, that makes sense. Thanks for the clarification.

PR #4545, opened 2021-04-08
Fix Tvheadend and ffmpeg for DMS 7

publicarray:
_Motivation:_ cherry-picked commits from dsm7-packages branch and added fixes to uriparser and tvheadend version number
_Linked issues:_  

### Checklist


publicarray:
On the dsm 6.x the version upgrade worked. (probably because the revision number is newer)

publicarray:
* ejabberd & erlang is out of scope in this PR.  (changes in `expat` caused them to build, and they failed to build before: #4333)
* chromaprint is proven difficult and IMHO should be addressed in a separate PR as well.

th0ma7:
@publicarray  I quickly went over this PR and at first glance things looked OK.
But indeed it failed to build miserably due to `expat`.  It looked like it was unable to download properly from sourceforge.net as the URL changed.  Can that be fixed so it builds?

I'll have another look at it over the week-end and from there might be able to propose experimental package on my github for the community to play with so we get some feedbacks.

cross/expat/Makefile
```diff
@@ -1,9 +1,10 @@
...
 PKG_VERS = 2.2.7
-PKG_EXT = tar.bz2
-PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = https://downloads.sourceforge.net/project/$(PKG_NAME)/$(PKG_NAME)/$(PKG_VERS)
-PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+PKG_EXT = tar.gz
+PKG_DIST_NAME = R_2_2_7.$(PKG_EXT)
+PKG_DIST_SITE = https://github.com/lib$(PKG_NAME)/lib$(PKG_NAME)/archive/refs/tags
+PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIR = lib$(PKG_NAME)-R_2_2_7/$(PKG_NAME)
```
hgy59, line 7: With the `tar.bz2` source of github you get exactly the same file as on sourceforge.
Then you can omit `expat_pre_configure` and revert the digests file.

```suggestion
PKG_DIST_SITE = https://github.com/libexpat/lib$(PKG_NAME)/releases/download/R_$(subst .,_,$(PKG_VERS))
```

hgy59:
I would like to see this merged asap, as ejabberd depends on expat too (#4333).

cross/expat/Makefile
hgy59: There are a lot of github repos that provide there custom release packages (that e.g. do not need configure to generate the Makefile). The *.zip and *.tar.gz source packages are autogenerated by github and are often the wrong choice to build the packages.

publicarray: Thanks, Done.

publicarray: We about to hit 1k forks on this repo! :partying_face:

cross/libaom/Makefile
```diff
@@ -52,3 +52,5 @@ endif
 ifeq ($(findstring $(ARCH),$(i686_ARCHS)),$(ARCH))
 CMAKE_ARGS += -DAOM_TARGET_CPU=x86
 endif
+
+ENV += CXXFLAGS=-D_GLIBCXX_USE_C99_MATH
```
th0ma7, line 56: It does not build under `armv7`.  Also I propose using standard CMAKE calls.
I propose the following instead, added :
```

# libaom specific cmake options
CMAKE_ARGS += -DCONFIG_AV1_ENCODER=1

CMAKE_ARGS += -DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_C99_MATH
ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
CMAKE_ARGS += -DENABLE_NEON=OFF
endif
endif
```

cross/tvheadend/Makefile
```diff
@@ -1,11 +1,11 @@
...
 PKG_EXT = tar.gz
-PKG_DOWNLOAD_METHOD = git
 PKG_GIT_HASH = 9ed76c0a176b055a57b6e8bd2e0b6e29409269a9
-PKG_DIST_SITE = https://github.com/tvheadend/tvheadend.git
-PKG_DIST_FILE = $(PKG_NAME)-git$(PKG_GIT_HASH).$(PKG_EXT)
-PKG_DIR = $(PKG_NAME)-git$(PKG_GIT_HASH)
+PKG_DIST_NAME = $(PKG_GIT_HASH).$(PKG_EXT)
+PKG_DIST_SITE = https://github.com/tvheadend/tvheadend/archive
+PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIR = $(PKG_NAME)-$(PKG_GIT_HASH)
```
th0ma7, line 9: No no no!!!
github 4.3 package dates from 2017.
`tvheadend` does not provide package version anymore.
We must use a git hash specific snapshot.
As such I'm using the date in conjunction with the git hash to generate a proper version schema.
Please revert this.

cross/tvheadend/digests
```diff
@@ -1,3 +1,3 @@
-tvheadend-git9ed76c0a176b055a57b6e8bd2e0b6e29409269a9.tar.gz SHA1 821634cc8a40375ab22885032ee959a81de11970
-tvheadend-git9ed76c0a176b055a57b6e8bd2e0b6e29409269a9.tar.gz SHA256 1378fcd819011b3161dc01ccc9b068099041cf4af7550777a6f1e64cacbaec41
-tvheadend-git9ed76c0a176b055a57b6e8bd2e0b6e29409269a9.tar.gz MD5 70f71d73fc5bb1da2253764109d327cc
+tvheadend-4.3.tar.gz SHA1 6d8f86a1f4a9318537b06d55a9aea76245c961c0
+tvheadend-4.3.tar.gz SHA256 d9ea0d1d1596efc5d6fed1595391a7cb2fc2442aa292fe0b57b9ab6910b32046
+tvheadend-4.3.tar.gz MD5 ff0b7102c72be18146f3f44051d2270e
```
th0ma7, line 3: Same as previous.

spk/tvheadend/Makefile
```diff
@@ -2,9 +2,9 @@ SPK_NAME = tvheadend
 SPK_SHORT_VERS = 4.3
 SPK_GIT_HASH = 9ed76c0
 SPK_GIT_DATE = 20201011
-SPK_VERS = $(SPK_SHORT_VERS).$(SPK_GIT_DATE)~$(SPK_GIT_HASH)
+SPK_VERS = $(SPK_SHORT_VERS).$(SPK_GIT_DATE)
```
th0ma7, line 5: Revert this for previous behavior

th0ma7 (commented):
Here's a first review.
Still compiling in the background in the lookout for issues.

th0ma7:
As for `chromaprint`, adding `CMAKE_ARGS += -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations"` helps significantly under DSM7 but still fails.  I believe the code change may be relatively easy to figure out to get this one going as well.

spk/tvheadend/Makefile
publicarray: I can revert everything but this. dsm7 requires digits not letters or symbols. I think I copied the error message in the git commit message:

> Failed to load pkg info, version=[4.3.20201011~9ed76c0-26], correct-format=[^\d+(\.\d+){0,5}(-\d+)?$]

We still have revision numbers and the date

hgy59:
as #4539 is now on the master branch, you have to remove ` >> ${INST_LOG}`  in `service-setup.sh` for all service_* functions to work.

spk/tvheadend/Makefile
th0ma7: I guess the hash could be discarded entirely so only the date is being kept.
Or add a bit of fun with converting hex to decimal using `echo $((16#9ed76c0))`

Still, the git hash needs to be kept for  the help web page in order to provide actual version from:
`TVH_VERS = $(SPK_SHORT_VERS)~$(SPK_GIT_HASH)`

spk/tvheadend/src/service-setup.sh
```diff
@@ -6,7 +6,7 @@ HTTPP=9981
 HTSPP=9982
 
 # Replace generic service startup, run service in background
-GRPN=`id -gn ${EFF_USER}`
+GRPN=$(id -gn ${EFF_USER})
 HOME_DIR="${SYNOPKG_PKGDEST}/var"
```
th0ma7, line 10: Looking further at the code without mentioning the backup functionality that needs reworking eventually
(as inherited from original package and not working so much)

Isn't there a `${SYNOPKG_PKGVAR}` variable?

spk/tvheadend/src/service-setup.sh
```diff
@@ -72,13 +66,13 @@ service_preupgrade ()
         done
         # Move recording directories from package root into var
         UPGRADE_CFG_DIR="${SYNOPKG_PKGDEST}/var/dvr/config"
```
th0ma7, line 68: Same here, `${SYNOPKG_PKGVAR}` can probably be used.
Also, same variable is also used below so probably best to move outside of function for using everywhere.

spk/tvheadend/src/service-setup.sh
```diff
@@ -72,13 +66,13 @@ service_preupgrade ()
...
-            DVR_DIR=`grep -e 'storage\":' ${file} | awk -F'"' '{print $4}'`
-            REAL_DIR=`realpath "${DVR_DIR}"`
-            TRUNC_DIR=`echo "${REAL_DIR}" | cut -c9-28`
-            CHECK_VAR=`echo "${REAL_DIR}" | cut -c29-32`
+            DVR_DIR=$(grep -e 'storage\":' ${file} | awk -F'"' '{print $4}')
+            REAL_DIR=$(realpath "${DVR_DIR}")
+            TRUNC_DIR=$(echo "${REAL_DIR}" | cut -c9-28)
+            CHECK_VAR=$(echo "${REAL_DIR}" | cut -c29-32)
             # Replace any remaining target links in log files first
             sed -i -e "s,/var/packages/tvheadend/target,/usr/local/tvheadend,g" "${file}"
```
th0ma7, line 77: And here it would be better off using `${SYNOPKG_PKGDEST}` par of sed.

spk/tvheadend/src/service-setup.sh
```diff
@@ -93,34 +87,34 @@ service_postupgrade ()
...
-        $CP "${SAVE_DIR}"/. "${HOME_DIR}" >> ${INST_LOG}
+        echo "Restoring recordings into ${HOME_DIR}"
+        $CP "${SAVE_DIR}"/. "${HOME_DIR}"
         $RM "${SAVE_DIR}"
     fi
 
     # Need to enforce correct permissions for recording directories on upgrades
-    echo "Adding ${GROUP} group permissions on recording directories:"  >> ${INST_LOG}
+    echo "Adding ${GROUP} group permissions on recording directories:"
     UPGRADE_CFG_DIR="${SYNOPKG_PKGDEST}/var/dvr/config"
```
th0ma7, line 97: Duplicate `UPGRADE_CFG_DIR`

spk/tvheadend/src/service-setup.sh
```diff
@@ -93,34 +87,34 @@ service_postupgrade ()
...
     if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
-        synoacltool -del "${SYNOPKG_PKGDEST}" >> ${INST_LOG} 2>&1
-        chown ${EFF_USER}:root "/var/packages/tvheadend" >> ${INST_LOG} 2>&1
-    else
-        chown ${EFF_USER}:${USER} "/var/packages/tvheadend/target" >> ${INST_LOG} 2>&1
+        synoacltool -del "${SYNOPKG_PKGDEST}"
+        chown ${EFF_USER}:root "/var/packages/tvheadend"
+    elif [ $SYNOPKG_DSM_VERSION_MAJOR == 6 ]; then
+        chown ${EFF_USER}:${USER} "/var/packages/tvheadend/target"
         set_unix_permissions "${SYNOPKG_PKGDEST}/var"
```
th0ma7, line 118: Again usage of `${SYNOPKG_PKGDEST}` and `${SYNOPKG_PKGVAR}` would be best.

spk/tvheadend/src/service-setup.sh
```diff
@@ -93,34 +87,34 @@ service_postupgrade ()
...
-    echo "Restore '${EFF_USER}' unix permissions on package system directories" >> ${INST_LOG}
+    echo "Restore '${EFF_USER}' unix permissions on package system directories"
     if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 6 ]; then
-        synoacltool -del "${SYNOPKG_PKGDEST}" >> ${INST_LOG} 2>&1
-        chown ${EFF_USER}:root "/var/packages/tvheadend" >> ${INST_LOG} 2>&1
-    else
-        chown ${EFF_USER}:${USER} "/var/packages/tvheadend/target" >> ${INST_LOG} 2>&1
+        synoacltool -del "${SYNOPKG_PKGDEST}"
+        chown ${EFF_USER}:root "/var/packages/tvheadend"
+    elif [ $SYNOPKG_DSM_VERSION_MAJOR == 6 ]; then
```
th0ma7, line 116: Looking at things, there hasn't been any pre-DSM6 packages released.  All that code can be removed entirely.

spk/tvheadend/src/service-setup.sh
```diff
@@ -19,48 +19,42 @@ GROUP=sc-media
...
-    BIN=${SYNOPKG_PKGDEST}/bin
-    $BIN/busybox --install $BIN
-    $BIN/delgroup "${USER}" "users" >> ${INST_LOG}
-    $BIN/deluser "${USER}" >> ${INST_LOG}
 }
 
 service_preupgrade ()
 {
     # For backwards compatibility on DSM6
     if [ $SYNOPKG_DSM_VERSION_MAJOR -gt 5 ]; then
```
th0ma7, line 41: DSM5 is not supported anymore in this package.

cross/tvheadend/Makefile
```diff
@@ -1,11 +1,11 @@
...
 PKG_VERS = 4.3
 PKG_EXT = tar.gz
-PKG_DOWNLOAD_METHOD = git
 PKG_GIT_HASH = 9ed76c0a176b055a57b6e8bd2e0b6e29409269a9
-PKG_DIST_SITE = https://github.com/tvheadend/tvheadend.git
+PKG_DIST_NAME = $(PKG_GIT_HASH).$(PKG_EXT)
+PKG_DIST_SITE = https://github.com/tvheadend/tvheadend/archive
 PKG_DIST_FILE = $(PKG_NAME)-git$(PKG_GIT_HASH).$(PKG_EXT)
-PKG_DIR = $(PKG_NAME)-git$(PKG_GIT_HASH)
+PKG_DIR = $(PKG_NAME)-$(PKG_GIT_HASH)
```
th0ma7, line 8: Not sure I follow, will this download the old 4.3 version or the version matching the git hash?
As I never saw that option of downloading a tar.gz from a git hash though the archive URL path.

th0ma7 (commented):
I noticed a few variables needing adjusting in the code below in order to get closer to standard.
Perhaps a bit of it is out of scope of this PR... 
But timing would be good I believe to go over them now.
Let me know what you think.

spk/tvheadend/src/service-setup.sh
publicarray: This `sed` command will work in DSM7 but the path won't be available. Why is it needed anyway? Why can't the app do its thing in `/var/packages/tvheadend`?

cross/tvheadend/Makefile
publicarray: It will download the state the repository is in at particular hash.

If you go to https://github.com/tvheadend/tvheadend/tree/9ed76c0a176b055a57b6e8bd2e0b6e29409269a9 and hit the green `code button` and hover over `download zip` you should see the URL I used. (I just changed the file extension)

publicarray:
@th0ma7 If you want you can add #4551 to this PR\*, so we can see green :heavy_check_mark: s  (I initially thought that it was more difficult, so I did not even attempt it, shame on me :) )

\* just checkout this branch and add your commits (I use [github's cli](https://cli.github.com/): `gh pr checkout 4545`)

cross/tvheadend/Makefile
th0ma7: thnx, nice to know.
I guess we should change other package in using that same method instead of git.

spk/tvheadend/src/service-setup.sh
```diff
@@ -19,108 +19,94 @@ GROUP=sc-media
 service_postinst ()
 {
     # Encrypt password
-    wizard_password=`echo -n "TVHeadend-Hide-${wizard_password:=admin}" | openssl enc -a`
+    wizard_password=$(echo -n "TVHeadend-Hide-${wizard_password:=admin}" | openssl enc -a)
 
     # Edit the password configuration according to the wizard
-    sed -i -e "s/@password@/${wizard_password}/g" ${SYNOPKG_PKGDEST}/var/passwd/a927e30a755504f9784f23a4efac5109
+    sed -i -e "s/@password@/${wizard_password}/g" ${SYNOPKG_PKGVAR}/passwd/a927e30a755504f9784f23a4efac5109
```
th0ma7, line 25: I just saw a post in my email (that I can no longer find) about:
`${SYNOPKG_PKGDEST}/var` vs `${SYNOPKG_PKGVAR}`
Where, depending of DSM version it points to `package/target/var` or just `package/var`.
Now I wonder, are we good with the changes in here?

th0ma7 (approved):
I've included PR #4551 directly in this PR with hope to get all green flags.
Other than the `${SYNOPKG_PKGDEST}/var` vs `${SYNOPKG_PKGVAR}` variable below which I now have doubts, everything else looks great!
Nice team work!

th0ma7:
@publicarray Awesome! There now only `ejabberd` failing for all arches but otherwise build is succesfull for all `ffmpeg` related packages.  woot!

hgy59:
@th0ma7 ejabberd is WIP in #4333 and waiting for `cross/expat` download fix with this PR.

hgy59:
regarding `SYNOPKG_PKGVAR` we have to pay attention.
This variable is provided only by the DSM 7 context.
And it is defined in the generic installer for use with other DSM versions too.

But `SYNOPKG_PKGVAR` is still restricted to DSM 7 in custom start-stop-status scripts or when a custom installer is used.

spk/tvheadend/src/service-setup.sh
th0ma7: Honestly I have no clue.
I guess this `sed` could just be removed entirely.

spk/tvheadend/src/service-setup.sh
publicarray: Yes, It's backwards compatible in the `service-setup` scrips only

publicarray: The setting isn't applied. It's another symptom of the change here: https://github.com/SynoCommunity/spksrc/commit/0eaccc5f0a45bfa91b505f7ed9750c987128c235#r49449742

spk/tvheadend/src/service-setup.sh
publicarray: I tried to find a sample in `SYNOPKG_PKGVAR}/dvr/log` but I don't see any. I guess I need to have a receiver and start a recording.

th0ma7: But I do :)

Got the following on my DSM6:
```
root@th0ma7-nas:/var/packages/tvheadend/target/var/dvr# tree -d
.
├── autorec
├── config
└── log

3 directories
```
Honestly I've been wishing to rework all this for a year but never took the time to do so.  The backup/recovery process does not work, period.

1. For instance the `/var/packages/tvheadend/target/var/config` references to `/usr/local/tvheadend` should be `sed`ed to `/var/package/tvheadend` (clearly the "`binary matches`" isn't good in the grep below):
```
# grep -R "/usr/local/tvheadend" *
Binary file bin/tvheadend matches
var/channel/config/65ab00d40fe4addffdf4babbc3b390d4:	"icon": "file://usr/local/tvheadend/.xmltv/",
var/channel/config/f87d3e9167eceb6d6f13e9ab7739f0cb:	"icon": "file://usr/local/tvheadend/.xmltv/",
var/channel/config/0cdf1b095014ab7990f9c8569a69270b:	"icon": "file://usr/local/tvheadend/.xmltv/",
var/channel/config/1fbe2fe151b5c6d9f555a69ce61af02f:	"icon": "file://usr/local/tvheadend/.xmltv/",
var/channel/config/3b1392ac0539b5b4a400b0d81d4e34eb:	"icon": "file://usr/local/tvheadend/.xmltv/",
var/config:	"chiconpath": "file://usr/local/tvheadend/.xmltv/",
var/config:	"muxconfpath": "/usr/local/tvheadend/share/tvheadend/data/dvb-scan/",
```
2. This entire directory (e.g. `/var/packages/tvheadend/target/var`) needs to be backup/restored along with creating a recovery `something-config-backup-date.tar.bz` in order to recover as needed.
My personal backup script: https://raw.githubusercontent.com/th0ma7/synology/master/tvheadend-backup.sh
3. There is a caching directory `/var/packages/tvheadend/target/cache` that needs to be recovered as well.  It would benefit from being located elsewhere under `/volume1/...` from a user defined param at installation time.
```
# du -hs cache var
24M	cache
14M	var
```

4. And also having a few scripts & configs that needs to be added manually that I should add to the default package...

BTW, just got my `tvheadend` and `ffmpeg` packages ready for DSM7... in theory ready for testing :)

th0ma7: And the good: both packages install.
And the bad: Got a `503 Forbidden` after the password prompt when accessing the `http://<host>:9981/extjs.html` so guessing the password entry capture didn't worked as expected...

publicarray: @th0ma7 `503 Forbidden` is fixed with 72be92802f353cdf43a8108669dedf983c84d8bb

publicarray: I've just rebased to get the changes from master. If you want to try again.

th0ma7: tried again but still `503 Forbidden`.
Also, I found what's really missing: setting the default storage option for recording.
This would allow removing almost the entire script and only focus on backuping the configs.
But, that could come into a later PR that I could work on once this one is in.

publicarray: Did you do a clean install/uninstall? Since the rebase I can always login with `admin:admin`

th0ma7: I ensured to do a `git pull` and rebuilt the `tvheadend` package for `kvmx64`, fully uninstall/install (there is a new option now to completely wipe the app), kept passwd to default (`admin`) and still no luck.
I'll try to investigate this further later on.

m4tt075:
@publicarray I just stumbled over this PR by accident. As you know, there are some "backwards compatibility" code chunks in the TVH `service-setup.sh` to recover and move recordings from the package root directory into the var directory.

Here the "history" on that: Years ago, I once produced a "broken" TVH version, that (by default and incorrectly) recorded into the root directory. When we realized the mistake, we published a corrected version containing all the "backwards compatibility" snippets. The faulty version was only online for a pretty limited timeframe.

I highly doubt that after all this time there are many (if any) faulty installations left, which never have been upgraded and where the recordings are still displaced. As such, if you want to, I believe that you can safely remove them. Happy maintaining...   ;-)

spk/tvheadend/src/service-setup.sh
th0ma7: I found what the issue is... It relates to the changes of `${SYNOPKG_PKGDEST}` vs `${SYNOPKG_PKGVAR}`...

Directories `accesscontrol` and `passwd` are now located under `/var/packages/tvheadend/target/var -> /volume1/@appstore/tvheadend/var`) where the `password/a927e30a755504f9784f23a4efac5109` looks like:
```
{
	"enabled": true,
	"username": "admin",
	"password2": "@password@",
	"wizard": true
}
```
Moving both directories (`accesscontrol` and `passwd`) to `/var/packages/tvheadend/var -> /volume1/@appdata/tvheadend`) then manually fixing the password solved the isse:
```
mv /volume1/@appstore/tvheadend/var/* /volume1/@appdata/tvheadend/.
wizard_password=$(echo -n "TVHeadend-Hide-${wizard_password:=admin}" | openssl enc -a)
sed -i -e "s/@password@/${wizard_password}/g" /var/packages/tvheadend/var/a927e30a755504f9784f23a4efac5109
```

th0ma7:
@m4tt075 thnx for the info and worth will definitively have another look for a subsequent release :)
On a similar note I intend to pre-adjust recording directory by asking the user at installation time directly...

th0ma7:
@publicarray it looks like when you rebased you discarded my chromaprint changes as it fails to build once again.

publicarray:
Sorry @th0ma7, adding it back now.

spk/tvheadend/src/service-setup.sh
th0ma7: @publicarray reminder on this.
I've looked at compile options but can't see something obvious that allows moving around theses two directories...
I'll try to have another look at it later and see if I can come-up with a fix somehow.
```Installation Paths
  --prefix=DIR$                  Installation root [/usr/local]
  --bindir=DIR                   Install binaries in DIR [${prefix}/bin]
  --libdir=DIR                   Install libraries in DIR [${prefix}/lib]
  --mandir=DIR                   Install man pages in DIR [${datadir}/man]
  --datadir=DIR                  Install data files in DIR [${prefix}/share]

```

th0ma7: @publicarray and @hgy59 I believe I have found where the issue is.
And may have found a second one actually... or food for thought.

**First issue:**
It relates to `$(STAGING_DIR)` where the entire content is being `tgz` and ends-up being installed over to `/var/packages/tvheadend/target/` (_appstore_).  Although it happens that `work-*/staging/var` containing both `accesscontrol` and `passwd` that should follow the $(SYNOPKG_PKGVAR) and end-up into `/var/packages/tvheadend/var` (_appdata_) where all the automatically generated files also end-up going.

An idea to circumvent this:
1. From the `Makefile`, install files under `staging/appdata` using a new variable such as `$(STAGING_SPKVAR)`.
2. installer script go over all files & directories in `staging/appdata`.  If `install` just copy to `/var/packages/tvheadend/var` else perhaps something as trivial as a `rsync` no-overwrite over the directory (which could be used for all cases actually).  Depending of DSM the destination folder is set from `SYNOPKG_PKGVAR` so it should always work.


**Second issue: (not sure about this one)**
I wonder if it would make sense to extend solution proposed for the first issue to add variables for `$(STAGING_SPKHOME)`, `$(STAGING_SPKETC)` and `$(STAGING_SPKTEMP)` and use `rsync` to copy files over the end-state location?

th0ma7: Or a simple work-around is to move the accesscontrol & passwd directory creation and file copying at the service-setup.sh level.

th0ma7: @hgy59 and @publicarray Got a working solution.
Will provide a separate PR to address this later today.

th0ma7:
While reworking PR #4579 I noticed the following in the logs:
```
2021/04/26 17:29:18 Fixing fontconfig links
2021/04/26 17:29:18 /bin/ln: failed to create symbolic link '/volume1/@appstore/tvheadend/etc/fonts/conf.d/*.conf': No such file or directory
2021/04/26 17:29:18 End service_postinst
```
which is a result of the following in `service_postinst`:
```
    # Fix fontconfig links
    CONFD_DIR="${SYNOPKG_PKGDEST}/etc/fonts/conf.d"
    FONTS_DIR="${SYNOPKG_PKGDEST}/share/fontconfig/conf.avail"
    echo "Fixing fontconfig links"
    for FONT_FILE in ${CONFD_DIR}/*.conf
    do
        FONT_NAME=$(basename "${FONT_FILE}")
        $LN "${FONTS_DIR}/${FONT_NAME}" "${CONFD_DIR}/${FONT_NAME}"
    done
```
It seems that somewhere somehow the `fontconfig` dependency got removed, which results as well in lacking `freetype`.  Not sure how much mandatory this is for TVH but this "might" explain inability to handle special characters I'm seeing from OTA channel capture due to lack of well supporting ISO-8859-1 Latin1 in some circumstances.

th0ma7 (approved):

th0ma7:
@publicarray and chance we can revisit this and complete this PR?  (I recall you where going to be away for a while)

EngMarc:
@th0ma7 let me know when you have something for testing on my DS1817+ that I just upgraded to DSM 7.

th0ma7:
@publicarray I'll be revisiting this PR over the next few days, starting with rebasing in order to get this merged asap considering the increasing noise over DSM7.

@hgy59 and @ymartin59 would be great if you can find cycles to go over it and do another round of review to confirm all looks OK?

th0ma7 (commented):
I now recall where we left...
Basically the PR was ready, besides this.

Issue is on DSM7: the following files gets installed in `/var/packages/tvheadend/target/var` while all the tvheadend tree ends-up going under `/var/package/tvheadend/var`.

As such, it doesn't work as the web server can't find the `access` and `passwd` files.

We need #4579 (or any other viable solution) to solve this.

th0ma7:
As this solves all build errors I recommend moving ahead and to merge it.
Unit testing is looking good and I'm planning for an upgrade of both ffmpeg (#4576) and tvheadend (#4680 + another coming shortly) anyway so additional testing will take place.
Work on #4579 is not mandatory for this portion to get merged to master.

@hgy59 and/or @ymartin59 are you ok with this PR?

ymartin59 (approved):

PR #4549, opened 2021-04-11
fix build and update rsnapshot

hgy59:
_Motivation:_  ~Add native dependencies to fix build of rsnapshot package~ Update and fix dependencies
_Linked issues:_  follow up of #4539

### Checklist


th0ma7:
@hgy59 Out of curiosity, I wonder why does it require a native `openssh` and `rsync` ?  Couldn't the docker debian default work?

hgy59:
> 
> 
> @hgy59 Out of curiosity, I wonder why does it require a native `openssh` and `rsync` ? Couldn't the docker debian default work?

Yes we could add it to the development-environment and the spksrc-docker-image.
As @ymartin59 always wanted to keep this image small, I already added native/rsync and native/zip required for rabbitmq some time ago.

But let us discuss it here: where should we set the limit of what tools are added to the development environment (i.e. the spksrc docker image) and what not?
- rsync
- zip
- (open)ssh
- cmake
- libicu
- ruby
- rust
- meson
- dotnet-sdk 
- java
- node
- file (this tool must have the same version as cross/file, so it might be better to add native/file)
-

publicarray:
I'd like to keep is small as well. I'd rather we update native/*libs. For me storage is Cheep. Internet bandwidth not so much. I update the container when I feel like so if i understand correctly there is the possibility that I could build with outdated libs. Which would be quite bad for packages that use native/openssl.

hgy59:
for rsnapshot this discussion is obsolete as the build dependencies could be removed.

PR #4554, opened 2021-04-14
Add lldpd spk package

ixs:
Add a lldpd package that runs the lldp daemon from <http://lldpd.github.io> allowing the synology device to be discovered on the network.

### Checklist


ixs:
@hgy59  I have been using it with < 7 but when after the upgrade, I stumbled over the problem you described and did not get around to fixing things yet.
Your change otherwise looks pretty good.

What is needed for the privilege file?

hgy59:
> What is needed for the privilege file?

After installation (and after repair or upgrade) the privilege file must be modified to replace the user definition "package" by "root".

To do so call `sudo sed -i 's/package/root/g' /var/packages/lldpd/conf/privilege` and finally start the package with `sudo synopkg start lldpd`.

But this needs further validation. For lldpd to work under DSM 7, I manually started the daemon once like this:
`/var/packages/lldpd/target/bin/lldpd -d`
Then (when no error was logged) terminated with Ctl+C and started with synopkg.

So there are some pending tasks:
- [ ] add wizard page for DSM 7 with instructions to run the package as root
- [ ] evaluate wheter the initial manual start is required or how it can be fixed otherwise
- [ ] didn't find a solution to get logs from lldpd

PR #4558, opened 2021-04-14
DSM7: revert var overwrite

hgy59:
_Motivation:_  Revert a breaking change of #4539

### Checklist

### Remarks
- revert overwriting files in ${SYNOPKG_PKGVAR} folder from target/var (introduced with #4539)

_Reference:_ [comment on spksrc.service.installer.dsm7 in  #4539](https://github.com/SynoCommunity/spksrc/commit/0eaccc5f0a45bfa91b505f7ed9750c987128c235?branch=0eaccc5f0a45bfa91b505f7ed9750c987128c235&diff=split#r49449742)

As linking into the merged PR is not very good, here is the comment of @publicarray 
> Dang I missed it, why change it? This is a regression or please let me explain how it was supposed to work.
> 
> DSM7:
> 
> `${SYNOPKG_PKGDEST}/var` => /var/packages/{package}/target/var
> `${SYNOPKG_PKGVAR}` => /var/packages/{package}/var
> 
> DSM6:
> `${SYNOPKG_PKGDEST}/var` => /var/packages/{package}/target/var
> `${SYNOPKG_PKGVAR}` => /var/packages/{package}/target/var or undefined (need to double-check)
> 
> If `${SYNOPKG_PKGDEST}/var` exists we want its contents to be coped to `${SYNOPKG_PKGVAR}` unless there are already files there. (think new dsm7 installation or upgrading from dsm6 to dsm7)
> 
> And we don't want to override the on next install since there are updated database/user files in there.
> 
> So please don't publish any DSM7 packages yet.

### DIFF
--- a/mk/spksrc.service.installer.dsm7
+++ b/mk/spksrc.service.installer.dsm7
@@ -167,7 +167,7 @@ postinst ()
     # copy target/var data to permanent storage
     # and don't override old configurations
     if [ -d ${SYNOPKG_PKGDEST}/var ] && [ "$(find ${SYNOPKG_PKGVAR} -mindepth 1 -not -name '*.log' -print)" = "" ]; then
-        $CP -RT ${SYNOPKG_PKGDEST}/var/. ${SYNOPKG_PKGVAR}
+        $CP -RT ${SYNOPKG_PKGDEST}/var/. ${SYNOPKG_PKGVAR} 2>&1 | install_log
     fi    
 
     call_func "service_postinst" install_log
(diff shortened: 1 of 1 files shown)


hgy59:
@publicarray just wondering what the `-RT` is needed for in `$CP -RT` .

as `CP=/bin/cp -rfp` the `-R` is obsolete as equal to `-r`
and `-T` means 'treat target as file` (--no-target-directory) that looks strange for me in this context.

publicarray:
@hgy59 I don't remember why, and I just tried it without the flags, and it still works. So I'm happy if you remove them here and on this line too.

https://github.com/SynoCommunity/spksrc/blob/0eaccc5f0a45bfa91b505f7ed9750c987128c235/mk/spksrc.service.installer.dsm7#L221

publicarray:
If you do want to override files it is possible by just removing the find statement all together.

PR #4562, opened 2021-04-15
Use installer resources for DSM6, update TCL package

hgy59:
_Motivation:_  Another Framework cleanup and a package update for TCL
_Linked issues:_  #4524, #4539

### Checklist

### Remarks
1. Revert `SPK_LINKS` Makefile variable and use `SPK_USR_LOCAL_LINKS` for DSM6 and DSM5.
2. Update TCL and make use of SPK_LINKS obsolete, and add 64-bit support for amd64 archs
3. Adjust remaining packages to use `SPK_USR_LOCAL_LINKS`  instead of `SPK_LINKS`
4. Redesign DSM6 installer to use resource definitions for Links, FwPorts, SHARES similar to DSM7
5. Only keep creation of `/usr/local/{package}` link to avoid breaking already installed packages that lean on this

### References
- usr-local-linker for DSM6: https://github.com/SynoCommunity/spksrc/pull/4539#discussion_r611113503

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -60,6 +60,12 @@ else ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 SPK_USER = $(SPK_NAME)
 endif
 
+ifeq ($(strip $(STARTABLE)),yes)
+# we only evaluate for STARTABLE=no
+# STARTABLE=yes is default (same as STARTABLE not defined)
+STARTABLE=
+endif
+
 # Recommend explicit STARTABLE=no
 ifeq ($(strip $(SSS_SCRIPT) $(SERVICE_COMMAND) $(SERVICE_EXE) $(STARTABLE)),)
 ifeq ($(strip $(SPK_COMMANDS) $(SPK_USR_LOCAL_LINKS)),)
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -177,6 +177,8 @@ ifneq ($(strip $(SERVICE_WIZARD_SHARE)),)
 		'."data-share" = {"shares": [{"name": $$share, "permission":{"rw":[$$user]}} ] }' $@ 1<>$@
 endif
 SERVICE_FILES += $(DSM_CONF_DIR)/resource
+
+# Less than DSM 6.0
 else
 ifneq ($(strip $(SPK_COMMANDS) $(SPK_USR_LOCAL_LINKS)),)
 	@echo "# List of commands to create links for" >> $@
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -60,12 +60,6 @@ else ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 SPK_USER = $(SPK_NAME)
 endif
 
-ifeq ($(strip $(STARTABLE)),yes)
-# we only evaluate for STARTABLE=no
-# STARTABLE=yes is default (same as STARTABLE not defined)
-STARTABLE=
-endif
-
 # Recommend explicit STARTABLE=no
 ifeq ($(strip $(SSS_SCRIPT) $(SERVICE_COMMAND) $(SERVICE_EXE) $(STARTABLE)),)
 ifeq ($(strip $(SPK_COMMANDS) $(SPK_USR_LOCAL_LINKS)),)
@@ -183,8 +177,15 @@ ifneq ($(strip $(SERVICE_WIZARD_SHARE)),)
 		'."data-share" = {"shares": [{"name": $$share, "permission":{"rw":[$$user]}} ] }' $@ 1<>$@
 endif
 SERVICE_FILES += $(DSM_CONF_DIR)/resource
+else
+ifneq ($(strip $(SPK_COMMANDS) $(SPK_USR_LOCAL_LINKS)),)
+	@echo "# List of commands to create links for" >> $@
+	@echo "SPK_COMMANDS=\"${SPK_COMMANDS}\"" >> $@
+	@echo "SPK_USR_LOCAL_LINKS=\"${SPK_USR_LOCAL_LINKS}\"" >> $@
+	@cat $(SPKSRC_MK)spksrc.service.create_links >> $@
 ifneq ($(findstring conf,$(SPK_CONTENT)),conf)
 SPK_CONTENT += conf
+endif
 endif
 
 # Less than DSM 6.0
--- a/spk/transmission/src/wizard/install_uifile
+++ b/spk/transmission/src/wizard/install_uifile
@@ -1,69 +1,125 @@
-[{
-    "step_title": "Basic configuration",
-    "items": [{
-        "type": "textfield",
-        "desc": "Download directory",
-        "subitems": [{
-            "key": "wizard_download_dir",
-            "desc": "Download directory",
-            "defaultValue": "/volume1/downloads",
-            "validator": {
-                "allowBlank": false,
-                "regex": {
-                    "expr": "/^\\\/volume\\w*[0-9]{1,2}\\\/[^<>: */?\"]*/",
-                    "errorText": "Path should begin with /volumename?/ where volumename can be 'volume' or also 'volumeUSB' and ? is the volume number (1-99)."
-                }
+[
+    {
+        "step_title": "Basic configuration",
+        "items": [
+            {
+                "type": "combobox",
+                "desc": "Please select a volume to use for the download folder",
+                "subitems": [
+                    {
+                        "key": "wizard_volume",
+                        "desc": "volume name",
+                        "displayField": "display_name",
+                        "valueField": "volume_path",
+                        "editable": false,
+                        "mode": "remote",
+                        "api_store": {
+                            "api": "SYNO.Core.Storage.Volume",
+                            "method": "list",
+                            "version": 1,
+                            "baseParams": {
+                                "limit": -1,
+                                "offset": 0,
+                                "location": "internal"
+                            },
+                            "root": "volumes",
+                            "idProperty": "volume_path",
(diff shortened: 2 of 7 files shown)


mk/spksrc.service.mk
```diff
@@ -182,6 +170,13 @@ ifneq ($(strip $(SERVICE_WIZARD_SHARE)),)
 		'."data-share" = {"shares": [{"name": $$share, "permission":{"rw":[$$user]}} ] }' $@ 1<>$@
```
publicarray, line 170: Something interesting to note:
> Since 7.0-41201, package center will create a symlink under `/var/packages/[package_id]/shares/` named by share folder pointing to share folder path.

mk/spksrc.service.mk
```diff
@@ -182,6 +170,13 @@ ifneq ($(strip $(SERVICE_WIZARD_SHARE)),)
...
 endif
 SERVICE_FILES += $(DSM_CONF_DIR)/resource
+else
```
publicarray, line 173: Can we make separation more obvious?
```suggestion

# Less than DSM 6.0
else
```

mk/spksrc.service.installer
```diff
@@ -1,12 +1,16 @@
 #!/bin/sh
```
publicarray, line 1: Should this file get a dsm6 postfix (for consistency) `spksrc.service.installer.dsm6` or can dsm6/7 share this file now? How much dsm 7 specific stuff is left anyway?

spk/rdiff-backup/Makefile
```diff
@@ -21,11 +21,6 @@ LICENSE  = GPLv2
 
 SERVICE_SETUP = src/service-setup.sh
 
-include ../../mk/spksrc.common.mk
-ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 SPK_USR_LOCAL_LINKS = bin:env/bin/rdiff-backup bin:env/bin/rdiff-backup-statistics
```
publicarray, line 24: Just curious, What makes you pick between `SPK_COMMANDS` vs `SPK_USR_LOCAL_LINKS = bin:` ? 

FYI when `ctl_stop` is absent or "Yes" the user can link and unlink the `/usr/local/*` stuff on package start/stop.

hgy59: with `SPK_COMMANDS` the folder is reused unter `/usr/local` i.e. `bin/file` will create `/usr/local/bin/file => /var/packages/{package}/target/bin/file` 

As `SPK_COMMANDS = env/bin/test` will create `/usr/local/bin/env/test => /var/packages/{package}/target/env/bin/test`
this is not expected, 
but with `SPK_USR_LOCAL_LINKS = bin:bin/env/test` you get
`usr/local/bin/test => /var/packages/{package}/target/env/bin/test`

SPK_COMMANDS = `sbin/test` will create a link
`/usr/local/sbin/test => /var/packages/{package}/target/sbin/test` (but I'm not shure whether this is in DSM7 the same).

mk/spksrc.service.mk
hgy59: I have to test shared folders with DSM6 anyway. I will look whether dsm6 resource worker will create such links too.

mk/spksrc.service.installer
hgy59: ✔️ 
current differences are:
- different path of FWPORTS_FILE
- save INST_VARIABLES is implemented in DSM6 only
- several syno_user/group functions are not (yet?) updated for DSM7
- DSM6 creates the folder `${SYNOPKG_PKGVAR}` and sets permissions
- DSM6 still creates the /usr/local/{package} folder in postinst and removes it in postuninst
- DSM7 has the default uninstall-wizard with the option to delete permanent storage data
- DSM6 removes the package user and group in postuninst
- DSM6 saves and restores the var folder between preupgrade and postupgrade
-

spk/rdiff-backup/Makefile
hgy59: Finally I must confess that we could remove the `SPK_COMMANDS` (but must adjust all packages that currently use it) with a little supplementary setting:
`SPK_COMMANDS = bin/file`
may be replaced by
`SPK_USR_LOCAL_LINKS = bin:bin/file`

hgy59: Or may be we can enhance `SPK_USR_LOCAL_LINKS` to behave like `SPK_COMMANDS` when no colon separater is used.

publicarray: Thanks for the info, yea AFAIK `sbin` is not supported with DSM7. 
We can add that to our long todo list for now since it's not critical and more of a cosmetic change.

mk/spksrc.service.installer
publicarray: Thanks, yea so sill a lot, Best keep them separate then :+1: 
Regarding syno_user/group functions I don't think they can ever be ported due to lower permissions. I'm fine with it since the UI from the DSM now has much better/fine-grained permissions through the "internal users"

publicarray (approved):
LGTM

hgy59:
> 
> 
> LGTM

I have to validate the `SHARE_PATH` (i.e. `SERVICE_WIZARD_SHARE`) still works for DSM 6

hgy59:
@publicarray I tend to redesign the shared folder handling for DSM6.

I did not succeed to mix full path (i.e /volume1/downloads) and shared folder name (i.e. downloads) with one installer.
Finally the breaking fact is that it is not possible to force back variables (i.e. wizard_shared_folder_name) into the enviroment so that the system uses it for the "data-share" resource definition.
I also tried to patch the conf/resource file with the share_name extracted from full path. The resource file exists at the final location in the postinst step and the patch was sucessful, but the installation failed and the name of the shared folder was not taken from the updated resource file _(it still took the name from the wizard variable when I defined both)_.
I even tried to use `/usr/syno/sbin/synopkghelper update ${package} data-share` ignoring that for data-share resources the documentation says it is not updatable (only port-configs seem to be updatable). Unsurprising error `Failed to update data-share of demoservice [0xD900 resource_api.cpp:331]`.
The synology dev docu says that the data-share resource worker activates resources at enable step but this seems to be wrong _(just another example that this docu is worth nothing)_.

The shared folder processing in the sc framework (i.e. with /volumeX/share_name) was good for DSM<=5.
With the sc framework we are suffering again on the incomplete implementation of the DSM6 development guide.

When creating a folder under /volumeX we do not really get a DSM shared folder _(this might be the cause for a lot of permission issues discussed here)_. Such folders are not displayed under shared folders and are deletable by the root user - shared folders cannot be deleted by root, but only within "Control Panel > Shared Folders".

IMHO we must migrate to "data-share" resource definitions for DSM6, 
This includes the requirement to adjust all wizards for packages that use shared folders.

The benefit would be real shared folders. And we could define default access for sc-download group in the resource file too.

Just another evening spent on digging into the swamp of DSM. 😩

publicarray:
@hgy59 I've pushed transmission package form dsm7-packages (feel free to remove it again). But I used this as my test package since I already did most of the work on it. In my tests it can install on a fresh dsm6 install and can upgrade from an older version of the transmission package on the dsm6. (both with the default settings/shared folder). I added a commit to set the `wizard_download_dir` variable when upgrading for the worker to succeed.

publicarray:
Hm, I tried the same method for the demoservice with, but I also get the error you describe above.

```
4]: share_default_get.c:347 failed to alloc share for [/volume1/downloads], errno=0x1200
2021-04-18T10:12:39+10:00 dsm6 synoscgi_SYNO.Core.Package.Installation_1_install[21904]: share_default_share_create.c:21 Failed to get defualt status for /volume1/downloads
2021-04-18T10:12:39+10:00 dsm6 synoscgi_SYNO.Core.Package.Installation_1_install[21904]: data_share.cpp:273 Failed to create /volume1/downloads [0x1200 share_alloc.c:68]
```

spk/transmission/src/wizard/install_uifile
```diff
@@ -1,69 +1,125 @@
...
+                        }
+                    }
+                ]
+            },
+            {
+                "type": "textfield",
+                "desc": "Download shared folder (using the volume chosen above)",
+                "subitems": [
+                    {
+                        "key": "wizard_download_dir",
```
hgy59, line 43: Aaah, that is the trick, to use separate wizard variables for the volume and the share name!

but this should work in the demoservice too.

publicarray: Only the folder name is used in the worker (not the full path), maybe we need to use `basename` before passing it on to `jq --arg share [..]`

publicarray: Example from documentation:
```json
"data-share": {
    "shares": [{
        "name": "music",
        "permission": {
             "ro": ["AudioStation"]
         }
    }]
}
```

publicarray: @hgy59  whichever way we go, (new wizards or patch to remove `volumeX` before passing the variable on) please remember to remove the if.
https://github.com/SynoCommunity/spksrc/blob/83f850f485c25db3bba65f80a6b95d57cb929b16/spk/demoservice/Makefile#L21-L25

ymartin59 (changes requested):
I do not appreciate such long PR with unrelated content.
Please split it into three distinct branches: links, tcl and transmission

hgy59:
so we gonna split this PR

hgy59:
@publicarray I would like to close this PR, as my work was merged with #4577 and #4578.
Just want to ask you, whether nothing is lost about your work for shared folders on transmission package here.

publicarray:
Thanks @hgy59 I'll hopefully get to it early next month. Just busy with work ATM

publicarray:
The transmission stuff is in #4719 closing this one.

PR #4566, opened 2021-04-18
Enhance checksum verification for multiple files

hgy59:
_Motivation:_  There are few packages that have arch specific downloads. Add support to validate such downloads with regular digests files.
_Linked issues:_  Separated (cherry picked) from #3932

### Checklist

### Remarks
This PR updates the checksum validation to work with digests files that have multiple downloaded files in it.
The creation of such digests files is not as easy as `make digests` as the make supports only one ARCH per call.
Therefor the creation of the digests file is specific for such packages.

- lidarr, radarr and libstdc++ have now digests files with multiple files
- those have a new target `update-digests` to create/update the digests file
- the digests file is now mandatory when using the regular download
- only the outdated cross/squidclamav is missing a digets file, but this package needs rework anyway, as the current version is not available for download anymore (v6.10 of 2015)


mk/spksrc.checksum.mk
```diff
@@ -39,18 +39,13 @@ pre_checksum_target: checksum_msg
 
 # validate file integrity with the provided digests.
 checksum_target: $(PRE_CHECKSUM_TARGET)
-	@if [ -f $(DIGESTS_FILE) ] ; \
-	then \
-	  cat $(DIGESTS_FILE) | (cd $(DISTRIB_DIR) && while read file type value ; \
+	@if [ -f $(DIGESTS_FILE) ] ; then \
+	  cat $(DIGESTS_FILE) | grep $(LOCAL_FILE) | ( \
+	  count=0 ; \
```
th0ma7, line 44: The `count` variable could be discarded entirely if there was a leading if statement to check if `grep` is not null.
As such there could be preceding this a simple check if/else such as:
```
if [ $(cat $(DIGESTS_FILE) | grep -c $(LOCAL_FILE) || true) -eq 0 ]; then
   $(MSG) "  Downloaded file $(LOCAL_FILE) is not in digests file"
else
   ...
```

th0ma7 (changes requested):
Proposed opportunity to simplify the code.

mk/spksrc.checksum.mk
hgy59: @th0ma7 thanks for the tip, I first wanted to cound for the 3 checksum types, but the regular checks do not count either.

hgy59: Finally I added the verification for at least 3 checksums.

th0ma7: Then again, the `count` does not provide that much as it does not validate each `md5`|`sha` types.
You could do instead, right from the get go:
```
# File missing in digests
if [ $(cat $(DIGESTS_FILE) | grep -c $(LOCAL_FILE) || true) -eq 0 ]; then
   $(MSG) "  Downloaded file $(LOCAL_FILE) is not in digests file"
   rm $(DOWNLOAD_COOKIE)
# Missing sha entries in digests
elif [ $(cat $(DIGESTS_FILE) | sed '/MD5/{/SHA1/{/SHA256/p;};}' | grep -c $(LOCAL_FILE) || true) -lt 3 ]; then
   $(MSG) "  Downloaded file $(LOCAL_FILE) has only $$count/3 checksum entries. Please update the digests file"
   rm $(DOWNLOAD_COOKIE)
# All looks good, proceed.
else
   cat $(DIGESTS_FILE) | grep $(LOCAL_FILE) | ( \
   ...
```

th0ma7 (changes requested):
Strongly suggesting validating the sha|md5 right from the start instead of using a counter.

mk/spksrc.checksum.mk
```diff
@@ -39,42 +39,47 @@ pre_checksum_target: checksum_msg
...
+	@if [ -f $(DIGESTS_FILE) ] ; then \
+	  # validate file in digests \
+	  if [ $$(cat $(DIGESTS_FILE) | grep -c $(LOCAL_FILE) || true) -eq 0 ]; then \
+	     $(MSG) "  Downloaded file $(LOCAL_FILE) is not in digests file" ; \
+	     exit 1 ; \
+	  # validate checksum entries in digests \
+	  elif [ $$(cat $(DIGESTS_FILE) | sed '/MD5/{/SHA1/{/SHA256/p;};}' | grep -c $(LOCAL_FILE)) -lt 3 ]; then \
+	     $(MSG) "  Downloaded file $(LOCAL_FILE) has less than 3 checksum entries. Please update the digests file" ; \
+	     exit 1 ; \
+	  # validated, proceed \
```
th0ma7, line 51: I was struggling making this work on my side... missing was the double `$` such as `$$(cat ...)`.
Great work!

mk/spksrc.checksum.mk
```diff
@@ -39,42 +39,47 @@ pre_checksum_target: checksum_msg
...
-	      [ -f $$file.wrong ] && rm $$file.wrong ; \
-	      mv $$file $$file.wrong ; \
-	      $(MSG) "    Renamed as $$file.wrong" ; \
-	      rm $(DOWNLOAD_COOKIE) ; \
-	      $(MSG) "    Download cookie removed to trigger the download again" ; \
-	      exit 1 ; \
-	    fi \
-	  done) ; \
+	@if [ -f $(DIGESTS_FILE) ] ; then \
+	  # validate file in digests \
```
th0ma7, line 43: Last suggestion, as all the `exit` cases are on top, this one could be using a similar approach and benefit from a `exit 1` like the others afterwards:
```
@if [ ! -f $(DIGESTS_FILE) ] ; then \
	  $(MSG) "No digests file for $(NAME)" ; \
	  exit 1 ; \
else \
```

th0ma7 (approved):
Last code change proposal otherwise great work!

mk/spksrc.checksum.mk
hgy59: This is not implemented with `exit 1` now and I doubt that we should change this.
The only package I found without digests file is cross/squidclamav. I wanted to add one, but the current source package is too old and not available for download anymore.

th0ma7:
LGTM.  Will rebase my PR #4144 using this when ever you feel ready to merge.

ymartin59:
@hgy59  @th0ma7  There is an ugly "copy-paste" pattern here for "update-digests" target... it would make sense to move it to a common makefile script in `mk/` folder

hgy59:
> 
> 
> @hgy59 @th0ma7 There is an ugly "copy-paste" pattern here for "update-digests" target... it would make sense to move it to a common makefile script in `mk/` folder

@ymartin59 you are absolutely right. I just "found" a generic solution.

PR #4575, opened 2021-04-24
generic solution for multifile digests (followup of #4566)

hgy59:
_Motivation:_  The previous solution was not handy and has duplicated code.
_Linked issues:_  #4566

### Checklist

###Remarks

Now a single `make digests` call works for packages with and without multiple source files to create the `digests` file.

### DIFF
--- a/cross/libstdc++/Makefile
+++ b/cross/libstdc++/Makefile
@@ -19,7 +19,7 @@ INSTALL_TARGET = libstdc++6_install
 include ../../mk/spksrc.archs.mk
 
 # use digests with multiple files
-PKG_DIST_ARCHS = amd64 armhf arm64 armel i386
+PKG_DIST_ARCH_LIST = amd64 armhf arm64 armel i386
 
 PKG_DIST_ARCH = amd64
 LIB_DIR = x86_64-linux-gnu
--- a/cross/libstdc++/Makefile
+++ b/cross/libstdc++/Makefile
@@ -1,7 +1,7 @@
 PKG_NAME = libstdc++6
 PKG_VERS = 6.3.0-18+deb9u1
 PKG_EXT = deb
-PKG_DIST_NAME = $(PKG_NAME)_$(PKG_VERS)_$(PKG_ARCH).$(PKG_EXT)
+PKG_DIST_NAME = $(PKG_NAME)_$(PKG_VERS)_$(PKG_DIST_ARCH).$(PKG_EXT)
 PKG_DIST_SITE = http://http.us.debian.org/debian/pool/main/g/gcc-6/
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 EXTRACT_PATH = $(WORK_DIR)/$(PKG_DIR)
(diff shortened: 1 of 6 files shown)


mk/spksrc.generate-digests.mk
```diff
@@ -3,8 +3,10 @@
 # include this file after the rule "all:"
 #
 
+ifeq ($(strip $(PKG_DIST_ARCHS)),)
+
 $(DIGESTS_FILE): download
-	@$(MSG) "Generating digests for $(NAME)"
+	@$(MSG) "Generate digests for $(NAME) - PKG_DIST_ARCH=$(PKG_DIST_ARCH)"
```
th0ma7, line 9: if `$(PKG_DIST_ARCHS)` is empty, calling `$(MSG)` with it will also be empty?
Ah, now I see, really not obvious of using `PKG_DIST_ARCH` with and without an `S`.
To differentiate them, perhaps something like `PKG_DIST_ARCH_LIST` instead.
(while I get of keeping things aligned with CPU ARCHS for instance)

cross/libstdc++/Makefile
```diff
@@ -14,60 +14,33 @@ HOMEPAGE = https://gcc.gnu.org/
...
+PKG_DIST_ARCH = armel
 LIB_DIR = arm-linux-gnueabi
 else ifeq ($(findstring $(ARCH),$(i686_ARCHS)),$(ARCH))
-PKG_ARCH = i386
+PKG_DIST_ARCH = i386
 LIB_DIR = i386-linux-gnu
 endif
 
-include ../../mk/spksrc.cross-cc.mk
+include ../../mk/spksrc.install-resources.mk
```
th0ma7, line 40: Would there be any way to revert this and keep using `cross-cc.mk` so from there it calls `install-resoureces.mk`?
That would keep the behavior further aligned with all other packages.
Perhaps even being able to include it in `cross-cmake.mk` and `cross-meson.mk` so it is accessible in all cases?

th0ma7 (changes requested):

cross/libstdc++/Makefile
hgy59: `spksrc.install-resources.mk` was introduced for packages that do not need to comile (i.e. CONFIGURE_TARGET = nop and COMPILE_TARGET = nop) and only need a custom install target.

th0ma7:
In turn, this also keep the multi file entries within a single `digests` file?

mk/spksrc.generate-digests.mk
hgy59: Yes we already have `*_ARCH` and `*_ARCHS` in `spksrc.archs.mk` and I don't like to use a diffeent approach here.

cross/libstdc++/Makefile
th0ma7: Gotcha, that explains.  Will it also work with other packages using `cross-cc.mk` and al?

mk/spksrc.generate-digests.mk
th0ma7: Agreed, and makes me wonder if we shouldn't revisit that again sooner or later to use a `_LIST` naming.... 
In this specific case it really is easy to mix things up and would benefit of getting additional clarity.

cross/libstdc++/Makefile
hgy59: Yes it works with all cross-*.mk, Only `install-resoureces.mk` was not including `../../mk/spksrc.generate-digests.mk` and had a copy of the previous digests creation code.

mk/spksrc.generate-digests.mk
hgy59: BTW I alredy added the new Makefile variables in the [wiki](https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables)

th0ma7: it's not a problem, it's just a little misleading.

th0ma7 (approved):

hgy59:
@th0ma7 finally to complete this PR I used the naming you proposed.

_It is always boring to touch packages just by updating framework code._

PR #4601, opened 2021-05-03
fix cmake cpu for hi3535

hgy59:


th0ma7:
Good catch!

hgy59:
It was ok in cross/librsync but failed after migration of librsync to spksrc.cross-cmake.mk

PR #4604, opened 2021-05-03
Python: Update python38 to v3.8.11 and python3 to v3.7.11

hgy59:
_Motivation:_  Python 3.8 update includes some security updates
_Linked issues:_  

### Checklist


native/python3/Makefile
```diff
@@ -31,7 +31,7 @@ PIP_NATIVE = $(WORK_DIR)/../../../native/$(PKG_NAME)/work-native/install/usr/loc
 python3_native_post_install: $(WORK_DIR)/python-native.mk
 	@$(MSG) Installing pip, setuptools, cffi and cross env
 	@$(RUN) wget https://bootstrap.pypa.io/get-pip.py
-	@$(RUN) $(PYTHON) get-pip.py "pip<20.3"
+	@$(RUN) $(PYTHON) get-pip.py "pip==21.1.1"
```
smaarn, line 34: Is this expected ? I would have thought you'd only need to change python 3.8 related resources.
The usage of a specific version is a good idea anyway.

smaarn (approved):
Only worried about python 3.7 potential impacts

cross/pip/Makefile
```diff
@@ -1,6 +1,6 @@
 PKG_NAME = pip
 # use the same version in native/python3/Makefile (python3_native_post_install)
-PKG_VERS = 20.2.4
+PKG_VERS = 21.1.1
```
hgy59, line 3: @smaarn The comment here is the reason for my update in native/python3/Makefile

smaarn: I see.
The only concern I have here is that this means this update impacts all python3 dependent packages. Even though it should be ok, shouldn't we have a "pip3" specialized cross package ? I do understand anyway it would really make this PR much heavier (potential side effects ...)

hgy59: We have a pip3 cross package the `cross/pip`. For python 2.7 we have `cross/pip_py2` as pip >= 21.0 is for python >= 3.

ymartin59:
From my point of view, it is difficult to maintain multiple python versions the way we did right now.
If any new 3.7 build is required, I would use a branch for that need - where dependencies and applications remain as-is for 3.7 (instead of creating copies of libraries/dependencies for 3.7 version)

smaarn:
> From my point of view, it is difficult to maintain multiple python versions the way we did right now.
> If any new 3.7 build is required, I would use a branch for that need - where dependencies and applications remain as-is for 3.7 (instead of creating copies of libraries/dependencies for 3.7 version)

Actually I'm not sure that python 3.7 is the most expensive here. For example it seems here that the issue stems from python2 packages (ffsync).
But the idea could anyway be applied to python 2, don't you think ?

native/python3/Makefile
hgy59: python 3.7 changes moved to #4637

hgy59:
> From my point of view, it is difficult to maintain multiple python versions the way we did right now.
> If any new 3.7 build is required, I would use a branch for that need - where dependencies and applications remain as-is for 3.7 (instead of creating copies of libraries/dependencies for 3.7 version)

There is no problem with Python 3.7 and 3.8 (or 3.9), but more and more packages are dropping support for Python 2.7.

The update of setuptools (and creating cross/setuptools_py2) is required for future update of cross/cryptography that will need setuptools-rust (depending on setuptools >= 46.1).

There is still a chance to migrate the remaining packages (see #4303) to Python 3.x.
Only for ffsync, that will stay on Python 2, a dedicated Python 2 branch is recommended.


_The next challenge will come with Python 3.10 and it's breaking changes like two digit minor version._

PR #4618, opened 2021-05-08
SynoAppsDocs : package aimed to document applications using DSM API

DigitalBox98:
_Motivation:_  The usage of DSM API can be understood via this package giving access to documentation relative to the DSM apps

This package is a no_arch because only shortcut to HTML documentation

One example is how MariaDB package is using the webapi to change the password (with ExtJS) : 

Access to the package class :
apps-docs

Access to the source for the usage of the webapi : 

api-usage

A must have for the ones who would like to propose in the future package with their own webapi + extjs interface directly integrated into DSM : )

### Checklist
- [x ] Build rule `all-supported` completed successfully


ymartin59:
For package with nothing to compile, I recommend to download "resources" from spk Makefile itself (to avoid cross/ structure). @hgy59 May you please help about it?

hgy59:
> 
> 
> For package with nothing to compile, I recommend to download "resources" from spk Makefile itself (to avoid cross/ structure). @hgy59 May you please help about it?

@ymartin59 I have no idea what you mean. I propose to use `spksrc.install-resources.mk` instead of  `spksrc.cross-cc.mk` in `cross/synoapppsdocs/Makefile`.

The only package I can see without cross package is syno-magnet, but with an uncommon download...

cross/synoappsdocs/Makefile
```diff
@@ -0,0 +1,24 @@
...
+COMMENT  = Syno Application Docs information
+
+GNU_CONFIGURE = 1
+
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = nop
+INSTALL_TARGET = synoappsdocs_install
+
+include ../../mk/spksrc.cross-cc.mk
+
```
hgy59, line 20: to install resources use:
```suggestion

INSTALL_TARGET = synoappsdocs_install

include ../../mk/spksrc.install-resources.mk

```

DigitalBox98: Ok this change is working fine : )

Issue #4473, opened 2021-03-05
tt-rss don't start after install


UberElectron:
### Setup
_Package Name:_ tt-rss
_Package Version:_ 20200829-12

_NAS Model:_ DS115j
_NAS Architecture:_ MARVELL Armada 370
_DSM version:_ DSM 6.2.3-25426 Update 3

### Expected behavior
Package should run after install or after I clicked the run button.

### Actual behavior
Package don't run with a message "Failed to run the package service".
I managed to find the following error:
`Exception while creating PDO object:SQLSTATE[HY000] [1045] Access denied for user 'ttrss'@'localhost' (using password: YES)`

### Steps to reproduce
I don't know. I'm using php5.6 with MariaDB 5.

#### Package log
I was unable to find any log. I have searched into `var/logs`, `usr/local/tt-rss/var`, `var/packages/tt-rss/target/var`.

UberElectron:
I have managed to get the webpage working. However, the package still fail to run every time.
I create the ttrss user and db in MariaDB manually and use `my.site/tt-rss/install` to initialize the db and generate a config file.
No user o db was created automatically, so that may be the reason it wasn't working.
I'm having issues updating the feed following the wiki `(https://tt-rss.org/wiki/UpdatingFeeds)` using the update daemon.
Do the package manage updates by himself so it would work if the package runs?

smaarn:
Hello @UberElectron ,

Just to be sure of the initial symptom: you were performing an install from scratch and the db wasn't appropriately setup ? (I'll work on this)

In theory the updating feeds is what's actually being handled by the service upon startup so you shouldn't be needing it. Would you be able to launch it manually by running the below command as the `http` user ?: 

```
/var/packages/tt-rss/target/bin/tt-rss-daemon
```

smaarn:
(Don't hesitate running the below command using `-x` in order to have debug logs e.g. `/bin/bash -x /var/packages/tt-rss/target/bin/tt-rss-daemon`).

Would it be possible to have the logs stored under `/var/packages/tt-rss/target/var/logs/` directory ?

smaarn:
Just checking on my end and it seems that latest version hasn't been published. You are using version 20200829-12, right ?

smaarn:
Ok, it seems that latest version hasn't been published. 

[...]

PR #4624, opened 2021-05-11
tt-rss: updated 20201114 -> 20210306

smaarn:
_Motivation:_  Fixing initial installation procedure + bump of underlying tt-rss version 
_Linked issues:_  #4473 

### Checklist


smaarn:
It seems that migration process is actually way more complicated than it was (everything goes through "environment variables" mechanisms which is a biiiig shift).

Closing this PR for now.

PR #4667, opened 2021-06-03
Mosh: add libutempter

hgy59:
_Motivation:_  This is a followup of the work started by @sn00pster
_Linked issues:_  #3583

### Checklist


hgy59:
@sn00pster to work on https://github.com/SynoCommunity/spksrc/pull/3583#issuecomment-456338491 I need to have a use case for detached sessions. 
Any guidance on how to use mosh and how to detach sessions would be appreciated.

fizzyade:
> @sn00pster to work on [#3583 (comment)](https://github.com/SynoCommunity/spksrc/pull/3583#issuecomment-456338491) I need to have a use case for detached sessions.
> Any guidance on how to use mosh and how to detach sessions would be appreciated.

I’m not sure what you mean? Mosh allows the session to float between connections, unlike ssh where if the connection dies, the session is terminated.

I don’t use Synology equipment any more.

hgy59:
> I’m not sure what you mean? Mosh allows the session to float between connections, unlike ssh where if the connection dies, the session is terminated.

I don't know mosh at all (and didn't find any guidance)
- do I have to start the mosh server on the diskstation?
- what app do I need on a mobile (android) device?
- how do I connect (protocol, credentials)?

fizzyade:
Mosh isn't a daemon.  

When you want to open a mosh connection to a server, it actually ssh's into the target machine and then spawns the mosh command which either starts a new or attaches to an existing session, the ssh connection is closed, it's literally only used to initiate a session.

it uses utmp to handle the sessions, libutempter is used to list the target machines sessions, because mosh connections are persistent until you physically log the session out (sessions remain through disconnects), so of your client crashes or doesn't' save the state, the sessions will hang around on the target.  utempter is used to list these on connection, so you know that they exist and that you can then close them.

No idea about an android client, I don't use android.  I'm sure somebody has created or provided a mosh binary for it.  I personally use iOS and the best client I've ever found is Blink, which is really good.

As above, sessions are initiated via ssh, mosh handles this for you.  You need ssh to start the connection, once logged in and a mosh session is created using the mosh binary, the communication then goes via UDP.

hgy59:
@fizzyade thank you very much for the detailed description.
So I will use mosh on my debian workstation and MobaXterm on windows to test mosh with a diskstation as I have no OSX device.

cross/libutempter/patches/001-Makefile-configurable-prefix.patch
```diff
@@ -0,0 +1,20 @@
...
+ INSTALL = install
+-libdir = /usr/lib
+-libexecdir = /usr/lib
+-includedir = /usr/include
+-mandir = /usr/share/man
++prefix ?= /usr
++libdir = $(prefix)/lib
++libexecdir = $(prefix)/lib
++includedir = $(prefix)/include
++mandir = $(prefix)/share/man
```
ymartin59, line 17: It is worth to submit this upstream to ease maintenance

ymartin59 (approved):

ymartin59:
@hgy59 Thanks. As far as I remember tmux may benefit from libutempter too.

Issue #4124, opened 2020-08-13
Domoticz upgrade from 4.0.10717-8 to 4.0.9700-8?


Sven79:
### Setup
_Package Name:_ Domoticz
_Package Version:_ 4.0.10717-8

_NAS Model:_ Synology DS216+
_NAS Architecture:_ 
_DSM version:_ DSM 6.2.3-25426 Update 2

### Actual behavior
I am currently running Domoticz 4.0.10717-8. (installed from http://www.jadahl.com/)
DSM Package center informed me to update the package to 4.0.9700-8, which is an older version. I would have expected an update to a 2020-version and not a downgrade?
(edit: newer version v2020.1.11804-8 is available from jadahl)

### Steps to reproduce
![image](https://user-images.githubusercontent.com/31131918/90130548-ed8e0180-dd6a-11ea-8723-4a72b54eb268.png)

BenjV:
This is not a SynCommunity package.
So you should ask Jahdal or ask it here:
https://www.domoticz.com/forum/viewtopic.php?f=47&t=7087&sid=12bd13c1eb0d680faa7ec36dc9a10ca1&start=540#top

Sven79:
Tnx for your reply.
Yes, I know, but somehow it is seen as a SynoCommunity package; that is what is strange to me.
I cannot recall that I installed Domoticz through SynoCommunity; I'm pretty sure I installed it from Jahdal. But somehow package center detects it is installed and links it to Synocommunity...

BenjV:
There is also a Domoticz package from the SynoCommunity and the package you have is made by Diaoul and not by Jahdal.
So you must have installed is from the SynoCommunity.

If you add this package center to your Nas you can install the latest package from Jahdal.

> http://www.jadahl.com/domoticz_stable_6.2_python/

or if you don't want python extensions

> http://www.jadahl.com/domoticz_stable_6.2/

bassduh:
I am having the same issue here.
200% sure I had installed it via Jadahl's repository.

Output of:
`head -n 10 /var/packages/domoticz/INFO`

`package="domoticz"`
`dsmappname="com.synocommunity.domoticz"`
`thirdparty="yes"`
`version="4.0.10717-8"`
`description="Domoticz Home Automation System. This version includes OpenZwave and USB drivers.If you like and use my `domoticz package, you might consider a donation with paypal. https://www.paypal.me/jumbotroll"

`arch="apollolake"`
`maintainer="Domoticz"`
[...]

Issue #4069, opened 2020-07-15
Boost library V1.72 (or V1.73) compile error when address model is 32 bits


Antori91:
Hello,

I'm a new user and try to cross-compile domoticz using the latest versions of the boost library (i.e 1.72.0 or 1.73.0). When address model is 64 bits (I've tried arch-braswell), it just works fine. But when address model is 32 bits (armada38x for example), the compilation fails. Here is the log:
```
antori@A320MK:~/spksrc/cross/domoticz$ make arch-armada38x
===>  Building package for arch armada38x
make[1]: Entering directory '/home/antori/spksrc/cross/domoticz'
===>  Processing dependencies of domoticz
make[2]: Entering directory '/home/antori/spksrc/cross/boost'
===>  Compiling for boost
[...]

PR #1769, opened 2015-06-22
update Domoticz to r2563 and add openzwave support

gkroone:
Here's a cleaned up PR for upgrading domoticz to the latest stable release: r2563

Note that although this version of Domoticz is marked as stable, it depends on a recent snapshot revision of openzwave, with limited retention on the website http://old.openzwave.com/snapshots/

PR #4674, opened 2021-06-11
Update domoticz to v2020.2

### LINKED ISSUE #4673: Failed to install Domoticz - status Stopped
### Setup
_Package Name:_ domoticz.v8.f15047[armada370-armada375-armada38x-alpine-alpine4k-armadaxp-comcerto2k-monaco].spk
_Package Version:_ Domoticz v4.9700-8

_NAS Model:_ DS218J
_NAS Architecture:_Marvel Armada 385 
DSM 6.2.4-25556

&nbsp;
--

_DSM version:_ 6.2.2.4

### Expected behavior
After installation manually of the pakkage in the pakkage centre the status should be ' Installed

### Actual behavior
Failed to install Domoticz - status Stopped

### Steps to reproduce
_1._ Pakkage centre - installation manually
_2._ Domoticz - Status Stopped
_3._ 

### Setup
_Package Name:_ 
_Package Version:_ 

_NAS Model:_ 
_NAS Architecture:_ 
_DSM version:_ 

### Expected behavior
Tell us what should happen

### Actual behavior
Tell us what happens instead

### Steps to reproduce
_1._ 
_2._ 
_3._ 

#### Package log
You find service log and other log files in `/var/packages/{package}/target/var`.
_
```
Insert the package log(s) here
```
#### Other logs
```
Insert log here
```

hgy59:
_Motivation:_  Domoticz does not run and does not use generic serivce support yet (and jadahl.com will not create further updates).
_Linked issues:_  #4069, #4124, fixes #4673, (older issues; #1478, #1769)

### Checklist

### TODO:
- [x] `cross/jsoncpp` fails for `arch-qoriq` (fails only in context of spk/domoticz, build succeeds in cross/jsoncpp)
   Finally solved by fixing the CMAKE CFLAGS for PPC Archs (ppc arch has multiple cflags, issue is is for newer cmake of native/cmake only as building in cross/jsoncpp native/cmake is not used)
- [x] domoticz shows undefined version information (`Domoticz V2020.2 (c)2012-1970 GizMoCuz` and `Build Hash: 0, Date: 1970-01-01 01:00:00`)
  This is by design, as domoticz does not support build from source package, but only within a git repository (see https://github.com/domoticz/domoticz/issues/4859)
  Solution: create custom domoticz version information. As we cannot support automatic updates anyway this solution helps to display reasonable version info.
- [x] add Open-ZWave support (this was hard to get, as both - openzwave and domoticz - have some nonstandard/unusual build requirements).

### Remarks
- update domoticz and add new dependencies
- use generic service setup (includes DSM7 compatibility)
- add native/cmake, as domoticz requires cmake >= 3.16.0 and debian development environment has 3.13 (closes #4321).
- add Makefile variable to use native cmake
- update boost to v1.66.0 (v1.69.0 was also successfully built, but reverted to v1.66.0, the minimal version required by domoticz)
- remove cross/busybox and remove dsm5->dsm6 (user) migration
- we will not bundle usb drivers with domoticz package (as jadahl.com did), to separate specific drivers from generic packages.
- we will not support automatic updates of domoticz, as there is no build environment available on the diskstations.
- newer version of domoticz (2021.1) fail to build
- deluge packages fail to build with the boost library update


hgy59:
domoticz is now ready for testing.
Packages for all archs are downloadable as artifacts of the latests github action build (Checks). All supported archs successfully built domoticz. But deluge fails now due to the boost library update.

Still todo:
- [ ] validate mosquitto package update
- [ ] optional: fix deluge build with boost update (deluge currently builds only for DSM7 for aarch64 and armv7)

BenjV:
@hgy59 Can you tell me were/how I can get a domoticz.spk to test with?

hgy59:
> 
> 
> @hgy59 Can you tell me were/how I can get a domoticz.spk to test with?

Under "Actions" select a terminated "Build" (not Lint) Action. Under "Artifacts" you can download a zip archive containing all successfully built spk files.
https://github.com/SynoCommunity/spksrc/actions/runs/933333272

BenjV:
I have installed and tested domoticz and it works fine  beside some minor issues.

Because the package is running as a user and not (like jahdal's package) as root the package has no privileges to access usb devices and that is of course needed.

I would suggest during installation of the package to create a group called "domotica" and make the user "sc-domoticz" a member of that group.
Further add a udev file like "70-domotica.rules" to the folder "/usr/lib/udev/rules.d"
The following lines should be in this file:
```
KERNEL=="ttyACM*", ACTION=="add", GROUP="domotica", MODE="0660"
KERNEL=="ttyUSB*", ACTION=="add", GROUP="domotica", MODE="0660"
```
And of course these rules should be reloaded with this command:
`udevadm control --reload-rules`

Some minor remarks:

Please add the loglevel to the service command, otherwise all measurement data domoticz is getting will be logged.
So add to SERVICE_COMMAND
`-loglevel 2 `
This will log only errors not all measurements

In the service-setup there are some definitions used that are available as package variables like:
```
USER="domoticz"
SERVICE_PORT="8084"
```
They are already defines in the INFO file, so it would be better to use the package variables for those like this:
```
USER=${SYNOPKG_PKGNAME}
SERVICE_PORT=${SYNOPKG_PKGPORT}
```
One last remark.
Maybe you could add the location of the database as a choice during installation via a wizard rule.
Lots of people would like the database on for example a usb disk, because the constant logging domoticz does so preventing the main disk to go to sleep.

hgy59:
@BenjV Thanks for your detailed testing.

The intention is to install the package for `user:group` =  `sc-domoticz:sc-domoticz`.

I was focused on DSM7 where the privileges file is created by the framework on demand.
This is now fixed for DSM6.

And I still have the focus on "advance packages for DSM7".

---

> Further add a udev file like "70-domotica.rules" to the folder "/usr/lib/udev/rules.d"

This will not be supported with DSM7 and must be configured manually by a user of the admin group.
As DSM7 (still with RC1) does not allow to use any different user/group in the `"ctrl-script"` section, this will be the new standard, and I want to keep this for DSM6 too.

---

> They are already defines in the INFO file, so it would be better to use the package variables for those like this:
> 
> ```
> USER=${SYNOPKG_PKGNAME}
> SERVICE_PORT=${SYNOPKG_PKGPORT}
> ```

`USER` is not used and `SERVICE_PORT` variable is already used for the `SERVICE_COMMAND`

---

> One last remark.
Maybe you could add the location of the database as a choice during installation via a wizard rule.
Lots of people would like the database on for example a usb disk, because the constant logging domoticz does so preventing the main disk to go to sleep.

Good idea. As we currently have no solution for shared folders in DSM7, I propose to postpone this for a later update.

---

> Please add the loglevel to the service command, otherwise all measurement data domoticz is getting will be logged.
So add to SERVICE_COMMAND in de service-setup
`-loglevel 2`
This will log only errors not all measurements

Okey, will add this.

BenjV:
I have tested domoticz only on DSM 7-41882

If you change the group name in the udev rules to sc-domoticz everything will work fine.
So those lines will become:
```
KERNEL=="ttyACM*", ACTION=="add", GROUP="sc-domoticz", MODE="0660"
KERNEL=="ttyUSB*", ACTION=="add", GROUP="sc-domoticz", MODE="0660"
```
If you do not add this to the package lots of people will not be able to create those rules and it is very simple for the package to do this during installation.

Below a part of the service-setup:
```
SERVICE_PORT="8084"
# start-stop-status script redirect stdout/stderr to LOG_FILE
LOG_FILE="${SYNOPKG_PKGVAR}/${SYNOPKG_PKGNAME}.log"
# Service command has to deliver its pid into PID_FILE
PID_FILE="${SYNOPKG_PKGVAR}/${SYNOPKG_PKGNAME}.pid"
# domoticz service definition

DOMOTICZ="${SYNOPKG_PKGDEST}/bin/domoticz"
DB_FILE="${SYNOPKG_PKGVAR}/domoticz.db"
WWW_PORT="-www ${SERVICE_PORT}"
WWW_ROOT="-wwwroot ${SYNOPKG_PKGDEST}/www"
WWW_DISABLE_HTTPS="-sslwww 0"
WWW_OPTIONS="${WWW_PORT} ${WWW_ROOT} ${WWW_DISABLE_HTTPS}"

SERVICE_COMMAND="${DOMOTICZ} -daemon ${WWW_OPTIONS} -dbase ${DB_FILE} -userdata ${SYNOPKG_PKGVAR} -pidfile ${PID_FILE} -noupdates -log ${LOG_FILE}"
```
As you can see the SERVICE_PORT is hard coded to 8084 and used in WWW_PORT which is used WWW_OPTIONS which is used in the SERVICE_COMMAND
So this line would be better.
WWW_PORT="-www ${SYNOPKG_PKGPORT}"


By the way the privileges files are not new, they also work exactly the same way on DSM 6 as on DSM 7, the only difference is that in DSM 7 they are mandatory.

hgy59:
> Can you tell me what you mean by this?
> Maybe I can help

A short description is under <sup>6)</sup> in #4524

The wizards for shared folders must be redesigned as the `"data-share"` resources allow only the name of a share (without `/volume?` and no subfolders), and DSM7 allows access to such folders only.

The target is to redesign the wizards of all packages with shared folders to use resouce definition for DSM6 and DSM7.

hgy59:
> As you can see the SERVICE_PORT is hard coded to 8084 and used in WWW_PORT which is used WWW_OPTIONS which is used in the SERVICE_COMMAND

The service port is ***NOT*** hardcoded, it is only defined in spk/domoticz/Makefile.

The top 5 lines in the `service-setup` file are created by the framework based on Makefile variables and generic installer code.

The `service-setup.sh` _source_ file starts with `# domoticz service definition` and uses the variable `SERVICE_PORT`.

hgy59:
> Just add the rules file to the package and move it during installation to /usr/lib/udev/rules.d folder and add this command to the script

Did you test this with DSM7?
IMHO under DSM7 the installer will not allow to access the folder `/usr/lib/udev/rules.d` as it does not run as `root`.

BenjV:
Yes I tested with DSM 7
And to run the crtl script as root you should use this privilege file.
Then all the installation script will run as root except the start-stop-status script in this case because I left it out.
```
{
	"defaults": {
		"run-as": "package"
	},
	"username": "sc-domoticz",
	"groupname": "sc-domoticz",
	"ctrl-script": [{
		"action": "preinst",
		"run-as": "root"
	},
	{
		"action": "postinst",
		"run-as": "root"
	},
	{
		"action": "preuninst",
		"run-as": "root"
	},
	{
		"action": "postuninst",
		"run-as": "root"
	},
	{
		"action": "preupgrade",
		"run-as": "root"
	},
	{
		"action": "postupgrade",
		"run-as": "root"
	}]
}
```

See page 86 of the DSM 7 development guide.
To my opinion this privilege file should be used for all packages, because lots of them need root privilege to correctly be installed.

hgy59:
> Yes I tested with DSM 7
> And to run the crtl script as root you should use this privilege file.
> This will ensure that all the installation scripts will be run as root and the package itself not.

The problem is, that DSM7 does not accept any ctrl script entry
```
	"ctrl-script": [{
		"action": "preinst",
		"run-as": "package"
	},
```

Even with only one entry like the one above, (or an empty ctrl-script array) the DSM7 installer fails with the Message, that the package requires root privileges.

@publicarray can you report this error to synology please?

BenjV:
Very strange, I checked several packages from Synology themselves and the all have ctrl_script settings in the privilege file.

hgy59:
> 
> 
> Very strange, I checked several packages from Synology themselves and the all have ctrl_script settings in the privilege file.

DSM7 packages from synology are encrypted (or signed?) so such packages are handled differently than thirdparty packages not published by synology.
As regarding the documentation this should be supported, I suppose this is really a bug in the DSM7 installer.

BenjV:
I found this statement from Synology.
https://kb.synology.com/en-global/DSM/tutorial/supported_third_party_packages_beta

So it seems they only allow package signed by Synology to run as root.
I don't know if it was intended that this was extended to the ctrl scripts but is is at the moment although the development documentation states is should be supported as @hgy59 says.

hgy59:
@BenjV just updated the service installation

- udev rules are now installed/uninstalled (for DSM<6 only)
- the log level is now set to error,status (you requested error only)

As I do not have devices available that can send data to domoticz, can you please validate that measurement values are kept away from the log with the current settings (I suppose that one of the debuglevels hardware,received,eventsystem would add measurement value logs).

BenjV:
@hgy59 
There are no artifacts for one of your newer build.
So I cannot download a package to test.

hgy59:
> 
> 
> @hgy59
> There are no artifacts for one of your newer build.
> So I cannot download a package to test.

I can build one for you. What arch do you need?

BenjV:
Een DS116 armada_38x

hgy59:
x64 and armv7 packages are available here: https://github.com/hgy59/spksrc/releases/tag/domoticz_2020.2

BenjV:
@hgy59 
This package is working fine on DSM 7
With the current loglevel no data measurements are added to the logfile.

ymartin59 (approved):

PR #4689, opened 2021-06-25
spksrc: Allow multiple DSM for all-supported

th0ma7:
_Motivation:_  As we progress into supporting DSM7 and as a perquisite to building recently merged PR, I want to be able to build both officially supported DSM in one shot using the regular `make all-supported`.  Also, downloading of distrib package files is still an issue with current code.  This PR is attempt to fix both issues:
a) allowing multi DSM builds and
b) fix parallel build download issue
c) fix parallel build of native/*
d) output individual `build-` log files for better tracking of each arch build
_Linked issues:_  

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

```
$ make setup
Creating local configuration "local.mk"...
Setting default toolchain version to DSM-6.1
Setting default toolchain version to DSM-7.0
```
That leads into the following after `make all-supported`:
```
$ tree ffmpeg -L 1
ffmpeg
├── build-88f6281-6.1.log
├── build-aarch64-6.1.log
├── build-aarch64-7.0.log
├── build-armv7-6.1.log
├── build-armv7-7.0.log
├── build-evansport-6.1.log
├── build-evansport-7.0.log
├── build-hi3535-6.1.log
├── build-qoriq-6.1.log
├── build-x64-6.1.log
├── build-x64-7.0.log
├── work-88f6281-6.1
├── work-aarch64-6.1
├── work-aarch64-7.0
├── work-armv7-6.1
├── work-armv7-7.0
├── work-evansport-6.1
├── work-evansport-7.0
├── work-hi3535-6.1
├── work-qoriq-6.1
├── work-x64-6.1
└── work-x64-7.0
```

PR #4691, opened 2021-06-26
native/nasm: Update native nasm from verstion 2.14.02 to 2.15.05

th0ma7:
_Motivation:_  While working on fixing parallel build (PR #4689) I came across an build issue with nasm (caused by the parallel build and not nasm itself, and fixed with https://github.com/SynoCommunity/spksrc/pull/4689/commits/0484c441bea5f9967e26da51f75c979ca39fab0f).  While further looking into things I tested updating to latest version which worked out all right.
_Linked issues:_  https://github.com/SynoCommunity/spksrc/pull/4689/commits/0484c441bea5f9967e26da51f75c979ca39fab0f

### Checklist


hgy59:
I have just started to redesign native packages to use installers for amd64/x86_64 target instead of compiling from sources.

So I would like to hold this back for a moment.

hgy59:
I just found that there exists no native nasm linux installer package.
So either we use deb package to install such native packages or keep to build from source. 😞

native/nasm/Makefile
```diff
@@ -1,8 +1,8 @@
 PKG_NAME = nasm
-PKG_VERS = 2.14.02
+PKG_VERS = 2.15.05
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = https://www.nasm.us/pub/nasm/releasebuilds/$(PKG_VERS)
+PKG_DIST_SITE = https://www.nasm.us/pub/$(PKG_NAME)/releasebuilds/$(PKG_VERS)
```
hgy59, line 5: _I am not a fan of using $(PKG_NAME) in url, as you cannot doubleclick the link. I know the links are invalid with $(PKG_VERS) anyway, but $(PKG_VERS) is variable, but $(PKG_NAME) is const within a package._ Just my own opignion.

hgy59 (approved):
as there are only deb packages for installation, let's keep building this from source.

native/nasm/Makefile
th0ma7: I am of the exact same opinion.  That specific change now reverted.

PR #4693, opened 2021-06-26
Install native packages

hgy59:
_Motivation:_  Omit compilation of native packages that support direct installation for linux amd64/x86_64
_Linked issues:_  #4674

### Checklist

## Remarks
We introduced native packages as we do not want to blow up the spksrc docker image with all the tools needed to build different packages.
Until now most of the native packages are compiled for the target linux system (debian, in the spksrc docker image) on demand, with the exception for the dotnet sdk packages.
With the introduction of native/cmake (#4674) I noticed that the build of cmake takes a lot of time (and resources). _The (powerful) github build action environment takes about 7 minutes to build cmake form sources (as of June 2021), and my personal build system more than 20 min._
The download and installation (i.e. extraction) takes only a tiny little bit of resources.
This was the kick off for this PR.


th0ma7:
Just a thought, what if instead we add deb build dependencies, e.g. force `apt install` of native dependencies as we see fit?  and leave the `native/*` only for version specific requirements we may have?

Something like: `DEPENDS = deb/cmake` where a `spksrc.deb.mk` would invoke an install of packages within the debian container.

hgy59:
> Just a thought, what if instead we add deb build dependencies, e.g. force `apt install` of native dependencies as we see fit? and leave the `native/*` only for version specific requirements we may have?
> 
> Something like: `DEPENDS = deb/cmake` where a `spksrc.deb.mk` would invoke an install of packages within the debian container.

IMHO we have specific versions in native packages only. As the debian cmake package is outdated (we already have cmake in the docker image) we added newer cmake to native.

I do not like to use apt-get within the container
- the package cache is cleaned when building the docker image, so we would have to call `apt-get update` each time befor installing such a package
- as I avoid to use `root` but a sepecific user without root access in the container this is a no go too.

th0ma7:
> IMHO we have specific versions in native packages only. As the debian cmake package is outdated (we already have cmake in the docker image) we added newer cmake to native.

We could probably use backports for that...  And to that effect allow adding ppa or other channels as needed.

> I do not like to use apt-get within the container
> 
> the package cache is cleaned when building the docker image, so we would have to call `apt-get update` each time before installing such a package

Unless the cache is entirely empty (and even then) we can specify the exact version to use, thus, avoiding the `update`.  Its no different than maintaining version in our makefiles.  And even if `update` was needed, it is faster than rebuilding a `native/*` and probably not that much longer than download an external .tar file.

> as I avoid to use `root` but a sepecific user without root access in the container this is a no go too.

A simple sudo can do the trick at docker image creation while making sure our user is part of the sudo group:
```
$ sudo perl -i.bak -pe 's/=.*$/=NOPASSWD: ALL/ if /^.*sudo.*ALL=/' /etc/sudoers
```
Just saying it may be best relying more on Debian packages as much as possible rather than downloading pre-built tar-balls when we can avoid it.

hgy59:
Now all native packages with downloadable binaries are installed instead of compiled from sources.
The remaining native packages could be installed from deb packages later.
`native/squid` is going to be obsolete by #4695

PR #4696, opened 2021-06-27
fix ffmpeg 4.3.2 for ppc853x

hgy59:
_Motivation:_  Fix build of ffmpeg 4.3.2 for arch-ppc853x to support DS-210+ et al.
_Linked issues:_  

### Checklist


th0ma7:
Good catch!  Is this for DSM-5.2 support?
I so I had intentionally removed that support to ensure this doesn't ends-up being a nightmare to support due to the size of this build.

cross/ffmpeg/Makefile
```diff
@@ -64,23 +82,31 @@ CONFIGURE_ARGS += --enable-libopus
...
 CONFIGURE_ARGS += --enable-libvpx
+endif
 
 DEPENDS += cross/libzmq
 CONFIGURE_ARGS += --enable-libzmq
 
+ifneq ($(findstring $(ARCH),$(OLD_PPC_ARCHS)),$(ARCH))
 DEPENDS += cross/shine
 CONFIGURE_ARGS += --enable-libshine
+endif
```
th0ma7, line 109: I'd rather have a single `ifneq` to deal with instead of multiple repeat of it.

cross/ffmpeg/Makefile
```diff
@@ -161,23 +171,24 @@ CONFIGURE_ARGS += --arch=arm64 --enable-neon --enable-thumb --disable-yasm --dis
...
+CONFIGURE_ARGS += --disable-asm
 CONFIGURE_ARGS += --arch=ppc
-endif
-
 ifeq ($(findstring $(ARCH),qoriq),$(ARCH))
-CONFIGURE_ARGS += --disable-asm --cpu=e500v2
+CONFIGURE_ARGS += --cpu=e500v2
+else
+CONFIGURE_ARGS += --cpu=e500 --disable-altivec
+endif
```
th0ma7, line 180: nice one :)

cross/ffmpeg/Makefile
hgy59: So you don't mind when I change the order to have all in one group for non OLD_PPC_ARCHS?

hgy59:
> If so I had intentionally removed it to ensure this doesn't ends-up being a nightmare to support due to the size of this build.

Yes I knew, that's why I waited with this, until you finished your work.

hgy59:
@th0ma7 just wondering about debug settings in cross/ffmpeg/Makefile (lines 22 and 24)
```
CONFIGURE_ARGS += --disable-debug 
CONFIGURE_ARGS += --enable-debug=1
```

I think one of those is obsolete.

cross/ffmpeg/Makefile
th0ma7: Nah, I don't mind.  I prefer a single group to manage that one exception.

th0ma7:
> @th0ma7 just wondering about debug settings in cross/ffmpeg/Makefile (lines 22 and 24)
> 
> ```
> CONFIGURE_ARGS += --disable-debug 
> CONFIGURE_ARGS += --enable-debug=1
> ```
> 
> I think one of those is obsolete.

Again good catch... and may explain the size of the package which I believe increased a bit.  I had enabled debug a while ago but it's not needed anymore.  Perhaps keeping the lines close by and commenting the `--enable` would be best to ease its activation at a later time if need be.

cross/ffmpeg/Makefile
hgy59: done

PR #4707, opened 2021-07-03
spk.mk: Fix native compile for parallel builds

th0ma7:
_Motivation:_  I made an error when trying to add file descriptor locking against `native` compiling for parallel builds.  The reason I had first hit the issue is that the native build hook wasn't properly being handled in the first place.  Code change removes the `flock` and call-up `$(MAKE) dependency-list` instead to generate full dependency list.  It looks for `native`, builds them and then move on to the parallel building of all archs to be processed.
_Linked issues:_  #4689

### Checklist

### DIFF
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -442,17 +442,12 @@ endif
 
 pre-build-native:
 	@$(MSG) Pre-build native dependencies for parallel build
-	@for depend in $(sort $(BUILD_DEPENDS) $(DEPENDS) $(OPTIONAL_DEPENDS)) ; \
+	@for depend in $$($(MAKE) dependency-list) ; \
 	do \
 	  if [ "$${depend%/*}" = "native" ]; then \
-	    exec 5> /tmp/native.$${depend}.lock ; \
-	    flock --timeout $(FLOCK_TIMEOUT) --exclusive 5 || exit 1 ; \
-	    pid=$$$$ ; \
-	    echo "$${pid}" 1>&5 ; \
 	    $(MSG) "Pre-processing $${depend}" ; \
 	    $(MSG) "  env $(ENV) $(MAKE) -C ../../$$depend" ; \
-	    env $(ENV) $(MAKE) -C ../../$$depend ; \
-	    flock -u 5 ; \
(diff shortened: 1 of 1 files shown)


mk/spksrc.spk.mk
```diff
@@ -442,17 +442,12 @@ endif
 
 pre-build-native:
 	@$(MSG) Pre-build native dependencies for parallel build
-	@for depend in $(sort $(BUILD_DEPENDS) $(DEPENDS) $(OPTIONAL_DEPENDS)) ; \
+	@for depend in $$($(MAKE) dependency-list) ; \
```
hgy59, line 445: Im would avoid to use `make dependency-list` as `dependency-list` target was introduced to to find dependencies based on changed files (for github build action) and the implementation is complex with recursive calls...
If you would like to eliminate duplicate dependencies, you could use `sort -u` for this.

To build native packages first:
- define `BUILD_DEPENDS native/...` instead of `DEPENDS native/...` in all packages
- change for loop here from
 `@for depend in $(sort $(BUILD_DEPENDS) $(DEPENDS) $(OPTIONAL_DEPENDS)) ; \`
  to
 `@for depend in $(sort -u $(BUILD_DEPENDS)) $(sort -u $(DEPENDS) $(OPTIONAL_DEPENDS)) ; \`

th0ma7: Pre-building `native` was first introduced with #4200, already using `make dependency-list` and working really well until a regression was later inserted with PR #4349.  Call to `pre-build-native` is only occurring against (`publish-`) `all-supported|latest`.

Proposed change does not work and is the reason I am asking to revert to previous working solution from #4200.  Also, adapting all Makefile for `BUILD_DEPENDS` adds a lot of unnecessary churn and won't be maintainable in the long run.  There will always be forgotten cases that are otherwise already well handled by `make dependency-list` at no additional cost.

The recursivness of `make dependency-list` is the exact reason it works so well and the only method that can make this to work (I don't know who wrote it, but just wow, simple and efficient!)  Plus, we reuses existing built-in spksrc functions.  Really complex cases such as `ffmpeg` with 50+ dependencies will never work otherwise without relying on this function.

hgy59: Ok I was the author of `dependency-list` target, happy to see it works for this usecase too. 😃

hgy59 (approved):

PR #4717, opened 2021-07-08
Fix setup-synocommunity command

publicarray:
Spaces where added at some point in local.mk. Why?

_Motivation:_  This is needed for GitHub actions, specifically when using it to publish: https://github.com/SynoCommunity/spksrc/pull/4491/files.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,10 +169,10 @@ dsm-%: local.mk
 	@grep -q "^DEFAULT_TC.*=.*$*.*" local.mk || sed -i "/^DEFAULT_TC =/s/$$/ $*/" local.mk
 
 setup-synocommunity: setup
-	@sed -i -e "s|PUBLISH_URL=.*|PUBLISH_URL=https://api.synocommunity.com|" \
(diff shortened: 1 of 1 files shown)


hgy59 (approved):

Makefile
```diff
@@ -169,10 +169,10 @@ dsm-%: local.mk
 	@grep -q "^DEFAULT_TC.*=.*$*.*" local.mk || sed -i "/^DEFAULT_TC =/s/$$/ $*/" local.mk
 
 setup-synocommunity: setup
-	@sed -i -e "s|PUBLISH_URL=.*|PUBLISH_URL=https://api.synocommunity.com|" \
```
smaarn, line 172: Why not using a regular expression matching any whitespace characters ?
Something such as `\s*` ?

smaarn: (Or `\\s*` depending on the escaping rules obviously...)

publicarray: Mostly because I don't care enough about this. I want to deploy packages not worry about spaces. You are right that is the correct way to do it of course. I just needed a quick fix at the time.

PR #4683, opened 2021-06-15
Update minio to 2021-09-09T21-37-07Z

fgma:
_Motivation:_  Update minio to 2021-06-14T01-29-23Z and prepare for DSM7.
_Linked issues:_  https://github.com/SynoCommunity/spksrc/issues/4524

Depends on https://github.com/SynoCommunity/spksrc/pull/4655 being merged first to compile.

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

### Checklist DSM 7
- [x] SERVICE_WIZARD_SHARE is not supported for DSM7. Shared folders must be declared in resources and must be the name of the share (without volume).
- [x] Invalid version number. DSM7 validates $(SPK_VERS)-$(SPK_REV) with regex [^\d+(\.\d+){0,5}(-\d+)?$]. So only digits and dots are allowed in SPK_VERS
- [x] SERVICE_EXE needs migration to SERVICE_COMMAND

I'm not sure about SERVICE_WIZARD_SHARE at the moment.

PR #4627, opened 2021-05-11
Add kiwix

hgy59:
_Motivation:_  Add package for kiwix
_Linked issues:_  closes #4079

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

### Remarks
new Makefile variable `SERVICE_WIZARD_SHARE_NAME` to create shared folder with resource file. The assigned wizard variable accepts the name of an existing shared folder or the folder will be created at installation time.  No folder must be in such variables (i.e. `download` instead of `volume1/download`). This works for DSM >= 6.1 only.

PR #4719, opened 2021-07-08
Transmission DSM7 support

### LINKED ISSUE #4313: Can‘t start transmission on DSM7.0

publicarray:
_Motivation:_  share worker supported since 6.0-5914
_Linked issues:_  #4562 closes #4313

### Checklist
- [x] Package upgrade completed successfully (dsm 6 & 7 individually) some users report a leftover firewall rule #4313
- [x] New installation of package completed successfully (dsm6 & 7)


houndtt:
great to see the work to bring this package to DSM7. looking forward to its release

publicarray:
@hgy59 I've added the ability to have DSM version specific wizard files. (needs French translation)
@BenjV already tested new installation procedure of the DSM7 package.

Selenuix:
> @hgy59 I've added the ability to have DSM version specific wizard files. (still needs French translation)
> @BenjV already tested new installation procedure of the DSM7 package.

Are the translations still needed? Do you accept PRs? I got no real way to test but I could always try to help.

publicarray:
Bonjour @Selenuix, PRs are always welcome. The Wizards are in a JSON format you can see the old translation I removed here https://github.com/SynoCommunity/spksrc/pull/4719/commits/fb8640d0f5eb7798effe862460172d22c7846d9e and the new ones can be found here https://github.com/SynoCommunity/spksrc/pull/4719/files#diff-10ef7b0d6d3b7deeb475da37e2b322ab6bdbaa8807b1146be0a9e531351fedf9

You can copy the contents into a new file appending the `_fre` to the file name. Mostly only the description needs translation. Any help is appreciated.

spk/transmission/src/service-setup.sh
```diff
@@ -70,17 +70,40 @@ service_postinst ()
...
+            #     mv "$OLD_DOWNLOAD_DIR"/* "$NEW_DOWNLOAD_DIR"
+            # fi
+            if [ -n "${OLD_INCOMPLETE_DIR}" ] &&  [ "$OLD_INCOMPLETE_DIR" != "$NEW_INCOMPLETE_DIR" ]; then
+                mkdir -p "$NEW_INCOMPLETE_DIR"
+                mv -nv "$OLD_INCOMPLETE_DIR"/* "$NEW_INCOMPLETE_DIR/"
+            fi
+            if [ -n "${OLD_WATCHED_DIR}" ] && [ "$OLD_WATCHED_DIR" != "$NEW_WATCHED_DIR" ]; then
+                mkdir -p "$NEW_WATCHED_DIR"
+                mv -nv "$OLD_WATCHED_DIR"/* "$NEW_WATCHED_DIR/"
+            fi
```
publicarray: @hgy59 what do you think about the migration? I can't decide whether It's better for the user to do this manually rather than the package.

BenjV: First of all I found a bug in the package.
The new Package is designed to use a "incomplete" folder, located on the new transmission share and be enabled it by default.
The setting for enabling this is not correct, in the setting.json this line is created:

> "incomplete-dir-enabled": @incomplete_dir_enabled@,

But that should be:

> "incomplete-dir-enabled": true,

Secondly I would suggest that the package moves the old *.part file to te new incomplete location.
Where the old .part are located can be found in te settings.json.

If you create a package for a ARM cpu (DS116) I would be happy do do the testing on DSM 7

publicarray: I can not reproduce the `@incomplete_dir_enabled@` issue. Is this from an upgrade or a fresh installation?

The part file migration is exactly the code snip it you're commenting on. So that's a +1 on the package to migrate those, Cool I'm happy with that.

Thanks I'll upload the new build. In the meantime you can always find the latest build at the bottom of a GitHub Action run. The artefacts contain all compiled architectures: https://github.com/SynoCommunity/spksrc/actions/runs/1177151421#artifacts

BenjV: Ok I tested the latest artifact package and it it correct now.
So I tested probably a older version.

spk/transmission/src/service-setup.sh
```diff
@@ -2,22 +2,22 @@
 # This gives tranmission the power to execute python scripts on completion (like TorrentToMedia).
 PYTHON_DIR="/usr/local/python"
 PATH="${SYNOPKG_PKGDEST}/bin:${PYTHON_DIR}/bin:${PATH}"
```
hgy59, line 4: As there is no `SPK_DEPENDS=python...` in this package, it is not defined which python should be used. 

For synocommunity packages the path `/usr/local/{package}` must not be used anymore, i.e. it must be defined as `/var/packages/{package}/target`.

I suppose that python2 should not be used anymore, so the package must be  `python3` or `python38`.

Proposal:
remove these lines and rely on python that is already installed and available in the path.

if no python is found in path, a lookup for synocommunity python3 and python38 could be used here to add it to the path for transmission.

publicarray: Thanks. Good catch!

BenjV: As of DSM 7 Python 3 is standard available and can be found via the path.

publicarray: >if no python is found in path, a lookup for synocommunity python3 and python38 could be used here to add it to the path for transmission.

@hgy59 I didn't do any fancy detection because that is what $PATH is for anyway. Don't need to do it twice and this way it still works if a python package is added or removed. Agreed, for DSM7 the build in Python should be good enough.

hgy59:
Can we now use `spk/transmission` as a template for wizard files with shared folders compatible with DSM 7?
This would help to finalize #4683, #4627 and all the download related packages like #4743 and others.

Zigwan06:
thank you for your work

BenjV:
@hgy59 , @publicarray 
I would suggest to use it for all packages, but at least not only for packages that create shared folders but also the ones who must read them.
Like Sonarr, Radarr and Sick(beard/chill) that reads files from the shares created by Nzbget and/or Transmission.
Their user should also be member of the syncommunity group.

So I would suggest to make all packages use the syncommunity group and not sc-packagename to keep things simple.

spyrodragon1:
Thanks for the work on this one. Just curious when the transmission update package actually be released via the synology apps section ?  Thanks again.

publicarray:
@hgy59  Yes we can. #4748 is interesting because I want to know what happens when multiple packages use the same shared folder. For DSM6 It also uses the old behaviour of creating shared folders. So existing DSM6 packages should continue to work. (sorry about the reverts I'm a  bit indecisive)

@BenjV > shared folders but also the ones who must read them.

I'm not sure on that. Everyone has a different way to organise their media. e.g. filebot for instance. Setting up the permissions for the packages that you use is easy enough already. I prefer users to be educated rather than forcing a specific way of doing things. IMHO we should be phasing out wizards not add more.

> all packages use the syncommunity group

You already convinced me see #4789

@ spyrodragon1  tops 48h for the CDN to catch up

houndtt:
was able to repair my transmission package successfully with this and it even retained all my previous torrents. great work folks!

gemmmo:
Just updated DS218 (rtd1296) to DSM7. Repair didn't work in Package Manager but manual install of the package from the synocommunity website completed without errors and everything works fine. Thank you!

isr25:
Thank you for your hard work!

gemmmo:
Not sure if this is the right place to ask, after installing transmission on DSM7 it shows up in Package Center and works when launched using the URL but there is no app shortcut in the main menu. Is there a way to fix this? For example if I uninstall Transmission and reinstall transmission will my torrent data be automatically restored? Thank you.

BenjV:
Yes if you uninstall Transmission and do not remove the data you can install Transmission again and all torrents will be restored.

gemmmo:
> Yes if you uninstall Transmission and do not remove the data you can install Transmission again and all torrents will be restored.

Thanks for the help! Reinstalled smoothly but the Transmission logo still doesn't show up in the Main Menu. I'll leave it since it works fine launching from URL and the migration was perfect from DSM7.

BenjV:
I have experienced that with more packages(also on DSM 6) and sometimes it just re-appears again and sometimes not.
Has noting to do with this package, maybe a reboot of your Nas will help.

gemmmo:
So far rebooting hasn't worked yet but I will see how it goes. Transmission is working great regardless, thank you for the help!

spyrodragon1:
Apologies for the stupid question but do i need to update to v7 of the synology software to see the transmission update or should i see it while I'm on v6 ? The update hasn't appeared in the package center yet on my NAS and I'm just wondering if the update just hasn't appeared yet ( and I should be more patient ) or it hasn't appeared because it won't appear until i upgrade to v7. Hope that makes some semblance of sense.  Thanks

BenjV:
Version 3.00-19 is available for both DSM 6 as well as DSM 7 as of yesterday
What architecture is your Nas?

You can always download it and do a manual install.
Download can be found here.
https://synocommunity.com/package/transmission

BenjV:
You need the geminilake version for a DS220+

spyrodragon1:
@BenjV - Sorry again for the stupid question but it says transmission 3.00-19 was the 1st November 2020 release on that page ?

Update openssl to 1.1.
Date
2020-11-01 17:01:28.615450

Apologies again if I'm missing something.....

BenjV:
Most likely a mistake in a year field.
This version is currently the latest build.

vjayer:
> You need the geminilake version for a DS220+ zie: https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model
@ spyrodragon1
sorry for the necroposting reply but just wanted to place this in case others have the same question. 
Gemini Lake is the Intel codename for the family that includes the Celeron J4025 
https://ark.intel.com/content/www/us/en/ark/products/197307/intel-celeron-processor-j4025-4m-cache-up-to-2-90-ghz.html
so your DS220+ should be supported

PR #4753, opened 2021-07-25
Update DSM7.0 Toolchains

publicarray:
* Newer GCC compiler


mk/spksrc.tc-vers.mk
```diff
@@ -64,4 +64,5 @@ endif
 
 ifeq ($(TC_VERS),7.0)
 TC_BUILD = 40000
```
hgy59, line 66: we could now set the build version to 41890

mk/spksrc.tc-vers.mk
```diff
@@ -64,4 +64,5 @@ endif
...
+TC_VERS = 7.0.0
```
hgy59, line 67: We might get unexpected side effects, as the DSM version is still 7.0 and not 7.0.0

publicarray:
FYI  DSM7.0 is still missing the GPL source code https://sourceforge.net/projects/dsgpl/files/Synology%20NAS%20GPL%20Source/

publicarray:
@hgy59 I noticed that Synology is also hosting them: https://archive.synology.com/download/ToolChain/toolchain/7.0-41890
Maybe we can add support for multiple mirrors?

publicarray:
Remember to do a `make dist-clean`

PR #4602, opened 2021-05-03
Add clientbuffer module to znc

jaloji:
_Motivation:_  A very usefull module who is still absent in the actual version of ZNC
_Linked issues:_  closes #4279

I saw that the makefile for ZNC compile all the cpp files who are present in module dir so I just have to add the .cpp file for client buffer in /modules dir and add clientbuffer.so in the PLIST file. All compilation are done and .spk packages can be found here: https://github.com/jaloji/spksrc/releases/tag/1.8.2-17

For each new updates of ZNC all the module must be recompiled with the new version.

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully

PR #4781, opened 2021-08-15
fix package user in privilege file

hgy59:
_Motivation:_   jq does not create valid json when username is already set
_Linked issues:_  This error popped up with the publish of #4602 

### Checklist

With already set `"username": "USER"` jq appends an `}` at the end of the privilege file and thus makes it invalid.

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -171,28 +171,28 @@ $(DSM_CONF_DIR)/resource:
 	$(create_target_dir)
 	@echo '{}' > $@
 ifneq ($(strip $(SERVICE_PORT)),)
-	@jq '."port-config"."protocol-file" = "$(DSM_UI_DIR)/$(SPK_NAME).sc"' $@ | sponge $@ 1<>$@
+	@jq '."port-config"."protocol-file" = "$(DSM_UI_DIR)/$(SPK_NAME).sc"' $@ | sponge $@
 endif
 ifneq ($(strip $(FWPORTS)),)
 # e.g. FWPORTS=src/foo.sc
 	@jq --arg file $(FWPORTS) \
-		'."port-config"."protocol-file" = "$(DSM_UI_DIR)/"+($$file | split("/")[-1])' $@ | sponge $@ 1<>$@
+		'."port-config"."protocol-file" = "$(DSM_UI_DIR)/"+($$file | split("/")[-1])' $@ | sponge $@
 endif
 ifneq ($(strip $(SPK_COMMANDS)),)
 # e.g. SPK_COMMANDS=bin/foo bin/bar
 	@jq --arg binaries '$(SPK_COMMANDS)' \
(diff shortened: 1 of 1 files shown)


hgy59:
@publicarray this is a very strange behaviour of `jq`.
It might be related to the  inplace modification with `1<>$@` , but I have no clue.

hgy59:
Without the fix, the privilege file was created like this:
[privilege.txt](https://github.com/SynoCommunity/spksrc/files/6988053/privilege.txt)

publicarray:
Yes you are 100% right. Same problem/solution here as well: https://github.com/SynoCommunity/spksrc/commit/7edf5f41d0d566e5392a80919552ad89d6aa2aa7

we might be better off adding the `sponge` utility from the moreutils package -> https://stackoverflow.com/questions/42716734/modify-a-key-value-in-a-json-using-jq-in-place

publicarray:
man sponge

>sponge - soak up standard input and write to a file
>
>sed '...' file | grep '...' | sponge file
>
>sponge reads standard input and writes it out to the specified file. Unlike a shell redirect, sponge soaks up all its input before >opening the output file. This allows constricting pipelines that read from and write to the same file.
>
>If no output file is specified, sponge outputs to stdout.

hgy59:
@publicarray it runs like a charm with sponge...
The dockerfile update is in a dedicated PR #4782, as the updated spksrc image must be available when we merge #4781.

hgy59:
> Same problem/solution here as well: [7edf5f4](https://github.com/SynoCommunity/spksrc/commit/7edf5f41d0d566e5392a80919552ad89d6aa2aa7)

Thanks for the link, I remembered little, but couldn't find it...

publicarray:
Thanks @hgy59 and no problem :smile:

mk/spksrc.service.mk
```diff
@@ -171,28 +171,28 @@ $(DSM_CONF_DIR)/resource:
 	$(create_target_dir)
 	@echo '{}' > $@
 ifneq ($(strip $(SERVICE_PORT)),)
-	@jq '."port-config"."protocol-file" = "$(DSM_UI_DIR)/$(SPK_NAME).sc"' $@ 1<>$@
+	@jq '."port-config"."protocol-file" = "$(DSM_UI_DIR)/$(SPK_NAME).sc"' $@ | sponge $@ 1<>$@
```
publicarray, line 174:
```suggestion
	@jq '."port-config"."protocol-file" = "$(DSM_UI_DIR)/$(SPK_NAME).sc"' $@ | sponge $@
```

Isn't the `1<>$@` redundant now?

hgy59: you are right...

hgy59:
@publicarray now it would be great use `jq` for the generation of `$(STAGING_DIR)/$(DSM_UI_DIR)/config:` to avoid the `python -m json.tool`

PR #4782, opened 2021-08-15
update deveopment environment

hgy59:
_Motivation:_  The development environment needs sponge command to use `jq` for in place modification.
_Linked issues:_  required by #4781, #4693

### Checklist

### Remarks
- add moreutils for sponge command
- add sudo to use docker image as regular user
This is a precondition to install native packages from repository as proposed in #4693.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 		libunistring-dev \
 		lzip \
 		mercurial \
+		moreutils \
 		ncurses-dev \
 		ninja-build \
 		php \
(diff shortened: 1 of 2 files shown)


publicarray:
LGTM but I still don't understand what the new user 'user' is for?

hgy59:
> LGTM but I still don't understand what the new user 'user' is for?

I always run the spksrc container with `docker run -it .... --user $(id -u):$(id -g) ... synocommunity/spksrc:latest`

So all file modifications in the mounted /spksrc folder in the container are done with my host user, and therefore I have full access without root on the host.
Without the user `user` inside the container the user is `I have no name!`.
I know that this works only due when the host user id is 1000 (the same as the user has inside the container).

Finally we need a dedicated user in the container that works with sudo without password if one want to avoid using root.

This is for a future update to eliminate to build native binaries (cmake, go, etc.) and to use `apt-get install` to install such prerequisites instead.

publicarray:
@hgy59 Thanks :+1: 

Rather than `--user $(id -u):$(id -g)` you could add the [`USER user` or `USER 1000:1000` ] (
https://docs.docker.com/engine/reference/builder/#user) to the image, so this is done by default? Maybe add a group for the id 1000 as well?

hgy59:
> Maybe add a group for the id 1000 as well?

The group 1000 (user) is created by default for the user `user`.

```
$ cat /etc/passwd | grep user
user:x:1000:1000:,,,:/home/user:/bin/bash
```

publicarray:
Is it possible to get the image on docker hub updated?

hgy59:
> Is it possible to get the image on docker hub updated?

@ymartin59 how to trigger an update of the docker hub image?

publicarray:
I used a github action to build the image: https://github.com/publicarray/spksrc/pkgs/container/spksrc and https://hub.docker.com/r/publicarray/spksrc

PR #4783, opened 2021-08-15
Parallel cmake and make

publicarray:
_Motivation:_  Make compiling faster. I've only tested `fish` and `umurmur` so far. (I have an [i7-10750H with 6 cores](https://www.intel.com/content/www/us/en/products/sku/201837/intel-core-i710750h-processor-12m-cache-up-to-5-00-ghz/specifications.html))

```sh
# fish before
real	2m29.187s
user	2m58.331s
sys	0m29.212s

# fish after
real	0m49.710s
user	4m31.117s
sys	0m35.185s

# fish ninja build (https://github.com/publicarray/spksrc/tree/cmake-ninja)
real	1m8.352s
user	6m17.500s
sys	0m47.527s

# fish https://github.com/SynoCommunity/spksrc/pull/4803
real    0m49.367s
user    4m33.264s
sys     0m35.001s
```

`fish` from 2m29s to 49 s

```sh
# python38 before 
real    10m44.831s
user    15m10.705s
sys 2m10.672s

# python38 after
real    6m6.519s
user    13m9.923s
sys 1m52.586s
```

### DIFF
--- a/cross/busybox/Makefile
+++ b/cross/busybox/Makefile
@@ -27,7 +27,7 @@ busybox_configure:
 
 .PHONY: busybox_compile
 busybox_compile:
-	$(RUN) CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) $(MAKE) busybox
+	$(RUN) CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) $(MAKE) -j$(NCPUS) busybox
 
 .PHONY: busybox_install
 busybox_install:
--- a/cross/busybox/Makefile
+++ b/cross/busybox/Makefile
@@ -11,8 +11,9 @@ HOMEPAGE = https://busybox.net
 COMMENT  = Fairly comprehensive set of programs needed to run a Linux system
 LICENSE  = GPL
 
+COMPILE_MAKE_OPTIONS += CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) busybox
+
 CONFIGURE_TARGET = busybox_configure
-COMPILE_TARGET = busybox_compile
 INSTALL_TARGET = busybox_install
 
 include ../../mk/spksrc.cross-cc.mk
@@ -24,10 +25,6 @@ busybox_configure:
 	  sed -i -r -e "s/^(# )?$$option.*$$/$$option=$$value/" $(WORK_DIR)/$(PKG_DIR)/configs/spksrc_defconfig ; \
 	done
 	$(RUN) CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) $(MAKE) spksrc_defconfig
-
-.PHONY: busybox_compile
-busybox_compile:
-	$(RUN) CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) $(MAKE) -j$(NCPUS) busybox
 
 .PHONY: busybox_install
 busybox_install:
--- a/mk/spksrc.compile.mk
+++ b/mk/spksrc.compile.mk
@@ -33,7 +33,7 @@ compile_msg:
 pre_compile_target: compile_msg
 
 compile_target:  $(PRE_COMPILE_TARGET)
-	@$(RUN) $(MAKE) $(COMPILE_MAKE_OPTIONS)
+	@$(RUN) $(MAKE) -j$(NCPUS) $(COMPILE_MAKE_OPTIONS)
 
 post_compile_target: $(COMPILE_TARGET)
 
--- a/mk/spksrc.compile.mk
+++ b/mk/spksrc.compile.mk
@@ -33,7 +33,7 @@ compile_msg:
 pre_compile_target: compile_msg
 
 compile_target:  $(PRE_COMPILE_TARGET)
-	@$(RUN) $(MAKE) -j$(NCPUS) $(COMPILE_MAKE_OPTIONS)
+	@$(RUN) $(MAKE) $(COMPILE_MAKE_OPTIONS)
 
 post_compile_target: $(COMPILE_TARGET)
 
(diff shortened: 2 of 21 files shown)


mk/spksrc.compile.mk
```diff
@@ -33,7 +33,7 @@ compile_msg:
 pre_compile_target: compile_msg
 
 compile_target:  $(PRE_COMPILE_TARGET)
-	@$(RUN) $(MAKE) $(COMPILE_MAKE_OPTIONS)
+	@$(RUN) $(MAKE) -j$(NCPUS) $(COMPILE_MAKE_OPTIONS)
```
publicarray, line 36: i'm not sure if this does anything

cross/busybox/Makefile
```diff
@@ -27,7 +27,7 @@ busybox_configure:
 
 .PHONY: busybox_compile
 busybox_compile:
-	$(RUN) CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) $(MAKE) busybox
+	$(RUN) CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) $(MAKE) -j$(NCPUS) busybox
 
 .PHONY: busybox_install
```
hgy59, line 32: most of the custom compile-targets adjusted here could be removed by definition of `COMPILE_MAKE_OPTIONS`
as the default compile_target executes `$(RUN) $(MAKE) $(COMPILE_MAKE_OPTIONS)`

busybox_compile can be removed by:
`COMPILE_MAKE_OPTIONS += CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX)  busybox`

this work only, when it does not matter that CROSS_COMPILE definition must not be left of $(MAKE) on the command line.

As spksrc.common.mk already defines `COMPILE_MAKE_OPTIONS += -j$(NCPUS)` this must not be added here again.

mk/spksrc.compile.mk
hgy59: yes it is obsolete, as `COMPILE_MAKE_OPTIONS += -j$(NCPUS)` is defined in spksrc.common.mk

native/elixir/Makefile
```diff
@@ -22,7 +22,7 @@ include ../../mk/spksrc.native-cc.mk
 
 .PHONY: elixir_compile
 elixir_compile:
-	$(RUN) PATH=$$PATH:$(ERLANG_BIN_DIR) $(MAKE) $(MAKE_COMPILE_OPTIONS)
+	$(RUN) PATH=$$PATH:$(ERLANG_BIN_DIR) $(MAKE) -j$(NCPUS) $(MAKE_COMPILE_OPTIONS)
```
hgy59, line 25: here the PATH definition must preceed the $(MAKE) call, but we can omit the `-j$(NCPUS)` , as it is already in `MAKE_COMPILE_OPTIONS`.

mk/spksrc.compile.mk
publicarray: Thanks

th0ma7:
Humm, rather interesting but there is a need to be really really careful with this... There are two things to be on the lookout:
1. Is this being called as a parallel build in the first place?
2. Will you overrun your total CPU core count?

Let's say you run `make -jX all-supported` (X being `nproc`).  This will result in parallel-build up to `npcroc` times of `arch-<arch>-<version>`.  Now if you add parallel builds within each `cross/*` you may end-up with a `npcroc` * `npcroc` resulting in overwhelming your system entirely.

Currently, as we call `-@MAKEFLAGS= $(MAKE) ...` to invoke parallel building of `arch-<arch>-<version>` so you won't know if a `-j` was passed over to make.  So there would be a need to check that first, and only then when assuming not already building in parallel (or not exceeding the total amount of core), add parallel building of `cross*`.

publicarray:
Hm that is a valid concert. I still think this is an overall improvement and adds consistency. AFAIK lots of packages are already running like you describe (npcroc * npcroc). I guess my focus is to make one arch builds faster to increase development productivity, everything else is secondary IMHO. As for overwhelming your system. It depends entirely how many cores you let docker use.

I guess my ultimate aim is to make one arch builds faster to increase development productivity and let the CI build all other architectures.

th0ma7:
I took a totally different approach (ref: https://github.com/SynoCommunity/spksrc/pull/4803/commits/b62495f9feaf93f4125ca63b5dd1b1bd47fa9b44), based on my current PR while fixing `cross/*` direct arch builds (PR #4803).

Running quick tests on my 12x core AMD show a significant reduction of build time for both regular `make` and `cmake`.  Actually, `zlib` was sooo fast that I didn't thought it had built in the first place:

### cross/SVT-AV1
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	2m51.662s
user	2m38.332s
sys	0m13.549s

$ time make arch-apollolake-7.0
real	0m32.660s
user	4m4.555s
sys	0m18.168s
```

### cross/ZLIB
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	0m3.024s
user	0m2.190s
sys	0m0.888s

$ time make arch-apollolake-7.0
real	0m0.876s
user	0m2.553s
sys	0m1.146s
```

publicarray:
**spk/fish**

Before (current master):

```sh
$ make -C spk/fish clean
$ time make -C spk/fish all-supported
real    22m45.112s
user    27m26.899s
sys     4m33.054s
```

After (this PR):

```sh
$ make -C spk/fish clean
$ time make -C spk/fish all-supported
real    7m33.658s
user    40m11.263s
sys     5m12.594s

$ make -C spk/fish clean
$ time make -j12 -C spk/fish all-supported
Out of memory (over 18GB)
```

publicarray:
Superseded by #4803

PR #4788, opened 2021-08-16
Use jq to generate the UI config file

publicarray:
From a conversation that started here: https://github.com/SynoCommunity/spksrc/pull/4781#issuecomment-899105000


publicarray:
@hgy59 Is this good to merge?

PR #4789, opened 2021-08-16
DSM7: Set synocommunity as the default group

publicarray:
It was a beta thing and never worked correctly, or more likely I did not know how it was supposed to work.

`SPK_GROUP` was never used (proof with `rg -g 'spk/**/Makefile' 'SPK_GROUP'` ) and should be `GROUP` anyway e.g. `sc-download` 

Files will have `$username:$groupname` as the package file owners.

Relates to conversation here: https://github.com/SynoCommunity/spksrc/issues/4524#issuecomment-877112301

starts around: https://github.com/SynoCommunity/spksrc/issues/4524#issuecomment-871166899 and https://github.com/SynoCommunity/spksrc/issues/4524#issuecomment-869634575

### DIFF
--- a/mk/spksrc.service.mk
+++ b/mk/spksrc.service.mk
@@ -256,12 +256,6 @@ $(DSM_CONF_DIR)/privilege:
 	@$(MSG) '(privilege) run-as: package'
 	@$(MSG) "(privilege) DSM >= 7 $(DSM_CONF_DIR)/privilege"
 # Apply variables to privilege file
-ifneq ($(strip $(GROUP)),)
-# Creates group but is different from the groups the user can create, they are invisible in the UI an are only usefull to access another packages permissions (ffmpeg comes to mind)
-# For DSM7 I recommend setting permissions for individual packages (System Internal User)
-# or use the shared folder resource worker to add permissions, ask user from wizard see transmission package for an example
-	@jq --arg packagename $(GROUP) '."join-pkg-groupnames" += [{$$packagename}]' $@ | sponge $@
-endif
 ifneq ($(strip $(SYSTEM_GROUP)),)
 # options: http, system
 	@jq '."join-groupname" = "$(SYSTEM_GROUP)"' $@ | sponge $@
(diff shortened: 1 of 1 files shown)


publicarray:
DSM6 to DSM7 migration renames the **OLD** group if the same group name is used in both. Like `sc-download` etc. The permissions stay in the old group of course since DSM7 has a different permission system for packages (`system internal user`)

The only use case I've found for the DSM7 package group is for packages to access files from another package in the same group. I don't know of any other use cases. 

Shared folders don't follow these groups since the `system internal user` setting overrides that (needs confirmation).

Screenshot from 2021-08-28 20-05-24

To reproduce in DSM7

1. make sure the group does not exist: `synogroup --del synocommunity` and possibly `synogroup --del synocommunity__PKG_` and just in case `synouser --del sc-filebrowser`
2. create a group in the control panel or from the command line: `synogroup --add synocommunity`
3. Install a package with the same group. The **old** group will be `synogroup --get synocommunity__PKG_` and the package uses `synocommunity`

I think that works for us, but I let other developers weigh in. @hgy59 @th0ma7?

publicarray:
@BenjV you've done testing on this. What do you think is the beast way to handle the new groups?

1. We keep the name and use the behaviour above or
2. We use a new environment var like `SPK_GROUP` for DSM7 only groups.

BenjV:
No if the user and group are expliciete named in the the privilege file those names are used no matter if the already exists or not.
Only when there is no privilege file (or user/group are missing) the package center uses the name of the package as defined in the INFO file and add _PKG_xxx to it if that username or group name already exists.

My suggestion would still be to use "synocommunity" as group for all packages for DSM 7

publicarray:
@BenjV I think we were saying the same thing. The only change is that in certain conditions in DSM7 the "control panel" group is renamed, not the group for the package if they have the same name.

I'm ok on using "synocommunity" with all packages, It's just not all need it. And it basically removes some of the security boundary that Synology implemented/intended with this model.

BenjV:
I did not experience any renaming of existing groups, only when there is no group definition in the privilege file then  the package center creates a unique name for the to be installed package.

You are right that is is a little less less secure to use the same group for all packages.
But this is only a little less security on the application level not on the OS level (still no root access).

But besides packages that use files created by other package there is another advantage.

There are a lot of package that use configuration files without the functionality to change them.
Like the settings.json file of transmission.
If all packages use the group syncommunity then another package could change those settings.
I was thinking of building a pyhon application what could do that for different packages as a kind of maintenance package.

publicarray:
> There are a lot of package that use configuration files without the functionality to change them.
Like the settings.json file of transmission.
> If all packages use the group syncommunity then another package could change those settings.
> I was thinking of building a pyhon application what could do that for different packages as a kind of maintenance package.

:+1: 

No need, checkout the filebrowser: https://github.com/SynoCommunity/spksrc/pull/4813

PR #4794, opened 2021-08-17
Add PDF documentation from Synology

publicarray:
I'd like to add the PDFs so that there is a bit of history and potentially when the files are removed or changed on Synology's servers they are forever in git. Feel free to add DSM5 if anyone still has it. Files are from https://www.synology.com/en-global/support/developer#tool

I don't mind if this is not merged. Feel free to vote :+1:  :-1: 

Most up-to-date documentation still at https://help.synology.com/developer-guide/


ymartin59:
Sorry but Git is not the right to store binaries as "archives"... soon a Git clone will become larger and build slower

publicarray:
Fair point @ymartin59! Would making a new git repository for these be acceptable instead? Or are you against them being in git in general?

publicarray:
The files update rairly, a few times in a coupple of years and are less than 2MB each. I've found it is usefull to have older versions of the docs.

ymartin59:
If you really want to keep a reference on it... put it into wiki repository `spksrc.wiki `... with a link from relevant "developer" page...

publicarray:
Ok if the past is anything to go by Synology is not good in keeping the same links. 

I'll just keep it in my own repo then

PR #4803, opened 2021-08-22
Fix per-arch builds & add multi-arch & single arch parallel builds

th0ma7:
_Motivation:_  Since code was changed to `all-supported` in order to allow full parallel builds and align github action workflow, it wasn't possible to build package simply using `make arch-apollolake` (required full version such as `make arch-apollolake-7.0`).  The following code change does:
* Builds all `$DEFAULT_TC` packages for same arch when no version passed
* Allows per arch parallel builds when calling `-jX` whith multiple archs passed in argument such as `make -j12 arch-evansport arch-apollolake-7.0`.  Can be disabled with `PARALLEL_MAKE=nop`
* Enable full parallel builds for single arch compilation at per-dependency level (e.g. `make arch-apollolake-7.0`)
* Adds the `build-<arch>-<version>.log` for arch builds (e.g. similar to group builds `all-suported`) for `spk/*` and `cross/*`

_Linked issues:_  #4783 (in theory it supersedes its parallel build functionality)

### Checklist

### spk/ffmpeg
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	100m14.571s
user	90m13.618s
sys	11m16.448s

$ time make arch-apollolake-7.0
real	27m35.783s
user	131m24.426s
sys	17m6.737s
```
### spk/fish
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	3m46.047s
user	3m9.937s
sys	0m37.642s

$ time make arch-apollolake-7.0
real	0m57.700s
user	4m50.890s
sys	0m59.453s
```
### spk/python38
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	13m39.728s
user	11m44.916s
sys	2m29.929s

$ time make arch-apollolake-7.0
real	7m16.851s
user	14m36.478s
sys	3m17.609s
```
### spk/rutorrent
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	20m39.915s
user	18m1.128s
sys	3m46.232s

$ time make arch-apollolake-7.0
real	6m17.894s
user	26m21.084s
sys	5m39.970s
```
### cross/bzip2
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	0m3.837s
user	0m3.586s
sys	0m0.290s

$ time make arch-apollolake-7.0
real	0m1.960s
user	0m4.219s
sys	0m0.436s
```
### cross/lua
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	1m6.893s
user	0m47.150s
sys	0m20.635s

$ time make arch-apollolake-7.0
real	0m24.443s
user	1m3.870s
sys	0m34.759s
```
### cross/svt-av1
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	2m51.662s
user	2m38.332s
sys	0m13.549s

$ time make arch-apollolake-7.0
real	0m32.660s
user	4m4.555s
sys	0m18.168s
```
### spk/umurmur
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	2m9.403s
user	1m36.813s
sys	0m30.935s

$ time make arch-apollolake-7.0
real	0m35.695s
user	2m29.563s
sys	0m47.700s
```
### cross/x265
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	8m59.325s
user	8m40.300s
sys	0m18.258s

$ time make arch-apollolake-7.0
real	3m18.044s
user	11m20.038s
sys	0m27.208s
```
### cross/zlib
```
$ time PARALLEL_MAKE=nop make arch-apollolake-7.0
real	0m3.024s
user	0m2.190s
sys	0m0.888s

$ time make arch-apollolake-7.0
real	0m0.876s
user	0m2.553s
sys	0m1.146s
```

### DIFF
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -57,19 +57,6 @@ SUPPORTED_ARCHS = $(sort $(filter-out $(ARCHS_DUPES), $(AVAILABLE_TOOLCHAINS)))
 #         all archs except generic archs
 LEGACY_ARCHS = $(sort $(filter-out $(addsuffix %,$(GENERIC_ARCHS)), $(AVAILABLE_TOOLCHAINS)))
 
-
-# Relocate to set conditionally according to existing parallel options in caller
-ifneq ($(PARALLEL_MAKE),)
-ifeq ($(PARALLEL_MAKE),max)
-NCPUS = $(shell grep -c ^processor /proc/cpuinfo)
-else
-NCPUS = $(PARALLEL_MAKE)
-endif
-ifeq ($(filter $(NCPUS),0 1),)
-COMPILE_MAKE_OPTIONS += -j$(NCPUS)
-endif
-endif
-
 # Terminal colors
 RED=`tput setaf 1`
 GREEN=`tput setaf 2`
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -67,3 +67,26 @@ version_le = $(shell if printf '%s\n' "$(1)" "$(2)" | sort -VC ; then echo 1; fi
 version_ge = $(shell if printf '%s\n' "$(1)" "$(2)" | sort -VCr ; then echo 1; fi)
 version_lt = $(shell if [ "$(1)" != "$(2)" ] && printf "%s\n" "$(1)" "$(2)" | sort -VC ; then echo 1; fi)
 version_gt = $(shell if [ "$(1)" != "$(2)" ] && printf "%s\n" "$(1)" "$(2)" | sort -VCr ; then echo 1; fi)
+
+# Set parallel build mode
+ifeq ($(PMAKE),)
+# If -j or -l argument passed user must
+# manually specify the value of PMAKE
+# as otherwise this will create too high load
+ifneq ($(strip $(filter -j% -l%, $(shell ps T $$PPID))),)
+PMAKE = nop
+ENV += PMAKE=nop
+else
+# If not set, force max parallel build mode
+PMAKE = max
+ENV += PMAKE=max
+endif
+endif
+
+# Enable stats over parallel build mode
+ifneq ($(filter 1 on ON,$(PSTAT)),)
+PSTAT_TIME = time -o $(PSTAT_LOG) --append
+else
+PSTAT_TIME =
+endif
+PSTAT_LOG = build.stats
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -57,6 +57,38 @@ SUPPORTED_ARCHS = $(sort $(filter-out $(ARCHS_DUPES), $(AVAILABLE_TOOLCHAINS)))
 #         all archs except generic archs
 LEGACY_ARCHS = $(sort $(filter-out $(addsuffix %,$(GENERIC_ARCHS)), $(AVAILABLE_TOOLCHAINS)))
 
+# Set parallel build mode
+ifeq ($(PARALLEL_MAKE),)
+# If not set but -j or -l argument passed, must
+# manually specify the value of PARALLEL_MAKE
+# as otherwise this will create too high load
+ifneq ($(strip $(filter -j% -l%, $(shell ps T $$PPID))),)
+PARALLEL_MAKE = nop
+ENV += PARALLEL_MAKE=nop
+# If not set, force max parallel build mode
+else
+PARALLEL_MAKE = max
+ENV += PARALLEL_MAKE=max
+endif
+endif
+
+# Set NCPUS based on PARALLEL_MAKE
+ifeq ($(PARALLEL_MAKE),nop)
+NCPUS = 1
+else ifeq ($(PARALLEL_MAKE),max)
+NCPUS = $(shell grep -c ^processor /proc/cpuinfo)
+else
+NCPUS = $(PARALLEL_MAKE)
+endif
+
+# Enable stats over parallel build mode
+ifneq ($(filter 1 on ON,$(PSTAT)),)
+PSTAT_TIME = time -o $(PSTAT_LOG) --append
+else
+PSTAT_TIME =
+endif
+PSTAT_LOG = build.stats.log
+
 # Terminal colors
 RED=`tput setaf 1`
 GREEN=`tput setaf 2`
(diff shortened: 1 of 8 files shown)


th0ma7:
@hgy59 This solves the issue (can't find from which #issue) where `make arch-<arch>` would not work anymore.  This relatively simple code change solves this.

Was there anything else to look into as well?

th0ma7:
@hgy59 and @publicarray this is good for your review.
Honestly I think I finally created the "nirvana" for me to build ffmpeg & all packages.
Really much happy with the proposed changes.
In theory supersedes #4783

publicarray:
Would you mind doing one benchmark with `all-suported`? I'm hoping to be little regression, but still like to see if any and how much. I'm using `fish` as a good baseline. No dependencies, but builds long enough, but not too long on a single core. (With `zlib` you are effectively benchmarking your storage medium and docker overhead 🤣)

th0ma7:
@publicarray I noticed one yesterday... I am currently figuring out how to best handle the `all-supported` part as I want it to be able to do both:
1. accept the `-jX` passed argument in order to build multiple archs in parallel (`PARALLEL_MAKE=nop`)
2. build all arch in sequence BUT each one built with `PARALLEL_MAKE=max`
3. and/or a mixture of both using `-l` (`--max-load`) parameter option

Preliminary work for this is in the work...

mk/spksrc.cross-env.mk
```diff
@@ -37,6 +37,7 @@ ENV += TC=$(TC)
 ENV += $(TC_ENV)
 endif
 
-#ifneq ($(COMPILE_MAKE_OPTIONS),)
-#ENV += MAKEFLAGS="$(COMPILE_MAKE_OPTIONS)"
-#endif
+# Set parallel options in caller
+ifeq ($(PARALLEL_MAKE),max)
+MAKEFLAGS += -j$(shell nproc)
```
publicarray, line 42: What happens when PARALLEL_MAKE is a number? IIRC PARALLEL_MAKE was used to limit how many cores to use. I guess it doesn't actually matter, since we can just limit docker to the number of cores we want to give it. Any thoughts @hgy59? If the new behaviour stays, rather than max, true/false would make more sense then.

publicarray (approved):
From my testing if `max` is used this works as intended and improves compile time. But I'm not too qualified to talk about the `.PARALLEL` and `.NOTPARALLEL` commands.

Are there any packages that don't want to have the `-j` option because it breaks something? This PR adds it by default, just something to consider. But I'm all for it :+1:

mk/spksrc.cross-env.mk
publicarray:
```
$ time PARALLEL_MAKE=3 make -C spk/fish arch-x64-7.0

real    3m8.738s
user    2m44.456s
sys     0m24.859s
```

mk/spksrc.common.mk
```diff
@@ -57,19 +57,6 @@ SUPPORTED_ARCHS = $(sort $(filter-out $(ARCHS_DUPES), $(AVAILABLE_TOOLCHAINS)))
...
 LEGACY_ARCHS = $(sort $(filter-out $(addsuffix %,$(GENERIC_ARCHS)), $(AVAILABLE_TOOLCHAINS)))
 
-
-# Relocate to set conditionally according to existing parallel options in caller
-ifneq ($(PARALLEL_MAKE),)
-ifeq ($(PARALLEL_MAKE),max)
-NCPUS = $(shell grep -c ^processor /proc/cpuinfo)
-else
-NCPUS = $(PARALLEL_MAKE)
-endif
```
publicarray, line 67: If we revert this bit (or do something similar) than we can use $NCPUS to get the correct number for -j. Saves an if statement for checking max every time. Looking at the code the cmake part doesn't handle PMAKE=4 only max or none at the moment.

publicarray (approved):

mk/spksrc.common.mk
th0ma7: Sure thing.  Note that I was hoping to complete this asap as I will be offline for another week or so.  I'll finish things up upon my return.

th0ma7: done, it actually simplify the code overall and allowed to fix cmake as well.

th0ma7:
@publicarray looks pretty good now.  I think all issues you found so far are now solved.

Item still remaining (for a later time) is to allow full parallel build within `cross/*`... For now it does handle `PARALLEL_MAKE` option but doesn't build multiple archs in parallel following user passed option `-jX`.

publicarray (approved):
I'm happy with this.
Before merging, I do want to give @hgy59 a little time to have a look if he wants to.

th0ma7:
@publicarray I was finally able to circumvent the issue with building from `cross/*` ...

Long story short, to use `PARALLEL_MAKE` flag it is mandatory to set `.NOTPARALLEL` from within `spksrc.cross-cc.mk`.  This way using both  `-j` with `PARALLEL_MAKE` works great from `spk/*` directory.  Side effect is, it doesn't honor the `-jX` flag when ran from the `cross/*` directory.

As such the `.NOTPARALLEL` is only set if `PARALLEL_MAKE` is not `nop`.  This allows using the `-jX` flag like previously but it won't honor that `-jX` flag when ran in conjunction with `PARALLEL_MAKE`.  Its one or the other, and only `cross/*` is affected.

And sure, such a change requires a few pair of eyes (@hgy59) to confirm all looks good.  I really wanted to complete this before disconnecting for the comming week.  If you find it relevant for a merge please proceed as you see fit in my absence, cheers!

th0ma7:
@publicarray while playing a bit with #4797 I noticed a bizarre failure that may be due to this PR.  I'm unable to reproduce locally, perhaps can you?

I'm thinking perhaps we should revert this PR and look at it a little further?  I could then add a "false" change to `ffmpeg` which is a perfect testing candidate to confirm auto-builds are still ok?

publicarray:
Hm what failures? Did you rebase your PRs on master I can build ffmpeg. Would it help to remove the -j flag if this is a package specific issue? To remove the `-j` flag you can use `$(filter-out -j%, $(TMP_MAKEFLAGS))` https://github.com/SynoCommunity/spksrc/commit/a4fc85ae59285ebd586c0d8df70f0657bdafe3bd

publicarray:
Do you mean `WARNING: none or old ffmpeg libraries were detected` ?

publicarray:
https://github.com/publicarray/spksrc/commit/41beb00b1142048e85f151a7b3a43a2f91eed2a2 should fix it

As a patch: https://github.com/publicarray/spksrc/commit/41beb00b1142048e85f151a7b3a43a2f91eed2a2.patch

th0ma7:
@publicarray actually, https://github.com/SynoCommunity/spksrc/pull/4849 solves it entirely.
And is ready for a quick review & merge.

PR #4752, opened 2021-07-23
Boost: Upgrade to 1.76

ohnoohweh:
Upgraded to Boost 1.76
Files from issue #4746

Motivation: Some programs require Boost 1.76. DSM7 comes with 1.68. SpkSrc has a repository for 1.66
Linked issues: SynoCommunity#4746

### Warning ####
I update the boost package to 1.76. However the programs using the boost lib may not compatible with this newer version

### Checklist
- [ ] Build rule `all-supported` completed successfully
make all-supported does not work on cross/boost
```
make all-supported
make: *** No rule to make target 'all-supported'.  Stop.

```

`make all-archs`

Seems like endless variants:

Ok means that libs are created. Compiling does not stop due to error.

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">

<html>
<head>
</head>

<body>

Arch/Version | 4.3 | 5.0 | 5.1 | 5.2 | 6.0.2 | 6.1 | 6.2 | 6.2.2 | 6.2.3 | 6.2.4 | 7.0
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
88f6281 | nok | ok | ok | ok | ok | ok | ok | ok | ok | ok | '-
aarch64 | '- | '- | '- | '- | '- | ok | ok | ok | ok | ok | ok
alpine | '- | '- | ok | ok | ok | ok | ok | ok | ok | ok | ok
apollolake | '- | '- | '- | '- | '- | ok | ok | ok | ok | ok | ok
armada370 |   |   |   |   |   |   |   |   |   |   |  
armada375 | '- | '- |   |   |   |   |   |   |   |   |  
armada37xx | '- | '- | '- | '- | '- | '- | '- |   |   |   |  
armada38x | '- | '- | '- |   |   | ok |   |   |   | ok  | ok 
armadaxp |   |   |   |   |   |   |   |   |   |   |  
armv7 | '- | '- | '- |   | '- |   |   |   |   |   | ok  
avoton | '- |   |   |   |   |   |   |   |   |   |  
braswell | '- | '- | '- |   |   |   |   |   |   |   |  
broadwell | '- | '- | '- | '- |   |   |   |   |   |   |  
broadwellnk | '- | '- | '- | '- | '- | '- |   |   |   |   |  
bromolow |   |   |   |   |   |   |   |   |   |   |  
cedarview |   |   |   |   |   |   |   |   |   |   |  
[...]

PR #4811, opened 2021-08-26
Boots upgrade

arnebjarne:
_Motivation:_  Upgrade mkvtoolnix so boost can be upgraded.
_Linked issues:_  https://github.com/SynoCommunity/spksrc/pull/4752

### Checklist


arnebjarne:
Quick question:
Since my DS is running 7.0 now - on which DSM's must packages use python2 and on which python3?
I can see that my 7.0 has both python2 and python3, but I wont use python2 if I can avoid it.
I guess some build dependencies is needed to make sure either v3 or v2 is used ?

PR #4813, opened 2021-08-27
Add Filebrowser

publicarray:
_Motivation:_  A file browser for the file system. Note It's for advanced users. Useful to change files/settings without SSH. Probably better than my synoedit package.

To get write permission the files/packages need to be in the `SynoCommunity` group. and be given `+rw` permissions for the group.

This should be obvious but:
**DO NOT forward the port to the public internet! Unless you've configured strong authentication, but I still would not do it.**

### Checklist


cross/filebrowser/Makefile
```diff
@@ -0,0 +1,27 @@
...
+GO_BIN_DIR = $(GO_SRC_DIR)/$(PKG_NAME)
+GO_LDFLAGS = -s -w
+
+PRE_COMPILE_TARGET = filebrowser_pre_compile
+
+include ../../mk/spksrc.cross-go.mk
+
+.PHONY: filebrowser_pre_compile
+filebrowser_pre_compile:
+	$(RUN) make build-frontend
```
hgy59, line 27: just a small improvement
```suggestion
	$(RUN) $(MAKE) build-frontend
```

spk/filebrowser/Makefile
```diff
@@ -0,0 +1,27 @@
+SPK_NAME = filebrowser
+SPK_VERS = 2.17.1
```
hgy59, line 2: very fluent, 2.17.2 available...

spk/filebrowser/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+MAINTAINER = publicarray
+DESCRIPTION = Web File Browser
+DISPLAY_NAME = File Browser
+HELPURL = https://github.com/filebrowser/filebrowser/issues
+LICENSE = Apache-2.0
+CHANGELOG = "First Version"
+
+SPK_GROUP=synocommunity
+GROUP=synocommunity
```
hgy59, line 17: just for the sake of consistency..
```suggestion
SPK_GROUP = synocommunity
GROUP = synocommunity
```

publicarray: Thanks :+1: I want to merge #4789 then having both is no longer necessary.

#4789 requires testing since the group in DSM7 behaves differently then in DSM6. From memory if the group in DSM6 is created and the upgraded to DSM7 it is created again, and the new group is renamed to something like `groupname__PKG__` because it already exists. But I might be wrong and unsure how to test this.

spk/filebrowser/Makefile
```diff
@@ -0,0 +1,28 @@
...
+
+SPK_GROUP = synocommunity
+GROUP = synocommunity
+
+WIZARDS_DIR = src/wizard/
+
+STARTABLE = yes
+SERVICE_USER = auto
+SERVICE_SETUP = src/service-setup.sh
+SERVICE_PORT = 8091
```
hgy59, line 24: Just a friendly reminder to add this port to the wiki page.

ymartin59 (approved):

spk/filebrowser/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = filebrowser
-SPK_VERS = 2.18.0
-SPK_REV = 2
+SPK_REV = 1
+SPK_VERS = 2.20.1
```
hgy59, line 3: any reason to change the order of those?
```suggestion
SPK_VERS = 2.20.1
SPK_REV = 1
```

publicarray: I have no idea how that happened. (magic keyboard shortcut?) Thanks for pointing that out.

PR #4815, opened 2021-08-27
feat: Update fish to 3.3.1

etcusrvar:
_Motivation:_  Update fish to 3.3.1

### Checklist


ymartin59 (approved):

hgy59 (commented):
is this really "gold" linker or just "old" linker?

publicarray:
@hgy59 I know yes really :rofl:, I've not heard about it before until my earlier encounter with the same issue here: https://github.com/SynoCommunity/spksrc/pull/4494#issuecomment-817045003

If you need proof: See the patch file:

`if ("${LD_VERSION}" MATCHES "GNU gold")`

`set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fuse-ld=gold")`

https://en.wikipedia.org/wiki/Gold_(linker)

etcusrvar:
Since we're removing the gold linker stuff from a patch named ...remove-gold-linker..., wouldn't it make sense to remove the patch file entirely?
Are the `add_compile_options...` and `source_group...` lines really needed?

publicarray:
No, The patch removes the lines from the cmake file in the fish source code. The lines begin with -. If they where + then we would add lines.

publicarray:
Those two lines give context about where in the source the patch should be applied. The file was auto generated.

Issue #4746, opened 2021-07-19
Boost upgrade to 1.76 possible ?


ohnoohweh:
### Setup
_Package Name:_  boost
_Package Version:_  1.66

_NAS Model:_  DS918+
_NAS Architecture:_  apollolake
_DSM version:_  7

### Expected behavior
I have a linux application based on Boost 1.76 (Websockets / SSL) that I would love to natively run on Synology.
DSM7 comes with Boost 1.68, however some SSL related functionalities in boost aren't developed yet, so compilation fails.

There is a boost lib in spksrc for boost, however using version 1.66. This compiles well.

### Actual behavior
I tired to change to 1.76 and compile, however compilation fails

./boost/core/cmath.hpp:141:12: error: ‘std::isfinite’ has not been declared
 using std::isfinite;
            ^~~~~~~~
./boost/core/cmath.hpp:142:12: error: ‘std::isnan’ has not been declared
 using std::isnan;
            ^~~~~
./boost/core/cmath.hpp:143:12: error: ‘std::isinf’ has not been declared
 using std::isinf;
            ^~~~~
./boost/core/cmath.hpp:144:12: error: ‘std::isnormal’ has not been declared
 using std::isnormal;
            ^~~~~~~~
./boost/core/cmath.hpp:145:12: error: ‘std::fpclassify’ has not been declared
 using std::fpclassify;
            ^~~~~~~~~~
./boost/core/cmath.hpp:153:12: error: ‘std::signbit’ has not been declared
 using std::signbit;

It seems to have some trouble with the c++ 2011 definitions above.

Is the currently used GCC 7.3.0 fresh enough for this task ?

### Steps to reproduce
_1._  Download develop environment
_2_. Change version to PKG_VERS = 1.76.0 in Makefile
        Change checksums in digest file to
        boost_1_76_0.tar.bz2 SHA1 8064156508312dde1d834fec3dca9b11006555b6
        boost_1_76_0.tar.bz2 SHA256 f0397ba6e982c4450f27bf32a2a83292aba035b827a5623a14636ea583318c41 
        boost_1_76_0.tar.bz2 MD5 33334dd7f862e8ac9fe1cc7c6584fb6d 
_2._  cd /spksrc/cross/boost
_3._ make arch-apollolake-7.0

[boost_1_76_0_buildlog.txt](https://github.com/SynoCommunity/spksrc/files/6841016/boost_1_76_0_buildlog.txt)

[...]

PR #4817, opened 2021-08-28
Boost upgrade

arnebjarne:
_Motivation:_  Current Boost is old. A few packages needs rework to compile
_Linked issues:_ https://github.com/SynoCommunity/spksrc/pull/4752

### Checklist


cross/boost/Makefile
```diff
@@ -1,11 +1,11 @@
 PKG_NAME = boost
-PKG_VERS = 1.66.0
+PKG_VERS = 1.76.0
 PKG_EXT = tar.bz2
 PKG_DIST_NAME = $(PKG_NAME)_$(subst .,_,$(PKG_VERS)).$(PKG_EXT)
 PKG_DIST_SITE = https://boostorg.jfrog.io/artifactory/main/release/$(PKG_VERS)/source
 PKG_DIR = $(PKG_NAME)_$(subst .,_,$(PKG_VERS))
 
-DEPENDS = cross/bzip2 cross/zlib
+DEPENDS = cross/bzip2 cross/zlib cross/python3
```
hgy59: Until now we have support to build boost with python2 or python3 binding.
And we have support to define what boost libraries have to be built.
We can build selective boost libraries. If a package needs the boost libraries for python, the package itself has to define cross/python2 or cross/python3 or cross/python38 as dependency before the dependency of cross/boost.

```suggestion
DEPENDS = cross/bzip2 cross/zlib
```

cross/boost/PLIST
```diff
@@ -1,2 +1,14785 @@
+rsc:include/boost/accumulators/accumulators.hpp
```
hgy59, line 1: PLIST ***must not*** contain include files....

cross/pip_py3/Makefile
```diff
@@ -0,0 +1,14 @@
+PKG_NAME = pip
```
hgy59, line 1: we already have pip for python3/python38, it is cross/pip.
We only have *_py2 cross packages to keep compatible with python2, that hopefully will be removed, when we drop python2 support.

mk/spksrc.cross-env.mk
```diff
@@ -6,6 +6,7 @@ ENV += INSTALL_PREFIX=$(INSTALL_PREFIX)
 
 ifeq ($(strip $(REQUIRE_KERNEL)),1)
 ENV += REQUIRE_KERNEL_MODULE="$(REQUIRE_KERNEL_MODULE)"
+ENV += REQUIRE_KERNEL_HEADERS="$(REQUIRE_KERNEL_HEADERS)"
```
hgy59: this might be from another PR...

hgy59:
@arnebjarne do we realy need an update of boost?
AFAICR I didn't update mkvtoolnix as newer versions are not compatible with the (old) compilers of the synology toolchains.

And there is already #4153 for deluge.

If you realy want to upgrade boost, I propose to keep the current version as cross/boost_1_66 (or boost_legacy, or similar)

I already tried an update some time ago, but didn't succeed. Unfortunately there is no real requirement for a newer boost version given in the original issue #4746. I just would answer with no, it's not possible (and not required).

hgy59 (commented):
sume suggestrions

hgy59:
@arnebjarne your PR is outdated as #5187 is merged into master.
mkvtoolnix is also updated in #5187.
probably your work on cross/rtorrent needs to be added in a new PR.
deluge is already WIP with #4153
domoticz is not yet in progress, but should be migrated to python 3.10 (@th0ma7 can you add domoticz to your list in #5043?), so in #5187 there is cross/boost_1.68 for domoticz 2020.2.

hgy59:
for different situations we now have several boost versions implemented with https://github.com/SynoCommunity/spksrc/pull/5187 https://github.com/SynoCommunity/spksrc/commit/a73ee99087efe492844c9683a19d87db9a2e2f2e

    cross/boost_1.68
    cross/boost_1.70
    cross/boost_1.78
    cross/boost (currently version 1.78)

PR #4821, opened 2021-08-29
SickChill: Use included Python in DSM7

tunisiano187:
_Motivation:_  Explain here what the reason for the pull request is.

### Checklist

### DIFF
--- a/spk/sickchill/Makefile
+++ b/spk/sickchill/Makefile
@@ -4,7 +4,7 @@ SPK_REV = 1
 SPK_ICON = src/sickchill.png
 
 DEPENDS = cross/$(SPK_NAME)
-SPK_DEPENDS = "python3>=3.6:git"
+SPK_DEPENDS = "git"
 PIP=pip3
 
 WHEELS = src/requirements.txt
--- a/spk/sickchill/Makefile
+++ b/spk/sickchill/Makefile
@@ -4,7 +4,12 @@ SPK_REV = 1
 SPK_ICON = src/sickchill.png
 
 DEPENDS = cross/$(SPK_NAME)
+include ../../mk/spksrc.common.mk
+ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 SPK_DEPENDS = "git"
+else
+SPK_DEPENDS = "python3>=3.6;git"
+endif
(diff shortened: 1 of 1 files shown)


spk/sickchill/Makefile
```diff
@@ -4,7 +4,7 @@ SPK_REV = 1
 SPK_ICON = src/sickchill.png
 
 DEPENDS = cross/$(SPK_NAME)
-SPK_DEPENDS = "python3>=3.6:git"
+SPK_DEPENDS = "git"
```
hgy59, line 7: You need at least make this dependent whether the package is created for DSM 7.

tunisiano187: Even if it's part of DSM 7 ?

hgy59:
```suggestion
include ../../mk/spksrc.common.mk
ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
SPK_DEPENDS = "git"
else
SPK_DEPENDS = "python3>=3.6;git"
endif
```

BenjV: @tunisiano187
The python package of the SynoCommunity is different package then the python package included in DSM.
You can install both of them on a Nas.

hgy59: > 
> 
> @tunisiano187
> The python package of the SynoCommunity is different package then the python package included in DSM.
> You can have both of them on a Nas.
> 
> Anything changed in the SynoCommunity package will not effect the Synology package at all.

Yes, but @tunisiano wants to use the included python on DSM 7 and not depend on synocommunity package AIUI.

BenjV: My misunderstood, my mistake!

tunisiano187: Thank you for the informations.
Yes the main goal is that.

publicarray (approved):

spk/sickchill/Makefile
BKSteve: if you do this and use the native python in DSM7 how does this affect the `sickchill/src/service-setup.sh` which contains `PYTHON_BIN="/var/packages/python38/target/bin"` in the first line?

Issue #4546, opened 2021-04-08
[Radarr] Won't start on DS715 after upgrade to version 20210311-15


schumi2004:
### Setup
_Package Name:_ Radarr
_Package Version:_ 20210311-15

_NAS Model:_ DS715
_NAS Architecture:_ Alpine
_DSM version:_ 6.2.4-25556

_Package Name:_ Mono 
_Package Version:_ 5.20.1.34-17

### Expected behavior
The package should start normally from Package Center

### Actual behavior
After clicking 'Run' in Package Center a windows pops up with the message 'Failed to run package service'.

### Steps to reproduce
_1._ Installed the given update, but also a clean installation causes this behavior. v20210311-15
_2._ Removing and manual installing version v20201214-13 makes it work again, upgrading then brakes it.

#### Package log
_find logfiles in `/var/packages/{package}/var/`, for DSM7: `/var/log/packages/{package}.log`, for DSM < 6.0: protocol in Package Center.
_
```
radarr.log 
Thu Apr  8 18:59:37 CEST 2021
Starting radarr command env HOME=/volume1/@appstore/radarr/var LD_LIBRARY_PATH=/volume1/@appstore/radarr/lib /volume1/@appstore/radarr/share/Radarr/bin/Radarr 

radarr_install.log 
Thu Apr  8 17:41:01 CEST 2021
===> Step preuninst. USER=radarr GROUP=sc-download SHARE_PATH=
Removing service configuration radarr.sc
Thu Apr  8 17:41:03 CEST 2021
===> Step postuninst. USER=radarr GROUP=sc-download SHARE_PATH=
Removing user sc-radarr
Thu Apr  8 17:43:56 CEST 2021
===> Step preinst. USER=radarr GROUP=sc-download SHARE_PATH=
Thu Apr  8 17:43:58 CEST 2021
===> Step postinst. USER=radarr GROUP=sc-download SHARE_PATH=
Installing service configuration /var/packages/radarr/conf/radarr.sc
Adding 'sc-radarr' to 'sc-download'
Group Name: [sc-download]
Group Type: [AUTH_LOCAL]
Group ID:   [65536]
Group Members: 
0:[autosub]
1:[nzbget-testing]
2:[sc-nzbdrone]
3:[sc-transmission]
4:[sc-bazarr]
5:[sc-radarr]
Invoke service_postinst
Granting 'sc-radarr' unix ownership on /volume1/@appstore/radarr/var/.config
[...]

PR #4824, opened 2021-08-30
Revert "dotnet: mark armadaxp and alpine as unsupported"

publicarray:
Reverts SynoCommunity/spksrc#4793

Thanks @AnonTester So  one .NET app leads to Invalid Instruction on one CPU and the other one on a  different CPU? I give up. I'll just mark it in the wiki pages. Maybe a future .NET / jellyfin / radarr update magically fixes it. 
https://github.com/SynoCommunity/spksrc/issues/4546#issuecomment-906818900


publicarray:
To summarise:

Jellyfin and armadaxp (DS414) do not work
Radarr and alpine (DS715) do not work

PR #4849, opened 2021-09-09
Use cmake+ninja by default

th0ma7:
_Motivation:_  With recent cmake it is possible to use ninja to build which is more efficient than using default cmake.
_Linked issues:_  This also address weird build issue with #4797

### Checklist


th0ma7:
@hgy59 , @publicarray , @ymartin59 this is now good to go.
It solves build issues currently against master and uses ninja always with cmake.

cross/libaom/Makefile
```diff
@@ -27,6 +27,9 @@ include ../../mk/spksrc.cross-cmake.mk
 # libaom specific cmake options
 CMAKE_ARGS += -DCONFIG_AV1_ENCODER=1
 
+# Disable tests
+CMAKE_ARGS += -DENABLE_TESTS=0
```
publicarray, line 31: Was it the tests that did not like being run in parallel? Otherwise, I wouldn't mind keeping them

publicarray:
Nice, thanks for this! FYI Ninja is only 1 second or so better in compiling fish and umurmur.

cross/libaom/Makefile
th0ma7: No, not related to parallel build (although at first I thought so).  With the usage of ninja all build failures went away.
I removed it as it reduces the compile time significantly (about 10%), went from 480+ down to 440ish ninja build actions.  Feel free to re-enable as you see fit.  Personally I have no need for it.

th0ma7:
> Nice, thanks for this! FYI Ninja is only 1 second or so better in compiling fish and umurmur.

Thnx.  Indeed, it ain't faster, just more robust and efficient and it resolved all build failures I had from github action.  It also allowed me to fix cmake as it only had c & c++ calls as now it auto-detects all cross tools correctly automatically.

PR #4851, opened 2021-09-10
Pimped up the python38

arnebjarne:
_Motivation:_  Improved support of modules availble in Ptyhon 3.8.
_Linked issues:_  

### Checklist


hgy59:
@arnebjarne what is the goal of this PR, what is missing without this?

IMHO the native python38 is built only for the pgen executable to run in the build environment. So there should be no need to pimp up native/python3.
If this is really needed, than it should depend on cross/zlib etc. (might not work from the scratch) to avoid duplicated native definitions. Or even better: use apt-get to install the expected versions instead of building native python (I have in mind to replace all native executables by installation in the dev env).

Again: never add include files in PLIST as this is the definition which files get installed with the target package on the diskstation and packages never will compile code on the diskstation.

arnebjarne:
> 
> 
> @arnebjarne what is the goal of this PR, what is missing without this?
> 
> IMHO the native python38 is built only for the pgen executable to run in the build environment. So there should be no need to pimp up native/python3.
> If this is really needed, than it should depend on cross/zlib etc. (might not work from the scratch) to avoid duplicated native definitions. Or even better: use apt-get to install the expected versions instead of building native python (I have in mind to replace all native executables by installation in the dev env).
> 
> Again: never add include files in PLIST as this is the definition which files get installed with the target package on the diskstation and packages never will compile code on the diskstation.

I noticed in the build log

```
Python build finished successfully!
_dbm                  _gdbm                 _lzma
_sqlite3              _tkinter              readline

```

But if these modules are never used when just reject my PR.
Also I looked at Debians and RedHats build parameters to see what was not in this build.

th0ma7:
@arnebjarne While working on #4902 @hgy59 made me notice this PR.

I had a quick look at it and here's a few comments:
1. We try to keep the `native/*` to the bare minimum.  This means that it only use is to provide a native tool to allow building the `cross/*` package as needed.  As such, adding optimizations to the `native/python38` or any other options is useless as this python binary is not being packaged but rather is only used as a mandatory helper-tool (like gcc) to proceed with the cross-compile building.  As such all other native `expat`, `gdbm`, `ncursew`, shouldn't be needed at all as we only need a basic python38 to help the build process afterwards.
2. You added a lot of patches to `readline`.  Two things comes to mind: a) why exactly? as this becomes really hard to maintain afterwards and b) if you believe this is really needed then I would suggest doing a separate PR to address this specifically which would allow keeping this PR clean & lean.
3. `cross/tcl` you removed a lot of `rsc:*` from the `PLIST`.  This is a case where the package do provide tcl helper scripts by default and really theses needs to stay part of the package.
4. `--use-deprecated=legacy-resolver` good catch.  I faced that issue as well into #4902
5. `gdbm` why `--disable-largefile`?  On the other hand the two other configure args (`--disable-static` and `--enable-libgdbm-compat`) are fine.
6. `pyNaCl` should never be built into the native python as we want a minimal native python anyway, change probably not needed.

At the end of it, the changes specific to `python38` looks otherwise fairly trivial and resume to (and please break into multi-line):
```
DEPENDS += cross/openssl
CONFIGURE_ARGS += --with-ssl-default-suites=openssl

DEPENDS += cross/gdbm
CONFIGURE_ARGS += --with-dbmliborder=gdbm:ndbm:bdb

DEPENDS += cross/expat
CONFIGURE_ARGS += --with-system-expat

CONFIGURE_ARGS += --with-computed-gotos=yes
CONFIGURE_ARGS += --with-system-ffi
CONFIGURE_ARGS += --enable-loadable-sqlite-extensions
CONFIGURE_ARGS += --enable-optimizations
```
It might be appropriate to include a few directly into my on-going PR... will see what I can do with it.

PR #4854, opened 2021-09-11
parallel build: First set of fixes

th0ma7:
_Motivation:_  Follow-up on parallel-build issues... 
- Fix logging to PSTAT at all MAKELEVEL
- PSTAT logging more specific with added [BEGIN] and [END] messages
- Re-uses @publicarray fix for cmake
- Move back to using cmake by default
- Move `libaom` _only_ to using ninja by default
- When building multiple specific archs, force building native in sequence first (e.g. `pre-build-native`)
- Make all possible `cross/*` to use default `COMPILE_TARGET` to allow parallel build (`bzip2`, `dropbear`, `lua`, `mpfr`,  `mtproxy`, `node-exporter`, `p7zip`, `python3`, `python38`, `rtorrent`, `tcp_wrappers`, `udns, unzip`)

_Linked issues:_  This should fix issue found in #4683

### Checklist


spk/ffmpeg/Makefile
```diff
@@ -1,6 +1,6 @@
 SPK_NAME = ffmpeg
 SPK_VERS = 4.3.2
-SPK_REV = 37
+SPK_REV = 38
```
publicarray, line 3: Kind reminder to test/build with `CMAKE_USE_NINJA=0` to check the cmake changes.
FYI you can now build without such a commit. In your fork, use the `Run Workflow` button: https://github.com/SynoCommunity/spksrc/wiki/Publish-a-package-from-Github#run-buildpublish-workflow

th0ma7: This is really really nice.  Great documentation btw, thnx.

cross/busybox/Makefile
```diff
@@ -27,7 +27,7 @@ busybox_configure:
 
 .PHONY: busybox_compile
 busybox_compile:
-	$(RUN) CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) $(MAKE) busybox
+	$(RUN) CROSS_COMPILE=$(TC_PATH)$(TC_PREFIX) $(MAKE) -j $(NCPUS) busybox
```
publicarray, line 30: Are we back to where we need to specify which custom compile target is build in parallel? Kind of like #4783 did?` it's easy to forget, so I liked the default and maybe have an easy opt out. At the same time the errors are not obvious when they are coursed by a parallel build, so maybe this is the best thing after all. CC/ @hgy59

th0ma7: Not at all!  `busybox` overrides default `COMPILE_TARGET` and as such doesn't honor the `MAKEFLAGS` as spawn-up with `-@MAKEFLAGS= $(MAKE)  ...`

Although, there may be a way for theses exceptions to force the `MAKEFLAGS` ...  I'll see if I can come up with something.

th0ma7: @publicarray I've looked at it and no, I haven't found any way to play with `MAKEFLAGS` for theses specific exceptions.  A such, packages exceptionally overriding the `COMPILE_TARGET` must add `-j $(NCPUS)` to parallel build within their `custom_compile_target`. 

For the `busybox` case, it was really easy to migrate to using default `COMPILE_TARGET` and thus parallel build by default (see https://github.com/SynoCommunity/spksrc/pull/4854/commits/e2935b623cfe7a115e8b8c859707d0e963357ec3)

spk/ffmpeg/Makefile
th0ma7: I made some mileage from your advice, I've set `CMAKE_USE_NINJA=0` to default and only forced `libaom` to build using `ninja`.  github action build outputs looks pretty good besides `ffsync` and `squidguard` (considering `minio` on dsm7 is being looked at in #4683)

Did any of the two ever built recently?

th0ma7:
I think I've shaken out all the remaining bugs.  Most of them occurred on a "virgin" tree wheras on an existing tree with pre-compile native and pre-downloaded distrib & toolchain it would not show.

@publicarray if you have a few spare cycles can you give it a shot with variations such as from `cross` and `spk` :
```
$ make all-supported
$ make -jX all-supported
$ PARALLEL_MAKE make all-supported
$ PARALLEL_MAKE make -jX all-supported
--
$ make arch-apollolake
$ make -jX arch-apollolake
$ PARALLEL_MAKE make arch-apollolake
$ PARALLEL_MAKE make -jX arch-apollolake
--
$ make arch-apollolake-7.0
$ make -jX arch-apollolake-7.0
$ PARALLEL_MAKE make arch-apollolake-7.0
$ PARALLEL_MAKE make -jX arch-apollolake-7.0
--
$ make arch-apollolake-7.0 arch-apollolake-6.1
$ make -jX arch-apollolake-7.0 arch-apollolake-6.1
$ PARALLEL_MAKE make arch-apollolake-7.0 arch-apollolake-6.1
$ PARALLEL_MAKE make -jX arch-apollolake-7.0 arch-apollolake-6.1
```
And perhaps also simply directly `make` under `toolchain` or `native`...

spk/ffmpeg/Makefile
th0ma7: Actually, an option to set something like `.BROKEN` empty file in the known broken spk could allow github action to NOT build them, using less CPU cycles and providing better outputs.

fgma:
@th0ma7 I've checked out this branch. Compiling minio for 6.1 works. 7.0 obviously fails which is expected because this is still the old version. 

The problem I found was easily reproducible in spksrc's master branch and is completely unrelated to my PR https://github.com/SynoCommunity/spksrc/pull/4683.

th0ma7:
> @th0ma7 I've checked out this branch. Compiling minio for 6.1 works. 7.0 obviously fails which is expected because this is still the old version.
> 
> The problem I found was easily reproducible in spksrc's master branch and is completely unrelated to my PR #4683.

I was also able to build your branch by copying over all the mk/* files from this PR.  Things seems to be looking good, thnx.  I'll wait for a review to confirm all is still ok before merging.

spk/ffmpeg/Makefile
th0ma7: @publicarray problem now solved.  Using `COMPILE_MAKE_OPTIONS` it was trivial to remove almost all `COMPILE_TARGET` calls making almost every dependencies to now "parallel" build.

th0ma7:
@hgy59 and @publicarray I was able to make all possible `cross/*` to use default `COMPILE_TARGET` to allow parallel build (`bzip2`, `dropbear`, `lua`, `mpfr`,  `mtproxy`, `node-exporter`, `p7zip`, `python3`, `python38`, `rtorrent`, `tcp_wrappers`, `udns, unzip`).

I've applied to same to the `native/*` (only had a handful).  I do not see anything else for now that I could make parallel further...

BTW, python is now much faster to compile, that's including the native build ... :)
```
$ time make arch-apollolake-6.1
real	8m24.407s
user	17m48.153s
sys	3m9.713s
```

cross/unzip/Makefile
```diff
@@ -12,16 +12,16 @@ COMMENT  = UnZip is an extraction utility for archives compressed in .zip format
 LICENSE  = http://infozip.sourceforge.net/license.html
 
 GNU_CONFIGURE = 1
```
hgy59, line 14: `GNU_CONFIGURE = 1` is obsolete with `CONFIGURE_TARGET = nop`

th0ma7: Good catch, I can indeed remove the `GNU_CONFIGURE`.  Want me to fix the other ones while at it?
```
spksrc@spksrc-debian10:~/makeflags/spksrc$ grep GNU_CONFIGURE */*/Makefile | grep 1 | while read LINE; do FILE=$(echo $LINE | cut -f1 -d:); grep -q "CONFIGURE_TARGET = nop" $FILE && echo $FILE; done
cross/libcap/Makefile
cross/p7zip/Makefile
cross/ps3netsrv/Makefile
cross/redis/Makefile
cross/unzip/Makefile
```

publicarray: A ripgrep version, I could not help myself: 

```bash
$ rg --pcre2 -U -l -g '**/**/Makefile' '^(?=[\s\S]*\bCONFIGURE_TARGET\s*=\s*nop\b)(?=[\s\S]*GNU_CONFIGURE\s*=\s*1)[\s\S]*$'
cross/p7zip/Makefile
cross/unzip/Makefile
cross/ps3netsrv/Makefile
cross/redis/Makefile
cross/libcap/Makefile
```

The regex is thanks to https://stackoverflow.com/questions/469913/regular-expressions-is-there-an-and-operator#469951

publicarray:
Thanks for this :+1: And I apologise for merging something that coursed these issues. I knew there was a risk but not to this degree.(I did my tests and those did pass, must have been lucky packages) Sorry. I think If only one or two packages needed a not parallel flag I still would have accepted the earlier PR...

This all started when I noticed compiling fish took too long 2.5 minutes as to less than a minute.

publicarray:
@th0ma7 please revert the `SPK_REV = ` changes.
building a few packages works. e.g. python38 in 6m11s

cross/unzip/Makefile
publicarray: Personally I would like to see more smaller PRs if possible. IMHO you can leave them

mk/spksrc.native-env.mk
```diff
@@ -3,8 +3,6 @@ PKG_CONFIG_LIBDIR = $(INSTALL_DIR)/$(INSTALL_PREFIX)/lib/pkgconfig
 INSTALL_DIR = $(WORK_DIR)/install
 INSTALL_PREFIX = /usr/local
 
-ENV += -u LDSHARED
-ENV += -u MAKEFLAGS
-ENV += -u PKG_CONFIG -u PKG_CONFIG_LIBDIR -u PKG_CONFIG_PATH
+ENV := -u LDSHARED -u PKG_CONFIG -u PKG_CONFIG_LIBDIR -u PKG_CONFIG_PATH $(ENV)
```
publicarray, line 6: I think `+=` is easier to understand than `:=`. Any reason for this change?

publicarray (approved):

publicarray:
Is it just me or does busybox just appear to hang there.
```
$ time make -C cross/busybox arch-x64-6.1
make: Entering directory '/spksrc/cross/busybox'
===>  Building package for arch x64-6.1
make build-arch-x64-6.1
make[1]: Entering directory '/spksrc/cross/busybox'
===>  Building package for arch x64-6.1
make[2]: Entering directory '/spksrc/cross/busybox'
===>  Set up toolchain 
env make --no-print-directory tc_vars > /spksrc/toolchain/syno-x64-6.1/work/tc_vars.mk 2>/dev/null;
===>  Downloading files for busybox
===>    File busybox-1.26.2.tar.bz2 already downloaded
===>  Verifying files for busybox
===>    Checking sha1sum of file busybox-1.26.2.tar.bz2
===>    Checking sha256sum of file busybox-1.26.2.tar.bz2
===>    Checking md5sum of file busybox-1.26.2.tar.bz2
/spksrc/cross/busybox/../../distrib/busybox-1.26.2.tar.bz2
===>  Processing dependencies of busybox
===>  Extracting for busybox
tar -xjpf /spksrc/cross/busybox/../../distrib/busybox-1.26.2.tar.bz2 -C /spksrc/cross/busybox/work-x64-6.1 
===>  Patching for busybox
patch -p0 < patches/100-start_stop_daemon.patch
patching file debianutils/start_stop_daemon.c
===>  Configuring for busybox
cp spksrc_defconfig /spksrc/cross/busybox/work-x64-6.1/busybox-1.26.2/configs/spksrc_defconfig
cat  | while IFS== read option value; do \
  sed -i -r -e "s/^(# )?$option.*$/$option=$value/" /spksrc/cross/busybox/work-x64-6.1/busybox-1.26.2/configs/spksrc_defconfig ; \
done


```

cross/unzip/Makefile
th0ma7: agreed, l'll do a different PR to manage that instaed.

th0ma7: > A ripgrep version, I could not help myself:

I so much need to learn to play with it :)

th0ma7:
> Thanks for this +1 And I apologise for merging something that coursed these issues. I knew there was a risk but not to this degree.(I did my tests and those did pass, must have been lucky packages) Sorry. I think If only one or two packages needed a not parallel flag I still would have accepted the earlier PR...
> 
> This all started when I noticed compiling fish took too long 2.5 minutes as to less than a minute.

Honestly this was on my todo list since day 1.  I'm glad it finally works... but it did took way too many cycles to get there (as usual I guess).  Now lets hope there isn't too many bugs lefts, but considering the amount of packages that succesfully built I think we are good to go!

mk/spksrc.native-env.mk
th0ma7: Reason is, all the `-u` must be always at first as otherwise it pseudo-silently fails.  As such `:=` becomes the only way to ensure that.

I'll add a comment in the code for this...

th0ma7:
> Is it just me or does busybox just appear to hang there.
> I think it's unrelated to this PR

It is unrelated, busybox fails to build on its own as it requires specific parameters.  You must use a package that depends on it to successfully build it.

cross/unzip/Makefile
publicarray: @th0ma7 If you are interested in diving into alternatives like ripgrep, I recently found my 2nd favourite: [`fd`](https://github.com/sharkdp/fd) an alternative to find and [`xh`](https://github.com/ducaale/xh) alternative httpie. For even more checkout: https://github.com/TaKO8Ki/awesome-alternatives-in-rust

mk/spksrc.native-env.mk
publicarray: Thanks
IMHO `unset variables` sounds better, but it doesn't matter that much

fgma:
Just updated my branch. LGTM now.

mk/spksrc.native-env.mk
th0ma7: done with https://github.com/SynoCommunity/spksrc/pull/4857

cross/unzip/Makefile
th0ma7: done with https://github.com/SynoCommunity/spksrc/pull/4857

publicarray:
Looking at the build logs we are not quite done yet unfortunately: https://github.com/SynoCommunity/spksrc/runs/3586108881
https://github.com/SynoCommunity/spksrc/runs/3586108956 

`java-11-openjdk` `rabbitmq` worked before not 100% sure on the others but `deluge` and `minio` should also have worked.

th0ma7:
Well, minio didn't for DSM7 on master, @fgma is working on solving that.
As for the others, I'll have a look at them (`java-11-openjdk`, `rabbitmq` and `deluge`).
As for `squidguard`, its been broken since quite a while.

@publicarray on that, would it be possible to mark some package as broken so they stop being built and fails?

publicarray:
I'm about to create a PR too, at least for java-11-openjdk

publicarray:
> @publicarray on that, would it be possible to mark some package as broken so they stop being built and fails?

I think fixing them or moving them to diyspk or removing them if they are no longer supported by us would do. But no I don't think disabling them is a good way to deal with this.

publicarray:
Reviewing those PRs that address some of those packages was on my next thing to in my list. At the moment there is sickchill & salt I'm still working on reviewing. Honestly I think I'm going to spend time reviewing PRs for the foreseeable feature, we are getting close to 100 open PRs

th0ma7:
Indeed there are way too many of them... Let me know where I can help (and I might as well just pick and choose).
In the meantime I wanted to focus on adressing `ffmpeg` and `tvheadend` pending issues... resulting in more PR.

mk/spksrc.native-env.mk
publicarray: I'm not convinced, at least this wrong for Erlang. Both `ENV +=` and `ENV :=` work the same for this package. Only -u MAKEFLAGS was missing

publicarray: Fix is in #4858

th0ma7: What needs to be happening for the later call is this:
```
env -u LDSHARED -u PKG_CONFIG -u PKG_CONFIG_LIBDIR -u PKG_CONFIG_PATH ...
```
But, as `PARALLEL_MAKE` was previously added to the environment prior to that, it becomes the first item in the list such as:
```
env PARALLEL_MAKE=max -u LDSHARED -u PKG_CONFIG -u PKG_CONFIG_LIBDIR -u PKG_CONFIG_PATH ...
```
And this ends-up throwing an error about the `-u` as unsetting variable needs to be first to the `env` call...

Question could then be, does `PARALLEL_MAKE` really needs to be part of `$(ENV)` or not?  If not then we could switch back to `+=` instead of `:=`.  Although I don't recall the exact reason why it needed to be into `$(ENV)`... perhaps something worth investigating?

publicarray: Ah ok thanks for the explanation :+1: we can leave it with `:=` then.

PR #4858, opened 2021-09-13
Fix build for java11-jdk, rabbitmq and CI cache 

publicarray:
_Motivation:_  followup to #4854

### Checklist


th0ma7:
it seems our path are crossing https://github.com/SynoCommunity/spksrc/pull/4859 ;)

publicarray:
I did try and tell you :) https://github.com/SynoCommunity/spksrc/pull/4854#issuecomment-918594624

cross/java-11-openjdk/Makefile
```diff
@@ -86,8 +86,8 @@ include ../../mk/spksrc.cross-cc.mk
 
 # Filter out any -jN since java's configure will autodetect number of CPU's
 # Use make JOBS=N to set number of CPU's in this package
-TMP_MAKEFLAGS:= $(MAKEFLAGS)
-MAKEFLAGS = $(filter-out -j%, $(TMP_MAKEFLAGS))
+TMP_MAKEFLAGS:= $(COMPILE_MAKE_OPTIONS)
+COMPILE_MAKE_OPTIONS = $(filter-out -j%, $(TMP_MAKEFLAGS))
```
th0ma7, line 90: You will not necessarily address all cases with this as `MAKEFLAGS` follows various `MAKELEVELS` and may still be in effect containing `-jX`.  As such in theory you could still get `make $(MAKEFLAGS) $(COMPILE_MAKE_OPTIONS)` and resulting in `make -jX JOBS=X` with a build failure.

Adding an exceptional `COMPILE_TARGET` such as `-@MAKEFLAGS= $(MAKE) $(COMPILE_MAKE_OPTIONS)` circumvent this entirely.

mk/spksrc.cross-env.mk
```diff
@@ -36,3 +36,7 @@ $(TC_VARS_MK):
 ENV += TC=$(TC)
 ENV += $(TC_ENV)
 endif
+
+ifeq ($(filter $(NCPUS),0 1),)
+COMPILE_MAKE_OPTIONS += -j$(NCPUS)
+endif
```
th0ma7, line 42: Similarily, `spksrc.cross-env.mk` ends-up being included at multiple levels (e.g. `spksrc.spk.mk`, `spksrc.cross-cc.mk`, etc..) leading into possibility of getting multiple `-jX` flags in conjunction with the `MAKEFLAGS` that do follow depending of where you stand for another extra potential `-jX` flag.

th0ma7:
Sorry to say but this is NOACK on this.  A `COMPILE_TARGET` exception within the `cross/java-11-openjdk` is the right path forward to not be impacting other parallel build pieces.

mk/spksrc.cross-env.mk
publicarray: Thanks, I found  that the -j flag originally was added in the `common.mk` the `cross-env` seemed the next logical choice, but you are right it is included elsewhere. I've moved it to `compile.mk`

publicarray:
Thanks for the feedback, I've made some changes. I was aiming for a more generic solution in case there are other packages that fail because of the parallel builds.

cross/java-11-openjdk/Makefile
```diff
@@ -82,12 +82,11 @@ COMPILE_MAKE_OPTIONS += product-images
```
th0ma7, line 82: You are still missing `COMPILE_MAKE_OPTIONS += JOBS=$(NCPUS)` to enable parallel build based on how openjdk uses it.

th0ma7 (approved):

th0ma7 (changes requested):
You are still missing `COMPILE_MAKE_OPTIONS += JOBS=$(NCPUS)` to enable parallel build based on how openjdk uses it.

I just say your other message: https://github.com/SynoCommunity/spksrc/pull/4859/files#r707719314

mk/spksrc.cross-env.mk
th0ma7: While your proposal looks lean & clean i still fear your change could lead into multiple additions of `-j` flag... and getting again the `-jX job-observer forced-reset` message...

publicarray: I don't see how, the `-jN` is in the same file as before, it's even gated to only add it when `$NCPUS` is greater than 1. I've compiled 3 packages and did not see the `-jX job-observer forced-reset` message yet.

publicarray:
I cant reproduce the rabbitmq error...  for deluge #4153 might be interesting

publicarray:
rabbitmq build -> https://github.com/publicarray/spksrc/actions/runs/1235665236

mk/spksrc.compile.mk
```diff
@@ -24,6 +24,10 @@ else
 $(POST_COMPILE_TARGET): $(COMPILE_TARGET)
 endif
 
+ifeq ($(filter $(NCPUS),0 1),)
+COMPILE_MAKE_OPTIONS += -j$(NCPUS)
+endif
+
```
th0ma7, line 30: Have you tried compiling with both `PARALLEL_MAKE=X` and `-jY` flag?
That is where I foresee a collision of your added flag within `COMPILE_MAKE_OPTIONS` and resetting the make build-in `MAKEFLAGS` jobs setting.

mk/spksrc.compile.mk
```diff
@@ -36,7 +40,7 @@ endif
 pre_compile_target: compile_msg
 
 compile_target:  $(PRE_COMPILE_TARGET)
-	@$(RUN) $(PSTAT_TIME) $(MAKE) -j$(NCPUS) $(COMPILE_MAKE_OPTIONS)
+	@$(RUN) $(PSTAT_TIME) $(MAKE) $(COMPILE_MAKE_OPTIONS)
```
th0ma7, line 43: If the `-j` flag ends-up going back here (which I would personally prefer), the previous `ifeq ($(filter $(NCPUS),0 1),)` would be a really nice addition down here.

And again, if the `-j` comes back here then this would mean re-using the `-@MAKEFLAGS= $(MAKE) ...` I did for java-11-openjdk https://github.com/SynoCommunity/spksrc/pull/4859/commits/9d59a0c6b353471f0b4a99e41e5c6cfc9bbd645f and consider it as an exception.

mk/spksrc.common.mk
```diff
@@ -72,6 +72,11 @@ ENV += PARALLEL_MAKE=max
 endif
 endif
 
+# Allow parallel make to be disabled per package
+ifeq ($(DISABLE_PARALLEL_MAKE),1)
+PARALLEL_MAKE = nop
+endif
```
th0ma7, line 78: I much like this, good idea!

mk/spksrc.compile.mk
publicarray: Yes I used the transmission package (`make -j12 TCVERSION=6.1 ARCH=x64 -C spk/transmission all-supported`). It complained about Openssl: `/spksrc/spk/transmission/work-x64-6.1/install//var/packages/transmission/target/lib/libssl.so: file not recognized: File format not recognized` Re-running the build command a few times "fixes" these errors.

mk/spksrc.compile.mk
publicarray: Thanks I think its better safe than sorry, so I'll do that. :+1:

mk/spksrc.compile.mk
publicarray: I recommend to only build without -jN for now. Maybe call it a beta feature. `make TCVERSION=6.1 ARCH=x64 -C spk/transmission all-supported` is already faster. :rocket:

mk/spksrc.compile.mk
publicarray: @th0ma7  If `-j` is back here and `$(NCPUS)` is filtered than https://github.com/SynoCommunity/spksrc/commit/9d59a0c6b353471f0b4a99e41e5c6cfc9bbd645f is not necessary. Since `-j` is not added with `DISABLE_PARALLEL_MAKE=1`.

publicarray:
Builds are failing due network issues: `failed: Connection timed out.`

th0ma7:
Food for thoughts:
- rename `PSTAT` to `PSTAT_ENABLE`
- Add `PSTAT_ENABLE = ON` to default `local.mk`
- replace all `ifeq PSTAT ... @$(MSG) MAKELEVEL: ... >> $(PSTAT_LOG)` with a function call to clean-up the code
- Add `time` to the default container image
- Add `$(PSTAT_TIME)` to $(MAKE) variable, again, to clean-up the code

mk/spksrc.native-env.mk
```diff
@@ -5,7 +5,7 @@ INSTALL_PREFIX = /usr/local
 
 # Unsetting variables MUST always be first
 # as otherwise it fails silently
-ENV := -u LDSHARED -u PKG_CONFIG -u PKG_CONFIG_LIBDIR -u PKG_CONFIG_PATH $(ENV)
+ENV := -u LDSHARED -u MAKEFLAGS -u PKG_CONFIG -u PKG_CONFIG_LIBDIR -u PKG_CONFIG_PATH $(ENV)
```
th0ma7, line 8: If I get this right, re-adding the `-u MAKEFLAGS` solved `rabbitmq`?

th0ma7 (approved):

mk/spksrc.native-env.mk
publicarray: Yes it it fixes native erlang

th0ma7:
@publicarray good job, thnx!

publicarray:
Thank you @th0ma7! Without you this would not have been possible.

PR #4859, opened 2021-09-13
java-11-openjdk: Fix parallel build

th0ma7:
_Motivation:_  A few left over to fix, namely rabbitmq, java and deluge...
_Linked issues:_  Follow-up on #4854

### Checklist

### DIFF
--- a/cross/java-11-openjdk/Makefile
+++ b/cross/java-11-openjdk/Makefile
@@ -87,11 +87,6 @@ ADDITIONAL_CFLAGS = -Os
 
 include ../../mk/spksrc.cross-cc.mk
 
-# Filter out any -jN since java's configure will autodetect number of CPU's
-# Use make JOBS=N to set number of CPU's in this package
-TMP_MAKEFLAGS:= $(MAKEFLAGS)
-MAKEFLAGS = $(filter-out -j%, $(TMP_MAKEFLAGS))
-
 .PHONY: java-11-openjdk_configure
 # Use CONFIGURE_ARGS instead of REAL_CONFIGURE_ARGS to ignore TC_CONFIGURE_ARGS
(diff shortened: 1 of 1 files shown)


cross/java-11-openjdk/Makefile
```diff
@@ -75,6 +76,8 @@ CONFIGURE_ARGS += --disable-hotspot-gtest
 # Disable all GUI related
 CONFIGURE_ARGS += --enable-headless-only
 
+COMPILE_MAKE_OPTIONS += JOBS=$(NCPUS)
```
publicarray, line 79: It automagically detects how many jobs should run, so you can remove the line

publicarray:
Thanks, I think your version looks cleaner, but I'm wondering if there should be a more obvious opt out for a package. Someone not knowing the framework would not see the difference between the default compile target `$(MAKE) -j$(NCPUS) $(COMPILE_MAKE_OPTIONS)` target and `$(MAKE) $(COMPILE_MAKE_OPTIONS)` unless they looked at the compile makefile.

I'm not sure that `$(PSTAT_TIME)` should be in the package makefile. I feel It's more for us framework developers and so package developers should not worry about that one IMHO.

cross/java-11-openjdk/Makefile
th0ma7: aaahhh, that I didn't know... still, i'd add it by default to be consistent but its just me.
@publicarray  and anyhow, I guess I can cancel this PR as yours superseeds it?

th0ma7:
@publicarray really a chat engine of some sort would be best as I guess I'm now puzzled on to which PR should go in :)

th0ma7:
Superseded by #4858 (Thnx @publicarray )

PR #4860, opened 2021-09-14
common.mk: Fix PSTAT_LOG value when not in cross/* or spk/*

th0ma7:
_Motivation:_  
1. PSTAT_LOG value is not redirected to the proper directory when ran outside of `spk/*` or `cross/*` for native, toolchain or other directories.  The following captures the parent pid to try to figure out where the "real" working directory is, and thus send PSTAT_LOG output always at the right place.
2. It now also shows the arch as being native, toolchain, toolkit, etc... when ran outside of `spk/*` or `cross/*`

_Linked issues:_  Follow-up to #4859

### Checklist

### DIFF
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -87,7 +87,12 @@ PSTAT_TIME = time -o $(PSTAT_LOG) --append
 else
 PSTAT_TIME =
 endif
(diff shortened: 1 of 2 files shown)


hgy59:
diyspk folder must be handled similar to spk folder.

Issue #4639, opened 2021-05-21
ffmpeg unable to play mpeg dash audio / radio stream


Rich28:
### Setup
_Package Name:_ ffmpeg
_Package Version:_ 4.3.1-35

_NAS Model:_ DS118
_NAS Architecture:_ ARMv8
_DSM version:_ 6.2.4-25556

### Expected behavior
MPEG dash (Dynamic Adaptive Streaming over HTTP) is now a very established approach of enabling high quality streaming of media content over the Internet and is used by various organizations for radio broadcasting.  I was under the impression that ffmpeg now supports MPEG dash, but nothing seems to occur when I attempt to play an mpeg dash audio stream with SynoCommunity's current version of ffmpeg.  

### Actual behavior
To demonstrate this behavior I have used one the BBCs, lower quality mpeg dash radio streams (https://a.files.bbci.co.uk/media/live/manifesto/audio/simulcast/dash/nonuk/dash_low/aks/bbc_radio_three.mpd), which I guess should be relatively reliable and isn't geofenced to the UK.  Referring to this BBC radio stream, the output from  from /usr/local/ffmpeg/bin/ffmpeg -i https://a.files.bbci.co.uk/media/live/manifesto/audio/simulcast/dash/nonuk/dash_low/aks/bbc_radio_three.mpd is as follows:- 

ffmpeg version 4.3.1 Copyright (c) 2000-2020 the FFmpeg developers
  built with gcc 4.9.4 (crosstool-NG crosstool-ng-1.22.0-169-g87eb166) 20150629 (prerelease)
[...]

PR #4866, opened 2021-09-16
ffmpeg: Add dash demuxer support (requires libxml2)

th0ma7:
_Motivation:_  Add support for DASH Demuxing
_Linked issues:_  #4639

### Checklist

### DIFF
--- a/cross/ffmpeg/Makefile
+++ b/cross/ffmpeg/Makefile
@@ -52,7 +52,7 @@ OPTIONAL_DEPENDS += cross/x265
 OPTIONAL_DEPENDS += cross/shine
 
 DEPENDS += cross/libxml2
-CONFIGURE_ARGS += --enable-demuxer=dash
+CONFIGURE_ARGS += --enable-demuxer=dash --enable-libxml2
 
 DEPENDS += cross/freetype
 CONFIGURE_ARGS += --enable-libfreetype
--- a/cross/ffmpeg/Makefile
+++ b/cross/ffmpeg/Makefile
@@ -52,7 +52,8 @@ OPTIONAL_DEPENDS += cross/x265
 OPTIONAL_DEPENDS += cross/shine
 
 DEPENDS += cross/libxml2
-CONFIGURE_ARGS += --enable-demuxer=dash --enable-libxml2
+CONFIGURE_ARGS += --enable-libxml2
+CONFIGURE_ARGS += --enable-demuxer=dash
 
 DEPENDS += cross/freetype
 CONFIGURE_ARGS += --enable-libfreetype
--- a/cross/ffmpeg/Makefile
+++ b/cross/ffmpeg/Makefile
@@ -51,6 +51,9 @@ OPTIONAL_DEPENDS += cross/x264
 OPTIONAL_DEPENDS += cross/x265
 OPTIONAL_DEPENDS += cross/shine
 
+DEPENDS += cross/libxml2
+CONFIGURE_ARGS += --enable-demuxer=dash
+
 DEPENDS += cross/freetype
 CONFIGURE_ARGS += --enable-libfreetype
 


th0ma7:
@publicarray quick question for you, I have noticed ` failed: Connection timed out.` happening lately and I am wondering if this is momentarily of due to the recent parallel build merge?

Another thought I had: could we cache the download files instead of re-downloading always?  Not that this fixes the issue but would reduce github-action compile time and avoid theses?

publicarray:
>happening lately and I am wondering if this is momentarily of due to the recent parallel build merge?

I have no idea what is causing it. Maybe GitHub has a new rate limiting system?

Regarding caching `/distrib` I'm sure this will course more issues than it solves (old files, unclean builds etc). Rather I would like to propose to download the files only once per run. At the moment every architecture run downloads the same file. To implement that we would need to know which files to download before we run make (not sure if this is possible).

th0ma7:
>  To implement that we would need to know which files to download before we run make (not sure if this is possible).

Well, that should be easy, no?  There already is a `make dependency-list` which currently allows building `native` first for instance.  There could be a call such as `make download-all-supported` to download everything needed for all supported archs... Would that suffice for you?

arnebjarne:
> 
> 
> > To implement that we would need to know which files to download before we run make (not sure if this is possible).
> 
> Well, that should be easy, no? There already is a `make dependency-list` which currently allows building `native` first for instance. There could be a call such as `make download-all-supported` to download everything needed for all supported archs... Would that suffice for you?

I guess a simple bash
```
test -f PACKAGE
```


> 
> 
> > happening lately and I am wondering if this is momentarily of due to the recent parallel build merge?
> 
> I have no idea what is causing it. Maybe GitHub has a new rate limiting system?
> 
> Regarding caching `/distrib` I'm sure this will course more issues than it solves (old files, unclean builds etc). Rather I would like to propose to download the files only once per workflow run. At the moment every architecture run downloads the same file. To implement that we would need to know which files to download before we run make (not sure if this is possible).

>Regarding caching /distrib

As long as you check the file against the digest checksum, caching ./distrub should be safe IMHO.

th0ma7:
> > Regarding caching /distrib
> 
> As long as you check the file against the digest checksum, caching ./distrib should be safe IMHO.

Actaully, you're right.  As all builds start with validation of the digests we're in theory good to go and `distrib/*` and its sub-directory could be entirely cached, reducing download significantly and, again, make the build faster.

arnebjarne:
> 
> 
> > > Regarding caching /distrib
> > 
> > 
> > As long as you check the file against the digest checksum, caching ./distrib should be safe IMHO.
> 
> Actaully, you're right. As all builds start with validation of the digests we're in theory good to go and `distrib/*` and its sub-directory could be entirely cached, reducing download significantly and, again, make the build faster.

I am not well trained in Git :) but itsnt there something about git submodules?
( https://git-scm.com/book/en/v2/Git-Tools-Submodules )
ie

i git clone https://github.com/SynoCommunity/spksrc
and get an empty ./distrib by default (to safe download time).

but if i do 
git clone --recurse-submodules https://github.com/SynoCommunity/spksrc
it could download all packages in https://github.com/SynoCommunity/spksrc/distrib
?

and i guess the github build thing have a direct access to the files via their infrastructure 

Just an idea.
BR,
Bjarne

hgy59:
> and i guess the github build thing have a direct access to the files via their infrastructure

No this does not work as you expect, no files in the distrib folder are under source control.

hgy59:
> As long as you check the file against the digest checksum, caching ./distrib should be safe IMHO.

To avoid deadlocks for the (rare) case of digests change for equal file name I like the idea to cache the download files within a workflow only.

th0ma7:
@hgy59 duno if it is due to your latest changes but builds are now all ok!

PR #4869, opened 2021-09-18
add github download action

hgy59:
_Motivation:_  Try to cache downloaded files in digests folder for faster run of github build action
_Linked issues:_  https://github.com/SynoCommunity/spksrc/pull/4866#issuecomment-922116351

### Remarks
This gives some improvements for the github build action, as the dependency-list is created only once in the prepare job and not for the download and every build job again.
The download job downloads all referenced source files and add those to the cache for the current workflow.
And the build jobs use the source files in distrib folder from that cache.

### DIFF
--- a/.github/actions/build.sh
+++ b/.github/actions/build.sh
@@ -25,6 +25,7 @@ echo "::endgroup::"
 echo "::group:: ---- find dependent packages"
 
 # filter for changes made in the spk directories and take unique package name (without spk folder)
+SPK_TO_BUILD+=" "
 SPK_TO_BUILD+=$(echo "${GH_FILES}" | tr ' ' '\n' | grep -oP "(spk)/\K[^\/]*" | sort -u | tr '\n' ' ')
 
 # filter for changes made in the cross and native directories and take unique package name (including cross or native folder)
@@ -50,10 +51,10 @@ done
 
 # fix for packages with different names
 if [ "$(echo ${SPK_TO_BUILD} | grep -ow nzbdrone)" != "" ]; then
-    SPK_TO_BUILD+=sonarr
+    SPK_TO_BUILD+=" sonarr"
 fi
 if [ "$(echo ${SPK_TO_BUILD} | grep -ow python)" != "" ]; then
-    SPK_TO_BUILD+=python2
+    SPK_TO_BUILD+=" python2"
 fi
 
 # remove duplicate packages
--- a/.github/actions/build.sh
+++ b/.github/actions/build.sh
@@ -1,9 +1,11 @@
 #!/bin/bash
 
-# github action to build the packages depending on changed files
+# Part of github build action
+# 
+# build the packages depending on evaluated packages (see prepare.sh)
 #
 # Functions:
-# - Build all packages depending on files defined in ${GH_FILES}.
+# - Build all packages depending on files defined in ${ARCH_PACKAGES} or ${NOARCH_PACKAGES}.
 # - Build for arch defined by ${GH_ARCH} (e.g. x64-6.1, noarch, ...).
 # - Successfully built packages are logged to $BUILD_SUCCESS_FILE.
 # - Failed builds are logged to ${BUILD_ERROR_FILE} and annotated as error.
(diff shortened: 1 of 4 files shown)


hgy59:
To see this in action see: https://github.com/SynoCommunity/spksrc/actions/runs/1249286414
There is only a small change after this (renamed download job from "Download referenced packages" to "Download sources".

publicarray:
Nice work. Do really you need the docker image to get package changes and download? Wouldn't it be quicker to just get dependencies you need (jq) from the host OS?

publicarray:
Or a smaller image? Ours is about 1GB and is downloaded 3 times

hgy59:
> 
> 
> Or a smaller image? Ours is about 1GB and is downloaded 3 times

No, it is downloaded 16 times due to the build-matrix. Unfortunately I couldn't find a solution to cache the used docker images 😞

publicarray:
Thanks, since the image is downloaded from GitHub’s own server it’s not too bad, 19 seconds

publicarray:
Im on a phone but this looks good to merge

publicarray (approved):

th0ma7:
Great job @hgy59 !

th0ma7:
Does that mean that downloads (e.g. `distrib`) will start being cached for all new github-action ?

hgy59:
> 
> 
> Does that mean that downloads (e.g. `distrib`) will start being cached for all new github-action ?

We could do this, but now the distrib-cache is bound to one workflow only. On every new github bulid action first the required files are downloaded and used by every arch specifique build. Only if you manually re-run the the same workflow the cached files are used.

publicarray:
Regarding caching the whole folder. There are some gotchas.

The CI doesn't update the cache with the same key later. So you can't incrementally add more files. 
There is a cache size limit. If I remember correctly it's 5GB. So if one digest is changed rebuilding the cache is costly. Most PRs do change at lest one digest so caching everything beforehand is not worth it.

Happy to be proven wrong 😃

th0ma7:
> The CI doesn't update the cache with the same key later. So you can't incrementally add more files.

humm, is it then possible to run a cache update on a regular basis (e.g. weekly/monthly) to cache "most" of `distrib` (e.g. focusing on DSM-6.1/7.0 toolchains & `cross/*/digests`)?  BTW, I have no clue how this side of things works so disregard as you see fit.

> There is a cache size limit. If I remember correctly it's 5GB. So if one digest is changed rebuilding the cache is costly. Most PRs do change at least one digest so caching everything beforehand is not worth it.

Indeed but all it's dependencies would still be cached.  I'm thinking of ffmpeg for instance... there are 50+ dependencies.  Changing one only means that 49+ are still cached and wouldn't be re-downloaded.  But then again the 5GB, no clue where this puts us in terms of capacity to cache `distrib`.

hgy59:
> Indeed but all it's dependencies would still be cached. I'm thinking of ffmpeg for instance... there are 50+ dependencies. Changing only one digests file means that the remaining 49+ are still cached and wouldn't be re-downloaded. But then again the 5GB, no clue where this puts us in terms of capacity to cache `distrib`.

The download of all ffmpeg dependencies takes only 2 minutes and is now done only once instead of  11 times on each arch-build.
When we cache the distrib folder over all build actions, we will not detect invalid download links.
BTW the cache is not related to digests, the files are cached by the filename.

The toolchains are already cached (not the downloaded files but the extracted work folders).

And if no build action is run for 7 days, both caches (toolchains and distrib folder) are deleted anyway...

publicarray:
> The download of all ffmpeg dependencies takes only 2 minutes and is now done only once instead of 11 times on each arch-build.

Thanks for implementing that. I think that is good enough for now. I don't think there is a major need to cache more. 

> BTW the cache is not related to digests, the files are cached by the filename

Yes but you can set the key to include a hash of all the digests (like the toolchains cache code does) and in doing so the cache becomes worthless if a single digest changes. (the key is different, so no cache is available to fit that key). And if you go the file name only route than yes we have had issues where the digest changed but not the filename.

> And if no build action is run for 7 days, both caches (toolchains and distrib folder) are deleted anyway...

True, but that is not a fix for the above problems.

publicarray:
Sorry If I'm a bit dismissive, maybe It's because when caching on the web, I've experienced many issues. One Idea I had was to set the cache key per package, That would solve the issues but probably requires rewriting most of the build workflow.

And welcome back @hgy59 :smile: !

hgy59:
I suppose the main issue was the parallel download of the same files due to the build matrix.

hgy59:
> And welcome back @hgy59 😄 !

Thanks, I had a great vacation, but I'll be very busy for the next few weeks...

Issue #4878, opened 2021-09-21
[SynoCli File Tools] rg throws "error while loading shared libraries: libpcre2-8.so.0"


screendriver:
### Setup
_Package Name:_ SynoCli File Tools
_Package Version:_ v2.3-10

_NAS Model:_ DS920+
_NAS Architecture:_ INTEL Celeron J4125
_DSM version:_ DSM 7.0-41890

### Expected behavior
`rg` should search for a regex pattern

### Actual behavior

```sh
rg: error while loading shared libraries: libpcre2-8.so.0: cannot open shared object file: No such file or directory
```

### Steps to reproduce
_1._ just type `$ rg`

publicarray:
I can reproduce, It's Interesting because when I build it locally (diyspk/) the lib is statically linked.

publicarray:
If you want you can download the build with the fix from the CI here: [packages.zip](https://nightly.link/publicarray/spksrc/actions/runs/1257302819/packages.zip)

PR #4879, opened 2021-09-21
Fix ripgrep #4878

publicarray:
When building synocli-file the dependencies from other packages include
PCRE2 and contaminate the build for ripgrep. The lib was found and
dynamically linked. Setting PCRE2_SYS_STATIC=1 ensures static linking

>Enabling the PCRE2 feature works with a stable Rust compiler and will attempt to automatically find and link with your system's PCRE2 library via pkg-config. If one doesn't exist, then ripgrep will build PCRE2 from source using your system's C compiler and then statically link it into the final executable. Static linking can be forced even when there is an available PCRE2 system library by either building ripgrep with the MUSL target or by setting PCRE2_SYS_STATIC=1.

From https://github.com/BurntSushi/ripgrep#building

_Linked issues:_  #4878

### Checklist

### DIFF
--- a/cross/ripgrep/Makefile
+++ b/cross/ripgrep/Makefile
@@ -12,9 +12,11 @@ DEPENDS =
 UNSUPPORTED_ARCHS += $(OLD_PPC_ARCHS)
 
 HOMEPAGE = https://github.com/BurntSushi/ripgrep
-COMMENT  = ripgrep recursively searches directories for a regex pattern 
+COMMENT  = ripgrep recursively searches directories for a regex pattern
 LICENSE  = public domain/Unlicense
 
+# https://github.com/SynoCommunity/spksrc/issues/4878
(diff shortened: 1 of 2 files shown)


cross/ripgrep/Makefile
```diff
@@ -12,9 +12,11 @@ DEPENDS =
 UNSUPPORTED_ARCHS += $(OLD_PPC_ARCHS)
 
 HOMEPAGE = https://github.com/BurntSushi/ripgrep
-COMMENT  = ripgrep recursively searches directories for a regex pattern 
+COMMENT  = ripgrep recursively searches directories for a regex pattern
 LICENSE  = public domain/Unlicense
 
+# https://github.com/SynoCommunity/spksrc/issues/4878
```
ymartin59, line 18: Please discard such comment as reference to issue is already in Git commit

ymartin59 (approved):

PR #4889, opened 2021-09-28
Add extract size to spk info file

publicarray:
This helps warn users in the rare case during the install when there isn't enough space to install the package.
https://help.synology.com/developer-guide/synology_package/INFO_optional_fields.html#field-name--extractsize


ymartin59 (approved):

hgy59:
I suppose we should introduce a Makefile variable to allow packages to increase the "extracsize",
i.e. for packages that download other packages at installation time (as do all the "*-custom" packages)
or to define the minimal disk space due to a package specific requirement.

When reading
_Description: This value indicates the minimal space to install a package. It will be used to prompt the user if there is enough free
space to install it._
it does not mean "extractsize" but it the minimal space required.

As we do **not** install packages in the system space (`install_type="system"`) and NAS devices have normally huge disk space this is not really mandatory...

PR #4870, opened 2021-09-18
try to fix sickchill

hgy59:
_Motivation:_  SickChill does not run on arm archs
_Linked issues:_  #4867

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

### Remarks
- avoid noarch due to arch specific requirements (wheels)
- remove python38 site-packages from wheelhouse
- adjust service-setup.sh for DSM 7

### TODO
- sickchill still runs on x64 (and may be on x86) only (fails to run armv7-6.1, aarch64-6.1)

Issue #4867, opened 2021-09-18
Wheel build as Python38 (cp38)


BKSteve:
I wish to know how to make a package build it's wheel as cp38 and not cp37 when doing `make all-supported`
[https://github.com/SynoCommunity/spksrc/pull/4835#issuecomment-920772718](https://github.com/SynoCommunity/spksrc/pull/4835#issuecomment-920772718)

This is a common error which x64 systems can resolve and download the tar.gz but when installing on some models (arm) it causes the system to fail it's install.
The configuration is to use Python 3.8 but this wheel for some reason is built with 3.7 and I wish to know how to resolve.
`ERROR: tornado-6.1-cp37-none-any.whl is not a supported wheel on this platform.`

From build spk.
```
Collecting tornado==6.1
  Downloading tornado-6.1.tar.gz (497 kB)
...
  Building wheel for tornado (setup.py): started
  Building wheel for tornado (setup.py): finished with status 'done'
  Created wheel for tornado: filename=tornado-6.1-cp37-cp37m-linux_x86_64.whl size=414483 sha256=8705a07c6363477174bceb6d1473adf870f4abbbd2f6cf76e2ea794607bacf24
  Stored in directory: /tmp/pip-ephem-wheel-cache-ssmx83xm/wheels/02/62/2c/f52c662d8ae374c4abda0c13ce432fb58eb1c75281a27b406c
```
Grateful for any guidance.

hgy59:
The error is in `spk/sickchill/Makefile` (I suppose you want to build sickchill?)
There must be something like `BUILD_DEPENDS = cross/python38 cross/setuptools cross/pip cross/wheel` to use python38 for the creation of cross compiled wheels. Otherwise it uses python from the path of the build environment i.e. the docker image.

hgy59:
Another error of the SickChill Package:
With dependency of wheels for tornado and wrapt the package is not _noarch_ anymore, i.e. those wheels need to be cross compiled and the packages get arch specific.

BKSteve:
But why do I need to do all this cross compilation when the x64 based devices load and run the package as is?
SickChill has the code to pull all the necessary components in and builds the site-packages / virtual env with all needed items.

It just these older ARM devices that have an issue and can't get past the initial basic install to actually run to pull all the pieces.
Just asking is all.

BKSteve:
> Another error of the SickChill Package:
[...]

Issue #4856, opened 2021-09-12
SickChill wont start after version 20210729-2 was installt on synolgy Dsm 6.2.4-25556


xmenxmen:
### Setup
_Package Name:_ SickChill 
_Package Version:_ version 20210729-2

_NAS Model:_ DS5113+
_NAS Architecture:_ 
_DSM version:_  6.2.4-25556

### Expected behavior
Tell us what should happen

### Actual behavior start
Tell us what happens instead It start and stop 1 min after

### Steps to reproduce
_1._ 
_2._ 
_3._ 

### Setup
_Package Name:_ 
_Package Version:_ 

_NAS Model:_ 
_NAS Architecture:_ 
_DSM version:_ 

### Expected behavior
Tell us what should happen

### Actual behavior
Tell us what happens instead

### Steps to reproduce
_1._ 
_2._ 
_3._ 

#### Package log
_Installer log: find logfile `/var/log/packages/{package}.log`, for DSM<6: `/var/packages/{package}/target/var/{package}_installer.log` and protocol in Package Center.
You find service log and other log files in `/var/packages/{package}/target/var`.
_
```
Insert the package log(s) here
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here
```

ejc1967:
Hi,

I encountered the same problem, tried upgrading from 20210329-1, as well as a clean install.
Both times SickChill starts, but stop again after less than a minute.
Only ERROR I could find is this: ERROR: SQLAlchemy-1.3.20-cp37-none-any.whl is not a supported wheel on this platform.
See included partial package log from the time of the update and clean install.
After reinstalling 20210329-1 and manually fixing the small env issue, SickChill works just fine again.

**Setup**
Package Name: SickChill
Package Version: version 20210729-2

NAS Model: DS3612xs
NAS Architecture: Bromolow (x64)
DSM version: DSM 6.2.4-25556 Update 2

**Package log**

2021/09/10 07:14:02	upgrade sickchill 20210329-1 Begin start-stop-status stop
Stopping sickchill ...
2021/09/10 07:14:22	upgrade sickchill 20210329-1 End start-stop-status stop ret=[0]
2021/09/10 07:14:33	upgrade sickchill 20210329-1 Begin preupgrade
2021/09/10 07:14:33	===> Step preupgrade. USER=sickchill GROUP=sc-download SHARE_PATH=
[...]

PR #4743, opened 2021-07-17
Sickchill dsm7

miigotu:
I was trying to rework this package last night and this is as far as I got. sickchill version needs pinned in the WHEELS line to the latest version once we get it to build.

Sickchill itself does not need cross compiled, just the dependencies. It would be cool if the framework did not require every wheel to be built from source, only build the ones that are not pure python and do not have a valid release that would work for the target arch.

build_depends and depends dont seem right, just need pynacl, poetry, and mediainfo on top of what is included with python38
### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

PR #4901, opened 2021-10-09
Sickchill DSM 6, 7

BKSteve:
To get SickChill a functional SPK in SynoCommunity 
Relates to
#4856 #4870 #4867 #4743 

The build is not perfect and has issues due to Python38 build causing cffi package build errors for a number of architectures, along with some rust build issues from that, there is also a lxml issue with man pages throwing an error in the build process.
To overcome most of the Python38 issues (other than cffi) the cross makefile lines 89 & 90 were changed to use newest versions of other packages required to build `"setuptools>=58.0.0"  "wheel>=0.36.2" "cffi>=1.14.5"`. Yes going >= was what i chose.
I'm sure there's a way to do this without affecting everybody else using cross/python38 and ask for input from others.

To build the spk is a 3 step process make `all-supported` then `spkclean` followed by `all-supported` provides the 11/11 architectures and DSM 6&7 builds which are fully functional, load and work. 
All the builds and testing are here [https://github.com/SickChill/SickChill/issues/7504](https://github.com/SickChill/SickChill/issues/7504) if you wish to check anything out.

This pull request has makefile numbered as version 3 for SynoCommunity sequence which is the testing 20210729-17.spk packages in 7504.

### Checklist
- [x] Build rule `all-supported` completed successfully after spkclean!


cross/dante-sockd/Makefile
```diff
@@ -14,10 +14,4 @@ LICENSE = BSD/CMU
 GNU_CONFIGURE = 1
 CONFIGURE_ARGS = --disable-client --without-libwrap --without-bsdauth --without-gssapi --without-krb5 --without-upnp --without-pam
 
-PRE_CONFIGURE_TARGET = dante_pre_configure
-
 include ../../mk/spksrc.cross-cc.mk
-
-.PHONY: dante_pre_configure
-dante_pre_configure:
-	$(RUN) autoreconf -i -f -v
```
th0ma7, line 23: This addition along with the patch allows it to build on DSM7.  Otherwise it does the following:
```
$ make arch-armv7-7.0
...
checking getaddrinfo() error symbols... configure: error: in `/home/spksrc/BKSteve/spksrc/cross/dante-sockd/work-armv7-7.0/dante-1.4.2':
configure: error: error: getaddrinfo() error value count too low
See `config.log' for more details
make[2]: *** [../../mk/spksrc.configure.mk:50: configure_target] Error 1
```

spk/sickchill/src/service-setup.sh
```diff
@@ -1,12 +1,12 @@
 PYTHON_DIR="/var/packages/python38/target"
 PIP=${SYNOPKG_PKGDEST}/env/bin/pip3
 PATH="${SYNOPKG_PKGDEST}/bin:${SYNOPKG_PKGDEST}/env/bin:${PYTHON_DIR}/bin:${PATH}"
-HOME="${SYNOPKG_PKGDEST}/var"
-VIRTUALENV="${PYTHON_DIR}/bin/virtualenv"
+HOME="${SYNOPKG_PKGVAR}"
+VIRTUALENV="${PYTHON_DIR}/bin/python3 -m venv"
```
th0ma7, line 5: As `${PYTHON_DIR}/bin` is already in your path, your `VIRTUALENV` call can simply use `python3` without the need of adding the path once more.

th0ma7 (commented):
That cffi thing still bugs me, there must be a way to avoid this compile issue... I tried a few things but wasn't able to identify the root cause yet.

cross/dante-sockd/Makefile
BKSteve: That's very interesting and somewhat odd.

th0ma7: Indeed but it's the only workaround I was able to find online...   Feel free to figure a better option, at least on master it now build ok with it.

BKSteve: I don't have the patch for dante-sockd and I can't get that error message when building armv7-7.0 like you show.
If make armv7 then I get the cffi error and second build after spkclean is good.
I'm a little confused now.

th0ma7: My guess is that your tree wasn't rebased against master as this update to version 1.4.3 is the last commit I pushed in and therefore your tree is removing this last commit on master.

The error I'm referring to is happening on all DSM7 builds and was caught on github-action from #4797.

th0ma7: So you could just discard all the dante changes and use what's on master.

cross/python38/Makefile
```diff
@@ -86,8 +86,8 @@ python38_post_install: $(WORK_DIR)/python-cc.mk
 	@$(RUN) $(PYTHON_NATIVE) -m crossenv $(STAGING_INSTALL_PREFIX)/bin/python$(PKG_VERS_MAJOR_MINOR) $(WORK_DIR)/crossenv/
 	. $(WORK_DIR)/crossenv/bin/activate && $(RUN) wget https://bootstrap.pypa.io/get-pip.py -O - | build-python
 	. $(WORK_DIR)/crossenv/bin/activate && $(RUN) wget https://bootstrap.pypa.io/get-pip.py -O - | python
-	. $(WORK_DIR)/crossenv/bin/activate && $(RUN) build-pip install "setuptools==44.1.0"  "wheel==0.36.2" "cffi==1.14.5"
```
smaarn, line 89: If I were you I would really avoid using `>=` mechanisms. It would make the build non repeatables.
Similarly it would be worth updating all `cross/*` equivalents to the same versions to be on the safe side.

th0ma7: Actually, this is superseeded by #4902 and should be removed from your PR.

cross/dante-sockd/Makefile
th0ma7: Please all `dante` related changes as this is most probably due to your branch not being rebased against master as this was merged part of PR #4898 and yours removes it entirely and it isn't related to Sickchill at all.

cross/python38/Makefile
BKSteve: Yes, the `>=` was an ugly work around during the testing and the errors being thrown that's why I mentioned in first post.

Now 3.8.12 is in master I'll update and rebase instead of using the combination of clone of th0ma7 3.8.12 branch from his repo along with my sickchill copied to that clone.

BKSteve:
Still active

BKSteve:
Still active, just deleted and recreated branch.

Here's my log of `88f6281` build where it looks like poetry is trying getting cffi from 35 down until is gets to 3.3.2 for success.
Line 17680 is the first Error: 
[Uploading build-88f6281.zip…]()

But does this constitute an actual issue? As a spk was built, so it should install and load. 

> Note, I only have x64 DSM so have to get others to test them.

PR #4845, opened 2021-09-06
Salt 3003.3 update

arnebjarne:
_Motivation:_  Current version is outdated
_Linked issues:_  

### Checklist
- [x] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4902, opened 2021-10-09
Update pyton3 to 3.7.12 & python38 to 3.8.12, fix cffi build issues and update requirements.txt

th0ma7:
_Motivation:_  Weird conjunction of elements leading to build issue for python38 where `cffi` wheel would fail to build.  This was due to too old wheel dependencies in conjunction with a automated newer pip which now requires `--use-deprecated=legacy-resolver` to download properly all wheels.  Also note that:
1. the `pip` version is now enforced to keep compatible with the `--use-deprecated` flag which will be removed eventually.  Otherwise down the road if pip continues being updated automatically to latest version by default it will stop working.
2. `native/python38` version `pip` (and other requirements) used is set to be identical to cross/python38
3. Installation of `setuptools`, `wheel` and `cffi` must be done in sequence in `python38/Makefile` as otherwise 64-bit platforms (aarch64 and x64 specifically) gets into this weird state where it tries to install `cffi` but `wheel` was just deleted in order to be updated as well resulting in `error: invalid command 'bdist_wheel'`.

_Linked issues:_  #4901, closes  #4851
1. Enable Link-Time Optimization except for armv5
2. Use system `expat`
3. Enable `gdbm`

### Checklist

### Error found
The python38 would silently fail on `cffi` for all archs except `x64`.  Previous silent error:
```
Building wheels for collected packages: cffi
  Building wheel for cffi (setup.py): started
  Building wheel for cffi (setup.py): finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /home/spksrc/rsync2/spksrc/spk/python38/work-armv7-7.0/crossenv/cross/bin/python -u -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-od4r_4f8/cffi_0b4316a0143b40518e05ae31396c9192/setup.py'"'"'; __file__='"'"'/tmp/pip-install-od4r_4f8/cffi_0b4316a0143b40518e05ae31396c9192/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-2n8hfkdp
       cwd: /tmp/pip-install-od4r_4f8/cffi_0b4316a0143b40518e05ae31396c9192/
  Complete output (13 lines):
  *******************************************************
  * Crossenv has leaked into another Python interpreter!
  * You should probably file a bug report.
  * Version 3.8.11 (default, Sep 28 2021, 16:27:25)
  [GCC 8.3.0]
  * Executable /home/spksrc/rsync2/spksrc/spk/python38/work-armv7-7.0/crossenv/cross/bin/python
  *******************************************************
  usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
     or: setup.py --help [cmd1 cmd2 ...]
     or: setup.py --help-commands
     or: setup.py cmd --help

  error: invalid command 'bdist_wheel'
  ----------------------------------------
  ERROR: Failed building wheel for cffi
  Running setup.py clean for cffi
Failed to build cffi
```

### DIFF
--- a/cross/java-11-openjdk/PLIST
+++ b/cross/java-11-openjdk/PLIST
@@ -195,7 +195,7 @@ lib:jvm/openjdk-11.0.12/lib/libawt_headless.so
 lib:jvm/openjdk-11.0.12/lib/libawt.so
 lib:jvm/openjdk-11.0.12/lib/libdt_socket.so
 lib:jvm/openjdk-11.0.12/lib/libextnet.so
-lib:jvm/openjdk-11.0.12/lib/libffi.so.7
+lib:jvm/openjdk-11.0.12/lib/libffi.so.8
 lib:jvm/openjdk-11.0.12/lib/libfontmanager.so
 lib:jvm/openjdk-11.0.12/lib/libharfbuzz.so
 lib:jvm/openjdk-11.0.12/lib/libinstrument.so


native/python38/Makefile
```diff
@@ -26,8 +26,8 @@ PIP_NATIVE = $(WORK_DIR)/../../../native/$(PKG_NAME)/work-native/install/usr/loc
 python38_native_post_install: $(WORK_DIR)/python-native.mk
 	@$(MSG) Installing pip, setuptools, cffi and cross env
 	@$(RUN) wget https://bootstrap.pypa.io/get-pip.py
-	@$(RUN) $(PYTHON) get-pip.py "pip==21.1.3"
-	@$(PIP) install "setuptools==49.6.0" "cffi==1.14.6" "crossenv==1.0"
+	@$(RUN) $(PYTHON) get-pip.py "pip==20.3.3"
```
BKSteve: You sure about using 20.3.3? Isn't it 21.2.4

hgy59:
> 
> 
> @hgy59 you're the last who was able to publish `homeassistant`. If you have a few spare cycles to help on this it would be much appreciated. Honestly I tried all permutations I could think of but always getting stuck with build issues only on `x64`.

You already found that the build of PyNaCL does not find the libffi.so file. It might be related to cross compilation, as for x64 there is no cross compilation required.

It is not supported to build python-whees in the cross folder. For further analysis of building python-wheels with minimal dependencies create a diyspk for pynacl like this:

```
SPK_NAME = pynacl
SPK_VERS = 1
SPK_REV = 1

BUILD_DEPENDS = cross/python38
DEPENDS = cross/PyNaCl

MAINTAINER = test
DESCRIPTION = 
DISPLAY_NAME = 
STARTABLE = no

HOMEPAGE   = 
LICENSE    = 
HELP_URL   = 

```

With diyspk makefile as above, the build of PyNaCl wheel succeeds. So it might be an issue with other dependencies in the homeassistant spk.


BTW _it would be great if we could reuse an already built python38 for BUILD_DEPENDS and/or DEPENDS on cross/python38 similar to ffmpeg in dependent packages (tvheadend, ...)_

cross/python38/Makefile
```diff
@@ -8,26 +8,57 @@ PKG_DIR = Python-$(PKG_VERS)
...
 GNU_CONFIGURE = 1
-CONFIGURE_ARGS  = --enable-shared --enable-ipv6 --without-ensurepip --enable-loadable-sqlite-extensions
-CONFIGURE_ARGS += ac_cv_buggy_getaddrinfo=no ac_cv_file__dev_ptmx=no ac_cv_file__dev_ptc=no ac_cv_have_long_long_format=yes
+
+CONFIGURE_ARGS  = --enable-shared
+CONFIGURE_ARGS += --enable-ipv6
+CONFIGURE_ARGS += --without-ensurepip
+CONFIGURE_ARGS += --enable-loadable-sqlite-extensions
+CONFIGURE_ARGS += --with-computed-gotos=yes
+CONFIGURE_ARGS += --enable-optimizations
```
hgy59, line 26: with optimizations it runs a serie of tests (some fail) and needs more resources to build.

Can we avoid this option for BUILD_DEPENDS and DEPENDS for other packages?
Only the build of spk/python38 should enable optimizations to spare build resources for other builds.

spk/python38/Makefile
```diff
@@ -20,7 +20,7 @@ DESCRIPTION_FRE = Langage de programmation Python.
 DESCRIPTION_SPN = Lenguaje de programación Python.
 STARTABLE = no
 DISPLAY_NAME = Python 3.8
-CHANGELOG = "1. Update to Python 3.8.11.<br/>2. Update OpenSSL to v1.1.1l"
+CHANGELOG = "1. Update to Python 3.8.12<br/>2. Update OpenSSL to v1.1.1l<br/>3. Update all requirements.txt"
```
hgy59, line 23: For the view of an end user we updated the wheels contained in the package.
```suggestion
CHANGELOG = "1. Update to Python 3.8.12<br/>2. Update OpenSSL to v1.1.1l<br/>3. Update included wheels"
```

hgy59:
finally removed cross/poetry build dependency.
poetry is a Dependency Management for Python.
It is intended to manage and install dependencies of Python projects,
but it creates it's own virtual environment on installation. This might be the reason for breaking the spksrc cross environment.

I have no clue why it was added to homeassistant some time ago.

cross/python38/Makefile
th0ma7: 1. I'll read a bit on that and confirm if this affect the lto as well.  Any thoughts on keeping one or the later or both?
2. I suggest we do this outside of this PR, perhaps with a focus of migrating all remaining apps to python38.  What do you think?

publicarray:
I think we should do everything possible to avoid the `--use-deprecated=legacy-resolver` flag. From my understanding pip has become stricter and without the flag newer pip exposes errors in existing requirements.txt

https://github.com/SynoCommunity/spksrc/pull/4845#issuecomment-917234727

more detailed: https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020

>Validate your current environment by running `pip check`. This will report if you have any inconsistencies in your set of installed packages

th0ma7:
> The build of cross/poetry tries to build cryptography until is succeeds with the latest version that does not depend on setuptools_rust, and probably has dependencies that conflict with later build of PyNaCl (did not analyze the details).

@hgy59 adding rust cross-compilers would be a really nice addition... along with a newer gcc.  Still on my TODO... someday.

Safihre:
> I think we should do everything possible to avoid the `--use-deprecated=legacy-resolver` flag. From my understanding pip has become stricter and without the flag newer pip exposes errors in existing requirements.txt
> 
> [#4845 (comment)](https://github.com/SynoCommunity/spksrc/pull/4845#issuecomment-917234727)
> 
> more detailed: https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020
> 
> > Validate your current environment by running `pip check`. This will report if you have any inconsistencies in your set of installed packages

What I learned in my effort in building SABnzbd spk is that it might also be a bug in pip. Because there was just no way to reproduce problems locally, even with identical command line parameters. Has to something with the crossenv or something..

Safihre (approved):

th0ma7:
@Safihre did it solve your issue you had?

th0ma7:
Good, merging and will release updated python packages.  Thnx for the review.

PR #4896, opened 2021-10-04
Bazarr: Initial release (0.9.9)

smaarn:
_Motivation:_  Creating initial bazarr spk
_Linked issues:_  Fixes #3535. Supersedes #4629 , #4453, #4547, #4751

Includes:

* New version of `gevent` (21.x)
* New version of `greenlet` (1.x)
* New version of `libev` (4.33 as opposed to current 4.22 used by `ffsync`)
* `zope.event`: 4.5.0
* `gevent-websocket`: 0.10.1

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [ ] New installation of package completed successfully

PR #4751, opened 2021-07-23
Bazarr: Initial release (0.9.6)

smaarn:
_Motivation:_  Creating initial bazarr spk
_Linked issues:_  Fixes #3535. Supersedes #4629 , #4453, #4547

Includes:

* New version of `gevent` (21.x)
* New version of `greenlet` (1.x)
* New version of `libev` (4.33 as opposed to current 4.22 used by `ffsync`)
* `zope.event`: 4.5.0
* `gevent-websocket`: 0.10.1

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4547, opened 2021-04-11
Feat/bazarr/0.9.3

smaarn:
_Motivation:_  Creating initial bazarr spk
_Linked issues:_  Fixes #3535. Supersedes #4453 

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

PR #4453, opened 2021-02-26
bazarr: initial release (0.9.1.1)

smaarn:
_Motivation:_  Creating initial bazarr spk
_Linked issues:_  Fixes #3535.

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4916, opened 2021-10-16
Bazarr 1.0.0

smaarn:
_Motivation:_  Creating initial bazarr spk
_Linked issues:_ #3535. 

Supersedes #4629 , #4453, #4547, #4751, #4896

Includes:

* `gevent-websocket`: 0.10.1
* `zope.event`: 4.5.0
* `webrtcvad`: 2.0.10

### Checklist
- [x] Package upgrade completed successfully (DSM 7.0 x64)
- [x] New installation of package completed successfully (DSM 7.0 x64)

### Binaries

Binaries can be found here: ~~https://github.com/smaarn/spksrc/releases/tag/bazarr-1.0.0-rc1~~ https://github.com/smaarn/spksrc/releases/tag/bazarr-1.0.0-rc2


spk/bazarr/src/service-setup.sh
```diff
@@ -0,0 +1,33 @@
...
+SVC_WRITE_PID=y
+SVC_CWD="${SYNOPKG_PKGDEST}/share/${PACKAGE}"
+
+export SYNOPKG_PKGDEST
+export SYNOPKG_PKGVAR
+
+service_postinst ()
+{
+
+    # Create a Python virtualenv
```
ymartin59, line 20: I guess this Python 3 setup has been copy-paste in many packages. It may be a good idea to collect it as a common function that package can just invoke. CC @hgy59

ymartin59 (approved):
LGTM. Thanks for this contribution

publicarray:
@smaarn While testing I found a small error in the installation log, doesn't seem to impact the package though. While fixing that I made some changes that make the package similar to how other python package install themselves and removed the bazzar-daemon script. The only behaviour change is that `source "${SYNOPKG_PKGVAR}/profile"` is not called anymore. But why was that in there anyway? As far as I can tell the file never existed. Please let me know what you think. Otherwise, this is good to merge.

smaarn:
> @smaarn While testing I found a small error in the installation log, doesn't seem to impact the package though. While fixing that I made some changes that make the package similar to how other python package install themselves and removed the bazzar-daemon script. The only behaviour change is that `source "${SYNOPKG_PKGVAR}/profile"` is not called anymore. But why was that in there anyway? As far as I can tell the file never existed. Please let me know what you think. Otherwise, this is good to merge.

Actually the profile logic was there for "advanced users" who wanted to customize environment variables. The main use case was for changing the encoding to use (in file names). The defaults enforced by Synology (in DSM 6 at least) wasn't systematically UTF8, so I had enforced a defaults to be UTF8 and added a profile mechanism.

publicarray:
@smaarn Thanks for the insight! I'm searched for other packages that do this and found that sonarr/sabnzbd has something similar for sometime (I don't recall anyone creating an issue about it). Other packages have the utf-8 wrapper removed. I think utf8 is a sane default, so I'm happy to leave it. I doubt anyone is going to change it. If someone opens an issue about it, we can always add it in later.

smaarn:
> @smaarn Thanks for the insight! I'm searched for other packages that do this and found that sonarr/sabnzbd has something similar for sometime (I don't recall anyone creating an issue about it). Other packages have the utf-8 wrapper removed. I think utf8 is a sane default, so I'm happy to leave it. I doubt anyone is going to change it. If someone opens an issue about it, we can always add it in later.

bazarr is forking its own subprocess. Environment variables may need being exported to be sure they're propagated, shouldn't they ?

publicarray:
Hm, Isn't using `env` enough? The subprocess should inherit from the parent. Is there a quick way to test this?

spk/bazarr/src/service-setup.sh
```diff
@@ -0,0 +1,32 @@
+FFMPEG_DIR="/var/packages/ffmpeg/target"
+PYTHON_DIR="/var/packages/python38/target"
+PATH="${SYNOPKG_PKGDEST}/bin:${SYNOPKG_PKGDEST}/env/bin:${PYTHON_DIR}/bin:${FFMPEG_DIR}/bin:${PATH}"
+PYTHON="${SYNOPKG_PKGDEST}/env/bin/python3"
+VIRTUALENV="${PYTHON_DIR}/bin/python3 -m venv"
+PIP=${SYNOPKG_PKGDEST}/env/bin/pip
+LANGUAGE="env LANG=en_US.UTF-8 LC_ALL=en_US.utf8"
```
smaarn, line 7: How about

```suggestion
LANGUAGE="export LANG=en_US.UTF-8 LC_ALL=en_US.utf8 && "
```

smaarn:
> Hm, Isn't using `env` enough? The subprocess should inherit from the parent. Is there a quick way to test this?

`env` is supposed to work only for the current environment not for forked commands, `export` guarantees that it's propagated to forked processes. But python may actually be circumvening this issue since bazarr is using `POpen` which is supposed to inherit it.

The quickest way would be to try retrieving some series / movies using French or Spanish custom symbols...

spk/bazarr/src/service-setup.sh
smaarn: The python doc actually seems to indicate this change should be useless.

publicarray:
Thanks. I tried a French movie and noticed no difference in behaviour.

Issue #3707, opened 2019-05-17
TvHeadend Login Fail on Synology


bigwamba:
_For new Package Requests, see the guidelines_  

### Setup
_Package Name:_ TvHeadend
_Package Version:_ 4.2.8-16

_NAS Model:_ Synology DS118
_NAS Architecture:_ 
_DSM version:_ 6.2.2-24922

### Expected behavior 
Should be able to login with web gui to reach control settings.

### Actual behavior
Login page never accepts any username/pw.

### Steps to reproduce
_1._ Attempt login to web gui from TvHeadend app in Synology DSM.
_2._ 
_3._ 

#### Package log
_Check Package Center or `/usr/local/{package}/var/`_
I'm unsure how to access relevant Synology logs.  Nothing seems to work.  Have been trying for days with Linux and OSX terminal apps.  Have tried many possible solutions from older posts e.g. reinstalling, thorough check of permissions etc.

Insert the package log here
I don't know where to find the TvHeadend package log.  Only very general system logs are present.

#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
Insert log here
```

bigwamba:
Sorry if I messed up anything on this post.  It's my first one on Github.  Help greatly appreciated.

m4tt075:
@bigwamba If completely uninstalling TVH and reinstalling it again does not help, you'll need the logs. Use `ssh` to log into your NAS, e.g. via `putty` and you will find them...

PR #4921, opened 2021-10-17
tvheadend Fall 2021 update & Fix 403 Forbidden

th0ma7:
_Motivation:_  Update `tvheadend` to latest available version and:
1. blabla
_Linked issues:_  #4758 and #3707

### Checklist


mk/spksrc.wheel.mk
```diff
@@ -44,34 +44,32 @@ pre_wheel_target: wheel_msg_target
...
-		fi ; \
+			fi ; \
+		done \
 	fi
 
 build_wheel_target: $(PRE_WHEEL_TARGET)
 	@if [ ! -z "$(WHEELS)" ] ; then \
 		$(foreach e,$(shell cat $(WORK_DIR)/python-cc.mk),$(eval $(e))) \
-		if [ ! -z "$(CROSS_COMPILE_WHEELS)" ] ; then \
+		if [ -f "$(WORK_DIR)/wheelhouse/cross-requirements.txt" ]; then \
```
Safihre, line 62: So what happens here? How does the rename from `requirements` to `cross-requirements` fixes things?

spk/python38/Makefile
```diff
@@ -12,7 +12,7 @@ DEPENDS += cross/lxml cross/pycrypto cross/pycurl cross/pyyaml
 DEPENDS += cross/msgpack-python cross/ruamel.yaml cross/immutables
 DEPENDS += cross/cryptography
 
-WHEELS = src/requirements.txt
+WHEELS = src/requirements-cross.txt src/requirements.txt
```
hgy59, line 15: Until we have a solution for how to detect whether a wheel is pure python before we built it, we could use dedicated variables to separate the requirements files.

```suggestion
WHEELS_PURE_PYTHON = src/requirements.txt
WHEELS_CROSS_COMPILE = src/requirements-cross.txt
```

hgy59: @th0ma7 explained this in https://github.com/SynoCommunity/spksrc/issues/4925#issuecomment-946304515

mk/spksrc.wheel.mk
th0ma7: its not only a rename but rather telling to cross-compile the ones in `cross` and "python" compile the other ones by default (e.g. `pure`).  Again for reference https://github.com/SynoCommunity/spksrc/issues/4925#issuecomment-946304515

th0ma7:
@hgy59 while playing with this now I see what the value of cross-linking python apps to `python38` like `ffmpeg` makes sense... that would reduce the CPU time overhead dramastically....

PR #4923, opened 2021-10-17
tt-rss: upgrade to 20210902 + migrate to maria DB 10 + add support for DSM 7 

smaarn:
_Motivation:_  Being able to operate under DSM 7 required Maria DB upgrade
_Linked issues:_  #4524 

Supersedes #4840 

Bump required PHP version to 7.2
Prefer using resource mechanism to inject db
Switch to Maria DB 10 and rely on its automatic Maria DB 5 => Maria DB 10 migration process
Fix initial installation phase which should trigger the schema update

### Checklist

### Tested scenarii

- [x] Upgrade DSM 6 => DSM 7
- [x] DSM 7 Upgrade as is (e.g. reinstalling the package)
- [x] DSM 7 Installation from scratch

Test packages can be found here: https://github.com/smaarn/spksrc/releases/tag/tt-rss_20210902


ymartin59:
@smaarn Great job.
I have a "personal" request. I would like to avoid to run MariaDB when PostgreSQL (included in DSM) is available for an application.
Would it be possible to select our database (mariadb or embedded postgresql) in setup wizard? CC @hgy59 @publicarray

smaarn:
That could be an option. There are several things to consider on which feedbacks would be much appreciated:
1. Is there a real officially supported pgsql version I should be considering ?
2. Should I try making the mysql dependency optional ?
3. Should I try performing migration automatically or should this be done only on new installation ?

Adding support for pgsql is soon going to be mandatory anyway since it's only being supported for maintenance and bugfixes.

ymartin59 (approved):

ewoks:
just a friendly poke for visibility here... As I see changes are approved, but there is no merge yet. Any specific reason for it or we can move forward and add few checkboxes in the table next to tt-rss [here](https://github.com/SynoCommunity/spksrc/issues/4524) ?

hgy59:
@smaarn you need to rebase to  master to fix the `dsmappname` in the INFO file. 
This will bring back the tt-rss link to the DSM Applications fixed with #4993.

spk/tt-rss/src/conf/resource
```diff
@@ -1,14 +1,79 @@
...
+    "drop-db-uninst": true,
+    "drop-user-uninst": true
+  },
+  "webservice": {
+    "services": [{
+      "service": "tt-rss",
+      "display_name": "Tiny Tiny RSS",
+      "type": "apache_php",
+      "root": "tt-rss",
+      "backend": 1,
```
hgy59, line 39: can we use apache 2.4 instead of 2.2?
```suggestion
      "backend": 2,
```

smaarn: If we go that path it means having different requirements depending on whether DSM 6 or not (currently it's only depending on the 2.2 version).

If that's an accepted approach I could do that.

smaarn: Wait, actually apache 2.2 is not even listed in the dependencies... That wouldn't be changing anything.

smaarn:
Flagged this PR as draft until I complete testing on the newer version which is
* Using apache 2.4
* Rebased

smaarn:
Latest binaries have been published here https://github.com/smaarn/spksrc/releases/tag/tt-rss_20210902_rc2

smaarn:
> @smaarn Great job.
> I have a "personal" request. I would like to avoid to run MariaDB when PostgreSQL (included in DSM) is available for an application.
> Would it be possible to select our database (mariadb or embedded postgresql) in setup wizard? CC @hgy59 @publicarray

Database selection could totally be possible in setup wizard **but** there is a mechanism which may prove to be awkward is the dependency selection mechanism which is not really working here.

TBH I wonder if, given the complexity, it wouldn't be simplier to 
1. have two packages (since now we're working on a "no-arch" package), one dedicated for mysql and one for pgsql, both replacing the former and conflicting with each other.
2. deprecate the -mysql one (if that's even possible)

I will try to have this done next year if you don't mind and would rather have this taken offside from this PR. WDYT ?

smaarn:
> @smaarn you need to rebase to master to fix the `dsmappname` in the INFO file.
> This will bring back the tt-rss link to the DSM Applications fixed with #4993.

@hgy59 just tested and it's indeed back :)

smaarn:
Out of curiosity, is there anything specific blocking this PR ? If yes, is there anything I can do to help ?

The postgresql project is indeed a topic but shouldn't it be considered separately ?

smaarn:
Rebased at head

smaarn:
Had to rebase at head since there were conflicts.

smaarn:
Rebased @ head to make sure to keep up with changes on the master branch.

smaarn:
Rebased @ head to handle conflict in merge.

Any objection to my merging ? (Knowing that obviously "just merging" won't be sufficient but at least it's a first towards releasing :) )

I am currently working on making the postgresql adaptation but I don't believe it would be relevant to mix it with this MR (which already changes a few things...)

spk/tt-rss/src/service-setup.sh
```diff
@@ -0,0 +1,169 @@
+#!/bin/sh
```
hgy59, line 1: this file is included into other scripts files and shebang is not required.
```suggestion
```

smaarn:
Rebased at head before merge by EOD (CET time).

hgy59:
@smaarn is this ready to be published?
_(sorry I was out for vacation from mid september to mid october)_

smaarn:
@hgy59 yed it should be ready to publish. Thanks a lot !

hgy59:
@smaarn to create the noarch package for SRM 1.1 I had to temporary deactivate the `REQUIRED_MIN_DSM = 6.0`

PR #4930, opened 2021-10-22
framework, syncthing: Implement DSM managed certificate for services (fixes #3436)

### LINKED ISSUE #3436: Syncthing: Use system certificates for Web GUI
_Package Name:_ syncthing
_Package Version:_ 0.14.49-15

### Expected behavior
Syncthing should show up under _Control Panel - Security - Certificates - Configure_ as an application whose server certificate can be configured.  This should especially allow to use the "system default" certificate that the DSM GUI uses as well.

At least point it at the "system default certificate" at `/usr/syno/etc/certificate/system/default/cert.pem` instead of the self-signed certificate that is generated upon Syncthing first start.

### Actual behavior
The Syncthing GUI uses its own self-signed SSL certificate. 

This has previously been discussed at https://github.com/SynoCommunity/spksrc/pull/3391#issuecomment-406567682 and https://github.com/SynoCommunity/spksrc/pull/3395#issuecomment-407229308.  Probably needs additional input / documentation from Synology on the service resource configuration.

Maybe the changes in https://github.com/SynoCommunity/spksrc/pull/3431 can give some inspiration, as GateOne seems to use the system default certificate after that PR.  Such an approach (copying the certificate and key) does not survive a renewal, though and probably has horrible security implications (copy of the private key in application directory).

acolomb:
_Motivation:_ DSM provides a way to manage TLS certificates for different applications through Control Panel --> Security --> Certificates.  In order for a package service to show up there, it must declare a resource named `certificate-config` pointing to a service definition.  This was reverse-engineered from Synology's WebDAVServer package and ported to our Syncthing package.  It allows to assign a TLS certificate for the Syncthing Web GUI, or request one from Let's Encrypt, which is then kept up to date on renewals.  The Syncthing process is automatically restarted to apply changes.

**The framework changes lay the foundation to integrate other packages with the certificate management UI.**

Also contains a small fix for the `privilege` file and `$SPK_NAME.sc` being created with executable permissions in the package.

_Linked issues:_ fixes #3436 

### To-Do / Further Discussion Needed

* The `ca_reloader.sh` script is based on the one from WebDAVServer, but the mechanism used there for restarting does not work with our `start-stop-status` script (on DSM6 at least).  Instead we use Syncthing's REST API for restarting.
[x] Because of that, the script is specific for Syncthing.  A more general implementation e. g. using `synopkg restart` could be considered, but ~may~ will not work because the `ca_reloader` script runs as the package service user (`sc-syncthing`).
* Installation of the `ca_reloader` script is currently handled by an extra install target, which is already in use for Syncthing.  If other packages start making use of this, copying should better be done in the framework,
* [x] ~The script must be tidied up, there is even a separate logging mechanism in place to debug what it does.~  I could not find where the `stdout` / `stderr` logs go during its execution.

* Syncthing could be configured to force HTTPS usage for its Web UI.  It depends on whether a DSM installation can ever run without any certificate whatsoever.  I think Synology always generates a self-signed cert for synology.com that is the system default.

* The field `display_name_i18n` is required for this resource configuration, unlike for example in DSM7's Web Service resource.  As there is no documentation in the Developer Guide about this, we just copy the regular `display_name` attribute, which seems to work so far.

* [x] ~Syncthing upstream has a new version available which should be included as well.  Not done yet because of unrelated build failure.~

* **Not tested on DSM7 at all.**

### Checklist
- [ ] Build rule `all-supported` completed successfully. **Only for the `spk/syncthing` package, haven't checked if the framework changes break anything else.** The `qoriq` architecture is failing, reported as unsupported.  That's not caused by this PR though.
- [x] Package upgrade completed successfully.
- [x] New installation of package completed successfully.
- [x] Verified that Syncthing is automatically restarted and uses the correct certificate after Let's Encrypt renewal or choosing a different certificate.


spk/syncthing/src/service-setup.sh
```diff
@@ -13,6 +13,13 @@ service_prestart ()
         SERVICE_COMMAND="${SERVICE_COMMAND} ${SYNCTHING_OPTIONS}"
     fi
 
+    # If the system has a TLS certificate for us, force its usage
+    cert_dir=/usr/local/etc/certificate/${SYNOPKG_PKGNAME}/${SERVICE_CERT}
+    if [ -f ${cert_dir}/cert.pem -a -f ${cert_dir}/privkey.pem ]; then
+    	ln -sf ${cert_dir}/cert.pem ${SYNOPKG_PKGVAR}/https-cert.pem
```
ymartin59, line 19: Same indentation request here - keep them consistent in the whole script

acolomb: Yea, sorry.  My editor is configured to use one tab instead of eight spaces when possible.  Fixed now.

acolomb:
I've just pushed some improvements and fixed the conflicts.  Would rather do a rebase onto master though, so we can keep the separate (cleaned up) commits.  Hang on a bit...

acolomb:
Syncthing upstream v1.18.4 was released today, so switched to that newer version.  Rebased to keep a clean history.

mk/spksrc.service.mk
```diff
@@ -319,11 +334,11 @@ endif
...
 ifeq ($(strip $(SERVICE_EXE)),)
 $(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege-installasroot
-	@$(dsm_script_copy)
+	@$(dsm_resource_copy)
 	@$(MSG) "(privilege) spksrc.service.privilege-installasroot"
 else
 $(DSM_CONF_DIR)/privilege: $(SPKSRC_MK)spksrc.service.privilege-startasroot
-	@$(dsm_script_copy)
+	@$(dsm_resource_copy)
 	@$(MSG) "(privilege) spksrc.service.privilege-startasroot"
```
publicarray, line 342: My guess is that this breaks the scripts for DSM 6

publicarray:
Unfortunately this is using an "undocumented" worker and is blocked by Synology on DSM7: `non-synology-signed package should not use internal worker`
Screenshot from 2021-11-11 22-34-30

publicarray:
While interesting, the official way to get HTTPS is to use their proxy/login portal/or whatever it's called on DSM7.

acolomb:
Wow, that is very unfortunate.  My ticket with Synology regarding the Developer Guide missing this information is still lingering in their "development", but I haven't heard back in a few weeks now.

publicarray:
Yea :cry: 
 Did you try the login portal? I think it provides similar functionality. An automated way would have been nicer.

acolomb:
No, I haven't investigated that at all yet.  Only seen that there is some mention of reverse proxies in the Guide and I think that might be an alternative approach.

Do you think we could still merge this and make it conditional on DSM<7?  I know several people would be very happy with this solution and I will continue trying to convince Synology about opening it up.

publicarray:
I'm honestly not sure. I think I like to rely on official support when we can. So if the alternate method with the "login portal" doesn't work than I would be more open to it. Maybe some of the other devs have a stronger opinion @th0ma7, @ymartin59  or @hgy59  might have something to say about it.

ymartin59 (approved):

ymartin59:
@publicarray I am OK with merge-and-publish a package update even if still restricted to DSM 6.2

mk/spksrc.service.mk
```diff
@@ -229,6 +238,12 @@ ifeq ($(strip $(USE_DATA_SHARE_WORKER)),yes)
 		'."data-share" = {"shares": [{"name": $$share, "permission":{"rw":[$$user]}} ] }' $@ | sponge $@
 endif
 endif
+ifneq ($(strip $(SERVICE_CERT)),)
+	@jq --arg service "${SERVICE_CERT}" \
+	    --arg reload "${SERVICE_CERT_RELOAD}" \
+	    --arg display_name "${DISPLAY_NAME}" \
+		'."certificate-config" = {"services": [{"display_name": $$display_name, "display_name_i18n": $$display_name, "service":$$service} ], "reloader-relpath": $$reload } | ."service-cfg" = {}' $@ | sponge $@
+endif
```
publicarray, line 246: In that case, this just needs be excluded from DSM7.

publicarray (changes requested):

mk/spksrc.service.mk
acolomb: Done in latest commit.

acolomb:
I'm still in contact with Synology regarding opening up that API for community packages.  Got as far as a "Junior Product Manager" from the German branch saying that the feature request was well received within the development team and they'd like to work with me on making it available.  Since then only more running around in circles between different inquiry and support tickets :roll_eyes:

mk/spksrc.service.mk
publicarray: Thanks

Just an FYI the /usr/local paths are not used anymore since DSM7  (DSM6.2 also uses some of it) as per the HFS: https://github.com/SynoCommunity/spksrc/wiki/DSM-7.0 (table at the bottom)

publicarray (approved):

publicarray:
Yea i know exactly what your going though..

mk/spksrc.service.mk
```diff
@@ -198,12 +205,14 @@ endif
 # - firewall rules/port definitions (DSM >= 6.0-5936)
 # - data share worker (DSM 7, optional for DSM 6)
 # - usr local links (DSM >= 6.0-5941)
+# - certificate config (DSM < 7, available only for Synology packages since DSM 7)
```
hgy59, line 208:
```suggestion
# - certificate config (DSM >= 7, available only for Synology packages since DSM 7)
```

mk/spksrc.service.mk
```diff
@@ -205,7 +205,7 @@ endif
 # - firewall rules/port definitions (DSM >= 6.0-5936)
 # - data share worker (DSM 7, optional for DSM 6)
 # - usr local links (DSM >= 6.0-5941)
-# - certificate config
+# - certificate config (DSM < 7, available only for Synology packages since DSM 7)
```
hgy59, line 208:
```
# - certificate config (DSM < 7, available only for Synology packages before DSM 7)
```

spk/syncthing/src/syncthing.sc
```diff
@@ -1,4 +1,4 @@
-[syncthing]
+[syncthing-bep]
```
hgy59, line 1: Does this work? Did you try to uninstall syncthing and the port reservation was removed?

Regarding the dev guide this must be `syncthing_bep`
_If your package needs more than one port description, you can define service_name2, service_name3, …_

mk/spksrc.service.mk
acolomb: That's not correct though.  We support it only for DSM 6, since the function is not available for community packages on DSM 7.

hgy59: I know, I know. It's more a semantic mismatch, as "since DSM 7" means "DSM >= 7".

mk/spksrc.service.mk
acolomb: So is that compatible with DSM 6 as well?  On my DS418j running latest DSM 6.2.something, the `/var/packages/syncthing/var` link does not exist.  The one under `target/var` does however.

What should I use for maximum compatibility in the future?  Note that the `ca_reloader.sh` script does not receive any of the `SYNOPKG_PKG*` variables unfortunately.  At least under DSM 6.

spk/syncthing/src/syncthing.sc
acolomb: There is a contradiction in the Developer Guide itself.  All examples use the underscore (`_`), whereas the description only allows `“a~z”, “A~Z”, “0~9”, “-”, “\”, “.”` (no underscore, only dash).

hgy59: The examples in the dev guide are correct. Only the service name does not allow `_`.

In the first example `[ss_findhostd_port]` is the service name `ss`
in the second example `[dhcp_udp]` is the service name `dhcp`.

acolomb: > Does this work? Did you try to uninstall syncthing and the port reservation was removed?

I just tried it on DSM 6.  During a package upgrade, the firewall rule stays unmodified and still displays the right description after the `[syncthing]` to `[syncthing-bep]` rename.  Subsequently uninstalling the package (without touching firewall rules) does remove the entry automatically.

acolomb: > The examples in the dev guide are correct. Only the service name does not allow `_`.

Sorry I don't understand at all.  AFAIU the whole "INI section" name within brackets is the service name.  I don't read anything in there about the first part before an underscore being somehow special.  It just says:

    service_name cannot be empty and can only include characters “a~z”, “A~Z”, “0~9”, “-”, “\”, “.”

`service_name` is the placeholder / "variable name", in our case `syncthing-bep`.  The value does not allow underscores, but of course they can use whatever placeholder string in the docs.  However the examples contradict their own rules.

I've played around with this a bit while testing the certificate implementation.  Certainly using a dash was not worse than an underscore, but I can't remember if there were any problems with using an underscore.  In the end I decided to take the value description as reference, disregarding their examples.

hgy59: IMHO the service name here is `syncthing` and not `syncthing-bep` as defined in `SERVICE_COMMAND`.

If you need multiple sections in *.sc files you can use servicename + `_` + postfix to create different sections with unique names. And this is optional if you want different descriptions for different ports of your service.

Sorry for not posting the first part of the dev description before:
```
Usually a package only has one unique service
name. If your package needs more than one port
description, you can define service_name2,
service_name3, …
```

It is not easy to write good documentation and synology badly uses `service_name` for the section.
It would make more sense with a better naming like `[service_postfix]` or `[service-name_postfix]`.

Summary:
The current syncthing package has the sections `[syncthing]`, `[syncthing-discovery]` and `[syncthing-webui]`.
This should be adjusted to `[syncthing_bep]`, `[syncthing_discovery]` and `[syncthing_webui]`.

acolomb: I still think using the underscore directly contradicts the rules given in the Developer Guide.  Apparently I'm not alone either, the `shairport-sync` spksrc package also uses a dash here.  While `rabbitmq` and `ffsync` use the underscore.

mk/spksrc.service.mk
acolomb: I think you're just missing the emphasis here.   It's meant to be:

> available **only** for **Synology** packages since DSM 7

and NOT

> available *only* for Synology **packages since DSM 7**

How could that be made clearer?

hgy59:
#4995 is merged into master, ready to rebase here...

spk/syncthing/src/service-setup.sh
```diff
@@ -13,6 +13,13 @@ service_prestart ()
         SERVICE_COMMAND="${SERVICE_COMMAND} ${SYNCTHING_OPTIONS}"
     fi
 
+    # If the system has a TLS certificate for us, force its usage
+    cert_dir=/usr/local/etc/certificate/${SYNOPKG_PKGNAME}/${SERVICE_CERT}
+    if [ -f ${cert_dir}/cert.pem -a -f ${cert_dir}/privkey.pem ]; then
+        ln -sf ${cert_dir}/cert.pem ${SYNOPKG_PKGVAR}/https-cert.pem
+        ln -sf ${cert_dir}/privkey.pem ${SYNOPKG_PKGVAR}/https-key.pem
+    fi
+
```
hgy59, line 22: Can we make this optional, i.e. activate with an option in the install wizard?

When using a web proxy with dyndns not hosted on the diskstation we don't want to use such certificates for TLS.
Probably a Checkbox in the Wizard like "[ ] Activate TLS with DSM integrated Certificate".
This would also allow to abort here with an error message, when certificate is not available.

acolomb: I like that idea, but not sure if it will work in all cases.  The `ca_reloader` script will be called whenever the *configured* certificate changes AFAIU.  By default, the "system" certificate is assigned, meaning the synology.com-derived certificate unless the user changed it.  Now if that one is automatically renewed by DSM, the ca_reloader script will replace Syncthing's self-signed certificate, regardless of whether the option was checked during installation.

There's no easy way to store such wizard settings permanently for later use, or is there?

Could you elaborate on the case where you think relying on Syncthing's own self-signed cert is the only solution?

spk/syncthing/src/syncthing.sc
acolomb: @hgy59 I'm still leaning toward the version with dashes, the more often I re-read Synology's docs on the matter.  Can you please reconsider my argumentation?

There are some other unresolved review comments.  Could you please look at those again as well?

acolomb:
> #4995 is merged into master, ready to rebase here...

I merged in the master branch for now.  Will rebase when all outstanding review comments are addressed.

Can we keep the SPK revision 23, or are you going to publish that one as is and this should become revision 24?

mk/spksrc.service.mk
acolomb: @publicarray I just added a check that will dynamically switch from `/var/packages/syncthing/var` to `/var/packages/syncthing/target/var` when necessary.  The former does not exist on my DSM 6.2 installation, and I've read several reports now that the latter does not work properly under DSM 7.

mk/spksrc.service.mk
acolomb: @hgy59 Re-phrased in latest commit, hope that's clearer now.

publicarray:
> One more thing: I noticed @Safihre removed himself from Syncthing's MAINTAINER field, replacing with generic SynoCommunity in f58b46c. Is that preferred or should I add myself back there?

You are welcome to add yourself back. I think that he might have removed you by mistake.

Safihre:
> > One more thing: I noticed @Safihre removed himself from Syncthing's MAINTAINER field, replacing with generic SynoCommunity in [f58b46c](https://github.com/SynoCommunity/spksrc/commit/f58b46cb66383286cf43c625f3bc17bffa8a8c2c). Is that preferred or should I add myself back there?
> 
> You are welcome to add yourself back. I think that he might have removed you by mistake.

@acolomb Sorry about that!! I just removed myself everywhere because I kept getting people asking me to fix packages I didn't touch for many years.

spk/syncthing/src/ca_reloader.sh
```diff
@@ -0,0 +1,28 @@
...
+    syncthing_webui)
+        # Forcefully overwrite certificate files by symlinks
+        ln -sf $CERT_DIR/cert.pem $CONF_DIR/https-cert.pem
+        ln -sf $CERT_DIR/privkey.pem $CONF_DIR/https-key.pem
+
+        # Required: set $HOME environment variable
+        HOME=$CONF_DIR
+        export HOME
+
+        $SYNCTHING cli --home=$CONF_DIR operations restart
```
Safihre, line 20: Have you tried `synopkg restart`? Just to be sure?

Safihre:
> While interesting, the official way to get HTTPS is to use their reverse proxy/login portal/or whatever it's called on DSM7.

@publicarray Are you sure about this? Isn't this the exact same as the DSM6 way? Looks the same to me on DSM7 so why shouldn't it work there?
@acolomb If you remove the DSM7 check, I can test it for you.

acolomb:
@Safihre that was already tested by @publicarray, see one of the comments above. It's blocked deliberately by Synology for foreign packages, the error message is quite clear on that.

spk/syncthing/src/ca_reloader.sh
acolomb: Yes, it works when run as `root`, but fails if run as user `sc-syncthing`.  As presumed in the original PR description.

Safihre: Right, maybe good to mark this off in the original PR description? I was using that as a guide of the changes and consideration in this PR.

acolomb: Done.

spk/syncthing/src/service-setup.sh
acolomb: Ping @hgy59.  Please revisit all open review comments here, there are some remaining questions.

spk/syncthing/Makefile
```diff
@@ -8,11 +8,11 @@ DEPENDS = cross/$(SPK_NAME)
 
 UNSUPPORTED_ARCHS = $(PPC_ARCHS)
 
-MAINTAINER = SynoCommunity
+MAINTAINER = acolomb
 DESCRIPTION = Automatically sync files via secure, distributed technology.
 DESCRIPTION_FRE = Synchronisation automatique de fichiers via une technologie sécurisée et distribuée.
 DISPLAY_NAME = Syncthing
-CHANGELOG = "Update syncthing to v1.18.5.  Add wizard page to preconfigure credentials for the Web GUI."
+CHANGELOG = "Update syncthing to v1.18.5.  Add wizard page to preconfigure credentials for the Web GUI.  Integrate with the certificate management UI in Control Panel."
```
miry, line 15: I beleive the package version should be bumped: https://github.com/SynoCommunity/spksrc/pull/4930/files#diff-27e7feeb4837853b261e74b395eb80e9e11bc0b1ab5a4f9425985437f4e515b2R3

acolomb: The revision was already bumped in #4995, but the package has not been published since then.  So I think we'll just keep piling up new features under the same `SPK_REV` until they are actually made available?

What's up with the link?  It only takes me to the diff view?

miry: > What's up with the link? It only takes me to the diff view?

It seems because I pointed to hidden line. My bad. https://github.com/SynoCommunity/spksrc/blob/b92382700ea70d436740523865e3b434e5aa9bad/spk/syncthing/Makefile#L3

acolomb:
Ping @publicarray @hgy59 @ymartin59 @th0ma7, anybody able to merge and publish this?  It's been sitting here ready for months now.  If there is anything else I should change / test, please say so.

publicarray:
Thanks, @acolomb, Sorry I totally missed this.

acolomb:
Thanks for merging @publicarray.  What's the prospect on publishing?  Can you do that as well?

BPtLNfxZWo:
thanks for your effort and progress. Any indication when this will be available in DSM 7?

acolomb:
Unfortunately no. Last reply was that the development team has the requirement on their radar and it will not get lost. But they can't give a timeline when or if it will be opened up. At least they seem to understand how it could be useful for third party developers.

BPtLNfxZWo:
Thanks for the update and transparency

airart320:
Any update on DSM7 compatibility?

acolomb:
No response from Synology whatsoever. The current DSM 7.0 Package Developer Guide still doesn't mention the certificate resource. As a possible alternative, it has a mechanism for packages to set up reverse proxies for some ports, but I don't know if that's enough for Syncthing's various protocols.

PR #4936, opened 2021-10-25
Fix minio share and inst variables

### LINKED ISSUE #4933: Minio install breaks when providing a password with special characters
### Setup
_Package Name:_ minio
_Package Version:_ updating from 2020-11-25T22-36-25Z-8 to 2021-09-18-9

_NAS Model:_ DS918+
_NAS Architecture:_ ?
_DSM version:_ 6.2

### Expected behavior
Wizard guides through package update steps prompts for root user and password. The password field should also support special characters or prevent you from entering them.

### Actual behavior
Wizard takes the input, but update fails. Because it stores the submitted values on disk, retrying the update / performing a repair fails and you are not prompted to change the values.

Default minio root user and pass are safe, but as I forgot what those were, I tried another password I frequently use on my syno and that contains some special characters.

### Steps to reproduce
_1._ perform and update of minio package (perhaps first install also triggers it)
_2._ when prompted for minio root user / pass, enter pass with special characters. In my case I expect it was because of '<' or '-'
_3._ update will fail and so will attempts to perform a repair

#### Package log
/var/log/packages/minio.log (i've masked the password)
```
2021/10/25 14:18:14     upgrade minio 2021.09.18-9 Begin preupgrade
/var/packages/minio/etc/installer-variables: line 6: syntax error near unexpected token `newline'
/var/packages/minio/etc/installer-variables: line 6: `WIZARD_ROOT_PASSWORD=***'
2021/10/25 14:18:14     upgrade minio 2021.09.18-9 End preupgrade ret=[2]

```

hgy59:
_Motivation:_  minio package fails to create the shared folder on DSM 6 and special characters cause failure in package
_Linked issues:_  closes #4933, https://github.com/SynoCommunity/spksrc/pull/4908#issuecomment-950217365

### Checklist

### Remarks
- as the packages using sc-download group on shared folders, I reverted the reactivatation of shared folders definition by resources for DSM 6 - after reverted by #4654
- a new Makefile variable `SERVICE_WIZARD_SHARE_RESOURCE` can be used to create shared folders from wizard variables with DSM resource worker definition only.
- add port 9001 for gui (console) as minio has now different ports for API and Console and we don't want dynamic ports that are random at each login.


hgy59:
> just it might be best for DSM6 packages not to use the resource worker for compatibility with existing dsm6 packages.

@publicarray as the minio package is not a download manager and has never used the `sc-download` group for shared folders, I introduced a new Makefile variable `SERVICE_WIZARD_SHARE_RESOURCE` to use the DSM resource linker to create the shared folders from wizard variables with DSM 6 too.

I will use this approach for #4627, when this PR got merged.

mk/spksrc.service.installer.dsm7
```diff
@@ -58,7 +58,12 @@ fi
 
 # Reload wizard variables stored by postinst
 if [ -r "${INST_VARIABLES}" ]; then
-    . "${INST_VARIABLES}"
+    # we cannot source the file to reload the variables, when values have special characters like <, >, ...
```
ymartin59, line 61: Just an idea: avoid copy-paste and collect command source code in a "generic" script - then sourced from specific DSM implementation, to reuse functions and reduce such maintenance effort

mk/spksrc.service.start-stop-status
```diff
@@ -61,9 +58,9 @@ start_daemon ()
                 SU=""
             fi
             if [ -z "${SVC_BACKGROUND}" ]; then
-                $SU ${SERVICE_SHELL} -c "${SVC_CD}${SERVICE_COMMAND}" >> ${OUT} 2>&1
+                $SU /bin/sh -c "${SVC_CD}${SERVICE_COMMAND}" >> ${OUT} 2>&1
```
ymartin59, line 61: The idea here is to be able to use shell specific (dash, bash...) instructions in "SERVICE_COMMAND". But I agree that this feature has never been used right now.

ymartin59 (approved):

mk/spksrc.service.installer.dsm7
hgy59: This is not so easy, as the service-installer files are copied as installer into the staging script folder.
To realize this idea, we would need to call functions for such duplicated code and put those into an additional script file to the package and soure this file into the service-installer as we already do for service-setup.

But I will give a try.

mk/spksrc.service.start-stop-status
hgy59: I first wanted to rename `SERVICE_SHELL` to `SVC_SHELL` for consistency.
But as this is not used anywhere, I decided to remove it.

mk/spksrc.service.installer.dsm7
hgy59: done.

unfortunately the functions are not usable within `service-setup.sh` and `upgrade_uifile.sh`

hgy59:
This is now ready for review and testing.

@fgma packages now support special characters in installer/wizard variables. Can you confirm that it is reasonable to use a static port for the console address?

mk/spksrc.service.mk
```diff
@@ -195,6 +235,12 @@ ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
 		'."data-share" = {"shares": [{"name": $$share, "permission":{"rw":[$$user]}} ] }' $@ | sponge $@
 endif
 endif
+ifneq ($(strip $(SERVICE_WIZARD_SHARE_RESOURCE)),)
```
publicarray, line 238: Since this does the same thing maybe a boolean (USE_NEW_SHARE_WORKER=true or OLD_SHARE=false) is a better way to deal with this? e.g  `if  TCVERSION >= 7.0 || USE_NEW_SHARE_WORKER`

hgy59: done with USE_DATA_SHARE_WORKER = yes

publicarray: Thanks, that is also a much better name :+1:

fgma:
@hgy59 Yes, this should be OK. I just left it out in the last PR to keep changes minimal.

PR #4919, opened 2021-10-17
update homeassistant to v2021.9.7

hgy59:
_Motivation:_  Update homeassistant
_Linked issues:_  #4905, #4833, #4820

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #4937, opened 2021-10-26
python 3.10.0

th0ma7:
_Motivations:_  
1. Update to newer python version, 
2. eliminate **ALL** usage of cross for wheels  (and therefore legacy wheel building process `spksrc.python-wheel.mk`)
3. and enhance default wheel set.

With this version I eliminated `pycrypto` as deprecated, only builds with legacy build process and since superseeded by `pyopenssl`, `paramiko`, and `cryptography`.

_Linked issues:_  

### Checklist


Safihre:
Can't we just skip 3.9? 3.10 is already out..

spk/python39/Makefile
```diff
@@ -18,9 +18,8 @@ DEPENDS += cross/sqlite
 DEPENDS += cross/curl
 ENV += PYCURL_CURL_CONFIG=$(STAGING_INSTALL_PREFIX)/bin/curl-config
 
-# Required for Pillow
-DEPENDS += cross/freetype cross/libjpeg cross/zlib
-ENV += STAGING_INSTALL_PREFIX=$(STAGING_INSTALL_PREFIX)
+# Patched cross-compiled wheel
+DEPENDS += cross/pillow
 
 WHEELS = src/requirements-cross.txt src/requirements.txt
```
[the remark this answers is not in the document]
th0ma7: Re-reading, now I better understand what you meant.

It happens that the default use-case (e.g. 3 out of 4) wheels are pure-python.  And if not called as such it often leads to `0.0.0` version issue (e.g. chicken or the egg situation).  As the default use-case is mostly pure-python I really much prefer keeping the behavior as it is (further more as packages have all been adjusted).

Still, I've added the ability to rename as needed with a pre-check if vars are defined.

th0ma7:
@hgy59 there's a lot in there, a few starting questions:
1. I would suggest working on a appropriate fix outside of this PR for all python*, what do you think?  We could fix all sub-packages as needed in one shot and come-up with a new/updated implementation of this fix.
2. Also, the good news is that now we know what was wrong and have a way to circumvent the problem.  If I read you correctly what you would like is that we review the implementation of the solution, right?
3. (unrelated) your thought on wetter we should we bump right away to 3.10?  I'm tempted to use "stable" 3.9 until 3.10 has received a few point releases.

I fully understand that the previous behavior was to force cross-compiling.  Although this behavior cannot be used anymore (e.g. `0.0.0` and `no satisfactory version found` error due to newer `pip`).  With that I believe that keeping this "legacy" behavior has no benefit as being broken.  I'm not saying that we could/should not reverse the cross/pure naming, but perhaps it's best to simply move away from it such as you propose:

> ```
> WHEELS_CROSS_COMPILE = src/cross-requirements.txt
> WHEELS_PURE_PYTHON = src/pure-requirements.txt
> ```

Just for file listing readability I would opt with `requirements-cross.txt` and `requirements-pure.txt` as they stay close-by.

> if $(WHEELS_PURE_PYTHON) is not defined you must not define it (line 18)

The current implementation allows passing both requirement files as well as individual wheel (e.g. `example==1.0`).  As most of the wheels are "pure" I opted to add them into the "pure" `wheelhouse` requirement file.  As such, unless we change the behavior, it currently needs to stay defined.  Do we want to keep this option? add notion of pure/cross in the wheel naming to add it the proper requirement file to be processed, food for thoughts I guess.

At the end of the implementation both pure & cross gets merged into the staging directory so I process it at package installation time (and not package gets missed).

> if $(WHEELS_CROSS_COMPILE) is not defined you must assign WHEELS_CROSS_COMPILE = $(WHEELS) (line 21) to keep the former behaviour.

We certainly can.  But should we as we know this behavior creates issues with newer pip (e.g. that legacy implementation is sort of broken).

> the confusing part is, that you use $(WHEELS_PURE_PYTHON) and $(WHEELS_CROSS_COMPILE) as file names in the $(WHEELHOUSE) folder (pre_wheel_target and build_wheel_target below). The requirement files in the wheelhouse should be independent of the files defined by the package Makefile (it is not even a must that those files are created in the wheehouse folder).

I've align to the previous behavior as the `requirements.txt` was already being copied over to the `wheelhouse` where it is read afterwards for pip processing.  I thought I would keep this similar and just add the `requirements-cross.txt` to it.  Indeed it probably should go elsewhere:  I was thinking of a `wheel-build` or `wheelhouse-build` directory or similar where things gets built into so we could track issues afterwards when a build failure occurs.

> the current solution will not work with multiple requirement files

Totally right, excellent point.  If this is deemed needed then (throwing out ideas) the file name could have "pure" or "cross" in it as a match, or have a comment `# PURE` or `# CROSS` or similar... or what if (really just a thought), all wheels in any requirement file (or direct parameter) wouldn't have a `cross:example1==1.0` or `pure:example2==0.666` and from there we simply grep|cut|awk|sed any files or wheel name passed and generate a cross/pure for processing?  With this we could set note prefix=pure as the most usual wheel?

th0ma7:
@hgy59 re-thinking things through, `cross:` | `pure:` prefix option seems like a reasonable idea for a redesign:
- Easy to implement
- Moves away from file name based approach
- No-prefix could still be an option (using something like `WHEEL_DEFAULT_PREFIX=<pure|cross>` variable)

If you share this view, here's a proposed plan:
1. review & merge this PR
2. new PR for the redesign affecting all python based packages
3. new PR to migrate all `python3` and `python38` packages to `python39`

Thoughts?

Safihre:
I feel quite strongly that we shouldn't use 3.9 while 3.10 is already out. Python releases are in beta testing for almost half a year before they are released. So major bugs are not to be expected.
If we take this approach of first 3.9 we are wasting time to transfer all packages there and then we have to it again for 3.10.

hgy59:
just to remember myself...
for homeassistant (#4421 and #4919) I reduced the package size from > 120MB to < 50MB by removing all pure python wheel from the spk.
The requirements.txt contains only wheels to cross compile. The pure pyhton wheels are splitted into two requirements files (postinst_default_config_requirements.txt, postinst_components_requirements.txt) that are used to install the pure python wheels from the index (pypi.org) at installation time. This was also a workaround for the 0.0.0 version problem of some packages.

th0ma7:
> I feel quite strongly that we shouldn't use 3.9 while 3.10 is already out. Python releases are in beta testing for almost half a year before they are released. So major bugs are not to be expected. If we take this approach of first 3.9 we are wasting time to transfer all packages there and then we have to it again for 3.10.

@Safihre I hear you!  I've created a single commit to migrate from 3.9 to 3.10 (in case we want to move back to 3.9).  Can you test it out and confirm it runs ok on your NAS?  I'll test on my `apollolake` later on to confirm all looks good.

th0ma7:
Weird build error on github that I can't reproduce on my end:
```
spk/python310$ ll ../../packages/*
-rw-r--r-- 1 spksrc spksrc 48896000 Nov  1 20:32 ../../packages/python310_88f6281-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 71813120 Nov  1 20:22 ../../packages/python310_aarch64-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 75827200 Nov  1 20:22 ../../packages/python310_aarch64-7.0_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 70615040 Nov  1 20:22 ../../packages/python310_armv7-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 73431040 Nov  1 20:22 ../../packages/python310_armv7-7.0_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 70676480 Nov  1 21:03 ../../packages/python310_comcerto2k-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 69795840 Nov  1 20:22 ../../packages/python310_evansport-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 73533440 Nov  1 20:31 ../../packages/python310_evansport-7.0_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 63303680 Nov  1 20:22 ../../packages/python310_hi3535-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 72396800 Nov  1 21:24 ../../packages/python310_kvmx64-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 76646400 Nov  1 21:47 ../../packages/python310_kvmx64-7.0_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 72263680 Nov  1 20:22 ../../packages/python310_qoriq-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 72396800 Nov  1 20:22 ../../packages/python310_x64-6.1_3.10.0-1.spk
-rw-r--r-- 1 spksrc spksrc 76625920 Nov  1 20:22 ../../packages/python310_x64-7.0_3.10.0-1.spk
```

mk/spksrc.archs.mk
```diff
@@ -28,6 +28,9 @@ PPC_ARCHS = powerpc ppc824x ppc853x ppc854x qoriq
 i686_ARCHS = evansport
 x64_ARCHS = $(GENERIC_x64_ARCH) apollolake avoton braswell broadwell broadwellnk bromolow cedarview denverton dockerx64 geminilake grantley purley kvmx64 v1000 x86 x86_64
 
+32bit_ARCHS = $(ARMv5_ARCHS) $(ARMv7_ARCHS) $(i686_ARCHS) powerpc ppc824x ppc853x ppc854x
+64bit_ARCHS = $(ARMv8_ARCHS) $(x64_ARCHS) qoriq
```
hgy59, line 32: qoriq is ppc32, i.e. 32 bit too.

spk/python310/Makefile
```diff
@@ -44,6 +44,14 @@ PYTHON_LIB_DIR = lib/python$(SPK_VERS_MAJOR_MINOR)
 
 POST_STRIP_TARGET = python310_extra_install
 
+include ../../mk/spksrc.archs.mk
+
+# Enfore 32-bit object-mode on 32-bit platforms
+# as otherwise it fails to build on github-action
+ifneq ($(findstring $(ARCH),$(32bit_ARCHS)),$(ARCH))
+ENV = OBJECT_MODE=32
```
hgy59: It's better not to overwrite the ENV (at least with #4949)
```suggestion
ENV += OBJECT_MODE=32
```

spk/python310/src/requirements.txt
```diff
@@ -0,0 +1,52 @@
+# Basic wheels
+setuptools==58.3.0
+pip==21.3.1
+wheel==0.37.0
+
+# cryptography dependencies
+idna==3.3
+ipaddress==1.0.23
```
Safihre, line 8: This one is now part of the base python library since 3.3 and can be removed.

spk/python310/src/requirements.txt
```diff
@@ -0,0 +1,52 @@
...
+certifi==2021.10.8
+chardet==4.0.0
+click==8.0.3
+distlib==0.3.3
+filelock==3.3.2
+Flask==2.0.2
+future==0.18.2
+html5lib==1.1
+idna==3.3
+importlib-metadata==4.8.1
```
Safihre, line 24: If a en ipaddress listed twice.

spk/python310/src/requirements.txt
```diff
@@ -0,0 +1,52 @@
...
+paramiko==2.8.0
+platformdirs==2.4.0
+pyasn1==0.4.8
+pycparser==2.20
+PyJWT==2.3.0
+pyOpenSSL==21.0.0
+pyparsing==3.0.1
+python-dateutil==2.8.2
+pytz==2021.3
+requests==2.26.0
```
Safihre, line 41: We should pay attention that any packages moving to python310 remove these now included packages.
Thanks for including them :)

Safihre (commented):
Great that the update to 3.10 worked! Will try to test later on my system.

spk/python310/src/requirements.txt
th0ma7: this one meaning `ipaddress` ?

spk/python310/src/requirements.txt
th0ma7: fully agree with you.
there where so many used everywhere that I tried to capture the mostly used ones.

spk/python310/src/requirements.txt
th0ma7: after looking for a bit, indeed `ipaddress` looks like already included.  removing.

spk/python310/src/requirements.txt
th0ma7: removing duplicate as well, thnx for noticing.

spk/python310/Makefile
th0ma7: good catch.  But in the end this is not necessary.  I was trying figure out where the issue came from and that option really didn't work out (and was a shot in the dark).  thnx.

th0ma7:
All changes included, thnx for your review @hgy59 and @Safihre.
I'm pretty pleased with the the look of the code, I'll let github do a last round and merge.
Next:
1. review `cross` vs `pure` part of the framework (I'll need your assistance on this @hgy59)
2. migrate python3 packages to python310
3. publish all new packages (including python310)
4. (optional) review `python3` and `python38` packages to remove all `cross/*` wheels like in python310

cross/pillow/Makefile
```diff
@@ -1,5 +1,5 @@
 PKG_NAME = Pillow
-PKG_VERS = 8.2.0
+PKG_VERS = 8.4.0
```
hgy59, line 2: unfortunately this breaks homeassistant (#4919) as it requires pillow==8.2.0.
IMHO we should include as less wheels as possible in the python* packages.
Or we could separate cross/pillow_py38 to version 8.2.0 for python38.

hgy59: fixed for homeassistant in #4919 by use of dedicated version in `cross/homeassistant.pillow`.

PR #4945, opened 2021-10-31
Mono and Python: Always use latest certs from trust store. Fix LE cert

publicarray:
Python: https://pypi.org/project/certifi/

Fixes: https://discord.com/channels/732558169863225384/732559466721181738/894569903737163836

### DIFF
--- a/spk/beets/src/requirements.txt
+++ b/spk/beets/src/requirements.txt
@@ -11,7 +11,7 @@ munkres==1.1.2
 
 # Plugin dependencies
 beautifulsoup4==4.10.0
-certifi==2019.11.28
+certifi
 chardet==3.0.4
 discogs-client==2.2.2
 langdetect==1.0.7
--- a/spk/borgbackup/src/requirements.txt
+++ b/spk/borgbackup/src/requirements.txt
@@ -1,7 +1,7 @@
 attrs==21.2.0
 #borgbackup==1.1.17
 borgmatic==1.5.18
-certifi==2021.5.30
+certifi
 charset-normalizer==2.0.4
 colorama==0.4.4
 idna==3.2
--- a/spk/duplicity/src/requirements.txt
+++ b/spk/duplicity/src/requirements.txt
@@ -30,7 +30,7 @@
 #duplicity==0.8.19
 
 ##### basic requirements #####
-certifi==2020.12.5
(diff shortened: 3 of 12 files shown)


spk/mono/src/service-setup.sh
```diff
@@ -1,7 +1,7 @@
 
 # mono service setup
 
-service_postinst ()
+service_prestart ()
 {
     # Sync ca certificates
     ${SYNOPKG_PKGDEST}/bin/cert-sync /etc/ssl/certs/ca-certificates.crt
```
publicarray, line 7: I still need to find out if this works with DSM7 as I'm not 100% sure about the permissions

publicarray: A new release would also fix this, but I think this is better in the long run.

publicarray: What was I thinking. mono isn't a daemon.

hgy59:
@publicarray I unsure whether is is worth to remove the explicit version from certifi dependency.

I still prefer to nail those to an explicit version with ==. As we cannot update a package without incrementing the SPK_REV we also can use a specific version of certify in the packages (obviously the current version at publishing date).
Without this, it is not possible to reproduce a specific spk version.

hgy59:
and it would be better not to update packages that are under development (octoprint, homeassistant, ...) to avoid merge conflicts or packages that fail to build on the master branch.
As there is a redesign ongoing to separate pure python and cross compiled wheels (@th0ma7), a lot of merge conflicts are predictable.

publicarray:
Yea fair point.

publicarray:
@hgy59 For consistency these should have been set. I don't mind which way we co, but I do prefer consistency.

https://github.com/SynoCommunity/spksrc/blob/9e02f76eb6bd5afffbead8b3903fa307ee174bd1/spk/fishnet/src/requirements.txt#L9

https://github.com/SynoCommunity/spksrc/blob/9e02f76eb6bd5afffbead8b3903fa307ee174bd1/spk/gateone/src/requirements.txt#L1-L6

https://github.com/SynoCommunity/spksrc/blob/9e02f76eb6bd5afffbead8b3903fa307ee174bd1/spk/python2/src/requirements.txt#L31

PR #4920, opened 2021-10-17
Sickchill 6 7 rebuild with Py3.8.12-6

BKSteve:
Continuation of #4901 after Py 38.12 #4902 

All 11/11 packages built on first pass. Errors remain in the logs pertaining to `cffi` and `setuptools_rust` during the `cross/poetry` build part of the process for `88f6281-6.1, armv7-6.1 and 7.0, hi3535-6.1, qoriq-6.1`
With `evansport` giving `libz.so: cannot open shared object file: No such file or directory` error during the Py3.8.12 build process.

PS. This is labelled as version 18 as that was my test, probably should be 2 or 3.

PR #4951, opened 2021-11-03
wheel.mk: Redesign proposal

th0ma7:
_Motivation:_  Relates to the python wheel fix (e.g. `0.0.0` and `invalid version found`), after a few exchanges was suggested to redesign python wheel process.

_Linked issues:_  #4921 

### Overall design
This proposal is a first attempt at this which does the following:
1. Allows using any filename (thus backward compatible with existing `requirements.txt`)
2. Allow using distinct files such that if the filename is `requirements-pure.txt` then its content is considered `pure`.
3. Same behavior applies to `requirements-cross.txt` as `cross` (similar behavior as now)
4. Allow using `pure` and/or `cross` prefix to python wheel definitions (with exception of dedicated files)
5. Allow using `WHEELS` variable with direct wheel names, with or without prefix (no-prefix defaults to `pure)
6. Allow re-defining `WHEEL_DEFAULT_PREFIX` variable for no-prefix (defaults set to `cross` unless `ARCH` is undefined wheras it gets defined to `pure`)

### How it works
When gathering all wheels to be processed it generates:
* `wheelhouse/requirement-pure.txt`
* `wheelhouse/requirement-cross.txt`
* `wheelhouse/requirement-abi3.txt` -> This allows generating cross-compiled wheels with limited python API/ABI

It then calls `pip` once per file to be processed (e.g. `cross` and `abi3` using pip from crossenv vs `pure` using pip from native) followed by the creation of a unique `requirements.txt` file containing all wheels for packaging in `staging/share/wheelhouse` used at package installation time.

#### `WHEEL_DEFAULT_PREFIX`
At the lower-level, by default the `WHEEL_DEFAULT_PREFIX` is always set to `cross`.  This means that no matter what the filename is to define requirements (except `requirement-pure.txt`) it will always cross-compile.  Although when `TC_ARCH` is undefined, the value of `WHEEL_DEFAULT_PREFIX` gets flipped to `pure` no-arch type packages.

### Finding details no. 1
@publicarray found out that call `pip` instead of `$(PIP)` solved a specific use case he had with `sickchill` (ref.: https://github.com/SynoCommunity/spksrc/pull/4920/files#r743494756).

I started investigating and ended-up noticing the `pip` difference between the calls where it would either point to `system`, `crossenv` or `native` but ALL where redirected to default `pip` which fallback to `native` (if native was indeed compiled).

After testing I now believe that most (but not all) failures where caused by this in the first place where using "native" `pip` would make "cross-compiled" wheels to fail as using a `pip` outside from the `crossenv`.

As such I've taken this a little further and was able to build 99% of `sickchill` wheels using only cross-compiling, while redirecting the 1% left over that still fails to using default `native` pip.  This is a rather different approach than prior (e.g. dividing all "pure" vs "cross") where in this case everything is "cross" by default and only the ones which fails (a handful) is being redirected to use "native" and compiled as "pure-python" wheels.

Just to confirm my theory, I've played with `python310`, `octoprint`, `beets` and `sickchill` and remixed the wheels into a single `requirements.txt` where the handful "pure-python" uses the `pure` prefix.

### Finding details no. 2
I made a new step into simplifying the wheel requirement management: adding `poetry` to the `crossenv`!
Immediately I was able to transform `sickchill` into a single & entirely cross-compile requirement file.
@hgy59 Yes there is still a need to manage `pure` vs `cross` wheels but the remaining "pure" exceptions is now really really thin.

### TODO
- [x] Update the wiki
- [x] Allow passing `pure` and `cross` prefix to direct wheel calls into `WHEELS` variable (for DEBUG purpose)
- [x] avoid the prefixes always (unless used for debug purpose, see above)
- [x] only support requirements files and not packages for the WHEELS-variables.
- [x] Allow re-using existing `requirements.txt`
- [x] Allow using distinct files for `pure` vs `cross` (@publicarray)
- [x] allow building `abi3` wheels (e.g @hgy59 fix the rename cross compiled wheels as `*none-any.whl`)
- [x] solve the workaround for the wrong arch name (@hgy59)

### Optional - Migrate packages to python 3.10
- [x] beets (@ymartin59)
- [x] borgbackup
- [x] duplicity
- [x] flexget
- [x] homeassistant (@hgy59)
- [x] mercurial
- [x] octoprint
- [x] rdiff-backup
- [x] sabnzbd (@Safihre)
- [x] sickchill (@BKSteve)
- [x] znc (@worstje)

### Package testing
- [x] beets - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] borgbackup: - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] duplicity: - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] flexget - `kvmx64-7.0` - installation: NO ERRORS, basic testing: need to understand configuration 1st
- [x] homeassistant - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] mercurial - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] python2 - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] python3 - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] python38 - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] python310 - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] octoprint - `kvmx64-7.0` - installation: NO ERRORS, basic testing: web portal setup wizard startup OK
- [x] rdiff-backup - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] sabnzbd - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] sickchill - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK
- [x] znc - `kvmx64-7.0` - installation: NO ERRORS, basic testing: OK

### Checklist


th0ma7:
@hgy59 to initiate discussions I've only focused on `python310` as it isn't yet in usage by any package.  If we come up with a design that looks fine then we could look into adapting all packages for using this.

Note that this proposal ended-up being much more straight-forward than expected to implement.  I suggest to pay attention to:
1. `python310` Makefile `WHEELS` variable definition
2. `requirements.txt` file where there is a mixture of `pure`, `cross` and no-prefix being used.

Please @Safihre , @publicarray, @ymartin59  and other fellow devs, feel free to join-in, inputs more than welcomed!

publicarray:
FYI the wiki needs updating: https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-packages

Safihre:
Still wonder if this is not just a bug in pip itself that it breaks for pure wheels if cross compiled. We are doing a lot of work here and it might be as simple as getting the base problem fixed at pip.. Isn't that worth checking?

th0ma7:
> FYI the wiki needs updating: https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-packages

@publicarray Added to the TODO.

th0ma7:
> Still wonder if this is not just a bug in pip itself that it breaks for pure wheels if cross compiled. We are doing a lot of work here and it might be as simple as getting the base problem fixed at pip.. Isn't that worth checking?

It certainly is worth checking.  Still, in the meantime the issue as been undergoing since quite a while, well hidden and needs a workaround it.

spk/python310/src/requirements.txt
```diff
@@ -18,33 +13,49 @@ distlib==0.3.3
...
-Unidecode==1.3.2
-urllib3==1.26.7
-virtualenv==20.9.0
-webencodings==0.5.1
-Werkzeug==2.0.2
-zipp==3.6.0
+pure:html5lib==1.1
+pure:idna==3.3
+pure:importlib-metadata==4.8.1
+pure:importlib-resources==5.3.0
```
publicarray, line 19: Since I'm a python noob I just want to comment on the style, I would prefer 2 separate files rather than prefixing. I know some text editors make is easier but still that's still a lot of text to change. It also departs what is otherwise is a valid requirments.txt that developers are already familiar with.

th0ma7: Good points.  This made me thinking, could we do both?

Commit below https://github.com/SynoCommunity/spksrc/pull/4951/commits/c9f1ef0dfe5d769779c93808b2d57749752d223f tries to alleviate this by:
a. allow using distinct files such that if the filename is `requirements-pure.txt` then consider it `pure`.
b. same behaviour applies to `requirements-cross.txt` as `cross` (similar to current)
c. for anything else, consider using `cross`|`pure` prefixes, if no prefix fallback to `pure`

This allows keeping already modified files to stay "as-is" because:
1. `requirements.txt` was used for `pure` with no prefix with recent changes, so still good to go using catch-all c) and sort of as backward compatible as we can get
2. `requirements-cross.txt` behavior stays as it is
3. and it allows a mixture of possibilities to play with to get things to work as you'd like where you can:
* Use the `WHEELS` variable with or direct package naming, with or without prefix
* Use the filename you desire (using prefix for non-default)
* Change default prefix
* etc.

publicarray: @th0ma7 🤩 Thanks 👍

th0ma7:
@hgy59 surprisingly it actually rebased like a charm!  If you can comment on the proposal once you have a few spare cycles would be much appreciated, cheers!

spk/octoprint/src/requirements.txt
```diff
@@ -14,14 +14,14 @@ itsdangerous==0.24
 Jinja2==2.8
 Markdown==2.6.2
 netaddr==0.7.17
-netifaces==0.10.4
+cross:netifaces==0.10.4
```
hgy59, line 17: I already have this on my branch for #4096, but since I rebased with master to get requirement.txt and requirements-cross.txt it fails to install with the cross compiled netifaces.
And I don't know what's wrong with it.

publicarray: @hgy59 when did you rebase? In https://github.com/SynoCommunity/spksrc/pull/4920#pullrequestreview-798644475 there is a fix that worked for sickchill.

hgy59:
some of my suggestions
- avoid the prefixes and use different requirements files for pure and cross (as @publicarray requested)
- we should only support requirements files and not packages for the WHEELS-variables. Some times ago I moved all package==version definitions into requiements files (but we never forced an error for such definitions)
- it might be a problem that we rename cross compiled wheels as *none-any.whl. This is an issue with pycryptodome and pycryptodomex (see homeassistant service-setup.sh) as those must have the "abi3" in the wheels name. But on the other side I left the none-any name in homeassistant and rename the cross compiled wheels at installation time before installing the wheels from whl files. This is also a workaround for the wrong arch name in the cross compiled wheels for armv7 (the wheels have  `arm` instead of `armv7l` in the wheels name before renaming in spksrc.wheels.mk.
- in homeassistant only cross compiled wheels are in the package, pure wheels are downloaded at installation time
- if we cannot fix the cross wheels name and must rename at installation time, it might be better to have different wheels folders in the packages for cross and pure wheels, otherwise it is not possible to distingish those.
- fixing wheel names at installation is easy as the arch must match `uname -m` on the target platform.

I give this collection of information here, as I will be offline for at least next week...

spk/octoprint/src/requirements.txt
hgy59: reinstall after uninstall solved the issue with netifaces.

miigotu:
I did not read the whole thread, but is it possible to use pyproject.toml instead of requirements.txt entirely (PEP518)? 

Why can't common wheels be built and stored in a repository, and pulled into the build for each project? 
See:
https://github.com/alpine-wheels/index
https://wheel-index.linuxserver.io/ubuntu/
https://wheel-index.linuxserver.io/alpine/
https://www.piwheels.org/simple (https://github.com/piwheels/piwheels)

I would much like to see the build system not rely solely on setuptools. We use poetry entirely for the whole update, build, and release process and having that as an option would be great. You could poetry build our wheel, then poetry export the requirements if necessary, then pip download the requirements from your index server (or git repository like the linuxserver one does) and zip up the spk and control files. Whole process would take a few seconds as long as there isn't a new version of a c-ext to be built.

th0ma7:
> I did not read the whole thread, but is it possible to use pyproject.toml instead of requirements.txt entirely (PEP518)?

Not sure if it is fully ready, reading on it mentions it can't install dependencies yet with it (but really I'm no close to be a pro in this).

> Why can't common wheels be built and stored in a repository, and pulled into the build for each project? See: https://github.com/alpine-wheels/index https://wheel-index.linuxserver.io/ubuntu/ https://wheel-index.linuxserver.io/alpine/ https://www.piwheels.org/simple (https://github.com/piwheels/piwheels)

@miigotu that is a really interesting option indeed...  Really not sure how to do that but that could be promising.

miigotu:
> > I did not read the whole thread, but is it possible to use pyproject.toml instead of requirements.txt entirely (PEP518)?
> 
> Not sure if it is fully ready, reading on it mentions it can't install dependencies yet with it (but really I'm no close to be a pro in this).
> 
> > Why can't common wheels be built and stored in a repository, and pulled into the build for each project? See: https://github.com/alpine-wheels/index https://wheel-index.linuxserver.io/ubuntu/ https://wheel-index.linuxserver.io/alpine/ https://www.piwheels.org/simple (https://github.com/piwheels/piwheels)
> 
> @miigotu that is a really interesting option indeed...  Really not sure how to do that but that could be promising.

What can't install dependencies with pyproject? poetry sure can, currently spksrc cannot (but should). Setuptools is going to be replaced or fully pyproject.toml supported sooner than later imo.

If you download Sickchill from got, you `poetry install` and you have the package built and all dependencies installed along with the package, in a fresh virtualenv. `poetry upgrade` upgrades all the dependencies in that venv.

publicarray:
> Setuptools is going to be replaced or fully pyproject.toml supported sooner than later imo.

After reading https://venthur.de/2021-06-26-python-packaging.html, I think there are going to be a few separate package managers for the foreseeable feature. I would love everything to be poetry but as long as the community hasn't chosen it as the main one... the python dependency saga continues.

th0ma7:
I consider the PR ready for a first review.  Many additional issues or caveats that where hit due to:
1. the wrong native `pip` called instead of cross
2. as well as fixing many wheels by including by default in python3* default `crossenv` the infamous `poetry`.

mk/spksrc.download.mk
```diff
@@ -157,7 +157,7 @@ download_target: $(PRE_DOWNLOAD_TARGET)
 	      else \
 	        $(MSG) "  wget $${url}" ; \
 	        rm -f $${localFile}.part ; \
-	        wget --secure-protocol=TLSv1_2 -nv -O $${localFile}.part -nc $${url} ; \
+	        wget --secure-protocol=TLSv1_2 --no-check-certificate -nv -O $${localFile}.part -nc $${url} ; \
```
publicarray, line 160: Please no.

Safihre: Agreed!!

th0ma7: good call, and no worry, this is **temporary** as there was a download failure due to a invalid certificate for `native/yasm` thus blocking the build.

publicarray: Ok thanks for the additional info, I mean you got to do what you got to do (for testing). But I really don't want it anywhere close to the main branch.

publicarray: Is it https://www.tortall.net/ with the invalid cert?

th0ma7:
It built successfully! woot!  now time to start testing the packages...

What's next (from @hgy59's list):
- [ ] avoid the prefixes
- [ ] use different requirements files for pure and cross (as @publicarray requested)
- [ ] only support requirements files and not packages for the WHEELS-variables.
- [ ] fix the rename cross compiled wheels as *none-any.whl (ref `pycryptodome` and `pycryptodomex` in homeassistant service-setup.sh as those must have the "abi3" in the wheels name)
- [ ] solve the workaround for the wrong arch name in the cross compiled wheels for armv7 (the wheels have arm instead of armv7l in the wheels name before renaming in spksrc.wheels.mk.
- [ ] use `homeassistant` related cross compiled wheel's naming fixing at installation time must match uname -m on the target platform.


1. As for the prefix specifically I'm ok for not using it anywhere.  Although I would suggest leaving the functionality around as this saved me a lot of efforts while debugging packages.
2. item 2 (individual files): TODO, very straight forward
3. item 3 (`WHEELS`): TODO
4. item 4-5-6 (wheel naming): requires removing change that kicked `ffmpeg` to build as there isn't enough cycles to go over the building process for all arches. Proposals:
- create a `spksrc.wheels-env.mk` where definitions gets moved there along with arch package prefix handling for renaming as necessary (or something similar)
- for the abi3 packages, could we create a abi3 dedicated package file to identify them to rename as needed?  or use a very specific prefix for them for the later renaming? (sharing ideas)

mk/spksrc.download.mk
publicarray: It's probably better to switch to a GitHub release then: https://github.com/yasm/yasm/releases

publicarray: I opened an issue just in case: https://github.com/yasm/yasm/issues/190

Safihre: This seems to be a recurring problem for them, so they suggest to use the GitHub release: https://github.com/yasm/yasm/issues/182#issuecomment-895487445

th0ma7: Can someone tackle this (will probably offline for a few days)?  I'll rebase this PR afterwards.

th0ma7:
@SynoCommunity/developers let's see how the github-action build come-out.  Help in testing packages welcomed!

Getting quite pleased with the code as it is currently.  For @hgy59 specifically, I've implemented the following to address some of your concerns:
1. a method to generate abi/api limited wheels (e.g. `abi3`) - currently use in `python310` as demonstration
2. a method to rename wheels to correspond to the arch `uname -m`
Feedback's on this would be much appreciated.

BKSteve:
I tested this with the same SickChill updates of \#4964 and built DSM6 and DSM7 x64. DSM6 version loaded, DSM7 did not with same `invalid file format` 
Build duration ~ same.

BKSteve:
Tried building with Py310 and ran into build issues with PyNaCl here's the full zipped log file.
[Uploading build-x64-7.0.zip…]()

th0ma7:
@SynoCommunity/developers all github-action green lights are on.  I've tested the python310 & sickchill package successfully, additional testers would be much appreciated.

I consider it now ready for a review for potential merging.  Next would be to complete transitioning all remaining python3* packages.

publicarray:
Thanks @th0ma7, nice work. Before merging I'd like to know if it works on other arches and  echo @BKSteve comment https://github.com/SickChill/SickChill/issues/7504#issuecomment-968216492

publicarray:
One concern is that there are lot's of package specific changes that will course conflicts with other package updates in other PRs e.g sickchill and beets

BKSteve:
You'll see on 7504 that there's still some checking and even one comment to resolve.

> I get invalid file format with the Python package: python310_evansport-7.0_3.10.0-2.spk

th0ma7:
> One concern is that there are lot's of package specific changes that will course conflicts with other package updates in other PRs e.g sickchill and beets

@publicarray indeed there are and they can easily be reverted as needed.  Package updating really is optional.

> You'll see on 7504 that there's still some checking and even one comment to resolve.
>> I get invalid file format with the Python package: python310_evansport-7.0_3.10.0-2.spk

@BKSteve that's exactly the type of comments I hoping for (but confirming it works would have been even better :)

BKSteve:
Update: By finding that DSM7 packages weren't installable `invalid file format` because they need to be whole number versions, without decimal places, I'm stepping a back from the bleeding edge of wheels as I'm out of my depth.

hgy59:
@th0ma7 you asked me to test python310 with some packages and referenced this PR.

I tried to build homeassistant on the current master that has python310 merged.

unfortunately the package `cross/gevent` fails to build (at least for armv7 and aarch64, but I think it will fail for all cross builds).
It tries always to find x86_64 cffi backend (`_cffi_backend.cpython-310-x86_64-linux-gnu.so`) instead of the arch specific one.

_It's hard to keep an overview of all the current work in progress, so may be this is already fixed in this PR..._

```
===>  Extracting for gevent
tar -xzpf /spksrc/cross/gevent/../../distrib/gevent-21.8.0.tar.gz -C /spksrc/spk/homeassistant/work-armv7-6.1
===>  Patching for gevent
===>  Configuring for gevent
===>  Compiling for gevent
  File "<string>", line 1, in <module>
  File "setup.py", line 481, in <module>
    run_setup(EXT_MODULES)
  File "setup.py", line 348, in run_setup
    setup(
  File "/spksrc/spk/homeassistant/work-armv7-6.1/install/var/packages/homeassistant/target/lib/python3.10/site-packages/setuptools/__init__.py", line 153, in setup
    return distutils.core.setup(**attrs)
  File "/spksrc/native/python310/work-native/install/usr/local/lib/python3.10/distutils/core.py", line 108, in setup
    _setup_distribution = dist = klass(attrs)
  File "/spksrc/spk/homeassistant/work-armv7-6.1/install/var/packages/homeassistant/target/lib/python3.10/site-packages/setuptools/dist.py", line 453, in __init__
    _Distribution.__init__(
  File "/spksrc/native/python310/work-native/install/usr/local/lib/python3.10/distutils/dist.py", line 292, in __init__
    self.finalize_options()
  File "/spksrc/spk/homeassistant/work-armv7-6.1/install/var/packages/homeassistant/target/lib/python3.10/site-packages/setuptools/dist.py", line 831, in finalize_options
    ep(self)
  File "/spksrc/spk/homeassistant/work-armv7-6.1/install/var/packages/homeassistant/target/lib/python3.10/site-packages/setuptools/dist.py", line 852, in _finalize_setup_keywords
    ep.load()(self, ep.name, value)
  File "/spksrc/spk/homeassistant/work-armv7-6.1/install/var/packages/homeassistant/target/lib/python3.10/site-packages/cffi/setuptools_ext.py", line 219, in cffi_modules
    add_cffi_module(dist, cffi_module)
  File "/spksrc/spk/homeassistant/work-armv7-6.1/install/var/packages/homeassistant/target/lib/python3.10/site-packages/cffi/setuptools_ext.py", line 49, in add_cffi_module
    execfile(build_file_name, mod_vars)
  File "/spksrc/spk/homeassistant/work-armv7-6.1/install/var/packages/homeassistant/target/lib/python3.10/site-packages/cffi/setuptools_ext.py", line 25, in execfile
    exec(code, glob, glob)
  File "src/gevent/libev/_corecffi_build.py", line 31, in <module>
    ffi = FFI()
  File "/spksrc/spk/homeassistant/work-armv7-6.1/install/var/packages/homeassistant/target/lib/python3.10/site-packages/cffi/api.py", line 48, in __init__
    import _cffi_backend as backend
ImportError: /spksrc/spk/homeassistant/work-armv7-6.1/crossenv/build/lib/python3.10/site-packages/_cffi_backend.cpython-310-x86_64-linux-gnu.so: cannot open shared object file: No such file or directory
make[3]: *** [../../mk/spksrc.python-wheel.mk:33: build_python_wheel] Error 1
make[3]: Leaving directory '/spksrc/cross/gevent'
make[2]: *** [../../mk/spksrc.depend.mk:51: depend_target] Error 2
make[2]: Leaving directory '/spksrc/spk/homeassistant'
make[1]: Leaving directory '/spksrc/spk/homeassistant'
user@spksrc_py310:/spksrc/spk/homeassistant$
```

th0ma7:
> @th0ma7 you asked me to test python310 with some packages and referenced this PR.

@hgy59 thnx this is much appreciated!

> I tried to build homeassistant on the current master that has python310 merged.
> unfortunately the package `cross/gevent` fails to build (at least for armv7 and aarch64, but I think it will fail for all cross builds). It tries always to find x86_64 cffi backend (`_cffi_backend.cpython-310-x86_64-linux-gnu.so`) instead of the arch specific one.

yup. sadly no surprises there.  This PR contains important fixes not yet in master.  I haven't ported `homeassistant` to python310 (I intentionally left it at Python3.8 where it builds fine) but I could take a quick rough start at it to pave the way for you if you see fit.  Also, feel free to join-in and add commits to this PR for `homeassistant` or other issues you may find.

> _It's hard to keep an overview of all the current work in progress, so may be this is already fixed in this PR..._

And let me know if I can make things easier to follow (and ideas on the how).  As `homeassistant` is quite of a beast on its own having this one fully ported to python310 would help set the path forward for others.

th0ma7:
@hgy59 I managed to get `homeassistant` to work well with Python 3.10.  Tested on my DS918 (x64-apollolake).
I intentionally downgraded `pillow` to 8.2 as it was blocking homeassistant to update it locally afterwards.
And `beets` would benefit from having access to `pillow` but it is a no-arch package.  So having it within python 3.10 solves that.

th0ma7:
@Safihre heads-up, I was able to build and test successfully `sabnzbd` against python 3.10.  Can you try it out to confirm?  Thnx in advance for your help :)

th0ma7:
@BKSteve heads-up, I was able to build and test successfully `sickchill` with this PR.  Can you give it a shot to confirm?  thnx in advance!

spk/homeassistant/Makefile
```diff
@@ -8,7 +8,6 @@ SPK_DEPENDS = "python310"
 BUILD_DEPENDS = cross/python310
 
 DEPENDS += cross/gevent
-DEPENDS += cross/homeassistant.pillow
```
hgy59, line 12: homeassistant still depends on cross/pillow

th0ma7: But it is, the proposal here is to have it under `python310` which makes it available not only for `homeassistan` but as well for `beets` which is noarch.

When testing, if `python310` provides it, it must be at version 8.2.0 as otherwise `homeassistant` tentative to downgrade it into its local crossenv fails, and thus the application overall doesn't work out.

th0ma7: BTW, I noticed that `octoprint` as a different method to install wheels at post-install which allows it to download + install a local copy independently from the version provided by default by the python package.  Just a thought if that could prove useful.

Otherwise, as is, the basic `homeassistant` testing I did looked promising.

th0ma7:
@ymartin59 with limited changes I was able to get `beets` to run with Python 3.10.
```
 /usr/local/bin/beet --version
beets version 1.5.0
Python version 3.10.0
no plugins loaded
```

th0ma7:
Hi all @SynoCommunity/developers , I now consider this PR ready for merge.

I propose merging this as-is and publish only `python310` package for a start.  I'd leave you to review/update packages you maintain as you see fit as you might want to take the opportunity to update or make additional changes.

Due to the changes to cross/pure/abi3 wheels there will probably some side-effects to packages not in scope of this PR build that I could not test.  Remaining packages that I intend to convert in a subsequent PR #4965 that I've started working on :
- `plexpy`
- `fishnet`
- `rutorrent`
- `salt-minion`
- `tvheadend`
- `znc`
- `bazarr`

Please let me know if you see an issue with this proposal and have an alternate approach we can use.

PR #4955, opened 2021-11-06
Python310 fixes

th0ma7:
_Motivation:_  Move python 3.10 fixes from #4951 into an individual PR
_Linked issues:_  Also relates to changes in #4920

### Checklist


th0ma7:
@hgy59 and all, while playing with @publicarray discovery related to `pip` I made some new progress that might alleviate most (not all) usage of separate requirements.txt files.  I'll try to demonstrate my findings once (and if) it turns out to work.

th0ma7:
Going back from the original build issues (sometimes hidden), there where cases where it would:
1. fail with `0.0.0` version or
2. build failures such as `outside of build environment` or 
3. unable to find an appropriate version

Yesterday @publicarray found out that call `pip` instead of `$(PIP)` solved a specific use case he had with `sickchill` (ref.: https://github.com/SynoCommunity/spksrc/pull/4920/files#r743494756).

I started investigating and ended-up noticing the `pip` difference between the calls where it would either point to `system`, `crossenv` or `native` but ALL where redirected to default `pip` which points to `native` (if native was indeed compiled).

After testing I now believe that most (but not all) failures where caused by this in the first place where using "native" `pip` would make "cross-compiled" wheels to fail as using a `pip` outside from the `crossenv`.

As such I've taken this a little further and was able to build 99% of `sickchill` wheels using only cross-compiling, while redirecting the 1% left over that still fails to using default `native` pip.  This is a rather different approach than prior (e.g. dividing all "pure" vs "cross") where in this case everything is "cross" by default and only the ones which fails (a handful) is being redirected to use "native".

Just to confirm my theory, I've played with `sickchill` (as the special case came from out there) and remixed the wheels.  I'm letting the github-action build to go though to check for other python package in the same situation of having one of the 1% wheel failure needing to be redirected to "native".

hgy59:
@th0ma7 pcre download is broken, you can use 3a57a4a196a71aa1ee180c730adf5701b1d367b6

unfortunately the download tries for about 30 minuntes to download from the broken url...

PR #4963, opened 2021-11-11
native/yasm: use github release & cross/libsodium: update digests

publicarray:
See: https://github.com/SynoCommunity/spksrc/pull/4951#discussion_r746493700

Just in-case the cert expires again

### Checklist
- [x] native build


th0ma7:
Why is it failing?  This seems like a trivial PR...

th0ma7:
@publicarray I suggest we simply merge this and start using github by default.  Are you ok with this?

publicarray:
I used minisign to verify the signature of libsodium :+1:

publicarray:
@th0ma7 Thank you!

publicarray:
Arch uses  `libsodium-1.0.18.tar.gz` not `libsodium-1.0.18-stable.tar.gz`. maybe that is part of the problem we are using the wrong file?. https://github.com/archlinux/svntogit-packages/blob/packages/libsodium/trunk/PKGBUILD

th0ma7:
I've switched to `-stable` a few weeks ago due to build issues elsewhere in the python dreamworld :)

publicarray:
https://github.com/jedisct1/libsodium/discussions/1124#discussioncomment-1628482

publicarray:
Based on he's response I think be should be using `libsodium-1.0.18.tar.gz`

Issue #4980, opened 2021-11-27
pyhon wheel not a supported on this platform - armv5tel


th0ma7:
### Setup
_Package Name:_ python310
_Package Version:_ not released yet

_NAS Model:_ synology_88f6282_213air
_NAS Architecture:_ armv5tel
_DSM version:_ 

### Issue
With the recent changes to how the wheel name are being managed it was expected that using `uname -m` as identifier in the naming was solving issues for `armv5tel`, `armv7` and `armv7l` but it seems original testing feedbacks wasn't sufficient to confirm that this change was working ok.

Issue currently being discussed here: https://github.com/SickChill/SickChill/issues/7672
Raised here in particular: https://github.com/SickChill/SickChill/issues/7672#issuecomment-980395077

Current wheel name example after being copied + renamed to staging:
```
$ find work*/staging/share/wheelhouse/lxml-*.whl
work-88f6281-6.1/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-arm_marvell_linux_armv5tel.whl
work-aarch64-6.1/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-aarch64_unknown_linux_aarch64.whl
work-aarch64-7.0/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-aarch64_unknown_linux_aarch64.whl
work-armv7-6.1/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-arm_unknown_linux_armv7.whl
work-armv7-7.0/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-arm_unknown_linux_armv7.whl
work-evansport-6.1/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-i686_pc_linux_i686.whl
work-hi3535-6.1/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-arm_cortexa9_linux_armv7l.whl
work-kvmx64-6.1/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-x86_64_pc_linux_x86_64.whl
work-kvmx64-7.0/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-x86_64_pc_linux_x86_64.whl
work-qoriq-6.1/staging/share/wheelhouse/lxml-4.6.3-cp310-cp310-powerpc_e500v2_linux_ppc.whl
[...]

PR #4981, opened 2021-11-27
[WIP] Wheel name fix

th0ma7:
_Motivation:_  With the redesign wheels no longer uses `none-any` suffix but rather the mathing `uname -m` of the NAS architecture machine name.  This PR address some of the shortfalls where wheels have extra arguments in their platform tag that needs to be reworked.
_Linked issues:_  #4980 and https://github.com/SickChill/SickChill/issues/7672#issuecomment-980395077

### Checklist

### DIFF
--- a/mk/spksrc.wheel-env.mk
+++ b/mk/spksrc.wheel-env.mk
@@ -47,11 +47,7 @@ ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS)),$(ARCH))
 PYTHON_ARCH = armv5tel
 endif
 
-ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS)),$(ARCH))
-PYTHON_ARCH = armv7
-endif
-
-ifeq ($(findstring $(ARCH),$(ARMv7L_ARCHS)),$(ARCH))
+ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
 PYTHON_ARCH = armv7l
 endif
 
--- a/mk/spksrc.wheel-env.mk
+++ b/mk/spksrc.wheel-env.mk
@@ -64,7 +64,7 @@ PYTHON_ARCH = x86_64
 endif
 
 ifeq ($(findstring $(ARCH),$(i686_ARCHS)),$(ARCH))
-PYTHON_ARCH += i686
+PYTHON_ARCH = i686
 endif
 
 install_python_wheel:
--- a/mk/spksrc.wheel-env.mk
+++ b/mk/spksrc.wheel-env.mk
@@ -90,7 +90,9 @@ install_python_wheel:
 		else \
 			for w in *.whl; do \
 				$(MSG) Copying to wheelhouse: $$(echo $$w | sed -E "s/(.*linux_).*(\.whl)/\1$(PYTHON_ARCH)\2/") ; \
-				cp -f $$w $(STAGING_INSTALL_WHEELHOUSE)/$$(echo $$w | sed -E "s/(.*linux_).*(\.whl)/\1$(PYTHON_ARCH)\2/") ; \
+				cp -f $$w $(STAGING_INSTALL_WHEELHOUSE)/$$(echo $$w \
+					| sed -E "s/(.*-).*(linux_.*\.whl)/\1\2/" \
+					| sed -E "s/(.*linux_).*(\.whl)/\1$(PYTHON_ARCH)\2/") ; \
 			done ; \
 		fi ; \
 	fi
--- a/mk/spksrc.wheel-env.mk
+++ b/mk/spksrc.wheel-env.mk
@@ -85,10 +85,9 @@ install_python_wheel:
 			done ; \
 		else \
 			for w in *.whl; do \
-				$(MSG) Copying to wheelhouse: $$(echo $$w | sed -E "s/(.*linux_).*(\.whl)/\1$(PYTHON_ARCH)\2/") ; \
-				cp -f $$w $(STAGING_INSTALL_WHEELHOUSE)/$$(echo $$w \
-					| sed -E "s/(.*-).*(linux_.*\.whl)/\1\2/" \
-					| sed -E "s/(.*linux_).*(\.whl)/\1$(PYTHON_ARCH)\2/") ; \
+				_new_name=$$(echo $$w | sed -E "s/(.*-).*(linux_).*(\.whl)/\1\2$(PYTHON_ARCH)\3/") ; \
+				$(MSG) Copying to wheelhouse: $$_new_name ; \
+				cp -f $$w $(STAGING_INSTALL_WHEELHOUSE)/$$_new_name ; \
 			done ; \
 		fi ; \
 	fi
--- a/spk/python310/Makefile
+++ b/spk/python310/Makefile
@@ -9,9 +9,11 @@ DEPENDS  = cross/$(SPK_NAME)
 # setup during package install
 DEPENDS += cross/setuptools cross/pip cross/wheel
 
-# Required for misc wheels
+# Required for lxml wheel
 DEPENDS += cross/libxml2
 DEPENDS += cross/libxslt
+
+# Required for PyYAML
 DEPENDS += cross/libyaml
 
 # Required for pycurl wheel
--- a/spk/python310/Makefile
+++ b/spk/python310/Makefile
@@ -1,7 +1,7 @@
 SPK_NAME = python310
 SPK_VERS = 3.10.0
 SPK_VERS_MAJOR_MINOR = $(word 1,$(subst ., ,$(SPK_VERS))).$(word 2,$(subst ., ,$(SPK_VERS)))
-SPK_REV = 3
+SPK_REV = 4
 SPK_ICON = src/python3.png
 
 DEPENDS  = cross/$(SPK_NAME)
--- a/spk/python310/src/requirements-crossenv.txt
+++ b/spk/python310/src/requirements-crossenv.txt
@@ -14,6 +14,7 @@ chardet==4.0.0
 charset-normalizer==2.0.7
 click==8.0.3
 cryptography==3.3.2
+Cython==0.29.24
 distlib==0.3.3
 Flask==2.0.2
 future==0.18.2
--- a/spk/python310/src/requirements-crossenv.txt
+++ b/spk/python310/src/requirements-crossenv.txt
@@ -14,7 +14,6 @@ chardet==4.0.0
 charset-normalizer==2.0.7
 click==8.0.3
 cryptography==3.3.2
-Cython==0.29.24
 distlib==0.3.3
 Flask==2.0.2
 future==0.18.2
@@ -25,7 +24,7 @@ immutables==0.16
 ipaddress==1.0.23
 itsdangerous==2.0.1
 Jinja2==3.0.2
-lxml==4.6.3
+lxml==4.6.4
 MarkupSafe==2.0.1
 msgpack-python==0.5.6
 netifaces==0.11.0


th0ma7:
Currently building and intent to make it available for `sickchill` users for testing on armv5 to confirm if this solves the issue.

hgy59:
@th0ma7 just found another issue.
`PYTHON_ARCH = armv7l` must be used for `ARMv7L_ARCHS` and for `ARMv7_ARCHS`. 
`PYTHON_ARCH = armv7` does not exist and is not valid.
The `L` in python arch `armv7l` is of different meaning than the `L` in ARMv7L (i.e. ARMv7L means not hf = hardfloat).

You can find the valid _python platform tags_ for _manylinux_ in [PEP-600](https://www.python.org/dev/peps/pep-0600/)
unfortunately there is none listed for ARMv5.

The wheel naming convention is defined in [PEP-427](https://www.python.org/dev/peps/pep-0427/)
The wheel filename is `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl`.
Our cross compiled wheels must use the platform tag `linux_${ARCH}` and ${ARCH} must match the cpu type returned by `uname -m` of the target platform.

I can't remember where I got the information about uname -m from, but the original error I solve in homeassistant package was caused by `crossenv` that used `arm` instead of `armv7l` for `ARMv7_ARCHS`.
It looks like an open issue in crossenv to evaluate the correct target platform (https://github.com/benfogle/crossenv/issues/76)

As the renamed whl was installable, the WHEEL file inside the *.whl can still define `Tag: cp310-cp310-aarch64_unknown_linux_gnueabi`. 
To successfully install `MarkupSafe-2.0.1-cp310-cp310-aarch64_unknown_linux_aarch64.whl` only the rename of the whl file to `MarkupSafe-2.0.1-cp310-cp310-linux_aarch64.whl` is required.

th0ma7:
@hgy59 I've adjusted the code in consequence, now re-building and will have to validate that the end-names do matches properly your findings (thnx btw)!

hgy59:
I built Python 3.10 with your changes.
And successfully installed octoprint 1.7.2 built with python310 dependency.

but I had to include `MarkupSafe==1.1.1`  and  `PyYAML==5.4.1` in the octoprint requirements file, as the versions of these wheels in python310 are incompatible.

This is a general issue. I suppose to include only the really basic wheels in python310. Otherwise we always have to deploy a lot of dependent packages. IMHO it is better to have the wheels duplicated in the dependent packages to keep them independent regarding the wheels.

The initial idea to include wheels in the python package was to include arch dependent wheels so some packages will be arch-independent (i.e. noarch), as all required cross-wheels are installed by the python package.

With this point of view we can remove all pure-python wheels from python310 package.
I suppose to remove additional wheels like
- Flask (incl. Flask_*, Werkzeug, ...)
- MarkupSafe
- lxml
- sqlalchemy

We only need cross wheels in Python 3.10 for the beets package (see below)

And we should document in the requirements file of Python310 which package depends on each entry - not only the reverse dependency in the requirements file of dependent packages (as commented with  `==> python310`).
Otherwise it is a mess to update any wheel in python310.

___
What is the reason to include cython?

___
Current noarch python dependent packages:
- beets           (python 3.10)
- plexpy-custom   (python 3.7)

Most of the noach python dependent packages depend on Python 2 (outdated anyway)
- couchpotatoserver
- couchpotatoserver-custom
- headphones
- headphones-custom
- htpcmanager
- lazylibrarian
- maraschino
- nzbhydra
- nzbmegasearch
- mylar
- plexivity
- salt-master
- saltpad
- sickbeard-custom
- sickrage
- subliminal

mk/spksrc.wheel-env.mk
```diff
@@ -90,7 +86,9 @@ install_python_wheel:
 		else \
 			for w in *.whl; do \
 				$(MSG) Copying to wheelhouse: $$(echo $$w | sed -E "s/(.*linux_).*(\.whl)/\1$(PYTHON_ARCH)\2/") ; \
-				cp -f $$w $(STAGING_INSTALL_WHEELHOUSE)/$$(echo $$w | sed -E "s/(.*linux_).*(\.whl)/\1$(PYTHON_ARCH)\2/") ; \
+				cp -f $$w $(STAGING_INSTALL_WHEELHOUSE)/$$(echo $$w \
+					| sed -E "s/(.*-).*(linux_.*\.whl)/\1\2/" \
+					| sed -E "s/(.*linux_).*(\.whl)/\1$(PYTHON_ARCH)\2/") ; \
 			done ; \
```
hgy59, line 92: this can be simplified to (and fixes MSG output)
```
			for w in *.whl; do \
				_new_name=$$(echo $${w} | sed -E "s/(.*-).*/\1/")linux_$(PYTHON_ARCH).whl ; \
				$(MSG) Copying to wheelhouse: $${_new_name} ; \
				cp -f $${w} $(STAGING_INSTALL_WHEELHOUSE)/$${_new_name} ; \
			done ; 
```

hgy59:
a minor typo
you can replace `PYTHON_ARCH += i686` by `PYTHON_ARCH = i686` in mk/spksrc.wheel-env.mk.

mk/spksrc.wheel-env.mk
th0ma7: done, slightly adapted.

th0ma7:
@hgy59 I am more than open in addressing this and figure out the best path forward.  Once defined I suggest we add this background to the wiki to set expectations to developers.

> but I had to include MarkupSafe==1.1.1 and PyYAML==5.4.1 in the octoprint requirements file, as the versions of these wheels in python310 are incompatible.

I often saw messages at installation that pip wasn't able to install the wheel in local env as another version was superseding it in python310.  But on the other hand the package ran fine (is that a false-positive then?).  On the other hand I saw sickchill doing exactly that with no issue.  One thing I'd like to come up with is documenting a proper pip call so it actually always can supersede wheels that comes from python310.  An example is `pillow` where python310 could provide the latest & greatest while allowing `homeassistant` to use another version as needed (current pip call in service-setup.sh doesn't allow that).

> This is a general issue. I suppose to include only the really basic wheels in python310. 

Question is: what are really basic wheels?

> Otherwise we always have to deploy a lot of dependent packages

Can you further elaborate on this, I'm not sure I understand?  I thought (and perhaps wrongly) that if most wheels are part of python default, updating it on a regular basis (along with its wheels) would simplify maintenance of individual apps packages, while also reducing space on the server as each python apps uses far less space.

> IMHO it is better to have the wheels duplicated in the dependent packages to keep them independent regarding the wheels.

What if (and really this is just a thought), keeping all cross by default in python (or as many as we can) and always enforcing downloading of pure-python wheels at installation time instead of providing too many of them in each packages?  I'm sure there is a corner-case here that I'm not thinking of... but what if?

> The initial idea to include wheels in the python package was to include arch dependent wheels so some packages will be arch-independent (i.e. noarch), as all required cross-wheels are installed by the python package.

Fully get that (i.e. beets).  My original thinking was making some mileage on that such as, can I push the really trivial wheels needed into the default python package (I was using `tvheadend` as an example)?  I mean any basic linux python install comes with a minimal set.  But this is where I'm sure where is the line?  I mean, a simple example are the following which are used everywhere, do we really have to repeat them in every packages?  Or should they be included in python default.  I prefer the later but where do we draw the line?
```
certifi==2021.5.30
chardet==4.0.0
requests==2.25.1
urllib3==1.26.5
```

> With this point of view we can remove all pure-python wheels from python310 package. I suppose to remove additional wheels like
>     * Flask (incl. Flask_*, Werkzeug, ...)
>     * MarkupSafe
>     * sqlalchemy
>     * lxml

`lxml` is cross-compiled...

> We only need cross wheels in Python 3.10 for the beets package (see below)
> 
> And we should document in the requirements file of Python310 which package depends on each entry - not only the reverse dependency in the requirements file of dependent packages (as commented with `==> python310`). Otherwise it is a mess to update any wheel in python310.

Agreed.  Probably worth the effort.

> What is the reason to include cython?

AFAIK cython is a C compiler for python wheel.  Having it built-into python allows to an application to download a wheel C source and try to compile it using cython.  This also presumes that python default comes with a minimal set of libraries that cython will be able to use and compile from.

> Current noarch python dependent packages:
>     * beets           (python 3.10)
>     * plexpy-custom   (python 3.7)
> 
> 
> Most of the noach python dependent packages depend on Python 2 (outdated anyway)
> 
>     * couchpotatoserver
>     * couchpotatoserver-custom
>     * headphones
>     * headphones-custom
>     * htpcmanager
>     * lazylibrarian
>     * maraschino
>     * nzbhydra
>     * nzbmegasearch
>     * mylar
>     * plexivity
>     * salt-master
>     * saltpad
>     * sickbeard-custom
>     * sickrage
>     * subliminal

Yeah, a lot of work ahead + migrating all remaining python3 apps to python310.

If I summarize, a few items that needs to be sorted out:
- [ ] ACTION 0: Confirm wheel management high-level strategy
- [ ] ACTION 1: Document wheel strategy to the wiki
- [ ] ACTION 2: Document in the wiki (and modify packages as needed) to allow pip to force using a different wheel version then what's provided by python default.  Also as we do now provide a fully functional `requirements.txt` file in apps packages, it should make use of that instead of installing the `wheelhouse` directory by default.
- [ ] ACTION 3: Document in the requirement file of Python310 which package depends on each entry (and vice-versa in package requirement files)

Related to the wheel management strategy:
1. Define what's are the basic wheels?  Items for consideration are: cross, basic common, cython + minimal library set, noarch requirements, anything else?
2. Can we enforce downloading pure python wheels always at installation time?

When I took a stab at python310 I used the following rule of thumb (that needs to be realigned as we see fit):
1. Provide all crossenv wheels in python whenever possible
2. Eliminate all cross wheels (e.g. spk/cross/*/Makefile) so it uses crossenv always.  That proved to be impossible but I made a lot of progress in that direction.
3. include wheels that are repeated in too many packages.  When ever a wheel was included into more than 3+ apps I was opting to include it in python directly

I will be easier if we define this now than later once python310 is fully deployed.

What's your thought?  Any idea as to how to document that "line" of what's in and what's not?  You already provided quite a few ideas but now how to articulate and document a strategy?

th0ma7:
So there is an interesting use case that might help guide how to manage wheels (and perhaps this is just me not yet having a good awareness on things). Ref.: https://github.com/SickChill/SickChill/issues/7672#issuecomment-981121703

`sickchill` still tries to build `lxml` while it is already present into python310, and ends-up failing miserably.  So mileage may vary depending of each apps.

hgy59:
> `sickchill` still tries to build `lxml` while it is already present into python310, and ends-up failing miserably. So mileage may vary depending of each apps.

Simply add lxml wheel to sickchill package ...

hgy59:
> > What is the reason to include cython?
> 
> AFAIK cython is a C compiler for python wheel. Having it built-into python allows to an application to download a wheel C source and try to compile it using cython. This also presumes that python default comes with a minimal set of libraries that cython will be able to use and compile from.

So cython can be removed, as there is no build environment available on the diskstations (that's why we deliver already cross compiled wheels).

hgy59:
> Question is: what are really basic wheels?

1. cross compiled wheels required by beets (possibly none)

This list may grow while migrating the following packages to python 3 (over the next n years ?)
> Most of the noach python dependent packages depend on Python 2 (outdated anyway)
>  -   couchpotatoserver
>  -   couchpotatoserver-custom
>  -   headphones
>  -   headphones-custom
>  -   htpcmanager
>  -   lazylibrarian
>  -   maraschino
>  -   nzbhydra
>  -   nzbmegasearch
>  -   mylar
>  -   plexivity
>  -   salt-master
>  -   saltpad
>  -   sickbeard-custom
>  -   sickrage
>  -   subliminal

hgy59:
> I often saw messages at installation that pip wasn't able to install the wheel in local env as another version was superseding it in python310.

IMHO the message tells, that is was not possible to uninstall the package in python3 site-packages not in the local env.
I always ignore such messages (probably we could tell pip not to try to update global site-packages).

sample of homeassistant
```
2021/08/17 07:40:00	  Attempting uninstall: ruamel.yaml
2021/08/17 07:40:00	    Found existing installation: ruamel.yaml 0.16.10
2021/08/17 07:40:00	    Not uninstalling ruamel.yaml at /volume1/@appstore/python38/lib/python3.8/site-packages, outside environment /volume1/@appstore/homeassistant/env
2021/08/17 07:40:00	    Can't uninstall 'ruamel.yaml'. No files were found to uninstall.
2021/08/17 07:40:01	  Attempting uninstall: PyYAML
2021/08/17 07:40:01	    Found existing installation: PyYAML 5.3
2021/08/17 07:40:01	    Not uninstalling pyyaml at /volume1/@appstore/python38/lib/python3.8/site-packages, outside environment /volume1/@appstore/homeassistant/env
2021/08/17 07:40:01	    Can't uninstall 'PyYAML'. No files were found to uninstall.
2021/08/17 07:40:05	  Attempting uninstall: psutil
2021/08/17 07:40:05	    Found existing installation: psutil 5.7.3
2021/08/17 07:40:05	    Not uninstalling psutil at /volume1/@appstore/python38/lib/python3.8/site-packages, outside environment /volume1/@appstore/homeassistant/env
2021/08/17 07:40:05	    Can't uninstall 'psutil'. No files were found to uninstall.
2021/08/17 07:40:21	  Attempting uninstall: MarkupSafe
2021/08/17 07:40:21	    Found existing installation: MarkupSafe 1.1.1
2021/08/17 07:40:21	    Not uninstalling markupsafe at /volume1/@appstore/python38/lib/python3.8/site-packages, outside environment /volume1/@appstore/homeassistant/env
2021/08/17 07:40:21	    Can't uninstall 'MarkupSafe'. No files were found to uninstall.
```

th0ma7:
The current cross wheels in python310 are:
```
Pillow-8.2.0-cp310-cp310-linux_armv5tel.whl
MarkupSafe-2.0.1-cp310-cp310-linux_armv5tel.whl
PyNaCl-1.4.0-cp310-cp310-linux_armv5tel.whl
PyYAML-6.0-cp310-cp310-linux_armv5tel.whl
SQLAlchemy-1.4.26-cp310-cp310-linux_armv5tel.whl
bcrypt-3.2.0-cp310-cp310-linux_armv5tel.whl
cffi-1.15.0-cp310-cp310-linux_armv5tel.whl
cryptography-3.3.2-cp310-cp310-linux_armv5tel.whl
greenlet-1.1.2-cp310-cp310-linux_armv5tel.whl
immutables-0.16-cp310-cp310-linux_armv5tel.whl
lxml-4.6.3-cp310-cp310-linux_armv5tel.whl
msgpack_python-0.5.6-cp310-cp310-linux_armv5tel.whl
netifaces-0.11.0-cp310-cp310-linux_armv5tel.whl
psutil-5.8.0-cp310-cp310-linux_armv5tel.whl
pycares-4.1.2-cp310-cp310-linux_armv5tel.whl
pycryptodome-3.11.0-cp35-abi3-linux_armv5tel.whl
pycryptodomex-3.11.0-cp35-abi3-linux_armv5tel.whl
pycurl-7.44.1-cp310-cp310-linux_armv5tel.whl
regex-2021.11.2-cp310-cp310-linux_armv5tel.whl
zope.interface-5.4.0-cp310-cp310-linux_armv5tel.whl
```

From the requirement files beets requires pillow.  And as an example tvheadend requires really trivial wheels (e.g. `certifi`, `chardet`, `requests`, `urllib3`) which I would much like them to be in default python310.

But if I understand you well, you would just provide setuptools, pip and wheel part of python310 (and perhaps pillow due to being dependency for beets)?  Not sure I fully get it.

Also, I suggest we merge this PR and open a new one to review default provided wheels there?

hgy59:
> 2\. Can we enforce downloading pure python wheels always at installation time?

I don't know whether a generic soultion is possible, as we have separated the pure wheels now.

I do this in homeassistant package in additional steps (2x pip with pure wheels requirements files)
For this at least to calls of pip are required
1. install --no-index *.whl
2. install -r pure-wheels-requirements.txt

service_postinst() of homeassistant contains:
```
    echo "Install packages from wheels"
    ${SYNOPKG_PKGDEST}/env/bin/pip install --no-deps --no-input --no-index ${wheelhouse}/*.whl

    echo "Install packages for default_config from index"
    ${SYNOPKG_PKGDEST}/env/bin/pip install --no-deps --no-input --requirement ${SYNOPKG_PKGDEST}/share/postinst_default_config_requirements.txt

    echo "Install packages for homeassistant.components from index"
    ${SYNOPKG_PKGDEST}/env/bin/pip install --no-deps --no-input --requirement ${SYNOPKG_PKGDEST}/share/postinst_components_requirements.txt
```

th0ma7:
Just a thought, what if you had a `requirements-postinstall.txt` that gets added to the resulting requirement file in the package and then at installation time it `pip install --index-url ${...wheelhousepath} -r requirements.txt` which would install locally found wheels + download any missing?

hgy59:
I try to explain  the dependency problem.

Say we have two packages with SPK_DEPENDS=python310
the first one requires MarkupSafe-1.1.1 and the second MarkupSafe-2.0.1

Say we first deliverd Python310 with MarkupSafe-1.1.1
and the first package does not include the MarkupSafe-1.1.1 wheel but uses the one delivered with Python310.

If we update MarkupSave in Python310 to 2.0.1 to allow the second Package to take MarkupSafe from Python package, this will break the first package.
- you cannot install the first package anymore, when Python310 is updated to the version with MarkupSafe-2.0.1

So you must not update wheels used by other packages in the Python package to not break those.
But your honest target was to only update Python packages to update the wheels ....

So from the view of a python dependent package I don't want to depend on wheels bundled with python.
Neither in the first package nor in the second in the example above.


One blocking reason is the DSM update process. An update is always a complete uninstall and a new installation with only backup/restore of the data (var) and not of the python virtual environments nor the installed wheels.


***And you cannot update wheels in packages by updating the python package!***
Even for a simple wheel like certifi you don't update the certifi package in already installed package environments. You always need to deliver an update of the dependent package and the user must update the python package before the dependent package to get use of the updated wheel.

So this is some more information why I would only add cross compiled wheels used by noarch packages into the python package.

hgy59:
@th0ma7 sorry, I will be more or less offline for the next week...

th0ma7:
Thnx for the explanation.  With this I might have found an alternative... mostly related to pure.

1. What if we add a variable to enable packaging of pure (and set it to **disable** by default).
2. In the `requirements-pure.txt` you include all you normally need (with `homeassistant` you could also include what's in your postinst)
3. At the building process it already generates a `requirements.txt` file and include all the built wheels (by default only the crossenv gets packageed)
4. Then, when the python app gets installed it uses `-r requirements.txt --index-url ${...wheelhousepath}`
5. As only the crossenv wheels are in the directory it installs them and downloads everything else (e.g. pure-python wheels)

This would trim down considerably the package size + reduce build time by a hell of a lot.

I'll merge this PR (as it fixes what it had to fix) and create a new one specifically to address pure-python using `homeassistant` as demo so you can tell me if this makes any sense.

Issue #4820, opened 2021-08-29
Update OpenSSL to 1.1.1l


hgy59:
There are two security issues in OpenSSL 1.1.1k

- CVE-2021-3711 [High severity] 
- CVE-2021-3712 [Moderate severity]

_Linked issues:_  https://github.com/SynoCommunity/spksrc/issues/4211#issuecomment-907757166

### Affected packages that need to be updated
- [x] borgbackup
- [ ] boxbackup-client
- [ ] deluge (indirect dependency: python = python2)
- [ ] domoticz
- [x] duplicity (indirect dependency: python38)
- [x] ejabberd
- [x] erlang
- [ ] ffsync (indirect dependency: python = python2)
- [x] flexget (indirect dependency: python38)
- [x] fossil-scm
- [x] git
- [ ] haproxy
- [x] homeassistant
- [ ] icecast
- [ ] irssi
- [ ] itools
- [ ] jackett
- [ ] memcached
- [x] mercurial (indirect dependency: python38)
- [x] monit
- [x] mosh
- [ ] mosquitto
- [ ] mtproxy
- [x] mutt
- [ ] ntopng
- [x] octoprint (indirect dependency: python38)
- [ ] python
- [x] python3
- [x] python38
- [x] rdiff-backup (indirect dependency: python3)
- [x] ruby
- [ ] rutorrent
- [x] sabnzbd (indirect dependency: python38)
- [x] salt-minion (indirect dependency: python3)
- [ ] shairport-sync
- [x] stunnel
- [x] synocli-disk
- [x] synocli-file
- [x] synocli-monitor
- [x] synocli-net
- [ ] transmission
- [ ] tvheadend
- [x] umurmur (migrated to mbedtls)
- [x] vim (indirect dependency: python38)
- [x] znc (indirect dependency: python38)

BenjV:
@hgy59 Great work!!
This Openssl update is important

hgy59:
outdated, another  CVE (high severity) is fixed with version 1.1.1n (#5161)

PR #4985, opened 2021-11-29
itools: update to libimediadevice 2.0 libraries

hgy59:
_Motivation:_  iOS tools are lacking the latest openssl update and have a 2.0 update for the libraries
_Linked issues:_  #4820

### Checklist

### Remarks
On DSM 7 the mount of connected devices is not supported without ssh into the device with root permissions.
Another issue with DSM 7 is that PythonModule Package by synology is not available.


hgy59:
@bxxxjxxg for DSM 6 we should update the python scripts to Python 3.10 or 3.11. Current versions are still Python 2. Do you have any capacity to do this?

Issue #4875, opened 2021-09-21
syncthing: Web GUI world-accessible without credentials by default


acolomb:
### Setup
_Package Name:_ syncthing
_Package Version:_ v1.17.0

### Expected behavior
The Syncthing Web GUI should not be accessible without any form of authentication right after installing the package.

Proposed solution is to include a new page in the install wizard that asks to choose a user name and password which will be written to the package's `config.xml` file before first service start.

### Actual behavior
Syncthing is accessible from any network interface without any form of authentication.  This is dangerous because with access to the Web GUI, any folder can be configured for sharing by a malicious third-party.  The files themselves may still be protected if the ACL rules are restrictive enough to not grant access to the user / group the syncthing process is running as.

This is based on a user complaint [on the Syncthing forum](https://forum.syncthing.net/t/the-danger-zone-free-files-compliments-of-this-ignoramus/17378/).  The offical project's stance is that such a dangerous default configuration should not be associated with the *Syncthing* brand name.

### Steps to reproduce
_1._ Install Syncthing SPK.
_2._ Point a web browser to the DS port TCP/8384.
_3._ Observe a big fat red warning that having this GUI accessible without authentication is dangerous.

acolomb:
Does anyone know where Synology's UIFILE format is currently documented? The link on the [wiki page](https://github.com/SynoCommunity/spksrc/wiki/UI-Develop) to the developer documentation is broken and so far I have not found any mention in the current (DSM 7.0) version of the Developer Guide.

PR #4995, opened 2021-12-09
syncthing: Ask for Web GUI credentials in installation wizard

acolomb:
The Syncthing Web GUI is currently accessible without any form of authentication right after installing the package, see #4875.  This implements the proposed solution to include a new page in the install wizard that asks to choose a user name and password which will be written to the package's `config.xml` file before first service start.

Note that the upgrade wizard does not include that step, deliberately.

Obviously this will conflict with #4930, so may want to merge both before publishing.  I will resolve conflicts when necessary / appropriate.

### Checklist


publicarray (approved):
LGTM. I just can't compile/test  on my NAS at the moment

PR #5018, opened 2021-12-22
fix PLIST for filenames with spaces

hgy59:
_Motivation:_  packages with files that have spaces in name can't be built.
_Linked issues:_  

### Remarks
Found this errors while working on a package for `Ombi` that contains a file `logo original.png`.
Got two errors:
- copy_target that uses tar failed
- When using PLIST.auto there are other errors

As latest stable ombi is still on .NET 5 and current versions with .NET 6 are prerelease (and didn't run), I want to publish this fix now.


th0ma7:
I'll have a look at it within a day or two.

th0ma7 (approved):
I haven't tried it per say but changes looks good (and nice additions actually).

PR #4998, opened 2021-12-13
Migrate remaining python3 packages to python310

th0ma7:
_Motivation:_  Migrate remaining python3 (3.7.x) and python38 packages to python310.
_Linked issues:_  Follow-up to #4986

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

### To migrate
- [x] `bazarr` - on top of #4992 - Tested: `kvmx64-7.0`, `armv7-7.0`, `armv5tel-6.1`
- [x] `fishnet` - Tested: `kvmx64-7.0`, `armv7-7.0`
- [x] `plexpy-custom` - Tested: `kvmx64-7.0`, `armv7-7.0`, `armv5tel-6.1`
- [x] `rutorrent`  *** <s>requires migrating to DSM7</s> update to DSM7 for a later PR ***  - Tested: `x64-6.1`
- [ ] `salt-minion` - update on-going #4845 - Moved to PR #5017

PR #5023, opened 2021-12-26
Wheel building: provide additional arguments

th0ma7:
_Motivation:_  Update to python wheel framework to enhance building usecases, mainly for `Pillow` and `numpy` but can be used for many other specificities needed by wheels through pip.  The end-state goal is to eliminate when ever possible the usage of `cross` to privilege `pip` with `crossenv` instead to follow more rigorously the default python build process.  Also (and most importantly), this provide more flexibility to each application to specify its own version instead of relying on the same `cross/*`.

In order to do so it needed a way to allow passing "per wheel" arguments and build FLAGS.  As such this PR does the following:
1. Cross-compile one wheel at a time instead of parsing the entire `requirements-crossenv|abi3.txt` at once.
2. (using item 1.) allow passing "per wheel" specific arguments to `pip`.  A good example is for `Pillow`.
3. (using item 1.) allow passing `CFLAGS`, `LDFLAGS`, `CPPFLAGS` and `CXXFLAGS` per wheel flags when cross-compiling
4. Pass `CPPFLAGS` and `CXXFLAGS` which where missing from the `pip` cross-compiling environment, needed for `numpy` for instance
5. Review and integrate code that used to be recopied over and over for `crossenv`, cross without `crossenv` and `abi3`.  Now it uses a single code-block to handle them all, eliminating code duplication.

_Linked issues:_  #4998

### numpy
- For it to work properly with Python 3.10 it requires the yet to be release v1.22.0rc3.
- There currently is an issue with DSM6 gcc <= 4.x, mainly with `arch-88f6281` related to `-std=c++11`
- Reference: https://github.com/numpy/numpy/issues/20664

### Pillow
- There where path issues resulting in a need to maintain a patch along with inability to directly build using `pip` (https://github.com/SynoCommunity/spksrc/blob/master/cross/pillow/patches/001-fix-paths.patch)
- Changes to the framework allow using a `WHEELS_BUILD_ARGS` that is being used to pass arguments to `pip` (renamed as `BUILD_ARGS` when building).  An example was provided by @radarhere https://github.com/radarhere/spksrc/commit/256770199a4e65cfe8a0b4effb706f4dfda37adf
- Reference: https://github.com/python-pillow/Pillow/issues/5799

### Checklist
- [x] Documentation in the wiki


th0ma7:
@hgy59 and @publicarray and others, the following is ready for review.  The high-level description of the PR tells it all.  I've also adapted the wiki accordingly.

Important to note that the enablement of wheel for python310 is only for having a full-test build on github-action.  They are to be re-disabled prior to merging.

PR #5025, opened 2021-12-29
WIP: Docker Raddar/Sonarr/Bazarr stack

publicarray:
_Motivation:_ With DSM 7.0 we can create docker packages and one combination that appears to be asked for often is the *arr stack. This early days but even if this doesn't work it can be used as a stepping stone to build other docker packages.

One behaviour I don't like about Synology's implementation even when I tag specific image versions, Docker still removes all images and redownloads them on a package update/install. This is just a bit over 1GB every time.

@Bokkoman thanks for your feedback and in general for the docker-compose template/guide

### Checklist
- [x] New installation of package completed successfully  (@Bokkoman reports it fails)
- [ ] Usability Tests: 
  - [x] Raddar
  - [x] Sonarr
  - [x] Bazarr
  - [ ] Plex - can't authorise, probably because I got the syntax wrong for the allowed local IPs
                - Plex claim is useless since it expires in 4 minutes, and it takes docker significantly longer to have everything up by then.


publicarray:
@th0ma7 would you be willing to give the package a test run ? I don't know why the installation fails for @bokkoman. Honestly with this package almost everything is handled by Synology's worker, so I have little insight other that It might be a network problem.

th0ma7:
I'll have a look at it within a day or two.

spk/demoservice/Makefile
```diff
@@ -38,7 +38,7 @@ PRE_COPY_TARGET = demoservice_pre_copy
 
 include ../../mk/spksrc.spk.mk
 
-.PHONY: demoservice_install
+.PHONY: demoservice_pre_copy
```
th0ma7, line 41: Good catch.  While at it, might be good creating a demodocker service?

bokkoman: Is it working for you?

th0ma7:
First off, really well done, I really like this.  Note that I'm not a *arr user although perhaps I should give it a try as this seems like nice pieces of code that could prove to be quite useful.

A few things I've noticed:
- without going into a wordsmithing look of things, I'm not convinced by the naming.  I would suggest perhaps adding "docker" in it so we know this is different than a normal DSM-like service?
- logs: there is nothing for many minutes up to "begin postinst".   If there was a was to pass docker logs (using tee perhaps) up to the regular /var/log/packages/*.log would be nice
- storage: is there a shared storage between the 4x dockers?  While trying to add a move I wasn't able to select a root folder within radarr, not even the `/data/media` that looks rw and shared among all containers.
- It is manually possible to create a docker shortcut for each of the running docker images: is it possible to create one automagically at install time?
- Plex: i tried to get to it, perhaps missing the port number or full URL?  For instance I was trying to get to the plex port number but ended-up into an xml file.

A few questions comes to mind:
- Is docker available on all arch supported by DSM7?
- Do they need to be bundled together, or not?  I see the value, it becomes really really easy to get them all running and tuned together this way.  But, is it needed or could the tuning still come with them bundled individually?
- How to maintain them afterwards?
- This ends-up creating a major gap between DSM6 and DSM7.  I could probably do the same and move tvheadend into a docker instance and stop maintaining DSM6 all-together.  It highlight a rather more profound question I guess.

Again, really well done.

publicarray:
Thanks for the valuable feedback 😄 
I agree on all accounts (naming things is hard) yea arr-docker-stack is a better name.

As for storage you need to give the user sc-arr-stack permissions for the docker folder. You also need a specific folder structure e.g `volumeX/docker/arr-stack/storage/movies/movietitle(year)/movietitle.mp4`

Unfortunately AFAIK only x64 devices have docker support. DSM7 is the first to provide this "worker". Previously this would require root access and command line usage with docker.

As for maintaining them that is a good question. I could use the "latest" tag and any manual reinstall will update them. Or we update them like any other package. I haven't thought about it too much yet. I might also ask Synology for more info since I haven't seen anyone make use of the worker yet. I'm missing a few things from docker-compose which this json is a mapping to.

publicarray:
Yea Plex is having issues with me too. And considering that it doesn't have hw acceleration I might remove it and rely on the official package

publicarray:
I'll work on the other stuff you pointed out soon.

bokkoman:
I wonder why it fails for me then.
Or did you create a new package in the meantime?

publicarray:
Hm 🤔, The only difference is the local ip gathering change that wasn't in the last build I gave you. (Last commit)

th0ma7:
Can't find the volume at all... (its on my kvmx64-7.0 vm).
```
root@DSM7:/volume1/@docker# find . -name storage
./plugins/storage
```

publicarray:
Is it possible that the nas is running out of ram?

bokkoman:
It keeps failing for me. I do already have images running from hotio, like Plex. I changed the port, but to no prevail.

publicarray:
> Can't find the volume at all... (its on my kvmx64-7.0 vm).
> 
> ```
> root@DSM7:/volume1/@docker# find . -name storage
> ./plugins/storage
> ```

Ups, Sorry, I called it `docker/media` it's called `/data/media` from inside docker.

publicarray:
> * It is manually possible to create a docker shortcut

No it isn't. One shortcut per package, unfortunately.

bokkoman:
> > Can't find the volume at all... (its on my kvmx64-7.0 vm).
> > ```
> > root@DSM7:/volume1/@docker# find . -name storage
> > ./plugins/storage
> > ```
> 
> Ups, Sorry, I called it `VolumeX/docker/arr-stack/storage/media` and it's linked to `/data/media` from inside docker.
> 
> Screenshot from 2022-01-01 15-20-58

If possible, I suggest putting the `/storage` into `/docker` directly, instead of in the arr-docker-stack.
And then rename it to `/media`.

publicarray:
> If possible, I suggest putting the `/storage` into `/docker` directly, instead of in the arr-docker-stack.
> And then rename it to `/media`.

Not possible with the current Docker worker. Everything must be inside the package name `arr-docker-stack` or in `/var/packages/arr-docker-stack/var` (hidden from the user) the folder is created by Synology

th0ma7:
@publicarray folder is there, I was just looking at the wrong spot :smile:

Issue #5024, opened 2021-12-27
Broken link to pkg for mono 5.8 armada370 arch


piejanssens:
See https://synocommunity.com/package/mono
5.8 armada370 points to this broken link: https://packages.synocommunity.com/mono/11/mono.v11.f15047%5Barmada370%5D.spk

piejanssens:
@hgy59

hgy59:
probably I did somthing wrong when cleaning up outdated versions of mono (as we had no space left on the server).
so I will have to rebuild and publish mono 5.8 

@publicarray are the files like `mono.v11.f15047%5Barmada370%5D.spk` gone on the server?

publicarray:
Yea @hgy59 the file is indeed missing. I tried to compile from 8b01fbbe559094679d7ea48c37b8e8acbdb22d03 but the toolchains had a different name/path

hgy59:
@publicarray I have a label `mono_5-8` on my repo that I will use to rebuild the packages.
And I will use generic archs, so the number of spk files will be less than before...

publicarray:
Thanks!

hgy59:
@piejanssens I have uploaded packages for DSM 6.x for generic x86_64, ARMv7 and aarch64 (armv8)  archs.
The link for your model is https://packages.synocommunity.com/mono/11/mono.v11.f15047%5Balpine-alpine4k-armada370-armada375-armada38x-armadaxp-comcerto2k-monaco%5D.spk

Unfortunatley the toolchains for ARMv5 (88f6281-6.1), ARMv7L (hi3535-6.1) and for DSM 5.x did not work.

hgy59:
@piejanssens I have updated mono 5.8 again (uploads from yesterday are replaced).

As I am currently working on adjustments to build packages for DSM 5.2 again, i have rebuilt the Mono 5.8 with current spksrc environment. With this the ARMv5 and ARMv7L packages are build successfully.

Improvements
- avoid build of static libraries and mcs docs (reduces the spk files from about 140MB to about 80 MB)

hgy59:
And I have successfully tested mono 5.8 with Radarr Radarr Ver. 0.2.0.1293 on DS-215j (armada370).

piejanssens:
Thanks @hgy59, I was able to install this version.

PR #5029, opened 2021-12-30
Minimize mono packages

hgy59:
_Motivation:_  mono packages contain obsolete static libraries 
_Linked issues:_  #5024

### Checklist

### Remarks
This contains the work done to rebuild Mono 5.8 packages (as those were lost in the repo, s #5024).
We can publish the updated Mono 5.20 too, to reduce the size of the published packages.


publicarray:
Nice work! Thanks for the comments & documentation. Roughly how much does this save?

hgy59:
>  Roughly how much does this save?


Mono 5.8
> 
>     * avoid build of static libraries and mcs docs (reduces the spk files from about 140MB to about 80 MB)

Mono 5.20
- from about 150MB to about 90 MB

PR #5047, opened 2022-01-04
ncurses: Update from version 6.1 to 6.3

th0ma7:
_Motivation:_  While looking into various packages I noticed that updates where available for both `gdbm` and `ncurse`.  Both updates are now in separate PR (`gdbm` PR #5050)
_Linked issues:_  N/A

### Checklist


hgy59:
just a hint: when you update `cross/ncurses` you probably want to update `cross/ncursesw` too. Those build from the same source code, the the regular version without and the `w` version with unicode (widechar) support.

th0ma7:
@hgy59 I'm also exchanging with ncurses maintainer as there is what I think is a bug with the location of the `pkgconfig` directory which does not seem to follow properly the `--with-pkg-config-libdir=$(INSTALL_PREFIX)/lib/pkgconfig` as it duplicate the full path under the working directory.

So pending for that fix from upstream before merging (where a temporary patch may be needed until the next update where it will have been mainlined).

th0ma7:
@hgy59 After various exchange with the maintainer upstream I finally found what the caveats are with the handling of the `pkgconfig` directory handling.  I've informed the maintainer upstream of my findings, which in the and can be summarized as needing to re-export `PKG_CONFIG_LIBDIR` so it is relative to the installation path.

I think it's now good and ready for merge but a second look would be appreciated.

hgy59:
currently reviewing...

I'm trying to remove cross/ncurses dependencies (i.e. replace dependency `cross/ncurses` by `cross/ncursesw`) 

- [ ] ejabberd 
- [ ] erlang 
- [x] gnupg  
- [ ] irssi  
- [ ] mosh  
- [x] plowshare  
- [ ] rutorrent  
- [ ] synocli-file  
- [ ] synocli-net  
- [ ] vim

th0ma7:
@hgy59 by rebasing against upstream it seems to have kept the revert for gdbm that I had moved in PR #5050

hgy59:
> @hgy59 by rebasing against upstream it seems to have kept the revert for gdbm that I had moved in PR #5050

Yes, this is kept (i.e. commit is skipped on rebase) as it contains the already merged commit.

hgy59:
I have now migrated all remaining packages to depend on `cross/ncursesw` instead of `cross/ncurses` - except the ruTorrent package, that depends on `cross/procps-ng` that fails to build with ncursesw dependency.

I have tested all those packages for successful build at least for arch-x64-6.1.

hgy59:
Finally your fix for `PKG_CONFIG_LIBDIR` and/or the update of ncurses/ncursesw seem to fix packages that probably could not used ncursesw before. 👏

th0ma7:
Wow, great work!  You solved quite a few out there!  All looks good to me.

Allow me a few cycles so I also migrating rutorrent to `ncursesw`.  Could it then allow renaming ncursesw to ncurses?

hgy59:
> Allow me a few cycles so I also migrating rutorrent to `ncursesw`. Could it then allow renaming ncursesw to ncurses?

I would propose to merge this PR asap, and to migrate rutorrent in a new PR.
If we ever can migrate rutorrent to avoid ncurses or to migrate to ncursesw we shall delete cross/ncurses and keep cross/ncursesw (no renaming). 

Optionally we could now cleanup the cross/ncursesw workaround (temp. download and extract to a renamed folder) as this was only required for synocli-packages depending on cross/ncurses and cross/ncursesw. If only rutorrent depends on cross/ncurses we don't need this workaround anymore.

I already tried to migrate rutorrent. It is `cross/procps-ng` that does not build with ncursesw (and I had no success to patch any file to work around).
You can configure procps-ng without ncurses, but this will skip the build of some of the tools.
I don't know which tools of procps-ng are mandatory for rutorrent.
Some of them are already available under DSM and some are not.

already available tools
- kill
- free
- pkill
- ps
- sysctl
- top
- uptime

not available under DSM 6+
- pgrep (available in cross/busybox)
- pmap (available in cross/busybox)
- pwdx (available in cross/busybox)
- slabtop
- tload
- watch (available in cross/busybox)
- w 

Most of those tools could be supported by cross/busybox, but first we need to know the requirements of rutorrent.

hgy59:
Currently spk/rutorrent already depends on cross/busybox and procutils of busybox support the following tools
- iostat
- lsof
- pgrep
- pmap
- watch

Somwhat confusion why cross/procps-ng was added.

hgy59:
Finally I removed cross/procps-ng and deactivates cross/ncurses.

th0ma7:
Really good job.  Anyway, its known that rutorrent requires more attention to make it available under DSM7 so additional details can be worked out from another PR.  Merging.

PR #5050, opened 2022-01-04
gdbm: Update from version 1.19 to 1.22

th0ma7:
_Motivation:_  While looking into various packages I noticed that updates where available for both `gdbm` and `ncurse`.  Both updates are now in separate PR (`ncurses` PR #5047)
_Linked issues:_  #5046

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

PR #5053, opened 2022-01-05
libjpeg: Update from version 9d to 9e and change URL source

th0ma7:
_Motivation:_  Build failures in #5050 due to changes in libjpeg related to the newer release version 9e (as it seems they also repackaged the older version 9d).  Using the opportunity to update the library.
_Linked issues:_  #5050

### Checklist

### DIFF
--- a/cross/libjpeg/Makefile
+++ b/cross/libjpeg/Makefile
@@ -1,17 +1,16 @@
 PKG_NAME = libjpeg
-PKG_VERS = 9d
+PKG_VERS = 09e
 PKG_EXT = tar.gz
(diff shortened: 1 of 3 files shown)


hgy59:
@th0ma7 it was probably a bad idea to change the download URL for libjpeg, as the source package was updated without changing the file name (bdbae354ebb4a3edce4a6e038436f59a139ad5a8)

th0ma7:
@hgy59 what triggered this is that the *sum was failing for the exact same reason you just pushed a this new PR moments ago.  It seems that both sources are re-deploying the same code with adjusted licensing, thus impacting the tar-ball *sum. :man_shrugging:

Issue #5030, opened 2021-12-30
Jellyfin service doesn't start on DS213j (armada370)


piwi82:
### Setup
_Package Name:_ Jellyfin 
_Package Version:_ 10.7.7-4

_NAS Model:_ DS213j
_NAS Architecture:_ armada370
_DSM version:_ 7.0.1-42218

### Expected behavior
When started, the package status should show status "Installed".

### Actual behavior
The package doesn't start and always shows status "Launch" in Package Center.
No Jellyfin process visible using `ps -aux`

### Steps to reproduce
_1._ Install Jellyfin package
_2._ Launch the package

#### Package log
_Installer log:_ `/var/log/packages/{package}.log`
```
2021/12/30 15:16:14     install jellyfin 10.7.7-4 Begin start-stop-status start
2021/12/30 15:16:14     install jellyfin 10.7.7-4 End start-stop-status start ret=[0]
2021/12/30 15:17:09     start jellyfin 10.7.7-4 Begin start-stop-status stop
2021/12/30 15:17:09     start jellyfin 10.7.7-4 End start-stop-status stop ret=[0]
2021/12/30 15:17:17     start jellyfin 10.7.7-4 Begin start-stop-status start
2021/12/30 15:17:17     start jellyfin 10.7.7-4 End start-stop-status start ret=[0]
2021/12/30 15:21:36     start jellyfin 10.7.7-4 Begin start-stop-status stop
2021/12/30 15:21:36     start jellyfin 10.7.7-4 End start-stop-status stop ret=[0]
2021/12/30 15:21:45     start jellyfin 10.7.7-4 Begin start-stop-status start
2021/12/30 15:21:45     start jellyfin 10.7.7-4 End start-stop-status start ret=[0]
```
#### Other logs
_E.g. `/var/log/messages` or `/var/log/synopkg.log`_
```
[...]

PR #5061, opened 2022-01-07
adjust INFO file to exclude archs that are not supported

hgy59:
_Motivation:_  Until now it is not possible to exclude specific archs from a generic arch
_Linked issues:_  solution for https://github.com/SynoCommunity/spksrc/issues/5030#issuecomment-1007364569

### TODO
- [ ] jellyfin: exclude armada370 and armadaxp from armv7
- [ ] radarr

### DIFF
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -113,6 +113,16 @@ ifneq ($(strip $(SPK_ICON)),)
 include ../../mk/spksrc.icon.mk
 endif
 
+# exclude unsupported archs from SPK_ARCH (for generic archs this is a list).
+SPK_ARCH := $(filter-out $(UNSUPPORTED_ARCHS), $(SPK_ARCH))
+ifeq ($(strip $(SPK_ARCH)),)
+  # packages without arch are invalid.
+  ifneq (,$(BUILD_UNSUPPORTED_FILE))
+    $(shell echo $(date --date=now +"%Y.%m.%d %H:%M:%S") - $(SPK_FOLDER): Arch '$(ARCH)' is not a supported architecture >> $(BUILD_UNSUPPORTED_FILE))
+  endif
+  @$(error Arch '$(ARCH)' is not a supported architecture)
+endif
+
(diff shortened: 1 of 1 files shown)


hgy59:
@th0ma7, @publicarray shall we apply this, to exclude specific archs within generic arch packages?

as discussed in #5030:

- with this update, generic arch packages will not be visible in DSM package center for unsupported archs. But a user can manually download the generic package and will get an "unsupported" error when manually installing the package.

- without this update, the package can be installed (manually or in the package center), but will fail to run on unsupported archs.

hgy59:
Please report packages with incompatible archs that must be excluded.

_I am not sure, whether `armadaxp` and `armada370` must be excluded from `armv7` for  `jellyfin`_

publicarray:
Sounds good to me. 👍

th0ma7:
Will this both "block" from being able to build for a specific arch and publish for that arch?
So it removes also that arch from arch-list passed to the web server when publishing?  Guessing through this:
`SPK_ARCH := $(filter-out $(UNSUPPORTED_ARCHS), $(SPK_ARCH))`

hgy59:
> Will this both "block" from being able to build for a specific arch and publish for that arch?

Yes

> So it removes also that arch from arch-list passed to the web server when publishing? Guessing through this:
> `SPK_ARCH := $(filter-out $(UNSUPPORTED_ARCHS), $(SPK_ARCH))`

Yes, the web server gets all information from the INFO file, there is no other arch-list passed to the web server.

th0ma7:
Sounds good to me as well.  I now wonder how I could use that for ffmpeg to differentiate older intel from new for the driver requirement... food for thoughts I guess.

th0ma7 (approved):

hgy59:
This is the framework enhancement only. Adjustments for jellyfin and other packages shall be provided in dedicates branches.

th0ma7:
@hgy59 yeah fully agree on that.  This was more of a theoretical question, nothing to be added in here otherwise.  cheers.

PR #4734, opened 2021-07-11
syncthing: update to v1.18.0

SheepKid12:
_Motivation:_  Update to current upstream version (v1.18.0).
_Linked issues:_ N/A

### Checklist
- [X] Build rule `all-supported` completed successfully
- [X] Package upgrade completed successfully
- [X] New installation of package completed successfully

PR #5076, opened 2022-01-13
Update syncthing to 1.18.6

miry:
## Description

Update package Syncthing from 1.18.5 to 1.18.6

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update

## Tophat

Tested upgrade on DSM 7.0 Deventon.
Spoted some magick. For some reason I had old version 1.17.0 instead of 1.18.5.
During the Manual installation the old config replaced with new one.
Then I tested uninstall and install again 1.18.6, and this time config remained the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ cd spksrc # Go to the cloned repository's root folder.
 docker run -it -v $(pwd):/spksrc ghcr.io/synocommunity/spksrc /bin/bash
 
 # If running on macOS:
-docker run -it -v $(pwd):/spksrc -e TAR_CMD="fakeroot tar" ghcr.io/synocommunity/spksrc /bin/bash
+docker run --rm --pull=always -it -v $(pwd):/spksrc -e TAR_CMD="fakeroot tar" ghcr.io/synocommunity/spksrc /bin/bash
 ```
 5. From there, follow the instructions in the [Developers HOW TO].
 
--- a/cross/syncthing/Makefile
+++ b/cross/syncthing/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = syncthing
-PKG_VERS = 1.18.5
+PKG_VERS = 1.18.6
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-source-v$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/syncthing/syncthing/releases/download/v$(PKG_VERS)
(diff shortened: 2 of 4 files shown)


hgy59:
please regard that there is already PR #4930 in progress that updates the build of the syncthing package.

miry:
@hgy59 I see changes PR #4930 should not conflict with current one.

miry:
@hgy59 can you help me to understand why in https://synocommunity.com/packages and Synology package list I see Syncthing 1.17.0 instead of current master version 1.18.5?

hgy59:
> @hgy59 can you help me to understand why in https://synocommunity.com/packages and Synology package list I see Syncthing 1.17.0 instead of current master version 1.18.5?

the #4995 is not yet published to https://synocommunity.com/packages. Only some of the maintainers can publish upload an activate) packages to https://synocommunity.com/packages.

As you see in https://github.com/SynoCommunity/spksrc/pull/4734#issuecomment-881779698 we don't creates such updates for syncthing as it uses the internal autoupdate.

For syncthing we only create new releases when synology introduces new package archs or the package specific installation changes.
So we should publish 1.8.5 as the installation has an updated wizard to configure credentials for the web UI.

But the problem is, that when a user updates syncthing with the internal update and we release a package with an updated revision (rev 23 for 1.18.5) the package center will tell that an update is available - but this will downgrade to 1.18.5.

acolomb:
I just stumbled over this PR by accident.  Would be nice to ping me as package maintainer, as I have stopped subscribing every notification from spksrc.

I'd recommend to close this in favor of #4930, where the upgrade to v1.18.6 is already included.  The changes from #4995 have not been published yet, that's also why I didn't bump the SPK_REV further.  #4930 is ready for merging, please do that and then publish.  I will keep pushing new Syncthing versions there until it's published.

A side note: When a user waits long enough after a new SPK version came out, the Syncthing release will *always* have updated to a more recent version than the one which will be installed by the SPK update.  That used to be a problem, but Syncthing upstream changed to better deal with "config file from the future" type errors and now does the internal upgrade check first.  So unless we implement an upgrade check right in the SPK installation scripts, this is always going to happen somewhere and it's not a major concern.  I might just add such a check in the future, using [the `--upgrade` command line option](https://docs.syncthing.net/users/syncthing.html#cmdoption-upgrade).

miry:
@acolomb To make sure that you review PRs related to Syncthing, I propose to update CODEOWNERS and github would assign you as reviewer.

acolomb:
Interesting concept, didn't know about that.  There are some caveats, though: I don't have write permission, so not qualified as owner.  And it just concerns pull requests, while issues regarding this specific package would be equally important to get notified of.

miry:
It is fine to make sure you approved PRs.

PR #5060, opened 2022-01-07
imagemagick: fix and update

hgy59:
_Motivation:_  fix imagemagic load modules and include configure.xml to provide build information
_Linked issues:_  closes #4729

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

PR #5078, opened 2022-01-14
Update and optimize common dependencies

hgy59:
## Description

update freetype and libxml2 and add code optimization to several common libraries.

While building imagemagick with #5060 I found that a lot of common libraries are not built yet with code optimization and a common c/c++ optimize option `"-O"` often reduces the generated code up to 30% (typic. ~10%).
I did not try other optimize options like -O2, -O3 or -Os as we have no experiance with the optimization of gcc in the synology toolchains yet.

This is a follow up to  #5060, where those changes are reverted to avoid triggering the build of all the dependent packages.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


publicarray:
For reference: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

PR #5080, opened 2022-01-16
make empty PLIST files in spk folders obsolete

hgy59:
## Description

remove error log of missing PLIST files in spk folders

Fixes # 
encouraged by @ymartin59 in discord
_Another topic: when reviewing packages, I find it "stupid" to create so many empty "PLIST" files - I guess the "makefiles framework" may be improved to adapt to the lack of PLIST file (and consider it as empty) - any volunteer to propose such improvement?_

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5048, opened 2022-01-04
Borgbackup v1.2.0 update

th0ma7:
_Motivation:_  Progressing towards releasing borgbackup using python310.
_Linked issues:_  https://github.com/SynoCommunity/spksrc/commit/72866f653fcff3f6862a417545a070f618c2d907#diff-7b228d5ca562f5bc3b0ebfbbe98d851ed882feab3721fbb04b57ad844256eb78
_Upstream issue:_ https://github.com/borgbackup/borg/issues/6105

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

PR #5092, opened 2022-01-22
avoid error output for packages without wheels

hgy59:
## Description

1. This is another fix to suppress obsolete build error output (like #5080) 
2. Ensure it exit with error when failure to build a wheel.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.wheel.mk
+++ b/mk/spksrc.wheel.mk
@@ -108,7 +108,7 @@ ifneq ($(strip $(WHEELS)),)
 	      $${abi3} \
 	      $${global_option} \
 	      --no-build-isolation \
-	      $${wheel} ; \
+	      $${wheel} || exit 1 ; \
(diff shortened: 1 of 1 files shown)


hgy59:
@th0ma7 as you are our python master, can you please verify that nothing is broken with this PR.

th0ma7:
@hgy59 I was hoping not to get such a title :facepalm: 
But I will happily have a look at it over the week-end

th0ma7 (approved):
Re-reading the patch, you simply more the `$(WHEEL)` verification at make level instead of using the shell code which solve the issue found.

Looks good to me (although I haven't tested it).

Also note that I removed the `|| true` at the end of the while loop so it does fail when a wheel does not build.

th0ma7:
Now tested against #5048 (using borgbackup 1.1.17 which fails on ppc) where it now properly exited with error on `qoriq` with the added change.  Looks good & merging.

th0ma7:
A successful run with as expected a failure against `qoriq` https://github.com/SynoCommunity/spksrc/actions/runs/1733975420

PR #5095, opened 2022-01-23
synokernel-cdrom: Package basic cdrom kernel modules

th0ma7:
## Description

Changes to spksrc framework related to kerenel modules & kernel builds along with providing a basic cd-rom kernel modules package example to demonstrate howto (original request from @gingerbeardman).  Once this is in I intent to also provide, usb-serial, usb-wifi and linuxtv dvb modules (based on pending PR I still maintain in my github).  Another user @erdoukki is also looking to using this for IPSet kernel modules.

Concept:
1. Provide basic kernel module package that user can then `insmod` and `rmmod` as they see fit.  Later down the road I may provide a helper script for people to use based on https://github.com/SynoCommunity/spksrc/pull/4420/files#diff-5d8e6e0ccc784c3a275658aad0fde11c873b122397414e4fa198ac51c219d99e)
2. When creating a kernel module package must be provided all DSMs sub-versions modules into that same package so it is compatible with all of them (e.g. a x64-DSM6.1 package contains modules built for all DSM 6.x sub-archs, same idea applies for DSM7).  In order to do so the framework must "loop" on all sub-arch available kernels, thus requiring additional changes to the framework.
3. (#5156) Due to the needed changes at no. 2. the github-action now uses `make arch-*` command line instead of `make TCARCH ...`.  There are corner-cases where builds does not exit with proper exit error signal.  After further looking into it this appear to be due to the `tee` pipe where the exit code is from the later command.  The way to properly handle this is through the `${PIPESTATUS[0]}` variable which provides the output from the first command from the pipe sequence.  In extra to that, all instances of `-@$(MAKE)` have been replaced with `@$(MAKE)` to allow passing errors + an additional workaround is being used in this case is to check if the resulting package has been created or not.  Demonstration of functionality on this can be observed through https://github.com/SynoCommunity/spksrc/pull/5095/files#diff-13feef9fa229fc052f6a5beb42cbdb780aef73b35f98395bc4bfa10839b75f6f (to be removed before merge)
4. Added the ability to pre-download kernel sources using github-action.  This in turns needed a few additional framework changes:
4.1 a new dependency listing function such as `make TCVERSION=6.1 dependency-kernel-list` where it now creates a list of dependency kernels when `KERNEL_REQUIRE_MODULE` is set.  This allows to pre-download all the necessary kernel sources along with specific toolchains so build do not fail due to too many connection on sourceforge later-on.   To do so I needed to add new variables to determine which kernels are available and supported (`AVAILABLE_KERNEL`, `AVAILABLE_KERNEL_VERSIONS`, `SUPPORTED_KERNEL_VERSIONS`).  Also using pre-determined `SUPPORTED_KERNEL_VERSIONS` as otherwise there simply are too many kernels to support and loop over for some archs such as `x64-6.x` and `armv7-6.x`.
4.2 (#5157) ability to download "only" the necessary toolchains: by default the download toolchain is always extracted.  This in turns takes a lot of disk space making github-action to fail due to lack of storage.
4.3 (#5159) added the ability to retry downloads using wget by adding `--timeout=30 --waitretry=0 --tries=5` options.
5. (#5153) `kernel` & `toolchain` header misc adjustments such as adding generic `kernel` archs for `armv7` and `aarch64`, removed non-existent `qoriq-7.0`, added missing `broadwell-7.0 and splitted `alpine` and `alpine4k`.

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update


.github/actions/build.sh
```diff
@@ -58,8 +58,8 @@ do
 
     if [ "${GH_ARCH%%-*}" != "noarch" ]; then
         # use TCVERSION and ARCH to get real exit codes.
-        echo "$ make TCVERSION=${GH_ARCH##*-} ARCH=${GH_ARCH%%-*} -C ./spk/${package}" >>build.log
-        make TCVERSION=${GH_ARCH##*-} ARCH=${GH_ARCH%%-*} -C ./spk/${package} ${MAKE_ARGS} |& tee >(tail -15 >>build.log)
+        echo "$ make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package}" >>build.log
+        make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package} ${MAKE_ARGS} |& tee >(tail -15 >>build.log)
```
hgy59, line 62: Sadly this will not work. My experience was that only `make TCVERSION=...  ARCH=...` returns the correct exit status as my comment above indicates.

th0ma7: Then I'll have to see what can be done to alleviate this... My guess is that `-@MAKEFLAGS= ` invokes a sub-shell and thus you just can't track process status.  There may be something to do about it.

Here's the situation (this really has been a very long-term goal and I think I'm now really close):
- Objective: Provide kernel module packages (cdrom, usb-dvb, usb-wifi, etc).
- Howto: Provide DSM6 and DSM7 packages that contains modules for all corresponding kernel versions such as, a DSM-6.1 `aarch64` package contains:
```
lib/modules/armada37xx-6.2.2/
lib/modules/armada37xx-6.2.2/4.4.59+/
lib/modules/armada37xx-6.2.2/4.4.59+/drivers/
lib/modules/armada37xx-6.2.2/4.4.59+/drivers/cdrom/
lib/modules/armada37xx-6.2.2/4.4.59+/drivers/cdrom/cdrom.ko
lib/modules/armada37xx-6.2.2/4.4.59+/drivers/scsi/
lib/modules/armada37xx-6.2.2/4.4.59+/drivers/scsi/sr_mod.ko
lib/modules/armada37xx-6.2.3/
lib/modules/armada37xx-6.2.3/4.4.59+/
lib/modules/armada37xx-6.2.3/4.4.59+/drivers/
lib/modules/armada37xx-6.2.3/4.4.59+/drivers/cdrom/
lib/modules/armada37xx-6.2.3/4.4.59+/drivers/cdrom/cdrom.ko
lib/modules/armada37xx-6.2.3/4.4.59+/drivers/scsi/
lib/modules/armada37xx-6.2.3/4.4.59+/drivers/scsi/sr_mod.ko
lib/modules/rtd1296-6.1/
lib/modules/rtd1296-6.1/4.4.15+/
lib/modules/rtd1296-6.1/4.4.15+/drivers/
lib/modules/rtd1296-6.1/4.4.15+/drivers/cdrom/
lib/modules/rtd1296-6.1/4.4.15+/drivers/cdrom/cdrom.ko
lib/modules/rtd1296-6.1/4.4.15+/drivers/scsi/
lib/modules/rtd1296-6.1/4.4.15+/drivers/scsi/sr_mod.ko
lib/modules/rtd1296-6.2/
lib/modules/rtd1296-6.2/4.4.59+/
lib/modules/rtd1296-6.2/4.4.59+/drivers/
lib/modules/rtd1296-6.2/4.4.59+/drivers/cdrom/
lib/modules/rtd1296-6.2/4.4.59+/drivers/cdrom/cdrom.ko
lib/modules/rtd1296-6.2/4.4.59+/drivers/scsi/
lib/modules/rtd1296-6.2/4.4.59+/drivers/scsi/sr_mod.ko
lib/modules/rtd1296-6.2.2/
lib/modules/rtd1296-6.2.2/4.4.59+/
lib/modules/rtd1296-6.2.2/4.4.59+/drivers/
lib/modules/rtd1296-6.2.2/4.4.59+/drivers/cdrom/
lib/modules/rtd1296-6.2.2/4.4.59+/drivers/cdrom/cdrom.ko
lib/modules/rtd1296-6.2.2/4.4.59+/drivers/scsi/
lib/modules/rtd1296-6.2.2/4.4.59+/drivers/scsi/sr_mod.ko
lib/modules/rtd1296-6.2.3/
lib/modules/rtd1296-6.2.3/4.4.59+/
lib/modules/rtd1296-6.2.3/4.4.59+/drivers/
lib/modules/rtd1296-6.2.3/4.4.59+/drivers/cdrom/
lib/modules/rtd1296-6.2.3/4.4.59+/drivers/cdrom/cdrom.ko
lib/modules/rtd1296-6.2.3/4.4.59+/drivers/scsi/
lib/modules/rtd1296-6.2.3/4.4.59+/drivers/scsi/sr_mod.ko
```

This mean that package is kernel-binary compatible to what-ever DSM you are running (except 6.2.4 which hasn't been released just yet, but in many cases previous version will still work).  All is left is for you to insmod the kernel-corresponding modules as you require.

But for this to work I adapted `spksrc.spk.mk` in order to loop over the sub-archs for generic packages.  There is no way I can do this otherwise which mean that current github-action calls must be converted to spksrc default make behavior for binary packages, at the very least for generic archs.

At first glance I see two options:
1. fix the underlying issue so all calls can be converted to `make arch-<arch>-<version>`
2. add a check for kernel module compiling and only then use `make arch-<arch>-<version>` calls.

th0ma7:
@hgy59 after a few readings it seems that it turns out that shell exit code within makefile are not being propagated to top-level makefile.  There are probably ways to circumvent this such as using `.SHELLSTATUS` but you where right, it's not that obvious to fix (something else on my todo list).

Still, something obvious caught my attention: the end-result.  For a build to succeed a `.spk` must also be generated.  I've added that to the github-action as a workaround and it seems to be working just fine.

Review on this PR would be appreciated, cheers.

erdoukki:
The build check failed on a out of topic error, on another package, with a pip version check reported error.

I wait for this PR merge to propose an IPSet package.
With kernel module, lib, and ipset util.

.github/actions/build.sh
th0ma7: @hgy59 all issues seems to be fixed now with the extra workaround I added (e.g. checking file result existence).  Besides adding an icon to the package I believe this is ready for your review (as there are a few framework changes).  Thnx in advance.

th0ma7:
> The build check failed on a out of topic error, on another package, with a pip version check reported error.

Thnx for reporting, the error is fully intended to demonstrate the change of error handling of exceptions.  It is expected to be reverted before merge.

gingerbeardman:
I need to try again to get this working.

th0ma7:
@hgy59 and @ymartin59 I'll be splitting this PR into smaller and less invasive ones and track them through this PR description.

th0ma7:
@hgy59 and @ymartin59 I've managed to split this PR into more manageable sub-PR.  Now really this should make a review much easier to do.  Your eye on this would be much appreciated, thnx in advance.

timrettop:
I tried this out to build a module and using `make all-supported` fails on x64-7.0 due to that kernel source not being available.  Is the solution simply to add x64-7.0 to the UNSUPPORTED_ARCHS in the Makefile, or is this a bug?

th0ma7:
> I tried this out to build a module and using `make all-supported` fails on x64-7.0 due to that kernel source not being available. Is the solution simply to add x64-7.0 to the UNSUPPORTED_ARCHS in the Makefile, or is this a bug?

@timrettop for your personal usage `make all-supported` is really not the way to go for you as it builds all arches for various DSM versions (in this case only 7.0 though...).  This also means downloading tons of irrelevant kernel sources and possibility for it to break along the way.

If really you want to build something for your own use you should rather use `make arch-<yourarch>-7.0` instead.  For instance I have a `apollolake` DS918+:
```
$ uname -a
Linux th0ma7-nas 4.4.59+ #25556 SMP PREEMPT Sat Aug 28 02:17:26 CST 2021 x86_64 GNU/Linux synology_apollolake_918+
```
I can just use `make arch-apollolake-7.0` and I'll end-up with a package containing only modules for my own NAS.  Hope this helps.

th0ma7:
@SynoCommunity/developers this is finally ready for review and prime time release.  This will lead into releasing more kernel module packages namely for: usbserial, usbwifi and finally linuxtv usb dvb.

All predecessor PR have already been merged over time in smaller bits, this really is the final piece.  Unless anyone sees a blocker with somthing in here I'm planning to merge this over the coming days.

Lastly, https://github.com/SynoCommunity/spksrc/wiki/FAQ-SynocliKernel-(usbserial,-linuxtv) will need to be fully reviewed as there where many small changes over time and the first kernel package is `cdrom` and not `usbserial` nor `linuxtv`.

spk/synocli-kernel/src/synocli-kernelmodule.sh
```diff
@@ -0,0 +1,347 @@
...
+   do
+      # first check if the name
+	  # matches to a path
+      if [ ! "$(echo $ko | grep '/')" ]; then
+         # Ensure to add .ko if needed
+         [ ! "$(echo $ko | grep '.ko$')" ] && ko=$ko.ko
+		 # Replace any '-' or '_' by '[-_]
+		 ko=$(echo ${ko} | sed -r 's/[-_]/[-_]/g')
+         # Find full module kernel object path
+         ko=$(find $KPATH -name $ko | awk -F'drivers/' '{print $2}')
```
timrettop, line 82: Why presume that the module will be in the drivers/ subdirectory?  Shouldn't this also support other kernel module paths (such as net/ etc? 
Line 136 also references it.

th0ma7: Excellent point, I've updated the code to be include one additional directory level to the variable, which in turn leave's room for `net/` or any other.

erdoukki:
Thanks you @th0ma7 for this great PR.
I was waiting for long time for it.
😉

th0ma7:
@erdoukki I created supporting documentation here https://github.com/SynoCommunity/spksrc/wiki/FAQ-SynocliKernel-(usbserial,-linuxtv)

And for your awareness, this is to be loaded on an original Synology kernel.  I don't know if Xpenology kernels are different or not (I might give it a try in a VM to check out).

PR #5096, opened 2022-01-23
homeassistant: Migrate to ujson & numpy using pip

th0ma7:
## Description

Pursuing work on migrating all python dependencies to regular `pip` building.  In this specific case trying to solve two issues:
1. `ujson` which in turns requires `setuptools-scm` to be added to the `crossenv`.  Hint provided by upstream through https://github.com/ultrajson/ultrajson/issues/496
2. `numpy` to standardize it using `pip` build process instead of `cross/numpy`.  Note that ARMv5 still require code fix https://github.com/numpy/numpy/issues/20664 but support for such old compiler is expected to end real soon.

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5099, opened 2022-01-25
Auto-handling of the gcc version through TC_GCC

th0ma7:
## Description

While working on a build issue with `comcerto2k` on DSM7 where it uses an older gcc version than other archs (ref. #5096), I thought of removing the manually provided `TC_GCC` to the benefit of automating it.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


th0ma7:
@hgy59 if I recall you're the one that last played with this code.  I came accros this and thought that would eliminate any remaining potential false-positives.

Note that I tried finding something similar for GLIBC but that ain't as easy as I thought.  That may be for another time.

ymartin59:
I wonder if it would be "enough" to retrieve information from `gcc -dumpversion` dynamically?

th0ma7:
@ymartin59 its exactly that:
```
	  if [ "$${target}" = "cc" ] ; then \
	    gcc_version=`eval $$(echo $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)$${source} -dumpversion)` ; \
	  fi ; \
```

th0ma7:
@hgy59 and @ymartin59 this is ready for merge.  I've rebased against master and recent changes on the toolchain & kernel.  This really is a relatively trivial change that removes all `TC_GCC` in favor of calling `gcc -dumpversion`, removing both cruft in our configurations and false-positives.

Before merging a quick review from on of you would be appreciated.

hgy59 (approved):
LGTM

was just wondering whether it would be possible to get the glibc version from the executables too.

(tried https://stackoverflow.com/questions/15134548/which-glibc-version-is-using-my-toolchain and https://stackoverflow.com/questions/9705660/check-glibc-version-for-a-particular-gcc-compiler but didn't find the matching version).

th0ma7:
I also tried but failed miserably in doing so for GLIBC... perhaps in a subsequent PR.
Thnx for your review.

Issue #4702, opened 2021-06-29
Mosquitto update deletes old config :-(


hmueller01:
### Setup
_Package Name:_ Mosquitto
_Package Version:_ 1.6.15-12

_NAS Model:_ RS816
_NAS Architecture:_ armada38x
_DSM version:_ 6.2.4-25556

### Expected behavior
Updating mosquitto from 1.6.12-11 should keep old config (/volume1/@appstore/mosquitto/etc/mosquitto) files!

### Actual behavior
The folder is deleted and replaced by a default folder and all previous settings are lost (conf, user, ca-files, ...)!
This is really bad, as loosing the ca_certificates is a disaster!!! :-( 

### Steps to reproduce
Press update in package manager ...

#### Package log
_Installer log: find logfile `/var/log/packages/{package}.log`, for DSM<6: `/var/packages/{package}/target/var/{package}_installer.log` and protocol in Package Center.
You find service log and other log files in `/var/packages/{package}/target/var`.
_
```
2021/06/29 15:27:40     upgrade mosquitto 1.6.12-11 Begin start-stop-status stop
Stopping mosquitto ...
2021/06/29 15:27:41     upgrade mosquitto 1.6.12-11 End start-stop-status stop ret=[0]
2021/06/29 15:27:41     upgrade mosquitto 1.6.12-11 Begin preupgrade
2021/06/29 15:27:41     ===> Step preupgrade. USER=mosquitto GROUP= SHARE_PATH=
2021/06/29 15:27:42     upgrade mosquitto 1.6.12-11 End preupgrade ret=[0]
2021/06/29 15:27:42     upgrade mosquitto 1.6.12-11 Begin preuninst
2021/06/29 15:27:42     upgrade mosquitto 1.6.12-11 End preuninst ret=[0]
2021/06/29 15:27:42     upgrade mosquitto 1.6.12-11 Begin /bin/rm -rf /volume1/@appstore/mosquitto
2021/06/29 15:27:42     upgrade mosquitto 1.6.12-11 End /bin/rm -rf /volume1/@appstore/mosquitto ret=[0]
2021/06/29 15:27:42     upgrade mosquitto 1.6.12-11 Begin postuninst
2021/06/29 15:27:42     upgrade mosquitto 1.6.12-11 End postuninst ret=[0]
2021/06/29 15:27:42     upgrade mosquitto 1.6.15-12 Begin preinst
[...]

PR #5117, opened 2022-02-03
[framework] syncthing: fix DSM7 package upgrade

### LINKED ISSUE #5116: syncthing: URGENT, package update deletes user config on DSM7.
### Is this a new Bug?

### Package Name

syncthing

### Package Version

v1.19.0-24

### Device Model

DS218j and others

### Device Architecture

AArch64 (ARMv8)

### Firmware Version

DSM 7.0.1

### What happened?

Reported on the [Syncthing forum](https://forum.syncthing.net/t/synocommunity-syncthing-package-update-1-19-0-24-for-synology/17961):

> Just registered to report that the package update above appears to have wiped the Syncthing config (other than the device ID) on my DS920+ running DSM7 and I have to recreate the shares etc. A word of caution to other users.
> 
> Just to clarify, Syncthing’s internal update mechanism had updated the binary to version 1.19.0 without any issue but a subsequent package update published by SynoCommunity and available from within DSM’s package center caused the problem.

PLEASE UNPUBLISH THE v1.19.0 UPDATE ASAP until this has been investigated.

### Reproduction steps

```bash
1. Update from previous SPK version v1.17.0-22 on DSM 7
2. Find the config file missing.
3. Have to recreate everything manually
```

### Install Log

```log
Dunno
```

### Service Log

_No response_

### Other Logs

_No response_

hgy59:
## Description

Fix DSM7 installer
- avoid overwrite of existing files in var folder
- fix recognition of dsm6 var folder

Fixes #5116

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.service.installer.dsm7
+++ b/mk/spksrc.service.installer.dsm7
@@ -4,7 +4,6 @@
 # - Not supported
 # DSM 6 -> 7 upgrade path:
 # - files are migrated from ${SYNOPKG_PKGDEST}/var to ${SYNOPKG_PKGVAR}
-# - Not supported (yet)
 
 # installer log is not writable, use for reference only.
 INST_LOG="/var/log/packages/${SYNOPKG_PKGNAME}.log"
@@ -69,23 +68,29 @@ call_func "initialize_variables"
 ### Functions library
 
 #
-# New DSM7 folders: copy no-verwrite + remove source
-# (@appdata) /var/packages/<package>/target/var -> /var/packages/<package>/target/../var
+# syncronize all files from var folder in spk (extracted to target/var)
+# (@appdata) /var/packages/<package>/target/var -> /var/packages/<package>/var
 #
 syno_sync_var_folder () {
+    # take all files that do not alredy exist and rename the other files to *.new to avoid overwriting
+    # existing configuration files.
+    # finally remove the target/var folder that is not used under DSM 7.
     if [ -d ${SYNOPKG_PKGDEST}/var -a "$(ls -A ${SYNOPKG_PKGDEST}/var 2>/dev/null)" ]; then
-        # Copy and delete all files except don't overwrite any existing
+    
+        echo "Install files from var folder"
+    
+        # Move all files except existing
         echo "$RSYNC --ignore-existing --remove-source-files ${SYNOPKG_PKGDEST}/var/ ${SYNOPKG_PKGVAR}"
         $RSYNC --ignore-existing --remove-source-files ${SYNOPKG_PKGDEST}/var/ ${SYNOPKG_PKGVAR}
 
         # Rename any remaining (thus pre-existing) files with .new suffix
         find ${SYNOPKG_PKGDEST}/var -type f -exec sh -c 'x="{}"; mv "$x" "${x}.new"' \;
 
-        # Force-copy and delete all .new files
+        # Move .new files (overwrite existing)
         echo "$RSYNC --remove-source-files ${SYNOPKG_PKGDEST}/var/ ${SYNOPKG_PKGVAR}"
         $RSYNC --remove-source-files ${SYNOPKG_PKGDEST}/var/ ${SYNOPKG_PKGVAR}
 
-        # Remove remaining directory
+        # Remove var folder extraced from spk
         $RM ${SYNOPKG_PKGDEST}/var
     fi
 }
@@ -198,12 +203,18 @@ preupgrade ()
     call_func "validate_preupgrade"
 
     # dsm6 -> dsm7
-    # Copy first to temporary directory if upgrading from DSM6 -> DSM7
-    # Then migrate data once to permanent storage at postupgrade
-    if [ -d ${SYNOPKG_PKGDEST}/var -a "$(ls -A ${SYNOPKG_PKGDEST}/var 2>/dev/null)" ]; then
-        echo "$RSYNC ${SYNOPKG_PKGDEST}/var/ ${TMP_DIR}" | install_log
-        $RSYNC ${SYNOPKG_PKGDEST}/var/ ${TMP_DIR} 2>&1 | install_log && chk=$?
-        [ "${chk}" = "0" ] && rm -fr ${SYNOPKG_PKGDEST}/var || true
+    if [ -d ${SYNOPKG_PKGDEST}/var ]; then
+        if [ ! -d ${SYNOPKG_PKGVAR} -o $(realpath ${SYNOPKG_PKGVAR}) = $(realpath ${SYNOPKG_PKGDEST}/var) ]; then
+            # Copy to temporary directory if upgrading from DSM6 -> DSM7
+            # Then restore once to permanent storage at postupgrade
+            if [ -d ${SYNOPKG_PKGDEST}/var -a "$(ls -A ${SYNOPKG_PKGDEST}/var 2>/dev/null)" ]; then
+                echo "Backup target/var folder used under DSM6" | install_log
+                echo "$RSYNC ${SYNOPKG_PKGDEST}/var/ ${TMP_DIR}" | install_log
+                $RSYNC ${SYNOPKG_PKGDEST}/var/ ${TMP_DIR} 2>&1 | install_log
+                # remove DSM6 target/var folder
+                rm -rf ${SYNOPKG_PKGDEST}/var 2>&1 | install_log
+            fi
+        fi
     fi
 
     call_func "service_preupgrade" install_log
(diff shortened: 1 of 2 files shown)


hgy59:
@th0ma7 with #4997 you introduced rsync calls to migrate the var folder of spk to the DSM7 var folder. This works as expected except for upgrade of packages created before.

Without this fix, the target/var folder (that was not removed in packages created without the rsync migration) was taken as dsm6 folder and existing config files were overwritten.
As it is not possible to modify already installed packages, the recognition of dsm 6 installation cannot only look for target/var folder but must check whether {package}/var is linked to target/var.

DSM 7 packages created before the rsync solution (like syncthing 1.17.0-22 created at 2021-06-19) do not delete the target/var folder.

hgy59:
@acolomb DSM 7 package updates are working now. I did intense testing under DSM 7 (including DSM 6 update simulation by recreation of former folder structure).

I am just wondering whether is was a good idea to create the sync folders in the package var folder (home). Before DSM7 this folder is backuped an restored at every package upgrade and it is removed when you uninstall the package.

A better approach would be to use a DSM shared folder.
- shared folders are never removed on uninstallation of a package 
- when you try to remove a shared folder in DSM, you get an error telling that the shared folder is in use by the respective package(s).
- you can give dedicated permissions to DSM users for the folders in such a share.
- on DSM < 7 there is no need to backup and restore the folders upon synocommunity package update (no issue with syncthing internal update).

acolomb:
Thanks @hgy59 for looking into it.  Sounds reasonable to me so far.

I don't know what you mean exactly by "creating sync folders".  Syncthing can be pointed at ANY existing folder path on the file system.  It is not limited to it's "own" sharing volume / folder or whatever.  Contrary to e.g. Dropbox where all shared stuff must live under one folder.  So if you mean the *default* location where new shared folders are suggested to be created, that should be pointed at a reasonable location.  I don't have enough overview of DSM to judge where that would be.

hgy59:
> So if you mean the _default_ location where new shared folders are suggested to be created, that should be pointed at a reasonable location. I don't have enough overview of DSM to judge where that would be.

Thanks for this information. For my tests I created two folders and for me it looked like the folder path is not editable.
I just created a new DSM shared folder and gave R/W access for the sc-syncthing user. This is working like a charm. So as you said, you have no limitations to choose the sync folders. 😄

spk/syncthing/Makefile
```diff
@@ -24,7 +24,6 @@ SERVICE_SETUP = src/service-setup.sh
 SERVICE_PORT = 8384
 SERVICE_PORT_TITLE = $(DISPLAY_NAME)
 SERVICE_CERT = syncthing_webui
-SERVICE_CERT_RELOAD = tools/ca_reloader.sh
```
acolomb, line 27: I think that variable is used under DSM 6 for setting up the certificate management.  I see why you removed it in the lines below, but for the other part we still need it.  Look in `spksrc.service.mk`, it's documented there.

spk/syncthing/Makefile
```diff
@@ -12,7 +12,7 @@ MAINTAINER = acolomb
 DESCRIPTION = Automatically sync files via secure, distributed technology.
 DESCRIPTION_FRE = Synchronisation automatique de fichiers via une technologie sécurisée et distribuée.
 DISPLAY_NAME = Syncthing
-CHANGELOG = "Update syncthing to v1.19.0.  Add wizard page to preconfigure credentials for the Web GUI.  Integrate with the certificate management UI in Control Panel."
+CHANGELOG = "Update syncthing to v1.19.0. Add wizard page to preconfigure credentials for the Web GUI. Integrate with the certificate management UI in Control Panel. Fix DSM 7 data migration."
```
acolomb, line 15: I routinely add the double spaces after a full-stop btw. because it's an accepted norm (one of several) in English writing.  And my editor (Emacs) treats it automatically when reflowing text.

acolomb (changes requested):

acolomb:
Damn, my review was 30 seconds too late! @hgy59

Please check out the comment about `SERVICE_CERT_RELOAD` before publishing!

hgy59:
oops I was too fast...

spk/syncthing/Makefile
hgy59: done in #5118

spk/syncthing/Makefile
hgy59: in #5118 I numbered the change log entries, so this gets obsolete.

th0ma7:
> @th0ma7 with #4997 you introduced rsync calls to migrate the var folder of spk to the DSM7 var folder. This works as expected except for upgrade of packages created before.
> 
> Without this fix, the target/var folder (that was not removed in packages created without the rsync migration) was taken as dsm6 folder and existing config files were overwritten with the default files of the former installation. As it is not possible to modify already installed packages, the recognition of dsm 6 installation cannot only look for target/var folder but must check whether {package}/var is linked to target/var.
> 
> DSM 7 packages created before the rsync solution (like syncthing 1.17.0-22 created at 2021-06-19) do not delete the target/var folder.

@hgy59 (late to reply, limited cycles available currently)

I recall testing this throughout but DSM6 to DSM7 isn't an easy thing to fully test.  If I understand this correctly, and this applies to synchting (but may apply elsewhere), after upgrading to DSM7, an existing DSM6 syncthing installation would look like this?
```
/var/packages/synchthing/target -> /volume1/@appstore/synchthing
/var/packages/synchthing/var -> /volume1/@appstore/synchthing/var
```
While a DSM7 package would normally look like (notice the appstore vs appdata):
```
/var/packages/synchthing/target -> /volume1/@appstore/synchthing
/var/packages/synchthing/var -> /volume1/@appdata/synchthing
```
Which could lead at upgrade time that rsync might try to re-sync over itself?  Am I painting the right picture?

... Actually no, as `var` data gets move to a temporary directory it would then overwrite the "new" packaged var content over the new files from the updated package in `target/var`... 

Would that mean:
1. First ensuring that `/var/packages/synchthing/var -> /volume1/@appdata/synchthing` does exists OR creating it otherwise.  Which would mean something such as:
```
if [ -h /var/packages/synchthing/var -a -d $(readlink /var/packages/synchthing/var) ]; then OK else create symlink.
```
2. Secondly, move the temporary backup to the "newly" created destination
3. Move along with the rsync as planned.

(re-reading your post, again)
> dsm6 folder and existing config files were overwritten with the default files of the former installation

Then rewinding for a bit, if that was the expected behavior, then my PR changed that behavior as updated var files do not overwrite pre-existing configurations but rather creates `.new` when a file is pre-existing.  Re-visiting that could would then lead to something like:
1. step 1: confirm var is ready
2. Do the rsync of target/var to var
3. re-use the rsync method but instead of creating `.new` create `.old` instead by moving old configurations over the new ones.

Is there a need for me to review the code change (as already merged)?  If so I can try to find additional spare cycles over the week-end and go through the code.

mk/spksrc.service.installer.dsm7
```diff
@@ -198,12 +203,18 @@ preupgrade ()
...
 
     # dsm6 -> dsm7
-    # Copy first to temporary directory if upgrading from DSM6 -> DSM7
-    # Then migrate data once to permanent storage at postupgrade
-    if [ -d ${SYNOPKG_PKGDEST}/var -a "$(ls -A ${SYNOPKG_PKGDEST}/var 2>/dev/null)" ]; then
-        echo "$RSYNC ${SYNOPKG_PKGDEST}/var/ ${TMP_DIR}" | install_log
-        $RSYNC ${SYNOPKG_PKGDEST}/var/ ${TMP_DIR} 2>&1 | install_log && chk=$?
-        [ "${chk}" = "0" ] && rm -fr ${SYNOPKG_PKGDEST}/var || true
+    if [ -d ${SYNOPKG_PKGDEST}/var ]; then
+        if [ ! -d ${SYNOPKG_PKGVAR} -o $(realpath ${SYNOPKG_PKGVAR}) = $(realpath ${SYNOPKG_PKGDEST}/var) ]; then
```
th0ma7, line 207: Interesting snipet, good catch.

hgy59:
> I recall testing this throughout but DSM6 to DSM7 isn't an easy thing to fully test. If I understand this correctly, and this applies to synchting (but may apply elsewhere),

All packages for DSM 7 are affected, that are created before 15 december 2021 (the date #4997 was merged) and are not updated after this date.

To fully understand what was going wrong, you have to consider the following:
1. only DSM 7 installer is affected
2. the DSM installer extracts the spk content (package.tgz) to the `target` folder. i.e. the var folder in the package goes to `target/var`
3. since  #4997 the `target/var` folder is deleted after sync into the _new_ var folder (those are never the same location under DSM 7)
4. before #4997 the `target/var` folder was not deleted by the DSM 7 installer
5. the temporary directory is used only for DSM6 -> DSM7 migration, the var folder sync goes directly from `target/var` to the _new_ var folder
6. Important: if we find a `target/var` folder in **preupgrade** (2. is not yet performed) it can be a leftover of a DSM 7 installation **or** from DSM 6
7. The sole error was in preupgrade to take the `target/var` that was left from a DSM 7 installer as DSM 6 var folder.
8. The correct recognition whether `target/var` is from DSM 6 in preupgrade fixes it all


_sorry for not responding to your detailed questions_

hgy59:
> Is there a need for me to review the code change (as already merged)? If so I can try to find additional spare cycles over the week-end and go through the code.

no need for review other than for fully understanding what's happening 😄

hgy59:
#4702 is probably realted (but it is for DSM 6)

PR #5126, opened 2022-02-09
Add gotify/server

thorpelawrence:
## Description

Add new package https://github.com/gotify/server. Based roughly on its [build instructions](https://gotify.net/docs/build)

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update


thorpelawrence:
Planned to test on armv7-6.2.4, was able to build `gotify/server` with `GOARCH=arm GOARM=7 go build` but seems like it's not working with this version of gcc, any ideas on how I might be able to fix it? So far haven't been able to build an spk because of this as the toolchain wasn't working with it

hgy59:
@thorpelawrence as x64 successfully builds, I tried to run in under DSM 7 on DS218+.
The server cannot create a data directory in the bin folder and therefore does not run (and there is no way to configure a different data folder).

thorpelawrence:
Thanks for testing, I didn't have an x64 Synology to test. Will see if I can create a patch or upstream change to configure data directory

thorpelawrence:
Thanks for all the help on this @hgy59, unfortunately I haven't been able to get it to compile with `make arch-armv7-6.2.4`, do you know if there's a way to cross-compile with a different toolchain that works with it?

hgy59:
As another approach we could try to use the prebuilt binaries of gotify-server and gotify-cli.

thorpelawrence:
> > Interesting, as far as I could see it was just generating Go code (no CGO dependency) so didn't think `packr` would generate anything platform specific.
> 
> Sorry this was just a guess. The "-m64" errors are a hint that the options are for the host but the gcc used is for the target arch.

I saw the -m64 somewhere in I think the CGO code for ARM builds, so seems like it might be an option that's applied that just doesn't seem to be working with this specific toolchain.

I'd be happy to use the pre built versions, seems like there's an armv7 version - I guess would be OK to just use that for platforms that don't compile?
I was trying to use `make all-supported` but seems like that doesn't exist anymore?

thorpelawrence:
Was able to get it working using the gotify-linux-arm-7 build from GitHub releases, but wasn't completely sure how to get this to work with other platforms properly

image

thorpelawrence:
In fact, didn't even notice but `gotify-cli` was compiled fine, just the issue with `gotify/server` which is what I replaced with the download instead of build. (Makes sense though as it's only Go code in cli, but there's some C code for sqlite inside `gotify/server`)

thorpelawrence:
Was able to get it working with this (quickly hacked together) change, but obviously only for armv7

```diff
diff --git a/cross/gotify-server/Makefile b/cross/gotify-server/Makefile
index fc46a0e1..7351618c 100644
--- a/cross/gotify-server/Makefile
+++ b/cross/gotify-server/Makefile
@@ -15,8 +15,8 @@ LICENSE  = MIT
 GO_SRC_DIR = $(WORK_DIR)/$(PKG_DIR)
 GO_BIN_DIR = $(GO_SRC_DIR)/$(PKG_NAME)
 
-
-PRE_COMPILE_TARGET = gotify_pre_compile
+COMPILE_TARGET = gotify_compile
+# PRE_COMPILE_TARGET = gotify_pre_compile
 
 CGO_ENABLED = 1
 GO_BUILD_ARGS += -ldflags "-w -s -X main.Version=$(PKG_VERS) -X main.BuildDate=$(shell date "+%F-%T") -X main.Commit=$(PKG_VERS) -X main.Mode=prod"
@@ -26,7 +26,12 @@ include ../../mk/spksrc.cross-go.mk
 ENV += NPM_CONFIG_USER=root
 PATH := $(WORK_DIR)/../../../native/nodejs/work-native/node/bin:$(PATH)
 
-.PHONY: gotify_pre_compile
+.PHONY: gotify_pre_compile gotify_compile
 gotify_pre_compile:
-	cd $(GO_SRC_DIR)/ui && npm install && npm run build
-	cd $(GO_SRC_DIR) && env $(filter-out GOARCH=%, $(ENV)) go run hack/packr/packr.go
+	# cd $(GO_SRC_DIR)/ui && npm install && npm run build
+	# cd $(GO_SRC_DIR) && env $(filter-out GOARCH=%, $(ENV)) go run hack/packr/packr.go
+
+gotify_compile:
+	curl -Lo $(WORK_DIR)/tmp.zip https://github.com/gotify/server/releases/download/v2.1.4/gotify-linux-arm-7.zip
+	unzip -o $(WORK_DIR)/tmp.zip gotify-linux-arm-7
+	mv gotify-linux-arm-7 $(GO_BIN_DIR)
```

thorpelawrence:
Awesome, thanks for the help @hgy59!

Any reason we couldn't just keep the compile for gotify-cli if that works? Or did it not work for all?

thorpelawrence:
> > Any reason we couldn't just keep the compile for gotify-cli if that works? Or did it not work for all?
> 
> We save some build resources (cpu, power) when just using the prebuilt binaries and not depend on native/go anymore. That's why I prefere to use prebuilt binaries of gotify-server for x64 archs too.

Makes sense, happy to use the prebuilt, thought you weren't able to get the executable bit setup but missed that commit.


Is this ready to be merged in that case?

thorpelawrence:
Hey, sorry to follow up but any updates on this?

thorpelawrence:
Thanks for all the help again @hgy59, do you know roughly when I can expect it to start showing up in my package manager? I'm on a DS218j with DSM 6.2.4

thorpelawrence:
Never mind, my synology network settings weren't working properly and only just figured out why, it's working now. Thanks again!

PR #5128, opened 2022-02-12
Remove x86 generic toolchains & kernels

th0ma7:
## Description

While playing with #5095 it became obvious that there are no reasons to maintain a non-existent `x86` generic arch group and generates unnecessary mixups when playing with kernel modules.  The generic x64 generic arch already serves this purpose.

Another approach would be to consider migrating `x86` to be i686 specific but as there is only `evansport` this would add an unnecessary complexity.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


hgy59:
@th0ma7 please do not remove the toolchains of x86 and do not remove x86 from x64 generic archs. x86 is not obsolete and is part of x86_64 (and not i386/i686).
As you can see in the [package architectures](https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures) those are required for the following models:
- DS712+
- DS411+II, DS1511+, DS2411+, RS2211+, RS2211RP+
- DS411+
- DS1010+, RS810+, RS810RP+
- DS710+

BTW I started to publish packages not only for ppc853x-5.2 but also for x86-5.2 and 88f6281-5.2 as far as those build without special effort.

I have some WIP (not yet published) to bring back packages for those models that support DSM 5.2 as latest version (ppc853x, x86, 88f6281).

hgy59:
The only refound of my dsm 5.2 work by now, was to _add latest DSM version for all models not supporting DSM 7_.

th0ma7:
> please do not remove the toolchains of x86 and do not remove x86 from x64 generic archs. x86 is not obsolete and is part of x86_64 (and not i386/i686). As you can see in the [package architectures](https://github.com/SynoCommunity/spksrc/wiki/Synology-and-SynoCommunity-Package-Architectures) those are required for the following models:
> 
> * DS712+
> * DS411+II, DS1511+, DS2411+, RS2211+, RS2211RP+
> * DS411+
> * DS1010+, RS810+, RS810RP+
> * DS710+

Shouldn't it be called `pineview` instead to follow other naming standard or is there a DSM specific reason to keep it using a generic `x86` naming?  I've made a quick PR to rename it accordingly #5129.

> BTW I started to publish packages not only for ppc853x-5.2 but also for x86-5.2 and 88f6281-5.2 as far as those build without special effort.
> 
> I have some WIP (not yet published) to bring back packages for those models that support DSM 5.2 as latest version (ppc853x, x86, 88f6281).

I must say that I am between flabergasted that you where able to revive package on old DSM5 and afraid of the burden of later maintenance... Curious to see what this will entails in hope its not too painful to keep maintaining them.

hgy59:
> Shouldn't it be called `pineview` instead to follow other naming standard or is there a DSM specific reason to keep it using a generic `x86` naming? I've made a quick PR to rename it accordingly #5129.

Synology names those x86. The arch in the package must be named like this.

PR #5152, opened 2022-03-18
update go to 1.18

fgma:
## Description

Update golang to 1.18
Avoid breaking build of syncthing by introduction of native/go_1.17

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


fgma:
Syncthing fails to build. [quic-go](https://github.com/lucas-clemente/quic-go) seems to be the problem. It got some bugfixes for go 1.18 support just a few days ago. There was no release yet so syncthing can't update its dependencies yet.

acolomb:
Would you mind holding this back until Syncthing upstream has adapted its dependencies to be compatible with Go v1.18, and I've updated the package here to include such upstream changes?

Or do you have an immediate need why we should rush out Go v1.18?

fgma:
Of course we have to wait until syncthing is compatible. I was optimistic it would just work. Now we know the current status which isn't too bad.

hgy59:
> Or do you have an immediate need why we should rush out Go v1.18?

@acolomb To avoid holding back updates of minio and gitea packages, I have added native/go_1.17 to build syncthing again.
So you have no stress to update syncthing (including the fix to keep syncthing internal updates on sc package updates).

acolomb:
Syncthing's upstream is currently working on 1.18 compatibility. Will hopefully be part of the next release. Alright I will take care of switching to it when ready.

fgma:
I'm not sure if I created some confusion but minio doesn't need go 1.18. My idea was to get it updated together with go to be prepared for possible future requirements and keep the number of spk releases limited.

As you've already put the effort into supporting go 1.18 und 1.17 in parallel we can go that route and use minio or any other package to test 1.18 before switching everything else over in future. That way we would also have a template for switching over to the next go release in future easily.

hgy59:
> I'm not sure if I created some confusion but minio doesn't need go 1.18.

No confusions, and I prefere to update to go 1.18 too.
As go 1.18 removes the support for TLS 1.0 and 1.1 protocols, it is also a security enhancement (and probably the cause for syncthing to fail).

fgma:
@hgy59 I've updated my PR for minio: https://github.com/SynoCommunity/spksrc/pull/5154

PR #5156, opened 2022-03-19
Fix make exit status

th0ma7:
## Description

This PR address the issue with the exit code on error.  
1. There are corner-cases where builds does not exit with proper exit error signal. After further looking into it this appear to be due to the tee pipe where the exit code is from the later command. The way to properly handle this is through the `${PIPESTATUS[0]}` using `bash` shell variable which provides the output from the first command from the pipe sequence.
2. In extra to that, all instances of `-@$(MAKE)` have been replaced with `@$(MAKE)` to allow passing errors
3. Additional github-action success test is being used to confirm that the resulting package has been created or not.

Part of #5095 PR split.

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5162, opened 2022-03-21
spk.mk: Fix all-supported and other parallel-build calls

th0ma7:
## Description

When parallel building do not stop on failures but continue on next arch.

Fixes #5156

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.cross-cc.mk
+++ b/mk/spksrc.cross-cc.mk
@@ -1,26 +1,6 @@
 # Default make programs
 #
 
-# When compiling from cross/* it is possible
-# build in parallel but only using all-supported
-# option.
-#  Ex: PARALLEL_MAKE=max make -j4 all-supported
-#
-# Manual multi-arch parallel build using both
-# -jX in conjunction with PARALLEL_MAKE=X flag
-# is NOT supported.
-#  EX: PARALLEL_MAKE=max make -j2 arch-aarch64-7.0 arch-x64-7.0
-# 
-# The .NOTPARALLEL is needed in order to disable
-# the -jX flag when called within cross/* only.
-# This does not affect building from spk/*
-#
-ifeq ($(filter all,$(subst -, ,$(MAKECMDGOALS))),)
-ifneq ($(PARALLEL_MAKE),nop)
-.NOTPARALLEL:
-endif
-endif
-
 # Common makefiles
 include ../../mk/spksrc.common.mk
 include ../../mk/spksrc.directories.mk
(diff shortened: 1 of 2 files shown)


hgy59:
@th0ma7 another bug that is fixed with this, occurres on building `cross/<package>`.

e.g. `cd cross/openssl; make arch-x64-7.0` fails with:
```
make[2]: Leaving directory '/spksrc/cross/openssl'
/bin/sh: 2: Bad substitution
make[1]: *** [../../mk/spksrc.cross-cc.mk:108: build-arch-x64-7.0] Error 2
make[1]: Leaving directory '/spksrc/cross/openssl'
make: *** [../../mk/spksrc.cross-cc.mk:104: arch-x64-7.0] Error 2
```

th0ma7:
Does it still build under github while capturing errors?  (Btw, disconnected atm)

hgy59:
As you can see in https://github.com/SynoCommunity/spksrc/runs/5636365391?check_suite_focus=true
The github action does now handle UNSUPPORTED_ARCHS as error, so it never will create packages that have UNSUPPORTED_ARCHS.
So my proposal: undo your changes and just leave it as it was...

th0ma7:
Btw, i had noticed something wrong with pre-check.mk as shell calls in it cannot work as theses are not "recipes".  Simply changing error to info or warning shows the underneath code issue.  The only reason it works is because the error forces exit.

Worst case you can revert the change for now until i can find cycles to review that further again.  But in the end i'll absolutely be needing it for kernel modules package builds (and good thing i broke it in little pieces to find issues before hand).

On the other hand if your added revert do work, you may have to do the same on other cross-*.mk, i'm not sure.

On that note, i communicated with Synology over the past few days and 6.2.4 kernels should appear starting today.  This is really the final missing piece to tie it all.

th0ma7:
> As you can see in https://github.com/SynoCommunity/spksrc/runs/5636365391?check_suite_focus=true The github action does now handle UNSUPPORTED_ARCHS as error, so it never will create packages that have UNSUPPORTED_ARCHS. So my proposal: undo your changes of #5156 and just leave it as it was...

@hgy59 while everything else seems fixed I just hit this issue myself... indeed `UNSUPPORTED_ARCHS` exit 2 as it should but make github-action to appear as failed.  Although that's probably the last remaining item I guess.  I'll see what can be done to circumvent this (as the exit status is actually ok).

PR #5163, opened 2022-03-21
spksrc: Fully use DSM7 var directory

th0ma7:
## Description

_Motivation:_  Follow-up to to both #4579 and #4797 with focus only on `var`.  This now uses previously merged PR such as:
- (#4997) Usage of rsync to migrate `target/var` to `var` in `spksrc.service.installer.dsm7`
- (#4997) Add a functionality that allows generating `.new` suffix new files when previous ones already exist.  User can then refer to them and adapt their own configuration as needed.

Current framework simply move files from `target/var` to `target/../var` at package installation/upgrade time.  While it's probably fine for most use-cases, this doesn't adapt all configuration files at build time using the "real" end-state path and most importantly the application isn't "aware" by default of the expected `var` directory path.

Solving this require using `localstatedir` option at configuration time.  Doing so enforces the application to have an awareness and use by default the new DSM7 `var` directory and thus, installs files at the proper location and adjust paths in configurations at build time.  _Not doing so leaves a bit of unknown on actual application state & behavior._

This PR aims at solving the current limitations by doing the following on top of rsync work (#4997):
1. enforcing the usage of `localstatedir` for DSM7 builds.  At package generation time, files installed under `<workdir>/install/var/package/<package>/var` are now being handled in order to be packaged and installed properly.

## Details
This new implementation is a major step above previous attempt as it:
1. requires **NO** changes at all in any packages.
2. is key to demonstrate feasibility to also migrate `etc` in a later work

diag-var-spksrc

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


Safihre:
I tried to review, since I know you put a lot of work in it, but I am not fully understanding (even after reading the previous PR) what this change will do?
Maybe you could explain it again shortly?

th0ma7:
Thnx @Safihre for looking into it.  It is now ready for review (you where a few minutes to early).  I've added a much better description + a diagram that explains it hopefully well.

Safihre (approved):
Based on your diagram, it looks good. All changes seem logical.
But @hgy59 or someone else is more an expert.

th0ma7:
@SynoCommunity/developers Friendly reminder for reviews.  Unless there is opposition to this PR I'll look into merging in the coming days.

mk/spksrc.spk.mk
```diff
@@ -269,7 +269,7 @@ $(DSM_LICENSE_FILE): $(LICENSE_FILE)
 $(WORK_DIR)/package.tgz: icon service
 	$(create_target_dir)
 	@[ -f $@ ] && rm $@ || true
-	(cd $(STAGING_DIR) && tar cpzf $@ --owner=root --group=root *)
+	(cd $(STAGING_DIR) && find . -mindepth 1 -maxdepth 1 -not -empty | tar cpzf $@ --owner=root --group=root --files-from=/dev/stdin)
```
hgy59, line 272: @th0ma7 
We must include empty directories and remove `-not -empty` in this line.
There are several packages that create (empty) directories in the staging area to have those created on package installation.
Identified so far:
- tt-rss creates var/log
- gateone creates ssl
- lirc creates var/log
- ntopng creates var/data
- haproxy creates var/crt
- monit creates var/events
...

hgy59: @th0ma7 you're right, subfolders under var are not a problem. But as I was building gateone packages, those where missing the empty ssl folder (target/ssl).

th0ma7: @hgy59 so indeed this would affect the ssl top-directory as being empty for `gateone`... Unless you convert that into `service-setup.sh` directory creation shell call instead?

hgy59: is there any need to exclude empty directories?

th0ma7: There was one in one of my previous (and more complex) PR where I was dealing with all other paths as well.  But with the current setup no, I don't believe so.  It most probably can be removed with no harm.

hgy59: As it is not a good solution to put certificates in target/ssl, I will update gateone to use target/var/ssl folder.

hgy59:
I have a package under development (urbackup) where the cross packages create config files in the var folder. PLIST contains entries like `rsc:var/urbackup/dataplan_db.txt`.

After rebase to include this PR, DSM 7 packages cannot be created anymore.
- The install folder for DSM 7 does not contain the `{packagename}/target/var` but only an empty `{packagename}/var` folder where for DSM 6 is does.
- The staging folder for DSM 7 is empty, so there is an error:
   ```
   ===>  [DSM7+] Copy & merge var and target/var to /spksrc/spk/urbackup/work-x64-7.0/staging/var
   tar: Cowardly refusing to create an empty archive
   ```

@th0ma7 any idea how to handle this?

th0ma7:
@hgy59 can you point me to your development branch i'll look into it probably tomorrow.

hgy59:
@th0ma7 the copy and merge of different var folders needs two command lines, as tar must be executed within each var folder separately.

hgy59:
The following existing cross packages have var folders in PLIST

- domoticz
- ejabberd
- mutt
- rsnapshot 
- saltgui (spk: salt-master)
- zap2epg (spk: tvheadend)

@th0ma7 shall I create a dedicated PR to merge this fix to master?

th0ma7:
Yes please do @hgy59 , I'll have a second look at it as well.

PR #5164, opened 2022-03-21
Separate packages for each arch

Safihre:
Instead of having 1 huge result, lets split the results based on `arch`.
Is there any reason not to do this?


Safihre:
Me too, but usually download speeds offered by Github are the limiting factor.

PR #5168, opened 2022-03-23
Disable pip version check during build

Safihre:
Save some extra time and logging.
Any objections @th0ma7 @publicarray ?


th0ma7:
@Safihre no objection at all... (I was actually considering it)

th0ma7 (approved):

PR #5165, opened 2022-03-21
update cross/openssl to v1.1.1n

hgy59:
## Description

- fixes CVE-2022-0778
related to: #5161

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5171, opened 2022-03-24
Python 3.10.4 Update

th0ma7:
## Description

Update to latest version 3.10.4 along with updating all wheels & cross-dependencies.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


spk/python310/Makefile
```diff
@@ -30,7 +30,7 @@ SPK_USR_LOCAL_LINKS = bin:bin/python3.10
 WHEELS = src/requirements-crossenv.txt src/requirements-pure.txt src/requirements-abi3.txt
 
 # Force building pure-python wheels
-#WHEELS_PURE_PYTHON_PACKAGING_ENABLE = 1
+WHEELS_PURE_PYTHON_PACKAGING_ENABLE = 1
```
Safihre, line 33: Why is this sometimes on and sometimes off?

spk/python310/Makefile
```diff
@@ -39,18 +39,18 @@ WHEELS = src/requirements-crossenv.txt src/requirements-pure.txt src/requirement
 ##
 
 # [gevent]
-# DEPENDS += cross/libev cross/c-ares
-# ENV += GEVENTSETUP_EMBED_CARES=FALSE
-# ENV += GEVENTSETUP_EMBED_LIBEV=FALSE
+DEPENDS += cross/libev cross/c-ares
```
Safihre, line 42: Same question, why are these sometimes on or off?

spk/python310/Makefile
th0ma7: By default it's always off.  This mean that at package install `pip` will install all local wheels and download the missing ones.  Provided wheels are the cross-compileds ones while the pure-python which are arch agnostic are being downloaded from pypi with pip.  This allow reducing greatly the package size considering we're short on disk space.

When I enable it is to demonstrate that the framework still works as expected and that it can build pure-python wheels and package them.  Once all lights are green I re-disable them.

spk/python310/Makefile
th0ma7: Same question but different answer...  By default python310 comes with no wheels (besides certificates).  Including wheels in python package ends-up breaking individual packages dependent on python310 as they cannot superseed the version provided by default with python310.  As such it was decided to not be including any and leave all of them in the sub-application packages.

But, there are many caveats at cross-compiling wheels.  As such I've included all the most common ones in python310 to test and demonstrate that cross-compiling do work at every updates.  This allows me to switch then ON then OFF simply to confirm builds are succesfull.  Application packages can then refer to python310 as it hosts the most common combinations and can be used as an example.

Hope this helps.

th0ma7:
@hgy59 this one is good to go.  Are we keeping track of updates in https://github.com/SynoCommunity/spksrc/pull/5165#issuecomment-1074529384 ?

spk/python310/Makefile
```diff
@@ -13,7 +13,7 @@ DESCRIPTION_FRE = Langage de programmation Python.
 DESCRIPTION_SPN = Lenguaje de programación Python.
 STARTABLE = no
 DISPLAY_NAME = Python 3.10
-CHANGELOG = "1. Update to python 3.10.2"
+CHANGELOG = "1. Update to python 3.10.4 <br/>2. Update to OpenSSL 1.1n"
```
hgy59, line 16:
```suggestion
CHANGELOG = "1. Update to python 3.10.4 <br/>2. Update to OpenSSL 1.1.1n"
```

PR #5173, opened 2022-03-24
kernel: Add DSM-6.2.4 kernels

th0ma7:
## Description

After multiple requests Synology finaly released DSM-6.2.4 source code!!!  Interestingly the kernel source code are on the new synology server but the toolchains stayed on sourceforge.  We can probably expect all new release to be on the new server.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


Safihre:
Can we not get rid of all the other 6.2 ones? Why do we need to keep those?
Same for the toolchains.

hgy59:
> Can we not get rid of all the other 6.2 ones? Why do we need to keep those? Same for the toolchains.

As those need very little resources - one Makefile and one digests file (< 600 Bytes) per arch and version, I have no problem to keep all toolchains and kernels in our repo.
Regard it as documentation of what synology provides to download.

Safihre:
Why are there so many more of 6.2.X than of the other ones?

th0ma7:
Simply because synology ended-up switching their numbering schema ... and there where many security issues as well (Meltdown/Spectre/L1TF) that needed backporting to various kernels while not really updating anything else in the DSM OS.

PR #5181, opened 2022-03-27
syncthing: Avoid overwriting newer binary from internal updater (fixes #5142)

### LINKED ISSUE #5142: Syncthing Not Starting, after update to 1.19.0.25
### Is this a new Bug?

### Package Name

Syncthing

### Package Version

1.19.0.26

### Device Model

DS620slim

### Device Architecture

x86_64

### Firmware Version

DSM 7.0.1-42218 Update 2

### What happened?

Having updated to version 1.19.0-25, when trying to start syncthing, i'm getting the following error in the sycnthing.log

```
WARNING: Failed to initialize config: config file version (36) is newer than supported version (35). If this is expected, use --allow-newer-config to override.
[monitor] 11:31:17 INFO: Syncthing exited: exit status 1

```

How can I resolve this?

### Reproduction steps

```bash
1. Got DSM Package Centre
2. Start Syncthing
3. After a few seconds, the syncthing service fails to start, and I end back at the option to run the service, checked the log file, and see the above error.
```

### Install Log

```log
Wed Mar  2 11:31:14 GMT 2022
Begin service_prestart
End service_prestart
Starting syncthing command /volume1/@appstore/syncthing/bin/syncthing serve --home=/volume1/@appdata/syncthing  
[start] 11:31:14 INFO: syncthing v1.19.0 "Fermium Flea" (go1.17.5 linux-amd64) root@publish 2022-02-03 19:51:15 UTC
[start] 11:31:14 WARNING: Failed to initialize config: config file version (36) is newer than supported version (35). If this is expected, use --allow-newer-config to override.
[monitor] 11:31:14 INFO: Syncthing exited: exit status 1
[start] 11:31:15 INFO: syncthing v1.19.0 "Fermium Flea" (go1.17.5 linux-amd64) root@publish 2022-02-03 19:51:15 UTC
[start] 11:31:15 WARNING: Failed to initialize config: config file version (36) is newer than supported version (35). If this is expected, use --allow-newer-config to override.
[monitor] 11:31:15 INFO: Syncthing exited: exit status 1
[start] 11:31:16 INFO: syncthing v1.19.0 "Fermium Flea" (go1.17.5 linux-amd64) root@publish 2022-02-03 19:51:15 UTC
[start] 11:31:16 WARNING: Failed to initialize config: config file version (36) is newer than supported version (35). If this is expected, use --allow-newer-config to override.
[monitor] 11:31:16 INFO: Syncthing exited: exit status 1
[start] 11:31:17 INFO: syncthing v1.19.0 "Fermium Flea" (go1.17.5 linux-amd64) root@publish 2022-02-03 19:51:15 UTC
[start] 11:31:17 WARNING: Failed to initialize config: config file version (36) is newer than supported version (35). If this is expected, use --allow-newer-config to override.
[monitor] 11:31:17 INFO: Syncthing exited: exit status 1
[monitor] 11:31:18 WARNING: 4 restarts in 4.203793074s; not retrying further
```

### Service Log

_No response_

### Other Logs

_No response_

acolomb:
## Description

Syncthing by default has its internal updater enabled, so the running version will usually be newer than what the SPK had installed.  When re-installing the SPK or upgrading after the previously running version has already updated itself to something newer, Syncthing may fail in some cases because it cannot read its configuration file.  It handles database upgrades gracefully, checking for upgrades first (unless disabled by preference).  But for the config, it just refuses to run if the stored file specifies a version number "from the future".

This can be avoided by retaining the binary during upgrades, if it is newer than the one contained in the SPK package.  Fixes #5142.

Also contains some minor refactoring of the `service-setup.sh` script.

**NOTE: This PR does not yet include the most recent Syncthing version.  During testing, this triggers the new code path immediately if Syncthing was previously running.  Before merging, the most recent release should be pulled in though.**

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update


acolomb:
I will bump the package revision when an up-to-date upstream version has been incorporated.  Please review / test already, and let me know when ready.  @publicarray thanks for your suggestion on Discord, I adapted it slightly.

zzlazz:
Thanks for pointing me to this @acolomb , but I have a n00b question: how do I use this info?

acolomb:
Wait for the CI build here to complete.  Then you can download a package matching your NAS model for testing.

Safihre:
Due to the number of updates and problems like these. I wonder if we can't just download the official Syncthing release during install instead of us compiling every release?
Just like we do for nzbget.

acolomb:
Sure. That's what upstream has been advocating for as well. I see one problem with it though: You will definitely need an Internet connection at the time of installation then. We should maybe include some binary in the SPK as fallback, for when someone downloads the package from our website for later offline installation.

I thought that compiling stuff during the package build was the preferred way for spksrc for some reasons?

hgy59:
> Due to the number of updates and problems like these. I wonder if we can't just download the official Syncthing release during install instead of us compiling every release?

It is not intended to create packages for every new syncthing release. IMHO we create new package releases only to update the installation environment as we did for DSM7 compatibility, DSM certificate integration, user definition in installation wizard, ...
When internal update is recommended/used, it is important not to _downgrade_ the binaries at package update.

BTW: most provided package binaries are not directly installable due to the need to link against system libraries like glibc, the non standard location of libraries. Or binaries are not available for specific archs like ARMv5, ARMv7L, PPC.

Safihre:
There is internet during install, because that's how the spk itself was downloaded.
Indeed this package would be more complicated than nzbget, which has 1 installer for all archs, while Syncthing has seperate ones.
@hgy59 I do think the ARM release of Syncthing probably works for the ARMv5 ones.
So let's ignore my suggestion for now, too much work probably.

acolomb:
> There is internet during install, because that's how the spk itself was downloaded.

Not necessarily, someone might have downloaded the SPK from the SynoCommunity repository onto a computer, then chose "Manual Install" in the Package Center, which would no longer work without a direct internet connection from the NAS.

D4nte:
Is it an option to disable the internal updater by default?

acolomb:
I wouldn't want to go that route, because Syncthing is a rather active project and although it strives for compatibility between most versions, missing out on updates and bug fixes would be unfortunate. It's essentially only useful in connection with other devices, and having the Synology devices always lurking behind on an older version will definitely cause troubles at some point. Given how long it takes to even get this fix for actual serious user issues published, I think needing to rely even more on developer action to maintain updates for the SPK will lead to a very slow update interval.

acolomb:
Since this contains a fix for actual user troubles, can we please get this merged soon? @hgy59 @publicarray

It now includes and supersedes the changes from #5246 as well.

hgy59:
@acolomb it is now published (built with go 1.18.1)

acolomb:
Great, thanks!

PR #5183, opened 2022-03-28
Update zlib as the previous version was deleted by authors

Safihre:


Safihre:
@hgy59 Sorry about that. I promise I will never just merge things anymore without a review! 
I was in a hurry to test my project...

PR #5194, opened 2022-04-04
Use shallow clone when including NZBGet tools

Safihre:
This lowers the size of the NZBGet SPK from 30MB to 14MB.
We could also just do this `git clone` during installation script, then the package would shrink to only a few MB.
But it would add the `git` dependency. Is that worth it?

### DIFF
--- a/spk/nzbget/Makefile
+++ b/spk/nzbget/Makefile
@@ -26,6 +26,10 @@ SERVICE_PORT_TITLE = NZBGet (HTTP)
 # Admin link for in DSM UI
 ADMIN_PORT = $(SERVICE_PORT)
 
+# Installer brings all the binaries
+COPY_TARGET = nop
+override ARCH=
+
 POST_STRIP_TARGET = nzbget_extra_install
 
 include ../../mk/spksrc.spk.mk
--- a/spk/nzbget/Makefile
+++ b/spk/nzbget/Makefile
@@ -36,5 +36,5 @@ include ../../mk/spksrc.spk.mk
 nzbget_extra_install:
 	install -m 755 -d $(STAGING_DIR)/var $(STAGING_DIR)/share/nzbget/scripts
 	# Include usefull scripts
-	cd $(STAGING_DIR)/share/nzbget/scripts && git clone https://github.com/clinton-hall/nzbToMedia.git
-	cd $(STAGING_DIR)/share/nzbget/scripts && git clone https://github.com/clinton-hall/GetScripts.git
+	cd $(STAGING_DIR)/share/nzbget/scripts && git clone --depth=1 https://github.com/clinton-hall/nzbToMedia.git
+	cd $(STAGING_DIR)/share/nzbget/scripts && git clone --depth=1 https://github.com/clinton-hall/GetScripts.git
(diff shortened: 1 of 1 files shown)


hgy59 (commented):
@Safihre are you using this package?

I am unsure about the installed scripts.
As those are python scripts, we probably have to make this package dependent on the Python (3.10) package?
Or we could search for available python at installation time. But this would not add package dependency to nzbget and the found python could be uninstalled and break the nzbget scripts.

I suppose that at least we have to add python to the path for nzbget and fix the shebang in the python scripts as `#!/usr/bin/env python` might not work.

Instead of cloning the script repos we could take the files only.
This is easy for GetScripts as all scripts are in one folder (and we can drop .gitignore)
But for nzbToMedia there is a folder structure with lib and other folders. One folder contains windows binaries that I am sure we can ignore, and getffmpeg.sh will not run on diskstation devices.
By taking the script files only (and removing win and test folders and some other files of nzbToMedia) the final spk size is shrinking to less than 8MB.

Safihre:
@hgy59 No, I don't use this package, but around 5 years ago I created the current version that `wget` the installer instead of compiling it.
But these scripts were included like this even before that, more than 7 years ago: https://github.com/SynoCommunity/spksrc/blame/master/spk/nzbget/Makefile#L39
So, if it hasn't been a problem since then... why try to fix it now?
I do know that NZBGet allows you to specify which Python to use from the configuration, so maybe people use that.

In any case, I just noticed it when inspecting the `nzbget.spk` and I was surprised by the huge size even though it doesn't include the installer itself.

Safihre:
@hgy59 I just tested, and it just works out of the box. Seems it's just using the Synology Python.
I also found that NZBGet installer includes 7zip and unrar, so we don't even need to include those.

What changes of the ones you suggested do you want to include?

hgy59:
@Safihre the current build fails, because there are no dependencies anymore to be added to the package.
to fix this, overwrite the copy_target in `spk/nzbget/Makefile` with:
```
COPY_TARGET = nop
```

_this must not be defined with my solution, where the scripts come from cross packages_

my proposal reduces the spk size from currently 11.5MB to 5.6MB

hgy59:
> But these scripts were included like this even before that, more than 7 years ago: https://github.com/SynoCommunity/spksrc/blame/master/spk/nzbget/Makefile#L39

cool, didn't see/use the _Blame_ button before...

Safihre:
@hgy59 I investigated more.
`nzbToMedia` includes an auto-update mechanism based on git. So if we use the `cross` approach, it would break. We could add `SPK_DEPENDS = "git"`, but since not all users of NZBGet use `nzbToMedia` it's maybe overkill.

So, I converted this to a `norach` since there's no more binaries. 
Seems enough for now, if you agree?

hgy59:
> We could add `SPK_DEPENDS = "git"`, but since not all users of NZBGet use `nzbToMedia` it's maybe overkill.

We don't need `SPK_DEPENDS = "git"`, when the synocommunity package `git` is installed, git is found in the path.
So it is a _soft_ dependency for those who want to use the autoupdate of `nzbToMedia`.

hgy59 (approved):
LGTM

PR #5195, opened 2022-04-05
Remove Release step from Build action

Safihre:
Since we never use it, let's remove it.


publicarray:
@Safihre Just to give some background, I used it to make these releases (see attached assets): https://github.com/publicarray/spksrc/releases so I thought it was a useful feature for others that made a fork and wanted to build their packages but did not want to send a PR or wait for us to publish. e.g. as a testing ground or software that won't be included. They could use GitHub to make a release (a GitHub workflow run assets are removed after a period of time)

Issue #5148, opened 2022-03-09
GateOne on DSM 4.2 (DS109J) needs Python


BRIOTPatrick:
### Is this a new Bug?

- [X] I checkd that the bug hasn't been reported before

### Package Name

Python

### Package Version

2.7.6-8

### Device Model

DS109J

### Device Architecture

PPC

### Firmware Version

4.2

### What happened?

I want to install GateOne on my DS109J on DSM 4.2.
When i try, a popup inform me i've to install Python 2.7.6-8, but this package isn't in your repository.

Could you help me to find it ?

Best Regard.
Bigvivi

### Reproduction steps

```bash
1.
2.
3.
...
```

### Install Log

```log
.
```

### Service Log

```log
.
```

### Other Logs

```log
.
```

hgy59:
@BRIOTPatrick are you sure you have DSM 4.2 ? The latest DSM for DS108J on the synology download pages is Version 4.0-2265 (2014-09-10).

BRIOTPatrick:
Sorry for the big mistake, effectively a DS109J ! And yes i'm on DSM 4.2-3259.

hgy59:
@BRIOTPatrick i was able to build python2 (2.7.18-26) for powerpc-4.2, but I had to exclude wheel for audio (cross/pyalsa, cross/pyaudio) and the zmq and uwsgi wheels (cross/pyzmq, cross/uwsgi).
I'm currently no aware whether gateone depends on one of those cross compiled wheels.
If you are interested, I can upload such a python2 package.

PR #5209, opened 2022-04-15
Framework cleanup

hgy59:
## Description

Some pending fixes and cleanups on my TODO list

## Checklist

- [x] Avoid icons in packages that do not integrate into DSM UI
- [x] Avoid overwriting build logs for successive make calls
- [x] Enable patches for noarch and native builds
- [x] Avoid irrelevant warnings when building wheels for python2
- [x] Consistently use BUILD_DEPENDS for native dependencies
       This triggers a lot of packages to build, but without code changes.
- [x] Use explicit PKG_DIST_FILE when PKG_DIST_NAME does not contain the package name
- [x] Shorten some github archive urls
- [x] Improve cross/x265 code generation
       This is the only code change in this PR - regarding ffmpeg (and dependents) and imagemagick packages

### Type of change

- [x] Small framework changes


cross/ffmpeg/Makefile
```diff
@@ -183,17 +181,19 @@ CONFIGURE_ARGS += --cpu=e500 --disable-altivec
 endif
 endif
 
+
+ifeq ($(findstring $(ARCH),$(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
+BUILD_DEPENDS = native/nasm
+NASM_PATH = $(relapath $(WORK_DIR)/../../../native/nasm/work-native/install/usr/local/bin)
+ENV += PATH=$(NASM_PATH):$$PATH
+endif
```
th0ma7, line 189: Have you found out why it failed for x64-6.1 and 7.0?

mk/spksrc.wheel.mk
```diff
@@ -85,9 +85,11 @@ ifneq ($(strip $(WHEELS)),)
 	      crossenvPIP=$$(. $(CROSSENV) && which pip) ; \
 	      $(MSG) "Python crossenv found: [$(CROSSENV)]" ; \
 	      $(MSG) "pip crossenv found: [$${crossenvPIP}]" ; \
-	   else \
+	   elif [ "$(PYTHON_VERSION)" != "2.7" ] ; then \
 	      $(MSG) "WARNING: Python crossenv NOT found!" ; \
 	      $(MSG) "WARNING: pip crossenv NOT found!" ; \
+	   else \
+	      $(MSG) "Python $(PYTHON_VERSION) uses pip: $${crossenvPIP}" ; \
```
th0ma7, line 92: As an FYI, I intent to revisit python 2.7 in a short future.  In the meantime (and perhaps also in the long-run) this change looks good.

th0ma7 (changes requested):
It currently fails on x64 which needs investigating.  Other than that all changes looks good.

cross/ffmpeg/Makefile
hgy59: finally yes (just a typo)

mk/spksrc.wheel.mk
hgy59: @th0ma7 I have python2 in progress for #5148.
As python2 spk contains some wheels that have vulnerabilities and are not updated for python 2 anymore, I am about to remove most of the wheels in that package (keep only pip, virtualenv and setuptools).
And probably we should publish python2 for DSM 7 too, as python2 by Synology does not contain pip, virtualenv and setuptools.

th0ma7: I'm good with that, and then we can publish ffsync

PR #5277, opened 2022-05-16
ffmpeg extras

th0ma7:
## Description

ffmpeg extras: Add extra version tag to easily identify running version along with encoders/decoders to ffmpeg

Includes a generic fix for packages that link to libatomic and must include this library from the toolchain in the package.

Fixes #5254

## Checklist

- [x] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5280, opened 2022-05-19
CMake rework use toolchain for cross-compiling

th0ma7:
## Description
I've hit a nasty issue using cmake when it requires to build "host" compatible sub-tools in order to be used within the build process to generate "target" binaries/libraries.  It hapens that the way I've originally implemented CMake framework was sort of a clone of the autoconf style where we set the environment as "target" therefore hiding "host" tools.  Current spksrc cmake framework hides the "host" tools which in turns always generate "target" binaries that can't be ran on the host (e.g. cannot run a `armv7` target compiled binary on `x86_64` host).

I've hit this issue while trying to build `codec2` as it requires some binaries to be natively compiled and used from the host to generate "target" results.  I first thought I could create a `native/codec2` and build a host-compatible `.../generate_codebook` but the `codec2` build process doesn't allow that.

This PR aims to change that entirely and make usage of CMAKE toolchain files.  What this does is defining the "target" environment for cross-compiling using a toolchain file per arch generated at build-time.  This allows keeping the "host" side of things clean (ex: `CC=/usr/bin/gcc`) similar to doing a "native" compiling (e.g. debian docker gcc,make,etc are still "available" by default to cmake environment).

More documentation available here: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html

Discussion on this was initiated while trying to figure out the build problem with `codec2` developers https://github.com/drowe67/codec2/issues/324

## How does it works:
1. At toolchain time it now also generates a `$(WORK_DIR)/$(ARCH)-toolchain.cmake` file containing all the generics that will be used by package building later-on (e.g. arch details and path to cross-compilers & tools)
2. At `spksrc.cross-cmake.mk` configure time it creates a copy of `$(WORK_DIR)/$(ARCH)-toolchain.cmake`  under `$(WORK_DIR)/$(PKG_DIR)` and adds a few mandatory bits such as `CMAKE_*_FLAGS`, `*RLINK*` options and assembly compiler path.  Creating a specific toolchain file per package was mandatory as **a)** flags variables are not accessible from toolchain step and **b)** all theses specific options varies from one package to another and **c)** using CMake it isn't possible to combine command line `-DCMAKE_*FLAGS` with a toolchain file (later always have precedence)
3. At configure time CMake is being called with `-DCMAKE_TOOLCHAIN_FILE=$(CMAKE_PKG_TOOLCHAIN_FILE)` but most importantly, the `$(ENV)` is being cleared off of cross-compilers & tools (e.g. `CC`, `CPP`, etc.) and from all the `*FLAGS` (similar to `spksrc.native-env.mk` but with sufficient differences that it cannot be reused).  This flag clearing step on `$(ENV)` is mandatory as otherwise they **a)** conflicts with the toolchain file, but **b)** most importantly, get used when building "host" compatible binaries (and a `x86_64` host binary will not compile with `arm` specific optimisations such as `-mcpu=8548 -mhard-float -mfloat-gprs=double`).

## In more details:
1. Migrate current `spksrc.cross-cmake.mk` to using toolchain files for cross-compiling (e.g. https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html)
4. Allow using `ADDITIONAL_CFLAGS`, `ADDITIONAL_CPPFLAGS`, `ADDITIONAL_CXXFLAGS` and `ADDITIONAL_LDFLAGS` keeping framework layout instead of `-DCMAKE_C_FLAGS=` ...
5. Add the ability to return to previous "legacy" method using `CMAKE_USE_TOOLCHAIN_FILE = OFF`
6. Add a new `spksrc.native-cmake.mk` functionality for building native cmake tools
7. Migrate `native/libmysqlclient` to using new `spksrc.native-cmake.mk`
8. Add `cross/codec2` which mandatory requires using toolchain files in order to generate a host compatible `generate_codebook` binary to complete its build process.  Also add to `ffmpeg`.
9. Add new `chromaprint-fftw` library independant from regular `chromaprint` package and using `fftw`, allowing to be enabled within `ffmpeg`.  This could in turn eventually allow removing `chromaprint` spk entirely.

Related to #5277 and https://github.com/drowe67/codec2/issues/324

## Testing
1. All `cross/*` dependent to `spksrc.cross-cmake.mk` have been built succesfully
2. Many `spk/*` directly or indirectly related to a `cmake.mk` dependent `cross` have been built succesfully
3. `ffmpeg` was in depth built & tested on `x64` (both DSM-6.1 and DSM7.0), `armv7` DSM-7.0 and `armv5` DSM-6.1

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [x] Package update


th0ma7:
@hgy59 and others I've hit a nasty issue using cmake.  It hapens that the way I've originally implemented it was sort of a clone of the autoconf style where we set the environment for the "target" and in essence we lost track of the "host".

I've hit this issue while trying to build `codec2` where it requires some binaries to be natively compiled and used from the host the generate "target" results.  Current spksrc cmake framework hides the "host" tools which in turns always generate "target" binaries that can't be ran on the host (e.g. cannot run a `armv7` target binary on `x86_64` host).

This PR aims to change that entirely and make usage of CMAKE toolchain files.  What this does is defining the "target" environment for cross-compiling BUT, you have to keep the "host" side of things clean similar to doing a "native" compiling.

More documentation available here: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html

Discussion on this was initiated while trying to figure out the build problem with `codec2` developers https://github.com/drowe67/codec2/issues/324

I'd be much interested in having proposals on how to test this change in-depth before merging.  For a start I intend to use `ffmpeg` as there are many many CMAKE components to it.  But other ideas are welcomed.

th0ma7:
@SynoCommunity/developers all blocker now fixed, this PR is now ready for review & merge.  It looks fairly large but in reality it isn't that much.  This should help significantly with complex cmake builds while avoiding `native/*` in many occasions.

th0ma7:
@hgy59 if you have a moment to review/comment would be much appreciated as my new `gstreamer` PR #5298 is dependent on this.  Much thnx in advance.

cross/cereal/Makefile
```diff
@@ -18,7 +18,8 @@ CMAKE_ARGS += -DWITH_WERROR=no
...
 include ../../mk/spksrc.cross-cmake.mk
 
-ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
-# fix compilation with newer compilers
-CMAKE_ARGS += -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_C99 -D_GLIBCXX_USE_C99_MATH" 
+# Fix compilation with newer compilers
+# Flags to be added to CMake toolchain file
+ifeq ($(call version_ge, $(call version_gcc), 7.5),1)
+ADDITIONAL_CXXFLAGS += -D_GLIBCXX_USE_C99 -D_GLIBCXX_USE_C99_MATH
 endif
```
hgy59, line 25: Didn't you want to move this to the "CMake toolchain file" in this PR?

hgy59:
@th0ma7 my review is still in progress...

Findings so far:
- `cross/libva`  fails to build for aarch64-6.1 - probably fails for all cross builds:
   ``` 
   checking for suffix of executables... 
   checking whether we are cross compiling... configure: error: in `/spksrc/cross/libva/work-aarch64-6.1/libva-2.14.0':
   configure: error: cannot run C compiled programs.
   If you meant to cross compile, use '--host'.
   See 'config.log' for more details
   ```

cross/cereal/Makefile
th0ma7: I thought of that... and I'm not sure about it as it's not "all" cases that needs it...  I'm willing to go as you see fit, worst case we can change it back later on.

Also, if I recall (and I may be wrong) this was used with GCC-7.3 for DSM7-beta.  If that was the case it may not be needed anymore with GCC-7.5 (again, this is a vague souvenir of the gcc version difference between beta and final, I'd need to double check to be sure).

th0ma7:
> my review is still in progress...

Thnx, much appreciated.

> `cross/libva` fails to build for aarch64-6.1 - probably fails for all cross builds:

@hgy59 in theory this really is only needed for x86_64 for i965 and iHD Intel drivers.  Still, I took a stab at it and converting it to meson (which we probably didn't have at the time) solved it all.  Should now work for all arches except `libva-utils` that won't build for armv5.

hgy59:
My verifications with building all cross packages using cmake builds are now successful.

We can work on `ADDITIONAL_CXXFLAGS += -D_GLIBCXX_USE_C99 -D_GLIBCXX_USE_C99_MATH`  later on (those flags are required only for packages using std:math).

I don't know whether the following lines in cross-cmake-env.mk are needed, as I could not find any generated `tc_cmake_vars.mk` file.
``` 
env $(MAKE) --no-print-directory -C ../../toolchain/$(TC) cmake_vars > $(TC_CMAKE_VARS_MK) ; \
``` 

BTW I would prefer to add _tc specific cmake variables_ to the already available `tc_vars.mk` file (if I understand your intentions correctly).

th0ma7:
> I don't know whether the following lines in cross-cmake-env.mk are needed, as I could not find any generated tc_cmake_vars.mk file.

I chosed that naming to be consistant with the existing `TC_VARS_MK` but this might be confusing as it stands.  A few lines up are the following:
```
TC_VARS_MK = $(WORK_DIR)/tc_vars.mk
TC_CMAKE_VARS_MK = $(WORK_DIR)/$(ARCH)-toolchain.cmake
```
I could probably rename it to something else such as `tc_vars.cmake` thus having a variable called `TC_VARS_CMAKE`.

Later-on at cmake configure it gets copied with all the necessary bits for building the specific sub-cross sources.  In there I'm using different variable names (e.g. `spksrc.cross-cmake-env.mk`):
```
CMAKE_TOOLCHAIN_FILE = $(ARCH)-toolchain.cmake
CMAKE_WRK_TOOLCHAIN_FILE = $(WORK_DIR)/$(CMAKE_TOOLCHAIN_FILE)
CMAKE_PKG_TOOLCHAIN_FILE = $(WORK_DIR)/$(PKG_DIR)/$(CMAKE_TOOLCHAIN_FILE)
```
And I just noticed that as CMake by default uses `-DCMAKE_TOOLCHAIN_FILE`.  As such that "naming" should not be used elsewhere to alleviate conflicting variables thus coming up with a different naming convention would be best.

Suggesting the following, at `spksrc.cross-env.mk` level, used at toolchain variable configuration writing:
```
TC_VARS_CMAKE = $(WORK_DIR)/tc_vars.cmake
```
Then at the cross building time, using:
```
CMAKE_TOOLCHAIN_NAME = $(ARCH)-toolchain.cmake
CMAKE_TOOLCHAIN_WRK = $(WORK_DIR)/$(CMAKE_TOOLCHAIN_NAME)
CMAKE_TOOLCHAIN_PKG = $(WORK_DIR)/$(PKG_DIR)/$(CMAKE_TOOLCHAIN_NAME)
```
@hgy59 what do you think?

th0ma7:
@hgy59 I've made the changes so I'll let it run to confirm all still builds OK (with exceptio of `python2` which is totally unrelated).  If all is good I'll merge afterwards.  Again thnx for you review, cheers!

th0ma7:
Merging, thnx again for the in-depth review @hgy59

th0ma7:
What problem have you found?

hgy59:
I got the error described here:
https://stackoverflow.com/questions/65680357/field-ctx-has-incomplete-type-error-when-cross-compiling-libwebsockets

This is related to cmake cross compilation, so I remembered your work done here.

After adding `CMAKE_USE_TOOLCHAIN_FILE = OFF` it successfully compiled, but I got a linker error (related to static library), that I was able to fix with `CMAKE_ARGS += -DLWS_WITH_STATIC=OFF`.
see  b53b37e2de19f4c31350acd6efd872a2d15c9cf5

th0ma7:
I'm glad I kept the old behavior just in case so we have a workaround for exceptions. With this pr i had tested all cross using cmake and out of the bunch only one misbehaved (which didn't surprised me as being tricky in the first place).

hgy59:
> I'm glad I kept the old behavior just in case so we have a workaround for exceptions
Me too ❤️

PR #5298, opened 2022-06-04
Add gstreamer package

th0ma7:
## Description

Add gstreamer package

Fixes #

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update


th0ma7 (commented):
Thnx! :)

cross/gst-plugins-bad/Makefile
```diff
@@ -0,0 +1,86 @@
...
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS = cross/gstreamer cross/gst-plugins-base cross/libxml2
+
+HOMEPAGE = https://gstreamer.freedesktop.org/
+COMMENT  = GStreamer is a library for constructing graphs of media-handling components.
+LICENSE  = LGPLv2.1
+
+# Require access to glib-mkenums Python script
+ENV += PATH=$(STAGING_INSTALL_PREFIX)/bin:$$PATH
```
hgy59, line 15: When this Makefile requires glib-mkenums, I recommend to include `cross/glib` as dependency and not to trust an indirect dependency.

th0ma7:
@hgy59 and fellow @SynoCommunity/developers I've been looking into ways of building `gst-omx` in order to achieve hardware acceleration on platforms that support it.  If a solution turns out working I might even be able to enable `--enable-omx` on `ffmpeg` as well.

I finally found some hints from the VideoStation sources here https://sourceforge.net/projects/dsgpl/files/Packages/DSM%206.1%20Package%20Release/ where Synology create a template file pointing directly to the pre-built OMX library to build up the various tools:
```
$ head gst-omx-virtual-video-station/config/synology/gstomx.conf.template
[omxmpeg2videodec]
type-name=GstOMXMPEG2VideoDec
core-name=__LIB_PREFIX__/libOMX_Core.so
component-name=OMX.realtek.video.dec
rank=258
in-port-index=0
out-port-index=1
sink-template-caps=video/mpeg,mpegversion=(int)[1, 2],systemstream=(boolean)false
src-template-caps=video/x-raw,format = (string) { NV12 },width=(int)[ 16, 4096 ],height=(int)[ 16, 4096 ],framerate=(fraction)[ 0/1, 2147483647/1 ]
```
Where `__LIB_PREFIX__` is:
```
sed "s,__LIB_PREFIX__,${LibPrefix},g" "${GST_OMX_CONF}.template" > "${GST_OMX_CONF}"
```
And `LibPrefix` is:
```
LibPrefix="/var/packages/$PackageName/target/lib/realtek"
```
And `PackageName` is `VideoStation` (probably now `CodecPack` since newer releases).

Long-story-short, I'm thinking of a way to download+extract the official Synology package matching the arch so I could then link to it.  I might even import the library all together which may be even simpler.  If this actually works, it may lead into enabling hardware decoding on many many platforms.

Two questions comes to mind:
1. Where or how to set that up in spksrc?  In a similar manner as we manager kernels/toolchain/toolkit?  or using cross with install_resources and multi-file digests
2. But first and most importantly, I just can't find how to extract Synology latest official spk such as https://archive.synology.com/download/Package/VideoStation/3.0.3-2084 or https://archive.synology.com/download/Package/CodecPack/2.0.0-1017

Your thoughts (and help on extracting spk) would be much appreciated :)

th0ma7:
For item no. 2 (e.g. extracting files from an Synology spk package) I think I have found all the ingredients... Its now a matter of making a short & sweet recipe out of it.

General howto, well made: https://www.cyberciti.biz/faq/pgp-tarball-file-signature-keys-verification/

Actual path where Synology store their keys on the nas:
```
root@th0ma7-nas:/usr/syno/etc/codesign# pwd
/usr/syno/etc/codesign
root@th0ma7-nas:/usr/syno/etc/codesign# ll
total 28
drwxr-xr-x  3 root root 4096 Aug 15  2021 .
drwxr-xr-x 64 root root 4096 Jun 10 10:34 ..
drwx------  2 root root 4096 Jun 10 14:45 .gnupg
-rw-r--r--  1 root root  175 Aug 15  2021 keyinfo-sys
-rw-r--r--  1 root root  224 Jun 10 14:45 keyinfo-user
-rw-r--r--  1 root root 3299 Aug 15  2021 keyring
-rw-r--r--  1 root root    1 Aug 15  2021 VERSION
```
Actual code that is/was used to validate the signed packages and extract them:  https://pastebin.com/3vBYv662

And finally Synology packages: https://archive.synology.com/download/Package

Optionally, this other pgp key might be useful: https://www.synology.com/en-global/support/security_pgp_key

AlexPresso:
Three libraries fail to load when calling `gst-launch-1.0`:
```
root@Kloudy:/var/packages/gstreamer/target/bin$ ./gst-launch-1.0 -s 1200 fakevideosink

(gst-plugin-scanner:30702): GStreamer-WARNING **: 16:32:36.941: Failed to load plugin '/var/packages/gstreamer/target/lib/gstreamer-1.0/libgstcairo.so': libcairo-gobject.so.2: cannot open shared object file: No such file or directory

(gst-plugin-scanner:30702): GStreamer-WARNING **: 16:32:37.000: Failed to load plugin '/var/packages/gstreamer/target/lib/gstreamer-1.0/libgsttaglib.so': /spksrc/spk/gstreamer/work-aarch64-7.0/install/var/packages/gstreamer/target/lib/libz.so: cannot open shared object file: No such file or directory

(gst-plugin-scanner:30702): GStreamer-WARNING **: 16:32:37.081: Failed to load plugin '/var/packages/gstreamer/target/lib/gstreamer-1.0/libgstpng.so': /spksrc/spk/gstreamer/work-aarch64-7.0/install/var/packages/gstreamer/target/lib/libz.so: cannot open shared object file: No such file or directory
```
I've checked the libraries and it seems like they're all present in `/var/packages/gstreamer/target/lib` but gstreamer is searching for them inside `/var/packages/gstreamer/target/lib/gstreamer-1.0`

publicarray:
Would this work as just another cross/ package? That way you don't need to modify the framework and this seems specific to one package, at least for now

th0ma7:
@AlexPresso I was finally able to resume work on this PR.  It isn't yet in a state I like but hardware acceleration on aarch64 using OMX "might" just work but there is no way for me to tell.

Just saying, I hadn't forgotten this... still on my pile :man_shrugging:

th0ma7:
1.20.5 refuses to build on arch using a kernel <= 3.4
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1765

th0ma7:
@AlexPresso @ner00 package is now ready for prime testing.
You can download the resulting zip files from auto-builds of github-action at https://github.com/SynoCommunity/spksrc/actions/runs/4097014300

AlexPresso:
Hi @th0ma7 thank you very much for your work ! I'm so happy to read that !

I tried to manually install the package on a DS218Play (DSM 7.1.1-42962 Update 2) but I get the following error : https://imgur.com/dT7Uf3Q

I'm wondering if it could be because the package is not signed ? Is there any security check to disable somewhere ?

th0ma7:
@AlexPresso lol :)  Is it an armv8 (i.e. aarch64) so you can test hardware acceleration (no clue if this works btw)?

BTW, gathering a bit of documentation here: https://github.com/SynoCommunity/spksrc/wiki/gstreamer

PR #5307, opened 2022-06-09
Update SABnzbd to 3.6.0

Safihre:
Update SABnzbd to 3.6.0.

I also limited the size of the `cross/p7zip` dependency, which is only used by SABnzbd.


cross/p7zip/PLIST
```diff
@@ -1,9 +1 @@
-rsc:bin/7z
-rsc:bin/7za
-rsc:bin/7zr
-lib:lib/p7zip/7z
-lib:lib/p7zip/7za
-lib:lib/p7zip/7zCon.sfx
-lib:lib/p7zip/7zr
-lib:lib/p7zip/7z.so
-lib:lib/p7zip/Codecs/Rar.so
+bin:lib/p7zip/7za
```
th0ma7, line 1: Two questions:
1. The path looks odd?  it's a `bin` in a `lib` folder?  And guessing staticaly compiled to be autonomous?
2. No other packages requires it?

th0ma7 (approved):

cross/p7zip/PLIST
Safihre: 1. Yeah strange indeed, it's just the output of the p7zip build for many years so I didn't look at it or update it.
2. When searching for `cross/p7zip` in the repo, it's only used by SABnzbd.

hgy59: This would not have worked, as `rsc:/bin/7za` was a link to `bin:lib/p7zip/7za`.

I have reworked `cross/p7zip` to build 7za (the default target) only.

hgy59:
@Safihre was sabnzbd ever able to use `7za` that could not be found in the path?

Safihre:
> @Safihre was sabnzbd ever able to use `7za` that could not be found in the path?

It was always able to find it, somehow it was in the `PATH`.

Safihre:
Ah, before it found:
```
2022-04-29 09:55:10,867::INFO::[SABnzbd:471] 7za binary... found (/bin/7z)
```
Now it finds:
```
2022-06-15 13:27:48,704::INFO::[SABnzbd:471] 7za binary... found (/volume1/@appstore/sabnzbd/bin/7za)
```

Thanks @hgy59

Safihre:
@th0ma7 For some reason the build fails on `cryptography`, even though it worked yesterday. 
I can't figure out why.. do you see the difference?
Today: https://github.com/SynoCommunity/spksrc/actions/runs/2501879929
Yesterday: https://github.com/Safihre/spksrc/actions/runs/2474706131

Safihre:
Right, the problem is in `setuptools==62.4.0`, which is installed by `get-pip`. See the changelog: https://github.com/pypa/setuptools/blob/main/CHANGES.rst

Strange, because after `get-pip` we install the old setuptools.. So seems it's keeping some old leftovers?

Safihre:
@th0ma7 Seems we made a mistake, the `get-pip.py "pip==21.3.1"` is not how you can call the script, it always installs the latest pip. 
However, it does have the ability to supply `--no-setuptools --no-wheel` parameters:
https://github.com/Safihre/spksrc/commit/c7348285420f73182c6a7115a34772ec05cf0fb7

But, it still doesn't work 😢 
https://github.com/Safihre/spksrc/runs/6919748449?check_suite_focus=true

Seems maybe the problem is this latest pip version..

th0ma7:
If I recall the only pip being installed as "latest" is at package installation time.
At build time I'll have to investigate a little more as all versions are fixed in `cross/python310` and `native/python310`.
I'll run a few tests while playing with deluge PR to find out.

Safihre:
> If I recall the only pip being installed as "latest" is at package installation time. At build time I'll have to investigate a little more as all versions are fixed in `cross/python310` and `native/python310`. I'll run a few tests while playing with deluge PR to find out.

No, that's not what happens. Check the build logs and it can be seen in the content of `get-pip.py`. It controls which version is installed, and is regularly updated: https://bootstrap.pypa.io/get-pip.py

If we want to control exact the pip version, we should use `python -m ensurepip` and install the bundled old version and then update it to our specified that version.

I think we should also really apply my patch for setuptools and wheel: https://github.com/Safihre/spksrc/commit/c7348285420f73182c6a7115a34772ec05cf0fb7

But it all still doesn't fix cryptography.. If I rerun my *exact* old build that once succeeded (so not a single line of code changed) it now fails.. So something changed upstream.

Safihre:
Ok, fixed it!

PR #4469, opened 2021-03-04
[wip] gcc 4.9.4 package

th0ma7:
_Motivation:_  Intention is to provide an initial `gcc` package in hope of eventually providing a newer version of the full gcc suite.  This provides other opportunities such as replacing entware gcc and ability to build perl bindings using cpan requiring a C compiler.
_Linked issues:_  

### Checklist
- [ ] Build rule `all-supported` completed successfully
- [ ] Package upgrade completed successfully
- [ ] New installation of package completed successfully

Note that due to the large amount of resource files needing to be provided I've create a new `PLIST.auto` functionality.  I allows using an empty `PLIST.auto` in `cross/<package>` where, when captured, the `spksrc.plist.mk` will look at files provided by `$(WORK_DIR)/$(PACK_NAME).plist` using `file` to determine its type and automatically create `PLIST` entries for it.

This should be quite useful for very large packages.  Changes are in commit https://github.com/SynoCommunity/spksrc/pull/4469/commits/c3035644d12046f1865c4d414c15c5258c51e91d

PR #5320, opened 2022-06-20
Add toolchains for DSM 7.1

hgy59:
## Description

- new archs for not yet known models:
  - add arch-broadwellntbap, arch-broadwellnkv2, arch-epyc7002 and arch-r1000 to generic x64 archs
  - add arch-rtd1619b to generic aarch64 archs
- DSM 7.1 comes with gcc 8.50 (except comcerto2k)
- remove obsolete TC_BUILD number from 7.0 Makefiles

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


th0ma7:
Are the toolchain differences sufficiently limited to only deplpy DSM7.0 packages?  Or will DSM7.1 be needed to be built as well?  If the later I fear we may be lacking disk space online.

Also, should we start considering removing DSM6 support?  I mean compiler wise it's a real pain to maintain...

hgy59:
> Are the toolchain differences sufficiently limited to only deplpy DSM7.0 packages? Or will DSM7.1 be needed to be built as well? If the later I fear we may be lacking disk space online.

For me it looks as we are going to drop DSM 7.0 packages some day and only support DSM 7.1 for DSM 7.x. The same ocurred with DSM 6.0, that was dropped for DSM 6.1 for all DSM 6.x.
I should plan this step for end of this year - or earlier when models are available, with the new archs (epyc7002, r1000, rtd1619b).

> Also, should we start considering removing DSM6 support? I mean compiler wise it's a real pain to maintain...

I would not drop DSM 6 support for the next 10 years, as the real benefit of our packages is to support old models, that do not get updates by synology.
I even investigate on ppc853x-5.2, 88f6281-5.2 and x86-5.2 for the models that only support DSM 5.2.
And there are some more models that will not support DSM 7.

With the harder restrictions on DSM 7 there is a good reason to stay on DSM 6 even for models that might be updated.

Further considerations
- In the near future we should enlarge (at least double?) the disk space of the repository
- It is good to drop pure python wheels from related packages to reduce package size, but this excludes usage on air gapped systems (isolated from the internet) - we are facing even problems with systems behind a proxy server
- the synorepo needs some investigations too
  - the main problem is, that DSM 6 packages are displayed on DSM 7 systems when no DSM 7 version is provided
  - without download counters it is not possible to guess whether a package is ever used 
  - @publicarray started to build a (promising) repo based on rust
- some packages are rock solid and support even gcc < 5 for decades (openssl, curl, python, ...) and other packages always use the latest compiler features (like mkvtoolnix)
- it might be a solution to build newer compilers for existing toolchains - but I have no idea wheter this is worth the effort or is possible without updating glib or even the kernel
-we are still facing problems with generic armv7 support for all armv7 cpus. This problem will fade away as new models are released with arm64 or x86_64 compatible cpus only (AFAICS)
- more and more applications support docker and packages for x86_64 cpus get obsolete
- above said, what about a docker package for arm64?
- ...

th0ma7:
> Further considerations
> 
> * In the near future we should enlarge (at least double?) the disk space of the repository

Doubling really is the bare minimum... 

> * It is good to drop pure python wheels from related packages to reduce package size, but this excludes usage on air gapped systems (isolated from the internet) - we are facing even problems with systems behind a proxy server

This is rather easy to do, just one variable set to `1` away from that.  Still disk space remains the primary issue

> * the synorepo needs some investigations too
>   * the main problem is, that DSM 6 packages are displayed on DSM 7 systems when no DSM 7 version is provided
>   * without download counters it is not possible to guess whether a package is ever used
>   * @publicarray started to build a (promising) repo based on rust

Signature could be of interest as well

> * some packages are rock solid and support even gcc < 5 for decades (openssl, curl, python, ...) and other packages always use the latest compiler features (like mkvtoolnix)
> * it might be a solution to build newer compilers for existing toolchains - but I have no idea wheter this is worth the effort or is possible without updating glib or even the kernel

I tried that and failed miserably.  Really that would solve all our issues of being able to update compilers on older platforms to match newer ones.  There are hints on howto here https://github.com/SynoCommunity/spksrc/pull/4469

Supporting older compilers is becoming more and more of an issue.  Further as kernel will now require c11 instead of c99 which in turns might impact many other projects that may lean towards that same direction.

> * we are still facing problems with generic armv7 support for all armv7 cpus. This problem will fade away as new models are released with arm64 or x86_64 compatible cpus only (AFAICS)
> * more and more applications support docker and packages for x86_64 cpus get obsolete
> * above said, what about a docker package for arm64?

Someone had put out an lxc package, there may be hints in there?

hgy59:
merged this to support new models DS1522+ and RS422+ with AMD Ryzen R1600 cpu.

PR #5333, opened 2022-06-28
Add cloudflared

karasevm:
## Description

Cloudflare Zero Trust provides an alternative to traditional VPN solutions without the need to expose your devices to the internet, it's been available on higher end models as a docker image, but this enables it's use on low end hardware.

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update


spk/cloudflared/Makefile
```diff
@@ -0,0 +1,23 @@
+SPK_NAME = cloudflared
+SPK_VERS = 2022.6.3
+SPK_REV = 1
+SPK_ICON = src/cf-logo-v-rgb.png
```
publicarray, line 4: I recommend to change the icon file name to match the package name.

spk/cloudflared/src/service-setup.sh
```diff
@@ -0,0 +1,18 @@
+TOKEN_FILE="${SYNOPKG_PKGVAR}/token"
```
publicarray, line 1: Why not use the official configuration file?

https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/local-management/configuration-file/

publicarray: https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/local-management/ingress/

karasevm: Correct, config file is used in locally managed tunnels and the token is used in tunnels managed through the dashboard. As far as I understand, remote management is the recommended set up. The service command is the same that's used in the systemd service created by following the official documentation.  

It seems there are also ways to use a credentials file with remote management, but they involve an interactive setup, passing premade configs, or using the API (related issues https://github.com/cloudflare/cloudflared/issues/645, https://github.com/cloudflare/cloudflared/issues/665).

spk/cloudflared/Makefile
karasevm: Done

publicarray:
Awesome work! Thanks for the PR, I'm just not 100% sure on the name because cloudflared does more than create a tunnel but on the other hand that is probably the most common use case.

PR #5352, opened 2022-07-05
synocli-file: fix include of nnn and rnm tools (#5351)

### LINKED ISSUE #5351: nnn 4.5 is not inside SynoCli File Tools v2.6-15 package as it is supposed to be
### Is this a new Bug?

### Package Name

SynoCli File Tools

### Package Version

2.6-15

### Device Model

Synology DS218+

### Device Architecture

x86_64

### Firmware Version

DSM 6.2.4 build 25556 update 6

### What happened?

nnn 4.5 is supposed to be inside the SynoCli File Tools v2.6-15 package, but it is not.

### Reproduction steps

```bash
1. Install SynoCli File Tools v2.6-15
2. Run a shell
3. Type "nnn"
```

### Install Log

```log
I have no longer the install log since I uninstalled SynoCli File Tools v2.6-15 and reinstalled SynoCli File Tools v2.5.13 (to get back "nnn")
```

### Service Log

_No response_

### Other Logs

_No response_

hgy59:
## Description

- unfortunately version_gcc is not available in spk/ Makefile
- use DSM version to evaluate supported gcc for rnm and nnn

Fixes #5351

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/spk/synocli-file/Makefile
+++ b/spk/synocli-file/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = synocli-file
 SPK_VERS = 2.6
-SPK_REV = 15
+SPK_REV = 16
 SPK_ICON = src/synocli-file.png
 
 # packages depending on cross/zlib must be defined later, see below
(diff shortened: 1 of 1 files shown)


hgy59:
@th0ma7 `version_gcc` does not work before the toolchain is extracted.
Probably we should go back to define GCC_VERSION in the toolchain Makefiles.
In most situations we need to get GCC_VERSION from TCVERSION and ARCH before any make target is processed.

th0ma7:
> @th0ma7 `version_gcc` does not work before the toolchain is extracted.
> Probably we should go back to define GCC_VERSION in the toolchain Makefiles.
> In most situations we need to get GCC_VERSION from TCVERSION and ARCH before any make target is processed.

I'd rather not, really.  In what case you'd need that for?  And could the check be done at a later time instead?

Issue #5081, opened 2022-01-16
Update Lidarr to use other TMP directory like Radarr


bakerboy448:
### Is this a new Bug?

- [X] I checkd that the bug hasn't been reported before

### Package Name

Lidarr

### Package Version

v20210322-7

### Device Model

Agnostic

### Device Architecture

x86_64

### Firmware Version

DSM7

### What happened?

USE_ALTERNATE_TMPDIR = 1 Is required  to be added due to the small /tmp directory causing update and database issues

https://github.com/Lidarr/Lidarr/issues/2583

Radarr Syno fix https://github.com/SynoCommunity/spksrc/commit/5a174ae655388df9bbbf6d9bb19a8f16662a9fcb

Radarr updater issue

https://github.com/SynoCommunity/spksrc/issues/4755
https://github.com/SynoCommunity/spksrc/issues/4621

### Reproduction steps

```bash
Known  issue

```

### Install Log

```log
Known issue
```

mreid-tt:
hey @bakerboy448, I went ahead and created PR https://github.com/SynoCommunity/spksrc/pull/5446 which implements this fix for an alternate TMP directory as it also resolves an issue with in-app updates (https://github.com/SynoCommunity/spksrc/issues/5303). The PR also brings the version of Lidarr up to the latest release (version 1.0.2).

hgy59:
Closed by #5446

PR #5400, opened 2022-08-21
Update Prowlarr v0.4.10.2111

wkobiela:
## Description

Update prowlarr - v0.4.10.2111

## Checklist

### Type of change

- [x] Package update 

Tested on DSM7 on virtualDSM and both fresh install and upgrade works fine


hgy59:
@wkobiela we don't create new packages for packages that use an internal updater. (`SPK_VERS = $(shell date +%Y%m%d)` shows that the package is not bound to a specific version, as it uses an internal updater).

Exceptions of this rule are:
- when the internal updater does not work anymore due to new dependencies
- when new archs are added that are not included in the released packages
- the DSM service integration needs an update (folders, user, port, etc.)


So my proposal for prowlarr:
- update unsupported archs (related to #5383)
- update to use alternate TMP dir (related to #5081)

wkobiela:
It seems that prowlarr is already using alternate TMP dir: https://github.com/SynoCommunity/spksrc/blame/master/spk/prowlarr/Makefile#L38

wkobiela:
@hgy59 do I also need to add `UNSUPPORTED_ARCHS` line in `spk/prowlarr/Makefile`? I've seen it in Radarr files and also in Jellyfin

bakerboy448:
> we don't create new packages for packages that use an internal updater.

This doesn't make any sense for the \*arrs.

So you're saying SynoComm policy is that the \*arrs packages should never be updated?  That means users will always have problems when they reinstall or uninstall and reinstall without purging existing app data.

the old versions that SynoComm provides are not compatible with the new databases and thus the apps will fail to function and users must manually purge their data which is a complicated task for some users

so are there no plans to land this PR then?

bakerboy448:
May as well update this to 0.4.7 rather than having it be an older version.
https://github.com/Prowlarr/Prowlarr/releases

Safihre (approved):

bakerboy448:
And may as well Bump to 0.4.8 as .2016 had Cardigann issues. Or just land as is and follow up with the updater rework and bump there

hgy59:
@bakerboy448 the automatic update on service start does not work in Prowlarr (it does work in Lidarr).
I have created an issue there (https://github.com/Prowlarr/Prowlarr/issues/1199)

hgy59:
@wkobiela I did some rework. Except the automatic update after package installation or package update, everything seems to be ok.
Can you have a look at it?

spk/prowlarr/src/wizard/upgrade_uifile
```diff
@@ -1,6 +1,6 @@
 [{
     "step_title": "Updating Prowlarr",
     "items": [{
-        "desc": "Keep Prowlarr up-to-date by using Prowlarr's built-in updater.<br>Navigate to System>Updates in the Prowlarr UI."
+        "desc": "Keep Prowlarr up-to-date by using Prowlarr's built-in updater.<br>Navigate to System>Updates in the Prowlarr UI.<br/><br/><b style=\"color: red\">Due to an error in Prowlarr, the automatic update in service start does not work, and you have to start the update manually.</b>"
```
bakerboy448, line 4: Id suggest referring to the prowlarr GHI # for easy tracking and future reference

spk/prowlarr/src/wizard/upgrade_uifile
```diff
@@ -1,6 +1,6 @@
 [{
     "step_title": "Updating Prowlarr",
     "items": [{
-        "desc": "Keep Prowlarr up-to-date by using Prowlarr's built-in updater.<br>Navigate to System>Updates in the Prowlarr UI.<br/><br/><b style=\"color: red\">Due to an error in Prowlarr, the automatic update in service start does not work, and you have to start the update manually.</b>"
+        "desc": "Keep Prowlarr up-to-date by using Prowlarr's built-in updater.<br>Navigate to System>Updates in the Prowlarr UI.<br/><br/><b style=\"color: red\">Due to an error in Prowlarr, the automatic update in service start does not work, and you have to start the update manually.</b><a href="https://github.com/Prowlarr/Prowlarr/issues/1199">Issue #1199</a>"
```
bakerboy448, line 4: lint failed

`spk/prowlarr/src/wizard/upgrade_uifile: line 4, col 300, found: 'INVALID' - expected: 'EOF', '}', ':', ',', ']'.`

wkobiela: Yeah I know, not sure how to include link in upgrade_uifile, probably just lack of backslash

bakerboy448:
the upgrade did not upgrade based on the version in system => status

hgy59:
@wkobiela as we now have the latest version in the package, you cannot see any update in Powlarr.
You only see the file `/var/packages/prowlarr/var/.config/Prowlarr/update_required` that is not deleted as service start.

The automatic update at startup will only work, when there is a version available, that has fixed this and you manually updated to such a version (or we include the fixed version in the package).

But first we need the confirmation that this will be fixed.

hgy59:
As https://github.com/Prowlarr/Prowlarr/issues/1199 is fixed now, we only need to wait for a new release of Prowlarr.

wkobiela:
I'm not sure that automatic update at startup is working, upgraded package manually and Prowlarr is still on previous version
image

wkobiela:
Configured automatic update and updated package manually, still on last version 
image

hgy59:
@wkobiela as I explained in #5511 the automatic update will only work in the future, i.e. when you update the Version 0.4.10.2111 to a future Version.

But you can verfy that the update mechanism is working. When you install v0.4.10.2111 again (over 0.4.10.2111) and the update_required file in the config folder is gone after this "update" than it will work with future versions too.

hgy59 (approved):

PR #5398, opened 2022-08-20
deluge 2 upgrade to python 3.10 fix for DSM7 (follow-up #4153)

th0ma7:
Co-Authored-by: hgy59 <hpgy59@gmail.com>
Co-Authored-by: Christophe David <david.christophe@protonmail.com>

## Description

Upgrade deluge and migrate to DSM7

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

### Testing
- [x] `x64-6.1` migrate/upgrade | install | upgrade
- [x] `x64-7.0` install | upgrade
- [x] `armv7-7.0` install | upgrade

PR #5410, opened 2022-08-29
Sabnzbd update to version 3.61

th0ma7:
## Description

This is a follow-up to #5398 which includes changes to the `service-setup.sh` script.  This update will ensure that the permission adjustments are being propagated to next package release.

Fixes n/a

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update

### DIFF
--- a/cross/par2cmdline/patches/88f6281/c++11-to-c++0x.patch
+++ b/cross/par2cmdline/patches/88f6281/c++11-to-c++0x.patch
@@ -1,0 +1,24 @@
+diff -uprN ../par2cmdline-0.8.1.orig/Makefile.am ./Makefile.am
+--- ../par2cmdline-0.8.1.orig/Makefile.am	2020-02-09 14:29:09.000000000 +0000
++++ ./Makefile.am	2022-08-29 10:53:46.532922839 +0000
+@@ -57,7 +57,7 @@ par2_SOURCES = src/par2cmdline.cpp \
+ par2_LDADD = libpar2.a
+ 
+ LDADD = -lstdc++
+-AM_CXXFLAGS = -Wall -std=c++11 $(OPENMP_CXXFLAGS)
++AM_CXXFLAGS = -Wall -std=c++0x $(OPENMP_CXXFLAGS)
+ 
+ EXTRA_DIST = PORTING ROADMAP \
+ 			 man/par2.1 \
+diff -uprN ../par2cmdline-0.8.1.orig/Makefile.in ./Makefile.in
+--- ../par2cmdline-0.8.1.orig/Makefile.in	2020-02-09 14:54:10.000000000 +0000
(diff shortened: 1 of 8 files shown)


Safihre:
Nice!

Safihre (approved):

cross/par2cmdline/patches/88f6281/c++11-to-c++0x.patch
```diff
@@ -0,0 +1,24 @@
+diff -uprN ../par2cmdline-0.8.1.orig/Makefile.am ./Makefile.am
+--- ../par2cmdline-0.8.1.orig/Makefile.am	2020-02-09 14:29:09.000000000 +0000
++++ ./Makefile.am	2022-08-29 10:53:46.532922839 +0000
+@@ -57,7 +57,7 @@ par2_SOURCES = src/par2cmdline.cpp \
+ par2_LDADD = libpar2.a
+ 
+ LDADD = -lstdc++
+-AM_CXXFLAGS = -Wall -std=c++11 $(OPENMP_CXXFLAGS)
++AM_CXXFLAGS = -Wall -std=c++0x $(OPENMP_CXXFLAGS)
```
Safihre, line 9: Python 3.11 requires C++17, so for a moment I thought this would exclude the 88f6281 arch when Python 3.11 is released. 
But seems it should still work, if the right flags are used.

hgy59: C++17 requirement will drop all DSM <7 builds.

Safihre: Typo, meant C++11!

th0ma7: even `c++11` will require dropping `gcc<4.9`.

BTW @Safihre I'm pushing this PR as I noticed a bug in `service-setup.sh` where the `incomplete`, `complete` and `watch` directories under `downloads` are not group write permitted which impacts other packages.  I've come-up with a method to alleviate that even on DSM7 such as https://github.com/SynoCommunity/spksrc/pull/5398/files#diff-0be67c7c2427a76d082fddbe7509dbf6111868c62f50d755bb1318131fa7dc1a

Therefore I took the opportunity to create an updated package so any new install do not block other packages.

But other than that this package served me as a perfect example to nail down `deluge` hopefully once and for all :)

th0ma7:
@Safihre would you have spare cycles to give this one a shot to confirm all is looking good?

Safihre:
Tested and it works, only tested DSM 7.

th0ma7:
> Tested and it works, only tested DSM 7.

Thnx, much appreciated!

PR #5414, opened 2022-09-05
imagemagick: update and support for additional image formats and fonts 

hgy59:
## Description

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

Issue #5401, opened 2022-08-21
version_gcc nested using version_* within ifeq not working


th0ma7:
### What happened?

The following does not work as expected and always turns into `else`:
```
ifeq ($(call version_ge, $(call version_gcc), 4.9),0)
...
else
...
fi
```

### Reproduction steps

```bash
ifeq ($(call version_ge, $(version_gcc), 7.5),1)
        @$(MSG) 7.5+
else ifeq ($(call version_ge, $(call version_gcc), 4.9),0)
        @$(MSG) 4.9+
else
        @$(MSG) -4.9
endif
```

While playing with the code I came out with a newer version of `version_gcc` that always works such as:
```
version_gcc = $(lastword $(shell grep TC_GCC $(PWD)/work-$(lastword $(subst =, ,$(filter ARCH=%,$(MAKEFLAGS))))-$(lastword $(subst =, ,$(filter TCVERSION=%,$(MAKEFLAGS))))/tc_vars.mk 2>/dev/null))
```

Note that the following call works as expected, making this more strange as it seems to only affect it under `ifeq`:
```
        @$(MSG) check 4.9: [$(call version_ge, $(version_gcc), 4.9)]
        @$(MSG) check 7.5: [$(call version_ge, $(version_gcc), 7.5)]
```

th0ma7:
@hgy59 have you encountered this?

hgy59:
AFAICR I saw such an issue when inverting the if...
My concern, for definition of gcc version dependent package support or gcc version related package dependencies, this solution does not work, as it requires the build (i.e. extraction) of the toolchain is done before.

th0ma7:
Indeed there is such a case but the weird part here is that a direct calls as expected:
```
@$(MSG) check 4.9: [$(call version_ge, $(version_gcc), 4.9)]
```
But when used within a `ifeq` it doesn't anymore...
```
ifeq ($(call version_gcc), 4.9.3)
```

th0ma7:
I found out why... the `ifeq` statement is being re-evaluated multiple times until the actual execution of that specific piece of code.  Playing with debugging output I was able to gather theses states during runtime:
```
1  MAKEFLAGS:  MAKECMDGOALS: spkclean TCVERSION: 5.2 ARCH:  WORK_DIR: /home/spksrc/deluge-th0ma7/spksrc/spk/deluge/work-all  4.9
[...]

PR #5417, opened 2022-09-06
Avoid default value for TCVERSION (followup to #5390)

hgy59:
## Description

The error was added by #5390
- avoid to set a default value for TCVERSION, as it might be evaluated later
- fixes the build of packages with REQUIRED_MIN_DSM (like mkvtoolnix, nodejs_v16)

Fixes mkvtoolnix and nodejs_v16 builds in #5414

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -54,19 +54,20 @@ else
 # different noarch packages
 SPK_ARCH = noarch
 SPK_NAME_ARCH = noarch
-ifeq ($(strip $(TCVERSION)),)
-# default: 3.1 .. 5.2
-TCVERSION = 5.2
-endif
+ifneq ($(strip $(TCVERSION)),)
 ifeq ($(call version_ge, $(TCVERSION), 7.0),1)
 SPK_TCVERS = dsm7
 TC_OS_MIN_VER = 7.0-40000
 else ifeq ($(call version_ge, $(TCVERSION), 6.1),1)
 SPK_TCVERS = dsm6
 TC_OS_MIN_VER = 6.1-15047
-else ifeq ($(call version_lt, $(TCVERSION), 3.0),1)
+else ifeq ($(call version_ge, $(TCVERSION), 3.0),1)
+SPK_TCVERS = all
+TC_OS_MIN_VER = 3.1-1594
+else
 SPK_TCVERS = srm
 TC_OS_MIN_VER = 1.1-6931
+endif
 else
 SPK_TCVERS = all
 TC_OS_MIN_VER = 3.1-1594


hgy59:
@th0ma7 it is always a mystery how make variables get evaluated...

th0ma7:
@hgy59 I played quite a bit with that and still there's a lot of black magic going on.  But I'll try to explain my findings (and please feel free to correct me as it goes).

First off, depending on how the `make` call is being made, it will result in multiple (re)evaluation of the variables.  For instance, calling `make all-supported` ends-up calling in parallel `make arch-<ABC>-X.Y` while emptying it's environment variables (although it does pass arguments).  This in turns calls `make TCVERSION=<version> ARCH=<arch>`, again without passing any environment variables (e.g. `-@MAKEFLAGS`).  But, forgot to mention, this in turns will also be evaluated through `dependency-list` and `dependency-flat` as they are used to make all `native` dependencies first and only then move on with the parallel building.

So lets take the `version_gcc` call that I recently removed (#5401) as this was too error prone, and lets say it is in a given `Makefile`: before getting anything done, 5x evaluation of that resulting value will have occured when using `make all-supported`, all resulting in bogus return value as the toolchain hasn't yet been installed nor a `tc_var.mk` generated (and `tc_vars.cmake` for what's worth).  Now here's the trick with `version_gcc` in particular, if you are re-invoking over an existing build (e.g. with an existing `$(WORK)` directory) the environment is still there and therefore the return value may well be right this time :)

Now is the interesting part, the environment such as ran using `$(RUN)` will have all you need in there e.g. you can use and abuse of any variable using a shell `${TC_GCC}`.  Now the question is: when can I make use of this?  The only "safe" place is after the `include ../../mk/spksrc.spk.mk` or `include ../../mk/spksrc.cross-*.mk`.  Reason is: it will have processed the `spksrc.tc.mk`, resulting in a valid environment at that stage only.  The caveat is that this might not meet all the usecases you have in mind where you wanted to validate something really early in your makefile.

So in turns this then mean that you would only adjust dependencies based on a shell environment value after the `include`.  Doing it before then may result in odd results.

Now, for the exception part, `TCVERSION` and `ARCH`.  Theses gets evaluated really early in the process as they are being used to locate and make use of the `toolchain/syno-<arch>-<version>`.  Also, theses gets evaluated as Makefile variables using `$(ARCH)` and `$(TCVERSION)` to the contrary of `tc_vars.mk` variables which are shell based.

In your usecase for your patch, the value of `$(TCVERSION)` will be empty by default when called using `make all-supported` or `make arch-<arch>-<version>` until it calls itself once more transforming it as `make TCVERSION=<version> ARCH=<arch>`.  Question is: will setting a default value be usefull or change it's behavior? probably not.  Reason is, it isn't needed for `all-supported` as it will gather the proper arch list values from `SUPPORTED_ARCHS` and `AVAILABLE_TCVERSIONS` from `spksrc.common.mk` and recall itself using multiple time in parallel using `arch-<arch>-<version>`.  And is it needed for the later?  no again as it will extract that info from the arguments passed to recall itself with  `make TCVERSION=<version> ARCH=<arch>`.  So being empty is OK until it's not.

Hope this helps or have I muddied the water?

hgy59:
@th0ma7 thanks a lot for the detailed findings (and writing it down).

BTW: my intent was to initialize TCVERSION only for noarch builds. _But it affected the prechecks for `REQUIRED_MIN_DSM`._

PR #5399, opened 2022-08-20
Add matrix-synapse (includes rust wheel build)

hgy59:
## Description

This is a package with cryptography wheel (cross compiled with rust)

Related to 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5435, opened 2022-09-20
python wheels using rustc

th0ma7:
## Description

python wheels using rustc - Using another approach from #5399

Extras: 
- [x] Update `python310` to version 3.10.7 (possibly 3.10.8? https://peps.python.org/pep-0619/)
- [x] Update `duplicity` to version 0.8.23
- [x] Fix `mysql-connector-c` which now fails to build
- [x] As needing to update `duplicity`, shoot for version 1.0.1 (tested on `x64-7.0`, `armv7-7.0`, `armv5-6.1`)

TODO: none left

Fixes #

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


th0ma7:
@publicarray reading back on one of your comments https://github.com/SynoCommunity/spksrc/pull/4971#issuecomment-968485250 : Any changes to the `Dockerfile` won't actually be ran unless merged to master?

Safihre:
> @publicarray reading back on one of your comments [#4971 (comment)](https://github.com/SynoCommunity/spksrc/pull/4971#issuecomment-968485250) : Any changes to the `Dockerfile` won't actually be ran unless merged to master?

Yes, but it's also only build when it's on `master`, so it's not really tested.

publicarray:
> > @publicarray reading back on one of your comments [#4971 (comment)](https://github.com/SynoCommunity/spksrc/pull/4971#issuecomment-968485250) : Any changes to the `Dockerfile` won't actually be ran unless merged to master?
> 
> 
> 
> Yes, but it's also only build when it's on `master`, so it's not really tested.

Yea I suppose the idea here was that you do build the docker container locally first and verify it before pushing the changes to master. However you can create a docker build in your fork if that is easier. yea it's not a good idea to do docker builds in PRs as we don't want to modify the published image until the changes land in master.

miigotu:
Is the rust env getting sourced?

Thos is how I do it in our docker file:
https://github.com/SickChill/SickChill/blob/master/contrib/docker/Dockerfile.rpi#L37

Dockerfile
```diff
@@ -86,6 +86,20 @@ RUN wget https://bootstrap.pypa.io/get-pip.py -O - | python3
...
+RUN CARGO_HOME=/usr/local rustup default stable
+# Install all rustc targets
+RUN CARGO_HOME=/usr/local rustup target install x86_64-unknown-linux-gnu
+RUN CARGO_HOME=/usr/local rustup target install i686-unknown-linux-gnu
+RUN CARGO_HOME=/usr/local rustup target install armv5te-unknown-linux-gnueabi
+RUN CARGO_HOME=/usr/local rustup target install armv7-unknown-linux-gnueabihf
+RUN CARGO_HOME=/usr/local rustup target install armv7-unknown-linux-gnueabi
+RUN CARGO_HOME=/usr/local rustup target install aarch64-unknown-linux-gnu
+RUN CARGO_HOME=/usr/local rustup target install powerpc-unknown-linux-gnu
+
```
miigotu, line 102: Is there a difference between
rustup target add $RUST_TARGET
And
rustup target install?

miigotu: I think the install command is for the `rustup toolchain install ` and for targets you do `rustup target add`, where the toolchain is obviously for the host doing the build, and the target is the end target when cross compiling.

th0ma7: Related to your first comment, and looking at `rustup target|toolchain --help`:
```
spksrc@spksrc:/tmp$ rustup toolchain --help
rustup-toolchain 
Modify or query the installed toolchains

USAGE:
    rustup toolchain <SUBCOMMAND>

FLAGS:
    -h, --help    Prints help information

SUBCOMMANDS:
    list         List installed toolchains
    install      Install or update a given toolchain
    uninstall    Uninstall a toolchain
    link         Create a custom toolchain by symlinking to a directory
    help         Prints this message or the help of the given subcommand(s)
```
vs `target`:
```
$ rustup target --help
rustup-target 
Modify a toolchain's supported targets

USAGE:
    rustup target <SUBCOMMAND>

FLAGS:
    -h, --help    Prints help information

SUBCOMMANDS:
    list      List installed and available targets
    add       Add a target to a Rust toolchain
    remove    Remove a target from a Rust toolchain
    help      Prints this message or the help of the given subcommand(s)
```
So while `install` does work, the actual call should be `add`.  Will change that.

miigotu: Yeah I wasn't looking at the actual rust clip help, I was reading their docs. They worded it a little difference on the rust book, so the toolchain I assumed was to build host tools, and the targets were for cross.


The env file just sets some variables for the includes and linker and path afaik.

mk/spksrc.tc.mk
```diff
@@ -121,6 +123,12 @@ tc_vars:
 	echo TC_ENV += CPPFLAGS=\"$(CPPFLAGS) $$\(ADDITIONAL_CPPFLAGS\)\" ; \
 	echo TC_ENV += CXXFLAGS=\"$(CXXFLAGS) $$\(ADDITIONAL_CXXFLAGS\)\" ; \
 	echo TC_ENV += LDFLAGS=\"$(LDFLAGS) $$\(ADDITIONAL_LDFLAGS\)\" ; \
+	echo TC_ENV += CARGO_HOME=\"/usr/local\" ; \
+	echo TC_ENV += CARGO=\"/usr/local/bin/cargo\" ; \
+	echo TC_ENV += RUSTC=\"/usr/local/bin/rustc\" ; \
+	echo TC_ENV += CARGO_TARGET=\"$(RUST_TARGET)\" ; \
```
messense, line 129:
```suggestion
	echo TC_ENV += CARGO_BUILD_TARGET=\"$(RUST_TARGET)\" ; \
```

mk/spksrc.tc.mk
```diff
@@ -121,6 +123,12 @@ tc_vars:
 	echo TC_ENV += CPPFLAGS=\"$(CPPFLAGS) $$\(ADDITIONAL_CPPFLAGS\)\" ; \
 	echo TC_ENV += CXXFLAGS=\"$(CXXFLAGS) $$\(ADDITIONAL_CXXFLAGS\)\" ; \
 	echo TC_ENV += LDFLAGS=\"$(LDFLAGS) $$\(ADDITIONAL_LDFLAGS\)\" ; \
```
messense, line 125:
```
        = note: /usr/bin/ld: /spksrc/toolchain/syno-armv7-7.0/work/arm-unknown-linux-gnueabi/bin/../arm-unknown-linux-gnueabi/sysroot/usr/libexec/../lib/Scrt1.o: relocations in generic ELF (EM: 40)
```

I think this error is caused by the `CFLAGS`/`CXXFLAGS` and etc. When rustc build and link the build script, it wrongly uses these target compiler/linker flags.

messense: See also https://github.com/yoctoproject/poky/blob/e6a9aab3ef52cfcbfbad17ad9de2e142ae32654f/meta/recipes-devtools/rust/rust.inc#L179-L185

th0ma7: @messense thnx a lot for spending a few cycles on this, your advices are much appreciated.

I gave that option a try and it failed with the exact same error.  I wonder, is there a way I can use a `config.toml` while building from `pip`?  I've tried various options but didn't found any that made it being handled (maybie due to pip building in a random `/tmp/*` directory).  Perhaps there is a method to pass the config file to pip using a global option of some sort?  This would ensure the environment is being set properly, always, for any given arch being built in parallel.

messense: > I wonder, is there a way I can use a `config.toml` while building from `pip`?

I don't think there is as of now, but most of cargo's configuration can be set by environment variable: https://doc.rust-lang.org/cargo/reference/config.html

messense: You can add a global `config.toml` in `CARGO_HOME` directory though.

miigotu:
So, I've been trying to get 32 bit arms builds working using docker buildx, and came across the same issue. There is a readdir bug in qemu for 32 bit arm that makes it's way all the way into the crossenv. This link is a docker build but I think the solution might be a similar problem in this cross toolchain. Also, rustup can't tell the architecture with an 32 bit arm architecture, so it can never find the correct compiler.

The fix for this is a mixture of the solution in the link here, and a patch to rustup I had done awhile back (I'll post that when I update the patch and test these fixes in miigotu/python/docker/images

https://github.com/matrix-org/synapse/issues/9403#issuecomment-780165457

th0ma7:
> So, I've been trying to get 32 bit arms builds working using docker buildx, and came across the same issue. There is a readdir bug in qemu for 32 bit arm that makes it's way all the way into the crossenv. This link is a docker build but I think the solution might be a similar problem in this cross toolchain. Also, rustup can't tell the architecture with an 32 bit arm architecture, so it can never find the correct compiler for arm6 and 7 and sends the objects to the wrong linker for arm 5/6

That's quite an interesting catch, although, the issue is also hapening on other archs and not only `armv7`.  I can reproduce a rather similar output using i686 (`arch-evansport-7.0`) and `aarch64` for instance.  All leading to a linking time issue.

miigotu:
Yeah I have the same issue with 386/686. I think it is related.

That's part of the patch for rustup is to get it to recognize the bitness as 32, it detects a 32 bit arch at 64 bit currently. I'll go quick and see if I can find my bug report from last year.

miigotu:
https://github.com/rust-lang/rustup/issues/2700

miigotu:
https://github.com/rust-lang/rustup/issues/2700

mk/spksrc.cross-rust.mk
```diff
@@ -31,93 +31,38 @@ TC = syno$(ARCH_SUFFIX)
...
-RUST_TARGET=powerpc-unknown-linux-gnu
-endif
-ifeq ($(RUST_TARGET),)
-$(error Arch $(ARCH) not supported)
-endif
-
-# Use distrib folder as cargo download cache
-CARGO_HOME_PATH=/spksrc/distrib/cargo
-ENV += CARGO_HOME=$(CARGO_HOME_PATH)
-ENV += PATH=:$(CARGO_HOME_PATH)/bin/:$(PATH)
```
miigotu, line 76: This is the contents of $CARGO_HOME/env:
```
#!/bin/sh
# rustup shell setup
# affix colons on either side of $PATH to simplify matching
case ":${PATH}:" in
    *:"$HOME/.cargo/bin":*)
        ;;
    *)
        # Prepending path in case a system-installed rustc needs to be overridden
        export PATH="$HOME/.cargo/bin:$PATH"
        ;;
esac

```

Adding the bin $CARGO_HOME/bin dir to the path is necessary

miigotu:
Man this is super annoying, this works locally to build armv7, but failed in github actions:
https://github.com/miigotu/python-docker-images/actions/runs/3156651360/workflow
from this dockerfile: https://github.com/miigotu/python-docker-images/blob/main/python/python-cryptography.dockerfile
`docker buildx build --platform linux/arm/v7 -f python/python-cryptography.dockerfile . --progress plain --ulimit nofile=262144:262144 --output=type=local,dest=/tmp/cryptography-wheels`
image

mk/spksrc.cross-rust.mk
th0ma7: Indeed but as the `CARGO_HOME` is being set to `/usr/local/` therefore `/usr/local/bin` is automagically being in the default `PATH`.

miigotu:
Same dockerfile works locally for all of them smh
`docker buildx build --platform linux/386,linux/amd64,linux/arm/v5,linux/arm/v7,linux/arm64/v8,linux/ppc64le,linux/s390x -f python/python-cryptography.dockerfile . --progress plain --ulimit nofile=262144:262144 --output=type=local,dest=wheels`

[cryptography-wheels.zip](https://github.com/SynoCommunity/spksrc/files/9686783/cryptography-wheels.zip)

th0ma7:
First thing to notice in here, the example you're refering to is using `venv` instead of `crossenv`.  Something I'm not sure I understand (still ramping-up on docker), is it starting a docker in qemu where host & target are the same?  Or is it cross-compiling from a `x86_64` host?

Anyhow, having a few pending patches on my end and trying to build using spksrc my "updated" docker environment where the host is `x86_64`.

My first thing to try is, is it even cross-compiling using rust?  Lets take `ripgrep` as an easy app to build:
```
# CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" cargo install ripgrep --target=aarch64-unknown-linux-gnu --force
# file rg
rg: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, with debug_info, not stripped
```
It works!  Now let's try that using `CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu` instead of enforcing `--target`:
```
# CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" cargo install ripgrep --force
# file rg
rg: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, with debug_info, not stripped
```
Still working!  So the theory is that I have a working `rustc` along with a proper recognition of my target `gcc` as default linker and `CARGO*` and `RUSTUP*` environment variables are also recognized and functional.


Now, following your Dockerfile example, along with using host default python3, still on `x86_64` host:
```
# python3 --version
Python 3.9.2
# apt-get update
# apt-get install python3-venv
# VENV_PATH=/opt/venv python3 -m venv $VENV_PATH --upgrade
# VENV_PATH=/opt/venv python3 -m pip install --upgrade setuptools pip
# VENV_PATH=/opt/venv python3 -m pip install --upgrade setuptools-rust wheel
# CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" PYO3_CROSS_LIB_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/lib/ PYO3_CROSS_INCLUDE_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/include/ VENV_PATH=/opt/venv python3 setup.py bdist_wheel --py-limited-api=cp36
```
Boom! working!  But hold-on, why is the wheel extension being `x86_64`?
```
# ll dist/cryptography-38.0.1-cp36-abi3-linux_x86_64.whl
-rw-r--r-- 1 root root 1214001 Sep 30 19:46 dist/cryptography-38.0.1-cp36-abi3-linux_x86_64.whl
```
After investigating a little further, looks like everything in the wheel file is ?OK?:
```
# file ./hazmat/bindings/_rust.abi3.so
./hazmat/bindings/_rust.abi3.so: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

Now let's try out the exact same thing, this time using our host native built python3.10:
```
# VENV_PATH=/opt/venv /spksrc/native/python310/work-native/install/usr/local/bin/python3 -m venv $VENV_PATH --upgrade
# VENV_PATH=/opt/venv /spksrc/native/python310/work-native/install/usr/local/bin/python3 -m pip install --upgrade setuptools pip
# VENV_PATH=/opt/venv /spksrc/native/python310/work-native/install/usr/local/bin/python3 -m pip install --upgrade setuptools-rust wheel

# CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" PYO3_CROSS_LIB_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/lib/ PYO3_CROSS_INCLUDE_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/include/ VENV_PATH=/opt/venv /spksrc/native/python310/work-native/install/usr/local/bin/python3 setup.py bdist_wheel --py-limited-api=cp36
```
Again, boom!  Although it works with a few deprecation warnings:
```
build/temp.linux-x86_64-cpython-310/cryptography.hazmat.bindings._openssl.c: In function ‘_cffi_d_ASN1_STRING_data’:
build/temp.linux-x86_64-cpython-310/cryptography.hazmat.bindings._openssl.c:7957:3: warning: ‘ASN1_STRING_data’ is deprecated [-Wdeprecated-declarations
```
Let's investigate further:
```
# ll dist/cryptography-38.0.1-cp36-abi3-linux_x86_64.whl
-rw-r--r-- 1 root root 2040082 Sep 30 20:10 dist/cryptography-38.0.1-cp36-abi3-linux_x86_64.whl
```
After unziping:
```
# find . -name "*.so"
./cryptography/hazmat/bindings/_openssl.abi3.so
./cryptography/hazmat/bindings/_rust.abi3.so
# file ./cryptography/hazmat/bindings/_openssl.abi3.so ./cryptography/hazmat/bindings/_rust.abi3.so
./cryptography/hazmat/bindings/_openssl.abi3.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=a1db744141c71550a25dac57f63994c81be4cf90, with debug_info, not stripped
./cryptography/hazmat/bindings/_rust.abi3.so:    ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```
So clearly a mismatch here.  But it did built sort of "succesfully" using a `venv`.  But clearly this wheel output is useless as having arch mismatchs at every levels.  But it did demonstrated that it can build, which is already a start.

Now, challenge really is, building using a `crossenv` ...

miigotu:
> First thing to notice in here, the example you're refering to is using `venv` instead of `crossenv`.  Something I'm not sure I understand (still ramping-up on docker), is it starting a docker in qemu where host & target are the same?  Or is it cross-compiling from a `x86_64` host?
> 
> Anyhow, having a few pending patches on my end and trying to build using spksrc my "updated" docker environment where the host is `x86_64`.
> 
> My first thing to try is, is it even cross-compiling using rust?  Lets take `ripgrep` as an easy app to build:
> ```
> # CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" cargo install ripgrep --target=aarch64-unknown-linux-gnu --force
> # file rg
> rg: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, with debug_info, not stripped
> ```
> It works!  Now let's try that using `CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu` instead of enforcing `--target`:
> ```
> # CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" cargo install ripgrep --force
> # file rg
> rg: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, with debug_info, not stripped
> ```
> Still working!  So the theory is that I have a working `rustc` along with a proper recognition of my target `gcc` as default linker and `CARGO*` and `RUSTUP*` environment variables are also recognized and functional.
> 
> 
> Now, following your Dockerfile example, along with using host default python3, still on `x86_64` host:
> ```
> # python3 --version
> Python 3.9.2
> # apt-get update
> # apt-get install python3-venv
> # VENV_PATH=/opt/venv python3 -m venv $VENV_PATH --upgrade
> # VENV_PATH=/opt/venv python3 -m pip install --upgrade setuptools pip
> # VENV_PATH=/opt/venv python3 -m pip install --upgrade setuptools-rust wheel
> # CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" PYO3_CROSS_LIB_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/lib/ PYO3_CROSS_INCLUDE_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/include/ VENV_PATH=/opt/venv python3 setup.py bdist_wheel --py-limited-api=cp36
> ```
> Boom! working!  But hold-on, why is the wheel extension being `x86_64`?
> ```
> # ll dist/cryptography-38.0.1-cp36-abi3-linux_x86_64.whl
> -rw-r--r-- 1 root root 1214001 Sep 30 19:46 dist/cryptography-38.0.1-cp36-abi3-linux_x86_64.whl
> ```
> After investigating a little further, looks like everything in the wheel file is ?OK?:
> ```
> # file ./hazmat/bindings/_rust.abi3.so
> ./hazmat/bindings/_rust.abi3.so: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
> ```
> 
> Now let's try out the exact same thing, this time using our host native built python3.10:
> ```
> # VENV_PATH=/opt/venv /spksrc/native/python310/work-native/install/usr/local/bin/python3 -m venv $VENV_PATH --upgrade
> # VENV_PATH=/opt/venv /spksrc/native/python310/work-native/install/usr/local/bin/python3 -m pip install --upgrade setuptools pip
> # VENV_PATH=/opt/venv /spksrc/native/python310/work-native/install/usr/local/bin/python3 -m pip install --upgrade setuptools-rust wheel
> 
> # CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" PYO3_CROSS_LIB_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/lib/ PYO3_CROSS_INCLUDE_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/include/ VENV_PATH=/opt/venv /spksrc/native/python310/work-native/install/usr/local/bin/python3 setup.py bdist_wheel --py-limited-api=cp36
> ```
> Again, boom!  Although it works with a few deprecation warnings:
> ```
> build/temp.linux-x86_64-cpython-310/cryptography.hazmat.bindings._openssl.c: In function ‘_cffi_d_ASN1_STRING_data’:
> build/temp.linux-x86_64-cpython-310/cryptography.hazmat.bindings._openssl.c:7957:3: warning: ‘ASN1_STRING_data’ is deprecated [-Wdeprecated-declarations
> ```
> Let's investigate further:
> ```
> # ll dist/cryptography-38.0.1-cp36-abi3-linux_x86_64.whl
> -rw-r--r-- 1 root root 2040082 Sep 30 20:10 dist/cryptography-38.0.1-cp36-abi3-linux_x86_64.whl
> ```
> After unziping:
> ```
> # find . -name "*.so"
> ./cryptography/hazmat/bindings/_openssl.abi3.so
> ./cryptography/hazmat/bindings/_rust.abi3.so
> # file ./cryptography/hazmat/bindings/_openssl.abi3.so ./cryptography/hazmat/bindings/_rust.abi3.so
> ./cryptography/hazmat/bindings/_openssl.abi3.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=a1db744141c71550a25dac57f63994c81be4cf90, with debug_info, not stripped
> ./cryptography/hazmat/bindings/_rust.abi3.so:    ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
> ```
> So clearly a mismatch here.  But it did built sort of "succesfully" using a `venv`.  But clearly this wheel output is useless as having arch mismatchs at every levels.  But it did demonstrated that it can build, which is already a start.
> 
> Now, challenge really is, building using a `crossenv` ...

I'm not building these with spksrc at all, because it isn't necessary. All pyo3 wheels built on a system with libc (non-alpine) will run on the same architecture they were built for, regardless of the host.

Buildx is a crossenv. The GitHub actions runs in an x86_64 host container always (ubuntu-latest is Ubuntu:jammy right now I believe). Buildx runs each build in a qemu cross ENV either determined by the --platform argument or possibly overridden on the FROM line of the dockerfile.

Venv is a totally different thing, that just isolates the python environment from the system python environment, so there are no conflicts with the system python packages or contamination of the build/install of what you are building. That could be left out entirely from the dockerfile and it would work, it just isn't best practices. 

So essentially, if this build worked in GitHub actions, very easily we could build every cross wheel in GitHub actions directly and then pass them into the spksrc build and build spksrc without needing rust at all.

It was just trying to prove that it isn't spksrc toolchain that is the problem, there is something wonky with GitHub actions that's making it not possible to build 32 bit builds. I'm going to break it back into cargo fetch on amd64 and then build on the --platform=$TARGETPLATFORM stage and see if that fixes it on GitHub actions.

If I get this built, I will create releases containing all of the built wheels, and you can reference them with --find-links and always avoid compile errors (and reduce storage usage and build time) on spksrc.

miigotu:
I'll add the bud log later, it absolutely is cross compiling the rust and c extensions.

mk/spksrc.cross-rust.mk
miigotu: I figured using cargo_home in a dir that has other files would conflict. There's a lib and cache dir and some other stuff in there too

Dockerfile
```diff
@@ -56,6 +57,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 		patchelf \
 		php \
 		pkg-config \
+		python2 \
```
miigotu, line 60: Are python3-venv and python3-pip not necessary here? I think python3-apt is also necessary for some packages to install their own depends. I'm not 100% familiar with the rest of the process in this docker.

th0ma7: No they're not needed.  How `spksrc` works is that it will build its own native python exactly matching the cross-compiled one that is built right afterwards.  So the host python always matches the target python.  Also, the other necessary tools such as `pip`, `crossenv`, etc. all get's intalled within either the native or cross versions of python, depending where it's needed.

th0ma7:
I gave it a shot on both Docker and a native VM to see if this isn't something in the host detection or similar... Sadly same result.

It's really easy to trigger with this PR, once the build is complete (e.g. `make arch-armv7-7.0`) do the following.  First setup your PATH to point towards the `crossenv` first:
```
$ export PATH=/spksrc/spk/python310/work-aarch64-7.0/crossenv/cross/bin:/spksrc/spk/python310/work-aarch64-7.0/crossenv/cross/bin:$PATH
```
Then build:
```
$ CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" \
CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu \
CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" \
PYO3_CROSS_LIB_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/lib/ \
PYO3_CROSS_INCLUDE_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/include/ \
python3 setup.py bdist_wheel --py-limited-api=cp36
```

Interestingly enough it gets a little farther using `RUSTFLAGS` such as:
```
RUSTFLAGS="--target=aarch64-unknown-linux-gnu -Clinker=/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" \
CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" \
PYO3_CROSS_LIB_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/lib/ \
PYO3_CROSS_INCLUDE_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/include/ \
python3 setup.py bdist_wheel --py-limited-api=cp36
```

A call using crossenv `pip` will also result with the same outcome:
```
CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-7.0/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc" \
CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu \
CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/rustup" \
PYO3_CROSS_LIB_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/lib/ \
PYO3_CROSS_INCLUDE_DIR=/spksrc/spk/python310/work-aarch64-7.0/install/var/packages/python310/target/include/ \
pip install cryptography==38.0.1
```

miigotu:
It's just crazy that I can build them with buildx locally but not on GitHub, with the same sources, commands, environment, and everything.


I'm going to try another method to compile right in the actions, rather than inside a dockerfile.

I can set up python in the action, set up rust toolchain (with multiple targets), and build/compile with setuptools-rust.

Now whether that gets around the bug in qemu (or GitHub, I'm not 100% sure the culprit yet), I don't know.

I suspect it could be because the action runner itself is in a container, and then we run another container inside that. Which I expect is why when I run the buildx command on my host machine it works, but on GitHub it doesn't. There's really no other difference between how I run here and there.

miigotu:
I imagine my next attempt to look something like this:

actions/setup-python (matrix with python 3.7-3.11-dev)
Then something similar to this:
https://github.com/actions-rs/example/blob/master/.github/workflows/cross_compile.yml
Then-
run: |
  pip download --no-deps :all: cryptography
  unzip
  cd and python setup.py bdist_wheel --py-limitef-abi=cp36 cryptography

miigotu:
@th0ma7 seems someone has already done what I was thinking, check this out: https://github.com/PyO3/setuptools-rust/pull/185/files

th0ma7:
> I imagine my next attempt to look something like this:
> 
> actions/setup-python (matrix with python 3.7-3.11-dev) Then something similar to this: https://github.com/actions-rs/example/blob/master/.github/workflows/cross_compile.yml Then- run: | pip download --no-deps :all: cryptography unzip cd and python setup.py bdist_wheel --py-limited-abi=cp36 cryptography

You can already do something close to this by using `cross/cryptography` which in turns doesn't use `pip` for building.  You simply have to add it to `DEPENDS` variable of `spk/python310` for a start.  Although the github workflow won't be the same.

messense:
Cross-post from https://github.com/PyO3/setuptools-rust/issues/294

This circles back to https://github.com/benfogle/crossenv/issues/65, after removing `LIBRARY_PATH` env var from `/spksrc/spk/python310/work-aarch64-7.0/crossenv/cross/bin/python` and adding `PYO3_CROSS_LIB_DIR=/spksrc/spk/python310/work-aarch64-7.0/crossenv/lib` env var for pyo3, the build succeeded for me locally.

th0ma7:
Good news!  I got it working https://github.com/PyO3/setuptools-rust/issues/294#issuecomment-1264353373

I have to think about the various code changes I need to do to have a working fix (or pending a new `setuptools-rust` release) but it does work both using `cross` and `pip` from a `crossenv`... there is light at the end of the tunnel finally!

miigotu:
Awesome, I hope that same thing fixes a straight build in gh actions too. Maybe this fix will bring back prebuilt wheels for arm32 and ppc64le

th0ma7:
@hgy59 got a working PR to generate rust-based wheels.  I still have a few details to polish but overall code is looking good.

The real thing in here it the updated docker image.  I propose pushing it into a different PR which in turn would allow me to fully test things in this one in particular (as I can't using github-action until it's merged to master, although I tested it manually).

Changes to mention related to the Docker image:
- Bumped to Debian 11 (e.g. stable)
- Installation by default of cargo/rustc environment

I did a lot of build testing and overall things is looking good.  Although I'm sure there will be things to adjust from migrating to Debian 11 resulting in a few bumps down the road that I'm confident we can address afterwards.

Thoughts?

miigotu:
Could you build the docker image and push it to your own dockerhub, and temporarily change the reference in this PR to get it to build? Then revert that change before squashing?

th0ma7:
No clue.  I could merge it on my own master and test from there then undo that last commit I guess.  But I believe the Docker part is ready for a merge.

miigotu:
From experience I suggest taking the extra step as a precaution. My workflows and dockerfile is nowhere near as complicated as they are on this repo, but I think they are a fairly decent process. And I think I have it all working well  many many times and turns out I'm wrong 😭

I guess here it won't break 20k+ installs instantly, but still haha.

th0ma7:
> > Yes, but it's also only build when it's on `master`, so it's not really tested.
> 
> Yea I suppose the idea here was that you do build the docker container locally first and verify it before pushing the changes to master. However you can create a docker build in your fork if that is easier. yea it's not a good idea to do docker builds in PRs as we don't want to modify the published image until the changes land in master.

@publicarray does it take a little while before the updated docker image gets used?

publicarray:
@th0ma7 I assume just the time it is required to build: https://github.com/SynoCommunity/spksrc/pkgs/container/spksrc

th0ma7:
@publicarray examining the build errors it does use the latest Docker image.  There's just a few details pending that I'm now fixing within the `python310` build itself.  Otherwise all is good, thnx.

cross/cryptography-legacy/Makefile
```diff
@@ -0,0 +1,14 @@
+PKG_NAME = cryptography
+PKG_VERS = 3.3.2
```
miigotu, line 2: 3.4.8 is the last version you can probably compile for legacy, with CRYPTOGRAPHY_DONT_BUILD_RUST.

idk if you want to update that package now so you don't have to touch this part again or if you want to leave it for another time.

miigotu:
Build passed! Great work!

cross/cryptography-legacy/Makefile
th0ma7: Thnx for the info but cryptography-legacy is really only meant for not breaking `python2` and `python38`, both of which are expected to be removed shortly.

miigotu: Only concern I would have, is some projects pin 3.4.8 (34.8) to make their packages continue to work on alternative hardware.

th0ma7: And everyone as the possibility to add it to their own `requirement-crossenv.txt` file as they see fit :)

Issue #5288, opened 2022-05-27
Creating Torrent file


MrOnyx:
I am using ruTorrent 3.10-10 on Synology DS1812+ 6.2.4-25556 Update 6 when creating torrent I got a message like this
i am not sure if torrent file is created properly

**/bin/php: /var/packages/rutorrent/target/lib/libz.so.1: no version information available (required by /bin/php)**
<img width="467" alt="Screenshot 2022-05-26 094145" src="https://user-images.githubusercontent.com/2565966/170685720-e14544de-9dfe-4d29-875d-f3948c5c2746.png">

PR #5455, opened 2022-10-16
rutorrent: DSM7 support + fix torrent creation wizard

smaarn:
## Description

Having rutorrent DSM7 ready + fix cryptic php error when trying to create torrent file via wizard

Fixes https://github.com/SynoCommunity/spksrc/issues/5288

## Checklist

- [x] New installation of package completed successfully (DSM 7)
- [x]  New installation of package completed successfully (DSM 6)
- [x] Package upgrade completed successfully (Manually install the package again) (DSM 7)
- [x] Package upgrade completed successfully (Manually install the package again) (DSM 6)

### Type of change

- [x] Package update
- [x] Bugfix


th0ma7:
@smaarn this is awesome, thnx for your assistance on this, really much appreciated.

A few things for considerations:
- the 2x services startup can now be handled within `service-setup.sh`.  `deluge` is a good example https://github.com/SynoCommunity/spksrc/blob/master/spk/deluge/src/service-setup.sh
- `rutorrent` 4.0-beta3 is now available, perhaps we can build around that and align with its release (or not)?

Let me know if/how I can provide assistance.

smaarn:
@th0ma7 will be having a look at the deluge folder conventions to see if I can match it easily with the defaults.

Regarding your "2x" services startup I'm not sure what you mean but will be having a look at how it's done in deluge.

The rutorrent 4.0 migration is definitely something we should be doing asap but I will be, for now, focusing on having the DSM 7 installation working and operational before diving into 4.0 migration (since their documentation on what was "changed" is pretty scarce I would tend to favor a conservative approach).

th0ma7:
> Regarding your "2x" services startup I'm not sure what you mean but will be having a look at how it's done in deluge.

I revisited this and you're right, there isn't "two" daemons but rather:
1. uses the default synology web server (where we could provide something else but is it worth it?)
2. calls a `screen` and invoke `rutorrent`.  That's done from a side script called `rtorrent-daemon`

So there really is only one daemon and it's call could be integrated into the `service-setup.sh` script instead of being left aside which would simplify maintenance later on.

BTW, if you want to play with the screen session, first identify it:
```
$ sudo su -s /bin/bash sc-rutorrent -c 'screen -ls'
Password: 
There is a screen on:
	22652..th0ma7-nas	(Detached)
1 Socket in /tmp/uscreens/S-sc-rutorrent.
```
Then attach to it:
```
$ sudo su -s /bin/bash sc-rutorrent -c 'screen -x'
```
To get help once connected use `CTRL-a + h` and to disconnect `CTRL-a + d`

smaarn:
Package tested manually under DSM 7: both install and upgrade works fine.

The open point is about DSM 6 not being broken (don't have a DSM 6 around so ... :/ ). Don't see why it would but then I changed a few things regarding the upgrade process which may have some unforeseen impacts.

Will be following up on this in the coming days.

th0ma7:
This is awesome work!  I still have a dsm6 so I'll test it over the coming days.  Again, great work and long overdue!

spk/rutorrent/src/rtorrent.rc
```diff
@@ -1,9 +1,9 @@
 scgi_port = localhost:8050
 directory = @download_dir@
-session = /usr/local/rutorrent/var/.session
+session = @var_dir@/.session
```
hgy59, line 3: my preferred way would be to use
`/var/packages/rutorrent/var`
this is availabel on all DSM versions, since we add DSM 7 compatible folders for DSM 5, 6 and SRM 1.2.

spk/rutorrent/src/rtorrent.rc
```diff
@@ -1,9 +1,9 @@
...
 schedule = watch_directory,5,5,load.start=@watch_dir@/*.torrent
 max_memory_usage = @max_memory@
-log.open_file = "rtorrent.log", "/usr/local/rutorrent/var/rtorrent.log"
+log.open_file = "rtorrent.log", "@var_dir@/rtorrent.log"
```
hgy59, line 6: dito

hgy59:
I propose to remove the configuration of the port range from the wizards.
As long as the firewall ports (in *.sc file or resources) are not adjusted to the ports from the wizard at installation time, it will not be possible to use these ports with activated firewall.

spk/rutorrent/src/rtorrent.rc
smaarn: I don't know why I thought it was differing depending on DSM versions.

Will be updating it (this will even simplify the code load !)

spk/rutorrent/src/wizard_templates/base_install_uifile.json
```diff
@@ -0,0 +1,100 @@
...
+                        "validator": {
+                            "allowBlank": true,
+                            "regex": {
+                                "expr": "/^\\\/volume[0-9]+\\\//",
+                                "errorText": "Path should begin with /volume?/ with ? the number of the volume"
+                            }
+                        }
+                    }
+                ]
+            },
```
hgy59, line 72: IMHO this share must be separated to volume and share name as above for the download location
OR you use the same volume as above (we did not test whether the creation of multiple shared folders on different volumes for the same package is supported by DSM).
OR you create the watch folder as a sub directory of the download shared folder, then you have to adjust at least the validation regex.

smaarn:
> I propose to remove the configuration of the port range from the wizards.
> As long as the firewall ports (in *.sc file or resources) are not adjusted to the ports from the wizard at installation time, it will not be possible to use these ports with activated firewall.

I would even add that altering the file configuration wouldn't make it work natively (meaning we would need to ensure that upon file configuration changes the firewall settings are updated... which could be a terrible thing to implement and do).

Doing that would require actually adding a rutorrent documentation section to be fair.

spk/rutorrent/src/wizard_templates/base_install_uifile.json
th0ma7: I would add, why not simply follow the exact same directory tree as deluge and enforce that by default such as:
```
downloads/
├── complete
├── incomplete
└── watch
```
There's a dsm6-7 "compatible" method to create the folders from deluge `service-setup.sh` script:
```
# Create download directories
install -m 0775 -o ${EFF_USER} -g ${GROUP} -d "${incomplete_folder}"
install -m 0775 -o ${EFF_USER} -g ${GROUP} -d "${complete_folder}"
install -m 0775 -o ${EFF_USER} -g ${GROUP} -d "${watch_folder}"
```

th0ma7:
@smaarn upgrading from DSM6 failed.  Overall the logs looks ok [rutorrent-DSM6-upgrade.txt](https://github.com/SynoCommunity/spksrc/files/9998006/rutorrent-DSM6-upgrade.txt) like it failed silently?

Capture d’écran du 2022-11-13 14-08-28

th0ma7:
Recap, success so far:
- clean install works great!
- can download/upload, overall it looks good
- kudos, kudos, kudos!

issues found so far:
- upgrade from earlier version on DSM6 fails
- max queue of 1 currently (e.g. only download 1x torrent at a time).  I believe to solve this we have to set `max_downloads_div` and `max_uploads_div` to `0`.
- Listening Port: ports are set by default to `6881-6999`.  I noticed under it the checked option: `Randomize port each time rTorrent starts`.  Hopefully the randomization is being done into the port range?  Also the port range is quite big, but that may be ok, duno.

smaarn:
It may actually be related to the former values used for the wizard which usef to be in the format /volume1/...

spk/rutorrent/src/wizard_templates/base_install_uifile.json
hgy59: We should use a different watch folder for each package, as we do not want to download a watched file by all watching applications (and it will depend on whether an application removes the files in the watch folder when download starts vs when the download is completed).

th0ma7: Good point, maybe worth defaulting it to `watch-<appname>` then?  I could make the changes to deluge while i have an opened pr.

smaarn:
> issues found so far:

> upgrade from earlier version on DSM6 fails
> missing geoIP plugin? It's actually there but it is missing the flag in the Peers torrent sub-tab.
... Need to check.
> max queue of 1 currently (e.g. only download 1x torrent at a time). I believe to solve this we have to set max_downloads_div and max_uploads_div to 0.
Can change the default but I believe this was already the case before. 
> Listening Port: ports are set by default to 6881-6999. I noticed under it the checked option: Randomize port each time rTorrent starts. Hopefully the randomization is being done into the port range? Also the port range is quite big, but that may be ok, duno.
It's indeed randomized within the specified range. Not too sure about the range size TBH.

TruuuC:
Hi,

The package seem works in my side, no issue.
Just had to install PHP 7.3 in addition of 7.4 already installed.

I also tried .htpasswd with success.

Thanks

smaarn:
Ok so some news:

1. DSM 6 upgrade actually highlighted an unexpected issue: the reload_inst_variables doesn't seem to work as expected => fix proposal included in this MR
2. I reworked the wizard mechanism to use shell generating wizards (to avoid all those JSON hanging around and to remain as much as possible compliant with standard DSM mechanisms) which allows for handling older versions' upgrade

With regards to the proposed folder structures I actually wonder if it's really worth having a watch directory **enforced** by default and worrying about concurrent application usages.
I would propose instead to provide with a checkbox "use watch folder" and have the default being the proposed convention (e.g. ./watch under the download directory).

mk/spksrc.service.installer.functions
```diff
@@ -98,7 +98,7 @@ reload_inst_variables ()
         for _line in $(cat "${INST_VARIABLES}"); do
             _key="$(echo ${_line} | awk -F'=' '{print $1}')"
             _value="$(echo ${_line} | awk -F'=' '{print $2}')"
-            declare "${_key}=${_value}"
```
smaarn, line 101: It seems that `declare ` on its own and used within a function wouldn't be overriding the parent context, making this call actually not updating the parent context. Unless I missed something ?

spk/rutorrent/Makefile
```diff
@@ -1,15 +1,31 @@
 SPK_NAME = rutorrent
 SPK_VERS = 3.10
-SPK_REV = 12
+SPK_REV = 13
 SPK_ICON = src/rutorrent.png
 DSM_UI_DIR = app
 
-SPK_DEPENDS = "python310:PHP5.6"
+SPK_DEPENDS = "python310:PHP7.3"
```
smaarn, line 7: Could totally migrate to PHP7.4 TBH. PHP7.3 only presents the advantage of being the same version as the default one shipped at the level of DSM 7 so it's homogeneous.

spk/rutorrent/Makefile
```diff
@@ -27,35 +53,64 @@ SERVICE_USER = auto
...
 ENV += BUSYBOX_CONFIG="$(BUSYBOX_CONFIG)"
 
 # some dependents require C++11 support
 UNSUPPORTED_ARCHS = $(ARMv5_ARCHS) $(OLD_PPC_ARCHS)
 # unsupported with cross/rtorrent
 UNSUPPORTED_ARCHS += $(ARMv7L_ARCHS)
 
+SYSTEM_GROUP = http
+
+ICON_DIR = $(STAGING_DIR)/$(DSM_UI_DIR)/images
```
smaarn, line 77: This is a hack to force the generation of icons. A rework of the framework could be useful to force them (otherwise DSM 7 specific resource workers wouldn't be using available icons)

spk/rutorrent/src/wizard_templates/base_install_uifile.json
smaarn: Actually I would propose to have a combination of

1. A checkbox allowing for usage of a watch directory (disabled by default)
2. An input component defaulted to the "watch" subdirectory

WDYT ?

I am actually not sure that rutorrent supports the complete / incomplete logics.

th0ma7: Sounds good to me.

As for the complete vs incomplete directories, if I recall you have to set the default directory to incomplete and use the move upon completion feature so it, well, move the resulting torrent to the complete folder once over.

One thing i noticed is the incredible performance difference between deluge vs rutorrent.  Deluge is numerous time more performant ... There may be config options to fine tune rutorrent...

Traveling this week so my availability may vary.

smaarn:
(flagging as draft since I did a few rework on the installation / upgrade procedure, still need to retest everything)

smaarn:
@hgy59 : removed the port customization feature
@th0ma7 : I have made several changes which should enable the upgrade from DSM 6

There are two open points:

1. The folders' separation logics
2. 4.0 upgrade

It turns out that 

1. Implementing the idea with the checkbox for the watch folder logic was actually... not feasible (the javascript framework doesn't allow me to do this it seems)
2. The 4.0 upgrade doesn't work as is (separated MR for that)

I will try to find some time to work on the folders' separation logic and 4.0 upgrade within the coming weeks.

As is, though (once the DSM 6 update scenario is validated), I believe it should be worth merging it. WDYT ?

th0ma7:
At first glance things look great.  But there seems to be a few corner-case to be dealth with.  Although I agree, 4.0 can wait for a subsequent release.

I tested the upgrade on DSM6.  Note that I uninstalled completely then reinstalled from our repo to have a clean installation.  Surprisingly I ended-up having the following error in the interface (although no error in the instlallation log):
```
[18.12.2022 16:31:02] _cloudflare: Plugin will not work. Webserver user can't access external program (python).
```
And trying to download archlinux didn't worked out... (stagnated at 0%) just weird.

Anyhow, afterward I've applied your update over it and sadly it failed.  What is weird is that there isn't any errors in the installation log file.

Capture d’écran du 2022-12-18 16-40-03

Looking a bit further I noticed the following in the `/var/log/messages`
```
2022-12-18T16:28:52-05:00 th0ma7-nas synoscgi_SYNO.Core.Package_2_list[16388]: pollingtask.cpp:506 PollingTask: unable to stop daemon when reloading
2022-12-18T16:28:52-05:00 th0ma7-nas synoscgi_SYNO.Core.Package_2_list[16388]: list.cpp:329 SYNO.Core.Package.list rollback to direct call
2022-12-18T16:30:27-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_install[24745]: resource_api.cpp:190 Acquire service-cfg for rutorrent when 0x0001 (done)
2022-12-18T16:30:37-05:00 th0ma7-nas synoacltool: synoacl_delete.c:43 failed to get acl of '/volume1/@appstore/rutorrent/tmp', errno=[8100]
2022-12-18T16:30:38-05:00 th0ma7-nas synoacltool: synoacl_delete.c:43 failed to get acl of '/var/services/web/rutorrent/share', errno=[8100]
2022-12-18T16:30:43-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Control_1_start[27182]: resource_api.cpp:190 Acquire php56 for rutorrent when 0x0001 (done)
2022-12-18T16:37:26-05:00 th0ma7-nas synoscgi_SYNO.Core.Package_2_list[9412]: list.cpp:329 SYNO.Core.Package.list rollback to direct call
2022-12-18T16:37:32-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_upload[10658]: process.cpp:236 Failed to run /volume1/@tmp/jiZ6Ci/WIZARD_UIFILES/upgrade_uifile_fre.sh, ret=[-1], Permission denied
2022-12-18T16:37:32-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_upload[10594]: pkgtool.cpp:1835 Failed to run /volume1/@tmp/jiZ6Ci/WIZARD_UIFILES/upgrade_uifile_fre.sh for rutorrent
2022-12-18T16:37:32-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_upload[10660]: process.cpp:236 Failed to run /volume1/@tmp/jiZ6Ci/WIZARD_UIFILES/upgrade_uifile.sh, ret=[-1], Permission denied
2022-12-18T16:37:32-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_upload[10594]: pkgtool.cpp:1835 Failed to run /volume1/@tmp/jiZ6Ci/WIZARD_UIFILES/upgrade_uifile.sh for rutorrent
2022-12-18T16:37:33-05:00 th0ma7-nas php: Signature file not found in archive.
2022-12-18T16:37:43-05:00 th0ma7-nas php: Signature file not found in archive.
2022-12-18T16:37:55-05:00 th0ma7-nas php: SYSTEM:	Last message 'Signature file not f' repeated 1 times, suppressed by syslog-ng on th0ma7-nas
2022-12-18T16:37:55-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_install[14832]: resource_api.cpp:289 Release php56 for rutorrent when 0x0002 (done)
2022-12-18T16:38:01-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_install[15473]: resource_api.cpp:289 Release service-cfg for rutorrent when 0x0002 (done)
2022-12-18T16:38:03-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_install[15741]: resource_api.cpp:190 Acquire data-share for rutorrent when 0x0002 (done)
2022-12-18T16:38:06-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_install[15872]: resource_api.cpp:190 Acquire service-cfg for rutorrent when 0x0002 (done)
2022-12-18T16:38:06-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_install[15872]: resource_api.cpp:190 Acquire data-share for rutorrent when 0x0002 (fail)
2022-12-18T16:38:06-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_install[11234]: pkginstall.cpp:1221 Failed to acquire resource after install rutorrent [0xD900 manager.cpp:204]
2022-12-18T16:38:11-05:00 th0ma7-nas synoscgi_SYNO.Core.Package_2_list[16312]: list.cpp:329 SYNO.Core.Package.list rollback to direct call
```

So my next attempt was at trying a full removal of existing and installing fresh with your new package.  From the get go there is an issue with the download folder...   Mine is located elsewhere so presuming all are using that as default won't work.  Better off asking the location as we used to.  I would suggest re-using the deluge method personnally.
```
2022/12/18 16:45:58	Adding 'sc-rutorrent' to 'sc-download'
2022/12/18 16:45:58	Group Name: [sc-download]
2022/12/18 16:45:58	Group Type: [AUTH_LOCAL]
2022/12/18 16:45:58	Group ID:   [65538]
2022/12/18 16:45:58	Group Members: 
2022/12/18 16:45:58	0:[sc-deluge]
2022/12/18 16:45:58	1:[sc-rutorrent]
2022/12/18 16:45:58	Begin service_postinst
2022/12/18 16:46:00	Granting 'sc-download' group permissions on /volume1/downloads
2022/12/18 16:46:00	(synoacltool.c, 140)Path not found 
2022/12/18 16:46:00	find: `/volume1/downloads': No such file or directory
2022/12/18 16:46:00	Granting 'sc-download' group basic permissions on /volume1/downloads
2022/12/18 16:46:00	(synoacltool.c, 140)Path not found 
2022/12/18 16:46:00	Python 3.10.8
```
Also later on in the logs the `${folder}` isn't interpreted from what I can see:
```
2022/12/18 16:47:01	find ${folder} -mindepth 1 -type d -exec synoacltool -enforce-inherit {} \;
2022/12/18 16:47:01	End service_postinst
2022/12/18 16:47:01	Granting 'sc-rutorrent' unix ownership on /volume1/@appstore/rutorrent/var
```

Hopefully I should have a few more cycles over the Holidays to play with this a help you out figuring out where the issue is.

spk/rutorrent/src/wizard_templates/base_install_uifile.json
smaarn: Sorry for the delays here.

First things first: my idea of having a checkbox is actually not feasible it seems with regards to the current Javascript framework which is used to handle the installer logics (short version: can't listen for a checkbox change to actually change visibility of a component, I can only do that on page changes)

I'm not (at all) an expert in rtorrent configuration, I may be having a look at this but it may take some time.

And for the performance part I must confess I don't use rutorrent (apart for testing, I'm more of a transmission user) and only had a crack at it to unblock some python logics so I can't really argue with you here :)

Will try having a look today.

smaarn:
> I tested the upgrade on DSM6. Note that I uninstalled completely then reinstalled from our repo to have a clean installation. 
There is an issue with the uninstallation: it doesn't purge directories (so you'd need to manually delete them :/)

Last time I saw this error with the _cloudflare plugin not being able to find python it was because of an issue with the installer where the php binary would refuse to run following the switch from a separate daemon bin to using the command variable (PATH tricks basically). I will try reverting that part.

> 2022-12-18T16:38:06-05:00 th0ma7-nas synoscgi_SYNO.Core.Package.Installation_1_install[15872]: resource_api.cpp:190 Acquire data-share for rutorrent when 0x0002 (fail)

This is bad news as it means that my "trick" for the upgrade didn't manage to actually work with the `data-share` worker. Do you think you could grab the contents of `conf/resource.own` file after your upgrade ?

> Also later on in the logs the ${folder} isn't interpreted from what I can see:

Actually this was more because I was echoing with resolving, but I can guarantee it was the right command. Will be fixing it though.

What I would propose is to:

1. Check how to revert the change about the script (it's causing the installer not to behave properly when configuring binary paths - python among others)
2. Check what's the behaviour in DSM 6 for the following installation cases:
  * Upgrade from an existing official version, checking the contents of the `conf/resource.own` (it will contain the fully resolved installation outcome)
  * Installing from scratch (after having ensure that the former directories were purged)

smaarn:
So the status of this MR is as follows:

1. Reverted the changes on the boot script (could be reworked but I'd first like to have DSM 7 support appropriately validated with DSM 6 still being supported)
2. In DSM 7 when upgrading I have the following logs of errors:

```
2023/02/04 17:53:56	upgrade rutorrent 3.10-13 Begin postinst
2023/02/04 17:53:56	Begin load_variables_from_file
/var/packages/rutorrent/scripts/installer: line 42: /var/packages/rutorrent/etc/installer-variables: Permission denied
2023/02/04 17:53:56	End load_variables_from_file
2023/02/04 17:53:56	Begin initialize_variables
2023/02/04 17:53:56	End initialize_variables
```

I have no clue what this is caused by, knowing that I have:

```
root@mallinas2:~# ls -la  /var/packages/rutorrent/etc/
total 12
drwxr-xr-x  2 sc-rutorrent synocommunity 4096 Feb  4 18:02 .
drwxr-xr-x 42 root         root          4096 Jan 21 12:25 ..
-rw-r--r--  1 sc-rutorrent synocommunity   48 Feb  4 18:02 installer-variables
```

smaarn:
Added help label as DSM 6 testing would be really appreciated (switching from an "old school" folder management to using data shared mechanism so... need to be sure it's still sound). I unfortunately don't have a DSM 6 version installed (and don't have the possibility to install one on my NAS)

mreid-tt:
> In DSM 7 when upgrading I have the following logs of errors:
> `[---snip---]`
> I have no clue what this is caused by, knowing that I have:

I've seen this recently as well and my guess it could be a change to the packager. See my experience in this https://github.com/SynoCommunity/spksrc/pull/5592#issuecomment-1416724363.

> Added help label as DSM 6 testing would be really appreciated (switching from an "old school" folder management to using data shared mechanism so... need to be sure it's still sound). I unfortunately don't have a DSM 6 version installed (and don't have the possibility to install one on my NAS)

I have an instance of DSM6 in a VirtualDSM container on an x64-based NAS. If this platform can work, let me know what specifically you would like assistance testing and I'll see what I can do.

spk/rutorrent/Makefile
th0ma7: I'm not sure I follow, what's missing exactly?

spk/rutorrent/src/bin/rtorrent-daemon
```diff
@@ -1,22 +1,34 @@
...
+  echo "This script requires the RUTORRENT_WEB_DIR environment variable to be set"
+  exit 1
+fi
+
+if [ "${SYNOPKG_PKGVAR}" == "" ]; then
+  echo "This script requires the SYNOPKG_PKGVAR environment variable to be set"
+  exit 1
+fi
+
+if [ "${SYNOPKG_PKGDEST}" == "" ]; then
```
th0ma7, line 15: Guessing `if [ -z "${string}" ]` could do as well.

spk/rutorrent/src/service-setup.sh
```diff
@@ -27,14 +32,12 @@ PID_FILE="${SYNOPKG_PKGVAR}/rtorrent.pid"
...
+
 validate_preinst ()
 {
     if [ "${SYNOPKG_PKG_STATUS}" == "INSTALL" ]; then
-        if [ ! -d "${wizard_download_dir}" ]; then
-            echo "Download directory ${wizard_download_dir} does not exist."
-            exit 1
-        fi
-        if [ -n "${wizard_watch_dir}" -a ! -d "${wizard_watch_dir}" ]; then
+        if [ -n "${wizard_watch_dir}" ] && [ ! -d "${wizard_watch_dir}" ]; then
```
th0ma7, line 40: FYI, I've migrated deluge's watch directory to be `watch-deluge` so rutorrent and deluge do not collide.

spk/rutorrent/src/service-setup.sh
```diff
@@ -61,59 +64,65 @@ fix_shared_folders_rights()
     local folder=$1
     echo "Fixing shared folder rights for ${folder}"
```
th0ma7, line 66: I'm assuming this only runs on DSM6?  For my own curiosity, can DSM7 installation method (e.g. with no direct intervention of file user/groups) be used as well for DSM6 (e.g. could it be possible to get rid of it entirely)?

spk/rutorrent/src/bin/rtorrent-daemon
```diff
@@ -1,22 +1,34 @@
 #!/bin/bash
```
th0ma7, line 2: As `service-setup.sh` now allows passing more than one daemon start|stop|status, could the `rtorrent-daemon` be merged within service-setup?

th0ma7 (approved):
Wow, just wow!  I will test it out later today but overall you did an amazing job on this!

I do have some optional changes and mainly questions.  The main items I'm wondering are:
1. Is it possible for DSM6 install to become closer to DSM7 installation to ease maintenance?  But you may well be using DSM7 specific new feature...
2. Can/should the `rtorrent-daemon` be somewhat merged within `service-setup.sh` (may not apply, for reference see deluge service script).

th0ma7:
Pertaining to the version 4.x upgrade, it now is officially out:
https://github.com/Novik/ruTorrent/releases/tag/v4.0.1-hotfix

To be included or not is a different question.  Having rutorrent ported to DSM7 really was the first objective.

spk/rutorrent/src/service-setup.sh
hgy59: we should keep the sh/ash compatible variant `if [ -n "${wizard_watch_dir}" -a ! -d "${wizard_watch_dir}" ]; then` and not use the bash syntax (&&) here (correct me if && is supported in ash).

spk/rutorrent/Makefile
smaarn: What's missing is that without this trick the icons are not automatically generated. Will rehave a look because framework has evolved lately but it seems icons were only generated in some very specific context.

spk/rutorrent/src/bin/rtorrent-daemon
smaarn: Actually, due to export constraints (playing with `PATH` and `LD_LIBRARY_PATH` it wasn't possible due to the requirement to use the native `php` binary which is available in the standard DSM PATH.

spk/rutorrent/Makefile
hgy59: looking at spksrc.icon.mk, icons are created when DSM_UI_DIR and SPK_ICON and SERVICE_PORT are defined and NO_SERVICE_SHORTCUT is not defined.

I suppose this will be fixed by adding to Makefile:
`SERVICE_PORT = 8050`

smaarn: Won't this be causing issues with the interactions with the way you can "open" the page ? (Checking)

hgy59: And probably you can use the SERVICE_PORT variable when patching the scgi_port in config.php at installation time.

hgy59: > Won't this be causing issues with the interactions with the way you can "open" the page ? (Checking)

I suppose it does not. It could change the generation of the app/config file, but you have a custom app/config file that is in src/app/config and is installed in rutorrent_extra_install. So you have to check that this file is still the same when adding SERVICE_PORT definition.

spk/rutorrent/Makefile
smaarn: Migrated to 7.4.

mreid-tt:
@smaarn, I took another look at this on DSM6.2 and the following line reoccurred from the previous log I sent up:
```
2023/02/04 14:21:40	find: `/volume1/downloads': No such file or directory
```

This struck me as strange that the downloads directory was not being created. Then I got to thinking that the installer is not prompting to create it. In fact, I didn't remember seeing any wizard prompts. Then I remembered an issue I came across when building Sonarr v4 for DSM6 (https://github.com/SynoCommunity/spksrc/pull/5524#issuecomment-1374689266), specifically, using shell scripts for the wizard files.

Your build seems to have the same issue as shown below:
```
% ls -all
total 80
drwxr-xr-x@ 6 mreid  staff   192 Feb 12 14:44 .
drwx------@ 9 mreid  staff   288 Feb 15 17:47 ..
-rw-r--r--@ 1 mreid  staff  9560 Feb 12 14:44 install_uifile.sh
-rw-r--r--@ 1 mreid  staff  9890 Feb 12 14:44 install_uifile_fre.sh
-rw-r--r--@ 1 mreid  staff  6532 Feb 12 14:44 upgrade_uifile.sh
-rw-r--r--@ 1 mreid  staff  6862 Feb 12 14:44 upgrade_uifile_fre.sh
```

Performing a `chmod g+x *.sh` on the files in the folder and then re-taring the archive I was able to successfully complete an install with DSM6. The successful log is shown here: https://pastebin.com/Ty3J74Hs.

Looking at some recently published packages such as [Jellyfin](https://github.com/SynoCommunity/spksrc/pull/5590) and [SABnzbd](https://github.com/SynoCommunity/spksrc/pull/5594) and examining their package contents I am now convinced that the permission setting for the wizard files is not working. Will need to look into this more and create a PR once the issue is diagnosed.

With your package however there are still some issues. When I have it installed on my VirtualDSM which has the admin address http://dsm62.local:5000/, I see that the URL for rutorrent is shown as http://dsm62.local/rutorrent/. The problem is that when I click the link it redirects to the DSM homepage and just gets me back to my admin address. I don't know if the webserver listening port is configured correctly to serve web clients.

th0ma7:
> Looking at some recently published packages such as [Jellyfin](https://github.com/SynoCommunity/spksrc/pull/5590) and [SABnzbd](https://github.com/SynoCommunity/spksrc/pull/5594) and examining their package contents I am now convinced that the permission setting for the wizard files is not working. Will need to look into this more and create a PR once the issue is diagnosed.

That could be easily fixed within the framework to enforce a chmod +x on the files prior to generating the .tar archive.

mreid-tt:
> That could be easily fixed within the framework to enforce a chmod +x on the files prior to generating the .tar archive.

The framework already has code for this:
https://github.com/SynoCommunity/spksrc/blob/5efc80869342049d9102672bb820639ea0853363/mk/spksrc.spk.mk#L394-L397

Also based on the build log for your [Build (x64-6.1)](https://github.com/SynoCommunity/spksrc/actions/runs/4157630720/jobs/7192290955#logs), it seems to be going into that section successfully:

```
  ===>  Create DSM Wizards
  /github/workspace/spk/rutorrent/work-x64-6.1/generated-wizards/upgrade_uifile.sh
  /github/workspace/spk/rutorrent/work-x64-6.1/generated-wizards/install_uifile.sh
  /github/workspace/spk/rutorrent/work-x64-6.1/generated-wizards/upgrade_uifile_fre.sh
  /github/workspace/spk/rutorrent/work-x64-6.1/generated-wizards/install_uifile_fre.sh
  ===>  Preparing conf
```

So either it is being set correctly and then overwritten or something else in the framework has changed. The last change to this section of code was in https://github.com/SynoCommunity/spksrc/pull/5383 but I can't see anything in there which would cause it to break.

th0ma7:
Actually I think I may have a clue of where the problem lies:
```
ifneq ($(wildcard $(DSM_WIZARDS_DIR)/*),)
```
Depending when it is being processed `wildcard` can have a different behavior depending of the value being evaluated at that specific time.  Also, in this case it doesn't provide any added value as both `find` commands won't induce any errors anyway if no files are found.

So a "potential" fix is to get rid of the `ifneq` call entirely and lead the find live a long free life :)

mreid-tt:
Hmm, the thing is though, when I checked another package before (redis) as in https://github.com/SynoCommunity/spksrc/pull/5524#issuecomment-1374808403, the build in the checks that are done here is different than the one that is published for download. So to me the build for publishing is working whereas the build for testing is not. I couldn't understand why the `SPK` files generated had different permissions but something is causing the permissions of the wizard files in the signed packages to be different that those of the unsigned ones.

mreid-tt:
> With your package however there are still some issues. When I have it installed on my VirtualDSM which has the admin address http://dsm62.local:5000/, I see that the URL for rutorrent is shown as http://dsm62.local/rutorrent/. The problem is that when I click the link it redirects to the DSM homepage and just gets me back to my admin address. I don't know if the webserver listening port is configured correctly to serve web clients (or is being overidden by the default webserver for DSM).

@smaarn, well I finally got your package to work in DSM 6.2. See screenshot below...

Screenshot 2023-02-16 at 5 43 47 AM

When I looked more into your Makefile, we have these lines:

https://github.com/SynoCommunity/spksrc/blob/46d1edbe3380f2f7b84fe4d73d04045ce1a7d898/spk/rutorrent/Makefile#L47-L49

However Apache was not being installed. To get your package to work, I had to take the following additional steps:

1. Install Web Station
2. Install Apache HTTP Server 2.4
3. Within Web Station > General Settings, change HTTP back-end server from the default Nginx to Apache HTTP Server 2.4

I don't know if this was an expectation for users to know this but perhaps either the wizard could be updated to inform them or maybe something in the `Makefile` or `conf/resource` needs to be updated so that these changes can be automatic.

Alternately, if there can be a configuration to make the web service compatible with Nginx that would be great!

mreid-tt:
I was also curious about these lines in the `conf/resource` file:

https://github.com/SynoCommunity/spksrc/blob/46d1edbe3380f2f7b84fe4d73d04045ce1a7d898/spk/rutorrent/src/conf/resource#L2-L7

In the developer guide the example just has "php" for the PHP INI as in:

> ### Syntax
> ```
> "php": {
> 	"php-ini": [{
> 		"relpath": "<ini-relpath>",
> 	}, ...],
> 	"fpm-conf": [{
> 		"relpath": "<conf-relpath>",
> 	}, ...]
> }
> ```

I wonder if this is a problem since in your `Makefile` you have:

https://github.com/SynoCommunity/spksrc/blob/46d1edbe3380f2f7b84fe4d73d04045ce1a7d898/spk/rutorrent/Makefile#L7

spk/rutorrent/src/conf/resource
```diff
@@ -4,5 +4,55 @@
 			"relpath": "etc/php/conf.d/com.synocommunity.rutorrent.ini",
 			"destpath": "/usr/local/etc/php56/conf.d"
 		}]
-    }
+    },
```
mreid-tt, line 7: Curious about these lines... in the developer guide the example just has "php" for the PHP INI as in:

> ### Syntax
> ```
> "php": {
> 	"php-ini": [{
> 		"relpath": "<ini-relpath>",
> 	}, ...],
> 	"fpm-conf": [{
> 		"relpath": "<conf-relpath>",
> 	}, ...]
> }
> ```

I wonder if this is a potential issue since in your `Makefile` you have:

https://github.com/SynoCommunity/spksrc/blob/46d1edbe3380f2f7b84fe4d73d04045ce1a7d898/spk/rutorrent/Makefile#L7

spk/rutorrent/src/service-setup.sh
smaarn: It seems that it's also running for DSM 7 as well and may still be required. Would need to check.

mreid-tt:
> Actually I think I may have a clue of where the problem lies:
> 
> ```
> ifneq ($(wildcard $(DSM_WIZARDS_DIR)/*),)
> ```
> 
> Depending when it is being processed `wildcard` can have a different behavior depending of the value being evaluated at that specific time. Also, in this case it doesn't provide any added value as both `find` commands won't induce any errors anyway if no files are found.
> 
> So a "potential" fix is to get rid of the `ifneq` call entirely and lead the find live a long free life :)

Based on some further research you seem to be spot on with this evaluation. I've created a patch for the framework under #5608 which should address this while keeping the original logic.

spk/rutorrent/Makefile
```diff
@@ -1,15 +1,24 @@
 SPK_NAME = rutorrent
 SPK_VERS = 3.10
-SPK_REV = 12
+SPK_REV = 13
 SPK_ICON = src/rutorrent.png
 DSM_UI_DIR = app
 
-SPK_DEPENDS = "python310:PHP5.6"
+SPK_DEPENDS = "WebStation;python310:PHP7.4"
```
mreid-tt, line 7: This syntax may not be correct. Also, having Webstation present may not be all that is needed to change the HTTP back-end server from the default Nginx to Apache HTTP Server 2.4. Any thoughts on getting PHP to work in either Nginx or Apache HTTP Server so that the installation will use whichever the user has already?

smaarn: > This syntax may not be correct.

Indeed, fixed it.

> Also, having WebStation present may not be all that is needed to change the HTTP back-end server from the default Nginx to Apache HTTP Server 2.4. 

No indeed this won't do that. Are we really looking to do this kind of things ?

> Any thoughts on getting PHP to work in either Nginx or Apache HTTP Server so that the installation will use whichever the user has already?

Well there are lots mechanism which could be used but is it really something we want to invest it ? We already need to support lots of different DSMs this may prove to be more complicated than needed (personally I don't care about nginx or apache). To be totally transparent I have limited availabilities so just bear with me if that's the target ;)

Would need to check why apache is needed though.

mreid-tt: Ideally users should be able to install a package and it work out of the box. Maybe changing the selected server is a config file to set without even needing to install WebStation? If it's onerous to change the selected HTTP back-end server then I would suggest that this requirement be explicit in the wizard UI files so that users will know when installing.

As to why Apache is needed, I couldn't say... curious as to your findings on this.

smaarn: > Ideally users should be able to install a package and it work out of the box. Maybe changing the selected server  is a config file to set without even needing to install WebStation? If it's onerous to change the selected HTTP back-end server then I would suggest that this requirement be explicit in the wizard UI files so that users will know when installing.

I see your point. Not too sure that this is part of this MR's scope though.

> As to why Apache is needed, I couldn't say... curious as to your findings on this.

Would you happen to have logs on the corresponding failures ? Because it may all simply be related to the configuration files not being appropriately injected.

mreid-tt:
@smaarn, so I tried your latest package with the updated dependencies and it works out of the box with DSM6 after all. I didn't need to install Apache HTTP Server and it seems to just work once Web Station is installed. See below video of the working install:


https://user-images.githubusercontent.com/943378/219782584-19845c09-b2b8-4437-85d2-b48afe7b42f9.mp4

spk/rutorrent/Makefile
mreid-tt: so this issue is no longer a problem based on my last comment below...

mreid-tt:
@smaarn, I'm not familiar with rutorrent but I was wondering if you had any thoughts on the errors seen in the log window:

```
Bad response from server: (404 [error,getplugins]) Not Found
```

Also, If I click on the Settings icon in the toolbar I get the following:

```
JS error: [http://dsm62.local/rutorrent/js/webui.js : 765] TypeError: this.systemInfo is undefined
```

These appear whether I use Nginx or Apache HTTP Server 2.4. Do you get these in your installation?

smaarn:
> @smaarn, I'm not familiar with rutorrent but I was wondering if you had any thoughts on the errors seen in the log window:
> 
> 
> 
> ```
> 
> Bad response from server: (404 [error,getplugins]) Not Found
> 
> ```
> 
> 
> 
> Also, If I click on the Settings icon in the toolbar I get the following:
> 
> 
> 
> ```
> 
> JS error: [http://dsm62.local/rutorrent/js/webui.js : 765] TypeError: this.systemInfo is undefined
> 
> ```
> 
> 
> 
> These appear whether I use Nginx or Apache HTTP Server 2.4. Do you get these in your installation?

I am not getting those. Need to recheck how the rights are set on the plugins directory. Do you think you could check how they are currently set ?

mreid-tt:
> I am not getting those. Need to recheck how the rights are set on the plugins directory. Do you think you could check how they are currently set ?

Sure, let me know what needs to be done. You can also hit me up on Discord for something a bit more real-time.

mreid-tt:
@smaarn, new build seems to resolve all remaining issues in DSM6. no errors in the interface when launched...

Screenshot 2023-02-19 at 11 43 35 AM

Installation log also looks clean -- https://pastebin.com/Ev0j4zUi.

mreid-tt:
I also did a test download [Test Torrent](https://academictorrents.com/details/d984f67af9917b214cd8b6048ab5624c7df6a07a) and confirmed that the application does download the torrent successfully...

Screenshot 2023-02-19 at 11 50 22 AM

Resulting in these in the downloads folder...

Screenshot 2023-02-19 at 11 51 12 AM

mreid-tt:
@smaarn, an upgrade from the current v3.10-10 on DSM is reporting as failed. The installation log seems to be fine though... https://pastebin.com/cJSQGY27.

mreid-tt:
@smaarn, your latest build was a success for upgrades from the current v3.10-10 on DSM6! See video...

https://user-images.githubusercontent.com/943378/219968412-db89aeaf-1431-4b0d-8f82-bf34022cc3d4.mp4

smaarn:
Thanks a lot for the help in testing !

spk/rutorrent/src/wizard_templates/base_install_uifile.json
mreid-tt: @smaarn, is this concern regarding the UI wizard logic now resolved?

spk/rutorrent/Makefile
mreid-tt: @smaarn, is this concern regarding the package icon generation now resolved?

spk/rutorrent/src/wizard_templates/base_install_uifile.json
smaarn: @mreid-tt no it's not. The wizard option is actually a dead-end and the actual rutorrent options need being tested.

I must confess I didn't consider investing more time in "normalizing" those options. TBH as a Transmission user I found it puzzling on my last upgrade to see the paths change so not sure it's something we want to force, is it ?

I mean, for an initial installation, it's arguable, but I'd say that for an existing installation it can be seen as a little bit "invasive" for a policy, is it not ?

spk/rutorrent/Makefile
smaarn: @mreid-tt issue should now be fixed. Thanks for following up on this and sorry for not answering earlier.

@hgy59  Found a much more more elegant way by the way to keep my app/config customization (by using an already existing framework option

smaarn:
4.0.2 upgrade comming in a follow up MR (here: https://github.com/SynoCommunity/spksrc/pull/5617 ) but with my limited availabilities didn't want to put it in the critical path. Here we have now a DSM 7 ready version to publish

th0ma7:
@smaarn was this published?  I don't see it online.  Let me know if you need assistance.

smaarn:
> @smaarn was this published?  I don't see it online.  Let me know if you need assistance.

@th0ma7 I don't believe this was published indeed.
TBH I don't have any idea on how to publish packages. Do you have any pointers on the procedure to follow ? (will be rechecking on my side if I find anything)

spk/rutorrent/src/wizard_templates/base_install_uifile.json
mreid-tt: > First things first: my idea of having a checkbox is actually not feasible it seems with regards to the current Javascript framework which is used to handle the installer logics (short version: can't listen for a checkbox change to actually change visibility of a component, I can only do that on page changes)

@smaarn, as a side note, I was working on a more robust uninstall UI wizard for ownCloud ([uninstall_uifile.sh](https://github.com/SynoCommunity/spksrc/blob/85c9a6c8609332caaea9295f26b1582ada6c739b/spk/owncloud/src/wizard/uninstall_uifile.sh)) and after going down a trial-and-error rabbit-hole, I can confirm the issue with checkboxes and component visibility. I did get it to partially listen for a checkbox to make another item visible using a validation function but there were two major issues:

1. The actual checkbox itself when clicked didn't do anything
2. The checkbox label did trigger the change but only after clicking elsewhere (remove focus)

Either way, this was not an ideal situation. I did however find a use-case to apply a validation function to a text field and based on the input, enable checkboxes (or other fields) as required. From a UX perspective it works because the user would want to click one of the empty checkboxes after entering a path and this action would make the text field loose focus – immediately enabling the checkboxes before the click is registered. I remembered this when I wrapped up and thought I'd share my findings.

Issue #5450, opened 2022-10-15
Sonarr uninstall not removing 'appdata' when erasing all data files


mreid-tt:
### Is this a new Bug?

- [X] I checked that the bug hasn't been reported before

### Package Name

Sonarr

### Package Version

20210717-19

### Device Model

DS916+

### Device Architecture

x86_64

### Firmware Version

DSM 7.1.1-42962 Update 1

### What happened?

When uninstalling Sonarr and selecting `Erase all of the package data files. (Not Recoverable)` the files in the AppData directory do not get removed. This results in an error on re-install of the package: `The DataMapper was unable to load the following field: 'Languages'`. This is as a result of mismatched configuration versions.

### Reproduction steps

```bash
1. Install Sonarr 20210717-19
2. Run internal updater to update to current version
3. Uninstall Sonarr
4. Reinstall Sonarr 20210717-19
5. Open Sonarr webpage
```

### Install Log

```log
`/var/log/packages/nzbdrone.log`

2022/10/14 20:59:43	install nzbdrone 20210717-19 Begin preinst
2022/10/14 20:59:43	===> Step preinst. USER=sc-nzbdrone GROUP=sc-download SHARE_PATH=
2022/10/14 20:59:43	install nzbdrone 20210717-19 End preinst ret=[0]
2022/10/14 20:59:43	install nzbdrone 20210717-19 Begin /bin/mkdir -p /volume1/@appstore/nzbdrone -m 755
2022/10/14 20:59:43	install nzbdrone 20210717-19 End /bin/mkdir -p /volume1/@appstore/nzbdrone -m 755 ret=[0]
2022/10/14 20:59:43	install nzbdrone 20210717-19 Begin /bin/rm -rf /volume1/@appstore/nzbdrone
2022/10/14 20:59:43	install nzbdrone 20210717-19 End /bin/rm -rf /volume1/@appstore/nzbdrone ret=[0]
2022/10/14 20:59:43	install nzbdrone 20210717-19 Begin /bin/mv -f /volume1/@tmp/synopkg/install.kmlfPO/package /volume1/@appstore/nzbdrone
2022/10/14 20:59:43	install nzbdrone 20210717-19 End /bin/mv -f /volume1/@tmp/synopkg/install.kmlfPO/package /volume1/@appstore/nzbdrone ret=[0]
2022/10/14 20:59:43	install nzbdrone 20210717-19 Begin /bin/rm -rf /var/packages/nzbdrone
[...]

PR #5456, opened 2022-10-16
Update sonarr 3.0.5 => 3.0.9

smaarn:
## Description

Just to keep up to date with latest version released of Sonarr.

## Checklist

- [x] ~~Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created~~

### Type of change

- [x] Package update

### DIFF
--- a/cross/sonarr/Makefile
+++ b/cross/sonarr/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = Sonarr
-PKG_VERS = 3.0.5.1144
+PKG_VERS = 3.0.9.1549
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME).main.$(PKG_VERS).linux.$(PKG_EXT)
 PKG_DIST_SITE = https://download.sonarr.tv/v3/main/$(PKG_VERS)
--- a/cross/sonarr/digests
+++ b/cross/sonarr/digests
@@ -1,3 +1,3 @@
-Sonarr.main.3.0.5.1144.linux.tar.gz SHA1 2335133139a46f1ce0d1b2cb1f566d30285d3d1e
-Sonarr.main.3.0.5.1144.linux.tar.gz SHA256 77aff6978d4dc80027979c815bd5a79579dd9bb50d39b1b2d14b3bb9e18058aa
-Sonarr.main.3.0.5.1144.linux.tar.gz MD5 675ce0ab2f1e643860ab1daab45b2842
+Sonarr.main.3.0.9.1549.linux.tar.gz SHA1 f475b4b070df9865c5ff777d532683f753151d97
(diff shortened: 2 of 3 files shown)


Safihre:
But why? The internal updater will just update it after the install?

smaarn:
Closing as redundant with internal updater and not adding any value besides.

bakerboy448:
> But why? The internal updater will just update it after the install?

Incorrect.

This PR is needed just like the other starr apps.

users who reinstall are required to manually purge appdata via ssh.  Why does the SynoCommunity require this manual work around in lieu of updating the base app images?

why is there suddenly an urge to no longer update the Starr apps to latest stables when historically that has never been the case?

bakerboy448:
Correction to the above. Due to this bug
https://github.com/SynoCommunity/spksrc/issues/5450

any users that reinstall sonarr MUST manually purge their appdata via ssh as uninstalling and purging will not.

additionally, if users don't enable auto update - and I don't believe the synology package does either - then sonarr would not autoupdate on install as claimed. It also does not appear update_required is set by the Syno sonarr package.  Thus Sonarr will absolutely never update on startup as claimed either.

smaarn:
@bakerboy448 if I'm reading you right this means I should be reopening this PR, did I get it right ?

bakerboy448:
correct @smaarn  :)

thanks for finally doing what SynoComm has not done since 3/27/2021

Safihre:
You seem to have quite some frustration with SynoCommunity.. We are just some people trying to do our best in our free time to work through the challenges posed by the Synology platform, all the different cpu architectures and cross compiling problems. There are a large number of packages that we attempt to support, but we can only do so much.. (I say "we", but I'm only a very small contributor). 

To your points:
There is no sudden urge to not update the Arr apps, historically we generally only updated the base versions when we updated something in the SynoCommunity framework or the scripts to start/stop the app. Otherwise we left them to auto update and that worked well for quite a few years.  
Releasing a new package every time there's a new version would take too much time and also disk space on our server. 
It seems DSM 7 introduced some new problems, like the AppData problem.
So let's try to fix that indeed. 
If additionally we need to configure auto update better, let's also do that. I am pretty sure that when I maintained the Sonarr package here a few years ago, auto update was enabled by default. If that changed, we should update accordingly.

So, let's fix the core problems.

bakerboy448:
> You seem to have quite some frustration with SynoCommunity.

because apps are continually not updated causing users headaches even though issues have been reported for years and then PRs that reduce user headaches are closed or ignored even though other apps have been updated in the same manner.  And the uninstall methods leave data behind requiring manual complex repairs for users, most synology users have no idea what terminal is nor how to use it.

The \*Arrs for years have never had auto update enabled by default.

We'll see how long it takes for the apps to be updated and if your words actually match your actions. So far continually PRs to update the apps have been closed or otherwise ignored.

mreid-tt:
hey @bakerboy448, looking at the comment on Sonarr not auto-updating after the package is installed, is there a separate issue or pull request for this? if not would you mind creating one as you seem to have a good understanding of the problem.

PR #5273, opened 2022-05-14
radarr: prevent downgrade on package update

hgy59:
## Description

- backup and restore existing radarr distribution to prevent downgrade
- adjust upgrade wizard (fre still needs translation)

To avoid troubles as in #5271 version 20220513-17 is removed in the repository.

Fixes #5271

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5467, opened 2022-10-30
Enlarge service start/stop timeout for *arr packages

### LINKED ISSUE #5133: DSM7 repair required for Radarr, Lidarr on reboot
I don't know if you guys are experiencing this but every time I reboot my NAS the *arr packages do not start. When I check in the package manager they all report that they require repair. I repair them and all is okay. While this is not a big deal it does add a bit of overhead which I didn't have under DSM6. Note that other community packages like SABnzbd and Transmission do not require such a repair on reboot.

hgy59:
## Description

- the package will be in status error and the package center will offer to repair it when timeout is too short (was 20 sec)
- fix service does start within timeout (as reported 60 seconds are enough)

Fixes  #5133

## Checklist

### Type of change

- [x] Bug fix


Safihre:
@hgy59 did you already release this?
Because based on feedback here, we should probably also update the base versions of these apps: https://github.com/SynoCommunity/spksrc/pull/5456#issuecomment-1295887770

hgy59:
@Safihre no, not released (published) yet.
For Radarr this is not an issue since #5273 prevents package downgrade.
Probably we can add such a prevention to sonarr, lidarr and prowlarr too?

PR #5485, opened 2022-11-10
Lidarr: avoid downgrade on package updates

hgy59:
## Description

- avoid downgrade on package updates 
- remove mono support
- enable automatic updates by default (on new package installation only)

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


spk/lidarr/src/service-setup-mono.sh
```diff
@@ -23,7 +23,7 @@ GROUP="sc-download"
 
 SERVICE_COMMAND="env PATH=${MONO_PATH}:${PATH} HOME=${HOME_DIR} LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib ${MONO} ${LIDARR}"
 SVC_BACKGROUND=y
-SVC_WAIT_TIMEOUT=60
+SVC_WAIT_TIMEOUT=90
```
bakerboy448, line 26: Is this file even needed anymore?

lidarr mono versions are EOL and unsupported.

lidarr v1 is dotnet and not mono

spk/lidarr/src/service-setup.sh
```diff
@@ -17,33 +19,45 @@ SVC_WAIT_TIMEOUT=90
...
-    mkdir -p ${CONFIG_DIR}/Lidarr
-    mv ${SYNOPKG_PKGDEST}/app/config.xml ${CONFIG_DIR}/Lidarr/config.xml
-    set_unix_permissions "${CONFIG_DIR}"
+    mkdir -p ${LIDARR_CONFIG_DIR}
+    mv ${SYNOPKG_PKGDEST}/app/config.xml ${LIDARR_CONFIG_DIR}/config.xml
+
+    echo "Set update required"
+    # Make Lidarr do an update check on start to avoid possible Lidarr
+    # downgrade when synocommunity package is updated
+    touch ${LIDARR_CONFIG_DIR}/update_required
```
bakerboy448, line 28: Also require users to have enabled auto updates (disabled by default)

but presumably most users would have it enabled or otherwise see the healthcheck saying to update

spk/lidarr/src/service-setup-mono.sh
hgy59: > lidarr mono versions are EOL and unsupported.

Thanks for the hint. Another caveat is, that you need to install mono, as the repository tells the DSM Package Manager that it depends on mono, even when lidarr based on dotnet is installed.

spk/lidarr/src/service-setup.sh
Safihre: How do we force enable Auto update for all the *arr programs? Is there a way?
Or at least for new installations.

hgy59: @bakerboy448 

the config.xml installed in the lidarr package has:
```xml
<UpdateMechanism>BuiltIn</UpdateMechanism>
```

Is this enough to  enable auto updates?

hgy59: Or does one need to configure an RSS indexer therefore?
```
ERROR: No indexers available with RSS sync enabled, Lidarr will not grab new releases automatically
```

hgy59: just found id (it was hidden in the advanced general settings)

If automatic updates are enabled, the config.xml contains:
```xml
<UpdateAutomatically>True</UpdateAutomatically>
```

Safihre:
@hgy59 can we do followup PR's to update Radarr and Sonarr and enable automatic updates?
Then we resolve most of @bakerboy448 concerns I think.

bakerboy448 (commented):
master should be the branch used, not develop.

And only non-defaults *should* need to be wrote to the config

mreid-tt:
hey @hgy59, I did a test to confirm that the downgrade does not occur on installation of this update. The methodology was as follows:

1. Fresh install of package version 20210322-7 from SynoCommunity
2. Enable TMP folder upgrades using this [comment](https://github.com/SynoCommunity/spksrc/issues/5303#issuecomment-1279819536)
3. Update Lidarr to version 1.1.0.2649 using internal updater
4. Upgrade package using version 20221111-10 from this PR

Once the above was done I was able to confirm that the installed version did not downgrade to Lidarr version 1.0.2.2592 which is included in the package but remained at the higher version.

Some screenshots are included below:
Screenshot 2022-11-14 at 8 42 28 AM
Screenshot 2022-11-14 at 8 42 36 AM

PR #5515, opened 2022-12-09
Sonarr: fixes for services and package modernisation

mreid-tt:
## Description

This replaces #5511 which built on the work from #5485 to address a number of issues in Sonarr as discussed in https://github.com/SynoCommunity/spksrc/pull/5471#issuecomment-1317381973 and summarised below:

1. Update package name to 'Sonarr v3'
2. Upgrade to Sonarr 3.0.9.1549
3. Avoid Sonarr downgrade on package update
4. Fix usage of TMPDIR and remove obsolete folders
5. Remove support for legacy (v2) versions

Fixes #5450, Closes #5471

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [x] Package update
- [x] includes a small framework change: include the STATUS (i.e. INSTALL, UPGRADE, UNINSTALL) in the installer log line created for each step

PR #5070, opened 2022-01-11
Mono: Fix TLS on DSM7

publicarray:
## Description
This should fix the trust store containing expired certificates in mono applications by using a different path. Dependent packages may need to be updated.

Fixes #5051

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #3803, opened 2019-11-12
Update Sonarr to v3

maxrogers:
# Test package for Sonarr V3 
### https://github.com/Safihre/spksrc/releases/tag/sonarr-v3-build 

⚠ To see which version you need: https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model 
⚠ Models that have `(x64)` listed on the wiki page should use the general x64 testing package.

### Checklist
- [x] Build rule `all-supported` completed successfully
- [x] Package upgrade completed successfully
- [x] New installation of package completed successfully

I just did a successful upgrade from Sonarr v2 to this package (v3) with no issues. It would be great to get a few experienced users to test and share their success/results going from v2 to v3 with this package. @Safihre can you make packages for people to test with? Thank you!

_Motivation:_  To add a v3 Sonarr package
_Reference:_  https://github.com/SynoCommunity/spksrc/pull/3518#issuecomment-552331259

PR #5520, opened 2022-12-12
[WIP] Sonarr: (dotNET) v4 build

### LINKED ISSUE #5051: Sonarr 3.0.6.1342 - Certificate validation errors after updating to Mono 5.20.1.34-18
### Setup
_Package Name:_ Sonarr
_Package Version:_ 20210717-19

_NAS Model:_ DS916+
_NAS Architecture:_ INTEL Pentium N3710
_DSM version:_ DSM 7.0.1-42218

### Expected behavior
Following the update of Mono to version 5.20.1.34-18, Sonarr is restarted. Once restarted, I should be able to check for updates without error from System -> Updates

### Actual behavior
On checking for updates I only get an error: Failed to fetch updates
I also note in the System -> Events a number of X509CertificateValidationService and SonarrErrorPipeline errors

### Steps to reproduce
_1._ Update Mono to latest version 5.20.1.34-18
_2._ Sonarr reboots as a dependent package
_3._ Errors show in the System -> Events

#### Package log
You find service log and other log files in `/var/packages/{package}/target/var`._

> nzbdrone_install.log -- https://pastebin.com/5EzrSXca
> nzbdrone.log -- https://pastebin.com/QjVGn1mu

#### Other logs

> synopkg.log -- https://pastebin.com/75sn1W1y

### Known workaround (see: https://github.com/SynoCommunity/spksrc/issues/5051#issuecomment-1009758625)

1. Within DSM, enable SSH service in _Control Panel > Terminal & SNMP_ and click apply
2. Using [Terminal](https://support.apple.com/en-gb/guide/terminal/apd5265185d-f365-44cb-8b09-71a064a42125/mac) (MacOS) connect to the NAS using `ssh -l [admin username] [NAS address]` or using [Putty](https://www.putty.org/) (Windows) connect to the network address of your NAS
3. Enter the required admin password and press enter
4. Enter the following command: `sudo /var/packages/mono/target/bin/cert-sync /etc/ssl/certs/ca-certificates.crt` and press enter
5. Enter the required admin password and press enter. When complete you should see the line _Import process completed_
6. Enter the following command: `sudo chmod -R a+rX /usr/share/.mono` and press enter
7. Disconnect the SSH session by typing `exit` and press enter
8. Within DSM, disable the SSH service in _Control Panel > Terminal & SNMP_ and click apply

mreid-tt:
## Description

This is an attempt to deploy a .NET build (v4) of Sonarr for Synology. As this is a major change, the package internal name has also changed from `nzbdrone` to `sonarr`. Unfortunately, this means that an upgrade from the previous deployment is not possible as the package name is no longer the same. If you try to deploy it, a conflict with the port used from the old package will be flagged. As such to test this deployment you should:

1. Uninstall the old Sonarr (but keep existing files)
2. Uninstall the SynoCommunity Mono
3. Install this new Sonarr
4. The old config files (if detected) will be migrated to the new deployment

Fixes #5051, #5450

## Checklist

### Type of change

- [x] Bug fix
- [x] New Package
- [x] Package update


mreid-tt:
@bakerboy448, this early test build was inspired by your comment:

> Sonarr on .Net also already exists for internal testing. https://github.com/Sonarr/Sonarr/tree/widowmaker

Additionally, based on this recommendation:

> For this SSL issue it seems that SynoCommunity's mono package has a bug and fails to sync certs upon install. This should be corrected as soon as it can given all the support issues it's causing.

There has been no apparent success based on https://github.com/SynoCommunity/spksrc/pull/5070

As such, this rough test build was created to determine the feasibility of using the early release of Sonarr on dotNET to resolve these issues.

bakerboy448:
Widowmaker is already not supported

The sonarr develop branch is v4 beta.

mreid-tt:
> The sonarr develop branch is v4 beta.

Thanks for the heads up. The PR has been updated with this build. Interested to hear your thoughts on the PR.

spk/sonarr/Makefile
```diff
@@ -1,22 +1,19 @@
-SPK_NAME = nzbdrone
+SPK_NAME = sonarr
 SPK_VERS = $(shell date +%Y%m%d)
-SPK_REV = 20
+SPK_REV = 22
```
hgy59, line 3: as this is a new package, we can start at revision 1
```suggestion
SPK_REV = 1
```

cross/sonarr/Makefile
```diff
@@ -1,21 +1,35 @@
...
 PKG_EXT = tar.gz
-PKG_DIST_NAME = $(PKG_NAME).main.$(PKG_VERS).linux.$(PKG_EXT)
-PKG_DIST_SITE = https://download.sonarr.tv/v3/main/$(PKG_VERS)
+PKG_DIST_NAME = $(PKG_NAME).develop.$(PKG_VERS).linux-$(PKG_DIST_ARCH).$(PKG_EXT)
+PKG_DIST_SITE = https://download.sonarr.tv/v4/develop/$(PKG_VERS)
 PKG_DIR = $(PKG_NAME)
 
 DEPENDS =
 
+UNSUPPORTED_ARCHS = $(i686_ARCHS) $(ARMv7L_ARCHS) $(ARMv5_ARCHS) $(PPC_ARCHS)
```
hgy59, line 10:
```suggestion
UNSUPPORTED_ARCHS = $(DOTNET_UNSUPPORTED_ARCHS)
```

spk/sonarr/Makefile
```diff
@@ -1,22 +1,19 @@
...
-
-SPK_DEPENDS = "mono>3.6"
+UNSUPPORTED_ARCHS = $(DOTNET_UNSUPPORTED_ARCHS)
 
 MAINTAINER = SynoCommunity
 DESCRIPTION = Sonarr is a PVR for newsgroup and torrent users. It can monitor multiple RSS feeds for new episodes of your favourite shows and will grab, sorts and rename them. It can also be configured to automatically upgrade the quality of files already downloaded if a better quality format becomes available.
 DESCRIPTION_FRE = Sonarr est un PVR pour les utilisateurs de groupes de discussion et torrents. Il peut surveiller plusieurs flux RSS pour les nouveaux épisodes de vos séries préférées et saisira, sortes et les renomme. Il peut également être configuré pour mettre à jour automatiquement la qualité des fichiers déjà téléchargés si un meilleur format de qualité devient disponible.
 DESCRIPTION_SPN = Sonarr es un PVR para los usuarios de grupos de noticias y torrents. Se puede controlar múltiples canales RSS para nuevos episodios de sus programas favoritos y se agarra, tipo y les cambia el nombre. También puede ser configurado para actualizar automáticamente la calidad de los archivos ya descargados si un formato de mejor calidad disponible.
 DISPLAY_NAME = Sonarr
```
hgy59, line 14: We must use different Display Names for nzbdrone and sonarr (internal package names). Otherwise there will be two different packages in the repository (and in DSM Package Manager) with the same name.

spk/sonarr/Makefile
```diff
@@ -1,22 +1,19 @@
...
 STARTABLE = yes
-CHANGELOG = "Enlarge the service start/stop timeout to 90 seconds."
+CHANGELOG = "1. Enlarge the service start/stop timeout to 90 seconds<br>2. Avoid Sonarr downgrade on package update<br>3. Implement fix for alternate TMP directory<br>4. Upgrade to Sonarr 4.0.0.87"
```
hgy59, line 16: As this is a new package we could define somthing like:
```suggestion
CHANGELOG = "Initial Sonarr v4 package with version 4.0.0.87. This package replaces a former Sonarr v3 installation (internally named nzbdrone)."
```

spk/sonarr/src/service-setup.sh
```diff
@@ -1,104 +1,89 @@
...
 {
-    mkdir -p ${CONFIG_DIR}
-    set_unix_permissions "${CONFIG_DIR}"
+    # use install_log to write to installer log file.
+    install_log "validate_prereplace ${SYNOPKG_PKG_STATUS}"
+
+    # check if the installed distribution is a legacy version
+    if [ -f "${LEGACY_SYNOPKG_PKGDEST}/share/NzbDrone/NzbDrone.exe" ]; then
+        # v2 installed
+        exit 1
```
hgy59, line 32: please give feedback to the user, why package update is not possible.
something like:
```
      echo "Update from NzbDrone (Sonarr v2.x) is not supported"
      exit 1
```

spk/sonarr/src/service-setup.sh
```diff
@@ -1,104 +1,89 @@
...
-    if [ -d "${LEGACY_CONFIG_DIR}" ]; then
-        echo "Moving ${LEGACY_CONFIG_DIR} to ${CONFIG_DIR}"
-        mv ${LEGACY_CONFIG_DIR} ${CONFIG_DIR} >> ${INST_LOG} 2>&1
-    fi
-    if [ ! -d ${CONFIG_DIR} ]; then
-        # Create, in case it's missing for some reason
-        mkdir -p ${CONFIG_DIR} >> ${INST_LOG} 2>&1
+    # use echo to write to the installer log file.
+    echo "service_prereplace ${SYNOPKG_PKG_STATUS}"
+
```
hgy59, line 44: In the generic service installer we already have
```
log_step "prereplace"
```

So there is already an entry in install log for the `prereplace` step.


As it might be relevant for all packages, I propose to add  `${SYNOPKG_PKG_STATUS}` to the install log in `log_step` function defined in `mk/spksrc.service.installer.functions` like:

```
log_step ()
{
    install_log "===> Step $1. STATUS=${SYNOPKG_PKG_STATUS} USER=${USER} GROUP=${GROUP} SHARE_PATH=${SHARE_PATH}"
}
```

spk/sonarr/Makefile
```diff
@@ -35,6 +32,17 @@ POST_STRIP_TARGET = sonarr_extra_install
 
 include ../../mk/spksrc.spk.mk
 
+# use alternate TMPDIR as /tmp might be too small and not accessible on DSM >= 7.1.
+USE_ALTERNATE_TMPDIR = 1
+
+# replace previous legacy package
+INSTALL_REPLACE_PACKAGES = nzbdrone
```
hgy59, line 39: We could nail the package replacement to specific versions. With the following change, there will be no need to handle existing sonarr v2 versions.
```suggestion
INSTALL_REPLACE_PACKAGES = nzbdrone>=20210329
```

bakerboy448:
> To continue supporting Sonarr v3 with mono dependency (for archs that do not support dotnet core), we need to keep (and rename) existing `cross/sonarr/*` and `spk/sonarr/*` to `cross/nzbdrone/*` and `spk/nzbdrone/*` (or use sonarr3 as name for former package, but internally it is called nzbdrone).
> 
> 
> 
> BTW:
> 
> We must not forget to adjust the specific nzbdrone/sonarr handling in `.github/actions/prepare.sh`.
> 
> 
> 
> 
> 
> 👍 👍 👍 for the implementation of package replacement ❗ 

Given nzbdrone is the legacy no longer supported v2.  I'd suggest having v3 as Sonarr3.  And the most recent Sonarr version (beta or not) as Sonarr.

Also Sonarr v4 should very clearly indicate it is BETA and not stable.

hgy59:
@mreid-tt just a small suggestion while you are working... 
I would like to keep the former icon for the sonarr3 package. This gives (a small) visual difference between the two packages.


Further suggestion:
Maybe we first need to publish an update of the sonarr3 package 
- everybody seeds the new display name for the current package before the package replacement
- we can use such a release for package replacement

bakerboy448:
> @mreid-tt just a small suggestion while you are working... 
> 
> I would like to keep the former icon for the sonarr3 package. This gives (a small) visual difference between the two packages.
> 
> 
> 
> 
> 
> Further suggestion:
> 
> Maybe we first need to publish an update of the sonarr3 package 
> 
> - everybody seeds the new display name for the current package before the package replacement
> 
> - The new display name is used on the installed package (and not only in the package repo)
> 
> - we can use such a release for package replacement
> 
> 

Sonarr v3 is the official stable version.  Sonarr v4 is still in beta and not released as stable.  No package replacement should be occurring until v4 is stable.

Either users can run the v3 stable Sonarr package. Or test the v4 beta.   But the v4 beta should not be forced as a replacement currently.

hgy59:
> Also Sonarr v4 should very clearly indicate it is BETA and not stable.

@bakerboy448 is there any schedule when Sonarr v4 will get stable?
- we should not publish v4 before it is stable
- we should not replace a stable v3 by a prerelease v4

mreid-tt:
> @mreid-tt just a small suggestion while you are working... I would like to keep the former icon for the sonarr3 package. This gives (a small) visual difference between the two packages.

Sure, that's not an issue, but I will use the larger 512x512px size to be consistent with the other packages.

> Further suggestion: Maybe we first need to publish an update of the sonarr3 package
> 
> * everybody seeds the new display name for the current package before the package replacement
> * The new display name is used on the installed package (and not only in the package repo)
> * we can use such a release for package replacement

Yes, this would be a good idea. Would you want me to implement the change of the external package name from `sonarr` to `sonarr3` into https://github.com/SynoCommunity/spksrc/pull/5515 and you can move that to be published? Then I would re-base this PR.

bakerboy448:
As always with the \*arrs there's never any ETA for a new release.

As a reminder/context - v3 was in "beta" (and working better than v2) for several years before it was finally released as stable and replaced nzbdrone.

mreid-tt:
hey @hgy59, I was experimenting with adjusting the port numbers of existing deployments as commented on here: https://github.com/SynoCommunity/spksrc/pull/3803/files#r596284050. Unfortunately, when the package builds, I don't get the correct `service-setup` file being generated in the package. It seems to be pulling from the old `spk/sonarr/src/service-setup.sh` code before it was moved to `spk/sonarr3/src/service-setup.sh`.

I don't know if this is the way GitHub is setup or that the changes in #5515 need to be merged into the master first before I can get a new package generated under the old name? Can't seem to continue development testing until this is resolved.

hgy59:
@mreid-tt is it required to use a different port for Sonarr v4?
I thought that we could use the same port, when a package is replaced by a new one.

Further Opinions
- we should not only replace sonarr v3 by v4 for DSM 7 but also for DSM 6
- for all arch/dsm version combinations that successfully build Sonarr v4 we should add the package replacement.
- Only DSM 5.x will stay on Sonarr v3
- currently we build DSM 6 packages for DSM 6.1-15047 - package replacement is supported from DSM 6.1-15117 (i.e. 6.1.2). So we have to ajust this for our DSM 6 packages. Both DSM 6.1 versions are fairly old (03/2017 and 09/2017) and all models supporting DSM 6.1 have updates at least for DSM 6.2.4.
- SRM 1.x is compatible with DSM 6 installer, but we have to approve this, before deploying packages with replacement of Sonarr v3.

mreid-tt:
@hgy59, from my testing, each time I try to install a package where the port number is not changed, the DSM reports the conflict and the installation cannot proceed. Now from what I read in the [documentation](https://global.download.synology.com/download/Document/Software/DeveloperGuide/Os/DSM/All/enu/DSM_Developer_Guide_7_enu.pdf) about `prereplace`, this is a new feature that is described as:

> It can be used to do data migration when install_replace_packages is defined in INFO for
package replacement. Package replacement will be aborted for non-zero returned value.

Unfortunately, when I look at the installation log, I don't see any evidence that the `prereplace` function is called. Strangely enough on my last test there is a log entry for `postreplace`. When I decompressed the [package](https://github.com/SynoCommunity/spksrc/suites/9864618877/artifacts/475855805) for DSM7, this is where I noted that the `prereplace` function is not present in the included `service-setup` file. As such, I don't really know if I need to force a port change or if the `prereplace` function will permit a package removal before it detects a conflict. Thus, I am at a loss of how to proceed.

Thanks for your additional comments:

> * we should not only replace sonarr v3 by v4 for DSM 7 but also for DSM 6

With regard to the DSM6 config, I'll add it once I have successfully tested a package replace under DSM7 (I don't have an active DSM6 deployment so I'll have to spend time setting one up). Given that in the DSM 6.0 [documentation](https://global.download.synology.com/download/Document/Software/DeveloperGuide/Firmware/DSM/6.0/enu/DSM_Developer_Guide_6_0.pdf), the `prereplace` and `postreplace` functions don't exist I don't know if these will be called at any point during the deployment. As such some workaround (like the alternate port install and port swap I've been tinkering with) would likely be needed there. Otherwise, users would have to manually remove their Sonarr v3 and then have the deployment pick up the config files left behind. This is what I've been trying to avoid but we'll see how the testing goes once I am able to resume.

> * for all arch/dsm version combinations that successfully build Sonarr v4 we should add the package replacement.

I thought the SynoCommunity spksrc platform would include the code I provided for all relevant architectures. The only adjustment I thought would be needed would be to add DSM6 support.

> * Only DSM 5.x will stay on Sonarr v3

Yes, that makes sense. I believe that we can just configure the build not to make anything for DSM5 (if not already configured to do so).

> * currently we build DSM 6 packages for DSM 6.1-15047 - package replacement is supported from DSM 6.1-15117 (i.e. 6.1.2). So we have to ajust this for our DSM 6 packages. Both DSM 6.1 versions are fairly old (03/2017 and 09/2017) and all models supporting DSM 6.1 have updates at least for DSM 6.2.4.

Ah, if the `prereplace` and `postreplace` functions are recognised in DSM 6.1 then this should be straightforward to test. As for DSM 6.0 we can work on that together.

> * SRM 1.x is compatible with DSM 6 installer, but we have to approve this, before deploying packages with replacement of Sonarr v3.

I'm not sure what SRM 1.x is exactly. Would this be the reason for the packager not including the correct `service-setup.sh` content?

mreid-tt:
> @mreid-tt is it required to use a different port for Sonarr v4?
> I thought that we could use the same port, when a package is replaced by a new one.

I did some more investigation and apparently, a method to deploy and avoid the port conflict check was attempted before in https://github.com/SynoCommunity/spksrc/pull/3803/commits/4d88fa9c2248c03a0e24bb45635f79e0e22d7e32. I've taken this logic and included it in the latest commit and we'll see how far that gets us.

mreid-tt:
@hgy59 So the package build worked correctly this time and the contents of the package file `service-setup` was consistent with the content from the revised `service-setup.sh`. I was also able to confirm that the package installed without flagging a conflicting port with the addition of the `checkport` code however the `prereplace` code still did not fire (see installation log: https://pastebin.com/6nRLNBKb). I need to get this resolved somehow.

Also, I don't quite know what should go into the functions themselves. For example, having `nzbdrone` listed in the `install_replace_packages` section should remove it but it doesn't as the app is still installed. So should the `postreplace` contain all the commands to manually remove the package? If you have any guidance here it would be appreciated.

mreid-tt:
hey @hgy59, so I did some further checks and found that the `prereplace` was in fact being run but appeared to be terminating prematurely. Based on the log at `/var/log/synopkg.log` (see https://pastebin.com/5e1qEvbJ), we note the following section:

```
2022/12/15 06:27:19 install   Begin prereplace
2022/12/15 06:27:19 stop nzbdrone: begin to stop version 20210717-19
2022/12/15 06:27:29 stop nzbdrone: stop version 20210717-19 failed, result 0
2022/12/15 06:27:29 install   End prereplace ret=[0]
```

What this appears to be is the function starting, passing the `validate_prereplace` successfully and then stopping in the `service_prereplace` function. Specifically on this line:

https://github.com/SynoCommunity/spksrc/blob/e13d4d621d14033353cb8fd9890c1167a4122d47/spk/sonarr/src/service-setup.sh#L54

I believe it attempts to stop the old NzbDrone but fails to complete because the command is not executed with elevated privileges. From checks elsewhere in the code, all of the scripts under DSM7 seem to be executed with the privilege of the package. The documentation mentions:

> ### Package Framework Changes
> **1. Force lower privilege for package**
All packages should provide conf/privilege with package in run-as explicitly. Any privileged operation should be accomplished via resource worker

I am not able to identify how to use this resource worker to give the script the required privilages in the `service_prereplace` function and have put some comments above it as a reminder that this needs resolution.

Hopefully you can shed some light on how to resolve this.

PR #5441, opened 2022-10-02
Docker: Install rustc & upgrade image to Debian 11 bullseye

th0ma7:
## Description

Docker: Install `rustc` & upgrade image to Debian 11 bullseye.  This PR is only aimed at adding `rustc` compiler onto our default docker image.  It's integration is within PR #5435 but it requires the docker image being updated in master in order to allow github-action to work.

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes small framework changes
- [x] This change requires a documentation update (e.g. Wiki)

### Testing HOWTO
Clone my copy of this PR:
```
$ git clone https://github.com/th0ma7/spksrc.git
$ cd spksrc
$ git checkout docker-rustc-debian11
```
Create the updated docker image (may need `$ sudo usermod -aG docker $USER` for proper user permissions):
```
$ docker build - < Dockerfile
...
Successfully built 882a0451f1bb
```
Access the docker image:
```
$ docker run -it -v $(pwd):/spksrc 882a0451f1bb /bin/bash
```
Build a `python310` package or what ever you are interested at testing:
```
root@2ff602f23b71:/spksrc# cd spk/python310
root@2ff602f23b71:/spksrc/spk/python310/# make all-supported
```
To remove all docker images afterwards:
```
$ docker container rm -f $(docker container ls -aq)
```

### Testing completed succesfully (`all-supported`)
- [x] ffmpeg

PR #5522, opened 2022-12-14
mono: fix build with updated dev. env.

hgy59:
## Description

This is another followup to #5441.
With the updated dev. env. default python is not python 2 anymore.
- patch makefile templates for explicit use of python 2

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/cross/mono/patches/002-explicit-use-of-python2.patch
+++ b/cross/mono/patches/002-explicit-use-of-python2.patch
@@ -1,0 +1,109 @@
+# To use python 2.x we must call python2 nowadays.
+# 
+--- scripts/submodules/versions.mk.orig	2019-07-16 18:16:12.000000000 +0000
++++ scripts/submodules/versions.mk	2022-12-14 00:04:43.894608574 +0000
+@@ -10,11 +10,11 @@
+ 
+ define ValidateVersionTemplate
+ #$(eval REPOSITORY_$(2):=$(shell test -z $(3) && echo $(1) || echo "$(3)"))
+-#$(eval DIRECTORY_$(2):=$(shell python $(SCRIPT) $(SUBMODULES_CONFIG_FILE) get-dir $(1)))
++#$(eval DIRECTORY_$(2):=$(shell python2 $(SCRIPT) $(SUBMODULES_CONFIG_FILE) get-dir $(1)))
+ #$(eval DIRECTORY_$(2):=$(shell test -z $(DIRECTORY_$(2)) && echo $(1) || echo $(DIRECTORY_$(2))))
+-#$(eval MODULE_$(2):=$(shell python $(SCRIPT) $(SUBMODULES_CONFIG_FILE) get-url $(1)))
+-#$(eval NEEDED_$(2)_VERSION:=$(shell python $(SCRIPT) $(SUBMODULES_CONFIG_FILE) get-rev $(1)))
+-#$(eval $(2)_BRANCH_AND_REMOTE:=$(shell python $(SCRIPT) $(SUBMODULES_CONFIG_FILE) get-remote-branch $(1)))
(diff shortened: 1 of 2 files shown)


mreid-tt:
@hgy59 was something changed with permissions? I tried to do a clean-install with this build and I didn't get any of the certificate import errors you mentioned in https://github.com/SynoCommunity/spksrc/issues/5051#issuecomment-1005163408.
The installation log also looks clean -- https://pastebin.com/YQ78RegU. I was curious if there was something specific which may have changed or if this was an anomaly on my end.

hgy59:
@mreid-tt This PR didn't change anything (IMHO) it fixed the build only, as python defaults now to python 3 and the build has scripts that need python 2.
#5070 mightfix this. Did you take mono from the repo or probably from #5070?

mreid-tt:
hmm, looking a bit more in the file system I think an uninstall and re-install doesn't in fact do a complete removal as there are many mono files left behind. This is what I suspect happened when I installed this build - the previous permissions set persisted and allowed the deployment to go in cleanly. I'll try that repo you mentioned but I think I'll need to start from a completely fresh Virtual DSM to do a true test

mreid-tt:
@hgy59, I've taken another look at this and while you are correct in that none of the files that changed in this PR should have affected the cert import I believe that something **has** changed. Take a look at this log file from a standard installation: https://controlc.com/d893f758. In it you see many lines like:

```
2023/02/12 15:23:35	System.UnauthorizedAccessException: Access to the path "/usr/share/.mono" is denied.
```

This is compared to the log file I posted previously (https://pastebin.com/YQ78RegU) which looks like this:

```
2022/12/23 03:16:13	Certificate added: CN=ACCVRAIZ1, OU=PKIACCV, O=ACCV, C=ES
```

Now I looked into why my results were inconsistent before and I found that the folder at `/usr/share/.mono` was not being removed when mono was uninstalled. However, I've manually removed it and recreated the issue with a clean install of the existing version in the repository. From there I uninstalled mono, rebooted and re-installed with the version from this build and the installation is still clean with the certificate files being added to `/var/packages/mono/target/var/.mono` as part of the installer.

Now, looking into the code I see this was something that the 'native' patches for mono should have fixed many years ago:
https://github.com/SynoCommunity/spksrc/blob/0b670e907070f5997569268409018ccaff8b3eae/native/mono/patches/001-relocate-specialfolders.patch#L16-L17

As such, I'm not sure why this is not working for the build in the repository. Perhaps something else in the framework prevented this from working at the last time mono was merged and published in #5029?

hgy59:
@mreid-tt we had times where patches did not work for native packages (and toolchains). (probably until #4852, i.e. December 2022)
So it might be that the published packages are missing this patch.

BTW we must update the patch to use `/var/packages/mono/var` instead of `/var/packages/mono/target/var` for DSM 7 compatibility.

mreid-tt:
> BTW we must update the patch to use `/var/packages/mono/var` instead of `/var/packages/mono/target/var` for DSM 7 compatibility.

Yes, I think this is the case based on the upgrade install I just tried. The logs are found here: https://pastebin.com/KmLxtnNL. In it we see that as part of the upgrade the certs are generated in `/var/packages/mono/var/` but then are automatically moved to `/var/packages/mono/var` with the line:

```
2023/02/12 16:52:57	Begin syno_sync_var_folder
2023/02/12 16:52:58	Install files from var folder
2023/02/12 16:52:58	/bin/rsync -avh --ignore-existing --remove-source-files /volume1/@appstore/mono/var/ /volume1/@appdata/mono
[---snip---]
2023/02/12 16:52:59	/bin/rsync -avh --remove-source-files /volume1/@appstore/mono/var/ /volume1/@appdata/mono
[---snip---]
2023/02/12 16:52:59	End syno_sync_var_folder
```

So the certificates are still there but are unable to be referenced. I'll try to author a new PR for this but this is way out of my comfort zone so I'll need a lot of reviewers.

PR #5424, opened 2022-09-11
Update toolchains

hgy59:
## Description

Still waiting for synology to provide toolchains for SRM 1.3
see https://community.synology.com/enu/forum/2/post/153407

Fixes #5402

### Type of change

- [x] Includes small framework changes

Issue #4790, opened 2021-08-16
[jellyfin] Unable to start service on ds414


jasonfenglu:
### Setup
_Package Name:_ jellyfin
_Package Version:_  10.7.6-3

_NAS Model:_ ds414
_NAS Architecture:_  Armadaxp
_DSM version:_ DSM 6.2.4-25556

### Expected behavior
It should be activated after turned on in GUI. 

### Actual behavior
Not being activated. 

### Steps to reproduce
_1._ Start jellyfin from Package Center

#### Package log
```
cat /var/packages/jellyfin/target/var/jellyfin.log

Mon Aug 16 12:50:56 CST 2021
Starting jellyfin command env LD_LIBRARY_PATH=/volume1/@appstore/jellyfin/lib /volume1/@appstore/jellyfin/share/jellyfin --service  --package-name synology  -d /volume1/@appstore/jellyfin/var/data  -C /volume1/@appstore/jellyfin/var/cache  -c /volume1/@appstore/jellyfin/var/config  -l /volume1/@appstore/jellyfin/var/log  -w /volume1/@appstore/jellyfin/web  --ffmpeg /var/packages/ffmpeg/target/bin/ffmpeg
[12:51:08] [INF] [1] Main: Jellyfin version: 10.7.6
[12:51:08] [INF] [1] Main: Environment Variables: ["[JELLYFIN_LOG_DIR, /volume1/@appstore/jellyfin/var/log]", "[DOTNET_BUNDLE_EXTRACT_BASE_DIR, /var/packages/jellyfin/target/var//dotnet_bundle_extract]"]
[12:51:08] [INF] [1] Main: Arguments: ["/volume1/@appstore/jellyfin/share/jellyfin.dll", "--service", "--package-name", "synology", "-d", "/volume1/@appstore/jellyfin/var/data", "-C", "/volume1/@appstore/jellyfin/var/cache", "-c", "/volume1/@appstore/jellyfin/var/config", "-l", "/volume1/@appstore/jellyfin/var/log", "-w", "/volume1/@appstore/jellyfin/web", "--ffmpeg", "/var/packages/ffmpeg/target/bin/ffmpeg"]
[12:51:08] [INF] [1] Main: Operating system: Linux
[12:51:08] [INF] [1] Main: Architecture: Arm
[12:51:08] [INF] [1] Main: 64-Bit Process: False
[12:51:08] [INF] [1] Main: User Interactive: True
[...]

PR #5524, opened 2022-12-16
Sonarr v4 beta

### LINKED ISSUE #5051: Sonarr 3.0.6.1342 - Certificate validation errors after updating to Mono 5.20.1.34-18

mreid-tt:
## Description

This is a new deployment of Sonarr v4 beta (.NET) for Synology which replaces #5520. As this is a major change, the package internal name has also changed from `nzbdrone` to `sonarr`. The upgrade steps from Sonarr v3 would thus be:

1. Installation of Sonarr v4 package
2. Automatic migration of previous configuration data
3. Automatic uninstallation of Sonarr v3
4. (optional) Uninstallation of Mono if no longer required

Fixes #5051

## Checklist

### Type of change

- [x] Bug fix
- [x] New Package


mreid-tt:
@hgy59, continuing the discussion in this new PR as the old one was closed after I rebased...

> @mreid-tt is it required to use a different port for Sonarr v4? I thought that we could use the same port, when a package is replaced by a new one.

I did some more investigation and apparently, a method to deploy and avoid the port conflict check was attempted before in https://github.com/SynoCommunity/spksrc/pull/3803/commits/4d88fa9c2248c03a0e24bb45635f79e0e22d7e32. I've taken this logic and included it in the latest commit and I was able to confirm that the package installed without flagging a conflicting port with the addition of the `checkport` field.

Thanks for your additional comments:

> * we should not only replace sonarr v3 by v4 for DSM 7 but also for DSM 6

With regard to the DSM6 config, I'll add it once I have successfully tested a package replace under DSM7. I may need some assistance with this as there seems to be a few other things typically done with that older OS.

> * for all arch/dsm version combinations that successfully build Sonarr v4 we should add the package replacement.

I thought the SynoCommunity spksrc platform would include the code I provided for all relevant architectures. Is there something more I would need to do?

> * Only DSM 5.x will stay on Sonarr v3

Yes, that makes sense. I believe that we can just configure the build not to make anything for DSM5 (if not already configured to do so).

> * currently we build DSM 6 packages for DSM 6.1-15047 - package replacement is supported from DSM 6.1-15117 (i.e. 6.1.2). So we have to ajust this for our DSM 6 packages. Both DSM 6.1 versions are fairly old (03/2017 and 09/2017) and all models supporting DSM 6.1 have updates at least for DSM 6.2.4.

Thanks for confirming that the `prereplace` and `postreplace` functions are recognised in DSM 6.1. As for DSM 6.0 we can work on that together.

> * SRM 1.x is compatible with DSM 6 installer, but we have to approve this, before deploying packages with replacement of Sonarr v3.

I'm not sure what SRM 1.x is exactly but from a bit of searching I see that SRM is some type of O/S like DSM? I came across it in a post on #5424 where you were asking Synology for toolchains for SRM 1.3. From the link you posted there, it seems that Synology eventually provided a response. Let me know what would be needed from my end.

mreid-tt:
@hgy59, so I did some deployment tests and the `prereplace` flag does not appear in the installation log (see: https://pastebin.com/6nRLNBKb). It does however seem to be called but appears to be terminating prematurely. Based on the log at `/var/log/synopkg.log` (see https://pastebin.com/5e1qEvbJ), we note the following section:

```
2022/12/15 06:27:19 install   Begin prereplace
2022/12/15 06:27:19 stop nzbdrone: begin to stop version 20210717-19
2022/12/15 06:27:29 stop nzbdrone: stop version 20210717-19 failed, result 0
2022/12/15 06:27:29 install   End prereplace ret=[0]
```

What this appears to be is the function starting, passing the `validate_prereplace` successfully and then stopping in the `service_prereplace` function. Specifically on this line:

https://github.com/SynoCommunity/spksrc/blob/e13d4d621d14033353cb8fd9890c1167a4122d47/spk/sonarr/src/service-setup.sh#L54

I believe it attempts to stop the old NzbDrone but fails to complete because the command is not executed with elevated privileges. From checks elsewhere in the code, all of the scripts under DSM7 seem to be executed with the privilege of the package. The documentation mentions:

> ### Package Framework Changes
> **1. Force lower privilege for package**
All packages should provide conf/privilege with package in run-as explicitly. Any privileged operation should be accomplished via resource worker

I am not able to identify how to use this resource worker to give the script the required privilages in the `service_prereplace` function and have put some comments above it for the executables which may need a resource worker.

I also did some further testing with the commands commented out but the remainder of the `prereplace` script does not seem to accomplish its intended function. Specifically, the old configuration files are not migrated.

I don't quite know what should go into the functions themselves. For example, having `nzbdrone` listed in the `install_replace_packages` section should remove it but it doesn't as the app is still installed. So should the `postreplace` contain all the commands to manually remove the package? If you have any guidance here it would be appreciated.

mreid-tt:
I've also done an upgrade test from 20221216-21 (Sonarr v3) to 20221217-1 (Sonarr v4) to compare DSM 6.2 to DSM 7.1 and found some interesting divergences. Most importantly, under DSM 6.2 the upgrade function actually works and removes the previous version.

**Video and logs from DSM 6.2:**

https://user-images.githubusercontent.com/943378/208226983-7f2ef62b-28e0-49a6-9539-55b917c93bb8.mp4

`/var/log/synopkg.log` -- https://pastebin.com/hUPaeFQi
`/var/log/packages/sonarr.log` -- https://pastebin.com/sntH0f2g

**Video and logs from DSM 7.1:**

https://user-images.githubusercontent.com/943378/208227031-661fa900-203b-404e-8c4a-053395487fc2.mp4

`/var/log/synopkg.log` -- https://pastebin.com/Gv1BBQU8
`/var/log/packages/sonarr.log` -- https://pastebin.com/60MehRyv

As you can see, in DSM 6.2 the replace actually works (unlike in DSM 7.1) even though the code is the same. This will need some further investigation. Unfortunately in both situations, the copying of the configuration files to the new version didn't happen so that will need to be looked into more.

I found this interesting tidbit in the `/var/log/bash_err.log`:

```
2022-12-16T21:07:16-04:00 DSM62 sh: BASH_ERR: builtin_error [ (16250)"/bin/sh -e /proc/self/fd/9" -> (16252)"/usr/syno/sbin/synopkgctl stop nzbdrone" -> (16254)"/bin/sh /var/packages/nzbdrone/scripts/start-stop-status stop" ] /var/packages/nzbdrone/scripts/start-stop-status:123: kill: (13202) - No such process
```

This suggested to me that the removal of Sonarr v3 perhaps took place before the `prereplace` function did its job and I'd need to defer its removal maybe? I'll run some more tests with that concept.

mreid-tt:
@hgy59, with some further testing I was able to confirm that the logic I wanted to use for the `prereplace` function was not required since the `install_replace_packages` field should trigger DSM to uninstall the old package anyway. As such, I've only retained the logic to migrate the old configuration data to the new installation if present and I have tested this successfully under DSM6 and DSM7.

One challenge remains (mentioned above) regarding the uninstall of the old package under DSM7. I've tried a number of variations including adding the `use_deprecated_replace_mechanism` field but the uninstall of `nzbdrone` never triggers. As such, under DSM7, the user will have to manually stop 'Sonarr v3' and then start 'Sonarr v4' to get the new deployment.

This I believe may be a bug in DSM7 itself because I've scoured the internet and can find no discussions on this point. If you have any insight into fixing this I believe that should be the last thing required for this PR.

mreid-tt:
@hgy59, any thoughts on the `install_replace_packages` field not uninstalling the "Sonarr v3" package under DSM7?

mreid-tt:
@hgy59, I opened a ticket with Synology support for the `install_replace_packages` not working in DSM7 and they came back with a fix. Apparently, in the INFO file, the versioning of the previous version to be replaced was not correct. As such once we changed ...
https://github.com/SynoCommunity/spksrc/blob/a160c01946153910f6731a433e0bd9792cdb411d/spk/sonarr/Makefile#L38
... to ...
https://github.com/SynoCommunity/spksrc/blob/473fff1a8c72312f796e81b0e8cb6b42ad7d90cb/spk/sonarr/Makefile#L38
... this resolved the issue. I've tested it and I believe the package is ready. Once you are comfortable with the changes, I'll squash the commits for you to merge.

mreid-tt:
I think I've finished all the refinements. The last two included the following:
1. Add a check for the older `var` location (nzbdrone version 20210329-19 on DSM6)
2. Only set an update check if not performing a Sonarr v3 migration

mreid-tt:
A question I have is regarding the installation, there is a screen titled "Attention! DSM Permissions" which says:
> "Permissions for all download-related packages are managed with the group **'sc-download'** in DSM."

Given this is no longer true in DSM7 should we change or remove this message? Or is there a way to have the screen display different information if running DSM7 versus DSM6?

hgy59:
> EDIT: I looked at the UI components in Radarr and they seemed fairly up to date so I copied these and changed the app name.

👍 

I suppose for the upgrade wizard we should add information that the package replaces the v3 package (and an extra advice to create a backup first).

It might be necessary to have a different upgrade wizard page when updating from v3 vs updating from v4.

To implement this, we need a shell script for the upgrade_uifile to create different wizard pages - you find example scripts in spk/redis/src/upgrade_uifile.sh or in minio, mosquitto and tt-rss (tt-rss has even different pages for DSM 6 and DSM 7).

mreid-tt:
> It might be necessary to have a different upgrade wizard page when updating from v3 vs updating from v4.
> 
> To implement this, we need a shell script for the upgrade_uifile to create different wizard pages - you find example scripts in spk/redis/src/upgrade_uifile.sh or in minio, mosquitto and tt-rss (tt-rss has even different pages for DSM 6 and DSM 7).

I've attempted to implement this and it appears to work under DSM7 however under DSM6, the script does not seem to be executed. I attempted to find why this may be and I think it is because the script files in the generated package do not have the execute bits set compared with the other examples you suggested.

Looking at this package we have:
```
% tar -tvf sonarr_x64-6.1_20230108-1.spk
[---snip---]
drwxr-xr-x  0 root   root        0 Jan  7 20:58 WIZARD_UIFILES/
-rw-r--r--  0 root   root      579 Jan  7 20:58 WIZARD_UIFILES/upgrade_uifile_fre
-rw-r--r--  0 root   root     4455 Jan  7 20:58 WIZARD_UIFILES/install_uifile.sh
-rw-r--r--  0 root   root     4721 Jan  7 20:58 WIZARD_UIFILES/install_uifile_fre.sh
-rw-r--r--  0 root   root      564 Jan  7 20:58 WIZARD_UIFILES/upgrade_uifile
```

This is very different than two of the other examples which have the execute bits set:
```
% tar -tvf redis.v13.f15047\[apollolake-avoton-braswell-broadwell-broadwellnk-broadwellntbap-bromolow-cedarview-denverton-dockerx64-geminilake-grantley-purley-kvmx64-v1000-x86-x86_64\].spk
[---snip---]
drwxr-xr-x  0 root   root        0 Dec 18 18:14 WIZARD_UIFILES/
-rwxr-xr-x  0 root   root      958 Dec 18 18:40 WIZARD_UIFILES/upgrade_uifile.sh
-rw-r--r--  0 0      0        1149 Dec 18 18:40 syno_signature.asc
```
```
% tar -tvf tt-rss.v15.f15047\[noarch\].spk 
[---snip---]
drwxr-xr-x  0 root   root        0 Nov 15 15:05 WIZARD_UIFILES/
-rw-r--r--  0 root   root      795 Nov 15 15:52 WIZARD_UIFILES/uninstall_uifile
-rwxr-xr-x  0 root   root     3016 Nov 15 15:52 WIZARD_UIFILES/upgrade_uifile.sh
-rw-r--r--  0 root   root     2106 Nov 15 15:52 WIZARD_UIFILES/install_uifile
-rw-r--r--  0 root   root     2290 Nov 15 15:52 WIZARD_UIFILES/install_uifile_fre
-rwxr-xr-x  0 root   root     2874 Nov 15 15:52 WIZARD_UIFILES/upgrade_uifile_fre.sh
-rw-r--r--  0 root   root      859 Nov 15 15:52 WIZARD_UIFILES/uninstall_uifile_fre
-rw-r--r--  0 0      0        1145 Nov 15 15:53 syno_signature.asc
```

I've done some light digging around and this setting of the execution bits should happen automatically based on this framework code:
https://github.com/SynoCommunity/spksrc/blob/5efc80869342049d9102672bb820639ea0853363/mk/spksrc.spk.mk#L394-L397

@hgy59 interested to get your thoughts on the above analysis. Is the reason the `uifile` scripts are not running under DSM6 because of the file permissions? If so, how can I fix it? If not, I'm interested to get your thoughts on the problem.

mreid-tt:
@hgy59 so apparently my analysis was correct. the build checks that are currently performed do not seem to set the execution bits. When I look at the run from redis: update to v7.0.7 #5526 ([Atifacts](https://github.com/SynoCommunity/spksrc/actions/runs/3726769511)), the package for x64-6.1 does not have the execution bits set either:
```
% tar -tvf redis_x64-6.1_7.0.7-13.spk 
[---snip---]
drwxr-xr-x  0 root   root        0 Dec 18 18:05 WIZARD_UIFILES/
-rw-r--r--  0 root   root      958 Dec 18 18:05 WIZARD_UIFILES/upgrade_uifile.sh
```
So I tried setting the bits manually with the following:
```
$ sudo -i
# mkdir temp
# tar xvf sonarr_x64-6.1_20230108-1.spk -C temp
[---snip---]
# chmod a+x temp/WIZARD_UIFILES/*.sh
# tar cf temp.spk -C temp $(ls temp/)
# exit
```
This resulted in an archive with the correct bits:
```
% tar -tvf temp.spk
[---snip---]
drwxr-xr-x  0 root   root        0 Jan  7 20:58 WIZARD_UIFILES/
-rw-r--r--  0 root   root      579 Jan  7 20:58 WIZARD_UIFILES/upgrade_uifile_fre
-rwxr-xr-x  0 root   root     4455 Jan  7 20:58 WIZARD_UIFILES/install_uifile.sh
-rwxr-xr-x  0 root   root     4721 Jan  7 20:58 WIZARD_UIFILES/install_uifile_fre.sh
-rw-r--r--  0 root   root      564 Jan  7 20:58 WIZARD_UIFILES/upgrade_uifile
```
With this I was able to successfully test the `uifile` scrips in DSM6. Assuming the actual published package will have the correct execution bits set, everything looks good to go following your final review.

mreid-tt:
I"m going to squash things down to get ready to publish. I also found a bug in my testing with upgrades using Sonarr v3 and I made a PR to address it.

cross/sonarr/Makefile
```diff
@@ -0,0 +1,35 @@
+PKG_NAME = Sonarr
+PKG_VERS = 4.0.0.344
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME).develop.$(PKG_VERS).linux-$(PKG_DIST_ARCH).$(PKG_EXT)
+PKG_DIST_SITE = https://download.sonarr.tv/v4/develop/$(PKG_VERS)
+PKG_DIR = $(PKG_NAME)
+
+DEPENDS =
+
+UNSUPPORTED_ARCHS = $(DOTNET_UNSUPPORTED_ARCHS)
```
lajuffermans, line 10: Also, is this correct? Mono started to annoy me so I tried building v4 myself. Initially it gave an "unsupported arch" when I tried to compile it for armadaxp(armv7) DSM 7.1.

When I removed this it compiled fine and it also runs fine. So is the `DOTNET_UNSUPPORTED_ARCHS` still up to date?

spk/sonarr/src/wizard/install_uifile.sh
```diff
@@ -0,0 +1,107 @@
...
+EOF
+}
+
+function upgrade_with_config
+{
+    cat <<EOF >"${SYNOPKG_TEMP_LOGFILE}"
+[{
+	"step_title": "Download a backup before upgrading",
+	"items": [{
+		"desc": "<strong style='color:red'>IMPORTANT:</strong> This update is an upgrade to Sonarr v3. We suggest that you download a manual backup as a precautionary measure before installing this update.<br><br>You can download a backup via Sonarr's built-in web-interface.<br>To do this, navigate to <b>System>Backup</b> in the Sonarr UI.<br><br>Create a backup and then download a copy of that backup before installing this upgrade."
```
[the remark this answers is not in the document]
mreid-tt: Good catch! The language has been revised for all three upgrade scenarios.

cross/sonarr/Makefile
mreid-tt: Yes, the `armadaxp` is not currently included as per:

https://github.com/SynoCommunity/spksrc/blob/175b8e87c5a170ceae5229994851ae4721511579/mk/spksrc.archs.mk#L53-L54

lajuffermans: Just FYI, Sonarr v4 does actually work on my DS414 (armadaxp). Sounds to me some applications do work and some don't.  😉

mreid-tt: Oh! I misinterpreted what you meant, thanks for clarifying. Perhaps you can take a look at the source issues https://github.com/SynoCommunity/spksrc/issues/4790 and https://github.com/SynoCommunity/spksrc/pull/4793. I don't have that model of NAS to do this myself but perhaps you can review the specific use cases and maybe re-test them to give a current position on this?

Looking across the codebase, this exclusion is currently made for jellyfin, lidarr, ombi, prowlarr as well as the dotnet-runtime. Ideally if you have something like Virtual DSM installed you can do some test deployments and gather current evidence to indicate the actual compatibility issues. With this evidence, a PR to include `armadaxp` in relevant .Net builds can be made (I'll help if you want).

mreid-tt: hey @lajuffermans, I've created a new issue for the testing of ARMv7 archs with ***arr** dotNET packages. I'd really appreciate your feedback to the issue to help with testing: https://github.com/SynoCommunity/spksrc/issues/5574

mreid-tt:
@hgy59, perhaps we can take another look at this one? I believe we should be able to merge this.

tupac55:
Yeah V4 of sonarr come with better language implantation so that would be very great if we can have a V4 update of the package.. docker is so much a pain to configure that Synocommunity package is a god send 

Thanks

mreid-tt:
@hgy59, is this good to merge?

bonebox:
Thanks for all your work on this @mreid-tt. How do we get this merged in?

publicarray:
Afaik the data migration only worked for DSM7. Did you test this on dsm6?

publicarray:
Otherwise well done for moving the package to the correct name 👍

mreid-tt:
> Afaik the data migration/package replacement only worked for DSM7. Did you test this on dsm6?

Yes, I tested this under DSM 6 and DSM 7 and both were eventually successful. There was an initial issue with package replacement under DSM 7 but this was because of the version specificity in the `INFO` file which was corrected. There was a second issue with wizard scripts not running under DSM 6 due to issues with file permissions and this was fixed with #5608.

The main difference with DSM 6 is that because it does a clean uninstall, the previous configuration files are removed on migration (no option to downgrade). In DSM 7 the older configuration files are left in the `/var/packages/nzbdrone/var` folder which allows you to revert to Sonarr v3 if you wanted to. What do you think about this?

spk/sonarr/Makefile
```diff
@@ -0,0 +1,57 @@
+SPK_NAME = sonarr
+SPK_VERS = $(shell date +%Y%m%d)
```
smaarn, line 2: Wouldn't it be better to freeze this ?

smaarn (commented):
Regarding the i18n versions of the installers wouldn't be worth using a template mechanism with i18n variables being injected ? (To avoid basically having to maintain separate wizard files which essentially do the same thing except for the text which are differing)

spk/sonarr/Makefile
mreid-tt: If I recall from previous comments from @hgy59, the dates are used with the `*arr` and other apps which have internal update mechanisms. In these apps, the internal version of the app can and does change, whereas fixed versions are used for apps where the package is tied to that version (like with `SABnzbd`).

mreid-tt:
> Regarding the i18n versions of the installers wouldn't be worth using a template mechanism with i18n variables being injected ? (To avoid basically having to maintain separate wizard files which essentially do the same thing except for the text which are differing)

I've not used i18n before but looking into it this seems to be a DSM 7 feature (doesn't seem to be present in the DSM 6 developer guide) and related to things like help files and config. I don't see any reference to be able to use this with wizard files. Do you have a specific example of where this could be used?

smaarn:
> I've not used i18n before but looking into it this seems to be a DSM 7 feature (doesn't seem to be present in the DSM 6 developer guide) and related to things like help files and config. I don't see any reference to be able to use this with wizard files. Do you have a specific example of where this could be used?

Actually I was more thinking about using a mechanism such as the one done for `rutorrent` (which could be extracted if deemed easier to maintain) which can be described in a nutshell as building wizard files by merging the following files:

1. A file specifying i18n variables
2. A base shell file which generates wizards using the variable defined in 1.

TBH I didn't know there was something about it in DSM 7 :D

mreid-tt:
> Actually I was more thinking about using a mechanism such as the one done for `rutorrent` (which could be extracted if deemed easier to maintain) which can be described in a nutshell as building wizard files by merging the following files:
> 
> 1. A file specifying i18n variables
> 2. A base shell file which generates wizards using the variable defined in 1.

An interesting proposal but I don't think this will be easier to maintain. Looking at the developer guide it is expecting the wizard UI files of the form `“install_uifile_[DSM language].sh”`. If this is the case you will still need to have individual files for each language required. Yes, this can be a shell that points to a base file which then imports the specific internationalised variables for the selected language (perhaps as a parameter) but that is a but too much abstraction in my view.

spk/sonarr/Makefile
```diff
@@ -0,0 +1,57 @@
+SPK_NAME = sonarr
```
th0ma7, line 1: So if I get this right (I ain't a sonarr user), the v3 is actually called nzbdrone while v4 will be called sonarr?

spk/sonarr/Makefile
```diff
@@ -0,0 +1,57 @@
...
+
+# use alternate TMPDIR as /tmp might be too small and not accessible on DSM >= 7.1.
+# TMPDIR is used for sonarr built in backup and update
+USE_ALTERNATE_TMPDIR = 1
+
+# replace previous legacy package
+INSTALL_REPLACE_PACKAGES = "nzbdrone>=20210329-19"
+USE_DEPRECATED_REPLACE_MECHANISM = yes
+# disable checking for port conflicts
+CHECKPORT = no
```
th0ma7, line 45: Shouldn't there also be a BETA flag somewhere in there?

spk/sonarr/Makefile
```diff
@@ -0,0 +1,57 @@
...
+WIZARDS_DIR = src/wizard/
+
+POST_STRIP_TARGET = sonarr_extra_install
+
+# use alternate TMPDIR as /tmp might be too small and not accessible on DSM >= 7.1.
+# TMPDIR is used for sonarr built in backup and update
+USE_ALTERNATE_TMPDIR = 1
+
+# replace previous legacy package
+INSTALL_REPLACE_PACKAGES = "nzbdrone>=20210329-19"
```
th0ma7, line 42: I guess this answers my question pertaining to the naming.

spk/sonarr/src/service-setup.sh
```diff
@@ -0,0 +1,88 @@
...
+CMD_ARGS="-nobrowser -data=${SONARR_CONFIG_DIR}"
+
+LEGACY_SPK_NAME="nzbdrone"
+LEGACY_SYNOPKG_PKGDEST="/var/packages/${LEGACY_SPK_NAME}/target"
+# check for legacy package data storage location
+if [ -d /var/packages/${LEGACY_SPK_NAME}/var ]; then
+    LEGACY_SYNOPKG_PKGVAR="/var/packages/${LEGACY_SPK_NAME}/var"
+else
+    LEGACY_SYNOPKG_PKGVAR="${LEGACY_SYNOPKG_PKGDEST}/var"
+fi
```
th0ma7, line 19: You could use `SYNOPKG_DSM_VERSION_MAJOR` instead to detect if being on DSM7 or later, or not.  This would satisfy your check.  Also I'd recommend doing a test -d on both cases  else either exit 1 or considering it isn't installed.

I was looking at https://global.download.synology.com/download/Document/Software/DeveloperGuide/Os/DSM/All/enu/DSM_Developer_Guide_7_enu.pdf and wondered if you could check if a package is installed or not but couldnt see any of interest in scripting variables at page 56.

Further testing `synopkg status` does work, not entirely as expected but does work:
```
$ synopkg status ffmpeg
{"aspect":{"active":{"status":"stop","status_code":263,"status_description":"failed to get unit status"}},"description":"Status: [263], package is stopped","package":"ffmpeg","status":"stop"}
th0ma7@th0ma7-nas:~$ echo $?
7
th0ma7@th0ma7-nas:~$ synopkg status toto
{"aspect":{"active":{"status":"stop","status_code":263,"status_description":"failed to get unit status"},"error":{"status":"non_installed","status_code":255,"status_description":"failed to locate given package"}},"description":"Status: [255], No such package","package":"toto","status":"non_installed"}
th0ma7@th0ma7-nas:~$ echo $?
255
```

spk/sonarr/src/wizard/install_uifile.sh
```diff
@@ -0,0 +1,115 @@
...
+
+function incompatible_upgrade
+{
+    cat <<EOF >"${SYNOPKG_TEMP_LOGFILE}"
+[{
+	"step_title": "ATTENTION! Incompatible upgrade detected",
+	"invalid_next_disabled_v2": true,
+	"items": [{
+		"type": "textfield",
+		"desc": "<strong style='color:red'>WARNING:</strong> This update is an upgrade to Sonarr v4 from Sonarr v3. We have detected that you have Sonarr v2 installed. We suggest that you upgrade your previous installation to Sonarr v3 before installing this upgrade.<br><br><b>Installation will abort.</b>",
```
th0ma7, line 11: Couldn't you simply tell something like, indeed incompatible as such if you continue your local legacy configs won't be ported over to the new installation and old install be deleted?  Just a thought...

th0ma7 (dismissed):
Overall looking quite good.  A few proposed changes proposed below.

spk/sonarr/Makefile
mreid-tt: Yes, that is correct

spk/sonarr/src/wizard/install_uifile.sh
mreid-tt: I would prefer that users take the decision to uninstall the old version on their own rather than doing it for them and losing their library data.

spk/sonarr/src/service-setup.sh
mreid-tt: I'm actually only interested in locating previous configuration files to migrate them. That is unless they are from Sonarr v2. I don't believe that checking the current DSM version will assist with that.

Regarding the recommendation to test with `-d` on the folders... I would normally agree with this however the tests later on for the specific files in `LEGACY_SYNOPKG_PKGDEST` and `LEGACY_CONFIG_DIR` are the substantive tests required. I wouldn't want to potentially have a null value for `LEGACY_SYNOPKG_PKGVAR` at this point and potentially invalidate those tests.

For the check for the presence of the older package, this would be ideal but given that it is a different package name the more direct method would be to check for the presence of the executable. This is done in the `validate_preinst` and is only done to exclude the upgrade from Sonarr v2.

cross/sonarr/Makefile
```diff
@@ -0,0 +1,33 @@
...
+# use digests with multiple files
+PKG_DIST_ARCH_LIST = x64 arm arm64
+
+include ../../mk/spksrc.archs.mk
+PKG_DIST_ARCH = x64
+ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS)),$(ARCH))
+PKG_DIST_ARCH = arm
+else ifeq ($(findstring $(ARCH),$(ARMv8_ARCHS)),$(ARCH))
+PKG_DIST_ARCH = arm64
+endif
```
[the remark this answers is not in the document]
mreid-tt: I've included an explicit check for x64 which should address this concern.

mk/spksrc.spk.mk
```diff
@@ -211,6 +211,15 @@ endif
...
 endif
+ifneq ($(strip $(INSTALL_REPLACE_PACKAGES)),)
+	@echo install_replace_packages=\"$(INSTALL_REPLACE_PACKAGES)\" >> $@
+endif
+ifneq ($(strip $(USE_DEPRECATED_REPLACE_MECHANISM)),)
+	@echo use_deprecated_replace_mechanism=\"$(USE_DEPRECATED_REPLACE_MECHANISM)\" >> $@
+endif
+ifneq ($(strip $(CHECKPORT)),)
+	@echo checkport=\"$(CHECKPORT)\" >> $@
+endif
```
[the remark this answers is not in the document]
mreid-tt: Agreed, the Wiki will be updated to articulate these commands. Essentially they are adding additional `INFO` commands from the developer guide into our framework.

spk/sonarr/Makefile
mreid-tt: I've added the beta flag as proposed.

hgy59: IMHO the BETA flag is to mark spk packages as beta and to the included package.

There was a discussion somewhere, and if I remember well, sonarr v4 will be beta for several years from now, but everyone that uses v4 will be aware of.

So if this package is proved to work, I would not mark it as beta.

spk/sonarr/Makefile
mreid-tt: I suppose I could set the date to the source file (which would be 20230306) but then we would need to maintain data on the source file both in `cross/sonarr` and `spk/sonarr`. This seems like a maintenance nightmare to me.

@hgy59, any thoughts on how we should proceed here? Are there any variables we can use to pull the date from the downloaded source file?

spk/sonarr/Makefile
mreid-tt: Thanks for the reminder. Yes, other packages like Lidarr and Prowlarr had operated as pre-release versions for some time without this flag so I'll remove it once the other comments are in.

spk/sonarr/Makefile
hgy59: > Nope, this won't work. There is then no way we can reproduce this spk afterwards. This date should be statically set and probably match the creation date of version 4.0.0.394 or corresponding to the md5sum of the commit associated to it.

@th0ma7 sorry to disagree. Packages that use app internal updaters have been published this way. The package version is taken from the date at build time (plus the usual REV).
- we do not use the real version number of the included app, since it is not applicable when the already installed package has updated to a newer version
- we need to create updates for such packages only, when the (generic) installer code or a dependen library in the package needs an update (for example: when new archs are available, or the dependen python changes from 310 to 311, ...)
- we could use the "release date" of the bundled app, but this would not change the fact that dependent modules where updated inbetween (for example python 3.10 had a minor version update)

IMHO we should keep this convention and use `SPK_VERS = $(shell date +%Y%m%d)` for packages with an internal updater.
cc @publicarray, @ymartin59 


BTW it is a good practice to add the version of the bundled app to the changelog of such packages. If there is no explicit version, this could be a git sha plus commit date.

mk/spksrc.spk.mk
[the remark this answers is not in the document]
mreid-tt: The Wiki has been updated with the three new variables. I've also sorted the sections of the [page](https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables) to make it easier to find things.

mreid-tt:
> Overall looking quite good. A few proposed changes proposed below.

Hey @th0ma7, any responses to the feedback on your requested changes?

AdamEldred:
I did compile and install this over a basic Sonarr v3 setup on DSM7 on a Realtek chip - no issues to report. Sonarr seems to self upgrade just fine after install as well. Nice work!

mreid-tt:
Hey @th0ma7, my review of the suggested changes show that the important ones were addressed (including updating the Wiki) and I've gone ahead an merged the PR. Do let me know if you can also assist with publishing.

th0ma7:
Overall looked good, nice work.
Publishing in progress: https://github.com/th0ma7/spksrc/actions/runs/4602153928

bonebox:
Worked great on a DS920+. Thank you!

grerppy:
Will there be an armada38x (DS216j) version for DSM6? Thanks

mreid-tt:
> Will there be an armada38x (DS216j) version for DSM6? Thanks

Based on the testing of dotnet *arr packages with ARMv7 archs we did (see #5574), the `armada38x` platform is only supported with DSM 7.1. DSM 6.x does not have the required libraries to run dotnet 6.x.

tupac55:
Thanks for the update everything went fine updating on my ds220+

PR #5527, opened 2022-12-18
Update SABnzbd to 3.7.1 and Python 3.11

Safihre:
## Description

Update SABnzbd to 3.7.1.

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5531, opened 2022-12-20
py310-py311: Update to latest wheel and rework numpy management

th0ma7:
## Description

Update to latest wheels and fully test all of them for both py3.10-3.11

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


spk/python310/src/requirements-crossenv.txt
```diff
@@ -8,98 +8,84 @@
...
 # Require --global-options arguments
 #  WHEELS_BUILD_ARGS = [Pillow] build_ext --disable-platform-guessing
-# Pillow==9.2.0
+Pillow==9.3.0
 
 # [psutil]
 # Last known version to build on armv5/ppc
-# psutil==5.9.1
+psutil==5.9.1
+#psutil==5.9.4
```
hgy59, line 56: psutil==5.9.4 builds for qoriq (ppc), version 5.9.2 had a bug that was fixed with https://github.com/giampaolo/psutil/pull/2156

th0ma7: ah! thnx, i was looking for exactly that!

hgy59:
@th0ma7 I am currently updating #5399 with actual python310 and get this warning while cross compiling wheels:
```
DEPRECATION: --no-binary currently disables reading from the cache of locally built wheels. In the future --no-binary will not influence the wheel cache. pip 23.1 will enforce this behaviour change. A possible replacement is to use the --no-cache-dir option. You can use the flag --use-feature=no-binary-enable-wheel-cache to test the upcoming behaviour. Discussion can be found at https://github.com/pypa/pip/issues/11453
```

This is just a hint, as I didn't analyze what you are working on here.

th0ma7:
> This is just a hint, as I didn't analyze what you are working on here.
> ```
> WARNING: Implying --no-binary=:all: due to the presence of --build-option / --global-option / --install-option. Consider using --config-settings for more flexibility.
> DEPRECATION: --no-binary currently disables reading from the cache of locally built wheels. In the future --no-binary will not influence the wheel cache. pip 23.1 will enforce this behaviour change. A possible replacement is to use the --no-cache-dir option. You can use the flag --use-feature=no-binary-enable-wheel-cache to test the upcoming behaviour. Discussion can be found at https://github.com/pypa/pip/issues/11453
> ```

@hgy59 thnx for the hint.  I've took a few cycles to analyse the issue and really there isn't yet any.  What is really frustrating is the complete lack of documentation of both `--config-settings` and `--use-feature=no-binary-enable-wheel-cache`

As such, these two non-issues will be handled at some other time, perhaps at a subsequent `pip` update where we won't have the choice to adapt.... or if I find time to ask developpers how to use them as replacement.

th0ma7:
@Safihre would you mind testing out py311 package resulting from the last commit?  (currently building)

th0ma7:
Both py310-py311 where successfully tested on DSM6 and DSM7.  Merging.

spk/python310/src/requirements-abi3.txt
```diff
@@ -8,5 +8,5 @@
 # Require:
 #    - CFLAGS=-sdt=c99 <  gcc-4.9
 #    - CFLAGS=-sdt=c11 >= gcc-4.9
```
hgy59, line 10: typo:  std (not sdt)

th0ma7: Good catch... Gladly its ok in the makefile :)

hgy59:
@th0ma7 shouldn't the wheels installed into crossenv be of the same version as in requirements-pure.txt files?
i.e. versions in
cross/python310/Makefile vs spk/python310/src/requirements-pure.txt
cross/python311/Makefile vs spk/python311/src/requirements-pure.txt

and setuptools should match the version in native/python310/Makefile and native/python311/Makefile

_so far only setuptools differ_
- setuptools (setuptools==65.6.3 vs setuptools==63.4.3)
- ...


When building crossenv wheels I faced _uninstall setuptools_ logs for every wheel...

th0ma7:
@hgy59 sharp eyes... indeed normally all versions are identical, from `native` to `cross` used to generate the `work-<arch>-version>/crossenv`.  Although `requirements-*` (e.g. resulting wheels being built) is less of an issue but still I do my best at keeping them identical.

Issue relates to numpy building from the `crossenv` (which in turns is recommended to have an identical version for `native`) for versions below <= 1.22.x, which affects only `armv5` and `armv7l`:
https://github.com/pypa/setuptools/issues/3549

With the newer py311 we can no longer support version 1.21.x for `armv5`.  Question then becomes, should we bump `setuptools` only for py311 and abandon numpy on older compilers (e.g. gcc <= 4.8 thus also affecting `armv7l`).

Lastly, I haven't tested with latest and greates `seutptools` along with latest updates available for older numpy branches.  Perhaps things have evolved slightly in terms of compatibility?

Safihre:
@th0ma7 It seems to work fine for #5527

PR #5543, opened 2022-12-29
curl: update to v7.87 and add optional features

hgy59:
## Description

update curl to v7.87.0

avoid blowing up cross/curl with additional features (introduced with #5495 for tvheadend)

- add BUILD_CURL_WITH variable to build optional features (libssh2, gnutls, mbedtls)
- avoid the optional features for all dependent packages as those are currently used with tvheadend only
- update tvheadend to build curl with features as before
- add `diyspk/curl/Makefile` to build curl with all features


th0ma7:
This is a neat idea.  If you need i can review it over the weekend.

cross/tvheadend/Makefile
```diff
@@ -11,7 +11,12 @@ PKG_DIR = $(PKG_NAME)-$(PKG_GIT_HASH)
 DEPENDS  = cross/openssl cross/gnutls
 DEPENDS += cross/uriparser cross/pcre2 cross/libiconv cross/fontconfig
 DEPENDS += cross/libhdhomerun cross/libdvbcsa cross/dvb-apps cross/pngquant
+DEPENDS += cross/libwebsockets
```
hgy59, line 14: @th0ma7 I have moved the `cross/libwebsockets` dependency from curl to tvh, assuming tvh uses libwebsockets directly.

`cross/libwebsockets` does not add any feature to curl. Does tvh need webservice support in curl/libcurl ?
If so, cross/libwebsockets can be removed at all, and tvh can build curl with `--enable-websockets` by adding `websockets` to `BUILD_CURL_WITH` variable.
But this is an experimental feature of curl.

th0ma7: I took the time to dig a little on this and I can't find anything in the code pertaining to libwebsockets library usage.  TVH does however create websocket but seems to be doing so on its own.  As such I don't believe libwesockets is relevant anymore.

cross/tvheadend/Makefile
```diff
@@ -11,7 +11,12 @@ PKG_DIR = $(PKG_NAME)-$(PKG_GIT_HASH)
...
+# To build curl with zstd feature define depends zstd before cross/curl
+DEPENDS += cross/zstd
 DEPENDS += cross/curl
+# curl build needs additional configuration for gnutls and libssh2
+export BUILD_CURL_WITH="gnutls libssh2"
```
th0ma7, line 19: While curl only has a `--with-zstd=PATH` and `--without-zstd`, this could also be added as a `BUILD_CURL_WITH` flag.

th0ma7 (approved):
LGTM.  Your proposal for adding functionality to curl is really neat.

Two proposed changes:
1. Removing websocket entirly from TVH
2. Moving zstd as BUILD_CURL_WITH option

hgy59:
@th0ma7 this is now ready to merge

_I did not succeed to add the optional protocol RTMP as librtmp from rtmpdump is not prepared for cross compilation_

PR #5118, opened 2022-02-03
syncthing: fix SERVICE_CERT_RELOAD

hgy59:
## Description

follow up of #5117
- revert to use SERVICE_CERT_RELOAD

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5550, opened 2023-01-04
restic rest-server 0.11.0 initial package

fgma:
## Description

New package for restic's rest-server:

https://restic.net/
https://github.com/restic/rest-server

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update

TODO:
- [ ] fix package update
- [ ] include certificate handling according to https://github.com/SynoCommunity/spksrc/wiki/Service-Support (SERVICE_CERT, SERVICE_CERT_RELOAD)
- [ ] add port to wiki (https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports)


fgma:
I will add the port to the wiki when I got some feedback if this PR is getting accepted and the port is OK to be used. 

I don't use the certificate handling myself but I guess this would be a nice feature for others. Would the syncthing implementation be a good starting point or is this not yet stable enough yet?

cross/rest-server/Makefile
```diff
@@ -0,0 +1,24 @@
+PKG_NAME = rest-server
+PKG_VERS = 0.11.0
+PKG_COMMIT_ID = 4741eec619a2b61d251ffbd0acce028620067579
```
hgy59, line 3: AFAICS this variable is not used
```suggestion
```

cross/rest-server/Makefile
```diff
@@ -0,0 +1,24 @@
...
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://github.com//restic/rest-server/releases/download/v$(PKG_VERS)
+PKG_DIST_FILE = $(PKG_DIST_NAME)
```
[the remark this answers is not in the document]
fgma: Fixed in https://github.com/SynoCommunity/spksrc/pull/5550/commits/f85d69a462510d7a7d1ca70d1ae1e01209f2d426

cross/rest-server/Makefile
fgma: Fixed in https://github.com/SynoCommunity/spksrc/pull/5550/commits/f85d69a462510d7a7d1ca70d1ae1e01209f2d426

cross/rest-server/Makefile
```diff
@@ -0,0 +1,24 @@
+PKG_NAME = rest-server
```
hgy59, line 1: Please name the package restic-rest-server (and rename the folder)

To have a variable for the downloaded file as rest-server, I propose to use an additional variable like PKG_REAL_NAME = rest-server. And use this new variable instead.

the name for the spk package should be restic-rest-server or restic (rest-server is just too common).

spk/rest-server/src/service-setup.sh
```diff
@@ -0,0 +1,50 @@
...
+# Reload wizard variables stored by postinst
+if [ -r "${INST_VARIABLES}" ]; then
+    # we cannot source the file to reload the variables, when values have special characters like <, >, ...
+    for _line in $(cat "${INST_VARIABLES}"); do
+        _key="$(echo ${_line} | awk -F'=' '{print $1}')"
+        _value="$(echo ${_line} | awk -F'=' '{print $2}')"
+        declare "${_key}=${_value}"
+    done
+fi
+
```
hgy59, line 17: we have the function `reload_inst_variables` availabel here (from mk/spksrc.servier.installer.functions)
so you should use this function

```suggestion
# Source installer variables and functions to be available for service start (and not only during installation)
INST_FUNCTIONS=$(dirname $0)"/functions"
if [ -r "${INST_FUNCTIONS}" ]; then
    . "${INST_FUNCTIONS}"
fi

reload_inst_variables
```

_but it is not available in upgrade_uifile.sh, there you have to keep it_

hgy59: It is probably not a god idea, to compose the SERVICE_COMMAND from wizard_variables. 

what about writing the ARGs into an options file at installation time and use this file to add additional arguments to the SERVICE_COMMAND?


There is a similar approach in spk/syncthing/src/service-setup.sh to configure additional arguments by the user and not by wizard varialbes.

Then the SERVICE_COMMAND must only be adjusted in service_prestart.

cross/rest-server/Makefile
fgma: Fixed in https://github.com/SynoCommunity/spksrc/pull/5550/commits/d176e86f04b4daaefd98357099096a53de51a68b

spk/rest-server/src/service-setup.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+if [ "${WIZARD_PROMETHEUS}" == "true" ]; then
+    ARGS="${ARGS} --prometheus"
+
+    if [ "${WIZARD_PROMETHEUS_NO_AUTH}" == "true" ]; then
+        ARGS="${ARGS} --prometheus-no-auth"
+    fi
+fi
+
+PORT=8500
```
hgy59: You can ommit to need the port definition here.

Add `SERVICE_PORT = 8500` to the Makefile and you will have the variable `SERVICE_PORT` with this port in service-setup.sh.

I suppose the there is no web UI available but the REST API only?
So you need to define  `NO_SERVICE_SHORTCUT` to avoid a link in DSM Ui, 
and you must remove `ADMIN_PORT = 8500` in the Makefile (that would create a link in the DSM Package Center to the non existing web page)

spk/rest-server/src/service-setup.sh
fgma: Your example code doesn't work for me. All wizard variables are empty. Did you test it?

spk/rest-server/src/service-setup.sh
fgma: Fixed in https://github.com/SynoCommunity/spksrc/pull/5550/commits/d8ea928913b6679d8e72201c7b8e9f61c764743d

fgma:
I've switched to multiselect. It looks a bit nicer and the internal handling is a bit easier.

spk/rest-server/src/service-setup.sh
fgma: I've found the problem in spksrc.service.installer.functions -> reload_inst_variables ()
```
reload_inst_variables ()
{
    # Reload wizard variables stored by postinst
        # we cannot source the file to reload the variables, when values have special characters like <, >, ...
        for _line in $(cat "${INST_VARIABLES}"); do
            _key="$(echo ${_line} | awk -F'=' '{print $1}')"
            _value="$(echo ${_line} | awk -F'=' '{print $2}')"
            declare "${_key}=${_value}"
        done
    fi
}
```

Declare needs the -g flag to make declared variables available outside the function. Should I change this or will this break something else?

hgy59: yes please, I have this issue elsewhere in the validate function...

fgma: Fixed in https://github.com/SynoCommunity/spksrc/pull/5550/commits/65187ff24eb49c68aab9e62f1a964d023c7b320b. I don't see any benefits in adding another config file for ARGS. The way it is everything is in one place using existing code. Adding another file just adds complexity.

fgma:
Updated to current master. Anything else left for me to do here?

fetzerch:
Thanks much for packaging this! I have been using *restic-rest-server* already a couple of months on a Synology box with a manual deployment. I now tried the package built in CI and it seems to work fine. Really looking forward to get this PR integrated.

I had to tweak two things manually to better support my usecase. Wanted to bring them up for discussion to see if it makes sense to include them as UI settings.

1. I'm backing up to an external USB drive, but I could not select `volumeUSB1` in the wizard. I was able to change that in `/usr/syno/etc/packages/restic-rest-server/installer-variables`.
2. I'm using https with a self signed certificate (`public_key` and `private_key` files are located in the backup repository root) and therefore I had to modify `/var/packages/restic-rest-server/scripts/service-setup` to start the service with `--tls`. It would be nice if there was an extra field in the wizard to add 'custom' options.

hgy59:
> 2\. I'm using https with a self signed certificate (`public_key` and `private_key` files are located in the backup repository root) and therefore I had to modify `/var/packages/restic-rest-server/scripts/service-setup` to start the service with `--tls`. It would be nice if there was an extra field in the wizard to add 'custom' options.

This triggers the discussion above to add a file for custom service arguments.

hgy59:
> Updated to current master. Anything else left for me to do here?

see  #5649 for shared folder handling.

probably this will solve the use case to use an usb drive for backups (when creating a shared folder on the usb drive before the installation of restic rest-server and chosing this share in the wizard).

fgma:
I'll add support for additional custom arguments. 

Concerning the certificates: Shouldn't we use the SERVICE_CERT/SERVICE_CERT_RELOAD for this use case?

hgy59:
> Concerning the certificates: Shouldn't we use SERVICE_CERT/SERVICE_CERT_RELOAD for this use case

Does this work for DSM 6 only? #5117 and #5118

fgma:
I have no idea. I don't use it. I've just read about it in the documentation and thought this the proper way to handle certificates.

fgma:
Custom arguments added. What about certificate handling? Is there an official way to do so?

fetzerch:
Any chance you could retrigger a build, I would like to try it out using the packages (but they're gone).

hgy59:
@fgma can you repase this PR?
It is not possible to trigger the gh-build-action again.

spk/restic-rest-server/Makefile
```diff
@@ -0,0 +1,30 @@
+SPK_NAME = restic-rest-server
+SPK_VERS = 0.12.1
+SPK_REV = 1
+SPK_ICON = src/logo.png
+DSM_UI_DIR = app
+
```
hgy59, line 6: A `DSM_UI_DIR` is not expected when `NO_SERVICE_SHORTCUT = yes` is configured

The `DSM_UI_DIR` is to provide a config file for short cuts in DSM UI and the required icons.
```suggestion

```

fgma: fixed

mreid-tt:
Hey @fgma, I noticed this related package was proposed some time ago but never quite made it over the finish line. If this is still something you’re interested in getting published, I’d be happy to work with you on it.

PR #5553, opened 2023-01-06
Bazarr: upgrade 1.1.2 => 1.1.4

smaarn:
## Description

This is upgrading bazarr 1.1.2 to 1.1.4.

## Checklist

- ~~[ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created~~

### Type of change

- [x] Package update


spk/bazarr/Makefile
```diff
@@ -29,6 +31,8 @@ SERVICE_SETUP = src/service-setup.sh
 SERVICE_PORT  = 6767
 SERVICE_PORT_TITLE = $(DISPLAY_NAME)
 
+SYSTEM_GROUP = sc-download
```
mreid-tt, line 34: I'm not sure this is valid. I think this would be used for the service user privilege file. Looking at the comments in the code, this seems to be limited to `http` and `system`...

https://github.com/SynoCommunity/spksrc/blob/175b8e87c5a170ceae5229994851ae4721511579/mk/spksrc.service.mk#L317-L318

I suggest that this line be removed. Also, under DSM7, `sc-download` is no longer used anyway. Can you elaborate on the problem you are trying to resolve with the group?

smaarn: For what it's worth I tested it and it did the trick.

The idea here was to be inline with DSM 6 conventions following a user comment on Discord. But following your comment I realize indeed that it's actually not relevant. 

Will revert.

smaarn: Reverted.

mreid-tt: Nice, perhaps you can update the description on the PR to reflect the current changes. Given this was building successfully before you can do another push to trigger a new build since there have been some changes recently to the repository which may have fixed the build process.

mreid-tt (approved):

Ketoth:
How can I update bazarr? can I do anything to help?

hgy59:
@smaarn just found some PRs of you that are ready to publish

PR #5554, opened 2023-01-08
Sonarr v3: fix block on legacy (v2) upgrades

mreid-tt:
## Description

This corrects a bug in Sonarr v3 (#5515) where an upgrade from legacy Sonarr v2 was not blocked (in DSM6). This results in the distribution containing two sets of binaries and AppData directories which can break the deployment. This bug also creates issues with upgrades to Sonarr v4 (#5524) since the correct installed version cannot be determined.

Once published, it is recommended that Sonarr v3 (Version 20221216-21) be unpublished to avoid future use.

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/spk/sonarr3/Makefile
+++ b/spk/sonarr3/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = nzbdrone
 SPK_VERS = $(shell date +%Y%m%d)
-SPK_REV = 21
+SPK_REV = 22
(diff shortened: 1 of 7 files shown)


mreid-tt:
@hgy59, when this bug fix is published please un-publish Sonarr v3 (Version 20221216-21) to avoid future use.

Issue #5549, opened 2023-01-03
minio update strategy for new data format


fgma:
I just tried to update the minio package to the current release.  Current versions of minio have changed the way data is stored. Old data will not be accepted when starting as seen in my log an also according to https://min.io/docs/minio/linux/operations/install-deploy-manage/deploy-minio-single-node-single-drive.html. 

The official migration strategy is described here: https://min.io/docs/minio/linux/operations/install-deploy-manage/migrate-fs-gateway.html

According to this documentation you need to setup a completely new instance with the new backend and transfer your data/users/permissions to the new instance manually using mc.

Currently I have no idea how to handle this for a regular user.

```
Tue Jan  3 23:54:04 CET 2023
Starting minio command /volume1/@appstore/minio/bin/minio server --quiet --console-address :9001 --anonymous /volume1/minio
ERROR Unable to use the drive /volume1/minio: Drive /volume1/minio: found backend type fs, expected xl or xl-single - to migrate to a supported backend visit https://min.io/docs/minio/linux/operations/install-deploy-manage/migrate-fs-gateway.html: Invalid arguments specified
```

**EDIT:** I just tested the latest minio release by creating a new bucket and adding some files on a new installation. The new file format is different and there is no direct mapping of a single file inside a bucket to a single file in the filesystem anymore. So one can't use minio as a simple upload interface reusing the raw bucket data from another program anymore.

hgy59:
We now have published the latest package version  `v2022.10.24-16` that works with deprecated "fs" mode.
Future updates will not work without migration, documented in our wiki: https://github.com/SynoCommunity/spksrc/wiki/MinIO.

PR #5430, opened 2022-09-14
allow all characters in installer variables

hgy59:
## Description

- redesign `reload_inst_variables` function to allow any character
- required to provide passwords with special characters like $ < \ > | " ' = :space: ...
- follow up to #4936 that fixed some of the special characters only

Fixes  #5428

### Type of change

- [x] Includes small framework changes

PR #5555, opened 2023-01-08
update minio to 2022-10-24T18-35-07Z

fgma:
## Description

Update minio to 2022-10-24T18-35-07Z. This is the last release supporting the legacy FS backend. Details see https://github.com/SynoCommunity/spksrc/issues/5549 

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


hgy59:
@fgma as #5430 is merged, it is possible to have space (and other special characters) in minio password entered in the install- or upgrade-wizard.
The format of the custom environment-variables has changed and I have updated the changelog in spk/minio/Makefile. Please keep this text when resolving the conflict for this tile.

fgma:
done

hgy59:
> Sorry, I've made a mistake and messed up the merge.

Yes, I think you still want to update to `2022-10-24T18-35-07Z` for compatibility with current package.

fgma:
Should be OK now.

hgy59:
> Anything left here?

We should add some information to the changelog, regarding that future versions will not be compatible with this one.
And we should know how the update/migration will work.

fgma:
I've updated the changelog. My idea was to get this update out so everyone has at least this version and new installations will not need a migration. Migration is a separate step for me.

Any ideas on how we should handle this? I don't see any chance doing this automatically because one would need the same amount of temporary storage used by data stored in minio (at least in the biggest bucket) which is not possible to guarantee.

fgma:
I've upgraded my whole setup to this version for production use. How are we going to continue from here? I'd like to continue upgrading to the latest minio release as soon as possible.

th0ma7:
@fgma and @hgy59 is there a hard requirement to #5649 or can this be merged and published as-is?

fgma:
@th0ma7 Not for me. I've used the manual shared folder creation on my system as a workaround.

hgy59:
> @fgma and @hgy59 is there a hard requirement to #5649 or can this be merged and published as-is?

The target of #5649 will be an independent solution to fix the shared folders.
By _independent_ I mean that I will have to create new Makefile variables to not touch existing implementations.
So this is not a blocker for any package that use the old version.
_(but AFAICR @fgma wants to have the new version to benefit from the feature to create shares on diferent volumes in advance)_.

fgma:
@hgy59 I'm don't really need https://github.com/SynoCommunity/spksrc/pull/5649 but I want to upgrade to the latest minio release which uses a new data format and the release in this PR ist the last one which supports the old format and the new format at the same time. For details have a look at https://github.com/SynoCommunity/spksrc/issues/5549

fgma:
Anything left for me to do do get this finally merged?

The currently available version has at least two problems:

1. No security updates for over a year.
2. Everyone who stores data in the old data format has to do a manual migration at some point in time. The longer this update takes the more users are affected.

We need a strategy to get this finally resolved.

th0ma7:
I may be able to provide a review over the weekend although from a strategy standpoint it needs a volunteer to take it over (unless that would be you, haven't followed this thread in particular)

fgma:
I can't decide how the migration path should look like. I've opened a thread for discussing the issue almost a year ago which received no reaction at all: https://github.com/SynoCommunity/spksrc/issues/5549

After doing my own migration I'm pretty sure any automation would result in a disaster. It took me multiple weeks for a few terrabytes because tools would timeout all the time and miss file etc. Maybe we need to create a new package or something like that.

Getting this update out would at least would save new users from a migration their data.

th0ma7 (approved):
LGTM

th0ma7:
So considering the other changes are in (thnx @hgy59) i guess this is a rather trivial update.  What I would suggest is adding a wizard with a single page to explain that this will be the LAST version compatible with the older format.  You would need to add that as well to the changelog so it is more obvious as info would be hands-on even prior to installing.

On top of that, I'd suggest creating a "new" package for the "new" format, perhaps called `minio-mc` or something similar to make a distinction between the two.  So for the older package you clearly state that this is the last version... then in the new package you explain that this is the newer one.

Guessing there will be issues to be sorted out relatively to ports in usage or else?   You could then create a wiki page in spksrc github section to explain how to migrate from `minio` to `minio-mc` ... 

Otherwise, as-is it looks fine.  The above proposal can be done at a later while creating the "new" package where the older one can be re-published with the added wizard and info in the changelog.

My two cents...

fgma:
I've added a step in the wizard explaining the situation to the user. This ready for me. 

After the merge I would create a new package based on a copy of the released version and upgrade it to the current minio release.

spk/minio/src/wizard/install_uifile
```diff
@@ -96,5 +96,22 @@
...
+         },
+         {
+            "desc": "To check if you're affected by this change check your buckets manually by looking into each folder on your diskstation. If there is exactly a single file per S3 object you need to migrate your data. Otherwise you're already running a single-node single-drive deployment an can upgrade to the new MinIO package in future without a migration."
+         },
+         {
+            "desc": "Because of the complexity and risks involved an automatic upgrade process is not available. The official MinIO documentation provides a guide for doing so called 'Migrate from Gateway or Filesystem Mode'"
+         },
+         {
+            "desc": "For a regular user it is suggested to run an instance of MinIO on a seperate system and migrate your data to this new instance. This system could be a desktop, single board computer or a server. If possible you can mount the storage from you diskstation over the network and use it directly during the migration process. After a successful migration replace the data of your current MinIO installation on your diskstation. Please stop MinIO before replacing your data and restart it afterwards."
```
hgy59, line 113:
```suggestion
            "desc": "For a regular user it is suggested to run an instance of MinIO on a seperate system and migrate your data to this new instance. This system could be a desktop, single board computer or a server. If possible you can mount the storage from your diskstation over the network and use it directly during the migration process. After a successful migration replace the data of your current MinIO installation on your diskstation. Please stop MinIO before replacing your data and restart it afterwards."
```

spk/minio/src/wizard/upgrade_uifile.sh
```diff
@@ -114,6 +114,23 @@ cat <<EOF > $SYNOPKG_TEMP_LOGFILE
...
+         },
+         {
+            "desc": "To check if you're affected by this change check your buckets manually by looking into each folder on your diskstation. If there is exactly a single file per S3 object you need to migrate your data. Otherwise you're already running a single-node single-drive deployment an can upgrade to the new MinIO package in future without a migration."
+         },
+         {
+            "desc": "Because of the complexity and risks involved an automatic upgrade process is not available. The official MinIO documentation provides a guide for doing so called 'Migrate from Gateway or Filesystem Mode'"
+         },
+         {
+            "desc": "For a regular user it is suggested to run an instance of MinIO on a seperate system and migrate your data to this new instance. This system could be a desktop, single board computer or a server. If possible you can mount the storage from you diskstation over the network and use it directly during the migration process. After a successful migration replace the data of your current MinIO installation on your diskstation. Please stop MinIO before replacing your data and restart it afterwards."
```
hgy59, line 131:
```suggestion
            "desc": "For a regular user it is suggested to run an instance of MinIO on a seperate system and migrate your data to this new instance. This system could be a desktop, single board computer or a server. If possible you can mount the storage from your diskstation over the network and use it directly during the migration process. After a successful migration replace the data of your current MinIO installation on your diskstation. Please stop MinIO before replacing your data and restart it afterwards."
```

hgy59:
@fgma sorry for my late review

Looking at the details of the [migration](https://min.io/docs/minio/linux/operations/install-deploy-manage/migrate-fs-gateway.html) I got some new ideas.

This minio package provides the wizard options to chose the shared folder independent of new installation or package update. In either case, you can use a folder with or without a minio configuration.

Since the current version and the version in this PR supports the _legacy FS_ and the _SNSD_ (Single-Node Single-Drive) mode, we should now use the SNSD-Mode when the shared folder does not contain a minio configuration.

The installer could check whether the shared folder already contains a minio configuration and either configure minio server for legacy FS or SNSD mode.
If the existing configuration is not already SNSD, this Minio version (2022-10-24T18-35-07Z) will display a deprecation warning in the installation wizard, and future versions will refuse the installation without former migration.

I suppose the minio server mode can be recognized with the file `.minio.sys/format.json` in the shared folder. For the legacy FS mode this looks like:
```json
{
  "version": "1",
  "format": "fs",
  "id": "05583150-3d04-4058-bc30-7e8600932b87",
  "fs": {
    "version": "2"
  }
}
```

I highly recommend that this update uses the SNSD mode by default so it is useful for the migration of existing installations.

@fgma what do you think about it?

hgy59:
@fgma The minio console displays the version `DEVELOPMENT.GOGET` under metrics - info - server.
Isn't there a way to build an official version that displays `2022-10-24T18-35-07Z`?

hgy59:
@fgma can you rebase this PR to current master so we have #5649 to apply the redesigned shared folder handling?

fgma:
@hgy59 I'll try to get the fixes done during the next days. 

Concerning bucket versioning: I've verified what you already noticed. When creating a new install even older versions of minio already create SNSD based data by default for a new installation. I've just tested this on version 2022-10-24T18-35-07Z and 2022-09-07T22-25-02Z (current version packaged) using official minio binaries on windows. The issue is mainly a problem for all long term users already having data stored inside minio using quite old releases for the initial install.

How does one gracefully stop an installation? We would need to dynamically detect the data version after the user has selected the storage location and then decide if we can continue or stop with an error. Is there is something similar in the codebase already?

spk/minio/src/wizard/install_uifile
fgma: fixed in https://github.com/SynoCommunity/spksrc/pull/5555/commits/6ecaa703bce6800a749d049e4c1d9a2c5091eefe

spk/minio/src/wizard/upgrade_uifile.sh
fgma: fixed in https://github.com/SynoCommunity/spksrc/pull/5555/commits/6ecaa703bce6800a749d049e4c1d9a2c5091eefe

fgma:
> @fgma The minio console displays the version `DEVELOPMENT.GOGET` under metrics - info - server. Isn't there a way to build an official version that displays `2022-10-24T18-35-07Z`?

fixed in https://github.com/SynoCommunity/spksrc/pull/5555/commits/6727bb7a37570705df37ae965f6a479a49762082

fgma:
Any feedback on this?

d8x:
How can we move forward with this?

hgy59:
This PR is lacking migration to redesigned shared folder handling.

```
SERVICE_WIZARD_SHARE = wizard_data_directory
USE_DATA_SHARE_WORKER = yes
```

should be replaced by
```
SERVICE_WIZARD_SHARENAME = wizard_data_directory
```

Then the install_uifile has to be adjusted (remove volume configuration)
and service-setup.sh must use SHARE_PATH variable instead of wizard_data_directory 

probably it is better to rename the variable wizard_data_directory to wizard_share_name to reflect the fact that it is the name of the share and not a full path anymore.

finally upgrade_uifile.sh and service-setup.sh have to be adjusted for the shared folder.

There is a good example of modifications in spk/owncloud/  by @mreid-tt

hgy59:
> How does one gracefully stop an installation? We would need to dynamically detect the data version after the user has selected the storage location and then decide if we can continue or stop with an error. Is there is something similar in the codebase already?

Sorry for my late reply.
There are the functions `validate_preinst` and `validate_preupgrade` that can be used for this kind of validation.

Just echo an "ERROR TEXT" and use "exit 1" in a validate function to refuse the installation or upgrade of the package (you find commented examples in spk/demoservice/src/service-setup.sh).

All such functions (service_* and validate_*) are called after wizard pages are completed and the installation or upgrade is started.

hgy59:
@fgma glad to tell you that this old pending PR is about to be ready to merge.

The redesinged shared folder handling is implemented and wizards are adjusted.

I shortened the Information about migration in the wizards and created a wiki page showing the detailed migration procedure.
(https://github.com/SynoCommunity/spksrc/wiki/MinIO)

Future versions only need a validation of the data format to refuse installation/upgrade if FS format is found in the shared folder.

I know you can't test this since you've already migrated your installation, but having a second set of eyes on it is still welcome.

th0ma7 (approved):
@hgy59 just got back, I haven`t build & tested but overall changes looked good to me.  And makes me wonder if the `vcs` related build options wouldn't be of interest to use for one of my other PR relatively to IGC stack.

IngmarStein:
Is this ready to go (followed by #6221)?

Issue #5439, opened 2022-09-28
[Package Request] restic 


stephane-archer:
### Software Name

restic 

### Brief Description

restic backup software

### Website

https://github.com/restic/restic

### Documentation

https://restic.readthedocs.io/en/latest/010_introduction.html

### Build/Installation documentation

https://github.com/restic/restic

### Source code

https://github.com/restic/restic

### OpenSource License

BSD-2-clause

ecatphi8:
Would it be possible to integrate resticprofile https://github.com/creativeprojects/resticprofile in the package ?
It looks like it would just be one extra binary.

rclone too would be awsome even it rclone might deserve its own package
https://rclone.org/

stephane-archer:
@hgy59 could you clarify why this is closed ? Is restic available ? I can only see rclone activities referenced on the issue

hgy59:
#6512 wrongly closed this PR

stephane-archer:
Amazing job thank you @fgma

PR #5557, opened 2023-01-09
restic: inital package of version 0.14.0

fgma:
## Description

Initial package of restic. Restic is a fast and secure backup program.

[https://restic.net/](https://restic.net/)

Fixes https://github.com/SynoCommunity/spksrc/issues/5439

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update


cross/restic/Makefile
```diff
@@ -0,0 +1,28 @@
+PKG_NAME = restic
+PKG_VERS = 0.15.0
+PKG_EXT = tar.gz
+PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
+PKG_DIST_SITE = https://github.com/restic/restic/releases/download/v$(PKG_VERS)
+EXTRACT_PATH = $(WORK_DIR)
```
hgy59, line 6: `$(WORK_DIR)` is the default value for `EXTRACT_PATH`, so you can omit this line.

fgma: Fixed in https://github.com/SynoCommunity/spksrc/pull/5557/commits/1976694e071212c765f57428d6a67276eeb81e7e

fgma:
Updated to current master. Anything else left for me to do here?

fgma:
Updated to latest release. Anything left for me to do?

mpgirro:
Thanks for implementing a package for restic! I'd be great if this gets merged and made available soon 🙂 

Could you also add [resticprofile](https://creativeprojects.github.io/resticprofile/) to this package? It's a profile manager for restic to enabled configuration file based backups.

mreid-tt:
Hi @fgma, thanks for your patience with this PR - it's been a long wait!

I've reviewed this against the newer PR #6764 and your implementation is preferred. To get this merged, could you please update to v0.18.1 (current latest)? You can also merge the current master to confirm the build still works.

Once updated, we can close #6764 as superseded. Thanks again for sticking with this!

spk/restic/Makefile
```diff
@@ -0,0 +1,21 @@
+SPK_NAME = restic
+SPK_VERS = 0.18.1
+SPK_REV = 1
+SPK_ICON = src/logo.png
+
+DEPENDS = cross/$(SPK_NAME)
```
mreid-tt:
```suggestion
DEPENDS = cross/restic
```

fgma: Thanks, just found it myself.

hgy59:
@mreid-tt the build for ARMv5 must use native/go_1.23. If it does not build with with go 1.23, ARMv5 must be declared unsupported.
I just have deactivated the published package for 88f6281-6.2.4.

@th0ma7 shall we add a guard the spksrc.go-env.mk to fail when go packages for ARMv5 do not use native/go_1.23?

kernel < 3.2  is not supported with go > 1.23
related info:
https://go.dev/doc/go1.23#linux

mreid-tt:
hey @fgma, note the ARMv5 and go version constraint. I guess when the next update is available you can experiment with the older go version and/or add an exclusion for ARMv5.

fgma:
I understand how to exclude ARMv5, but how to I force a certain version to build with a different go version? If you have some remarks how to do this for an update just leave a comment here as a reference for the next update.

mreid-tt:
It would be to change this line:

https://github.com/SynoCommunity/spksrc/blob/11aa36a1c4e6751bfc992c63822cbb5e2ad293a4/cross/restic/Makefile#L8

to something like:
```
BUILD_DEPENDS = native/go_1.23
```

fgma:
I guess this would lead to unfixed security issues for all other architectures because 1.23 will not receive security fixes anymore as 1.25 is the current stable release.

mreid-tt:
> I guess this would lead to unfixed security issues for all other architectures because 1.23 will not receive security fixes anymore as 1.25 is the current stable release.

I don't know enough about the go language and how this may affect builds but if this is a concern then a simple exclusion of ARMv5 is also a solution.

hgy59:
> I don't know enough about the go language and how this may affect builds but if this is a concern then a simple exclusion of ARMv5 is also a solution.

you need to add 
```
BUILD_DEPENDS = native/go_1.23
NATIVE_GO = native/go_1.23
else
BUILD_DEPENDS = native/go
NATIVE_GO = native/go
endif
```

as we have in node-exporter, duf, fzf, gdu, and others.

hgy59:
> > I guess this would lead to unfixed security issues for all other architectures because 1.23 will not receive security fixes anymore as 1.25 is the current stable release.
> 

Neither DSM, the kernels, the toolchains (compiler, system libraries et. al.), ... get updates for most of the diskstation models.
It is always the responsibility of the users to ensure secure usage.

PR #5567, opened 2023-01-19
Add support for DS223 and build for DSM 7.1

hgy59:
## Description

Newer models like DS223 do not support DSM 7.0 but come with DSM 7.1+
Add the new arch `rtd1619b` to generic ARMv8 archs.

### Type of change

- [x] Includes framework changes => Update default DSM 7.x to DSM 7.1


toolchain/syno-rtd1619b-7.2/Makefile
```diff
@@ -0,0 +1,12 @@
+TC_ARCH = rtd1619b
```
th0ma7, line 1: Ah, this is the new arch, and finally with a decent kernel version!  Let's hope they end-up switching to an supported lts for x64

th0ma7 (changes requested):
Overall lgtm.  Until Synology release 7.1 kernel sources I strongly suggest we keep the default dsm7 version to 7.0... unless there is a mandatory requirement for that?

hgy59:
> Overall lgtm. Until Synology release 7.1 kernel sources I strongly suggest we keep the default dsm7 version to 7.0... unless there is a mandatory requirement for that?

Independent of the default DSM Version (github action and `make setup`) I will start to publish packages for DSM 7.1 instead of DSM 7.0. 
IMHO there is some risk that x23 Models (that ship with DSM >= 7.1.1) fail to run builds for DSM 7.0.
DSM 7.1 will be needed for some time as DSM 7.2 will drop support for some ARMv7 arch and for evansport. And DSM 7.0 is obsolete IMO.

PR #5570, opened 2023-01-21
Fix compatibility with POSIX shell in SRM (ash)

### LINKED ISSUE #5548: SRM version of Cloudflared not working
### Is this a new Bug?

### Package Name

Cloudflared

### Package Version

v2022.12.1-5

### Device Model

MR2200ac

### Device Architecture

ARMv7

### Firmware Version

SRM 1.3.1-9316

### What happened?

I can install the cloudflared package but impossible to start after the token validation.

Package ask for update

### Reproduction steps

```bash
1. Install package
2. Token
3. Try to start
...
```

### Install Log

```log
N/A
```

### Service Log

_No response_

### Other Logs

_No response_

publicarray:
## Description

Rather than using arrays I'm using the newline character for separation, it is more hacky but should still work. ~~I still need to test if deluge works with this.~~ I have confirmed that Cloudflared works with this change on the SRM.

Fixes #5548

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update


publicarray:
Alternatively if this approach is too hackey we could use a token character to separate multiple SERVICE_COMMANDs e.g. `:`

hgy59:
@publicarray what about the `reload_inst_variables` function in installer.functions? Is uses declare to inject the variables from the `${INST_VARIABLES}` file and might fail on SRM too. _(there is a fix in #5550 to declare globally (-g))_

publicarray:
Thanks @hgy59 I'll take a look

publicarray:
I think using `export` should work instead of `declare -g`

hgy59:
> I think using `export` should work instead of `declare -g`

@publicarray you are right. The `reload_inst_variables function` was introduced with the minio package. There is no need to use `declare` and I have redesigned this function to take the file to read variable declarations as parameter.

As I already did a redesign to allow special characters in such variables (#5430) there is a conflict that must be resolved.

Finally PR #5555 can benefit of the new function `load_variables_from_file`.

PR #5582, opened 2023-01-28
Fix rustc builds on github-action

### LINKED ISSUE #5581: gihub build action broken
####  The github build action is broken since some time

The first failing run was https://github.com/SynoCommunity/spksrc/actions/runs/3974432067 on `Sat, 21 Jan 2023 10:34:33 GMT`

The latest sucessfully run was https://github.com/SynoCommunity/spksrc/actions/runs/3979813908 and this was after the first failing one on `Sun, 22 Jan 2023 13:05:32 GMT`

From then on, all builds fail for all archs except aarch64-6.1 and aarch64-7.1
The builds fail only, when python is involved as DEPENDS or BUILD_DEPENDS.

Before the first failing, there was #5567 merged, that changed the DSM 7 version from 7.0 to 7.1 for the github build action. IMHO this is not related to the failure, as after this, the build action https://github.com/SynoCommunity/spksrc/actions/runs/3974183723 on `Sat, 21 Jan 2023 09:18:33 GMT` and others succeeded.

Regarding the logs, the failing might be related to #5508 when the docker image was updated to install rust on the fly. See a discussion in 
https://github.com/SynoCommunity/spksrc/pull/5478#issuecomment-1399597940

one of the errors appears while building python wheels:
```
ERROR: Could not build wheels for maturin, which is required to install pyproject.toml-based projects
```
The reason is:
```
2023-01-22T22:04:43.1539599Z       error: can't find Rust compiler
```

####  related error log: 
```
2023-01-22T22:04:42.6251008Z Building wheels for collected packages: maturin
2023-01-22T22:04:42.6264913Z   Building wheel for maturin (pyproject.toml): started
2023-01-22T22:04:43.1397982Z   Building wheel for maturin (pyproject.toml): finished with status 'error'
2023-01-22T22:04:43.1480634Z   error: subprocess-exited-with-error
2023-01-22T22:04:43.1480988Z   
2023-01-22T22:04:43.1481403Z   × Building wheel for maturin (pyproject.toml) did not run successfully.
2023-01-22T22:04:43.1481759Z   │ exit code: 1
2023-01-22T22:04:43.1530264Z   ╰─> [33 lines of output]
2023-01-22T22:04:43.1530897Z       ModuleNotFoundError while trying to load entry-point editable_wheel: No module named 'setuptools.command.editable_wheel'
2023-01-22T22:04:43.1531301Z       running bdist_wheel
2023-01-22T22:04:43.1531589Z       running build
2023-01-22T22:04:43.1532054Z       running build_py
2023-01-22T22:04:43.1532289Z       creating build
2023-01-22T22:04:43.1532638Z       creating build/lib.linux-x86_64-cpython-310
2023-01-22T22:04:43.1533062Z       creating build/lib.linux-x86_64-cpython-310/maturin
2023-01-22T22:04:43.1533525Z       copying maturin/__init__.py -> build/lib.linux-x86_64-cpython-310/maturin
2023-01-22T22:04:43.1534183Z       copying maturin/import_hook.py -> build/lib.linux-x86_64-cpython-310/maturin
2023-01-22T22:04:43.1534757Z       copying maturin/__main__.py -> build/lib.linux-x86_64-cpython-310/maturin
2023-01-22T22:04:43.1535059Z       running egg_info
2023-01-22T22:04:43.1535348Z       creating maturin.egg-info
2023-01-22T22:04:43.1535678Z       writing maturin.egg-info/PKG-INFO
2023-01-22T22:04:43.1536103Z       writing dependency_links to maturin.egg-info/dependency_links.txt
2023-01-22T22:04:43.1536539Z       writing requirements to maturin.egg-info/requires.txt
2023-01-22T22:04:43.1536959Z       writing top-level names to maturin.egg-info/top_level.txt
2023-01-22T22:04:43.1537382Z       writing manifest file 'maturin.egg-info/SOURCES.txt'
2023-01-22T22:04:43.1537789Z       reading manifest file 'maturin.egg-info/SOURCES.txt'
2023-01-22T22:04:43.1538162Z       reading manifest template 'MANIFEST.in'
2023-01-22T22:04:43.1538788Z       writing manifest file 'maturin.egg-info/SOURCES.txt'
2023-01-22T22:04:43.1539069Z       running build_ext
2023-01-22T22:04:43.1539302Z       running build_rust
2023-01-22T22:04:43.1539599Z       error: can't find Rust compiler
2023-01-22T22:04:43.1539837Z       
2023-01-22T22:04:43.1540300Z       If you are using an outdated pip version, it is possible a prebuilt wheel is available for this package but pip is not able to install from it. Installing from the wheel would avoid the need for a Rust compiler.
2023-01-22T22:04:43.1540745Z       
2023-01-22T22:04:43.1540939Z       To update pip, run:
2023-01-22T22:04:43.1541162Z       
2023-01-22T22:04:43.1541433Z           pip install --upgrade pip
2023-01-22T22:04:43.1541665Z       
2023-01-22T22:04:43.1541899Z       and then retry package installation.
2023-01-22T22:04:43.1542144Z       
2023-01-22T22:04:43.1542772Z       If you did intend to build this package from source, try installing a Rust compiler from your system package manager and ensure it is on the PATH during installation. Alternatively, rustup (available at https://rustup.rs) is the recommended way to download and update the Rust compiler toolchain.
2023-01-22T22:04:43.1543361Z       [end of output]
2023-01-22T22:04:43.1543577Z   
2023-01-22T22:04:43.1543884Z   note: This error originates from a subprocess, and is likely not a problem with pip.
2023-01-22T22:04:43.1544238Z   ERROR: Failed building wheel for maturin
2023-01-22T22:04:43.1544499Z Failed to build maturin
2023-01-22T22:04:43.1544964Z ERROR: Could not build wheels for maturin, which is required to install pyproject.toml-based projects
```

####  local builds do not fail (tested again with current docker image)
```
ghcr.io/synocommunity/spksrc       latest                 d98ac8b9ac48   3 weeks ago     1.04GB
```

hgy59:
## Description

This tries to fix #5581

- fix export of CARGO_HOME and RUSTUP_HOME
- fix issue when toolchain caching keeps rustc status file (`toolchain/syno-<arch>-<tcversion>/work/.rustc_done`) in cases where distrib folder install of cargo/rustc isn't existing.

### Type of change

### DIFF
--- a/spk/python310/src/requirements-crossenv.txt
+++ b/spk/python310/src/requirements-crossenv.txt
@@ -22,6 +22,7 @@ immutables==0.19
 MarkupSafe==2.1.1
 msgpack-python==0.5.6
 netifaces==0.11.0
+psutil==5.9.4
 regex==2022.10.31
 SQLAlchemy==1.4.45
 zope.interface==5.5.2
@@ -31,48 +32,45 @@ zope.interface==5.5.2
 # Require environment variables
 #  GEVENTSETUP_EMBED_CARES=FALSE
 #  GEVENTSETUP_EMBED_LIBEV=FALSE
-gevent==22.10.2
+# gevent==22.10.2
 
 # [lxml]
 # Depends: libxml2, libxslt
-lxml==4.9.2
+# lxml==4.9.2
 
 # [mysqlclient]
 # Depends: mysql-connector-c, mariadb-connector-c
 # Require environment variables
 #  MYSQLCLIENT_CFLAGS
 #  MYSQLCLIENT_LDFLAGS
-mysqlclient==2.1.1
+# mysqlclient==2.1.1
 
 # [Pillow]
 # Require --global-options arguments
 #  WHEELS_BUILD_ARGS = [Pillow] build_ext --disable-platform-guessing
-Pillow==9.3.0
-
-# [psutil]
-psutil==5.9.4
+# Pillow==9.3.0
 
 # [pycares]
 # Depends: c-ares
 # Require environment variables
 #  PYCARES_USE_SYSTEM_LIB=1
-pycares==4.3.0
+# pycares==4.3.0
 
 # [pycurl]
 # Depends: curl
 # Require environment variables
 #  PYCURL_CURL_CONFIG
-pycurl==7.45.2
+# pycurl==7.45.2
 
 # [PyNaCl]
 # Depends: cross/libsodium
 # Require environment variables
 #  SODIUM_INSTALL=system
-PyNaCl==1.5.0
+# PyNaCl==1.5.0
 
 # [PyYAML]
 # Depends: libyaml
-PyYAML==6.0
+# PyYAML==6.0
 
 # [rencode]
 # Updated fork of the project
@@ -80,7 +78,7 @@ PyYAML==6.0
 git+https://github.com/totaam/rencode.git@f6254ab26161f90b9c5e97915b9193fee805fc1f#egg=rencode==1.0.7
 
 # [ujson]
-#   - Require setuptools-scm in cross/python310 crossenv
+#   - Require setuptools-scm in cross/python3* crossenv
 ujson==5.6.0
 
 # [webrtcvad]
--- a/.github/actions/build.sh
+++ b/.github/actions/build.sh
@@ -28,6 +28,11 @@ echo "===> TARGET: ${GH_ARCH}"
 echo "===> ARCH   packages: ${ARCH_PACKAGES}"
 echo "===> NOARCH packages: ${NOARCH_PACKAGES}"
 
+# Remove rust toolchain status file to enfore re-installing cargo/rust
+# This fixes isees on github-action where toolchain caching omits the
+# actual installation state of cargo/rust within the distrib folder
+rm -f toolchain/syno-${GH_ARCH}/work/.rustc_done
+
 if [ "${GH_ARCH%%-*}" = "noarch" ]; then
     build_packages=${NOARCH_PACKAGES}
 else
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -3,9 +3,9 @@
 
 # Add cargo for rust compiler to default PATH
 ifneq ($(BASE_DISTRIB_DIR),)
-export CARGO_HOME = $(BASE_DISTRIB_DIR)/cargo
-export RUSTUP_HOME = $(BASE_DISTRIB_DIR)/rustup
-export PATH := $(BASE_DISTRIB_DIR)/cargo/bin:$(PATH)
+export CARGO_HOME=$(BASE_DISTRIB_DIR)/cargo
+export RUSTUP_HOME=$(BASE_DISTRIB_DIR)/rustup
+export PATH:=$(BASE_DISTRIB_DIR)/cargo/bin:$(PATH)
 endif
 
 RUST_TOOLCHAIN ?= stable
--- a/mk/spksrc.tc-rust.mk
+++ b/mk/spksrc.tc-rust.mk
@@ -39,8 +39,13 @@ rustc_msg:
 	@$(MSG) "- default PATH: $(PATH)"
 
 pre_rustc_target: rustc_msg
-	@$(MSG) "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path"
-	@curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path
+	@$(MSG) "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path" ; \
+	exec 5> /tmp/tc-rustc.lock ; \
+	flock --timeout $(FLOCK_TIMEOUT) --exclusive 5 || exit 1 ; \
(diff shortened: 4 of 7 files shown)


hgy59:
@th0ma7 installation of rustc does not work with github builds.
It does not work, since rustc is installed with the toolchains and the toolchains are cached.
We would need a different cache for the distrib/rustup folder, but there is not checksum file, to trigger changes.
We might use distrib/cargo/CACHEDIR.TAG as checksum file, but I don't know whether this would cover changes in the distrib/rustup folder too.

I will try to add caches for rustup and cargo. But rustup will never be updated when installed only with toolchains (as toolchain never will be updated). _I am trying this on PR #5284_

Any idea on how to install rustc independent of the toolchains?

th0ma7:
@hgy59 just got back from vacation... Although I did read the thread on my way back and may have an idea...

If caching is the issue then why not stop caching?  Let me explain:
- Part of previous changes was the ability to "download-only" the toolchain packages (part of one of my kernel changes, would need to find this back)
- Keeping this in mind, why not simply download the toolchain `tar.gz` files and stop caching the `spksrc/toolchain` directory entirely?
- This would ensure to start by mdsum + extracting the pre-downloaded toolchain + installing rustc

With this there is almost no changes needed from a github-action point of view with the exception of:
1. Enforcing the download of all required toolchain files
2. Stop caching `spksrc/toolchain`
3. Start caching `spksrc/distrib/toolchain` (maybie not even needed as `distrib` probably already there)

As such, at every build spksrc would validate mdsum of the pre-downloaded toolchain files, extract then install rustc and voilà!

th0ma7:
Looking at the code I believe `actions/prepare.sh` needs adjusting so all archs `toolchain/syno-${arch}` gets added on top of the `DEPENDENCY_LIST` variable.  But in order to do that the `workflows/build.yml` needs adjusting presumably so `matrix.arch` variable gets somehow available to the `prepare` job ...

And really I'm not that much versed into github-action files...

hgy59:
> And really I'm not that much versed into github-action files...

The problem is not the github build action per se.

The problem is, that we currently install rust toolchains only once, when a toolchain is installed.
The toolchains will never get updated. So newer rust compiler or different toolchain versions (stable/nightly) get only installed by cleaning the extracted toolchains.

We can define the RUST-TOOLCHAIN (arch) names at time of tc_vars.mk file creation (as we do now for cmake and other flags/settings), but we must not install rustc and cargo at this time.

IMHO we need something that we can define as dependency on demand.

th0ma7:
> The problem is, that we currently install rust toolchains only once, when a toolchain is installed.

Indeed and thus toolchains could only be downloaded instead of extracted and therefore rustc install will get trigerred as expected.

> The toolchains will never get updated.

Indeed, neither will their `.tar.gz` cached under `distrib/toolchain/*`.

> So newer rust compiler or different toolchain versions (stable/nightly) get only installed by cleaning the extracted toolchains.

Or by lack of being extracted in the first place.

> We can define the RUST-TOOLCHAIN (arch) names at time of tc_vars.mk file creation (as we do now for cmake and other flags/settings), but we must not install rustc and cargo at this time.

The easiest really was to use it part of the default docker image like cmake or meson although with total image size issue.  I still think we would be better off to not be installing and caching the toolchain directory & extract later rather than tring to circumvent the issue differently.

> IMHO we need something that we can define as dependency on demand.

Do you mean something totally new like a `make rust-setup` or similar?

th0ma7:
> So newer rust compiler or different toolchain versions (stable/nightly) get only installed by cleaning the extracted toolchains.

Actually, just tought of this, we could simply alterate the toolchain status with the following:
```
spksrc/toolchain/syno-x64-6.1$ rm -f work/.rustc_done
spksrc/toolchain/syno-x64-6.1$ make
```
So guessing that would be broadly speaking would result simply into the following:
```
spksrc$ rm -f toolchain/syno-*/work/.rustc_done
```

Still, to limit possibilities of collisions a `flock` call similar to downloads could be made to serialize the rustc install so archs do not overlap one over the other.

th0ma7:
@hgy59 alternate fix is working out.  I've looked into the logs and rust install gets called as expected.  I suggest you incorporate changes to [.github/actions/build.sh](https://github.com/SynoCommunity/spksrc/pull/5583/files#diff-ef7be95265181888899b756eca483d03f5385ce2f53284f22a6fa57b17446bc4) and [mk/spksrc.tc-rust.mk](https://github.com/SynoCommunity/spksrc/pull/5583/files#diff-12c84b7c2347e4b3255f6413b745504941b6bb080470d8bac3b410164b583739) within this PR so I can rebase #5583 against master afterwards and confirm fix still works as expected.

Remaining is disabling/enabling properly the wheel builds so build is all green.

th0ma7 (approved):
LGTM

spk/python310/src/requirements-crossenv.txt
```diff
@@ -22,6 +22,7 @@ immutables==0.19
 MarkupSafe==2.1.1
 msgpack-python==0.5.6
 netifaces==0.11.0
+psutil==5.9.4
```
hgy59, line 25: @th0ma7 is psutil added by intention?
_just to be sure_

th0ma7: Nope, just reordered.  The sole intent is to facilitate wheel building testing when uncommenting requirements-crossenv.txt and serves as examples.

But by default crossenv wheels are always disabled and not packaged with py3.1*.  Only a minimal set of pure wheels are included.

PR #5585, opened 2023-02-01
update jellyfin 10.8.9 and dotnet to 6.0.405

mxwi:
## Description

- update jellyfin version to 10.8.9
- update dotnet to 6.0.405

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #5584, opened 2023-02-01
Lidarr: Update dotnet build for DSM 7.2

mreid-tt:
## Description

This includes the following fixes:

1. Update dotnet build for DSM 7.2
2. Update compatible ARMv7 archs
3. Add support for x86 (32-bit) architecture
4. Remove outdated mono builds

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] Package update

PR #5590, opened 2023-02-02
jellyfin: add note for unsupported archs to changelog

hgy59:
## Description

followup to #5585
to clarify that most archs (32-bit archs) are not supported anymore

### DIFF
--- a/spk/jellyfin/Makefile
+++ b/spk/jellyfin/Makefile
@@ -8,13 +8,14 @@ DSM_UI_DIR = app
 
 DEPENDS = cross/libstdc++ cross/jellyfin cross/jellyfin-web
 
-UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7_ARCHS) $(ARMv7L_ARCHS) $(i686_ARCHS)
+# x64 and armv8 archs are supported only.
+UNSUPPORTED_ARCHS = $(32bit_ARCHS)
 
 MAINTAINER = stevenliuit
 DESCRIPTION = "The Free Software Media System. It is an alternative to the proprietary Emby and Plex."
 DISPLAY_NAME = Jellyfin
 STARTABLE = yes
-CHANGELOG = "1. Update jellyfin to 10.8.9<br>2. Update dotnet 6.0.405"
+CHANGELOG = "1. Update jellyfin to 10.8.9<br/>2. Update dotnet to 6.0.13<br/>3. Remove support for 32-bit archs."
 HOMEPAGE = https://jellyfin.org
 HELPURL = https://jellyfin.org/docs/general/server/settings.html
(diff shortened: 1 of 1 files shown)


hgy59:
@mreid-tt this should help to avoid questions: users will see the change log, when they want to manually download the package.

hgy59:
and I adjusted the dotnet version in the change log (6.0.405 is the sdk version and 6.0.13 is the dotnet version).

spk/jellyfin/Makefile
```diff
@@ -8,13 +8,14 @@ DSM_UI_DIR = app
...
-UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7_ARCHS) $(ARMv7L_ARCHS) $(i686_ARCHS)
+# x64 and armv8 archs are supported only.
+UNSUPPORTED_ARCHS = $(32bit_ARCHS)
 
 MAINTAINER = stevenliuit
 DESCRIPTION = "The Free Software Media System. It is an alternative to the proprietary Emby and Plex."
 DISPLAY_NAME = Jellyfin
 STARTABLE = yes
-CHANGELOG = "1. Update jellyfin to 10.8.9<br>2. Update dotnet 6.0.405"
+CHANGELOG = "1. Update jellyfin to 10.8.9<br/>2. Update dotnet to 6.0.13<br/>3. Remove support for 32-bit archs."
```
mreid-tt, line 18: I saw a similar modification from @publicarray regarding `<br/>`. If you can, I'd really like your help validating the need for this because I have two bits of information which conflict with this:

1. The example for the `CHANGELOG` in our own Wiki page on [Available Variables in spk/ Makefiles](https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables#available-variables-in-spk-makefiles) shows `<br>` separating multiple entries
2. The W3C Markup Validation Service has some documentation on [Trailing slashes](https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes) which indicates that "_a trailing slash to mark start tags as self-closing_" ... "_does not concur with what the HTML standard states_"

As such, I'd really like some clarity on the correct use of HTML with void elements in the change log.

mreid-tt:
Thanks for the amendment. The PR looks good but I have one question regarding the HTML syntax given a similar note from @publicarray in https://github.com/SynoCommunity/spksrc/pull/5584#discussion_r1094191010.

spk/jellyfin/Makefile
hgy59: Reading the docs the trailing slash is only a problem for tags with attributes. 
So I leaf it as is (and updated the wiki page)

mreid-tt: Thanks for the response. I'll correct all my pending PRs based on this.

PR #5529, opened 2022-12-20
Radarr: fix package update and version bump

mreid-tt:
## Description

This includes the following fixes:

1. Upgrade to Radarr 4.3.2.6857
2. Avoid modification of config.xml on package update (see #5514)
3. Update compatible ARMv7 archs
4. Add support for x86 (32-bit) architecture

Fixes #5363

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [x] Package update

PR #5587, opened 2023-02-02
Prowlarr: fix dotnet archs and version bump

mreid-tt:
## Description

This includes the following fixes:

1. Update Prowlarr to v1.1.3.2521
2. Update compatible ARMv7 archs
3. Add support for x86 (32-bit) architecture

Fixes #5542

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [x] Package update

PR #5110, opened 2022-01-29
[New Package] Readarr

EckPhi:
## Description

motivation: New e-book collection manager in *arr flavour, Readarr.

Closes: #4801 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5592, opened 2023-02-04
Radarr: fix for service setup

mreid-tt:
## Description

1. Amends service setup `GROUP` variable to exclude DSM7 (deprecated in DSM7)

Follow up of #5529
Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/spk/radarr/src/service-setup.sh
+++ b/spk/radarr/src/service-setup.sh
@@ -12,8 +12,9 @@ CMD_ARGS="-nobrowser -data=${RADARR_CONFIG_DIR}"
 # SPK_REV 15 has it in the wrong place for DSM 7
 LEGACY_CONFIG_DIR="${SYNOPKG_PKGDEST}/var/.config"
 
-# for DSM < 7 only:
-GROUP="sc-download"
+if [ ${SYNOPKG_DSM_VERSION_MAJOR} -lt 7 ]; then
+    GROUP="sc-download"
+fi
 
 SERVICE_COMMAND="env HOME=${HOME_DIR} LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib ${RADARR} ${CMD_ARGS}"
 SVC_BACKGROUND=y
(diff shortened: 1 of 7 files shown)


hgy59:
@mreid-tt (cc @publicarray @th0ma7 )
I doubt that it makes sense to define the `GROUP` variable for DSM 7+.
This variable is only used in `set_unix_permissions`.
It is only required when the installer runs as root user.

In DSM 7+ the user and group is taken from the privilege resource file only.
default previleges are username=sc-{package} groupname=sc-{package}
A package can define custom privileges with in `conf/privilege` and `CONF_DIR = conf`

For DSM 7 we do not have any other option - and the installer does not accept any "ctrl-script" entries (not even `"run-as": "package"`). So we can only use "username", "groupname" and "join-groupname"
For DSM 6 we two options
1. (default) run installer as root and adjust owner (and group) in post install
2. use the DSM 7 way

If we investiate into this, I propose to to use option 2 for all DSM 6 packages that do not need root access for installation and/or to run the package service(s).

GROUP Variable:
We should update the framework (i.e. the service and installer related files in the spksrc/mk folder) in a way, that no GROUP variable must be defined in the service-setup.sh script of packages. 

In spksrc.service.mk I see that we already have the variable "GROUP" - so we should apply a fix to avoid the need to define GROUP in service-setup.sh
Currently this GROUP variable is defined only when SERVICE_WIZARD_GROUP is used.

I propose to introduce a Makefile variable SERVICE_GROUP that can be used to define the group for the installers for all cases where no group comes from the install wiazrd.

If we really need a common group for multiple packages (i.e. "synocommunity" a.k.a "sc-download") we can use the Makefile variables SERVICE_GROUP or SERVICE_WIZARD_GROUP.


just some thoughts...

mreid-tt:
@hgy59, thanks for reviewing the proposed change. To give you a bit of background, this was part of my efforts to standardise the configurations of all the `*arr` packages following the work done on the new package for [Readarr](https://github.com/SynoCommunity/spksrc/pull/5110).

While I was running some tests on the build I kept seeing these line in the `/var/log/packages/readarr.log`:

```
2023/02/04 06:41:19	Begin load_variables_from_file
/volume1/@tmp/synopkg/install.1gJWjF/scripts/installer: line 42: /var/packages/readarr/etc/installer-variables: No such file or directory
2023/02/04 06:41:19	End load_variables_from_file
[---snip---]
2023/02/04 06:41:25	Begin load_variables_from_file
/var/packages/readarr/scripts/installer: line 42: /var/packages/readarr/etc/installer-variables: No such file or directory
2023/02/04 06:41:25	End load_variables_from_file
```

When I looked at the particulars of the file I found the following details:

```
$ ls -all /var/packages/readarr/etc/
total 4
drwxr-xr-x 1 sc-readarr synocommunity  38 Feb  4 06:41 .
drwxr-xr-x 1 root       root          194 Feb  4 06:41 ..
-rw-r--r-- 1 sc-readarr synocommunity  20 Feb  4 06:41 installer-variables
[---snip---]
$ cat /var/packages/readarr/etc/installer-variables 
GROUP=synocommunity
```

I then took a look again at the logs for an update I was building for [Prowlarr](https://github.com/SynoCommunity/spksrc/pull/5587). To my surprise, in `/var/log/packages/prowlarr.log`, the exact same lines kept coming up:

```
2023/02/04 06:31:23	Begin load_variables_from_file
/volume1/@tmp/synopkg/install.2VRFL9/scripts/installer: line 42: /var/packages/prowlarr/etc/installer-variables: No such file or directory
2023/02/04 06:31:23	End load_variables_from_file
[---snip---]
2023/02/04 06:31:38	Begin load_variables_from_file
/var/packages/prowlarr/scripts/installer: line 42: /var/packages/prowlarr/etc/installer-variables: Permission denied
2023/02/04 06:31:38	End load_variables_from_file
```

In fact, this situation was even weirder, because this was the log from a [build](https://github.com/SynoCommunity/spksrc/actions/runs/4086611112) I did that had a different approach for this code:

```
if [ ${SYNOPKG_DSM_VERSION_MAJOR} -lt 7 ]; then
    GROUP="sc-download"
fi
```

When looking into the directories, I confirmed that the output was as expected:

```
$ ls -all /var/packages/prowlarr/etc/
total 0
drwxr-xr-x 1 sc-prowlarr synocommunity   0 Feb  4 06:31 .
drwxr-xr-x 1 root        root          180 Feb  3 14:40 ..
```

So the file being complained about wasn't even there, yet it gave me messages about `Permission denied`. This is why I switched to at least writing out the file to give a `No such file or directory` error instead... or so I thought. Checking on the `/var/log/packages/radarr.log` for this build I see the same strange logs:

```
2023/02/04 06:32:31	Begin load_variables_from_file
/volume1/@tmp/synopkg/install.LBYrsK/scripts/installer: line 42: /var/packages/radarr/etc/installer-variables: No such file or directory
2023/02/04 06:32:31	End load_variables_from_file
[---snip---]
2023/02/04 06:32:48	Begin load_variables_from_file
/var/packages/radarr/scripts/installer: line 42: /var/packages/radarr/etc/installer-variables: Permission denied
2023/02/04 06:32:48	End load_variables_from_file
```

I have a permission denied for a file that does exist with the correct permissions as far as I could tell:

```
$ ls -all /var/packages/radarr/etc/
total 4
drwxr-xr-x 1 sc-radarr synocommunity  38 Feb  4 06:32 .
drwxr-xr-x 1 root      root          180 Feb  3 14:40 ..
-rw-r--r-- 1 sc-radarr synocommunity  20 Feb  4 06:32 installer-variables
[---snip---]
$ cat /var/packages/radarr/etc/installer-variables 
GROUP=synocommunity
```

As such, I'm back to square one and the results seem inconsistent. Further, this does not seem to happen to some of the packages in the current repository, so my guess is something has changed in the build process.

I'll take a look at your suggestions above and get back to you on them. I'm not deeply familiar with the framework so it will take me a bit to get up to speed.

th0ma7 (commented):
Nice catch.  Fix sounds appropriate.

mreid-tt:
> Nice catch. Fix sounds appropriate.

Actually, from testing, that was not the fix at all. The main issue is that when this was introduced in https://github.com/SynoCommunity/spksrc/pull/5570, there was the presumption that the `call_func` accepted a second parameter for `${INST_VARIABLES}`. This was not correct since in the installer for DSM5 it accepted no second parameter and the DSM6/7 versions of the installer, these accepted the second argument as the log file. This is why we were getting all sorts of file errors for the `installer-variables` file because it was attempting to use it as a log file. This corrects the framework such that the arguments passed are handled correctly.

hgy59:
@mreid-tt thanks for exploring this...

mreid-tt:
> @mreid-tt thanks for exploring this...

@hgy59, happy to take the time to figure this out. There have been a few packages which have been merged since #5570 was introduced. I don't know if we'll have to re-publish these packages to remove the bug from the repository. I'll hold off on merging any other of my PRs until this one is reviewed and merged.

hgy59 (approved):
LGTM 
_(BTW I will be offline for about a week)_

mreid-tt:
> LGTM _(BTW I will be offline for about a week)_

Thanks for the review. As discussed in #5587, I've moved the framework changes to #5602 for approval.

spk/radarr/src/service-setup.sh
```diff
@@ -24,43 +25,43 @@ service_postinst ()
     echo "Set update required"
     # Make Radarr do an update check on start to avoid possible Radarr
     # downgrade when synocommunity package is updated
-    touch "${RADARR_CONFIG_DIR}/update_required" 2>&1
+    touch ${RADARR_CONFIG_DIR}/update_required 2>&1
 
-    if [ "${SYNOPKG_DSM_VERSION_MAJOR}" -lt 7 ]; then
+    if [ ${SYNOPKG_DSM_VERSION_MAJOR} -lt 7 ]; then
```
th0ma7, line 30: BTW, keeping `"<var>"` in if statement changes the error msg (although exit status remains the same).

Using `"<var>"`:
```
./toto.sh: 3: [: Illegal number: 
```
without `""`:
```
./toto.sh: 3: [: -lt: unexpected operator
```

th0ma7 (approved):
Besides the to be or not to be `""`, LGTM.

spk/radarr/src/service-setup.sh
hgy59: BTW `SYNOPKG_DSM_VERSION_MAJOR` will never be empty, since we have a safeguard at the beginning of all generated service-setup files in the packages.

```
### Generic variables and functions
### -------------------------------

if [ -z "${SYNOPKG_PKGNAME}" ] || [ -z "${SYNOPKG_DSM_VERSION_MAJOR}" ]; then
  echo "Error: Environment variables are not set." 1>&2;
  echo "Please run me using synopkg instead. Example: \"synopkg start [packagename]\"" 1>&2;
  exit 1
fi
```

hgy59: and quoting in other palces will depend on whether DSM installation directories might ever contain spaces....

mreid-tt:
> Besides the to be or not to be `""`, LGTM.

Thanks for the review. I decided to roll-back my changes for the `""` because on looking more into the syntax checking this is to _prevent globbing and word splitting_ ([SC2086](https://www.shellcheck.net/wiki/SC2086)). Now ideally we should but the contents of these variables are manually set and shouldn't contain _spaces, line feeds, glob characters and such_.

The other reason is that I wanted the `service-setup.sh` to be as identical to other `*arr` packages so that comparison and keeping them in sync would be easier. Changing them all to adhere to [SC2086](https://www.shellcheck.net/wiki/SC2086) would have been a chore and not something that we currently apply consistently to the rest of the framework. Maybe this can be revisited in the future.

publicarray:
> not something that we currently apply consistently 

Just be careful as there may be places where adding or removing `"` can break things in subtle ways

mreid-tt:
@th0ma7, @publicarray any chance you guys could assist in getting this published?

th0ma7:
@mreid-tt https://github.com/th0ma7/spksrc/actions/runs/4188757062
I'll enable them online once completed.

th0ma7:
sorry for taking so long but finally enabled online, cheers!

mreid-tt:
@th0ma7, thanks for this. looking at the website https://synocommunity.com/package/radarr many of the DSM 7.x packages seem to be missing. i hope this is a temporary gap while the server updates its cache?

th0ma7:
I just checked and all now looks ok

PR #5594, opened 2023-02-06
Update SABnzbd to 3.7.2

Safihre:
I wonder if we can get rid of the whole DSM 6 -> DSM 7 upgrade path?
It requires users that have a non-default `download_dir` to change it every time during the upgarde.
We don't do these things for packages like `nzbget` either..

But maybe that can be changed for the next major release.


mreid-tt (approved):

mreid-tt:
Installed and tested successfully on DSM7.1.

mreid-tt:
Thanks for publishing this. I can confirm via re-downloading the original test files in https://github.com/sabnzbd/sabnzbd/pull/2401 that this release fixed my issue. There was a bug identified in the framework that was patched after you published this package (https://github.com/SynoCommunity/spksrc/pull/5602). It would result in errors in the installation log:
```
2023/02/14 04:38:34	Begin load_variables_from_file
/var/packages/sabnzbd/scripts/installer: line 42: /var/packages/sabnzbd/etc/installer-variables: Permission denied
2023/02/14 04:38:34	End load_variables_from_file
```
These can be ignored as they will go away next time the package is built (for v3.7.3 for example).

I did however also notice these in the log file:

```
2023/02/14 04:38:34	Begin service_postupgrade
2023/02/14 04:38:35	/var/packages/sabnzbd/scripts/service-setup: line 112: shopt: -d: invalid option
2023/02/14 04:38:35	shopt: usage: shopt [-pqsu] [-o] [optname ...]
2023/02/14 04:38:35	End service_postupgrade
```

Perhaps this can be looked into in the future to see what is causing this.

PR #5225, opened 2022-04-22
synocli-devel: New package - debugging, target python wheel building

th0ma7:
## Description

New package to help debugging including gdb, binutils and LLVM

## Checklist

- [x] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5600, opened 2023-02-11
Misc updates (fribidi, glib, glibmm, pango, libsigc++)

th0ma7:
## Description

Misc updates (fribidi, glib, glibmm, pango, libsigc++):
- This PR contains a change to the framework exporting `TC_GCC`, `TC_GLIBC` and `TC_KERNEL` variables.  In turns this finally allows using `DEPENDS =` based on versions of said key toolchain libraries
- This PR is a first attempt at creating entirely generic cross dependencies that fall back to sub-cross dependencies:
```
glib
├── glib-latest (TC_KERNEL >= 3.10 && TC_GGG >= 7.5)
└── glib-2.66

glibmm
├── glibmm-latest (TC_GCC >= 7.5)
└── glibmm-2.66

libsigc++
├── libsigc++-latest (TC_GCC >= 7.5 => requires c++17)
└── libsigc++-2.12 (requires c++11)
```

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


th0ma7:
@hgy59 I know you're off for the week but I believe you will much like this proposed change.  The slight framework change finally allows using `$(TC_GCC)`, `$(TC_GLIBC)` and `$(TC_KERNEL)` natively.

Using this, we can now play with various versions using dependencies based on needed compiler & kernel.  This should make our life way way much easier to maintain package in the longer term.

th0ma7:
@hgy59 if you have a moment for commenting on this.  This framework change is a mandatory to get kernel headers right for `strace` part of `synocli-devel` package #5225 on top of providing updated glib & all for `gstreamer` part of #5298

hgy59:
I already had an 👁️ on this and would also have proposed to pin an additional glib version. This si a neat update to have several pinned and a latest version of cross package(s).
And I just checked it, whether it does not conflict with #5611 😄 .

hgy59 (approved):
LGTM

th0ma7:
Thnx, much appreciated :)

hgy59:
@th0ma7 sorry to bother you, but I didn't find any reason why glib-latest is restricted to kernels >= 3.10.

```
glib
├── glib-latest (TC_KERNEL >= 3.10 && TC_GGG >= 7.5)
└── glib-2.66
```

Currently glib-latest is version 2.84.2 and in the Makefile only OLD_PPC_ARCHS are defined as unsupported.
And it builds without any issue for all archs except OLD_PPC_ARCHS, ARMv5_ARCHS and x86-5.2 (all toolchains with c11 and c++11 support work).
So what was the reason to build glib-latest only for DSM >= 7 and Kernel >= 3.10?

th0ma7:
> sorry to bother you, but I didn't find any reason why glib-latest is restricted to kernels >= 3.10.

@hgy59 Don't be, always a pleasure :)

> ```
> glib
> ├── glib-latest (TC_KERNEL >= 3.10 && TC_GCC >= 7.5)
> └── glib-2.66
> ```
> 
> Currently glib-latest is version 2.84.2 and in the Makefile only OLD_PPC_ARCHS are defined as unsupported. And it builds without any issue for all archs except OLD_PPC_ARCHS, ARMv5_ARCHS and x86-5.2 (all toolchains with c11 and c++11 support work). So what was the reason to build glib-latest only for DSM >= 7 and Kernel >= 3.10?

`glib` is actually dependent on `glibc` which is turn is dependent on the `kernel` version... not forgetting as well `gcc` version requiring either C99 or C11 depending of the version.  Playing a bit with chatgpt I was able to get this table:

### GLib, glibc, Linux Kernel and GCC Minimums

| GLib version | Release year | Requires glibc ≥ | Implies Linux kernel ≥ | Requires GCC ≥ | Notes |
|--------------|--------------|------------------|-------------------------|----------------|-------|
| **2.50**     | 2016         | 2.16 (2012)      | 2.6.32                  | 4.5            | First C99-only GLib release |
| **2.56**     | 2018         | 2.25 (2017)      | 3.2                     | 4.8            | Needs `clock_gettime()` and POSIX.1-2008 APIs |
| **2.64**     | 2020         | 2.28 (2018)      | 3.2                     | 4.9            | Stronger POSIX.1-2008 reliance |
| **2.66**     | 2020         | 2.28 (2018)      | 3.2                     | 4.9            | DSM 6.x often stops here |
| **2.70**     | 2021         | 2.28 (2018)      | 3.2                     | 5              | Fully Meson-based build system |
| **2.74**     | 2022         | 2.28 (2018)      | 3.2                     | 5              | Requires proper C11 compiler (GCC ≥ 5) |
| **2.80**     | 2024         | 2.28 (2018)      | 3.2                     | 5              | Same baseline, stricter compiler needs |
| **2.84**     | 2025         | 2.28 (2018)      | 3.2                     | 5              | Current stable, no stricter kernel requirement |

One must consider that stable version are even, dev odd.  Looking further, only one arch on dsm7 still uses an older glibc: comcerto2k using a glibc 2.20 and kernel 3.2.101.  So testing against kernel 3.10 instead of glibc in turns affects also other DSM7 archs still using a 3.2 kernel, namely evansport (and armada but hidden behind generic armv7 using kerrnel 3.10).

Looking further, DSM-7.1 mostly uses glibc 2.26, the switch to 2.36 only hapens with DSM-7.2.  With that a proper testing and versioning should then really be:
```
glib
├── glib-latest (TC_GLIBC >= 2.28 && TC_GCC >= 5) = DSM-7.2
├── glib-2.62 (TC_GLIBC >= 2.25 && TC_GCC >= 4.8) = DSM-7.1
├── glib-2.54 (TC_GLIBC >= 2.16 && TC_GCC >= 4.5) >= DSM-5 (gcc-4.5 is pretty much the first version supporting C99)
└── ...
```
So you are right questionning it.  On its own `TC_KERNEL >= 3.10` isn't entirely wrong as it is the minimal DSM-7.2 kernel version, but gcc should have been `TC_GCC >= 8.5` to ensure that.  All in all, both version and tests using glibc should be fixed adequately based on this info.

hgy59:
@th0ma7 thanks for the detailed information.

But why can I build glib 2.84 for all DSM 6 archs (except ARMv5)?
none of those has gcc >= 5.
For example hi3535-6.2.4 has gcc 4.8.3 and glibc 2.19.
Why does it successfully build glib?

I can only guess
- cross/glib-latest does not use features of glibc > 2.19
- we probably use only a subset of glib that bilds well with gcc < 5
-

PR #5477, opened 2022-11-03
deluge: Update to latest libtorrent wheel version 1.2.18 and 2.0.8

th0ma7:
## Description

deluge: Update to latest libtorrent wheel version 1.2.18 and 2.0.8

This PR provides the following:
- Update to deluge libtorrent library
- Fix of boost build pertaining to python
- Fix issue with deluge on DSM7 pertaining to special characters being converted to `.`
- Fix Error: skipping tracker announce (unreachable)

A framework change is also included:
- Addition of new variable `GCC_DEBUG_INFO` that when set to `1` will change the optimization flags in order to include `debug_info` library symbols

Might fix #5452 & #5452 - `debug_info` enabled on DSM6 for follow-up

## Checklist

- [x] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [x] Includes small framework changes
- [x] This change requires a documentation update (e.g. Wiki)

PR #5616, opened 2023-02-21
Transmission: Update to v4.0.2

### LINKED ISSUE #5595: [Package Request] Bump Transmission to 4.0.2
### Software Name

Transmission 4.0.2

### Brief Description

Bump Transmission version from 3.00 to 4.0.2.

### Website

https://transmissionbt.com/

### Documentation

https://github.com/transmission/transmission/blob/main/docs/README.md

https://github.com/transmission/transmission/blob/main/docs/Building-Transmission.md

### Source code

https://github.com/transmission/transmission

### OpenSource License

other (please specify below)

### Other License

https://github.com/transmission/transmission/blob/main/COPYING

mreid-tt:
## Description

1. Update Transmission to version 4.0.2
2. Patch for alternate volume selection

Fixes #5595

## Checklist

### Type of change

- [x] Package update


cross/transmission/Makefile
```diff
@@ -1,26 +1,24 @@
 PKG_NAME = transmission
-PKG_VERS = 3.00
+PKG_VERS = 4.0.0
 PKG_EXT = tar.xz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/transmission/transmission/releases/download/$(PKG_VERS)
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
-DEPENDS = cross/zlib cross/openssl cross/curl cross/libevent
+DEPENDS =
```
hgy59: out of the build log files, it depends on curl

```suggestion
DEPENDS = cross/curl
```

cross/transmission/Makefile
```diff
@@ -5,7 +5,7 @@ PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/transmission/transmission/releases/download/$(PKG_VERS)
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
-DEPENDS = cross/curl cross/libevent
+DEPENDS = cross/curl cross/libevent cross/python38
```
hgy59, line 8: whenever python is required, use at least python310

mreid-tt: noted, but I don't think the issue has anything to do with python. going to try a different approach based on this consistent error across almost all the builds:
```
gmake[6]: *** No rule to make target 'third-party/utp/lib/libutp.a', needed by 'daemon/transmission-daemon'.  Stop.
```

mreid-tt: rolled back to original dependencies.

mreid-tt:
@hgy59, thanks for your comments. I think I'm stuck with this now and need help. To help with things I've rolled back and squashed a number of interim commits to make the PR more understandable to others.

This is the brief summary:

1. Initially did a standard binary and version update
2. Based on your [comment](https://github.com/SynoCommunity/spksrc/issues/5595#issuecomment-1424731289), I borrowed some code from another package to omit older compilers
3. Based on the initial build issues, I removed the previous patch and updated the build method
4. There was a number of errors reporting `No rule to make target 'third-party/utp/lib/libutp.a', needed by 'daemon/transmission-daemon'`
5. To address this a PR from [transmission/transmission#4810](https://github.com/transmission/transmission/pull/4810) was imported as patches which forced the `libutp` library to build as static
6. Now the linker is reporting `ld: warning: libcares.so.2, needed by .../libcurl.so, not found`

From what I can tell the `-rpath` configurations are correct elsewhere in the log so I'm not sure how to address this. It may be possible we may have to wait until Transmission 4.0.1 is released which incorporates the code I imported as well as many other changes to help with successful builds.

mreid-tt:
I took another look at this to try to diagnose the current error. The full text of the errors in Build (x64-7.1) are:

```
/github/workspace/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/../lib/gcc/x86_64-pc-linux-gnu/8.5.0/../../../../x86_64-pc-linux-gnu/bin/ld: warning: libcares.so.2, needed by /github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/libcurl.so, not found (try using -rpath or -rpath-link)
/github/workspace/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/../lib/gcc/x86_64-pc-linux-gnu/8.5.0/../../../../x86_64-pc-linux-gnu/bin/ld: warning: libz.so.1, needed by /github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/libcurl.so, not found (try using -rpath or -rpath-link)
```

In breaking this down it seems to be saying that the GNU linker (`ld`) located in `/github/workspace/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/bin/` is reporting that:
1. `libcurl.so` located in `/github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/` cannot find `libcares.so.2`
2. `libcurl.so` located at the same path cannot find `libz.so.1`

Based on these errors for linking CXX executable subprocess-test and transmission-daemon, the build stops.

Now higher up in the build, these files seem to have already been installed with these log entries:
```
-- Installing: /github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/libz.so.1.2.13
```
and...
```
libtool: install: /usr/bin/install -c .libs/libcares.so.2.5.1 /github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/libcares.so.2.5.1
libtool: install: (cd /github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib && { ln -s -f libcares.so.2.5.1 libcares.so.2 || { rm -f libcares.so.2 && ln -s libcares.so.2.5.1 libcares.so.2; }; })
```

In both cases, the files (or their links) are in the path `/github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/` which is the same path as `libcurl.so`. As such I can't understand why it can't find these files.

hgy59:
@mreid-tt probably we should wait for an updated version (4.0.1?) as there are some early production issues...

some suggestions so far:
- transmission should be able to build in /cross/transmission folder (but the TC_GCC version check does not work, so I prefer the `REQUIRED_MIN_DSM = 7.0` variant).
- and all dependencies should be in cross/transmission/Makefile, not in spk/transmission/Makefile
- BUILD_DEPENDS is for build tools and static libraries only, for dependent shared libraries you must use DEPENDS
- AFAICT it is not necessairy to patch the CMakelist file, as you can overwrite the default settings with
  ```
     CMAKE_ARGS  = -DCMAKE_BUILD_TYPE=RelWithDebInfo
     CMAKE_ARGS += -DENABLE_DAEMON=ON
     CMAKE_ARGS += -DINSTALL_DOC=OFF
     CMAKE_ARGS += -DINSTALL_LIB=ON
     ....
   ```
   et al.
- when there are linker errors with cmake it sometimes helps to define `CMAKE_USE_TOOLCHAIN_FILE = OFF`  (not yet veryfied, since I struggled with libutp - tried to build from bittorent/libutp repo)
- AFAICT the dependency of any python is not required
- when you work on branches and expect to get commits from others on your branch, you must avoid to (rebase and) force push - as it breaks the branches on their clones.
- ...

mreid-tt:
@hgy59, I've also been reviewing the issues reported in the upstream repo (which is how I found the code for the patch). While there are a few tickets (now closed) with build issues, I haven't found a solution yet, so perhaps waiting on 4.0.1 would be best. Alternatively we could open a build issue there but I would struggle to properly describe the environment so I'd need your help if we were to go that route.

> some suggestions so far:
> 
> * transmission should be able to build in /cross/transmission folder (but the TC_GCC version check does not work, so I prefer the `REQUIRED_MIN_DSM = 7.0` variant).

From what I've observed the TC_GCC version check does work in the `cross/transmission` folder but does not in the `spk/transmission` folder (seems to omit all builds). I agree with just setting a required minimum OS, as that has less include files. Will update.

> * and all dependencies should be in cross/transmission/Makefile, not in spk/transmission/Makefile
> * BUILD_DEPENDS is for build tools and static libraries only, for dependent shared libraries you must use DEPENDS

I'm a bit unclear on which type of dependencies to attach, whether they should be `DEPENDS` or `BUILD_DEPENDS`. I didn't get a sense from the [How to build Transmission](https://github.com/transmission/transmission/blob/main/docs/Building-Transmission.md) readme what kind of libraries were required. What I did find however was that if I included Python it would cause that section of the build to fail if you placed it in `cross`. Many other packages include it as a `BUILD_DEPENDS` at the `spk` path and when I did the same it allowed that part of the build to complete. Still learning about dependencies but I'll move the others back to the `cross` path.

> * AFAICT it is not necessairy to patch the CMakelist file, as you can overwrite the default settings with
>   ```
>      CMAKE_ARGS  = -DCMAKE_BUILD_TYPE=RelWithDebInfo
>      CMAKE_ARGS += -DENABLE_DAEMON=ON
>      CMAKE_ARGS += -DINSTALL_DOC=OFF
>      CMAKE_ARGS += -DINSTALL_LIB=ON
>      ....
>   ```
>   et al.

I'll try that but the main issue was that without the patches it was not building the static library `third-party/utp/lib/libutp.a` at all. No errors, just no attempt to do so. I'll do some more research on `CMAKE_ARGS` and see what I can find. Not very familiar with cmake at all and was trying to "wing it".

> * when there are linker errors with cmake it sometimes helps to define `CMAKE_USE_TOOLCHAIN_FILE = OFF`  (not yet veryfied, since I struggled with libutp - tried to build from bittorent/libutp repo)

Hmm, I can try and see what happens. Will need to include in my cmake research that I've yet to finish.

> * AFAICT the dependency of any python is not required

There was one error noted when building the `libevent` portion that flagged Python as missing but including it hasn't resolved the error so I guess it can be excluded. Was only including based on the [How to build Transmission](https://github.com/transmission/transmission/blob/main/docs/Building-Transmission.md) page. 

> * when you work on branches and expect to get commits from others on your branch, you must avoid to (rebase and) force push - as it breaks the branches on their clones.
> * ...

My apologies. I'll stop trying to clean-up until the problems are resolved.

mreid-tt:
@hgy59, I've done a few more tests and here are the results:

1. Tried a series of the `CMAKE_ARGS` however setting a static library build did not work
2. A new release of transmission-4.0.1 came out so no need for patches (hooray!)
3. Tried your suggestion for fixing linker errors but the build failed earlier so I reverted
4. Re-added the `TC_GCC` version check because it was failing for comcerto2k-7.1

All in all we are still stuck at the previous error I examined but with a much cleaner PR. Hopefully this should help with further diagnosis. Thanks again for your assistance.

hgy59:
> 4\. Re-added the `TC_GCC` version check because it was failing for comcerto2k-7.1

good when it works. Otherwise you can define
```
REQUIRED_MIN_DSM = 7.0
UNSUPPORTED_ARCHS = comcerto2k
```

mreid-tt:
@hgy59, so I enabled some more verbose output on the linker and the raw build command used was as follows:

```
/github/workspace/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu-g++ -I/github/workspace/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/usr/include -I/github/workspace/spk/transmission/work-x64-7.1/install//var/packages/transmission/target/include    -O3 -DNDEBUG CMakeFiles/subprocess-test.dir/subprocess-test-program.cc.o -o subprocess-test  -Wl,-rpath,/var/packages/transmission/target/lib:/github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib ../../libtransmission/libtransmission.a -lpthread ../../third-party/libdeflate.bld/pfx/lib/libdeflate.a /github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/libssl.so /github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/libcrypto.so /github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib/libcurl.so ../../third-party/libpsl.bld/pfx/lib/libpsl.a ../../third-party/libnatpmp.bld/pfx/lib/libnatpmp.a ../../third-party/miniupnpc.bld/pfx/lib/libminiupnpc.a ../../third-party/dht.bld/pfx/lib/libdht.a ../../third-party/libutp.bld/libutp.a ../../third-party/libb64.bld/src/libb64.a -lm ../../third-party/jsonsl/libjsonsl.a ../../third-party/wildmat/libwildmat.a ../../third-party/libevent.bld/pfx/lib/libevent.a 
```

The section of interest would be this:

`-Wl,-rpath,/var/packages/transmission/target/lib:/github/workspace/spk/transmission/work-x64-7.1/install/var/packages/transmission/target/lib`

In this we see two paths, the second of which is the explicit install location (discussed above) of the missing files in question. Is there a way to see the state of that directory when the command is being executed? Are the files still there? Are the permissions to access them correct? I really have no more ideas of how to fix this.

mreid-tt:
@hgy59, just as an update I tried building this manually both on Debian and Ubuntu and they both built okay. I also examined the specific line that failed for the linker and the parameters are pretty much the same. In fact the SynoCommunity build basically had additional include directories and more aggressive optimizations paths as the only differences.

I'm wondering now if it's somehow the dependencies which are different. I did note that the SynoCommunity one did seem to have to build a lot more things before compiling the project. This suggests that there were maybe fewer dependencies present? Don't know if you had any further thoughts on resolving my build issue.

mreid-tt:
@hgy59, so a friend suggested I try to use the same environment as the runner-image that is used ([Ubuntu 22.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md)) to build the package. This is in the hopes that I could re-create the fault.

In looking at the prerequisites for [Building Transmission](https://github.com/transmission/transmission/blob/main/docs/Building-Transmission.md) on Ubuntu, only a few of them match exactly (build-essential, automake, autoconf, libtool, pkg-config) as packages available to a fresh install of Ubuntu Server 22.04.2. The other prerequisites did not match exactly and were not installed (libcurl4-openssl-dev, libgtk-3-dev, intltool, libglib2.0-dev, libevent-dev, libminiupnpc-dev, libappindicator3-dev).

Based on this and what I see triggering in `spksrc`, I automated all of the setup in the following script:

```
#!/bin/bash

read -t 5 -p "Preparing to setup the environment..."
echo "..."
cd ~
mkdir -p Downloads
cd Downloads

# install matching packages from Ubuntu 22.04 runner-image
sudo apt install build-essential=12.9ubuntu3
sudo apt install automake=1:1.16.5-1.3
sudo apt install autoconf=2.71-2
sudo apt install libtool=2.4.6-15build2
sudo apt install pkg-config=0.29.2-1ubuntu3

# cmake 3.25.2 not available so install latest
sudo apt install cmake

read -t 5 -p "Preparing to deploy additional packages..."
echo "..."

# download remaining packages from source
wget https://c-ares.org/download/c-ares-1.18.1.tar.gz
tar -xzvf c-ares-1.18.1.tar.gz
cd c-ares-1.18.1
./configure --prefix=/opt/target 
make
sudo make install
cd ..

wget https://www.zlib.net/fossils/zlib-1.2.13.tar.gz
tar -xzvf zlib-1.2.13.tar.gz
cd zlib-1.2.13
./configure --prefix=/opt/target 
make
sudo make install
cd ..

wget https://www.openssl.org/source/openssl-1.1.1t.tar.gz
tar -xzvf openssl-1.1.1t.tar.gz
cd openssl-1.1.1t
./config --prefix=/opt/target --openssldir=/usr/local/ssl
make
make test
sudo make install
cd ..

wget https://curl.se/download/curl-7.87.0.tar.xz
tar -xJvf curl-7.87.0.tar.xz
cd curl-7.87.0
./configure --prefix /opt/target --disable-static --enable-ipv6 --with-openssl=/opt/target --enable-ares=/opt/target --with-ca-bundle=/etc/ssl/certs/ca-certificates.crt --with-ca-path=/usr/share/ca-certificates/mozilla
make
sudo make install
cd ..

read -t 5 -p "Preparing to build app..."
echo "..."

wget https://github.com/transmission/transmission/releases/download/4.0.1/transmission-4.0.1.tar.xz
tar -xJvf transmission-4.0.1.tar.xz
cd transmission-4.0.1
mkdir -p build
cd build
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo .. -DCMAKE_PREFIX_PATH=/opt/target
make
cd ../..
```

Again, I tried to match the build behaviour as close as I could from the logs. Unfortunately, I was not able to reproduce the error and the process built successfully. In trying to better understand I took a look at the `cross/curl` which is the only dependency in the build currently. I don't know if the above suggests something different about the dependencies but I believe I've matched them as closely as possible. Let me know if this helps in any way.

voidfel:
Hello. Any update on when this will be fixed?

mikedld:
Playing with it locally, I see that `--rpath-link` argument isn't being passed to g++ when linking Transmission executables (or zlib library for that matter, which also uses CMake). When manually adjusting the invocation to include the flag, linking succeeds.

It looks like `mk/spksrc.cross-cmake.mk` makes an effort of passing this flag by setting `CMAKE_LD_FLAGS`, but that variable doesn't seem to be used by CMake Correct way would be to set `CMAKE_{STATIC,SHARED,EXE}_LINKER_FLAGS`, which results in a green build when the following patch is applied:

```patch
--- a/mk/spksrc.cross-cmake.mk
+++ b/mk/spksrc.cross-cmake.mk
@@ -54,7 +54,8 @@ endif
     echo 'set(CMAKE_C_FLAGS "$(CFLAGS) $(CMAKE_C_FLAGS) $(ADDITIONAL_CFLAGS)")' ; \
     echo 'set(CMAKE_CPP_FLAGS "$(CPPFLAGS) $(CMAKE_CPP_FLAGS) $(ADDITIONAL_CPPFLAGS)")' ; \
     echo 'set(CMAKE_CXX_FLAGS "$(CXXFLAGS) $(CMAKE_CXX_FLAGS) $(ADDITIONAL_CXXFLAGS)")' ; \
-    echo 'set(CMAKE_LD_FLAGS "$(LDFLAGS) $(CMAKE_LD_FLAGS) $(ADDITIONAL_LDFLAGS)")' ; \
+    echo 'set(CMAKE_EXE_LINKER_FLAGS "$(LDFLAGS) $(CMAKE_EXE_LINKER_FLAGS) $(ADDITIONAL_LDFLAGS)")' ; \
+    echo 'set(CMAKE_SHARED_LINKER_FLAGS "$(LDFLAGS) $(CMAKE_SHARED_LINKER_FLAGS) $(ADDITIONAL_LDFLAGS)")' ; \
     echo
     @echo "# define library rpath" ; \
     echo "set(CMAKE_INSTALL_RPATH $(subst $() $(),:,$(CMAKE_INSTALL_RPATH)))" ; \
```

mreid-tt:
> Playing with it locally, I see that `--rpath-link` argument isn't being passed to g++ when linking Transmission executables (or zlib library for that matter, which also uses CMake). When manually adjusting the invocation to include the flag, linking succeeds.

Much thanks @mikedld, we are definitely further along in the build with your patch with the x64 build successfully completing. The x86 and ARM builds seem to be still failing but it seems to be a new error. Will take a look at it a bit later to see what it may be.

th0ma7:
@mikedld much thnx for finding this!  Now merged to master.  :tada:

mikedld:
The new issue seems to be [cmake/cmake#18840](https://gitlab.kitware.com/cmake/cmake/-/issues/18840). I see that you're building but not running the unit tests, and so not providing the target emulator. If you're not interested in running tests, maybe building them shouldn't be done either, and so you can add `-DENABLE_TESTS=OFF` to `CMAKE_ARGS` to avoid test discovery failure that's happening on non-native archs there.

mreid-tt:
@mikedld, I can't thank you enough for rendering assistance to resolve my build issues. Now onto some package testing...

mreid-tt:
hey @hgy59, I've completed testing of this package on DSM 7.1. It also successfully retains all configurations and functions from an upgrade from v3.00-20. Let me know if we are good to merge.

mreid-tt:
> @mreid-tt, I don't think there was any change in behavior. There are (and were before) `download-dir` and `incomplete-dir` options to configure. The former is the "complete" dir you're looking for.

I think I found the issue. Our UI wizard assumes a parent directory and does not allow sub-directories. I'll see what I can do to make this cleaner for new installs...

Screenshot 2023-03-12 at 1 50 37 PM

mreid-tt:
So I attempted to fix the UI wizard but I think it needs an overhaul. The wizards aren't showing the correct screen on DSM 7.1 (the one above should be for DSM 6) and processing of migrations can be more robust. I also don't like the idea in DSM 6 of users having to create their own folders and permissions; they should have the same experience as with DSM 7. So I'm going to re-work the UI wizards and do some more in-depth testing.

tothf:
I've just seen that there was a [breaking change](https://github.com/transmission/transmission/commit/9c436d87ddac1f4e15797feeb84c5eb9474a4ae6) in Transmission 4.0.1 which has been fixed in 4.0.2. I think this can cause issues with remote clients which is used by many who runs Transmission on Synology.
Maybe 4.0.2 should be the version to bump to instead of 4.0.1.

mreid-tt:
@hgy59, I've revised and simplified the logic in the service setup and successfully tested upgrade migrations. The UI wizard is also much simpler and handles the folder relocation I alluded to above. I realise I spent some time ensuring DSM 6 compatibility but then realise this doesn't build for DSM 6 :)

Anyway it should be good to go but I am also aware of the flag that @tothf raised on v4.0.2. Hopefully it will be out soon and the digests can be updated quickly.

mreid-tt:
hey @hgy59, in light of this https://github.com/SynoCommunity/spksrc/pull/5455#discussion_r1020961644 you made, should we consider having a custom watch dir for transmission from this version?

th0ma7:
If that can help for `deluge` I've forced a new watch dir being `watch-deluge`.  Guessing similar could apply in this case.
https://github.com/SynoCommunity/spksrc/pull/5477/files#diff-f09010c00042a8dd69d09e60cb426f5df14730cdf8c043c373ecb8990450e423

mreid-tt:
> If that can help for `deluge` I've forced a new watch dir being `watch-deluge`. Guessing similar could apply in this case.

Thanks @th0ma7, I've added the change for the installer and upgrade scripts. Initially I considered removing the old watch directory but I decided to leave it in case there were other apps using that path.

mreid-tt:
So a few updates on the package:

1. Based on feedback from @tothf, the upgrade wizard has been removed as moving around directories should not be an objective of an in-place update. This would be consistent with the behaviour of the desktop app as well as many other packages
2. The package has been upgraded to version 4.0.2! With that, the removal of the unit tests from the build as suggested by @mikedld to workaround build failures with version 4.0.1 has been rolled back

The package should now be in its final state and ready for approval.

raphoun:
So there is no way to get the package on DSM 6?

mreid-tt:
> So there is no way to get the package on DSM 6?

Not to my knowledge based on this https://github.com/SynoCommunity/spksrc/issues/5595#issuecomment-1424731289

mreid-tt:
@hgy59, based on the work you are doing in #5595, I've included a fix for installs where the package and the download paths are on different volumes.

I was able to complete a clean install and uninstall as well as a clean install with download dir on a different volume as well as an upgrade install. I think everything is ready to go.

hgy59:
> So there is no way to get the package on DSM 6?

You need to use docker for Transmission 4 on DSM 6 (https://hub.docker.com/r/linuxserver/transmission/tags)
There will never be a regular package for version 4 with DSM 6, since the compiler in the toolchains provided by Synology are too old.
Another solution would be to use an additional layer of abstration, like the servarr team does for the *arr packages. But this is beyond the capabilities of the synocommunity dev. environment.

smaarn (commented):
Not much comments on my side, I'm only not really convinced about the policy of forcing users to use watch, incomplete and complete directories when they already have an existing installation. On my end I found this a little bit annoying, but I may be the only one here ;)

spk/transmission/src/service-setup.sh
```diff
@@ -14,119 +14,55 @@ TRANSMISSION="${SYNOPKG_PKGDEST}/bin/transmission-daemon"
...
+    if [ "${SYNOPKG_PKG_STATUS}" = "INSTALL" ]; then
+        # Capture wizard variable
+        TXN_DNLOAD=${wizard_download_dir:=volume1/downloads}
+        # Check that the path exists, if not use path in package shares
+        if [ ! -d "${TXN_DNLOAD}" ]; then
+            TXN_DNLOAD=$(realpath "/var/packages/${SYNOPKG_PKGNAME}/shares/${wizard_download_share}")
         fi
+        TXN_FUNCTS=("complete" "incomplete" "watch")
+        TXN_FOLDRS=("complete" "incomplete" "watch-transmission")
+        TXN_PATHS=()
```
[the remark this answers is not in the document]
mreid-tt: You are correct that we shouldn't be manipulating existing setups on upgrade. This was something that @tothf was very clear on in his note https://github.com/SynoCommunity/spksrc/issues/5595#issuecomment-1470307364. I actually agreed with this perspective which is why all the upgrade scripts were removed. These folders are only created on first install of a new deployment. As this is wrapped in a `SYNOPKG_PKG_STATUS` check for `INSTALL` this will not fire for upgrades.

spk/transmission/src/service-setup.sh
```diff
@@ -14,119 +14,55 @@ TRANSMISSION="${SYNOPKG_PKGDEST}/bin/transmission-daemon"
...
-                set_syno_permissions "${INCOMPLETE_DIR}" "${GROUP}"
-            fi
-            if [ -n "${WATCH_DIR}" ] && [ -d "${WATCH_DIR}" ]; then
-                set_syno_permissions "${WATCH_DIR}" "${GROUP}"
+        # Edit the configuration according to the wizard
+        sed -e "s|@username@|${wizard_username:=admin}|g" \
+            -e "s|@password@|${wizard_password:=admin}|g" \
+            -i "${CFG_FILE}"
+        i=0
+        while [ $i -lt ${#TXN_FUNCTS[@]} ]; do
```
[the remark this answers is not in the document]
mreid-tt: The difference in the iteration is to keep the folders and the functions separate. Specifically, in the configuration file there is a line for the watch folder for `watch_dir_enabled` which can only work if we reference the function name array rather than the folder name array. Keeping the use of the variables separate for me was a way to avoid errors (i.e. the first loop only references `TXN_FOLDRS` where as the second loop only references `TXN_FUNCTS` and the two are not mixed).

mreid-tt:
> Not much comments on my side, I'm only not really convinced about the policy of forcing users to use watch, incomplete and complete directories when they already have an existing installation. On my end I found this a little bit annoying, but I may be the only one here ;)

See note above which clarifies that the folder structure is only enforced on first install only. Are we good to merge?

spk/transmission/src/service-setup.sh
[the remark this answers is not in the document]
smaarn: I meant using the same kind of loop actually (e.g. here you are using a cursor based loop while for the folders one you used a "for each" loop).

spk/transmission/src/service-setup.sh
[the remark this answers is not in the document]
smaarn: My bad !

Indeed it's fine for a simple install.

spk/transmission/src/service-setup.sh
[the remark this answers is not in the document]
mreid-tt: Oh! No specific reason. I supposed I could change the first loop to be a cursor based loop as well and change the line:

```
folder="$TXN_DNLOAD/$item"
```

to

```
folder="$TXN_DNLOAD/${TXN_FOLDRS[$i]}"
```

Doesn't really matter to me.

cross/transmission/Makefile
```diff
@@ -1,26 +1,20 @@
...
+UNSUPPORTED_ARCHS = comcerto2k
 
-ifeq ($(findstring $(ARCH), ppc824x ppc853x),$(ARCH))
-ifneq (5,$(firstword $(sort $(TCVERSION) 5)))
-# toolchains before version 5 require the old linux quota version
-# related issue: https://github.com/transmission/transmission/issues/591
-ADDITIONAL_CFLAGS = -D_LINUX_QUOTA_VERSION=1
-endif
-endif
+CMAKE_ARGS  = -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
```
th0ma7, line 18: This cannot just be `Release`, you actually need the deb info files as while?

spk/transmission/Makefile
```diff
@@ -11,15 +11,16 @@ DESCRIPTION_FRE = Transmission est un client BitTorrent simple et rapide. Vous p
 DESCRIPTION_SPN = Transmission es un cliente BitTorrent simple y rápido. Puedes controlarlo remotamente mediante su interfaz web o alguna aplicación dedicada.
 STARTABLE = yes
 DISPLAY_NAME = Transmission
-CHANGELOG = "1. Update openssl to 1.1.1n.<br/>2. Deploy dedicated package for DSM7 on comcerto2k (DS414j)."
+CHANGELOG = "1. Update Transmission to v4.0.2.<br/>2. Update openssl to 1.1.1t."
```
th0ma7, line 14: I noticed openssl being remove from `DEPENDS` above in `cross/transmission/Makefile`

th0ma7 (approved):
Besides the two comments, overall LGTM

cross/transmission/Makefile
mreid-tt: I don't quite understand what you mean. The `CMAKE_ARGS` was taken directly from [How to Build Transmission](https://github.com/transmission/transmission/blob/main/docs/Building-Transmission.md). Apologies but I'm not too familiar with `cmake`, is there a specific change you are proposing?

spk/transmission/Makefile
mreid-tt: You are correct. I do think that `openssl` is automatically pulled in with `curl` but I can remove it from the change log since it's not explicitly referenced.

cross/transmission/Makefile
th0ma7: Normally we build with CMAKE using `Release` type by default.  But indeed documentation here says otherwise so guessing it's all good.  Note that by it's name I'm guessing it suggest doing a build and preparing a `debian` sub-directory to be used to create a `.deb` package afterwards or from the get go, something which clearly doesn't apply when creating a Synology `.spk` package.

spk/transmission/Makefile
th0ma7: You are right, it is part of curl.  To me options are to add it as a dependency and therefore add a relevant comment, or nothing at all (others may see things differently).

mreid-tt: The change log has been updated to remove mention of `openssl`.

mreid-tt:
Hello @th0ma7, I just wanted to let you know that the PR has been merged successfully. Thank you for taking the time to review and approve it. Could you kindly assist with publishing it? Your help is greatly appreciated.

mreid-tt:
> Now activated online, cheers.

thanks @th0ma7 !

develop7:
@th0ma7 it shows up in synocommunity UI, but doesn't in DSM's (7) "Packages" application (`geminilake` arch). Is there a way to trigger package sources update?

th0ma7:
It usually takes somewhere between 24-48h to show up.  Otherwise you can download and install it manually.

PR #5619, opened 2023-02-23
ownCloud: Update to version 10.12

### LINKED ISSUE #2710: owncloud upgrade path to 9.1
I plan to propose 3 PRs for each owncloud version required to upgrade current 8.1 package: each for major versions: 8.2, 9.0 and 9.1 and expect to get them work properly for both DSM 5.2 and 6.0 (not yet ready for 6.1)

My objective is also to address related issues: #1979 #2409 #2136 #2072

### LINKED ISSUE #3580: OwnCloud Missing
Hello,

Sorry to be a nuisance. I'm running a little DS118. I'm in the Community tab of the Package Center, but ownCloud is missing. Has it been withdrawn?

Thanks.

mreid-tt:
## Description

1. Update ownCloud to version 10.12 (_not upgradeable from v8.1.1-7_)
2. Use SQLite database for faster performance
3. Use shared folder for data directory
4. Framework fix for icon generation

Fixes # 
Closes #2710
Closes #3580

## Checklist

### Type of change

- [x] Package update


hgy59:
@mreid-tt please migrate `installer.sh` to `service-setup.sh`. "installer.sh" is the (very old) version before introduction of the _generic service installer_ and is not supported with DSM 7+ anymore.

spk/owncloud/Makefile
```diff
@@ -1,16 +1,16 @@
 SPK_NAME = owncloud
-SPK_VERS = 8.1.12
-SPK_REV = 8
+SPK_VERS = 10.11
+SPK_REV = 9
 SPK_ICON = src/owncloud.png
 DSM_UI_DIR = app
 
-DEPENDS = cross/busybox cross/$(SPK_NAME)
+DEPENDS = cross/$(SPK_NAME)
```
hgy59, line 7: good to remove cross/busybox dependency! ❤️

spk/owncloud/src/service-setup.sh
```diff
@@ -3,7 +3,7 @@
 PACKAGE_NAME="com.synocommunity.packages.${SYNOPKG_PKGNAME}"
 
 # Others
-INSTALL_DIR="/usr/local/${SYNOPKG_PKGNAME}"
+INSTALL_DIR="/var/packages/${SYNOPKG_PKGNAME}/target"
```
hgy59, line 6: please avoid use of INSTALL_DIR, use `${SYNOPKG_PKGDEST}` instead.

spk/owncloud/src/bin/owncloud-daemon
```diff
@@ -5,13 +5,12 @@ PACKAGE="owncloud"
...
-WEB_DIR="/var/services/web"
-PATH="${INSTALL_DIR}/bin:${PATH}"
+WEB_DIR="/var/services/web_packages"
+PATH="${SYNOPKG_PKGDEST}/bin:${PATH}"
 BUILDNUMBER="$(/bin/get_key_value /etc.defaults/VERSION buildnumber)"
 USER="$([ "${BUILDNUMBER}" -ge "4418" ] && echo -n http || echo -n nobody)"
-OWNCLOUD="${INSTALL_DIR}/bin/owncloud.sh"
-PID_FILE="${INSTALL_DIR}/var/owncloud.pid"
+OWNCLOUD="${SYNOPKG_PKGDEST}/bin/owncloud.sh"
+PID_FILE="${SYNOPKG_PKGDEST}/var/owncloud.pid"
```
hgy59, line 13: and this must be 

```suggestion
PID_FILE="${SYNOPKG_PKGVAR}/owncloud.pid"
```

hgy59: as this is the default, it can be omitted.

mreid-tt:
@hgy59, thanks for the tips while I work on this. So far I've gotten it to install cleanly on DSM7 as well as uninstall (with DB backup). I still have to work on the upgrade logic and update the UI to use modern elements as well as include additional prompts.

One strange thing is that even though I have an icon file defined, the package is not generating any of the icons in the `app/images` folder which doesn't seem to get created. Any thoughts on why this may be?

I also noted from the DSM 7.0 Developer Guide that the Application Config file (at `app/config`) does not recognise properties like `protocol`, `port` or `grantPrivilege` so I've removed these from the file. Instead I've made an addition to the `postinstall` script to amend the Apache configuration file for the same effect.

Thinking to my next steps, I don't know how I'll approach upgrades. The previous versions were based on a MySQL backend but with my performance testing I found the initial setup taking ~20x longer to complete when using MariaDB versus a standalone SQLite database. As such. I've moved to that back end. With the differing backends I don't know that a direct upgrade would be possible. Interested to hear your thoughts.

mreid-tt:
@hgy59, so I took another look at he icons that were missing from `app/images` and found that in the framework file `mk/spksrc.icon.mk`, it did an explicit check for the 'SERVICE_PORT' variable. If this was absent it would not build the icons. This package however is running on the default port 80 so this variable would not be present. I've included a proposal to check for the 'ADMIN_URL' variable as an alternate and if present it would pass. This seems to work well.

As for the upgrade strategy, this version cannot be upgraded from the existing one in the repository (v8.1.1) and would need a minimum of v8.2.11 as starting point. As such, I've included a check for this to block such upgrades. I did however write a fairly elaborate upgrade sequence which includes data and configuration preservation as well as database backup archiving. Tests so far on DSM7 work well for a variety of scenarios including different data paths.

The next thing I think I need to update is the UI wizard files to include the relevant details of what is going on. I'm going to see if I can incorporate some scripting to check for some of the validations included in the `service-setup.sh`. For example ensuring that for a database backup on uninstall that the destination directory is writable. Don't know if that is possible but I'll see what I can find. Once that is done I believe the package should be good to publish.

mreid-tt:
@hgy59, So I've come to a blocker with the DSM 6 testing. A number of the PHP scripts are very picky about the owner of the files matching the owner of the running process. Now this isn't a problem on DSM 7 with the resource worker populating the `web_packages` folder but on DSM 6 (where this needs to be done manually) this is an issue since a lot of the times it runs as root.

Ideally if there were a way to get both the installer and the running process to run as `sc-owncloud` but in a way that still allows the functions to manipulate the files in the `web` folder (which will likely still need root privileges), that would resolve all my issues. Any thoughts on an approach for this?

ymartin59:
I just ask the question: is this worth maintaining DSM 6 support? We may decide to only deliver package for DSM 7.

hgy59:
> @hgy59, So I've come to a blocker with the DSM 6 testing. A number of the PHP scripts are very picky about the owner of the files matching the owner of the running process. Now this isn't a problem on DSM 7 with the resource worker populating the `web_packages` folder but on DSM 6 (where this needs to be done manually) this is an issue since a lot of the times it runs as root.
> 
> Ideally if there were a way to get both the installer and the running process to run as `sc-owncloud` but in a way that still allows the functions to manipulate the files in the `web` folder (which will likely still need root privileges), that would resolve all my issues. Any thoughts on an approach for this?

You always can provide a privilege file for DSM 6 with the same content as the one generated for DSM 7 (i.e. "run-as" : "package")

I highly recommend to migrate DSM 6 packages to use the same privileges (and shared folder handling) for all packages that are fully supported on DSM 7+.

Unfortunately for the upgrade of existing packages this must be done in two steps
In the first step the package must still have root access for the installer, but the service runs as package user. This is to change ownership of all package files for the package user.
If this is done correctly, then future packages can be deployed without any root access.

We currently have installer files for dsm5, dsm6 and dsm7.
For this new approach we will need another installer file for dsm6 without root permissions. This will look similar to the dsm7 installer except for DSM 7 specific features. I.e. all the permission related functions will not be supported.

When those DSM 6 packages use shared folders, it should (or must?) use the DSM 7 version and define `USE_DATA_SHARE_WORKER = yes` (see #5649).

hgy59:
> I just ask the question: is this worth maintaining DSM 6 support? We may decide to only deliver package for DSM 7.

My answer is YES. 

There is a reasonable count of models that are limited to DSM 6.
_(And there will be other models, that will not support the upcoming DSM 7.2)_


Perhaps we should update the DSM 6 package to DSM 6.2.4 instead of 6.1 as soon, Synology will drop support for DSM < 6.2.4 (and remove all packages for unsupported models) see (https://www.synology.com/en-global/products/status/EOL-announcement-for-legacy-dsm).

mreid-tt:
@hgy59, thanks for the suggestion with using a privilege file. Looking at its use in the developer guide I had a thought to perhaps move all the DSM6 specific stuff requiring elevated privileges to a separate 'tool' which would run as root but the rest of the service setup would run as the package. I don't think I did it correctly though. Is this a good solution?

mreid-tt:
I've tried to move the function that I needed run as the package to a command file but regardless of whether I define it as a tool or an executable, the log shows it not running as the package:

```
Console has to be executed with the user that owns the file config/config.php
Current user: root
Owner of config.php: http
Try adding 'sudo -u http ' to the beginning of the command (without the single quotes)
2023/03/11 18:14:50	realpath: 'Console has to be executed with the user that owns the file config/config.php
Current user: root
Owner of config.php: http
Try adding \'sudo -u http \' to the beginning of the command (without the single quotes)': No such file or directory
```

Am I defining it correctly?

https://github.com/SynoCommunity/spksrc/blob/556773e36cab20938f4e8e433a237b6dd5b4147f/spk/owncloud/src/conf/privilege#L32-L35

Am I calling it correctly in my script?

https://github.com/SynoCommunity/spksrc/blob/556773e36cab20938f4e8e433a237b6dd5b4147f/spk/owncloud/src/service-setup.sh#L13

mreid-tt:
I've completed testing for DSM 6. I had to implement a number of workarounds to get the web interface to run as the package user but it should be relatively clean. As the executable mentioned above is still not working, I included a few fail-safes to allow for the command to fail gracefully and not lose user data.

All in all it's a bit hack-y but it should be resilient and clean-up after itself on uninstall. Welcome to feedback from reviewers. Also, if the correct setup for executable with package permissions could be resolved that would be great!

mreid-tt:
@hgy59, I believe I've resolved the issue by using a helper function in the service setup. Testing on DSM 6 and DSM 7 has been successful for all the required checklist items. Ready for final review.

spk/owncloud/src/bin/owncloud-daemon
```diff
@@ -7,26 +7,24 @@ WEB_DIR="/var/services/web_packages"
 # for backwards compatability
 if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 7 ];then
     WEB_DIR="/var/services/web"
+	if [ -z ${SYNOPKG_PKGDEST} ]; then
+		SYNOPKG_PKGDEST="/var/packages/owncloud/target"
+	fi
```
hgy59: SYNOPKG_PKGDEST is always defined, when the service is started by synopkg (as SYNOPKG_DSM_VERSION_MAJOR is).

mreid-tt: I've removed it from the daemon, should it be removed from all the wizard files? I found that sometimes in the interface parameters like SYNOPKG_PKGDEST_VOL weren't always known so I added all of these fail-safes.

mreid-tt: Regarding the environment variables available to the UI wizard I realize they are not the same depending on the platform and stage. So I created a generic UI wizard script to extract the various variables (https://pastebin.com/xDdbTYLV). I was them able to collate the results as follows:

DSM 6.2: https://pastebin.com/n3e5jbRr
DSM 7.1: https://pastebin.com/7nKEC4eN

I believe this can be a useful resource for other devs to know what variables are and are not available when creating UI wizards. I'll audit the variables in use against this and prune what doesn't need to be there.

mreid-tt:
hey @hgy59, in light of this https://github.com/SynoCommunity/spksrc/pull/5455#discussion_r1020961644 you made, should we consider having a custom watch dir for transmission from this version?

mreid-tt:
@ymartin59, reaching out to you as the listed maintainer of ownCloud for your review of this PR. Do let me know if you have any comments or questions.

mreid-tt:
Hey @hgy59, just revisiting some of your earlier comments...

> I highly recommend to migrate DSM 6 packages to use the same privileges (and shared folder handling) for all packages that are fully supported on DSM 7+.

I believe the DSM 6 package should now be consistent with DSM 7 norms. Let me know if I missed anything.

> When those DSM 6 packages use shared folders, it should (or must?) use the DSM 7 version and define `USE_DATA_SHARE_WORKER = yes` (see #5649).

Actually, for this package there is not a shared folder created. The user can either leave their data folder in the package's 'var' folder or specify their own path. However now that I think about it more it may be preferable to create a separate share for owncloud data which would then have a consistent path between DSM 6 and DSM 7. Would you recommend I go in this direction?

hgy59:
> However now that I think about it more it may be preferable to create a separate share for owncloud data which would then have a consistent path between DSM 6 and DSM 7. Would you recommend I go in this direction?

Some more arguments for a shared folder:
- a shared folder is not touched by package updates - the var folder on DSM < 7 is backuped and restored on package updates (by our framework) and that makes no sense for larger data stores (imagine some GB or even TB)
- the shared folder is not deleted on package uninstall (not even with the "remove all package data" option in our DSM 7 uninstall wizard)
- DSM protects the deletion of a shared folder created for a package as long as the package is not uninstalled.

mreid-tt:
@hgy59, excellent points. I'll start a modification for it. I will not be including a volume selector however given that it is broken currently. If the work being done in #5649 is completed soon, I'll look to bring that enhancement into it.

hgy59:
> * DSM protects the deletion of a shared folder created for a package as long as the package is not uninstalled.

I have to update this statement:

- DSM protects the deletion of a shared folder created for a package as long as the package is running.

If you have a shared folder created for a package and want to delete the shared folder, DSM complains that the package is using the share and you have to stop it. When you stop the package (service), than you can delete the shared folder (you have to confirm that you will not be able to recover any data and enter your password). If you do this, the service cannot be starteted anymore, until you recreate the shared folder and add access rights for the service user.

mreid-tt:
Thanks for the clarification. I've completed the change from user input full path for data directory to the creation of a user selected shared folder. All tests so far have been successful.

mreid-tt:
@hgy59, I was doing some final re-testing this time under DSM 6 and found some potentially strange logs with a clean install:

**/var/log/packages/owncloud.log** 

```
2023/03/18 11:43:19	===> Step postinst. STATUS=INSTALL USER=owncloud GROUP=http SHARE_PATH=owncloud
2023/03/18 11:43:19	Begin save_wizard_variables
2023/03/18 11:43:19	End save_wizard_variables
2023/03/18 11:43:19	Configuring owncloud
2023/03/18 11:43:19	Get Share defualt config failed.
2023/03/18 11:43:19	Skip granting 'http' group permissions on /owncloud as the directory does not reside in '/volumeX'. Set manually if needed.
2023/03/18 11:43:19	Begin service_postinst
```

**/var/log/synopkg.log** 

```
2023/03/18 11:43:10	install owncloud 10.11-9 End /bin/mv -f /volume1/@tmp/pkginstall/conf /var/packages/owncloud/conf ret=[0]
2023/03/18 11:43:18	install owncloud 10.11-9 Begin postinst
2023/03/18 11:45:07	(system) trigger Apache2.4 2.4.54-0017 Begin start-stop-status stop
```

**/var/log/messages** 

```
2023-03-18T11:43:09-04:00 DSM62 synoscgi_SYNO.Core.Package.Installation_1_install[26428]: resource_api.cpp:190 Acquire data-share for owncloud when 0x0001 (done)
2023-03-18T11:43:15-04:00 DSM62 synoscgi_SYNO.Core.Package.Installation_1_install[26566]: resource_api.cpp:190 Acquire service-cfg for owncloud when 0x0001 (done)
2023-03-18T11:43:18-04:00 DSM62 synoscgi_SYNO.Core.Package.Installation_1_install[26566]: resource_api.cpp:190 Acquire data-share for owncloud when 0x0001 (done)
2023-03-18T11:43:19-04:00 DSM62 synoshare: share_default_get.c:347 failed to alloc share for [], errno=0x1200
2023-03-18T11:45:12-04:00 DSM62 synopkgctl: resource_api.cpp:289 Release usr-local-linker for Apache2.4 when 0x0000 (done)
```

Based on the above it seems to be a call to `synoshare` which seems to originate in `spksrc.service.installer.dsm6` around this section:

https://github.com/SynoCommunity/spksrc/blob/af3c312591d5883fb61561134d07187e771cfc9a/mk/spksrc.service.installer.dsm6#L276-L281

What are your views on the above? Is this something I need to diagnose and fix? So far the app seems to be behaving normally so I can't detect any ill effects. This may be because I have a custom method to fix folder permissions post-install. Let me know if you think this would prevent merging this PR.

mreid-tt:
> What are your views on the above? Is this something I need to diagnose and fix? So far the app seems to be behaving normally so I can't detect any ill effects.

@hgy59, with the new ownCloud 10.12 released last week I was able to update the app and perform additional testing. Apparently, the installer was creating a folder at `/owncloud` as a result of how we handle shared folders in DSM 6. This is directly related to the fix you were working on in PR #5649. To resolve this I added the undocumented field `wizard_volume` to the installer UI wizard which resolved the issue. I commented it accordingly so it could be removed once your PR is complete.

mreid-tt:
All testing for upgrading from the previous version 10.11 to version 10.12 has been completed on DSM 6 and DSM 7. Minor improvements have been committed and the package update is now ready for review.

spk/owncloud/Makefile
```diff
@@ -1,44 +1,49 @@
...
+SPK_DEPENDS = "WebStation:PHP7.4:Apache2.4"
+
+REQUIRED_MIN_DSM = 6.0
 
 MAINTAINER = ymartin59
 DESCRIPTION = ownCloud is a personal cloud which runs on your own server and gives you freedom and control over your own data.
-ADMIN_URL = /owncloud/
+CHANGELOG = "1. Update ownCloud to 10.12.<br/>2. Use faster SQLite database.<br.>3. Use shared folder for data directory."
```
smaarn, line 14: I believe there is a typo in your change log.

```suggestion
CHANGELOG = "1. Update ownCloud to 10.12.<br/>2. Use faster SQLite database.<br/>3. Use shared folder for data directory."
```

spk/owncloud/src/app/config
```diff
@@ -1,15 +1,14 @@
...
-            "grantPrivilege": "local"
-        }
-    }
+	".url": {
+		"com.synocommunity.packages.owncloud": {
+			"allUsers": true,
+			"desc": "ownCloud",
+			"icon": "images/owncloud-{0}.png",
+			"port": "443",
+			"protocol": "https",
```
smaarn, line 8: Doesn't this bit depend on local configuration ?

While the http may always work I'm not sure the https would always be working, right ?

spk/owncloud/src/service-setup.sh
```diff
@@ -0,0 +1,507 @@
...
+    return $OCC_EXIT_CODE
+}
+
+service_prestart ()
+{
+    # Replace generic service startup, fork process in background
+    echo "Starting owncloud-daemon at ${SYNOPKG_PKGDEST}/bin" >> ${LOG_FILE}
+    COMMAND="${SYNOPKG_PKGDEST}/bin/owncloud-daemon"
+    ${COMMAND} >> ${LOG_FILE} 2>&1 &
+    echo "$!" > "${PID_FILE}"
```
smaarn, line 74: Why would we need to replace the generic service startup ?
It would seem to me that the command here is pretty simple is it not ? (I may have missed something here)

spk/owncloud/src/service-setup.sh
```diff
@@ -0,0 +1,507 @@
...
+}
+
+service_preinst ()
+{
+    :
+}
```
smaarn, line 80: Could be removed, right ?

spk/owncloud/src/service-setup.sh
```diff
@@ -0,0 +1,507 @@
...
+
+exec_occ() {
+    PHP="/usr/local/bin/php74"
+    OCC="${OCROOT}/occ"
+    OCC_ARGS=()
+    for arg in "$@"; do
+        OCC_ARGS+=("$arg")
+    done
+    COMMAND="${PHP} ${OCC} ${OCC_ARGS[@]}"
```
smaarn, line 57: Just curious, what's the difference with the below

```suggestion
    COMMAND="${PHP} ${OCC} $*"
```

spk/owncloud/src/service-setup.sh
```diff
@@ -0,0 +1,507 @@
...
+    if [ ${SYNOPKG_DSM_VERSION_MAJOR} -lt 7 ]; then
+        OCC_OUTPUT=$(/bin/su "$EFF_USER" -s /bin/sh -c "$COMMAND")
+    else
+        OCC_OUTPUT=$($COMMAND)
+    fi
+    OCC_EXIT_CODE=$?
+    echo "$OCC_OUTPUT"
```
smaarn, line 64: Open question: why are we saving the output in some variable to then echo it ? Comparing to letting the command issue its output ?

spk/owncloud/src/service-setup.sh
```diff
@@ -0,0 +1,507 @@
...
+            ${SED} -i "s|^user = http$|user = ${SUBST_TEXT}|g; s|^listen.owner = http$|listen.owner = ${SUBST_TEXT}|g" "${TEMPDIR}/${WS_TMPL_FILE}"
+            ${MV} ${WS_TMPL_PATH}/${WS_TMPL_FILE} ${WS_TMPL_PATH}/${WS_TMPL_FILE}.bak
+            rsync -aX ${TEMPDIR}/${WS_TMPL_FILE} ${WS_TMPL_PATH}/ 2>&1
+            ${RM} ${TEMPDIR}/${WS_TMPL_FILE}
+            CFG_UPDATE="yes"
+        fi
+        # Check for ownCloud Apache config
+        if [ ! -f "/usr/local/etc/apache24/sites-enabled/owncloud.conf" ]; then
+            echo "Add Apache config for ownCloud"
+            rsync -aX ${SYNOPKG_PKGDEST}/web/owncloud.conf /usr/local/etc/apache24/sites-enabled/ 2>&1
```
smaarn, line 145: I understand that rsync is powerful but how is that relevant compared to a plain old cp ?

smaarn (commented):
A couple of open questions on my side (I may lack of context and am more curious than anything here as the installation procedure handles a **lot** of things !)

Found very interesting the bits about the handling of DSM 6 WebStation configuration integrations which is very enlightening !

spk/owncloud/src/service-setup.sh
mreid-tt: Yes, these can be removed.

spk/owncloud/src/service-setup.sh
mreid-tt: I didn't pick it for any specific reason, just seen `rsync` being used much more in other packages to copy files. In this case the `-aX` will preserve attributes and extended attributes when copying files.

spk/owncloud/src/service-setup.sh
mreid-tt: I was not familiar with that syntax actually. This is certainly a lot simpler and I can look to implement and test.

spk/owncloud/src/service-setup.sh
mreid-tt: No reason, I suppose that within the loop we can just do:

```
$COMMAND
```

I can look into making it simpler.

spk/owncloud/src/service-setup.sh
mreid-tt: Because the service startup is not actually stating anything (it's just calling a dummy cron config), it needed to have something to have a PID file. As such I took the example from our Wiki - [Custom service process startup](https://github.com/SynoCommunity/spksrc/wiki/Service-Support#custom-service-process-startup) to get the daemon to work. If there is a better way to represent a start/stop on a PHP-based website I'm open to it.

spk/owncloud/src/app/config
mreid-tt: This shouldn't be in there at all since in the developer guide there are no attributes for `port` or `protocol` for the config file. I found however that not putting them in causes the browser to go to the DSM port (:5000 or :5001) when using DSM 6. So I put these in to get it to behave. You are right they could also be `80` and `http` but it doesn't seem to make a difference as it does http anyway and I redirect to https in the Apache conf file. I can change it if you feel strongly about it.

spk/owncloud/Makefile
mreid-tt: Yes, you are correct. Fixed in my local copy and will commit once all the other comments are in.

mreid-tt:
> A couple of open questions on my side (I may lack of context and am more curious than anything here as the installation procedure handles a **lot** of things !)
> 
> Found very interesting the bits about the handling of DSM 6 WebStation configuration integrations which is very enlightening !

Thanks so much for your detailed review. Yes, the scripts became quite long trying to handle all the DSM 6 stuff. I also relied on some other code examples to help me in areas where I was not too familiar with shell scripting which explains some of the inefficiencies you flagged.

DSM 6 WebStation was a pain to get configured manually which is why I guess they added that resource worker for web service in DSM 7. I'm glad you found it enlightening. :-)

spk/owncloud/src/app/config
smaarn: TBH I don't have any strong feeling about it, I was only worried about users who wouldn't have configured https on their Synology could see odd behaviour. If it does the trick, fine by me ;)

spk/owncloud/src/service-setup.sh
smaarn: If you want I did something of the kind for `tt-rss` (there is an update daemon script which `really` needs being started to have RSS feeds updated)

spk/owncloud/src/service-setup.sh
smaarn: Which is equivalent to `cp --preserve=mode,ownership,timestamps,xattr` but I give you that it does the trick then :)

spk/owncloud/src/app/config
mreid-tt: Okay, I'll leave as is then. My other motivation was to make it consistent with this part of the Makefile:

https://github.com/SynoCommunity/spksrc/blob/d1d3dfda78b5e170724a9712b22a19e05a0cfead/spk/owncloud/Makefile#L27-L29

spk/owncloud/src/service-setup.sh
```diff
@@ -0,0 +1,487 @@
...
+            ${MV} ${WS_CFG_PATH}/${PHP_CFG_FILE} ${WS_CFG_PATH}/${PHP_CFG_FILE}.bak
+            rsync -aX ${TEMPDIR}/${PHP_CFG_FILE} ${WS_CFG_PATH}/ 2>&1
+            ${RM} ${TEMPDIR}/${PHP_CFG_FILE}
+            CFG_UPDATE="yes"
+        fi
+        # Check for backup PHP template
+        if [ -f "$WS_TMPL_PATH/$WS_TMPL_FILE.bak" ]; then
+            echo "Restore PHP template from backup"
+            ${RM} ${WS_TMPL_PATH}/${WS_TMPL_FILE}
+            ${MV} ${WS_TMPL_PATH}/${WS_TMPL_FILE}.bak ${WS_TMPL_PATH}/${WS_TMPL_FILE}
```
smaarn, line 308: Two questions (may be naive):

1. Since `${MV}` is a shortcut for `mv -f` I suspect the initial rm is redundant, right ?
2. Wouldn't this principle of "restoring" former configuration file be fragile in case of other packages using the same mechanisms ? (thinking about tt-rss here because I would be glad to use the same tricks)

smaarn (commented):
On my side there are only two things:

1. Backup mechanism which I'm not necessarily sure of (overwriting it twice makes the initial backup potentially broken and potential issues if the same principle is reused by other mechanisms)
2. Usage of the single `=` when testing values in shell scripts which is not really inline with existing conventions from what I see. But that may be C/C++ devs PTSD (not sure if DSM supports bash in all its flavours for the supported versions but that's another topic).

Will give a crack at installing it (though it would only validate DSM 7 installations)

smaarn:
One open point: is this really an arch dependent package ? I don't see in the package any binary install, wouldn't it be worth flagging it as an explicit no-arch package ? (in the same manner as for cops for example).

mk/spksrc.icon.mk
```diff
@@ -14,7 +14,7 @@ ICON_COOKIE = $(WORK_DIR)/.$(COOKIE_PREFIX)icon_done
 # Icons are only needed with "DSM UI configuration" (see spksrc.service.mk)
 ifneq ($(strip $(DSM_UI_DIR)),)
 ifneq ($(strip $(SPK_ICON)),)
-ifneq ($(strip $(SERVICE_PORT)),)
+ifneq ($(or $(strip $(SERVICE_PORT)), $(strip $(ADMIN_URL))),)
```
th0ma7, line 17: Curious, what's this for?

spk/owncloud/src/bin/owncloud-daemon
```diff
@@ -0,0 +1,32 @@
...
+# Main loop
+while true; do
+    # Update
+    echo "Updating..."
+    ${OWNCLOUD} >> "${HOME_DIR}/daemon.log" 2>&1
+
+    # Wait
+    echo "Waiting ${SLEEP_TIME} seconds..."
+    sleep ${SLEEP_TIME}
+done
```
th0ma7, line 32: Just a thought, there is a reference to `crond.service` in the user manual... wonder if that couldn't be replaced by using it?  Although nowhere I see how to actually make usage of it... 

However I did found something interesting here (duno if it is worth investing time?) https://www.reddit.com/r/synology/comments/10iokvu/description_of_synocrond_tasks/?onetap_auto=true

spk/owncloud/src/wizard/upgrade_uifile.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
th0ma7, line 1: For all the upgrade & install uifile, as there is a lot of code would it make sense to extract it from the file so at `PRE_STRIP_TARGET` or some other point you concatenate them?  That would simplify the overall maintenance aftewards.

th0ma7 (approved):
LGTM

spk/owncloud/src/service-setup.sh
mreid-tt: Thanks for your feedback. You are correct that the `${RM}` may be redundant but I guess I was just trying to be safe. As for the config restoration, you are correct in that it may be fragile. I'm going to try to make it a bit more robust.

mreid-tt:
> One open point: is this really an arch dependent package ? I don't see in the package any binary install, wouldn't it be worth flagging it as an explicit no-arch package ? (in the same manner as for cops for example).

You are correct, I don't believe this package compiles anything or is compiled beforehand since it seems to be a bunch of PHP scripts. I'll take a look at making a no-arch version.

mk/spksrc.icon.mk
mreid-tt: This is so that the icons are built for this package since there is no specific `SERVICE_PORT`. Didn't know if there was a better way to do this. Thoughts?

spk/owncloud/src/bin/owncloud-daemon
mreid-tt: Hmm, I'm not sure trying to setup anything in `crond` is worthwhile since no real action is required for the website to be visible. I was looking at the cops package that @smaarn mentioned and there doesn't seem to be any daemon at all. Maybe if I just remove all of it?

smaarn: Actually the package to look at would be `tt-rss` if looking for a daemon service to run.

spk/owncloud/src/service-setup.sh
smaarn: At worse you could leave that to the next (there is already a lot of work in this MR).

Personally I would only have shown a message "post install" to indicate that some settings were changed and leave those untouched when uninstalling (apart from the PHP profile removal, that is).

smaarn (approved):
Only remaining remarks are really nitpicking IMHO 

1. making it a no-arch could have some added value I believe.
2. considering using a custom service command to avoid having to handle the PID logic thingy could be nice

spk/owncloud/src/service-setup.sh
mreid-tt: @smaarn, I tried using the `SERVICE_COMMAND` however the package failed to start (even when I removed the while loop in the daemon). In comparing with `tt-rss`, the daemon file there included the `--daemon` option as part of the execution. Unfortunately, the OCC utility does not have such an option. I've since reverted to the original code so that the package start/stop can function.

spk/owncloud/src/service-setup.sh
mreid-tt: I've changed the restoration of the configuration explicitly change the new string back to the old string using `sed` so it doesn't rely on the presence of a backup file.

spk/owncloud/src/wizard/upgrade_uifile.sh
mreid-tt: Could you expand on this? Are you suggesting that I move the code which may be common into a separate file and import it using something like `source`? If so, not all the code is the same and I've tried to minimise what's in each file. If not, I'm not very familiar with `PRE_STRIP_TARGET` so I may have misinterpreted your meaning.

spk/owncloud/src/bin/owncloud-daemon
mreid-tt: Similar to my other comment on `SERVICE_COMMAND`, I looked at `tt-rss`, and the daemon file there includes a `--daemon` option as part of the execution. Unfortunately, the OCC utility does not have such an option which is why I suspect the original author used this while loop to keep the execution running. Can't think of another way to get a PID file so I'll leave it as is for now.

mreid-tt:
> Only remaining remarks are really nitpicking IMHO
> 
> 1. making it a no-arch could have some added value I believe.
> 2. considering using a custom service command to avoid having to handle the PID logic thingy could be nice

Thanks for the feedback. For these points:

1. The no-arch config has been added and seems to work well (tested with DSM 6).
2. The `SERVICE_COMMAND` didn't work as the OCC utility does not have a `--daemon` or similar option.

mreid-tt:
> On my side there are only two things:
> 
> 1. Backup mechanism which I'm not necessarily sure of (overwriting it twice makes the initial backup potentially broken and potential issues if the same principle is reused by other mechanisms)
> 2. Usage of the single `=` when testing values in shell scripts which is not really inline with existing conventions from what I see. But that may be C/C++ devs PTSD (not sure if DSM supports bash in all its flavours for the supported versions but that's another topic).
> 
> Will give a crack at installing it (though it would only validate DSM 7 installations)

Thanks for the feedback. For your points:

1. Your comment on the backup and restore function being fragile due to potential overwriting was noted. It has been fixed with an explicit replacement using `sed`.
2. My understanding from https://www.shellcheck.net/wiki/SC2272 is that comparisons in shell should use a single `=` with the comparison statement in square braces `[]`.

mreid-tt:
Hello @th0ma7, I just wanted to let you know that the PR has been merged successfully. Thank you for taking the time to review and approve it. Could you kindly assist with publishing it? Your help is greatly appreciated.

th0ma7:
@mreid-tt I kicked off the build + publish.  I may be able to follow-up tomorrow or the day after to enable online. https://github.com/th0ma7/spksrc/actions/runs/4538303401

th0ma7:
Well that was fast (probably due to being noarch now).
Build + upload + enablement done.
cheers!

th0ma7:
@hgy59 I just noticed that noarch builds uses `7.0-40000` ... Not an issue but unaligned with the 7.1 for per-arch builds.

mreid-tt:
> Build + upload + enablement done.

thanks @th0ma7 !

th0ma7:
@mreid-tt  this is odd as the build process worked out all-right https://github.com/th0ma7/spksrc/actions/runs/4538303401/jobs/7997117592

I ended-up publishing manually this morning.  It's now activated, should appear shortly thereafter.  cheers.

mreid-tt:
> @mreid-tt this is odd as the build process worked out all-right https://github.com/th0ma7/spksrc/actions/runs/4538303401/jobs/7997117592
> 
> I ended-up publishing manually this morning. It's now activated, should appear shortly thereafter. cheers.

@th0ma7, just to close the loop on this odd behaviour, this was because the server gave an internal server error which was not handled (we have since handled it in #5701). The logs confirm this as follows:

[Build (noarch-7.0)](https://github.com/th0ma7/spksrc/actions/runs/4538303401/jobs/7997117592#logs) succeeds!

```
http --verify=no --ignore-stdin --auth ***: POST https://api.synocommunity.com/packages @/github/workspace/spk/owncloud/../../packages/owncloud_noarch-dsm7_10.12-9.spk
  {"package": "owncloud", "version": "10.12-9", "firmware": "7.0-40000", "architectures": ["noarch"]}
```

[Build (noarch-6.0)](https://github.com/th0ma7/spksrc/actions/runs/4538303401/jobs/7997117453#logs) fails

```
http --verify=no --ignore-stdin --auth ***: POST https://api.synocommunity.com/packages @/github/workspace/spk/owncloud/../../packages/owncloud_noarch-all_10.12-9.spk
  <html>
    <head>
      <title>Internal Server Error</title>
    </head>
    <body>
      <h1><p>Internal Server Error</p></h1>
      
    </body>
  </html>
```

The root cause for this I believe we have come to the answer to in https://github.com/SynoCommunity/spkrepo/issues/95 with a fix proposed and in review.

mreid-tt:
hey @publicarray, based on issue #5717, it appears that the publish process seems to have been corrupted with the database reporting an upload which is not present. I would appreciate your assistance in deleting version 10.12-9 of ownCloud from the repository. Once done I'll manually re-publish the files one at a time to avoid a race condition.

hgy59:
@mreid-tt I can delete this for you.
shall I delete package for all (i.e. DSM >= 3.1 <6) too?
_(it is not activated and the Makefile declares `REQUIRED_MIN_DSM = 6.0`)_

mreid-tt:
hey @hgy59, thanks much, for now remove all builds under version 10.12-9. The build for DSM 3.x is a bit weird and may need an adjustment in the Makefile on the next update. I believe it may have something to do with this line:

https://github.com/SynoCommunity/spksrc/blob/19ec801005714a15a28bba1dd0e9a025fdef6a1f/spk/owncloud/Makefile#L41-L42

spk/owncloud/src/wizard/upgrade_uifile.sh
mreid-tt: This I believe is finally resolved in https://github.com/SynoCommunity/spksrc/pull/5912 with the implementation of wizard templates.

spk/owncloud/src/bin/owncloud-daemon
mreid-tt: This has been removed in https://github.com/SynoCommunity/spksrc/pull/5912 and is more aligned to the webservice example shown in https://github.com/SynoCommunity/spksrc/tree/master/spk/demowebservice.

mk/spksrc.icon.mk
mreid-tt: This was subsequently enhanced in https://github.com/SynoCommunity/spksrc/pull/5649#issuecomment-1751843885.

Issue #5421, opened 2022-09-09
ffmpeg download not working


vespino:
### Is this a new Bug?

- [X] I checkd that the bug hasn't been reported before

### Package Name

ffmpeg

### Package Version

4.4.2-46

### Device Model

DS1821+

### Device Architecture

x86_64

### Firmware Version

7.1.1-42962

### What happened?

I use ffmpeg to download tv show streams. On DSM 7.1-42661 this used to work fine, but recently I upgraded to 7.1.1-42962 and files downloaded won't play on Kodi.

### Reproduction steps

```bash
/volume1/@appstore/ffmpeg/bin/ffmpeg -y -i "https://cdn-rtlvod-h8.akamaized.net/ba232bae-2341-4e1d-a8c7-2908f7620cab/8fc709e3-4fdd-4bf3-967e-2ac75e8062a2.ism/QualityLevels(5859744)/Manifest(video,format=m3u8-aapl,filter=Quality,encryption=cbc)" -i "https://cdn-rtlvod-h8.akamaized.net/ba232bae-2341-4e1d-a8c7-2908f7620cab/8fc709e3-4fdd-4bf3-967e-2ac75e8062a2.ism/QualityLevels(125368)/Manifest(audioname,format=m3u8-aapl,filter=Quality,encryption=cbc)" -c copy "test2.mp4"

I ran this command on a Synology with DSM 7.1-42661 and on one with 7.1.1-42962. The first does play on Kodi 19.3, the second doesn't.
```

### Install Log

```log
From DSM 7.1-42661
2022/09/09 11:33:03     install ffmpeg 4.4.2-46 Begin preinst
2022/09/09 11:33:03     Begin reload_inst_variables
2022/09/09 11:33:03     End reload_inst_variables
2022/09/09 11:33:03     Begin initialize_variables
2022/09/09 11:33:03     End initialize_variables
2022/09/09 11:33:03     ===> Step preinst. USER= GROUP= SHARE_PATH=
2022/09/09 11:33:03     install ffmpeg 4.4.2-46 End preinst ret=[0]
2022/09/09 11:33:03     install ffmpeg 4.4.2-46 Begin /bin/mkdir -p /volume1/@appstore/ffmpeg -m 755
2022/09/09 11:33:03     install ffmpeg 4.4.2-46 End /bin/mkdir -p /volume1/@appstore/ffmpeg -m 755 ret=[0]
[...]

PR #5620, opened 2023-02-25
[WIP] ffmpeg5 + ffmpeg6 packages

### LINKED ISSUE #5085: how to update FFmpeg 5.0
The official website has been updated with ffmpeg5.0, when will we update it?

th0ma7:
## Description

FFMPEG5 Jellyfin optimized package + FFMPEG6 default

Fixes #5085, #5421

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [x] Package update


th0ma7:
@hgy59 and @mreid-tt  and all, looking for help!

Having issues specifically with x86_64 and i686 builds... I've created an ffmpeg bug issue at https://trac.ffmpeg.org/ticket/10219 with hope to find a solution.

Getting multiple `undefined reference to` only on those two archs :
```
LD	ffmpeg_g
LD	ffprobe_g
libavcodec/libavcodec.so: undefined reference to `ff_h264chroma_init_x86'
libavfilter/libavfilter.so: undefined reference to `ff_spp_init_x86'
libavcodec/libavcodec.so: undefined reference to `ff_ttaencdsp_init_x86'
libswresample/libswresample.so: undefined reference to `swri_audio_convert_init_x86'
...
```
Hints at solving this issue would be much appreciated :shrug:

hgy59:
> Hints at solving this issue would be much appreciated 🤷

Sometime it helps to force cross compilation for i686 (see cross/erlang/Makefile) - just guessing...

nyanmisaka:
Maybe nasm is too old for the new ffmpeg, which passes the configure check but there's no `X86ASM` in the ffmpeg5 build logs. Thus all `ff_*_x86` asm functions are missing at LD stage. And asm was disabled in ARM builds so it's not affected.

nyanmisaka:
Disabling x86 asm can be a stopgap but it causes software decoders and encoders to be extremely slow.

Better to figure out how to re-enable it later. The low-power chips on Synology devices can benefit a lot from this.

th0ma7:
> Disabling x86 asm can be a stopgap but it causes software decoders and encoders to be extremely slow.
> 
> Better to figure out how to re-enable it later. The low-power chips on Synology devices can benefit a lot from this.

@Nyaran thnx for the pointer.  Indeed disabling x86 asm solved the immediate issue.  I've updated the ffmpeg track issue to reflect this in hope to get some help.

nyanmisaka:
I wonder why asm was disabled in arm and ppc?

This is unlikely to be an upstream issue since ffmpeg 5.0 was released more than a year ago.

Maybe `--x86asmexe=/path/to/nasm` helps.

th0ma7:
> I wonder why asm was disabled in arm and ppc?

This pre-date me taking over the maintenance of ffmpeg.  I never touched that portion of the build process.  Now that this isn't working all right for x86 neither, perhaps it's a good time to review it... similarly, asm is only available for x86 for x265 as well.

> Maybe `--x86asmexe=/path/to/nasm` helps.

I tried that and it didn't worked out neither.

nyanmisaka:
NEON asm works for arm processors. Those asm code were written for some DSP and decoders.

https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/arm/Makefile

```
  --disable-armv5te        disable armv5te optimizations
  --disable-armv6          disable armv6 optimizations
  --disable-armv6t2        disable armv6t2 optimizations
  --disable-vfp            disable VFP optimizations
  --disable-neon           disable NEON optimizations
```

th0ma7:
@nyanmisaka I've been playing more & more with ASM on various front and really, it works just well for `aarch64` but it simply isn't called at all for `i686` and `x86_64` even though it normally should...  w e i r d :shrug:

spk/ffmpeg4/Makefile
```diff
@@ -1,8 +1,8 @@
-SPK_NAME = ffmpeg
+SPK_NAME = ffmpeg4
 SPK_VERS = 4.4.3
-SPK_REV = 47
+SPK_REV = 48
```
hgy59, line 3: This is not a good idea. I propose to keep `SPK_NAME = ffmpeg`
Otherwise this is a new package and you can start with SPK_REV = 1
But then all packages depending on ffmpeg (SPK_DEPENDS = "ffmpeg") will not use further updates (and we will need to keep the latest ffmpeg package in the repository).

SPK_NAME should stay ffmpeg, only DISPLAY_NAME may be changed to ffmpeg4 (or ffmpeg 4).

th0ma7:
@hgy59 and @nyanmisaka finally found the build issue... to which we already had a fix for since 4.x era.  It hapens that `configure` script and `ffbuild/common.mak` uses `ARCH` variable which in turns gets polluted by spksrc framework usage of `ARCH` variable... Re-using existing https://github.com/SynoCommunity/spksrc/blob/master/cross/ffmpeg/patches/0001-replace-arch-env-by-ffmpeg_arch.patch solved the matter.

I can now finally move on to next steps :rocket:

spk/ffmpeg4/Makefile
th0ma7: The issue is more complex than that...

The github build scripts also refers to ffmpeg for it to be pre-process and keept prior to building dependent packages in order to re-use existing libraries.  That also needs fixing so a similar behavior is kept for ffmpeg5 and ffmpeg6 which adds complexity as comskip, chromaprint and tvheadend are in various stages of porting over to ffmpeg >= 5.x.

With that in mind, and based on python3 experience, I decided to rename it all to ffmpeg4 along with adjusting dependent packages.  Once all dependent packages are migrated over the new naming scheme I'll then be able to start porting them over to a newer ffmpeg.  Adding to the blur is jellyfin which requires ffmpeg 5.x for it to run smoothly (and to which there isn't any spk dependencies up until now :shrug: ).

Renaming it all to be `ffmpeg<version>` is to me the best route as they will have to be maintained in parallel for a while.  As for `SPK_REV`, I can probably switch it to `1` but on the other hand if gives the idea that `ffmpeg4` version 48 really is the update from `ffmpeg` version 47.  So I'd rather keeping it as-is.

nyanmisaka:
@th0ma7 Nice to hear that. I've seen that patch but didn't understand what it does. We are also updating [patches for 6.0](https://github.com/jellyfin/jellyfin-ffmpeg/pull/223) but it's not finalized yet.

th0ma7:
@nyanmisaka if I may recommend, would be nice to keep patches for older ffmpeg releases.  I'm still using v4 and now v5 versions of the patches.  Pertaining to v6, any insight as of when the patches may be ready?  As I'll probably wait for that before releasing it.

spk/ffmpeg4/Makefile
hgy59: @th0ma7 you can name all folders ffmpeg4 only the SPK_NAME sould stay ffmpeg.
This is similar to the python2 package, where SPK_NAME is still python.

nyanmisaka:
@th0ma7 4.4.x is for the old Jellyfin 10.7, it's best to keep the current patches as they are. Downstream only needs to pull the corresponding version of the patch series from the `debian/patches` folder.

5.1.x will continue to be maintained for a long time since it is LTS and is required by Jellyfin 10.8.

6.x slightly breaks compatibility with Jellyfin 10.8, we will release a new Jellyfin version to adapt it. Before that make the Jellyfin package depend on ffmpeg5.

nyanmisaka:
Just a FYI some FFmpeg options and runtimes used by Jellyfin are not available in spksrc.

```
Intel Compute-Runtime
https://github.com/intel/compute-runtime
# required by OpenCL tone-mapping on Intel iGPU
# build time deps: 
 * libva https://github.com/intel/libva
 * gmmlib https://github.com/intel/gmmlib
 * level-zero https://github.com/oneapi-src/level-zero
 * intel-graphics-compiler https://github.com/intel/intel-graphics-compiler
```

```
--enable-opencl
# required by OpenCL tone-mapping on Intel iGPU
# build time deps: https://github.com/KhronosGroup/OpenCL-Headers
# linked to: https://github.com/KhronosGroup/OpenCL-ICD-Loader
```

```
--enable-libvpl
# [optional] sucessor of the libmfx, used by FFmpeg 6+
# build time deps: https://github.com/oneapi-src/oneVPL
# run time:
 * UHD6xx and older: libmfxhw64.so* https://github.com/Intel-Media-SDK/MediaSDK
 * UHD7xx, Xe and newer: libmfx-gen.so* https://github.com/oneapi-src/oneVPL-intel-gpu
```

```
--enable-ffnvcodec \
--enable-cuda \
--enable-cuda-llvm \
--enable-cuvid \
--enable-nvdec \
--enable-nvenc
# required by hardware transcoding on NVIDIA GPU (jellyfin 10.8 requires n11.0.10.2+)
# build time deps: https://github.com/FFmpeg/nv-codec-headers
```

th0ma7:
@nyanmisaka thnx for sharing this, that will save me quite a bit of time.

From the get go only a fraction of this will be available to all platforms.

Intel:
I already reached the end of the road for gcc<=4.9.* in regards to `intel-media-driver` which is a mandatory requirement to `intel-media-sdk`.  Therefore I doubt that level-zero, IGC or oneVPL will ever build for DSM6 but still worth a try.  Althought that should be relatively feasible for DSM7+ which uses gcc-7.5 and newer.

nVidia:
I have no clue where to start with... additional pointers or examples would be much appreciated.  For instance, does it only need the `nv-codec-headers` for it to enable ffmpeg build options?  Also I see there are other branches available for older driver versions which lead to the need of knowing what version Synology uses in theses case to match things up (I own a few Syno NAS but none using nvidia so can't really test or find out).

nyanmisaka:
> Intel: I already reached the end of the road for gcc<=4.9.* in regards to `intel-media-driver` which is a mandatory requirement to `intel-media-sdk`. Therefore I doubt that level-zero, IGC or oneVPL will ever build for DSM6 but still worth a try. Althought that should be relatively feasible for DSM7+ which uses gcc-7.5 and newer.

https://gcc.gnu.org/projects/cxx-status.html#cxx14

cpp14 is required by the VPL runtime, which is not fully supported in gcc-4.9 but you can still give it a try.

> nVidia: I have no clue where to start with... additional pointers or examples would be much appreciated. For instance, does it only need the `nv-codec-headers` for it to enable ffmpeg build options? Also I see there are other branches available for older driver versions which lead to the need of knowing what version Synology uses in theses case to match things up (I own a few Syno NAS but none using nvidia so can't really test or find out).

Compiling cuda kernel requires cpp14 too, this part is handled by clang instead of the nvcc that comes with cuda toolkit.

So `--enable-cuda-llvm` indicates that clang 3.4+ is required https://clang.llvm.org/cxx_status.html

I'm not sure if clang is available in Syno toolchains. You can lower the priority of support this as not many people use NVIDIA GPUs on Syno. Other than that you just need to make install the `nv-codec-headers` header files.

spk/ffmpeg4/Makefile
```diff
@@ -0,0 +1,36 @@
...
+SPK_ICON = src/ffmpeg.png
+CHANGELOG = "1. Change package directory to ffmpeg4<br/>2. Enable openh264<br/>3. Enable twolane<br/>4. Enable aarch64 asm neon optimization<br/>5. Update dav1d to version 1.1.0<br/>6. Update frei0r to version 2.2.0<br/>7. Update libvpx to version 1.13<br/>8. Update x264 to latest git commit<br/>9. Update SVT-AV1 to version 1.4.1<br/>10. Update libass to version 0.17.1"
+
+CONF_DIR = src/conf
+
+DEPENDS = cross/ffmpeg4
+
+# Enforce using ffmpeg4 directory
+INSTALL_PREFIX = /var/packages/ffmpeg4/target
```
th0ma7, line 12: @hgy59 so I followed-up your advice and kept the `ffmpeg` naming for ffmpeg4.  While builds and everything works out all right it hapens that it ends-up getting installed in `/var/packages/ffmpeg/target`, hence not using the `ffmpeg4` directory name, hence generating an error at usage:
```
$ ./ffmpeg
./ffmpeg: error while loading shared libraries: libavdevice.so.58: cannot open shared object file: No such file or directory
```
The obvious fix is to either use the default directory name of `ffmpeg` or rename to `ffmpeg4`.... 

Or the best outcome, do you know if there a way to enforce the target installation directory name at package install?

hgy59: @th0ma7 first you have to check, whether there is a wrong library rpath in ffmpeg executable
```
# readelf -d /var/packages/ffmpeg/target/bin/ffmpeg

Dynamic section at offset 0x3c408 contains 36 entries:
 0x0000000000000001 (NEEDED)             Shared library: [libavdevice.so.58]
 0x0000000000000001 (NEEDED)             Shared library: [libavfilter.so.7]
 0x0000000000000001 (NEEDED)             Shared library: [libavformat.so.58]
 0x0000000000000001 (NEEDED)             Shared library: [libavcodec.so.58]
 0x0000000000000001 (NEEDED)             Shared library: [libavresample.so.4]
 0x0000000000000001 (NEEDED)             Shared library: [libpostproc.so.55]
 0x0000000000000001 (NEEDED)             Shared library: [libswresample.so.3]
 0x0000000000000001 (NEEDED)             Shared library: [libswscale.so.5]
 0x0000000000000001 (NEEDED)             Shared library: [libavutil.so.56]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x000000000000000f (RPATH)              Library rpath: [/var/packages/ffmpeg/target/lib]
```

If the RPATH is wrong (i.e. uses ffmpeg4 instead of ffmpeg) we probably have an error in the `install_correct_lib_files` target of spksrc.install.mk.

If there is no common solution to fix this, you can use patchelf in a custom POST_STRIP_TARGET in spk/ffmpeg/Makefile.
As ffmpeg is not a service, it is not possible to add LD_LIBRARY_PATH for all use cases. So it is better to patch the ffmpeg executable with the correct RPATH (you can find an example this in spk/dotnet-runtime/Makefile).

hgy59: BTW the solution above is at build time, not at instsallation time.

th0ma7: And that is where the issue is: I'm building ffmpeg4 so the destination directy ends-up being `ffmpeg4` ... But at installation time it gets installed into `ffmpeg` which matches the package name.  So yeah, the rpath is wrong once installed there for sure.

Question is, how can I change the installation directory as well?  I've looked in Syno dev guide and found nothing obvious so far.

hgy59: I don't know what's going wrong with the build, but the install directory (INSTALL_PREFIX et al.) must use `SPK_NAME` that is actually `ffmpeg` and not `ffmpeg4`.

th0ma7: Exacly.  I've looked into the synology guide and there is no way to change that behavior (e.g. point towards another installation directory).  I've also looked into creating a symlink betwen `/var/packages/ffmpeg` and `/var/packages/ffmpeg4` but that won't work for DSM7 as requiring escalated privileges.

As such I've reverted `ffmpeg4` to be using both the `ffmpeg` name and using it's default installation directory name.  This also led me to readjust the calls for reusing libraries within comskip, chromaprint & tvheadend so the ability to easily switch from one version of ffmpeg to the next is maintained while managing the version 4 exception.

Next build should be now OK

th0ma7:
This should now good for merging.  I'll take one last stab at it upon my return and do further testing.

If @hgy59 you can have one last look at it would be nice.  My main fear is at filling up our server online...

nyanmisaka:
@th0ma7 We updated our patches with some tiny fixes. It'd be great if you can pull the them again before merging.
- 5.1: https://github.com/jellyfin/jellyfin-ffmpeg/tree/jellyfin-5.1
- 6.0: https://github.com/jellyfin/jellyfin-ffmpeg/tree/jellyfin

spk/tvheadend/Makefile
```diff
@@ -25,7 +25,7 @@ DESCRIPTION = Tvheadend is a TV streaming server and recorder for Linux, FreeBSD
 DESCRIPTION_FRE = Tvheadend est un serveur de streaming et enregistreur TV pour Linux, FreeBSD et Android prenant en charge DVB-S, DVB-S2, DVB-C, DVB-T, ATSC, ISDB-T, IPTV, SAT IP et HDHomeRun comme sources d\'entrée. Tvheadend offre le streaming HTTP, HTSP et SAT IP.
 DISPLAY_NAME = Tvheadend
 STARTABLE = yes
-CHANGELOG = "1. Update to latest git version d885496 as of November 28th 2022<br/>2. Update openssl to 1.1.1s<br/>3. Updated to use FFMPEG 4.4.3<br/>4. Updated libhdhomerun to version 20221031<br/>5. iptv_auto: Add support for m3u channel-number tag with HDhomeRun<br/>6. Include enhanced curl<br/>7. Use newer libdvbcsa with CPU optimizations<br/>8. Add OSCam support"
+CHANGELOG = "1. Update to latest git version d885496 as of November 28th 2022<br/>2. Update openssl to 1.1.1s<br/>3. Updated to using ffmpeg4 version 4.4.3<br/>4. Updated libhdhomerun to version 20221031<br/>5. iptv_auto: Add support for m3u channel-number tag with HDhomeRun<br/>6. Include enhanced curl<br/>7. Use newer libdvbcsa with CPU optimizations<br/>8. Add OSCam support"
```
hgy59, line 28:
```suggestion
CHANGELOG = "1. Update to latest git version d885496 as of November 28th 2022<br/>2. Update openssl to 1.1.1t<br/>3. Updated to using ffmpeg4 version 4.4.3<br/>4. Updated libhdhomerun to version 20221031<br/>5. iptv_auto: Add support for m3u channel-number tag with HDhomeRun<br/>6. Include enhanced curl<br/>7. Use newer libdvbcsa with CPU optimizations<br/>8. Add OSCam support"
```

th0ma7:
> @nyanmisaka  We updated our patches with some tiny fixes. It'd be great if you can pull the them again before merging.

All patches are now updated.  Finalizing the testing and should be ready for publishing before end of week.

PR #5596, opened 2023-02-10
Dotnet runtime: Update to v6.0.14

mreid-tt:
## Description

This PR contains the following:

1. Update to dotnet-runtime v6.0.14 (security patches)
2. Refinement of unsupported armv7 archs

Follow up of #5301 and #5574
Fixes #5315

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [x] Package update

PR #5623, opened 2023-02-25
avoid dotnet packages for SRM

hgy59:
## Description

This is a follow up to #5596
- SRM armv7 archs do not support dotnet


hgy59:
I have an RT1900ac with a BROADCOM Northstar plus BCM58622 CPU.
dotnet -info gives a coredump, so it is not supported.

I suppose that ipq806x and dakota CPU are not supported either.

PR #5655, opened 2023-03-05
update meson build

hgy59:
- avoid WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.

### Type of change


th0ma7 (approved):
LGTM

PR #5702, opened 2023-04-20
Update git to 2.40.1 and include contrib/subtree module

Fisiu:
## Description

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


hgy59:
@Fisiu is it possible to update git to v2.40.1?
And probably we can add git subtree as requested in #5726?

Fisiu:
git bumped to 2.40.1, subtree included. Not sure if there is macro for `make` however `@$(RUN) make ` works fine.

cross/git/Makefile
```diff
@@ -38,7 +39,13 @@ include ../../mk/spksrc.cross-cc.mk
 git_pre_configure:
 	@$(RUN) autoconf
 
+.PHONY: git_post_compile
+git_post_compile:
+	@$(MSG) Compiling contrib subtree
+	@$(RUN) make -C contrib/subtree/
```
hgy59, line 45: I prefere to use the MAKE variable and to add MAKE_COMPILE_OPTIONS. This is not mandatory here, as there is no compilation but only script processing for making subtree (but might be for other contributions)

```suggestion
	@$(RUN) $(MAKE) -C contrib/subtree/ $(MAKE_COMPILE_OPTIONS)
```

cross/git/Makefile
```diff
@@ -38,7 +39,13 @@ include ../../mk/spksrc.cross-cc.mk
...
+
 .PHONY: git_post_install
 git_post_install:
+	@$(RUN) make -C contrib/subtree/ $(INSTALL_MAKE_OPTIONS)
```
hgy59, line 49: same here
```suggestion
	@$(RUN) $(MAKE) -C contrib/subtree/ $(INSTALL_MAKE_OPTIONS)
```

spk/git/Makefile
```diff
@@ -29,6 +29,9 @@ SPK_COMMANDS += bin/git-shell
 SPK_COMMANDS += bin/git-upload-archive
 SPK_COMMANDS += bin/git-upload-pack
 SPK_COMMANDS += bin/scalar
+
+ifneq ($(findstring $(ARCH),$(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
 SPK_COMMANDS += bin/git-lfs
+endif
```
hgy59, line 35: It is not a problem to add a command that not exist to SPC_COMMANDS, it will be simply ignored. But for the sake of cleanness you are right.

But let's reduce the arch evaluation and include the optional dependency into this (and remove it above)

```suggestion
ifneq ($(findstring $(ARCH),$(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
DEPENDS += cross/git-lfs
SPK_COMMANDS += bin/git-lfs
endif
```

Fisiu:
@hgy59, thanks for suggestions - all applied.

hgy59:
@Fisiu I have two more suggestions

----

There are some non-stripped binaries in the `libexec/git-core/` folder. By adding the following lines to `cross/git/PLIST`, we can strip those and reduce the package size a little bit:
```
bin:libexec/git-core/git
bin:libexec/git-core/git-daemon
bin:libexec/git-core/git-http-backend
bin:libexec/git-core/git-http-fetch
bin:libexec/git-core/git-http-push
bin:libexec/git-core/git-imap-send
bin:libexec/git-core/git-remote-http
bin:libexec/git-core/git-sh-i18n--envsubst
bin:libexec/git-core/git-shell
bin:libexec/git-core/scalar
```

---

Another suggestion is to use hardlinks and remove the `NO_INSTALL_HARDLINKS=1` option from the `INSTALL_MAKE_OPTIONS`. (i.e. to remove the declaration of `INSTALL_MAKE_OPTIONS` in cross/git/Makefile at all, as the remaining value is the default).
And to remove the patch `001-git-less-hardlinks.patch`.
With hardlinks some binary files are installed as links and this shrinks the package size by about 15%.
I have tested git with hardlinks on DSM 6 and 7 and cannot find any issues. But I have no clue why the option to disable hardlinks was added to the early git package (at least since version 1.8.1.5 on 11.03.2013).
This suggestion is not fully approved (and I will have to test it with DSM 5.2 too).
What do you think about this?

Fisiu:
@hgy59 
Changes applied. Indeed, spk file size has been reduced by few MB. I did some initial testing on DSM 7.1 and spk with hardlinks seems to be working as intended.

cross/git/PLIST
```diff
@@ -6,6 +6,16 @@ lnk:bin/git-upload-archive
...
+bin:libexec/git-core/git-http-fetch
+bin:libexec/git-core/git-http-push
+bin:libexec/git-core/git-imap-send
+bin:libexec/git-core/git-remote-http
+bin:libexec/git-core/git-sh-i18n--envsubst
+bin:libexec/git-core/git-shell
+bin:libexec/git-core/scalar
 rsc:libexec/git-core/
 rsc:share/git-core/
+rsc:share/gitweb/
```
hgy59, line 20: @Fisiu is there any benefit of including gitweb in the package?

We intentionally removed this in #5035. It was required by our git-server package that is not maintained for a long time.
As we have a gitea package and Synology provides a git-server package, there is no reason to keep this piece of code in the package.

Fisiu: I copied one line to much from generated PLIST. Should be good now.

PR #5727, opened 2023-05-11
rTorrent: improve session saving support

stickz:
## Description

This commit improves session saving support for rTorrent to prevent unnecessary loss of torrent files. It also increases the session saving interval from 20 minutes to 1 hour to reduce disk i/o usage with thousands of torrents. Session saving is a very intensive task that can take minutes. Torrents fail to save into session by default when added and can be lost if a crash occurs between saving intervals. These two fixes improve the performance and stability of the rTorrent client.

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/spk/rutorrent/src/rtorrent.rc
+++ b/spk/rutorrent/src/rtorrent.rc
@@ -13,3 +13,5 @@ dht.mode.set = auto
 dht.port.set = 6881
 protocol.pex.set= yes
 port_range = 6881-6999
+schedule2 = session_save, 1200, 3600, ((session.save))
+method.set_key = event.download.inserted, 2_save_session, ((d.save_full_session))


spk/rutorrent/src/rtorrent.rc
```diff
@@ -13,3 +13,5 @@ dht.mode.set = auto
 dht.port.set = 6881
 protocol.pex.set= yes
 port_range = 6881-6999
+schedule2 = session_save, 1200, 3600, ((session.save))
+method.set_key = event.download.inserted, 2_save_session, ((d.save_full_session))
```
smaarn, line 17: Open question: this will only be applied on new installation and won't be applied on upgrade. Is this something we want or would we want to try to amend existing installation's configurations on upgrade ?

stickz: Yes, session saving could present challenges for NAS environments. It highly possible to consume more resources writing session files with torrent information to disk than actually seeding torrents. RAID 5 in particular is very suspectable to this issue.

This and the potential to lose torrents is why I would recommend amending existing installs.

PR #5730, opened 2023-05-14
openssl3: update to v3.1.0

hgy59:
## Description

As openssl 1.1.1 will be EOL in a few months, it is time to provide a openssl v3 package for all archs.

_OpenSSL 1.1.1 was released on 11th September 2018, and so it will be considered EOL on 11th September 2023. It will no longer be receiving publicly available security fixes after that date._

Currently all of the spk packages depending on openssl use version 1.1.1.
Shortly I will provide an update of erlang that supports openssl v3.
Since openssl v1.1 and v3 have some differences in the API it will take some time, until all packages and their dependencies are updated (if ever).

## Checklist

### Type of change

- [x] Library update

### DIFF
--- a/cross/openssl3/Makefile
+++ b/cross/openssl3/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = openssl3
-PKG_VERS = 3.0.7
+PKG_VERS = 3.1.0
 PKG_EXT = tar.gz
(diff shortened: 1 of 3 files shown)


Safihre:
Python should already support OpenSSL 3.x, so we could apply it there?

PR #5236, opened 2022-04-26
minio: update to 2022-05-19T18-20-59Z (security update)

fgma:
## Description

Update minio to latest release.

Fixes CVE-2022-24842 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5192, opened 2022-04-03
Upgrade cops 1.0.1 => 1.1.3 + DSM 7

smaarn:
## Description

Two objectives:

1. Bring cops to DSM 7
2. Upgrade cops (while we're at it

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update
- [x] Includes small framework changes

Binaries are available here: https://github.com/smaarn/spksrc/releases/tag/cops-1.1.3-rc-2

PR #5754, opened 2023-05-28
fix github action to build noarch 6.1

hgy59:
## Description

fix for publish of noarch packages (like #5236 
- noarch build for DSM 6.x must be at least version 6.1

### Type of change


hgy59:
This is the fix discussed in https://github.com/SynoCommunity/spksrc/pull/5192#issuecomment-1565725639

PR #5701, opened 2023-04-20
Framework: Add publish validation check

mreid-tt:
## Description

This PR was motivated by inconsistencies observed in the package publish process. It was noted that when the package is sent to the repository, there was no checking to confirm that the package was successfully received and added. This PR seeks to add this functionality.

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5758, opened 2023-05-29
Framework: display the response of package publish

hgy59:
## Description

- Bring back the log of the http response that was removed by "publish validation check (#5701)"

### Type of change


hgy59:
@mreid-tt thanks to the validation check added in #5701 the http status code is displayed in the publish log.
Unfortunately the response is not shown anymore.
This brings the output of the http response back

hgy59:
@mreid-tt after reviewing this piece of code I finally found that everyhing was there - but in the case of an error the $response variable must be shown (and not the uninitialized $output variable)...

mreid-tt (approved):

PR #5734, opened 2023-05-17
cloudflared: update to 2023.5.1

karasevm:
## Description

Update to 2023.5.1
### New Features
- You can now stream your logs from your remote cloudflared to your local terminal with `cloudflared tail <TUNNEL-ID>`. This new feature requires the remote cloudflared to be version 2023.4.1 or higher.
- Migrated to devincarr/quic-go to unblock go 1.20 support as the old lucas-clemente/quic-go will not get go 1.20 support.

### Notices
- Due to the nature of QuickTunnels (https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/do-more-with-tunnels/trycloudflare/) and its intended usage for testing and experiment of Cloudflare Tunnels, starting from 2023.3.2, QuickTunnels only make a single connection to the edge. If users want to use Tunnels in a production environment, they should move to Named Tunnels instead. (https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-guide/remote/#set-up-a-tunnel-remotely-dashboard-setup)

Upstream changelog: https://github.com/cloudflare/cloudflared/blob/master/RELEASE_NOTES

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5759, opened 2023-05-29
cloudflared: fix package

hgy59:
## Description

follow up to #5734
- remove DSM_UI_DIR to avoid creation of unused icon files
- escape () in package description

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/cross/cloudflared/Makefile
+++ b/cross/cloudflared/Makefile
@@ -4,7 +4,7 @@ PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/cloudflare/cloudflared/archive
 PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIR =  $(PKG_NAME)-$(PKG_VERS)
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 BUILD_DEPENDS = native/go
 
--- a/spk/cloudflared/Makefile
+++ b/spk/cloudflared/Makefile
@@ -2,14 +2,13 @@ SPK_NAME = cloudflared
 SPK_VERS = 2023.5.1
 SPK_REV = 8
 SPK_ICON = src/cloudflared.png
-DSM_UI_DIR = app
 
-DEPENDS = cross/$(SPK_NAME)
+DEPENDS = cross/cloudflared
 UNSUPPORTED_ARCHS = $(PPC_ARCHS)
 
 MAINTAINER = karasevm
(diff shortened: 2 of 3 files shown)


hgy59:
@karasevm I wanted to publish #5734, but then I found some issues.

As of the name _cloudflared_ (d like daemon) I found that this package runs a service.
Is cloudflared a service running on port 7844?

If so, we must define the SERVICE_PORT for this package.

hgy59:
@karasevm to document my suggested chanes I added a dedicated commit.

additionally we have to document the used port (7844) on our wiki page, to avoid use of this port by new packages (see https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports).

hgy59:
> 7844 is the outbound destination port. The main idea behind tunnels is that you don't need to receive any incoming connections, which allows it to work in networks with firewalls and NATs.

@karasevm thanks for clarification

> - DNS-DoH proxy that is disabled by default but, if explicitly enabled without custom port specified, will try to use port 53

As port 53 is a privileged one, this needs root to be executed, so keep disabled by default.

karasevm:
@hgy59 The token is just a base64 encoded JSON with account number, tunnel UUID, and tunnel secret. I'm not sure if the length varies, in my case it is always 185 chars. I suppose setting the `minLength` to 150 and checking that it is valid base64 should cover most cases.

```json
"validator": {
    "minLength": 150,
    "regex": {
        "expr": "/^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=|[A-Za-z0-9+\/]{4})$/",
        "errorText": "Not a valid token."
    }
}
```

hgy59:
@karasevm I see some warnings and errors in cloudflared.log

```
2023-06-01T15:22:04Z WRN The user running cloudflared process has a GID (group ID) that is not within ping_group_range. You might need to add that user to a group within that range, or instead update the range to encompass a group the user is already in by modifying /proc/sys/net/ipv4/ping_group_range. Otherwise cloudflared will not be able to ping this network error="Group ID 207066 is not between ping group 1 to 0"
2023-06-01T15:22:04Z WRN ICMP proxy feature is disabled error="cannot create ICMPv4 proxy: Group ID 207066 is not between ping group 1 to 0 nor ICMPv6 proxy: socket: permission denied"
...
2023/06/01 17:22:04 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 832 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Receive-Buffer-Size for details.
```

As I see those on DSM 6 and DSM 7 I suppose it simply can be ignored.
Or do we need further investigation?

PR #5780, opened 2023-06-13
build a native file to avoid version conflicts

jisakiel:
This makes newer files (i.e. that of bookworm) still allow compiling older file versions.

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update


th0ma7:
Can you further elaborate on the requirement here?  Isn't file command already available within the debian container image?

hgy59:
@jisakiel thanks for contributing.
AFAICR at the time of writing __ it was not possible to build a different version of file, not even by using a native/file of the same version.
But since you show that this is possible, we can now update file to the latest version (i.e. 5,44).

jisakiel:
> Can you further elaborate on the requirement here? Isn't file command already available within the debian container image?

root@9d94521c089f:/spksrc# file --version
file-5.44

I'm trying a bookworm based debian dockerfile I'll share as well.

hgy59:
> I had deliberately done this independently of a bump, so that I could verify it working and it's easier to revert if needed :)

I tried to verify this, but it does not work with file v5.45.
Still get the error:
```
Making all in magic
make[5]: Entering directory '/spksrc/cross/file/work-x64-6.2.4/file-5.45/magic'
Cannot use the installed version of file (5.39) to
cross-compile file 5.45
Please install file 5.45 locally first
```

The requirement _the build environment must have file installed of the same version_ is still mandatory.

It is not enough to have a native file in the path.

hgy59:
@jisakiel I found a solution to use native/file in #5897. It needs a patch to create the magic.mgc file with the native version.

hgy59:
@jisakiel thank you for triggering this. It is now integrated with #5897

PR #5797, opened 2023-07-01
[WIP] bazarr: update 1.2.1 => 1.2.4

smaarn:
## Description

Update bazarr 1.2.1 => 1.2.3

Notes: 

* Python 3.11 is not officially supported yet at the bazarr level (experimental support).
* Python 3.11 can't be supported for old DSMs so if we are to provide with a Python 3.11 version that would be only for "recent" DSMs

## Checklist

### Type of change

- [x] Package update


th0ma7:
Related to greenlet, you can upgrade to version 2.x although you will require:
```
# archs without C++11 support are not supported:

# [greenlet]
ifeq ($(call version_lt, ${TC_GCC}, 5.0),1)
WHEELS_CPPFLAGS = [greenlet] -std=c++11
endif
```
OR you also can provide geenlet 1.x for GCC < 4.9, and `-std=c++11` for GCC < 5.0 and using default for the remaining.

Python 3.10 and 3.11 are due for an update.  I'll update that example accordingly so you can easily reuse as you see fit.

smaarn:
Superseded by https://github.com/SynoCommunity/spksrc/pull/5820

PR #5807, opened 2023-07-09
Fix GitHub prepare action

hgy59:
## Description

This PR was initiated by the fact, that the prepare job of the github build action takes a large amount of time.
Local analysis showed that it takes > 60 seconds to evaluate the dependency-list for each package that depends on ffmpeg.

before, the "Evaluate dependencies" step of the "Prepare for Build" job took about 6 to 10 minutes.
with this fix, it takes about 3 minutes.

## Checklist

### Type of change

### DIFF
--- a/.github/actions/prepare.sh
+++ b/.github/actions/prepare.sh
@@ -37,17 +37,20 @@ done
 for package in ${DEPENDENT_PACKAGES}
 do
     echo "===> Searching for dependent package: ${package}"
-    packages=$(echo "${DEPENDENCY_LIST}" | grep -w "${package}" | grep -o ".*:" | tr ':' ' ' | sort -u | tr '\n' ' ')
+    packages=$(echo "${DEPENDENCY_LIST}" | grep " ${package} " | grep -o ".*:" | tr ':' ' ' | sort -u | tr '\n' ' ')
     echo "===> Found: ${packages}"
     SPK_TO_BUILD+=${packages}
 done
 
 # fix for packages with different names
 if [ "$(echo ${SPK_TO_BUILD} | grep -ow nzbdrone)" != "" ]; then
-    SPK_TO_BUILD+=" sonarr3"
+    SPK_TO_BUILD=$(echo "${SPK_TO_BUILD}" | tr ' ' '\n' | grep -v "nzbdrone" | tr '\n' ' ')" sonarr3"
 fi
 if [ "$(echo ${SPK_TO_BUILD} | grep -ow python)" != "" ]; then
-    SPK_TO_BUILD+=" python2"
+    SPK_TO_BUILD=$(echo "${SPK_TO_BUILD}" | tr ' ' '\n' | grep -v "python" | tr '\n' ' ')" python2"
+fi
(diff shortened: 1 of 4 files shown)


cross/chromaprint/Makefile
```diff
@@ -9,13 +9,13 @@ HOMEPAGE = https://acoustid.org/chromaprint
...
 
-OPTIONAL_DEPENDS = cross/ffmpeg4
+OPTIONAL_DEPENDS = cross/ffmpeg$(FFMPEG_VERSION)
 
 # compiler too old
 UNSUPPORTED_ARCHS = $(ARMv5_ARCHS) $(OLD_PPC_ARCHS)
 
 ifneq ($(wildcard $(FFMPEG_DIR)),)
-CMAKE_RPATH = /var/packages/ffmpeg$(FFMPEG_VERSION)/target/lib
+CMAKE_RPATH = /var/packages/ffmpeg$(subst 4,,$(FFMPEG_VERSION))/target/lib
```
hgy59, line 18: @th0ma7 I suppose without this fix the chromaprint package is not able to find the ffmpeg libs at runtime.

th0ma7: You are totally right, good catch.

.github/actions/prepare.sh
```diff
@@ -37,17 +37,20 @@ done
...
 if [ "$(echo ${SPK_TO_BUILD} | grep -ow nzbdrone)" != "" ]; then
-    SPK_TO_BUILD+=" sonarr3"
+    SPK_TO_BUILD=$(echo "${SPK_TO_BUILD}" | tr ' ' '\n' | grep -v "nzbdrone" | tr '\n' ' ')" sonarr3"
 fi
 if [ "$(echo ${SPK_TO_BUILD} | grep -ow python)" != "" ]; then
-    SPK_TO_BUILD+=" python2"
+    SPK_TO_BUILD=$(echo "${SPK_TO_BUILD}" | tr ' ' '\n' | grep -v "python" | tr '\n' ' ')" python2"
+fi
+if [ "$(echo ${SPK_TO_BUILD} | grep -ow ffmpeg)" != "" ]; then
+    SPK_TO_BUILD=$(echo "${SPK_TO_BUILD}" | tr ' ' '\n' | grep -v "ffmpeg" | tr '\n' ' ')" ffmpeg4"
```
th0ma7, line 53: If I get this right, you're removing all `ffmpeg{456}` from the list but only adding `ffmpeg4` at the end of list?  That would then by default disable `ffmpeg{56}` builds, no?

hgy59: No, this is not the case, it just renames ffmpeg to ffmpeg4. (`-w ffmpeg` is a full word filter and does not match any of ffmpeg4, ffmpeg5 and ffmpeg6).

The error that is fixed with this is the following:
The modification of a file in the spk/ffmpeg4 folder triggers the build of ffmpeg package (not ffmpeg4) and I saw errors saying the folder spk/ffmpeg does not exist.
This custom handling is required for all packages where the package name does not match the spk folder (see lines above for python2 and sonarr3).

th0ma7: Makes full sense, another nice catch :)

PR #5817, opened 2023-07-21
cops: upgrade to 1.4.2

smaarn:
## Description

Original project has gone quiet for some time now (few years).

A fork has been created and maintained since which includes an OOM fix when trying to download "big files" and adds support for latest supported PHP versions.

This MR is therefore aiming for

1. Switching to the fork
2. Upgrading PHP requirement (7.4) to be usable on DSM 7
3. Add support for `.htaccess` migration

Additionally all its wizards are switched to using mustache templates.

## Checklist

### Type of change

- [x] Bug fix
- [x] Package update


spk/cops/src/wizard_templates/install_uifile.yml
```diff
@@ -0,0 +1,13 @@
...
+EXISTING_CALIBRE_DIRECTORY_LABEL: "Calibre directory"
+COPS_CATALOG_TITLE_DESCRIPTION: "Catalog title."
+COPS_CATALOG_TITLE_LABEL: "Catalog title"
+
+COPS_CONFIGURATION_SECOND_STEP_TITLE: "COPS configuration (next)"
+DO_YOU_WANT_TO_USE_COPS_WITH_A_KOBO_DESCRIPTION: "Do you want to use COPS with a Kobo eReader?"
+DO_YOU_WANT_TO_USE_COPS_WITH_A_KOBO_LABEL: "Kobo"
+
+DSM_PERMISSIONS_TITLE: "Attention! DSM Permissions"
+DSM_PERMISSIONS_TEXT: "Permissions are managed with the group <b>http</b> in DSM.<br>The group 'users' is no longer used as of DSM 6.<br>Package user (= sc-cops) will not appear on most UI settings.<br>Including the following:<br>- Application privilege's permission viewer<br>- FPT's chroot user selector<br>- File Stations's<br>- Change owner<br>- Shared Links Manager -> Enable secure sharing<br><br>The only exceptions are:<br>- Control Panel > Shared Folder > Edit > Permission > System internal user<br>- ACL editor<br>"
```
hgy59, line 13: Please provide a link to the permission management wiki page. This is mandatory and avoids discussions as in #5832.

The text _The group 'users' is no longer used as of DSM 6._ and more is related to the DSM5 to DSM 6 migration and is obsolete.

smaarn: Indeed. Fixed.

spk/cops/src/wizard_templates/install_uifile.yml
```diff
@@ -0,0 +1,13 @@
...
+EXISTING_CALIBRE_DIRECTORY_LABEL: "Calibre directory"
+COPS_CATALOG_TITLE_DESCRIPTION: "Catalog title."
+COPS_CATALOG_TITLE_LABEL: "Catalog title"
+
+COPS_CONFIGURATION_SECOND_STEP_TITLE: "COPS configuration (next)"
+DO_YOU_WANT_TO_USE_COPS_WITH_A_KOBO_DESCRIPTION: "Do you want to use COPS with a Kobo eReader?"
+DO_YOU_WANT_TO_USE_COPS_WITH_A_KOBO_LABEL: "Kobo"
+
+DSM_PERMISSIONS_TITLE: "Attention! DSM Permissions"
+DSM_PERMISSIONS_TEXT: "Package user and group will not appear on most UI settings. Please read <a target=\"_blank\" href=\"https://github.com/SynoCommunity/spksrc/wiki/Permission-Management\">Permission Management</a> for details."
```
hgy59, line 13: @smaarn I you wonder, why you get invalid packages with the latest change, you have to qote the `\` and the `"` like this

`target=\\\"_blank\\\"`

to generate a valid json file

```suggestion
DSM_PERMISSIONS_TEXT: "Package user and group will not appear on most UI settings. Please read <a target=\\\"_blank\\\" href=\\\"https://github.com/SynoCommunity/spksrc/wiki/Permission-Management\\\">Permission Management</a> for details."
```

spk/cops/src/wizard_templates/install_uifile_fre.yml
```diff
@@ -0,0 +1,13 @@
...
+EXISTING_CALIBRE_DIRECTORY_LABEL: "Répertoire contenant Calibre"
+COPS_CATALOG_TITLE_DESCRIPTION: "Titre du catalogue."
+COPS_CATALOG_TITLE_LABEL: "Titre du catalogue"
+
+COPS_CONFIGURATION_SECOND_STEP_TITLE: "Configuration de COPS (suite)"
+DO_YOU_WANT_TO_USE_COPS_WITH_A_KOBO_DESCRIPTION: "Voulez vous utiliser COPS avec votre liseuse Kobo ?"
+DO_YOU_WANT_TO_USE_COPS_WITH_A_KOBO_LABEL: "Kobo"
+
+DSM_PERMISSIONS_TITLE: "Attention! DSM Permissions"
+DSM_PERMISSIONS_TEXT: "L'utilisateur et le group du package n'apparaîtront pas sur la plupart des UI réglages. Veuillez vous référrer à <a target=\"_blank\" href=\"https://github.com/SynoCommunity/spksrc/wiki/Permission-Management\">Permission Management</a> pour plus de détails."
```
hgy59, line 13: the same here
```suggestion
DSM_PERMISSIONS_TEXT: "L'utilisateur et le group du package n'apparaîtront pas sur la plupart des UI réglages. Veuillez vous référrer à <a target=\\\"_blank\\\" href=\\\"https://github.com/SynoCommunity/spksrc/wiki/Permission-Management\\\">Permission Management</a> pour plus de détails."
```

hgy59:
@smaarn unfortunately the github lint action cannot not find json syntax errors here, as the wizard files are generated upon build...

It would be great to have a syntax check of the generated wizard files, but it is not worth to include `npm install jsonlint` in the Dockerfile...

spk/cops/src/service-setup.sh
```diff
@@ -5,19 +5,15 @@ PACKAGE_NAME="com.synocommunity.packages.${SYNOPKG_PKGNAME}"
...
+WEB_DIR="/var/services/web_packages"
+SECURITY_SETTINGS_FILE="${WEB_DIR}/${SYNOPKG_PKGNAME}/${SECURITY_SETTINGS_FILE_NAME}"
 CFG_FILE="${WEB_DIR}/${SYNOPKG_PKGNAME}/${CFG_FILE_NAME}"
-BUILDNUMBER="$(/bin/get_key_value /etc.defaults/VERSION buildnumber)"
 
 USER="http"
 GROUP="http"
-PHP_CONFIG_LOCATION="$([ "${BUILDNUMBER}" -ge "7135" ] && echo -n /usr/local/etc/php56/conf.d || echo -n /etc/php/conf.d)"
 
 service_preinst ()
```
hgy59, line 18: @smaarn to show the error message to the user, `service_preinst` must be replaced by `validate_preinst`.

otherwise the installation does not terminate and the error message is shown in the intallation logs only.

So this package requires that the shared folder is created before the installation of cops and the calibre data must already exist in the shared folder.
This is worth to mention in the install wizard too (not only the need of trailing `/`).

A more enhanced wizard would test the existance of the folder and metadata.db in a wizard validation function of the first page....

hgy59:
@smaarn I have now a working version of COPS on DSM 7.2.

The last failure was, that I did not install Apache 2.4. 
Without the package, there was an installation error logged in `/var/log/synopkgmgr.log` with the error `"failed to acquire postinst worker"` for cops.
Can we add `Apache2.4` to `"SPK_DEPENDS"` ?

hgy59:
> @smaarn I have now a working version of COPS on DSM 7.2.
> 
The online viewer is not yet working.
When I click the 👁️ eye icon available for ebup books, I get redirected to a page showing status 500. (the url looks like `/cops/epubreader.php?data=214&db=`)

The same ebook works with `cops 1.1.3-6` on DSM 5.2.

The only error log found is in `/var/log/nginx/error.log` (ip addresses replaced by {client-ip} and {server-ip})
```
2023/08/06 16:01:44 [error] 6170#6170: *11699 open() "/var/services/web/favicon.ico" failed (2: No such file or directory), client: {client-ip}, server: , request: "GET /favicon.ico HTTP/1.1", host: "{server-ip}", referrer: "http://{server-ip}/cops/epubreader.php?data=214&db="
```

smaarn:
Will be checking for adding a json check. I am wondering if a simple jq  processing wouldn't be doing the trick.

spk/cops/src/conf/resource
```diff
@@ -3,13 +3,14 @@
...
       "display_name": "COPS",
+      "icon": "app/images/cops-{0}.png",
       "type": "apache_php",
       "root": "cops",
       "backend": 2,
       "php": {
         "profile_name": "cops",
         "profile_desc": "PHP profile used by cops",
-        "backend": 6,
+        "backend": 11,
```
hgy59, line 13: @smaarn additionally the "zlib" extension is required to view (uncompress) epub books.

hgy59:
> > @smaarn I have now a working version of COPS on DSM 7.2.
> 
> The online viewer is not yet working. When I click the 👁️ eye icon available for ebup books, I get redirected to a page showing status 500. (the url looks like `/cops/epubreader.php?data=214&db=`)
> 
> The same ebook works with `cops 1.1.3-6` on DSM 5.2.
> 
> The only error log found is in `/var/log/nginx/error.log` (ip addresses replaced by {client-ip} and {server-ip})
> 
> ```
> 2023/08/06 16:01:44 [error] 6170#6170: *11699 open() "/var/services/web/favicon.ico" failed (2: No such file or directory), client: {client-ip}, server: , request: "GET /favicon.ico HTTP/1.1", host: "{server-ip}", referrer: "http://{server-ip}/cops/epubreader.php?data=214&db="
> ```

this can be fixed by adding the "zlib" extension for php

_got this error with cops 1.1.3 on DSM 6 without zlib extension
`clsTbsZip ERROR with the zip archive: Unable to uncompress file "META-INF/container.xml" because extension Zlib is not installed.`_

hgy59:
@smaarn can we please downgrade the required php to v7.4 (instead of 8.2) to get compatibility with DSM 7.1 and DSM 6?

smaarn:
> @smaarn can we please downgrade the required php to v7.4 (instead of 8.2) to get compatibility with DSM 7.1 and DSM 6?

@hgy59 unfortunately we don't have much choice here. The forked version actually requires PHP 8.2. Another option would be to do this in two steps:

1. An intermediary version which would be fixing the zlib issue, the installation wizard and OOM (using a patch file)
2. A final version doing the migration to the fork

WDYT ?

hgy59:
> > @smaarn can we please downgrade the required php to v7.4 (instead of 8.2) to get compatibility with DSM 7.1 and DSM 6?
> 
> @hgy59 unfortunately we don't have much choice here. The forked version actually requires PHP 8.2. Another option would be to do this in two steps:
> 
>     1. An intermediary version which would be fixing the zlib issue, the installation wizard and OOM (using a patch file)
> 
>     2. A final version doing the migration to the fork
> 
> 
> WDYT ?

Regarding the [Changelog](https://github.com/mikespub-org/seblucas-cops/blob/main/CHANGELOG.md), php8 is required for cops >= 1.4 (but not for 1.3.6).

So since the last published version of cops is broken on DSM 6, it would be good to have a fixed version (it is not only the missing zlib extension entry).

spk/cops/src/service-setup.sh
smaarn: Had a quick look at the Synology APIs but couldn't find any function which would allow such a thing. We'll be doing that async.

In the meantime I fixed the invalid override.

spk/cops/src/conf/resource
smaarn: That's great ! I must confess I never had tried that feature.

This is now fixed.

smaarn: (Thanks for the hint)

smaarn:
> Regarding the [Changelog](https://github.com/mikespub-org/seblucas-cops/blob/main/CHANGELOG.md), php8 is required for cops >= 1.4 (but not for 1.3.6).

Alas, it seems that the 1.2.0 is already about migrating to PHP 8.0.

> 
> So since the last published version of cops is broken on DSM 6, it would be good to have a fixed version (it is not only the missing zlib extension entry).

That's a fair point. Will be making an intermediary fixed version.

spk/cops/src/service-setup.sh
hgy59: > Had a quick look at the Synology APIs but couldn't find any function which would allow such a thing. We'll be doing that async.
> 

We ca do this later.

BTW:
wizard is only documented in the DSM6 developer guide

a validation function is documented as
```json
{
  "step_title": "Step2",
  "items": [{
    "type": "multiselect",
    "desc": "a check group",
    "subitems": [{
      "key": "check1",
      "desc": "Check button 1"
    }, {
      "key": "check2",
      "desc": "Check button 2",
      "defaultVaule": true,
      "validator": {
        "fn": "{var v=arguments[0]; if (!v) return 'Check this';return true;}"
      }
    }]
  },
```

hgy59:
> > Regarding the [Changelog](https://github.com/mikespub-org/seblucas-cops/blob/main/CHANGELOG.md), php8 is required for cops >= 1.4 (but not for 1.3.6).
> 
> Alas, it seems that the 1.2.0 is already about migrating to PHP 8.0.

Indeed there is no note in the changelog, that PHP 8 is required, there are only entries about PHP 8 compatibility.
cops 1.3.6 runs successfully with PHP 7.4 (changed php backend from 11 to 8 in the resource file and did a fresh install after removing PHP8.2 package).

spk/cops/src/wizard_templates/install_uifile
```diff
@@ -0,0 +1,51 @@
...
+    "items": [
+        {
+            "desc": "{{{PHP_CONFIGURATION_REQUIREMENTS_EXPLANATION}}}"
+        },
+        {
+            "desc": "<ul><li>gd</li><li>intl</li><li>pdo_sqlite</li><li>sqlite3</li><li>zlib</li></ul>"
+        }
+    ]
+},
+{{/IS_DSM_6}}
```
smaarn, line 43: #5819 is allowing this

smaarn:
Bumped to 1.4.2 which is supposed to officially support PHP 7.4

Generated packages here: https://github.com/smaarn/spksrc/releases/tag/cops-1.4.2-rc-1

Will be testing out the newly generated binaries right away.

smaarn:
Fixed installer PHP instructions for DSM 7

New binaries: https://github.com/smaarn/spksrc/releases/tag/cops-1.4.2-rc-2

smaarn:
PTAL @hgy59 I upgraded to 1.4.2 and fixed a few mustache issues (basically DSM 6 and DSM 7 weren't appropriately separated)

hgy59:
@smaarn just a small issue with DSM 6:

the installation logs show:
```

2023/08/27 20:00:55     ===> Step postinst. STATUS=INSTALL USER=http GROUP=http SHARE_PATH=
2023/08/27 20:00:55     Begin save_wizard_variables
2023/08/27 20:00:55     End save_wizard_variables
chown: missing operand after '/var/packages/cops/etc/installer-variables'
Try 'chown --help' for more information.
2023/08/27 20:00:56     Adding '' to 'http'
2023/08/27 20:00:57     Group Name: [http]
2023/08/27 20:00:57     Group Type: [AUTH_LOCAL]
2023/08/27 20:00:57     Group ID:   [1023]
2023/08/27 20:00:57     Group Members:
```

So for DSM < 7 we need `SERVICE_USER` definition, otherwise `sc-cops` is not created, and not available for shared folder permissions.

Please add this to the Makefile (just above the `SERVICE_SETUP = src/service-setup.sh`
```
SERVICE_USER = auto
```

hgy59 (approved):
LGTM now

PR #5800, opened 2023-07-03
Python 3.10.12 and 3.11.4 update

th0ma7:
## Description

Python 3.10.12 and 3.11.4 update

Fixes n/a

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5818, opened 2023-07-22
py310-311: Re-include openssl 1.x until all py-spk migrated over

th0ma7:
## Description

py310-311: Re-include openssl 1.x until all py-spk migrated over

Fixes #5800

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/spk/python310/Makefile
+++ b/spk/python310/Makefile
@@ -13,7 +13,7 @@ DESCRIPTION_FRE = Language de programmation Python.
 DESCRIPTION_SPN = Lenguaje de programación Python.
 STARTABLE = no
 DISPLAY_NAME = Python 3.10
-CHANGELOG = "1. Update to Python 3.10.12<br/>2. Update to OpenSSL 3.1.1<br/>3. Re-include python 1.x until all packages are migrated"
+CHANGELOG = "1. Update to Python 3.10.12<br/>2. Update to OpenSSL 3.1.1<br/>3. Re-include openssl 1.1 until all packages are migrated"
 
 HOMEPAGE = https://www.python.org
 LICENSE  = PSF
--- a/spk/python310/Makefile
+++ b/spk/python310/Makefile
@@ -1,7 +1,7 @@
 SPK_NAME = python310
 SPK_VERS = 3.10.12
 SPK_VERS_MAJOR_MINOR = $(word 1,$(subst ., ,$(SPK_VERS))).$(word 2,$(subst ., ,$(SPK_VERS)))
-SPK_REV = 16
+SPK_REV = 17
 SPK_ICON = src/python3.png
 
 DEPENDS  = cross/$(SPK_NAME)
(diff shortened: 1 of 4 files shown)


spk/python310/Makefile
```diff
@@ -13,7 +13,7 @@ DESCRIPTION_FRE = Language de programmation Python.
 DESCRIPTION_SPN = Lenguaje de programación Python.
 STARTABLE = no
 DISPLAY_NAME = Python 3.10
-CHANGELOG = "1. Update to Python 3.10.12<br/>2. Update to OpenSSL 3.1.1"
+CHANGELOG = "1. Update to Python 3.10.12<br/>2. Update to OpenSSL 3.1.1<br/>3. Re-include python 1.x until all packages are migrated"
```
Safihre, line 16: Typo, not python but openssl.

th0ma7: thnx, fixed.

Safihre (approved):

hgy59:
@th0ma7 I doubt that this migration path will work.
If you build python with openssl 1.1 and 3.x you are not sure whether, when building a package for sickchill, the cryptography wheel will be linked against openssl 1.1 or 3.x.
But you must ensure that the cryptography wheel delivered with the python packages link against openssl 3.x

The problem is, that packages depending on cryptography wheel from installed python, install the wheels once into the venv at installation time.
If we cannot guarantee that the cryptography wheel links against openssl 3.x, we will never be able to get rid of openssl 1.1 libraries.

My suggestion was that this error exists only in packages that do not include cryptography as cross compiled wheel but install it from site-packages of installed python.
But I got the same error in homeassistant like https://github.com/SynoCommunity/spksrc/pull/5800#issuecomment-1646507327 when updating python.
Though this could be fixed by adding `export LD_LIBRARY_PATH=${SYNOPKG_PKGDEST}/lib` to `service-setup` as the homeassistant package includes the cryptography wheel and it's dependencies (i.e. libssl.so.1.1)

If we cannot force cryptography wheel to depend on openssl 3.x I propose to update all packages depending on cryptography wheel now instead of building python packages with openssl 1.1 and 3.x.

mreid-tt:
So, should this PR be reverted?

th0ma7:
@hgy59 you got an excellent point and in fact I am totally in agreement with your last statement: the real fix is to re-release all python packages that has a relationship towards openssl, which is mostly due to cryptography.  Further as we can now easily package newer cryptography versions.

This being said, this PR really only is a "patch" so packages that aren't yet updated continue working (e.g. still able to find appropriate `openssl.so.1.1` file), that's it.  Honestly this only is a temporary fix in order to have something "working" ASAP so online packages do no break than waiting for a full-fledge fix, further as the openssl3 only version was downloaded and already caused issues.

@mreid-tt I don't believe it's needed although I'll create a new PR to enforce using openssl3 only for all python packages who may require it, thus bumping all package versions.

PR #5819, opened 2023-07-22
Add mustache to dev env

smaarn:
## Description

mustache is a very simple template language which could be used to easily generate internationalized wizards.

This is an attempt to integrate it in the dev environment and would be a prerequisite for https://github.com/SynoCommunity/spksrc/pull/5817

Homepage: https://mustache.github.io/

## Checklist

### Type of change


smaarn:
TBH there is this still the documentation which could be added about the "how to use it".

hgy59:
@smaarn I had to break the identation to limit the single RUN command to install packages.

and added zip package for a future removement of native/zip...

@th0ma7, @publicarray, @mreid-tt can you have a second look on the wizard generation from templates?

mk/spksrc.spk.mk
th0ma7, line 1: Nicely done.

hgy59 (approved):
LGTM

hgy59:
@smaarn 
BTW normally I used to add a custom build step to install packages before they are available in the dev env with
```
   sudo apt-get update
   sudo apt-get install ruby-mustache
```

but due to the sudo file error (fixed here) this did not work in spksrc containers on my newer dev machine with debian bookworm.

mreid-tt:
@smaarn, I see the wizards in cops and I like them. I'd like to incorporate in the next update to owncloud. Will the template also work for `install_uifile.sh` files?

PR #5801, opened 2023-07-04
deluge: Update to 2.1.1.dev61 and libtorrent 2.0.9

th0ma7:
## Description

deluge: Update to 2.1.1.dev61 and libtorrent 2.0.9

Depends on #5800

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5820, opened 2023-07-22
Migrate all packages to openssl3 and python 3.11

th0ma7:
## Description

Fixes #5800, #5818

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update

### Package testing (x86_64)
- [x] python310
- [x] python311
- [x] bazarr
- [x] beets
- [x] borgbackup
- [x] deluge  --> Already completed part of #5801
- [x] duplicity
- [x] flexget
- [x] headphones --> Migrated to using "develop" branch https://github.com/rembo10/headphones/issues/3320
- [x] homeassistant (homeassistant v2023.1.7 must stay on Python 3.10)
- [x] mercurial
- [x] octoprint
- [x] plexpy-custom
- [x] rdiff-backup
- [x] rutorrent --> Already migrated,  `cross/libtorrent-rakshasa`  needed migrating
- [x] sabnzbd
- [x] salt-master
- [x] salt-minion
- [x] sickchill --> Already migrated to py311
- [x] znc
- [x] <s>tvheadend</s>   --> Tested and works ok with py311 + openssl 3.1.1 (no cross-compiled python wheel dependency) - Migration to `openssl3` to be completed through subsequent PR including ffmpeg

### Package migrated to `spksrc.python.mk`
- [x] <s>python310</s> --> Not Applicable
- [x] <s>python311</s> --> Not Applicable
- [x] bazarr
- [x] beets
- [x] borgbackup
- [x] deluge
- [x] duplicity
- [x] flexget
- [x] headphones --> Nothing to do
- n/a homeassistant (homeassistant v2023.1.7 must stay on Python 3.10) 
- [x] mercurial
- [x] octoprint
- [x] plexpy-custom --> Nothing to do
- [x] rdiff-backup
- [x] rutorrent --> Nothing to do
- [x] sabnzbd
- [x] salt-master
- [x] salt-minion
- [x] sickchill
- [x] znc
- [x] <s>tvheadend</s> --> Nothing to do

### Build rule `all-supported` completed successfully
- [x] python310
- [x] python311
- [x] bazarr
- [x] beets
- [x] borgbackup
- [x] deluge
- [x] duplicity
- [x] flexget
- [x] headphones --> Nothing to do
- [x] homeassistant
- [x] mercurial
- [x] octoprint
- [x] plexpy-custom --> Nothing to do
- [x] rdiff-backup
- [ ] rutorrent --> Nothing to do
- [x] sabnzbd
- [x] salt-master
- [x] salt-minion
- [x] sickchill
- [x] znc
- [ ] <s>tvheadend</s> --> To be tested through another PR

### PUBLISHING DAY 1
- [ ] bazarr
- [ ] beets
- [ ] borgbackup
- [ ] deluge
- [ ] duplicity
- [ ] flexget
- [ ] headphones
- [ ] mercurial
- [ ] octoprint
- [ ] plexpy-custom --> Nothing to do
- [ ] rdiff-backup
- [ ] sabnzbd
- [ ] salt-master
- [ ] salt-minion
- [ ] sickchill
- [ ] znc

### PUBLISHING DAY 3
- [ ] python311
- [ ] homeassistant
- [ ] rtorrent
- [ ] sabnzbd

### PUBLISHING DAY 5
- [ ] python310

### Note to self
How to get the list of supported language standards:
https://stackoverflow.com/questions/34836775/compiler-standards-support-c11-c14-c17
```
$ gcc -v --help 2> /dev/null | sed -n '/^ *-std=\([^<][^ ]\+\).*/ {s//\1/p}'
...
c++0x
c++11
c++14
c++17
c++1y
c++1z
c++2a
c++98
...
```


hgy59:
@th0ma7 homeassistant can't be updated to python311 without update of homeassistant to at least 2023.5.x.
This is already WIP in #5757.

hgy59:
@th0ma7 octoprint needs to be updated to v1.9.0 at least to work with python311 (I have already a local build with octoprint 1.9.0, but latest is v1.9.2).

th0ma7:
> @th0ma7 homeassistant can't be updated to python311 without update of homeassistant to at least 2023.5.x. This is already WIP in #5757.

Thnx for the heads-up, in the meantime I'll re-map it to py310 but enforce it to use newer `openssl3` version.  We can check afterwards if it's worth publishing or not depending of you WIP PR.

hgy59:
@th0ma7 found that crypography < 40.0 has static binding to openssl and prebuilt wheels of version 38.0.3 are linked to openssl 1.1.

currently trying to build cryptography==38.0.3 with cross/cryptography_38 and cross/openssl3 for homeassistant 2023.1.7.

_just wondering that with cryptography==39.0.0 openssl 1.1 support was dropped:
BACKWARDS INCOMPATIBLE: Support for OpenSSL 1.1.0 has been removed. Users on older version of OpenSSL will need to upgrade._

spk/homeassistant/Makefile
```diff
@@ -5,7 +5,7 @@ SPK_ICON = src/homeassistant.png
 
 HACS_VERS = 1.31.0
 
-SPK_DEPENDS = "python310"
+SPK_DEPENDS = "python310>=3.10.12-18"
```
hgy59, line 8: IMHO this is not required.
I successfully tested homeassistant v2023.1.7-21 with python 3.10.12-16.

th0ma7: I thought this would make a demarcation point on all python packages whereas from this point on they've been tested to work with openssl3.

spk/homeassistant/Makefile
```diff
@@ -83,6 +83,11 @@ ENV += PYO3_CROSS_INCLUDE_DIR=$(STAGING_INSTALL_PREFIX)/include/
 # [cryptography]
 DEPENDS += cross/cryptography_38
 
+# [uamqp]
+# ignore deprecation warnings when building with openssl 3
+# "cc1: all warnings being treated as errors"
+WHEELS_CFLAGS = [uamqp] -Wno-deprecated-declarations
+
```
hgy59, line 90: @th0ma7 unfortunaltely this didn't work 
A solution for uamqp is documented here: https://github.com/Azure/azure-uamqp-python/issues/276#issuecomment-924890556
but it requires to patch the source code (and the patch file isn't available anymore)
I will try to create cross/uampq to build this wheel.

hgy59: It is not possible to build uamqp with DSM 6 toolchains.
Therefore homeassistant 2023.1.7 must  be built with openssl 1.1.1.

th0ma7: I may have found it from the Debian repository... 
https://salsa.debian.org/python-team/packages/azure-uamqp-python/-/blob/debian/1.5.0-1/debian/patches/OpenSSL3.patch

th0ma7:
> headphones has a known issue with python 3.11 I got the same error as in [rembo10/headphones#3320](https://github.com/rembo10/headphones/issues/3320)

@hgy59 Reading through the thread is appears that the `develop` branch has a working fix for it.  So I simply switched to this branch and at first glance things looks like working well.  Can you confirm on your end?

hgy59:
@th0ma7 sorry, I missed that you already fixed the codec2 URL...

hgy59:
@th0ma7 fixed protobuf wheel for duplicity.
The build (github action in your fork) still fails for arch-hi3535-6.2.4 and arch-88f6281-6.2.4.

This is strange since my local build successfully builds all archs (except OLD_PPC_ARCHS) like:
- duplicity_hi3535-6.2.4_1.2.3-10.spk
- duplicity_88f6281-6.2.4_1.2.3-10.spk
- duplicity_88f6281-5.2_1.2.3-10.spk
- 

---

for OLD_PPC_ARCHS even Python 3.11 fails to build with configure warning:
```
Platform "powerpc-none-linux-gnuspe" with compiler "gcc" is not supported by the
CPython core team, see https://peps.python.org/pep-0011/ for more information.
```

th0ma7:
@hgy59 I've hit a really interesting case where `greenlet` >= 2.x would fail using gcc-4.9 at first attempt but work on the second try when resuming the build process?!

It hapends that it had worked for DSM7 using newer gcc and resuming DSM6 build would re-use the pip cache, thus finding the already resulting wheel.

In this specific case, greenlet also needed `-fpermissive` to work correctly on gcc-4.9.

th0ma7:
>  fixed protobuf wheel for duplicity. The build (github action in your fork) still fails for arch-hi3535-6.2.4 and arch-88f6281-6.2.4.

@hgy59 thnx catching that.

Note that with that train of thought I've added a section in the top summary for `build all-supported` to confirm all modified packages do build as expected for all archs.  If you have a few spare cycles feel free to test build a few.  In the meantime I'm working on the last bit (e.g. `salt-*`) in hope of getting it to work (confirmed to be compatible, just missing the howto exacly)...  Much thnx in advance (and if other want to chime in and help much appreciated :smiley: )

hgy59:
> @hgy59 I've hit a really interesting case where `greenlet` >= 2.x would fail using gcc-4.9 at first attempt but work on the second try when resuming the build process?!
> 
> It hapends that it had worked for DSM7 using newer gcc and resuming DSM6 build would re-use the pip cache, thus finding the already resulting wheel.

That was my finding too! 
Finally all archs worked (except qoriq). I thought that greenlet 2.0.1 failed but greenlet 2.0.2 succeeded (but flexget requires greenlet==2.0.1)

BTW I have updated flexget to 3.7.10 too but was delayed by validation whether the transmission integration works, and found that the wheel installation does not work as expected (cross env wheels are not taken from the package but downloaded from index, and pip cache is ignored on updates with DSM 6 due to wrong user [install runs as root but all folders under var are changed to sc-flexget])

hgy59:
> Note that with that train of thought I've added a section in the top summary for `build all-supported` to confirm all modified packages do build as expected for all archs. If you have a few spare cycles feel free to test build a few.

Just popped up that it would be very useful if we could reuse a prevously built python (incl crossenv) when building packages with wheels. Otherwise you need to build python for each package and arch...

hgy59:
ffmpeg et. al. fail to build as nasm.us is currently down (https://updownradar.com/status/nasm.us)

th0ma7:
> > Note that with that train of thought I've added a section in the top summary for `build all-supported` to confirm all modified packages do build as expected for all archs. If you have a few spare cycles feel free to test build a few.
> 
> Just popped up that it would be very useful if we could reuse a prevously built python (incl crossenv) when building packages with wheels. Otherwise you need to build python for each package and arch...

That is something i'd really really much like doind, sinilar to ffmpeg.... That would save so many build cycles.

I'll tentatively give that a shot and see if i can make that to work.  It's the perfect pr for doing that...

th0ma7:
@hgy59 I got the basics to prepare the build environment working using `salt-master` as prototype in order to always re-use `spk/python311` pre-built libraries and `crossenv`.  Currently for an unknown reason it completes the preparation portion then stops.  Although working directory shows that it is all set for it to work.

I'll have more cycles tonight to further look into it, but feel free to dig-in a little in the meantime if you have spare cycles.

mk/spksrc.python.mk
```diff
@@ -0,0 +1,49 @@
...
+ifeq ($(strip $(PYTHON_VERSION)),)
+export PYTHON_VERSION = 311
+endif
+
+# set default spk/python* path to use
+export PYTHON_DIR = $(realpath $(shell pwd)/../python$(PYTHON_VERSION)/work-$(ARCH)-$(TCVERSION)/install/var/packages/python$(PYTHON_VERSION)/target)
+
+# always define SPK_DEPENDS
+SPK_DEPENDS := "python$(PYTHON_VERSION)>=$(shell sed -n 's/^SPK_VERS = \(.*\)/\1/p' $(shell pwd)/../python$(PYTHON_VERSION)/Makefile)-$(shell sed -n 's/^SPK_REV = \(.*\)/\1/p' $(shell pwd)/../python$(PYTHON_VERSION)/Makefile)"
+
```
hgy59, line 18: IMHO SPK_DEPENDS must be defined by the respective package.

This would not work for packages that depend not only on python (like `SPK_DEPENDS = "python311:git"`)
And probably it is not wanted to include the python package revision in the dependency.

mk/spksrc.python.mk
```diff
@@ -0,0 +1,49 @@
...
+ifneq ($(wildcard $(PYTHON_DIR)),)
+export ADDITIONAL_LDFLAGS = -Wl,--rpath-link,$(PYTHON_DIR)/lib -Wl,--rpath,/var/packages/python$(PYHTON_VERSION)/target/lib
+PRE_DEPEND_TARGET = python_pre_depend
+else
+BUILD_DEPENDS += cross/python311
```
hgy59:
```suggestion
BUILD_DEPENDS += cross/python$(PYTHON_VERSION)
```

mk/spksrc.python.mk
```diff
@@ -0,0 +1,49 @@
+###
+### Reuse Python libraries
+###
+
+# Default python version to use
+ifeq ($(strip $(PYTHON_VERSION)),)
+export PYTHON_VERSION = 311
+endif
```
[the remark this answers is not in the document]
hgy59: This will not work, we already have a variable PYTHON_VERSION that is 2.7, 3.8, 3.10 or 3.11.

found this code in spksrc.wheel.mk
```
	   elif [ "$(PYTHON_VERSION)" != "2.7" ] ; then \
```

and this error in the github build log:
```
023-07-25T12:48:07.9263087Z sed: can't read /github/workspace/spk/salt-master/../python3.11/Makefile: No such file or directory
```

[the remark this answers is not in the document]
th0ma7: > This will not work, we already have a variable PYTHON_VERSION that is 2.7, 3.8, 3.10 or 3.11.
> 
> found this code in spksrc.wheel.mk
> ```
> 	   elif [ "$(PYTHON_VERSION)" != "2.7" ] ; then \
> ```
> 
> and this error in the github build log:
> ```
> 023-07-25T12:48:07.9263087Z sed: can't read /github/workspace/spk/salt-master/../python3.11/Makefile: No such file or directory
> ```
> 
> EDIT:
> and this is echoed to `python-cc.mk` in cross/python???/Makefile
> ```
> @echo PYTHON_VERSION=$(PKG_VERS_MAJOR_MINOR) >> $@
> ```
> 

I followed what was used for ffmpeg.  Perhaps then `PYTHON_TARGET` ?

If a specific mk file for python works out I'll reorganize ffmep in a subsequent PR to use the same workflow.

hgy59:
@th0ma7 I have a working version of `spksrc.python.mk`.

just needs some verifications....

hgy59:
verification done.

needed some investigation as the *.so files in the cross compiled wheels have binary differences when reusing prebuilt python.

The difference is caused by different `RPATH` in the so files (i.e. Library Path).
Tested with spk/octoprint:
- regular build: `RPATH` [/var/packages/octoprint/target/lib]
- built with reuse of python311:  `RPATH` [/var/packages/octoprint/target/lib:/var/packages/python311/target/lib]

IMHO this is not an issue that must be fixed since the package folder has precedence over the python311 folder.

hgy59:
> verification done.
> 
> needed some investigation as the *.so files in the cross compiled wheels have binary differences when reusing prebuilt python.
> 
> The difference is caused by different `RPATH` in the so files (i.e. Library Path). Tested with spk/octoprint:
> 
>     * regular build: `RPATH` [/var/packages/octoprint/target/lib]
> 
>     * built with reuse of python311:  `RPATH` [/var/packages/octoprint/target/lib:/var/packages/python311/target/lib]
> 
> 
> IMHO this is not an issue that must be fixed since the package folder has precedence over the python311 folder.

There is a possible source of runtime errors for cross compiled wheels:
Due to the python wheel cache (in distrib/pip/wheels) we must ensure that there is no re-use of wheels built for another package. Or those will fail to find dependent libraries (like libssl.so) due to the wrong RPATH.
As long as only one package is built by github action, this is not an issue.
But for local builds it only works when the pip wheel cache is cleared for every new package built with (cross) wheels.
And when the github build action builds multiple packages with cross compiled wheels, the pip wheel cache must be cleared for each package.

The only solution is to disable the pip wheel cache (we can keep the cache for source files in `distrib/pip/http`)

@th0ma7 any other suggestion?

th0ma7:
First off, great work @hgy59 .  I'll have a look at the details but this was way overdue :smile: 

> There is a possible source of runtime errors for cross compiled wheels: Due to the python wheel cache (in distrib/pip/wheels) we must ensure that there is no re-use of wheels built for another package. Or those will fail to find dependent libraries (like libssl.so) due to the wrong RPATH. As long as only one package is built by github action, this is not an issue. But for local builds it only works when the pip wheel cache is cleared for every new package built with (cross) wheels. And when the github build action builds multiple packages with cross compiled wheels, the pip wheel cache must be cleared for each package.
> 
> The only solution is to disable the pip wheel cache (we can keep the cache for source files in `distrib/pip/http`)
> 
> @th0ma7 any other suggestion?

Actually I do and code change may be in reach: you might recall the pre-download wheel sources PR I had prepared a while back https://github.com/SynoCommunity/spksrc/pull/4971.  It's been ready for merging since more than a year but I was waiting for the python package cleanup to be over before doing that.

Proposed solution is a mix-bag of the following:
1. Download wheel source packages (and thus allow pre-downloading it for github-action to populate and re-use the cache).  Theses will then populate the `distrib` folder (or we may want to populate a `distrib/python-wheel` folder instead)?
2. Move the pip cache into `$(WORK_DIR)` so it becomes per-arch & version dependant.  What value this bring you might ask?  Well when doing a `make wheelclean` you keep it's cache thus expediting subsequent builds.  Also this both solves the `rpath` issue you've identified and the one I noticed with `greenlet` where at the second build it would pick-up the wheel from a similar arch but succesfully built for another's DSM version instead.

Also in sight, opportunities for improvement:
3. When invoking `wheelclean` also clean up the corresponding sub-directory for `cross/*` relatively to the `wheelhouse/requirements-cross.txt` content (if it exists).  For instance this would allow cleaning-up the `cryptography` build folder and it's status files, so at the next make it gets rebuilt resulting in a package that has all wheels (as otherwise any wheels generated from `cross/*` will have been discarded after a `wheelclean`.
4. Add a `make wheeldeepclean` that would also clean-up the local arch pip cache.

@hgy59 what do you think?

mk/spksrc.python.mk
```diff
@@ -0,0 +1,49 @@
...
+
+# set default spk/python* path to use
+export PYTHON_DIR = $(realpath $(shell pwd)/../python$(PYTHON_VERSION)/work-$(ARCH)-$(TCVERSION)/install/var/packages/python$(PYTHON_VERSION)/target)
+
+# always define SPK_DEPENDS
+SPK_DEPENDS := "python$(PYTHON_VERSION)>=$(shell sed -n 's/^SPK_VERS = \(.*\)/\1/p' $(shell pwd)/../python$(PYTHON_VERSION)/Makefile)-$(shell sed -n 's/^SPK_REV = \(.*\)/\1/p' $(shell pwd)/../python$(PYTHON_VERSION)/Makefile)"
+
+ifneq ($(wildcard $(PYTHON_DIR)),)
+export ADDITIONAL_LDFLAGS = -Wl,--rpath-link,$(PYTHON_DIR)/lib -Wl,--rpath,/var/packages/python$(PYHTON_VERSION)/target/lib
```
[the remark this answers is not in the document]
th0ma7: From the look of the code this is now solved.

mk/spksrc.python.mk
th0ma7: I have a different view on this, at least food for thoughts...  I'm doing something in `tvheadend` that sort of solves this and would automate the populating of the SPK_DEPENDS variable value.

Something like the following in `spksrc.python.mk` would ensure python is always included in the SPK_DEPENDS:
```
PYTHON_DEPENDS = python$(PYTHON_VERSION)>=$(shell sed -n 's/^SPK_VERS = \(.*\)/\1/p' $(shell pwd)/../python$(PYTHON_VERSION)/Makefile)-$(shell sed -n 's/^SPK_REV = \(.*\)/\1/p' $(shell pwd)/../python$(PYTHON_VERSION)/Makefile)
ifneq ($(wildcard $(SPK_DEPENDS)),)
SPK_DEPENDS := "$(SPK_DEPENDS):$(PYTHON_DEPENDS)"
else
SPK_DEPENDS = "$(PYTHON_DEPENDS)"
endif
```

th0ma7:
@hgy59 Looking at this I found there are a few corner-cases:
1. Rust based wheels uses `$(STAGING_INSTALL_PREFIX)` when defining python library and include path.  Somehow another variable needs to be used, one that can flip based on building using `spksrc.spk.mk` or `spksrc.python.mk`.
2. To add to the issue, `cross/*` rust based wheels will need to be worked out as well
3. Further, `spksrc.python-wheel.mk` uses a `PYTHONPATH` that probably needs re-adjusting as well automagically.  Potentiall a `spksrc.python-flag.mk` could be use to define a few key elements to be re-used by all `spksrc.python*.mk` afterwards.
4. I faced an issue with building `lxml` ... Not sure bu this may be due to the `ADDITIONAL_LDFLAGS` being overriden in `spksrc.wheel.mk` when calling pip.
5. I first thought that automating `pkgconfig/*.pc` and `$(WORK_DIR)/*_done` symlinks so everything from python311 gets linked, it hapens that `libpng` cannot find `zlib`... resulting in a first exception, then `rdiff-backup` requires `bzip2` binaries resulting in a second exception.  On one hand more we can re-use libraries comming from the python package smaller are the package sizes and lower our CPU usage when building packages... But how many exceptions will we have to manage?

All in all, devils are in the details... and python wheel building has a hell lot of them.  I'll have to sleep on it but I wonder, while I believe that conceptually `spksrc.python.mk` is heading in a right direction (and saves us tons of wasted cpu cycles and makes spksrc part of a greener IT and resulting packages much smaller), it probably will require a lot more efforts to solve all use-cases... or I may be wrong and these will be easily fixed.

I don't think these are wasted efforts as `spksrc.python.mk` can be kept for later and migrated packages can be move back to using `spksrc.spk.mk` instead.   But this might be just too much for this already big PR ...although it can be kept as proof of concept and re-splitted later on for merging...

Need to sleep on it :last_quarter_moon_with_face:

kashifkhan:
@hgy59 I work on the azure event hubs & service bus python libraries. I noticed that this PR was linked to an issue for uamqp and was wondering if that is required due to homeassistant? If that's the case we did issue updates to both libraries for home assistant, so that uamqp would no longer be a requirement. We have a new all python based AMQP library and that no longer requires compiling uamqp.

hgy59:
> @hgy59 I work on the azure event hubs & service bus python libraries. I noticed that this PR was linked to an issue for uamqp and was wondering if that is required due to homeassistant? If that's the case we did issue updates to both libraries for home assistant, so that uamqp would no longer be a requirement. We have a new all python based AMQP library and that no longer requires compiling uamqp.

Yes, it is for homeassistant as you can see in the commit https://github.com/SynoCommunity/spksrc/pull/5820/commits/682222ad2e63fdb1819339354b96bd86846bd801

This is still required for homeassisant 2023.1.7 linked to openssl 3. This version needs to (cross) compile uamqp wheel.

Future versions of homeassistant package will benefit of new python libraries for azure integrations and not need this anymore.

th0ma7:
@hgy59 I was able to cover a lot of use-cases.  This should already speed-up considerably the build time.  Left to be sorted out is rust-wheels built with `spksrc.wheel.mk` to which I already have a good idea how to proceed (ref comment above https://github.com/SynoCommunity/spksrc/pull/5820#issuecomment-1660128963).

Now let's see if this was sufficient to already impact positively the github-action builds... wait & see how things turns out with current changes...

th0ma7:
Looking at the github-action build logs it seems `ifneq ($(wildcard $(PYTHON_DIR)),)` is not being handled... weird.

Although good news, it does process python311, 310, ffmpeg456 right at first, so that part is good.

hgy59:
@th0ma7 homeassistant with python310 fails to build cross/nmap when using pyhton.mk.

it does not find the openssl headers
```
In file included from nse_main.cc:19:0:
nse_openssl.h:8:24: fatal error: openssl/bn.h: No such file or directory
 #include <openssl/bn.h>
                        ^
compilation terminated.
make[4]: *** [Makefile:119: nse_main.o] Error 1
make[4]: *** Waiting for unfinished jobs....
make[4]: Leaving directory '/spksrc/spk/homeassistant/work-x64-6.2.4/nmap-7.94'
```

I suppose we need a solution to force the build of `cross/openssl3` within homeassistant and not reusing openssl3 of python310...

th0ma7:
@hgy59 a few quick options comes to mind:
1. there is a `--with-openssl=DIR` that could be leveraged.
2. `nmap` could be added by default to py310-311
3. leave `homeassistant` using `spk.mk`

My two cents?

th0ma7:
So again I was able to make a lot of mileage.  From my testint (and it will require a full re-testing) I'm now able to build everything except `homeassistant` and `znc`.

Related to `homeassistant` I'm getting an error with `nmap` that I was able to fix with:
```
+# In case OpenSSL is linked from another package working directory
+ifneq ($(strip $(OPENSSL_PREFIX)),)
+CONFIGURE_ARGS += --with-openssl=$(OPENSSL_STAGING_PREFIX)
+ADDITIONAL_LDFLAGS += -Wl,--rpath-link,$(OPENSSL_STAGING_PREFIX)/lib -Wl,--rpath,$(OPENSSL_PREFIX)/lib
+endif
```
Although the `ADDITIONAL_LDFLAGS` shouldn't have been needed as covered either in `spksrc.python.mk` or `spksrc.wheel-env.mk` ... so something is fishy.

Similarly I get errors in finding openssl libraries with `uamqp` and `znc` such as:
```
/home/spksrc/py-openssl3/spksrc/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/../lib/gcc/x86_64-pc-linux-gnu/8.5.0/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -lssl
/home/spksrc/py-openssl3/spksrc/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/../lib/gcc/x86_64-pc-linux-gnu/8.5.0/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -lcrypto
collect2: error: ld returned 1 exit status
```
I'm optimistic and presuming we're now really close.

Lastly, from my testing github-action does take care of re-using python libraries as expected.  There is just too much to build for the 6h00 duration even by optimizing the python builds (ffmpeg* probably kills it, perhaps worth migrating the culprit commit to a later PR with tvheadend & ffmeg?).

hgy59:
@th0ma7 as I have merged openssl update just now, you can probably rebase again and abort the current build actions...

th0ma7:
@hgy59 thnx for the update, and now rebased, but forgot to cancel previous job before hand...

th0ma7:
@hgy59 now we have a much clearer picture of the situation :)

th0ma7:
@hgy59 `nmap` now fixed and also confirmed that export are neededs.  End result with current code from my build host is good as the two `rpath` are built-in to the resulting binaries as to be expected.
```
libtree -v ncat
ncat 
├── libm.so.6 [ld.so.conf]
├── libc.so.6 [ld.so.conf]
├── libdl.so.2 [ld.so.conf]
├── libssl.so.3 not found
├── libcrypto.so.3 not found
└── libpcap.so.1 not found
    ┊ Paths considered in this order:
    ┊ 1. rpath:
    ┊    depth 1
    ┊    /var/packages/homeassistant/target/lib
    ┊    /var/packages/python310/target/lib
    ┊ 2. LD_LIBRARY_PATH was not set
    ┊ 3. runpath was not set
    ┊ 4. ld config files:
    ┊    /usr/lib/x86_64-linux-gnu/libfakeroot
    ┊    /usr/local/lib
    ┊    /usr/local/lib/x86_64-linux-gnu
    ┊    /lib/x86_64-linux-gnu
    ┊    /usr/lib/x86_64-linux-gnu
    ┊    /lib32
    ┊    /usr/lib32
    ┊    /libx32
    ┊    /usr/libx32
    ┊ 5. Standard paths:
    ┊    /lib
    ┊    /lib64
    ┊    /usr/lib
    ┊    /usr/lib64
Error [ncat]: Not all dependencies were found
```

th0ma7:
@hgy59 there is only homeassistant left... was there a reason to stick with cryptography <= 40?  I just can't get it to build properly using the shared path with python311 while version >= 40 work great.

hgy59:
> @hgy59 there is only homeassistant left... was there a reason to stick with cryptography <= 40? I just can't get it to build properly using the shared path with python311 while version >= 40 work great.

@th0ma7 yes, we must leave homeassistant 2023.1.7 on python310. 
My update to 2023.7.3 (#5757) is almost ready.

th0ma7:
@hgy59 yes, the need to keep using `python310` for `homeassistant`, that was clear to me already until your PR is complete.

Question is related to `cryptography_38`, I haven't been able to make it work using `spksrc.python.mk`.  Is there a mandatory requirement to keep using that version specifically or can it move to >=40 which works well using python.mk?

hgy59:
> Question is related to `cryptography_38`, I haven't been able to make it work using `spksrc.python.mk`. Is there a mandatory requirement to keep using that version specifically or can it move to >=40 which works well using python.mk?

Yes we need exactly this version. Homeassistant is bound to it.

th0ma7:
@hgy59 K, then I'll leave homeassistant as is for now.   As needed i'll assist you with your other pr as you see fit.

Also, for this pr I'll revisit the other items we had exchanged on (pip cache, pre-download wheels, etc) and see what can be done within this pr (or thru a subsequent).

In the meantime if you can test out resulting packages to confirm all is good would be much appreciated.

hgy59:
@th0ma7 regarding the pip cache:
I wonder whether it is possible to clean a pip cache so that it only removes files in the {pip-cache}/wheels directory and not the downloaded source files in the {pip-cache}/http directory.
I don't know the state of what is discussed in https://github.com/pypa/pip/issues/7372.

We could leave the pip-cache in the distrib/pip folder, when it is possible to call pip to clean the cache for wheels of a specific arch only while keeping the cached source files.

Otherwise we loose the original idea to cache source files in the distrib folder as we do for regular packages.

th0ma7:
@hgy59 i'm looking into that where we cache the source files in distrib/pip and cache the wheels in $(WORK_DIR)/pip.  Code is in progress, should be ready relatively soon for yours to review.

hgy59:
@th0ma7 latest build failed to build `spk/znc` for `aarch64-7.1` due to missing openssl (other builds succeeded to create znc package - or failed with timeout and didn't build znc therefore...)
My local build successfully creates all supported znc packages.

So this seems to be a temporary issue, not worth any further investigation.

th0ma7:
> @th0ma7 latest build failed to build `spk/znc` for `aarch64-7.1` due to missing openssl (other builds succeeded to create znc package - or failed with timeout and didn't build znc therefore...)
> My local build successfully creates all supported znc packages.
> 
> So this seems to be a temporary issue, not worth any further investigation.
> 

Thnx for looking into it :)

th0ma7:
@hgy59 if all goes well there is only one last item remaining: ensuring that `make wheelclean` it also cleans-up wheels generated from `cross/*` so they get rebuilt at next make call (as otherwise they got clearned-up and will be missing from the resulting package).

Also FYI, new commands:
- `make wheeldownload` ... now called from github-action to pre-download all wheel sources
- `make wheelcleancache` which also remove the `$(WORK_DIR)/pip` cache
- `make wheelcleanall` which also remove the `distrib/pip` wheel source directory

.github/actions/download.sh
```diff
@@ -19,3 +21,20 @@ else
...
+echo ""
+
+# Download python wheel sources files
+build_packages="${NOARCH_PACKAGES} ${ARCH_PACKAGES}"
+
+if [ -z "${build_packages}" ]; then
+    echo "===> No wheels to download. <==="
+else
+    for package in ${build_packages}
+    do
```
hgy59, line 34: just for the sake of formatting ...
```suggestion
    for package in ${build_packages}; do
```

mk/spksrc.wheel.mk
```diff
@@ -35,6 +35,49 @@ else
...
+	@if [ -n "$(WHEELS_2_DOWNLOAD)" ] ; then \
+		for wheel in $(WHEELS_2_DOWNLOAD) ; \
+		do \
+			$(MSG) "Downloading wheels from $$wheel ..." ; \
+			# BROKEN: https://github.com/pypa/pip/issues/1884 ; \
+			# xargs -n 1 $(PIP_SYSTEM) $(PIP_DOWNLOAD_ARGS) 2>/dev/null < $$wheel || true ; \
+			while IFS= read -r requirement ; \
+			do \
+				if [ "$$(grep -s egg <<< $${requirement})" ] ; then \
+					name=$$(echo $${requirement#*egg=} | cut -f1 -d=) ; \
```
hgy59, line 50: I would like the `for ; do` and `while ; do` on the same line here too (as we used to do in `if ; then`)
forgive me my pedantry 😄

hgy59:
@th0ma7 pytho310 fails now to build for OLD_PPC_ARCHS.
I really like to have python310 on my DS210+ with ppc853x-5.2 (I have accepted that python311 is not supported anymore for this model).

if fails to build wheels for `rpds-py` so I suppose it is depending on cryptography.

I confess that DSM 5.2 versions of python310 packages have not been published since `3.10.10-14`

hgy59:
@th0ma7 oops, you removed the list for build all-supported in the PR description (wanted to enter the success for python310).

Don't we need such verification as the build actions do not complete within 6 hours?

th0ma7:
> pytho310 fails now to build for OLD_PPC_ARCHS.

I'll have a look at it in the coming days.  I also noticed another regression where building `python31*` from `cross/` does not work anymore.  It may already be the case in master and unrelated to this PR, duno.

> you removed the list for build all-supported in the PR description (wanted to enter the success for python310).

Yeah, probably better to put them back in place... 

Lastly, testing github-action changes isn't really possible until merged.  As such there may be a small regression to fix after merging, I'll test that out afterwards to be sure

th0ma7:
> pytho310 fails now to build for OLD_PPC_ARCHS. I really like to have python310 on my DS210+ with ppc853x-5.2 (I have accepted that python311 is not supported anymore for this model).

@hgy59 I thought I was going to be able to help you out but no, the toolchain files are no longer available... :shrug:

hgy59:
> @hgy59 I thought I was going to be able to help you out but no, the toolchain files are no longer available... 🤷

I'll bring those toolchains back with #5840

hgy59:
@th0ma7 whow, a big step forward yesterday.

So I will restart my overall build testing now. 
A python update is only expected to 3.11.5 (scheduled last monday) but not for 3.10.

And I will try to have a deeper look into the github action code.

th0ma7:
@hgy59 thnx but honestly I was pretty much at that point since a little while from a building perspective.

One things I figured though is that moving pip cache to `$(WORK_DIR)/pip` was absolutely necessary.  This ensures that resulting wheels do always have the adequate rpath, something that otherwise could be broken using a shared workspace.

If you do have additional spare cycles to help (so we can put this one to a close), left todo are:
1. ensure make wheelclean also cleans-up the corresponding `cross/*` such as cryptography.
2. figure out the build error when building python directly from cross

Also from a github-action perspective :
1. confirming the wheel source download do occurs (currently it doesn't, and I believe it won't until merged... but unsure @publicarray might know?)
2. confirming if, when building a specific package for publishing, does it also build the relevant python (or ffmpeg for that matter) first, then the needed package and only uploading the right spk at then end?

hgy59:
> 2\. figure out the build error when building python directly from cross

analyzing the build error with cross/python311/Makefile

in the following line in `python311_post_install` target, the variable `$${CC}` seems to be empty, giving the error `__main__.py: error: argument --cc: expected one argument`
```
	@$(RUN) $(PYTHON_NATIVE) -m crossenv $(STAGING_INSTALL_PREFIX)/bin/python$(PKG_VERS_MAJOR_MINOR) --cc $${CC} --cxx $${CXX} --ar $${AR} --sysroot $${SYSROOT} --env LIBRARY_PATH= --manylinux manylinux2014 $(WORK_DIR)/crossenv/
```

BTW I removed the `CONFIGURE_ARGS += --disable-test-modules` part to get comparable build log files.
Here is the failing part:

succeeding in spk/python311
```
make[4]: Leaving directory '/spksrc/spk/python311/work-aarch64-6.2.4/Python-3.11.4'
mkdir -p /spksrc/spk/python311/work-aarch64-6.2.4/Python-3.11.4/build/lib.linux-aarch64-3.11
cp -R /spksrc/spk/python311/work-aarch64-6.2.4/../../../native/python311/work-native/Python-3.11.4/build/lib.linux-x86_64-3.11 /spksrc/spk/python311/work-aarch64-6.2.4/Python-3.11.4/build/lib.linux-aarch64-3.11/../
. /spksrc/spk/python311/work-aarch64-6.2.4/crossenv/bin/activate && cd /spksrc/spk/python311/work-aarch64-6.2.4/Python-3.11.4 && env PYO3_CROSS_LIB_DIR=/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/lib/ PYO3_CROSS_INCLUDE_DIR=/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/include/ OPENSSL_LIB_DIR=/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/lib/ OPENSSL_INCLUDE_DIR=/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/include/ PKG_CONFIG_LIBDIR=/spksrc/spk/python311/work-aarch64-6.2.4/install//var/packages/python311/target/lib/pkgconfig WORK_DIR=/spksrc/spk/python311/work-aarch64-6.2.4 INSTALL_PREFIX=/var/packages/python311/target TC=syno-aarch64-6.2.4 SYSROOT="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/aarch64-unknown-linux-gnueabi/sysroot" LD="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-ld" LDSHARED="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-gcc -shared" CPP="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-cpp" NM="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-nm" CC="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-gcc" AS="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-as" RANLIB="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-ranlib" CXX="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-g++" AR="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-ar" STRIP="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-strip" OBJDUMP="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-objdump" OBJCOPY="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-objcopy" READELF="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-readelf" CFLAGS="-I/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/aarch64-unknown-linux-gnueabi/sysroot/usr/include -I/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/include -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -L /spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/lib -I /spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/include" CPPFLAGS="-I/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/aarch64-unknown-linux-gnueabi/sysroot/usr/include -I/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/include " CXXFLAGS="-I/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/aarch64-unknown-linux-gnueabi/sysroot/usr/include -I/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/include " LDFLAGS="-L/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/aarch64-unknown-linux-gnueabi/sysroot/lib -L/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/lib -Wl,--rpath-link,/spksrc/spk/python311/work-aarch64-6.2.4/install/var/packages/python311/target/lib -Wl,--rpath,/var/packages/python311/target/lib " CARGO_HOME="/spksrc/distrib/cargo" RUSTUP_HOME="/spksrc/distrib/rustup" CARGO_BUILD_TARGET="aarch64-unknown-linux-gnu" CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_AR="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-ar" CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="/spksrc/toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bi
n/aarch64-unknown-linux-gnueabi-gcc" TC_GCC=$(eval $(echo /spksrc/spk/python311/work-aarch64-6.2.4/../../../toolchain/syno-aarch64-6.2.4/work/aarch64-unknown-linux-gnueabi/bin/aarch64-unknown-linux-gnueabi-gcc -dumpversion) 2>/dev/null || true) TC_GLIBC=2.20 TC_KERNEL=4.4.59+ wget https://bootstrap.pypa.io/get-pip.py
--2023-08-11 09:33:23--  https://bootstrap.pypa.io/get-pip.py
Resolving bootstrap.pypa.io (bootstrap.pypa.io)... 146.75.116.175, 2a04:4e42:8d::175
Connecting to bootstrap.pypa.io (bootstrap.pypa.io)|146.75.116.175|:443... connected.
HTTP request sent, awaiting response... 200 OK
```

failing in cross/python311:
```
make[3]: Leaving directory '/spksrc/cross/python311/work-aarch64-6.2.4/Python-3.11.4'
mkdir -p /spksrc/cross/python311/work-aarch64-6.2.4/Python-3.11.4/build/lib.linux-aarch64-3.11
cp -R /spksrc/cross/python311/work-aarch64-6.2.4/../../../native/python311/work-native/Python-3.11.4/build/lib.linux-x86_64-3.11 /spksrc/cross/python311/work-aarch64-6.2.4/Python-3.11.4/build/lib.linux-aarch64-3.11/../
usage: __main__.py [-h] [--cross-prefix CROSS_PREFIX] [--system-site-packages]
                   [--clear] [--clear-cross] [--clear-build] [--clear-both]
                   [--without-pip] [--without-cross-pip]
                   [--relative-toolchain] [--cc CC] [--cxx CXX] [--ar AR]
                   [--config-var CONFIG_VAR] [--env ENV] [--sysroot SYSROOT]
                   [--sysconfigdata-file SYSCONFIGDATA_FILE]
                   [--manylinux MANYLINUX] [--machine MACHINE] [-v]
                   [--version]
                   HOST_PYTHON ENV_DIR [ENV_DIR ...]
__main__.py: error: argument --cc: expected one argument
make[2]: *** [Makefile:164: python311_post_install] Error 2
```

th0ma7:
At first glance it sounds like shell type variables aren't available from cross?

hgy59:
> @th0ma7 pytho310 fails now to build for OLD_PPC_ARCHS. I really like to have python310 on my DS210+ with ppc853x-5.2 (I have accepted that python311 is not supported anymore for this model).
> 
> if fails to build wheels for `rpds-py` so I suppose it is depending on cryptography.

@th0ma7 the same error for `rpds-py` occurred when building python311 for arch-armv7-1.2 (i.e. for SRM 1.2)
As we do not have a toolchain for SRM 1.3 yet, we will not be able to provide python310/311 for SRM (router) models.

_you will need the latest update from master with fix for DSM 5.2 and SRM 1.2 toolchains to validate this._

th0ma7:
@hgy59 I believe all remaining opened items are now closed.

> And I will try to have a deeper look into the github action code.

Have you had any cycles to review that part?  Left was ensuring that python always gets built, first, when building a specific package for publishing (same applies for ffmpeg which I always published manually as unsure of its actual behavior).

Other than that, I believe it's now ready for merge.  Actually I suggest the remaining github-action part could be left aside and done in a subsequent PR while updating to 3.11.5.  Looking at the release schedule it should have been released last monday: `3.11.5: Monday, 2023-08-07` but it isn't available yet.

hgy59:
> Left was ensuring that python always gets built, first, when building a specific package for publishing (same applies for ffmpeg which I always published manually as unsure of its actual behavior).

I already validated that python is always built first (and before ffmpeg*).

I did not check, whether python will be published again, when publishing a dependent package from github action. Publishing the same python package again will result in an error. Would this be a breaking issue? (I never use the github action to publish packages. I always make local builds to test before publishing...)

th0ma7:
> I already validated that python is always built first (and before ffmpeg*). (done some days ago white_check_mark )

Excellent, thnx.

> Would this be a breaking issue? 

humm, I guess we could merge and test github-action publishing and confirm from there.  Which makes me wonder, there a hell of a lot of packages to publish... 

Also, I just recall, there's still `salt-master` to fix at startup (or mark both salt-* as broken or deleting them alltogether).  Any thoughts on this in particular?

hgy59:
@th0ma7 after merging this PR, what is the schedule?

IMO we must first publish the dependent packages. This does not break anything, as those cannot be installed before the python updates are installed.
We should publish the python updates a day or two later to ensure all dependent packages are available for update in DSM package center.

Probably we have to add a note in the changelog of the python311 package that it has to be installed before dependent packages, but dependent packages might not work until updated. So the updates of the dependent packages should be available at this time.

After writing the above, I would adjust the schedule like this:
1. Publish all python dependent packages, that depend on python311
2. After a day or two publish updated python311 package (this will not break existing packages that currently depend on python310) (is there a package that currently depends on python311? Only for this, an immediate update is required after python311 update)
3. Some days later we can publish the updated homeassistant package. 
4. A day or two after publishing homeassistant, publish updated python310 - hopefully all users have updated the packages that now depend on python311.

In the additional note in the changelog of python311 only the packages that already depend on python311 must be mentioned

th0ma7:
@hgy59 you proposal sounds decent enough... I have rephrased it in the top comment to confirm we're on the same track (or adjust as need be).  Also, from what I recall only `rtorrent` and `sabnzbd` where already using python311 but needs to be confirmed.  As for `tvheadend`, lets keep it as is for now (i.e. do not publish it) as I'll rework that one in another PR #5834.

BTW, salt-master is now fixed :)

hgy59:
> BTW, salt-master is now fixed :)

I just installed salt-minion and salt-master and both are running.
I had to accept the pending key on salt-master with
```
 ./salt-key -a {host-name-of-local-system} -c /var/packages/salt-master/etc
```

to get rid of the log warnings int salt-master:
```
... authentication failed from host  {host-name-of-local-system}, the key is in pending and needs to be accepted with salt-key -a  {host-name-of-local-system}
```

Just wondering whether the salt-master package should also install the SaltGUI?
My tests are on DS218 (aarch64) with DSM 6.2.4
So I just copied the folder `/var/packages/salt-master/var/saltgui` to `/var/services/web/saltgui` and get a login mask when browsing the http://{nas-ip}/saltgui (I have not a clue what credentials are required):

SaltGUI-login


So probably the salt-* packages require another cycle...

hgy59:
> @hgy59 you proposal sounds decent enough... I have rephrased it in the top comment to confirm we're on the same track (or adjust as need be). Also, from what I recall only `rtorrent` and `sabnzbd` where already using python311 but needs to be confirmed.

I had to rethink the deployment, and only `deluge, `rtorrent`, `sabnzbd` and `sickchill` have to be published before the new python311 package.
All other packages (currently running on python 310) are not affected before by the first update cycle.

After this first cycle, we have (plenty of) time, to publish the remaining packages.
This gives us extra time for verification (and fixes?) of the remaining packages...

I have to validate the list of packages that are already published with python311 dependency.

hgy59 (commented):
LGTM so far

hgy59:
> I have to validate the list of packages that are already published with python311 dependency.

I have validated that only `deluge`, `rtorrent`, `sabnzbd` and `sickchill` where migrated to python311 before this PR.
Therefore I have updated the publishing schedule in the PR description

th0ma7:
The original `salt-*` never had saltgui included.  It is I who included it during the last release as a [WIP] as I never finished integrating that part (ain't my cup of thea at figuring web integration)

As for the deployment timeline, I was guessing it may be best starting with `python310` an `python311` directly but I forgot about the openssl3 update.  So yeah, guessing your proposal above makes sense.  Would you mind updating the top description for day 1-3-5 publishing act accordingly?

th0ma7:
@hgy59 nope, re-thinking about it and the python310 portion won't work...

Day 3:
- python311

Day 5:
- all package that is migrating to python311
- homeassistant

Day 7:
- python310

hgy59:
> @hgy59 nope, re-thinking about it and the python310 portion won't work...

oops, you are so right!

but I have no problem to postpone the homeassistant and python310 update until all python311 migrated packages are published (even when this is takes time until the end of this year...)

So it would be better to work on #5757 (this falls into the cycle of all packages migrated to python311)
This said, we would never have to publish homeassistant 2023.1.7 and python310 with openssl 3 (regarding existing packages).

th0ma7:
@hgy59 wondering (while it is building), why not release python311 at the same time at day 1?  I fear otherwise we may get false-positive reports where a package can't install due to a missing dependency.

hgy59:
> @hgy59 wondering (while it is building), why not release python311 at the same time at day 1? I fear otherwise we may get false-positive reports where a package can't install due to a missing dependency.

It is because we must ensure that the updates for `deluge`, `rutorrent`, `sabnzbd` and `sickchill` are available in the DSM package center when the python311 update is installed. Otherwise the packages will be broken until the updates are available.
Assuming that not all packages are visible in DSM package center at the same time when published together.

IMO it is better to get reports that updates are not installable as the python311 update is not yet available instead of reports that packages are broken after the python311 update when the required updates are not yet available.

To try to avoid such reports we must add a note in the changelog of `python311`, that updates for `deluge`, `rtorrent`, `sabnzbd` and `sickchill` must be available before python311 is updated, to avoid to break those packages.

th0ma7:
`deluge`, `rtorrent`, `sabnzbd` and `sickchill`  now published.  Remaining is `python311` to which feel free to add the liner you see fit in the changelog (not sure of its pertinence but I trust your jugement).  I'll publish it probably tomorrow night or monday, then move with the others over the week.

`homeassitant`?  I leave it aside entirely until your PR is ready?

hgy59:
> `homeassitant`? I leave it aside entirely until your PR is ready?

Yes, I plan to be ready with homeassistant 2023.7.3 in the near future...

But just found an issue with the this python311 PR.
I want to build homeassistant with include of `spksrc.spk.mk`.
But this now skipps the build of python311 (and openssl3 and may be other cross dependencies)  when a build of the same ARCH-TC is available in spk/python311.

But this is not possible anymore. It does not build python311 in spk/homeassistant and fails to create the wheels.
I would expect this behaviour only when `spksrc.python.mk` is included but not with `spksrc.spk.mk`.

When comparing to ffmpeg* this is (or was?) only using an existing spk/ffmpeg* when it exists, otherwise it the ffmpeg* is built locally.
I do not expect the same for python311, because we can force the use of prebuilt python311 (and abort with an error if it does not exist). But when I do not want a prebuilt python311, I do not expect that an existing prebuilt python311 is used and I do not expect at all, that it does not build python311 locally.

th0ma7:
That is really weird?!  No changes got into py311 nor py310 with this PR that relates to this.  Also this PR currently build homeassistant "as-usual" from `spk.mk` ... Also yes, it does look for the existence of a `spk/python311/work-<arch>-<version>/` using `ifneq ($(wildcard $(PYTHON_PACKAGE_ROOT)),)` ... 

Having a branch I could look into?

th0ma7:
I believe I may know why, when replacing BUILD_DEPENDS with PYTHON_VERSION and no spk/python311 exists it simply fails to build.  So in spk.python.mk there should be an else where BUILD_DEPENDS gets re-introduced.

In our use cases we haven't hit that but it should have been an obvious case to look into.

th0ma7:
@hgy59 so following-up on #5841, at first glance this looks like specific to `armv5` and `qoriq`.  Testing on my end I had success on `x64-6.2.4`, `x64-7.1` and `armv7-7.1` where all is working as expected.  It hapend I do have a good old `armv5` (boy it takes time to install packages...) and there really is a problem with the ssl library.

Thus, situation is not DSM6 specific, and I'd beleive not python specific as it is the same version as prior (`3.11.4`).

Although it has somehow to be `openssl3` related:
- either issue with old `glibc`
- or issue with older `compiler` requiring additional switches?

... Therefore, before releasing anything else from this PR I'd rather figure out where the issue lies.  So currently investigating...  Out of curiosity, having a `qoriq` nearby?

hgy59:
or just drop python311 for ARMv5 (as we did for OLD_PPC_ARCHS)

th0ma7:
@hgy59 honestly, I'm thinking along the same line and just go ahead with deleting the armv5 packages online and mark it as such in python311. 

I'd push things even a bit further, declaring `qoriq` as deprecated as well from a py311 standpoint.  Unless you (or @SynoCommunity/developers) has one close by that we could try to isolate the segfault core dump from?

Safihre:
We should then also mark all Python packages (like SABnzbd) as not supporting those arch's anymore.

hgy59:
> has one close by that we could try to isolate the segfault core dump from?

IMO a segfault indicates that the bigendian arch is not considered by the source of the affected library.
My single PPC (bigendian) model is DS210+ with ppc853x-5.2, that has the same issue on openssl3 for python311 as ARMv7 (lacking libatomic due to gcc < 4.8).

The only models with qoric are DS213+, DS413.

---

I like to support old models, as it helps to keep those usable and to reduce resource wasting...
As long as code does not depend on kernel versions, we might get further by creation of our own toolchains with support for newer gcc (follow up of #5225 and #4469) - and we should have a look at [crosstool-ng](https://crosstool-ng.github.io/).

Or we reinvest into chroot packages (at least for models without docker support) - _but this might not be a solution for (32-bit) PPC archs (and ARMv5 was only supported until debian 9 aka stretch with EOL 2022.06)_

th0ma7:
Available options for `armv5` are:
1. to `if`, `else` with `openssl` v1 and have python311 packages being supported until it gets deprecated (which is due to september 11th 2023, almost tomorrow).  And even perhaps afterwards with a message on its changelog stating that it is now unsupported somehow, or that this is the last version to be supplied for ARMv5?
2. mark is as deprecated right away and delete packages online - and add unsupported flags to `python311`

As for `qoriq` options relatively to `sabnzbd` are (I presume this is cryptography related):
1. Find where the issue is?  FYI, endianness looks OK:
```
$ file ./Cheetah/_namemapper.cpython-311-powerpc-linux-gnuspe.so ./cryptography/hazmat/bindings/_rust.abi3.so
./Cheetah/_namemapper.cpython-311-powerpc-linux-gnuspe.so: ELF 32-bit MSB shared object, PowerPC or cisco 4500, version 1 (SYSV), dynamically linked, with debug_info, not stripped
./cryptography/hazmat/bindings/_rust.abi3.so:              ELF 32-bit MSB shared object, PowerPC or cisco 4500, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```
2. similar as with armv5 for python311, to `if`, `else` but this time relatively to `cryptography` to keep using the older C version?  And for this one as well, mark it as being the last version to be supplied for qoriq
3. mark is as deprecated right away and delete packages online - and add unsupported flags to `cross/cryptography`

------------------------------------
As for supporting newer gcc, I've tried that and came to the conclusion that to build our own cross-compiler we have to restart from scratch and go with the stage1 and stage2 builds.  It's a hell of a road to get to something that works.  crosstool-ng may be a hell lot more easier to use?

hgy59:
@th0ma7 
I am trying to build homeassistant (currently 2023.8.2) that depends on `Pillow==10.0.0` and I get the following error:
```
===>  _PYTHON_HOST_PLATFORM=aarch64-unknown-linux-gnueabi /spksrc/spk/python311/work-aarch64-6.2.4/crossenv/cross/bin/pip wheel --disable-pip-version-check --no-binary :all: --find-links /spksrc/spk/homeassistant/../../distrib/pip --cache-dir /spksrc/spk/homeassistant/work-aarch64-6.2.4/pip --no-deps --wheel-dir /spksrc/spk/homeassistant/work-aarch64-6.2.4/wheelhouse --no-index --global-option=build_ext --global-option=--disable-platform-guessing --global-option=--enable-freetype --global-option=--enable-jpeg --global-option=--enable-zlib --no-use-pep517 --no-build-isolation Pillow==10.0.0
DEPRECATION: --build-option and --global-option are deprecated. pip 23.3 will enforce this behaviour change. A possible replacement is to use --config-settings. Discussion can be found at https://github.com/pypa/pip/issues/11859
WARNING: Implying --no-binary=:all: due to the presence of --build-option / --global-option.
Looking in links: /spksrc/spk/homeassistant/../../distrib/pip
Processing /spksrc/distrib/pip/Pillow-10.0.0.tar.gz
ERROR: Disabling PEP 517 processing is invalid: project specifies a build backend of backend in pyproject.toml
```

If I remove the following line in `spksrc.wheel.mk`:
```
         pip_global_option=$${pip_global_option}" --no-use-pep517" ; \
```
then cross build of the wheel does not work. It takes headers in the build system instead of the toolchain:
`/usr/include/bits/floatn.h:87:9: error: unknown type name ‘__float128’`


Any hint?

th0ma7:
Yup, i've hit that same issue a few weeks ago while updating python packages.  I had to stick with the previous 9.x version as it didn't work.

My investigations led me to exactly the same point.  I would argue either trying out from cross and/or opening a ticket on their github project page...

hgy59:
> My investigations led me to exactly the same point. I would argue either trying out from cross and/or opening a ticket on their github project page...

Thanks, I prefer to stay on homeassistant 2023.7.3 for the moment.

th0ma7:
> I am trying to build homeassistant (currently 2023.8.2) that depends on `Pillow==10.0.0` and I get the following error:

@hgy59 I've looked at this and may have found a few interesting bits... some of it pertains to next release of `numpy` that will have to migrate to `scikit-build-core` (which in turns uses cmake) or `meson-python` (which uses meson+ninja).  It's totally unclear how to pass our cross-environment files to them so they do use Synology toolchain properly.

Building python wheels is a mix-bag of magic, vodoo and luck... it seems that starting with newer pillow and numpy we just got luckier :)

hgy59:
> Yup, i've hit that same issue a few weeks ago while updating python packages. I had to stick with the previous 9.x version as it didn't work.
> 
> My investigations led me to exactly the same point. I would argue either trying out from cross and/or opening a ticket on their github project page...

I tried agaein to build  `Pillow==10.0.0`
It is as easy as removing `--no-use-pep517` from `pip_global_option` in `spksrc.wheel.mk`.

The error message tells, that this option is not valid when building with `pyproject.toml`
```
ERROR: Disabling PEP 517 processing is invalid: project specifies a build backend of backend in pyproject.toml
```

I think we should fully rely on PEP 517 and remove the  `--no-use-pep517`  option when building wheels.

th0ma7:
There will be some wheels that will refuse to build.  I recall when i added it initially there where many failures.

Although situation may have changed since whereas only a few may still fail for which we could then add the option (e.g. disabling pep517) on a per wheel basis.

hgy59:
> There will be some wheels that will refuse to build. I recall when i added it initially there where many failures.
> 
> Although situation may have changed since whereas only a few may still fail for which we could then add the option (e.g. disabling pep517) on a per wheel basis.

OK, one issue I found with PEP517:
```
      In file included from /usr/include/stdlib.h:55:0,
                       from /spksrc/spk/python311/work-armv7-6.2.4/install/var/packages/python311/target/include/python3.11/Python.h:23,
                       from src/_imagingft.c:22:
      /usr/include/bits/floatn.h:75:1: error: unknown machine mode ‘__TC__’
```

Cross builds need investigation. The error above shows that it takes include files from the system, instead of the toolchain.

th0ma7:
@hgy59 was that from pillow build or another one entirely?

th0ma7:
@hgy59 if I recall you had looked into `salt-master` along with enabling the `saltgui` (which probably needs updating).

Is this something you have cycles for to complete before I publish `salt-*` packages?  Otherwise I could probably publish them as-is, but unsure of the value of it?

hgy59:
> @hgy59 was that from pillow build or another one entirely?

yes, it is when building wheel Pillow==10.0.0 for homeassistant with arch-armv7-6.2.4 without the `--no-use-pep517` parameter in spksrc.wheel.mk.

The code in Python.h (from spk/python311/work-armv7-6.2.4/install/...) looks like this
```h
// stdlib.h, stdio.h, errno.h and string.h headers are not used by Python
// headers, but kept for backward compatibility. They are excluded from the
// limited C API of Python 3.11.
#if !defined(Py_LIMITED_API) || Py_LIMITED_API+0 < 0x030b0000
#  include <stdlib.h>
#  include <stdio.h>              // FILE*
#  include <errno.h>              // errno
#  include <string.h>             // memcpy()
#endif
```

mreid-tt:
@th0ma7, are there any other packages to publish from this? Just wondering if we should change the label to "published" or not.

th0ma7:
@mreid-tt all packages have been migrated and published.  The only remaining items relates to :
1. there are the `salt-*` packages that requires attention.  Re-using someone else's code I was able to migrate it from python2 to python3, then to python 3.11.  But under the hood I ended-up noticing it is missing the actual web gui enablement which never was provided with the package (not even with python2).  So I did introduced the web portion whereas it's now packaged but I didn't enabled it as web apps isn't really my area I can contribute much.  I'd need someone else to look at that specifically or we just call it a day and stop supporting it (as I doubt it ever fully worked without the web portal being enabled anyway?!?).
2. figuring out how to get powerpc to work with rust as otherwise, similarly to armv5, that arch will have to become unsupported.  I've spent numerous hours on it and may be close to a result ... that will hopefuly actually work.
3. before too long we will have to reconcile how to manage (or not) the `--no-use-pep517` flag using `pip`.  As mentionned previously, saying that cross-compiling python wheel is a challenge in an understatement as the Python community doesn't agree or commit to a standard toolset (ref. https://lwn.net/Articles/924114/)

So long-story-short:
1. `salt-*` is in need for someone else to look into it or we decide to drop it
2. if I can find a few more cycles I may have a tentative solution, probably a bit sloppy at first
3. when @hgy59 looks into the next homeassistant update we'll need to revisit the `--no-use-pep517` flag issue

mreid-tt:
@th0ma7 thanks for the update. I've never used salt or dug very deeply into python packages but I have touched on a few with web gui components. If you have a branch or a PR that I can look at I can try to help.

For now based on your comments, I'll leave the tags on this PR as is.

PR #5829, opened 2023-08-04
erlang: update to 25.3

hgy59:
## Description

This PR updates the Erlang ecosystem packages (Erlang, Elixir, RabbitMQ, ejabberd) and fixes critical DSM 7 upgrade issues.

### Version Updates

| Package | Old Version | New Version |
|---------|-------------|-------------|
| Erlang | 25.0.4 | 25.3.2.21 |
| Elixir (native) | 1.14.0 | 1.15.8 |
| RabbitMQ | 3.10.7 | 3.11.28 |
| ejabberd | 21.07 | 23.10 |

### Key Improvements

**ejabberd**
- Fixed configuration paths for DSM 7+ (`SYNOPKG_PKGVAR` instead of `SYNOPKG_PKGDEST`)
- Added dynamic upgrade wizard that detects missing admin ACL and prompts for credentials
- Added automatic data migration from old paths during DSM 7 upgrades
- Added CLI wrapper with proper permission handling for `ejabberdctl`
- Added web interface shortcut in DSM

**RabbitMQ**
- Fixed configuration for DSM 7
- Added DSM 7 upgrade data migration (cookie and var folder)
- Added CLI wrappers with proper permission handling for all management tools
- Added web interface shortcut to RabbitMQ Management UI (port 15672)

**Erlang**
- Fixed OpenSSL include order to resolve build issues
- Disabled parallel make for Elixir bootstrap to fix race condition

## Checklist

### Type of change

- [x] Package update


mreid-tt:
hey @hgy59, based on your https://github.com/SynoCommunity/spksrc/pull/6886#issuecomment-3714732755, this should be merged before any other updates to Erlang. From what I see the latest version in the 25.3.x train is [OTP 25.3.2.21](https://github.com/erlang/otp/releases/tag/OTP-25.3.2.21). Would you like me to bump to the latest versions compatible with this ([RabbitMQ 3.11.28](https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.11.28) and [ejabberd 25.10](https://github.com/processone/ejabberd/releases/tag/25.10)), test and publish for you? Any other things need to be completed in this PR?

hgy59:
> From what I see the latest version in the 25.3.x train is [OTP 25.3.2.21](https://github.com/erlang/otp/releases/tag/OTP-25.3.2.21)

I tried to build Erlang 25.3.2.21 but it fails (conflicts with openssl) the latest 25.3.x that builds is 25.3.2.7 (we already have 25.3.2.5 that is almoste the same).

rabbitmq that is compatible with erlang 25.3.x fails to build, this must be fixed first.

mreid-tt:
hey @hgy59, may I add some commits to your PR to get the build to work again?

hgy59:
> hey @hgy59, may I add some commits to your PR to get the build to work again?

Yes, please, if you have a solution...

mreid-tt:
> Yes, please, if you have a solution...

My solutions have beed added. As you will note I've gone to the latest compatible versions as follows:

- erlang to 25.3.2.21 (with OpenSSL include fix)
- rabbitmq to 3.11.28
- ejabberd to 23.10

The native libraries have also been updated to latest compatible versions:

- native/elixir: update to 1.15.8 (with disable parallel make fix)
- ~native/erlang: update to 25.3.2.21~

hgy59:
> * native/erlang: update to 25.3.2.21

This can stay on 25.3 as before. It must not match the cross/erlang version (this is different to other systems like native/python*).
Only major.minor version must match, patch version is not relevant.

mreid-tt:
> This can stay on 25.3 as before. It must not match the cross/erlang version (this is different to other systems like native/python*).
> Only major.minor version must match

Can you explain why this may be? Happy to roll it back but was curious about this.

hgy59:
> Can you explain why this may be? Happy to roll it back but was curious about this.

I did it this way since first erlang package and it was confirmed by an AI query.
AI confirmed that it is not officially documented, but says

The "major.minor must match" requirement is essentially this combination of:
- Patch-level compatibility guarantee
- No guarantee regarding minor boundaries
- Years of build experience in the Erlang/OTP project

hgy59:
@mreid-tt thanks for your fixes!

hgy59:
just found that rabbitmq never worked on DSM 7+...

hgy59:
ejabberd (also the current version) does not work in DSM 7+.

But the fix is not so easy as for rabbitmq.
ejabberd installs some files and folders to ejabberd/target/var instead of ejabberd/var.
This comes from the `localstatedir` i.e. `INSTALL_PREFIX_VAR` and is related to the spksrc framwork.
Also patching Makefile.in (by patch) and sbin/ejabberdctl (by post_install_target) might need adjustment...

mreid-tt:
> ejabberd installs some files and folders to ejabberd/target/var instead of ejabberd/var.
> This comes from the `localstatedir` i.e. `INSTALL_PREFIX_VAR` and is related to the spksrc framwork.
> Also patching Makefile.in (by patch) and sbin/ejabberdctl (by post_install_target) might need adjustment...

If you can spell out the precise issues to be resolved I can try to tackle them for you.

EDIT: hey @hgy59, I've proposed a fix. Feel free to revert it if this was not what you were taking about.

mreid-tt:
hey @th0ma7, in examining the issue above the AI that I was using suggested that there is an issue in the framework as follows:

---

**The Framework Issue**

The root cause is an include order problem in `mk/spksrc.cross-cc.mk`:

Line 25 - directories included FIRST
```
include ../../mk/spksrc.directories.mk
```

Line 28 - common (with version_ge function) included SECOND
```
```

In `spksrc.directories.mk` (lines 64-67), there's logic that depends on `version_ge`:

```
ifeq ($(lastword $(subst /, ,$(INSTALL_PREFIX))),target)
INSTALL_PREFIX_VAR = $(shell dirname $(INSTALL_PREFIX))/var
endif
endif
```

Since `version_ge` isn't defined yet when this is evaluated, the call returns empty, the condition fails, and `INSTALL_PREFIX_VAR` falls back to `$(INSTALL_PREFIX)/var` (which includes `/target/`).

**Options for Framework Fix**

1. Reorder includes in spksrc.cross-cc.mk - Move `spksrc.common.mk` before `spksrc.directories.mk`
    - Risk: Could break other packages that depend on current include order
    - Impact: Needs careful testing across all package types
2. Move `version_ge` to a separate early-include file - Create `spksrc.functions.mk` that's included first
    - Risk: Low, but adds complexity
    - Impact: Clean separation of concerns
3. Remove version_ge dependency from directories.mk - Use shell comparison instead:

   `ifeq ($(shell [ "$(TCVERSION)" != "" ] && printf '%s\n' "$(TCVERSION)" "7.0" | sort -VCr && echo 1),1)`
    - Risk: Slightly less readable
    - Impact: Self-contained fix
4. Leave as-is, document the limitation - Packages needing `INSTALL_PREFIX_VAR` in cross context must use explicit paths
    - Risk: None to framework
    - Impact: Packages like ejabberd need workarounds

**Which packages could be affected?**

Any cross package that:

- Has a `POST_INSTALL_TARGET` that references `INSTALL_PREFIX_VAR`
- Installs config/data files that need DSM 7+ var path handling

Examples include: dar, lm-sensors, phpmemcachedadmin, smartmontools.

---

This of course needs additional validation but I thought I'd share with you to see if it sounds correct with your familiarity of the framework.

hgy59:
@mreid-tt good catch, ../../mk/spksrc.common.mk must always be included before use of `version_ge` et. al.

When `INSTALL_PREFIX_VAR` will be fixed, we will need to handle the files that are installed to the var folder differently, because the PLIST file can handle only files that are installed under package/target (all under package/var must be installed in spk Makefile and can't be installed in cross Makefile).
And when this is fixed, we wouldn't need to replace /target/var by /var in sbin/ejabberdctl.

th0ma7:
@mreid-tt indeed good catch.  I suggest you make the change, either here and confirm it runs as expected or through a standalone PR with a few test build to confirm.

hgy59:
> @mreid-tt indeed good catch. I suggest you make the change, either here and confirm it runs as expected or through a standalone PR with a few test build to confirm.

This is not a small framework change
IMHO we have to split spksrc.common.mk into two files
one that is included before include of spksrc.directories.mk that does not depend on any directory (like CURDIR, BASEDIR, WORK_DIR, ...)
and the otherone that is included after include of spksrc.directories.mk 

I would like a dedicated PR and continue here with a workaround that does not need this fix.

Except we can avoid the use of version_ge in spksrc.directories.mk by definition of INSTALL_PREFIX_VAR the same for all DSM versions...

th0ma7:
@hgy59 can or can't?  DSM7's var environment is indeed different, thus this variable.  Personally the option 1 of reordering is limited in scope and would suffice for now.  spksrc.common.mk could be added to the META issue and done at a later time into its own PR indeed.

hgy59:
> DSM7's var environment is indeed different,

The generic installer of spksrc ensures the `$(INSTALL_DIR)/var` folder for all DSM versions.
So all DSM version have a DSM 7 compatible var folder.

mk/spksrc.directories.mk
th0ma7, line 1: ## TL;DR;
@hgy59 i started writing this being absolutely sure you were breaking things up for DSM6 with this small change... further i dig into it less I'm sure its the case - a few install + upgrade testing is needed to confirm its actually all is good.

EDIT: `spksrc.copy.mk` also needs changing if you want to enable this behavior by default...

## Long-version:
DSM6 = `/var/packages/<name>/target/var`
DSM7 = `/var/packages/<name>/var`

With the following change you're enforcing DSM6 `var` to be installed in the wrong path... as the `@appdata` does not exist.

Although interestingly now DSM6 allow using `/var/packages/<name>/var` but in reality DSM6 `var` symlink actually point to `target/var` ... I don't recall this was the case when I made this change for DSM7 a while back but that might have been an oversight of me.

`git` on DSM7:
```
th0ma7@th0ma7-nas:/var/packages/git$ ls -la
total 40
drwxr-xr-x  5 root root  4096 Nov 22 12:09 .
drwxr-xr-x 72 root root  4096 Jan  4 12:09 ..
drwxr-xr-x  2 root root  4096 Nov 22 12:08 conf
-rw-r--r--  1 root root     0 Nov 22 12:09 enabled
lrwxrwxrwx  1 root root    21 Oct 22 06:14 etc -> /volume1/@appconf/git
lrwxrwxrwx  1 root root    21 Oct 22 06:14 home -> /volume1/@apphome/git
-rw-r--r--  1 root root 17417 Oct 22 06:14 INFO
drwxr-xr-x  2 root root  4096 Oct 20 15:31 scripts
lrwxrwxrwx  1 root root    22 Oct 22 06:14 share -> /volume1/@appshare/git
lrwxrwxrwx  1 root root    22 Oct 22 06:14 target -> /volume1/@appstore/git
lrwxrwxrwx  1 root root    21 Oct 22 06:14 tmp -> /volume1/@apptemp/git
lrwxrwxrwx  1 root root    21 Oct 22 06:14 var -> /volume1/@appdata/git
drwxr-xr-x  2 root root  4096 Oct 20 15:31 WIZARD_UIFILES
```
And `git` on DSM6:
```
th0ma7@th0ma7-armv5:/var/packages/git$ ls -la
total 36
drwxr-xr-x 4 root root  4096 Jan 10 21:13 .
drwxr-xr-x 6 root root  4096 Jan 10 21:13 ..
drwxr-xr-x 2 root root  4096 Jan 10 21:13 conf
-rw-r--r-- 1 root root     0 Jan 10 21:13 enabled
lrwxrwxrwx 1 root root    26 Jan 10 21:13 etc -> /usr/syno/etc/packages/git
-rw-r--r-- 1 root root 17622 Jan 10 21:13 INFO
drwxr-xr-x 2 root root  4096 Oct 20 14:31 scripts
lrwxrwxrwx 1 root root    22 Jan 10 21:13 target -> /volume1/@appstore/git
lrwxrwxrwx 1 root root    26 Jan 10 21:13 var -> /volume1/@appstore/git/var
```

Issue is, there isn't any migration operation from `target/var` to `target/../var` in the DSM6 install scripts.  I first thought this was going to break existing packages upon their upgrade... but maybe not... 

dsm7 script has a `syno_sync_var_folder` which migrates the data from `target/var` to `var` ... but as the data is already in the right location in the package (i.e. already is in `target/var`), in theory no move is needed.  And since the symlink exists, paths expected at compile time should still work, surprisingly ?!??

And also suggesting a clearer code for `$(shell dirname $(INSTALL_PREFIX))/var` 
to be using instead `$(abspath $(INSTALL_PREFIX)/../var)`

Further digging into it... spksrc.copy.mk does a hook to capture this special `var` directory such as:
```
copy_target: SHELL:=/bin/bash
copy_target: .SHELLFLAGS := -o pipefail -c
copy_target: $(PRE_COPY_TARGET) $(INSTALL_PLIST)
        @$(MSG) [DSM7+] Copy target to staging, discard var directory
        @(mkdir -p $(STAGING_DIR) && cd $(STAGING_INSTALL_PREFIX) && tar cpf - $$(cat $(INSTALL_PLIST) | sed -e '/^.*:var\/.*/d' -e 's/^.*://g')) | \
          tar xpf - -C $(STAGING_DIR)
        @$(MSG) "[DSM7+] Copy and merge var and target/var to $(STAGING_DIR)/var"
        @if [ "$$(cat $(INSTALL_PLIST) | sed -n 's?^.*:var/??p')" ] ; then \
          mkdir -p $(STAGING_DIR)/var ; \
          (cd $(STAGING_INSTALL_PREFIX)/../ && tar cpf - $$(eval ls -d $$(cat $(INSTALL_PLIST) | sed -n 's?^.*:var/??p' | sed -e 's?^?{var,target/var}/?') 2>/dev/null)) | \
            tar xpf - -C $(STAGING_DIR)/var --strip-components=1 --transform='s!^target/!!' ; \
        fi
else
        @$(MSG) [DSM6] Copy target to staging
        @(mkdir -p $(STAGING_DIR) && cd $(STAGING_INSTALL_PREFIX) && tar cpf - $$(cat $(INSTALL_PLIST) | cut -d':' -f2)) | tar xpf - -C $(STAGING_DIR)
endif
```
This means that under DSM6 these files would not be packaged... thus `copy.mk` needs changing as well.

mreid-tt:
> > DSM7's var environment is indeed different,
> 
> The generic installer of spksrc ensures the `$(INSTALL_DIR)/var` folder for all DSM versions. So all DSM version have a DSM 7 compatible var folder.

I've also run a few tests with option 3 presented above and this also works just fine. We may want to consider this as an equally minor change that maintains the existing structural differences between DSM 6 and DSM 7 if that is a concern.

mk/spksrc.directories.mk
hgy59: we have the links for DSM 5 and 6 since #4727
And I can't remember any issue. The packages must ensure to use /var/packages/packagename/var and not the resolved link in the code an in scripts anyway.

IMHO copy_target should not need to handle a target/var folder, no package should have files from cross Makefile in the target/var folder (i.e. having :var in PLIST), such files must be copied by the spk Makefile (extra_install) directly to the staging dir.
The only difference is that under DSM 6 providing new file under var as *.new on package upgrade must be done explicit (like in service_preinst of ejabberd).

Not installing files to target/var is fixed for ejabberd is fixed in the subsequent commit.

th0ma7: > we have the links for DSM 5 and 6 since #4727 And I can't remember any issue. The packages must ensure to use /var/packages/packagename/var and not the resolved link in the code an in scripts anyway.

Gotcha, I figure they didn't existed when I added the var facility management.

> IMHO copy_target should not need to handle a target/var folder, no package should have files from cross Makefile in the target/var folder (i.e. having :var in PLIST), such files must be copied by the spk Makefile (extra_install) directly to the staging dir.

I disagree, while rather rare there are files provided from some of the cross such as:
```
spksrc@spksrc13:~/master/spksrc/cross$ grep var */PLIST
ejabberd/PLIST:rsc:var/ejabberd.yml
ejabberd/PLIST:rsc:var/ejabberdctl.cfg
ejabberd/PLIST:rsc:var/inetrc
imapfilter/PLIST:rsc:var/config.lua
imapfilter/PLIST:rsc:var/extend.lua
intel-libvpl/PLIST:rsc:etc/vpl/vars.sh
lm-sensors/PLIST:rsc:var/sensors3.conf
mutt/PLIST:rsc:var/Muttrc
mutt/PLIST:rsc:var/mime.types
mutt/PLIST:rsc:var/mime.types.dist
phpmemcachedadmin/PLIST:rsc:var/phpmemcachedadmin.config/
rsnapshot/PLIST:rsc:var/rsnapshot.conf.default
saltgui/PLIST:rsc:var/saltgui/**
slang/PLIST:lib:lib/slang/v2/modules/varray-module.so
smartmontools/PLIST:rsc:var/smartd.conf
smartmontools/PLIST:rsc:var/smartd_warning.sh
zap2epg/PLIST:rsc:var/epggrab/zap2epg/conf/zap2epg.xml
```

> The only difference is that under DSM 6 providing new file under var as *.new on package upgrade must be done explicit (like in service_preinst of ejabberd).

That's exactly what's being done automatically in `spksrc.service.installer.dsm7`:
```
#
# syncronize all files from var folder in spk (extracted to target/var)
# (@appdata) /var/packages/<package>/target/var -> /var/packages/<package>/var
#   
syno_sync_var_folder () {
    # take all files that do not alredy exist and rename the other files to *.new to avoid overwriting
    # existing configuration files.
    # finally remove the target/var folder that is not used under DSM 7.
    if [ -d ${SYNOPKG_PKGDEST}/var -a "$(ls -A ${SYNOPKG_PKGDEST}/var 2>/dev/null)" ]; then

        echo "Install files from var folder"

        # Move all files except existing
        echo "$RSYNC --ignore-existing --remove-source-files ${SYNOPKG_PKGDEST}/var/ ${SYNOPKG_PKGVAR}"
        $RSYNC --ignore-existing --remove-source-files ${SYNOPKG_PKGDEST}/var/ ${SYNOPKG_PKGVAR}

        # Rename any remaining (thus pre-existing) files with .new suffix
        find ${SYNOPKG_PKGDEST}/var -type f -exec sh -c 'x="{}"; mv "$x" "${x}.new"' \;
    
        # Move .new files (overwrite existing)
        echo "$RSYNC --remove-source-files ${SYNOPKG_PKGDEST}/var/ ${SYNOPKG_PKGVAR}"
        $RSYNC --remove-source-files ${SYNOPKG_PKGDEST}/var/ ${SYNOPKG_PKGVAR}

        # Remove var folder extraced from spk
        $RM ${SYNOPKG_PKGDEST}/var
    fi
}
```
 
> Not installing files to target/var is fixed for ejabberd is fixed in the subsequent commit.

Remaining issue from your change is that packages refererring to `rsc:var/<file>` in their `PLIST` simply won't be packaged anymore on DSM6.  Quick testing result in error - tested against `diyspk/lm-sensors`:
```
make[5]: Leaving directory '/home/spksrc/master/spksrc/diyspk/lm-sensors/work-aarch64-6.2.4/lm-sensors-3-6-0'
find /home/spksrc/master/spksrc/diyspk/lm-sensors/work-aarch64-6.2.4/install//var/packages/lm-sensors/target// \! -type d -printf '%P\n' | sed 's?^target/??g' | sort | \
  diff /home/spksrc/master/spksrc/diyspk/lm-sensors/work-aarch64-6.2.4/lm-sensors.plist.tmp -  | grep '>' | sed 's?> ??g' > /home/spksrc/master/spksrc/diyspk/lm-sensors/work-aarch64-6.2.4/lm-sensors.plist
make[4]: Leaving directory '/home/spksrc/master/spksrc/cross/lm-sensors'
make[3]: Leaving directory '/home/spksrc/master/spksrc/cross/lm-sensors'
===>  Creating target installation dir of lm-sensors
===>  [DSM6] Copy target to staging
tar: var/sensors3.conf: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
make[2]: *** [../../mk/spksrc.copy.mk:59: copy_target] Error 2
make[2]: Leaving directory '/home/spksrc/master/spksrc/diyspk/lm-sensors'
===>  20260111-150225 MAKELEVEL: 01, PARALLEL_MAKE: max, ARCH: aarch64-6.2.4, NAME: lm-sensors [END]
make[1]: *** [../../mk/spksrc.supported.mk:73: build-arch-aarch64-6.2.4] Error 1
make[1]: Leaving directory '/home/spksrc/master/spksrc/diyspk/lm-sensors'
make: *** [../../mk/spksrc.supported.mk:62: arch-aarch64-6.2.4] Error 2
```

The following fixes things and allows using the same behavior for all DSM:
```
spksrc@spksrc13:~/master/spksrc/mk$ git diff spksrc.copy.mk
diff --git a/mk/spksrc.copy.mk b/mk/spksrc.copy.mk
index 0a1c32b82..2181eec77 100644
--- a/mk/spksrc.copy.mk
+++ b/mk/spksrc.copy.mk
@@ -44,20 +44,15 @@ pre_copy_target: copy_msg
 copy_target: SHELL:=/bin/bash
 copy_target: .SHELLFLAGS := -o pipefail -c
 copy_target: $(PRE_COPY_TARGET) $(INSTALL_PLIST)
-	@$(MSG) [DSM7+] Copy target to staging, discard var directory
+	@$(MSG) "Copy target to staging, discard var directory"
 	@(mkdir -p $(STAGING_DIR) && cd $(STAGING_INSTALL_PREFIX) && tar cpf - $$(cat $(INSTALL_PLIST) | sed -e '/^.*:var\/.*/d' -e 's/^.*://g')) | \
 	  tar xpf - -C $(STAGING_DIR)
-	@$(MSG) "[DSM7+] Copy and merge var and target/var to $(STAGING_DIR)/var"
+	@$(MSG) "Copy and merge var and target/var to $(STAGING_DIR)/var"
 	@if [ "$$(cat $(INSTALL_PLIST) | sed -n 's?^.*:var/??p')" ] ; then \
 	  mkdir -p $(STAGING_DIR)/var ; \
 	  (cd $(STAGING_INSTALL_PREFIX)/../ && tar cpf - $$(eval ls -d $$(cat $(INSTALL_PLIST) | sed -n 's?^.*:var/??p' | sed -e 's?^?{var,target/var}/?') 2>/dev/null)) | \
 	    tar xpf - -C $(STAGING_DIR)/var --strip-components=1 --transform='s!^target/!!' ; \
 	fi
-else
-	@$(MSG) [DSM6] Copy target to staging
-	@(mkdir -p $(STAGING_DIR) && cd $(STAGING_INSTALL_PREFIX) && tar cpf - $$(cat $(INSTALL_PLIST) | cut -d':' -f2)) | tar xpf - -C $(STAGING_DIR)
-endif
 
 post_copy_target: $(COPY_TARGET)
 
```
I also suggest this `$(abspath $(INSTALL_PREFIX)/../var)` such as:
```
spksrc@spksrc13:~/master/spksrc/mk$ git diff spksrc.directories.mk
diff --git a/mk/spksrc.directories.mk b/mk/spksrc.directories.mk
index 6f9a39034..9e7bc5b9a 100644
--- a/mk/spksrc.directories.mk
+++ b/mk/spksrc.directories.mk
@@ -61,10 +61,8 @@ endif
 # when invoking make from under spk/*.  Setting var when
 # test-building dependencies from under cross/* is unecessary.
 #
 ifeq ($(lastword $(subst /, ,$(INSTALL_PREFIX))),target)
-INSTALL_PREFIX_VAR = $(shell dirname $(INSTALL_PREFIX))/var
-endif
+INSTALL_PREFIX_VAR = $(abspath $(INSTALL_PREFIX)/../var)
 endif
 ifeq ($(strip $(INSTALL_PREFIX_VAR)),)
 INSTALL_PREFIX_VAR  = $(INSTALL_PREFIX)/var
```

mreid-tt: Superseded by #6906

mreid-tt:
hey @hgy59 — hope you don’t mind, I was trying to help get this one over the line and ready for publishing. I saw your earlier concern about RabbitMQ not working on DSM 7, so I did some upgrade testing on my end.

During testing, I found that `SPK_DEPENDS` needed consistent decimal version formatting; otherwise, upgrades would fail. I’ve adjusted that accordingly. I also added wrappers for the existing CLI tools to make RabbitMQ easier to test, and set `HOME` to `SYNOPKG_PKGVAR` so Erlang can correctly locate the cookie file.

Based on my DSM 7.1 testing, everything appears to be working now.

mreid-tt:
hey @hgy59, I did some further investigation and noted that for ejabberd the upgrade process doesn't seem to work properly. Here are the two scenarios I noted:

For DSM 6: The default admin is migrated but not enabled as an admin because this section is missing in the `ejabberd.yml` file:
```
acl:
  admin:
    user:
      - "admin@localhost"
```

For DSM 7: The `ejabberd.yml` file is also missing the addition of the above section but the default admin account does not seem to be migrated at all. In the log file (even when I manually add the above section), I see this:
```
[warning] <0.547.0>@ejabberd_web_admin:process2/2:215 Access of <<"admin@localhost">> from <<"::ffff:192.168.10.159">> failed with error: <<"inexistent-account">>
```

So there seems to be a bit more work needed where upgrades are concerned. New installs seem to work with DSM 6 and 7.

mreid-tt:
## Summary of Changes

I've tested and extended this PR with additional fixes for upgrade scenarios for ejabberd. The changes are added in the commit https://github.com/SynoCommunity/spksrc/pull/5829/commits/8ee216ab25c6da13ae95f48b9b0a3ab136e1ee57.

### New Features

**Admin ACL Migration for Upgrades**
- Added dynamic upgrade wizard (`upgrade_uifile.sh`) that detects when the admin ACL is missing from `ejabberd.yml`
- Wizard prompts for existing admin credentials and validates them against the ejabberd database before proceeding
- On successful validation, the admin ACL section is automatically added to the config file

**CLI Wrapper Script**
- Added `bin/ejabberdctl` wrapper for `SPK_COMMANDS` that handles running as the package user
- Wrapper uses `sudo -u sc-ejabberd` with proper HOME environment for cookie access
- Provides user-friendly prompts when sudo authentication is required

### Bug Fixes

**DSM 7 Data Migration**
- Fixed HOME path from `SYNOPKG_PKGDEST` to `SYNOPKG_PKGVAR` for proper cookie location on DSM 7
- Added `service_save` and `service_restore` functions to migrate data from `target/var` to `var` (required because DSM 7 uses different paths: `@appstore` vs `@appdata`)
- Cookie file is also migrated from old HOME location (`target/.erlang.cookie`) to new location (`var/.erlang.cookie`)

**Upgrade Validation**
- `validate_preupgrade` temporarily starts ejabberd to validate admin credentials against the live database
- Uses correct HOME path based on where the erlang cookie exists (handles both old and new installations)
- Properly stops ejabberd after validation regardless of success or failure

### Technical Details

- Upgrade wizard only appears when `acl: admin: user:` is missing or empty in the existing config
- Config file detection handles both DSM 6 (`/var/packages/ejabberd/var/`) and DSM 7 (`/var/packages/ejabberd/target/var/`) paths
- AWK-based config modification inserts admin ACL section after `loopback:` block within the `acl:` section
- All ejabberdctl commands during upgrade use hardcoded installed paths since `SYNOPKG_PKGDEST` points to the new package being installed, not the current installation

### Testing

Built and tested on:
- DSM 7.1 x64 - upgrade from previous version with missing admin ACL ✓
- DSM 6.2.4 x64 - upgrade from previous version with missing admin ACL ✓

mreid-tt:
hey @hgy59 - I've reviewed, tested, and enhanced this PR. The changes are now ready for merging.

**Testing completed:**
- Fresh installs of ejabberd and RabbitMQ on DSM 7.1 x64
- Upgrades from existing installed versions on DSM 7.1 x64
- Verified ejabberd admin ACL migration wizard works correctly
- Verified RabbitMQ data migration during upgrade
- Confirmed web interface shortcuts function properly for both packages

**Functional testing:**

*Erlang:*
```
$ erl -s erlang halt
Erlang/OTP 25 [erts-13.2.2.16] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:1] [jit:ns]
```

*RabbitMQ:*
```
$ rabbitmqctl status
Status of node rabbit@DSM71 ...
RabbitMQ version: 3.11.28
Erlang configuration: Erlang/OTP 25 [erts-13.2.2.16] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:1] [jit:ns]
Crypto library: OpenSSL 3.5.4 30 Sep 2025
Plugins: rabbitmq_management, rabbitmq_management_agent, rabbitmq_web_dispatch, amqp_client, cowboy, cowlib
```

*ejabberd:*
```
$ ejabberdctl status
The node ejabberd@localhost is started with status: started
ejabberd 23.10 is running in that node
```

**Additional improvements made:**
- ejabberd: Dynamic upgrade wizard that detects missing admin ACL and validates credentials before upgrade proceeds
- ejabberd: Automatic migration of Erlang cookie and data from old paths (`target/var`) to new (`var`) on DSM 7
- ejabberd: CLI wrapper (`ejabberdctl`) with proper HOME and permission handling
- RabbitMQ: DSM 7 upgrade data migration for cookie and var folder
- RabbitMQ: CLI wrappers for all management tools with proper permission handling
- RabbitMQ: Web interface shortcut to Management UI (port 15672)
- Reorganized Makefiles for consistency

The PR description has been updated to reflect all changes. Do you have any other comments or changes before we merge?

PR #5862, opened 2023-08-30
tt-rss: upgrade 20221230 => 20230828

smaarn:
## Description

Upgrade tt-rss to latest version and migrate wizard to wizard templates

## Checklist

### Type of change

- [x] Package update


mreid-tt:
hey @smaarn, I was doing a review of this for you and I found that I was able to successfully install, upgrade and run this under DSM 7. The only thing I noted was that the service icon was a generic one.

Under DSM 6, during installation it seems to require the user to manually edit the default PHP profile. I also noted that the HTML code for the formatting on the wizard is showing up as text (which was not the case in the previous version). As for the default PHP profile modification under DSM 6, you might want to consider the workarounds I created for ownCloud and more recently for Selfoss (in https://github.com/SynoCommunity/spksrc/pull/5916). These allow for the creation of PHP profiles and other web settings in DSM 6. These are automatically created in DSM 7 using the resource worker but can be manually done under DSM 6 using the workarounds.

Under DSM 6, I noted however that the package does not start. The daemon which is triggered does not seem to like being run as root as per these entries in the `daemon.log`:
```
Please fix errors indicated by the following messages:

 * Please don't run this script as root.

You might want to check tt-rss wiki or the forums for more information.
Please search the forums before creating new topic for your question.
```

I was able to trigger a start using the command [`sudo -u http /var/packages/tt-rss/target/bin/tt-rss-daemon`] and then the program behaved correctly. Without this daemon running it does not seem to automatically refresh subscribed feeds. You can perhaps look at code in the [service-setup.sh](https://github.com/SynoCommunity/spksrc/blob/master/spk/owncloud/src/service-setup.sh) of ownCloud for examples of running as other users in DSM 6.

spk/tt-rss/Makefile
```diff
@@ -11,20 +11,17 @@ DESCRIPTION = Tiny Tiny RSS is an open source web-based news feed \(RSS/Atom\) r
 DESCRIPTION_FRE = Tiny Tiny RSS est un agrégateur et lecteur web de flux RSS et Atom open source conçu pour vous permettre de lire des nouvelles de n\'importe quel endroit, tout en se sentant aussi proche d\'une application de bureau que possible.
 ADMIN_URL = /tt-rss/
 DISPLAY_NAME = Tiny Tiny RSS
-CHANGELOG = "1. Update Tiny Tiny RSS to Git revision c30b24d09f (20221230) <br /> 2. Fixed broken DSM 7 initial installation procedure. <br /> 3. Upgrade PHP version to 7.4"
+CHANGELOG = "Update Tiny Tiny RSS to Git revision afd04d141c (20230828)"
 
 HOMEPAGE   = https://tt-rss.org/
 LICENSE    = GPL
```
mreid-tt, line 18: I noted in your build log for DSM 6 the following error:
```
  ===>  NOT creating /github/workspace/spk/tt-rss/work-dsm6/conf/privilege
  ===>  (privilege) DSM <= 6 and SERVICE_USER undefined
```

I would suggest that you add code similar to this to resolve this issue and allow the user `sc-tt-rss` to be created under DSM 6: https://github.com/SynoCommunity/spksrc/pull/5916/commits/cb744189368cbc21791bbc9e22c95d048cad29dc

spk/tt-rss/Makefile
```diff
@@ -11,20 +11,17 @@ DESCRIPTION = Tiny Tiny RSS is an open source web-based news feed \(RSS/Atom\) r
...
 
-WIZARDS_DIR = src/wizard/
+WIZARDS_TEMPLATES_DIR = src/wizard_templates
 CONF_DIR = src/conf/
 
 SERVICE_SETUP = src/service-setup.sh
```
mreid-tt, line 22: I noted in your build log for DSM 6 the following error:
```
  ===>  NOT creating /github/workspace/spk/tt-rss/work-dsm6/conf/privilege
  ===>  (privilege) DSM <= 6 and SERVICE_USER undefined
```

I would suggest that you add code similar to this to resolve this issue and allow the user `sc-tt-rss` to be created under DSM 6: https://github.com/SynoCommunity/spksrc/pull/5916/commits/cb744189368cbc21791bbc9e22c95d048cad29dc

mreid-tt:
@smaarn, another thing I noted was the prevalence of script files rather than consolidating most actions under the `service-setup.sh` as recommended by the [wiki](https://github.com/SynoCommunity/spksrc/wiki/DSM-7.0). Perhaps you can review a similar journey I took with ownCloud under https://github.com/SynoCommunity/spksrc/pull/5619 to simplify the script and daemon structures. Let me know if you need any assistance with this.

mreid-tt:
@smaarn, regarding the icon being generic in DSM 7, I believe this is because you don't have it defined in your resource file. Do take a look at the package developer guide or my example in ownCloud on its use.

P.S. As a suggestion you can also upgrade the icon in your package to the larger 512px one from [here](https://git.tt-rss.org/fox/tt-rss.git/plain/images/favicon-512px.png).

mreid-tt:
@smaarn, please note that I have updated the approach to DSM 6 in https://github.com/SynoCommunity/spksrc/pull/5916. The main change is to make all the web folders owned by the `http` user to avoid conflicts with different packages modifying the PHP FPM in DSM 6.

hgy59 (commented):
my package testing found some issues:

- wrong variable for mysql call
- on DSM 6 the db cannot be created 
- DSM 6 wizard files need tripple mustache brackets for `WEBSTATION_PHP_PROFILE_DESC` (install and upgrade)

--- 
in service-setup.sh the ttrss user pw must be used (not the db root pw):
replace
```
        "${MYSQL}" -u "${MYSQL_USER}" -p"${wizard_mysql_password_root}" "${MYSQL_DATABASE}" < "${WEB_DIR}/${PACKAGE}/schema/ttrss_schema_mysql.sql"
```
by
```
        "${MYSQL}" -u "${MYSQL_USER}" -p"${wizard_mysql_password_ttrss}" "${MYSQL_DATABASE}" < "${WEB_DIR}/${PACKAGE}/schema/ttrss_schema_mysql.sql"
```

otherwise there is an installation error logged:
```
2023/11/10 11:50:33     Begin service_postinst
2023/11/10 11:50:33     ERROR 1045 (28000): Access denied for user 'ttrss'@'localhost' (using password: YES)
```

_(since `ttrss_schema_mysql.sql` is an empty file, this did not break the installation)_

--- 

On DSM 6 the ttrss database tables are not created.
installation log shows:
```
2023/11/10 13:14:37     Begin service_postinst
2023/11/10 13:14:43     PHP Parse error:  syntax error, unexpected ':', expecting '{' in /volume1/web/tt-rss/update.php on line 23
2023/11/10 13:14:43     Parse error: syntax error, unexpected ':', expecting '{' in /volume1/web/tt-rss/update.php on line 23
2023/11/10 13:14:43     End service_postinst
```

extract of update.php (with line numbers):
```
    23          function make_stampfile(string $filename): bool {
    24                  $fp = fopen(Config::get(Config::LOCK_DIRECTORY) . "/$filename", "w");
    25
    26                  if (flock($fp, LOCK_EX | LOCK_NB)) {
    27                          fwrite($fp, time() . "\n");
    28                          flock($fp, LOCK_UN);
    29                          fclose($fp);
    30                          return true;
    31                  } else {
    32                          return false;
    33                  }
    34          }
```

I suppose that the guess_php function does not work as expected (on my DSM 6 the default php version is 5.6.11, on my working DSM 7.2 the default php version is 8.1.9)

confirmed that guess_php = `php56`
and php_options =  `-c /var/packages/WebStation/etc/php56/php.ini` 
on my DSM 6 installation.

For DSM 6 php 7.4 can be hard coded (in settings.sh), but for php_options it needs some effort (or as @mreid-tt proposed a php profile should be created too).

---

spk/tt-rss/Makefile
smaarn: Indeed. Don't know why it was broken in the first place. Fixing this.

smaarn:
> @smaarn, another thing I noted was the prevalence of script files rather than consolidating most actions under the `service-setup.sh` as recommended by the [wiki](https://github.com/SynoCommunity/spksrc/wiki/DSM-7.0). Perhaps you can review a similar journey I took with ownCloud under #5619 to simplify the script and daemon structures. Let me know if you need any assistance with this.

Will be having a look since there has been a few changes in the framework since but not too sure about it.

smaarn:
> > On DSM 6 the ttrss database tables are not created.
> 
> Ok, my fault. the default php profile in WebStation was the "Default Profile ( PHP 5.6 )" profile. After switching to "Default Profile (PHP 7.4)" the tt-rss successfully runs on DSM 6.
> 
> There are still installation log warnings, but the db is created sucessfully.
> 
> ```
> 2023/11/10 16:29:18     Begin service_postinst
> 2023/11/10 16:29:23     PHP Warning:  PHP Startup: Unable to load dynamic library 'mcrypt.so' (tried: /usr/local/lib/php74/modules/mcrypt.so (/usr/local/lib/php74/modules/mcrypt.so: cannot open shared object file: No such file or directory), /usr/local/lib/php74/modules/mcrypt.so.so (/usr/local/lib/php74/modules/mcrypt.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
> 2023/11/10 16:29:23     PHP Warning:  PHP Startup: Unable to load dynamic library 'mysql.so' (tried: /usr/local/lib/php74/modules/mysql.so (/usr/local/lib/php74/modules/mysql.so: cannot open shared object file: No such file or directory), /usr/local/lib/php74/modules/mysql.so.so (/usr/local/lib/php74/modules/mysql.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
> 2023/11/10 16:29:24     [15:29:24/27014] Lock: update.lock
> 2023/11/10 16:29:24     [15:29:24/27014] Proceeding to update without confirmation.
> 2023/11/10 16:29:24     [15:29:24/27014] Loading base database schema...
> 2023/11/10 16:29:51     [15:29:51/27014] Migration finished, current version: 147
> 2023/11/10 16:29:52     End service_postinst
> ```
> 
> anyway, the package should install a php profile for DSM 6 and not depend on the default profile,

Indeed. Will be having a look at the approach used by @mreid-tt on this topic. I thought at first it was an acceptable trade-off to be honest. 

Will be flagging this MR as in progress until this gets done.

Thanks a lot for your feedbacks @mreid-tt @hgy59 !

hgy59:
@smaarn another issue is, that the db backup at uninstall does not work (the db is already removed in postuninst).

This can easily be fixed by moving the backup from postuninst to preuninst like this:
```bash
service_preuninst ()
{
    # Export database
    if [ "${SYNOPKG_PKG_STATUS}" == "UNINSTALL" ]; then
        if [ -n "${wizard_dbexport_path}" ]; then
            ${MKDIR} ${wizard_dbexport_path}
            ${MYSQLDUMP} -u root -p"${wizard_mysql_password_root}" ${MYSQL_DATABASE} > ${wizard_dbexport_path}/${MYSQL_DATABASE}.sql
        fi
    fi
}

service_postuninst ()
{
    if [ "${SYNOPKG_DSM_VERSION_MAJOR}" -lt 7 ]; then
      # Remove the web interface
      ${RM} ${WEB_DIR}/${PACKAGE}
    fi
}
```

smaarn:
Ok so here goes the current status:

1. Uninstallation backup procedure is now fixed
2. Gave a crack at @mreid-tt 's way of handling things for the profile management on DSM 6 but unfortunately can't test it. Would need a guinea pig to test it and validate that the "automatic" way of detecting PHP binary and configuration work (if it doesn't work for the binary we can hardcode it but for the profile... I could need some help here :| )
3. Remove the weird attempt at running some SQL migration script (it's actually useless as pointed out by @hgy59 since the file is now... empty and we have an another procedure to run upgrades)
4. I added the mention of the icon the `conf/resource` file but I believe it should be related to the `app/config` resource and it's actually there. I'm at a loss here. 
5. Fixed mustache templates
6. Fixed DSM 6 user which wasn't declared appropriately

I'll try to see if during the week I can figure out this damn icon thing.

mreid-tt:
@smaarn, so the DSM 6 install kinda works but there are a couple of issues:

1. The Apache conf file seems to be missing for import as well as the code to import it
2. The files in the web directory need to be owned by the `http` user and group
3. The daemon still seems to be manually called in the Makefile. What this seems to result in is that even when you stop the package, the daemon is still writing to the log. Also when you try to re-start the package you get the error: `Can't create lockfile (update_daemon.lock)`
4. There is a lot of additional code in the Makefile to write out the PHP extensions to a file `/web/tt-rss.json` but you could have just put it into the source file `/src/6.0-/tt-rss.json`. This additional complexity is not needed, and the source file could just be in `/src/web` along with the missing Apache conf file
5. The daemon when called needs to run as the `http` user. This can be achieved by (a) adding a `privilege` file for DSM 6 which overrides the default run-as user to be `root` and (b) modifying the last line of the daemon to run as the correct user (e.g. `/bin/su "http" -s /bin/sh -c "exec $(guess_php) $(php_options) ${TTRSS} --daemon >>"${LOGS_DIR}/daemon.log" 2>&1"`).

For the last point, I would get rid of all the all the complexity as well as the dependencies for the daemon script. It attempts to evaluate a number of things which are not required. Additionally the function call for the daemon is expected to be very simple based on the [Update daemon](https://tt-rss.org/wiki/UpdatingFeeds#update-daemon) documentation.

For testing I use a VirtualDSM 6 deployment in Virtual Machine Manager. This is the easiest way I've found to test older DSM versions that I don't use in my production NAS. If you need any assistance with the above let me know.

smaarn:
> For testing I use a VirtualDSM 6 deployment in Virtual Machine Manager. This is the easiest way I've found to test older DSM versions that I don't use in my production NAS. If you need any assistance with the above let me know.

I have already tried that but since it only supports btrfs file system I'm pretty stuck here (unless I decide to upgrade completely or wipe out completely my personal NAS, both options not on the table at the moment).

> The Apache conf file seems to be missing for import as well as the code to import it (this results in the web page not rendering the PHP code)

Would you be able to tell me what an `ls -la /var/services/web/tt-rss/index.php` command returns ?
There has never been any real specific apache configuration file as far as I'm aware so it's weird.

> The files in the web directory need to be owned by the http user and group
As far as I remember only a few of them needed to be really owned by the http user and group. I could change that to be honest. Unfortunately it would need another round of testing.

> The daemon still seems to be manually called in the Makefile. As a result, even when you stop the package, the daemon doesn't actually stop and still writes to the log. Also when you try to re-start the package you get the error: Can't create lockfile (update_daemon.lock)

I think one is not necessarily related to the other but I need to check how the update daemon is now implemented to see what goes on here.

> There is a lot of additional code in the Makefile to write out the PHP extensions to a file /web/tt-rss.json but you could have just put it into the source file /src/6.0-/tt-rss.json. This additional complexity is not needed, and the source file could just be in /src/web along with the missing Apache conf file.

Actually the whole point was to avoid having to list the extensions lists in three different locations:

1. The command which runs the daemon (since it needs to have some PHP extensions enabled which are not enabled by default)
2. The DSM 7 configuration file
3. The DSM 6 specifics

>  The daemon when called needs to run as the http user. This can be achieved by (a) adding a privilege file for DSM 6 which overrides the default run-as user to be root and (b) modifying the last line of the daemon to run as the correct user (e.g. /bin/su "http" -s /bin/sh -c "exec $(guess_php) $(php_options) ${TTRSS} --daemon >>"${LOGS_DIR}/daemon.log" 2>&1"). I haven't been able to find another way since the SERVICE_USER is not allowed to be anything other than "auto"

I would have expected this to be handled natively by the framework. Will be looking into it.

hgy59:
The current version does not install correctly on DSM 6.

- install log shows error `--suffix: command not found`
- installation hangs on restart apache (tried to manually start apache in Package Center, but it shows `Another package is being configured, installed, or restored. Please try again later.`)

so the package center of my system is currently  locked...


install log:
```
2023/11/12 11:25:48     ===> Step postinst. STATUS=INSTALL USER=http GROUP= SHARE_PATH=
2023/11/12 11:25:48     Begin save_wizard_variables
2023/11/12 11:25:48     End save_wizard_variables
2023/11/12 11:25:48     Begin service_postinst
2023/11/12 11:26:03     /var/packages/tt-rss/scripts/service-setup: line 83: --suffix: command not found
2023/11/12 11:26:03     Set Apache as the back-end server
2023/11/12 11:26:04     Add PHP profile for tt-rss
2023/11/12 11:26:04     Restart Apache to load new configs
```

mreid-tt:
> > The Apache conf file seems to be missing for import as well as the code to import it (this results in the web page not rendering the PHP code)
> 
> Would you be able to tell me what an `ls -la /var/services/web/tt-rss/index.php` command returns ? There has never been any real specific apache configuration file as far as I'm aware so it's weird.

This Apache conf file was required as part of my workaround. The file is supposed to be placed in `/usr/local/etc/apache24/sites-enabled/` and content is supposed to look like this:

```
Alias "/tt-rss" "/var/services/web/tt-rss"

<Proxy "unix:/run/php-fpm/php-com-synocommunity-packages-tt-rss.sock|fcgi://tt-rss">
    ProxySet timeout=3600
</Proxy>

<Directory "/var/services/web/tt-rss">

	<FilesMatch "\.(php[345]?|phtml)$">
		SetHandler "proxy:fcgi://tt-rss"
	</FilesMatch>

	<IfModule dir_module>
		DirectoryIndex index.php index.htm index.html
	</IfModule>
</Directory>
```

For the command you wanted me to run, please note the file ownership would look different because I'd already changed it to `http:http`. Nonetheless the output is:

```
$ ls -la /var/services/web/tt-rss/index.php
-rwxrwxr-x 1 http http 11519 Aug 28 02:58 /var/services/web/tt-rss/index.php
```

> > The files in the web directory need to be owned by the http user and group
> 
> As far as I remember only a few of them needed to be really owned by the http user and group. I could change that to be honest. Unfortunately it would need another round of testing.

Perhaps you are right. I just change all the files in the web directory to `http:http` for safety. The testing will only be required for DSM 6 since the ownership change should not be done for DSM 7. The code examples I have to change ownership only triggers under DSM 6.

> > The daemon still seems to be manually called in the Makefile. As a result, even when you stop the package, the daemon doesn't actually stop and still writes to the log. Also when you try to re-start the package you get the error: Can't create lockfile (update_daemon.lock)
> 
> I think one is not necessarily related to the other but I need to check how the update daemon is now implemented to see what goes on here.

You may be right. I've tried a variation where I use the [Custom service process startup](https://github.com/SynoCommunity/spksrc/wiki/Service-Support#custom-service-process-startup) example to modify the service setup to look like this:

```
service_prestart ()
{
	COMMAND="${SYNOPKG_PKGDEST}/bin/tt-rss-daemon"
	su http -s /bin/sh -c "${COMMAND}" >> ${LOG_FILE} 2>&1 &
	echo "$!" > "${PID_FILE}"
}
```

... and then have a cut down daemon command that just has:

```
exec $(guess_php) $(php_options) ${TTRSS} --daemon >>"${LOGS_DIR}/daemon.log" 2>&1
```

... and this seems to resolve the issue of the daemon still running after I stop the package in the Package Center.

> > There is a lot of additional code in the Makefile to write out the PHP extensions to a file /web/tt-rss.json but you could have just put it into the source file /src/6.0-/tt-rss.json. This additional complexity is not needed, and the source file could just be in /src/web along with the missing Apache conf file.
> 
> Actually the whole point was to avoid having to list the extensions lists in three different locations:
> 
> 1. The command which runs the daemon (since it needs to have some PHP extensions enabled which are not enabled by default)
> 2. The DSM 7 configuration file
> 3. The DSM 6 specifics

Hmm, I do only think the extensions are only required in the DSM 6 and DSM 7 config files. Launching using the recommedation in [Update daemon](https://tt-rss.org/wiki/UpdatingFeeds#update-daemon) like this has no issue:

```
sudo -u http php74 /var/services/web/tt-rss/update.php --daemon
```

> > The daemon when called needs to run as the http user. This can be achieved by (a) adding a privilege file for DSM 6 which overrides the default run-as user to be root and (b) modifying the last line of the daemon to run as the correct user (e.g. /bin/su "http" -s /bin/sh -c "exec (guessphp)(php_options) ${TTRSS} --daemon >>"${LOGS_DIR}/daemon.log" 2>&1"). I haven't been able to find another way since the SERVICE_USER is not allowed to be anything other than "auto"
> 
> I would have expected this to be handled natively by the framework. Will be looking into it.

This was a bad take on my part and the example above is a better solution.

mreid-tt:
@smaarn, I noted that you closed this PR. Would you be opening a new one or would you like me to continue your work under a new PR?

smaarn:
> @smaarn, I noted that you closed this PR. Would you be opening a new one or would you like me to continue your work under a new PR?

@mreid-tt I won't be having time to look at this today. Closing the MR was an error, sorry.

Regarding the extensions the way it works is that tt rss update script actually forks another php process (which is set to be virtual php in order to be able to configure the extensions)
By itself the update script can be launched natively but I thought it would be simpler to use the "same" php as much as possible.

Will be trying to have a look over the coming weeks.

hgy59:
@mreid-tt, @smaarn yesterday I had a version fully working on DSM 6.

- had to add `SERVICE_USER = auto` since a custom privilege file in DSM 6 does not work out of the box (and we need root for all stages except start/stop).
- so the service will run ass ${EFF_USER} and I had to run `update-schema` with this user (for correct file permissions on the lock file)
- added the variable `GROUP=http` in service-setup.sh
- the fix permissions (DSM 6) part uses `chown ${EFF_USER}:${GROUP}...`
- the chmod +x ...index.php was not required
- and it needed a fix in `mk/spksrc.service.start-stop-status` to run the service as package user 
  `if [ -n "${USER}" -a -n "${SYNOPKG_DSM_VERSION_MAJOR}" -a "$SYNOPKG_DSM_VERSION_MAJOR" -lt 6 ]; then`
  must be
  `if [ -n "${USER}" -a -n "${SYNOPKG_DSM_VERSION_MAJOR}" -a "$SYNOPKG_DSM_VERSION_MAJOR" -lt 7 ]; then`

so my final part of the working service_postinst was
```bash
    if [ "${SYNOPKG_DSM_VERSION_MAJOR}" -lt 7 ]; then
      # permissions must be adjusted before update-schema is called
      chown -R "${EFF_USER}:${GROUP}" "${WEB_DIR}/${PACKAGE}"
      chown -R "${EFF_USER}:${GROUP}" "${LOGS_DIR}"
      chmod -R 777 "${WEB_DIR}/${PACKAGE}/lock"
      chmod -R 777 "${WEB_DIR}/${PACKAGE}/feed-icons"
      chmod -R 777 "${WEB_DIR}/${PACKAGE}/cache"
    fi

      if [ "${SYNOPKG_DSM_VERSION_MAJOR}" -lt 7 ]; then
        # run as package user when installer runs as root
        SU="su ${EFF_USER} -s"
      else
        SU=""
      fi
      $SU /bin/sh -c ${SYNOPKG_PKGDEST}/bin/update-schema
    fi
```


If we do not get the DSM 6 profile working, it would be better to depend on the default php profile.
Another approach I found in phpMyAdmin package of Synology. In this package for DSM 6, a php.ini file contained in the package is copied into a specific folder (but this is done for nginx, not apache, so it may or may not work for apache 2.4)

smaarn:
@hgy59 I was wondering if

> `if [ -n "${USER}" -a -n "${SYNOPKG_DSM_VERSION_MAJOR}" -a "$SYNOPKG_DSM_VERSION_MAJOR" -lt 6 ]; then`
> must be
> `if [ -n "${USER}" -a -n "${SYNOPKG_DSM_VERSION_MAJOR}" -a "$SYNOPKG_DSM_VERSION_MAJOR" -lt 7 ]; then`

Shouldn't be actually instead:

> `if [ -n "${USER}" -a -n "${SYNOPKG_DSM_VERSION_MAJOR}" -a "$SYNOPKG_DSM_VERSION_MAJOR" -gt 6 ]; then`

I reverted the changes enforcing the DSM 6 appropriate WebStation configuration to get back to a "sound" state. @mreid-tt feel free to elaborate from here if you got time. At this stage I'm going to "sleep on it" and maybe check during the week if I manage to find some spare cycles (this time for real).

hgy59:
> Shouldn't be actually instead:
> 
> > `if [ -n "${USER}" -a -n "${SYNOPKG_DSM_VERSION_MAJOR}" -a "$SYNOPKG_DSM_VERSION_MAJOR" -gt 6 ]; then`

No, under DSM 7 only the privilege file defines the user (package user) to run the services and su is not executable due to restrictions.

the whole block would look like:
_the else block is for `user is set by conf/privilege` for DSM 6 **and** DSM 7_
```bash
        if [ -n "${service}" ]; then
            if [ -n "${SVC_NO_REDIRECT}" ]; then
                OUT="/dev/null"
            else
                OUT="${LOG_FILE}"
            fi
            if [ -n "${USER}" -a "$SYNOPKG_DSM_VERSION_MAJOR" -lt 7 ]; then
                if [ -z "${SVC_CWD}" ]; then
                    SVC_CD=""
                else
                    SVC_CD="cd ${SVC_CWD}; "
                fi
                if [ "$(whoami)" = "root" ]; then
                    SU="su ${EFF_USER} -s"
                else
                    SU=""
                fi
                if [ -z "${SVC_BACKGROUND}" ]; then
                    $SU /bin/sh -c "${SVC_CD}${service}" >> ${OUT} 2>&1
                else
                    $SU /bin/sh -c "${SVC_CD}${service}" >> ${OUT} 2>&1 &
                fi
            else
                # DSM user is set by conf/privilege
                if [ -n "${SVC_CWD}" ]; then
                    cd "${SVC_CWD}"
                fi
                if [ -z "${SVC_BACKGROUND}" ]; then
                    ${service} >> ${OUT} 2>&1
                else
                    ${service} >> ${OUT} 2>&1 &
                fi
            fi
            if [ -n "${SVC_WRITE_PID}" -a -n "${SVC_BACKGROUND}" -a -n "${PID_FILE}" ]; then
                [ $i -eq 1 ] && echo -ne "$!" > ${PID_FILE} || echo -ne " $!" >> ${PID_FILE}
            else
                wait_for_status 0 ${SVC_WAIT_TIMEOUT:=20}
            fi
        fi
```

mreid-tt:
> I reverted the changes enforcing the DSM 6 appropriate WebStation configuration to get back to a "sound" state. @mreid-tt feel free to elaborate from here if you got time. At this stage I'm going to "sleep on it" and maybe check during the week if I manage to find some spare cycles (this time for real).

@smaarn, I cloned your PR into a new one (https://github.com/SynoCommunity/spksrc/pull/5923) and committed a number of changes to implement some things including:

1. Added the `STARTABLE` variable to the Makefile
2. Added alternate conf dir for DSM 6 to the Makefile
3. Deleted all legacy scripts from `src/bin` and removed references in the Makefile
4. Added the `privilege` and `resource` files for DSM 6 and reformatted the JSON files
5. Modified the service setup as follows:
   - Moved the log directory to reside with the web package files
   - Changed PHP reference to a static PHP7.4 path
   - Created function to handle update schema requests
   - Modified the `service_prestart ()` to be a self-contained daemon
   - Re-introduced the logic for DSM 6 web configuration
   - Enhanced web configuration with correct rsync arguments for compact code (continuing your efforts)
   - Included web configuration check to work around multiple PHP profiles
   - Moved function `service_preupgrade ()` to `service_save ()` to ensure files are backed up
   - Moved function `service_postupgrade ()` to `service_restore ()` for consistent restoration
6. Added hi-res icon for package
7. Removed PHP config references from install and upgrade wizards

The summary of code changes are shown [here](https://github.com/SynoCommunity/spksrc/compare/SynoCommunity:397c7a1de366543f60b2a42478a1f8ae1523f8cd...SynoCommunity:5dfface97bd2b37c19c7479e466ee99680ab6f6c).

Thus far on DSM 6 the package is able to successfully install and configure rss feeds; upgrade with the same version and retain configurations; and uninstall, cleaning up Apache and PHP configs. Once you are comfortable you can merge the new commits into your PR.

Some things remain to be refined in the package based on my observation:

1. The upgrade, the save function seeks to backup files `/var/services/web/tt-rss/feed-icons/*.ico`, however I've noted that (a) the cached feed icons seem to be in `/var/services/web/tt-rss/cache/feed-icons/` and (b) they are just numbered files without file extensions
2. On upgrade, the restore function seeks to move files from `@tmp/synopkg.tmp/tt-rss/var/tt-rss/feed-icons/*.ico` and `@tmp/synopkg.tmp/tt-rss/var/tt-rss/themes.local/*` however these may not exist and create errors in the log
3. The upgrade wizard screen for when there is no migration required is not required in my view. Unless the keys in the function `without_migration` are required, that step can be removed. If the keys are required then some other text can be added
4. I haven't done any testing with upgrades from previous versions of `tt-rss` on DSM 6, nor have I done any testing under DSM 7 (though I don't expect it to work any differently)

mreid-tt:
Closed to PR https://github.com/SynoCommunity/spksrc/pull/5923

Issue #5847, opened 2023-08-15
sabnzbd 4.0.3-64 core dumped at startup on qoriq


zaadstra:
### Is this a new Bug?

- [X] I checked that the bug hasn't been reported before

### Package Name

sabnzbd

### Package Version

sabnzbd 4.0.3-64

### Device Model

DS213+ Qoriq

### Device Architecture

PPC

### Firmware Version

6.2.4 latest

### What happened?

Update sabnzbd 4.0.3-63 to sabnzbd 4.0.3-64.  Then got an error "cannot start package service".

Some analysis has been done in [issue 5841](https://github.com/SynoCommunity/spksrc/issues/5841). Issue seemed related at first but it is different. First thougts by @th0ma7 

> Your pip calls are working out just fine, meaning you do have openssl working ok. The issue in your case really is the core dumped which I presume is due to cryptography wheel. I'll have to investigate a bit further.

I uploaded the logs to pastebin, both for Python 3.11 and sabnzbd.

python311.log ds213+ https://pastebin.com/Lw5NGb1S

sabnzbd.log ds213+ https://pastebin.com/uCqQFh04

2023/07/22 00:01:44 sabnzbd 4.0.2-62 -> 63
2023/08/14 13:25:52 sabnzbd 4.0.3-63 -> 64

(fails to start)
2023/08/14 13:35:38 start sabnzbd 4.0.3-64 Begin start-stop-status start
/var/packages/sabnzbd/scripts/start-stop-status: line 31: 15558 Illegal instruction (core dumped) ${service} >> ${OUT} 2>&1
2023/08/14 13:36:05 start sabnzbd 4.0.3-64 End start-stop-status start ret=[1]
2023/08/14 13:36:06 (system) trigger sabnzbd 4.0.3-64 Begin start-stop-status stop
2023/08/14 13:36:06 (system) trigger sabnzbd 4.0.3-64 End start-stop-status stop ret=[0]

(reboot)

2023/08/14 13:55:26 start sabnzbd 4.0.3-64 Begin start-stop-status start
/var/packages/sabnzbd/scripts/start-stop-status: line 31: 12265 Illegal instruction (core dumped) ${service} >> ${OUT} 2>&1
2023/08/14 13:55:56 start sabnzbd 4.0.3-64 End start-stop-status start ret=[1]
2023/08/14 13:55:57 (system) trigger sabnzbd 4.0.3-64 Begin start-stop-status stop
[...]

PR #5879, opened 2023-09-11
Enable rust powerpc-unknown-linux-gnuspe for qoriq

th0ma7:
## Description

1. Add `powerpc-unknown-linux-gnuspe` rust toolchain for properly supporting `qoriq` arch
2. Enable `debug_symbols` for qoriq (and x64 to help diagnosing) - TEMPORARY

Helps debugging #5684 and #5847
Fixes #

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### Objective
Objective is when building not getting `lwsync`:
```
powerpc-e500v2-linux-gnuspe-objdump -d ./cryptography/hazmat/bindings/_rust.abi3.so | grep lwsync
```

### Additional Change(s)
Introduction of  `$(ADDITIONAL_RUSTFLAGS)` that allows passing, well, additional flags to rustc at build time, similarly to other existing flags.

### DIFF
--- a/toolchain/syno-qoriq-6.2.4/Makefile
+++ b/toolchain/syno-qoriq-6.2.4/Makefile
@@ -19,6 +19,6 @@ include ../../mk/spksrc.tc.mk
 qoriq_post_fix_target:
 	@cd $(WORK_DIR)/$(TC_TARGET)/bin ; \
 	for gnutool in $$(ls -1); do \
-	  [ ! -h "powerpc-linux-gnuspe-$${gnutool##*-}" ] && ln -s $${gnutool} "powerpc-linux-gnuspe-$${gnutool##*-}" ; \
-	  [ ! -h "powerpc-unknown-linux-gnuspe-$${gnutool##*-}" ] && ln -s $${gnutool} "powerpc-unknown-linux-gnuspe-$${gnutool##*-}" ; \
+	  [ ! -L "powerpc-linux-gnuspe-$${gnutool##*-}" ] && ln -sf $${gnutool} "powerpc-linux-gnuspe-$${gnutool##*-}" ; \
+	  [ ! -L "powerpc-unknown-linux-gnuspe-$${gnutool##*-}" ] && ln -sf $${gnutool} "powerpc-unknown-linux-gnuspe-$${gnutool##*-}" ; \
 	done
--- a/toolchain/syno-qoriq-6.2.4/Makefile
+++ b/toolchain/syno-qoriq-6.2.4/Makefile
@@ -9,6 +9,16 @@ TC_DIST_SITE_PATH = PowerPC%20QorIQ%20Linux%202.6.32
 TC_TARGET = powerpc-e500v2-linux-gnuspe
 TC_SYSROOT = $(TC_TARGET)/sysroot
 TC_EXTRA_CFLAGS = -mcpu=8548 -mhard-float -mfloat-gprs=double
-TC_RUSTFLAGS = -C target-cpu=e500 target-feature=-hard-float,-spe
+TC_RUSTFLAGS = -Ctarget-cpu=e500
+
+POST_FIX_TARGET = qoriq_post_fix_target
 
 include ../../mk/spksrc.tc.mk
+
+.PHONY: qoriq_post_fix_target
+qoriq_post_fix_target:
+	@cd $(WORK_DIR)/$(TC_TARGET)/bin ; \
+	for gnutool in $$(ls -1); do \
+	  [ ! -h "powerpc-linux-gnuspe-$${gnutool##*-}" ] && ln -s $${gnutool} "powerpc-linux-gnuspe-$${gnutool##*-}" ; \
+	  [ ! -h "powerpc-unknown-linux-gnuspe-$${gnutool##*-}" ] && ln -s $${gnutool} "powerpc-unknown-linux-gnuspe-$${gnutool##*-}" ; \
+	done
--- a/toolchain/syno-qoriq-6.2.4/Makefile
+++ b/toolchain/syno-qoriq-6.2.4/Makefile
@@ -9,5 +9,6 @@ TC_DIST_SITE_PATH = PowerPC%20QorIQ%20Linux%202.6.32
 TC_TARGET = powerpc-e500v2-linux-gnuspe
 TC_SYSROOT = $(TC_TARGET)/sysroot
 TC_EXTRA_CFLAGS = -mcpu=8548 -mhard-float -mfloat-gprs=double
+TC_RUSTFLAGS = -Ctarget-cpu=e500
 
 include ../../mk/spksrc.tc.mk
--- a/toolchain/syno-qoriq-6.2.4/Makefile
+++ b/toolchain/syno-qoriq-6.2.4/Makefile
@@ -9,6 +9,6 @@ TC_DIST_SITE_PATH = PowerPC%20QorIQ%20Linux%202.6.32
 TC_TARGET = powerpc-e500v2-linux-gnuspe
 TC_SYSROOT = $(TC_TARGET)/sysroot
 TC_EXTRA_CFLAGS = -mcpu=8548 -mhard-float -mfloat-gprs=double
-TC_RUSTFLAGS = -Ctarget-cpu=e500
+TC_RUSTFLAGS = -C target-cpu=e500 target-feature=-hard-float,-spe
 
 include ../../mk/spksrc.tc.mk
--- a/toolchain/syno-qoriq-6.2.4/Makefile
+++ b/toolchain/syno-qoriq-6.2.4/Makefile
@@ -9,7 +9,7 @@ TC_DIST_SITE_PATH = PowerPC%20QorIQ%20Linux%202.6.32
 TC_TARGET = powerpc-e500v2-linux-gnuspe
 TC_SYSROOT = $(TC_TARGET)/sysroot
 TC_EXTRA_CFLAGS = -mcpu=8548 -mhard-float -mfloat-gprs=double
-TC_RUSTFLAGS = -Ctarget-cpu=e500
+TC_RUSTFLAGS = -C target-cpu=e500
 
 POST_FIX_TARGET = qoriq_post_fix_target
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 	bison \
 	build-essential \
 	check \
-	clang-13 \
 	cmake \
 	curl \
 	cython3 \
@@ -51,7 +50,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 	libssl-dev \
 	libtool \
 	libunistring-dev \
-	llvm-13 \
 	lzip \
 	mercurial \
 	moreutils \
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 	bison \
 	build-essential \
 	check \
+	clang-13 \
 	cmake \
 	curl \
 	cython3 \
@@ -50,6 +51,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 	libssl-dev \
 	libtool \
 	libunistring-dev \
+	llvm-13 \
 	lzip \
 	mercurial \
 	moreutils \
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -19,20 +19,15 @@ endif
 #   https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html#stage-2-the-truly-current-compiler
 RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 2
 
-ifeq ($(TC_RUSTUP_TOOLCHAIN),)
-TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
+# When calling directly from
+# toolchain/syno-<arch>-<vertion>
+# ARCH variable is still unset
+ifeq ($(ARCH),)
+ARCH=$(TC_ARCH)
 endif
 
-RUST_TARGET =
 # map archs to rust targets
-ifeq ($(findstring $(ARCH), $(x64_ARCHS)),$(ARCH))
-RUST_TARGET = x86_64-unknown-linux-gnu
-endif
-ifeq ($(findstring $(ARCH), $(i686_ARCHS)),$(ARCH))
-RUST_TARGET = i686-unknown-linux-gnu
-endif
 ifeq ($(findstring $(ARCH), $(ARMv5_ARCHS)),$(ARCH))
-# may be not supported for cargo
 RUST_TARGET = armv5te-unknown-linux-gnueabi
 endif
 ifeq ($(findstring $(ARCH), $(ARMv7_ARCHS)),$(ARCH))
@@ -45,9 +40,22 @@ ifeq ($(findstring $(ARCH), $(ARMv8_ARCHS)),$(ARCH))
 RUST_TARGET = aarch64-unknown-linux-gnu
 endif
 ifeq ($(findstring $(ARCH), $(PPC_ARCHS)),$(ARCH))
+RUST_FORCE_TOOLCHAIN = 0
 RUST_TARGET = powerpc-unknown-linux-gnuspe
 TC_RUSTUP_TOOLCHAIN = stage$(RUSTUP_DEFAULT_TOOLCHAIN_STAGE)-$(RUST_TARGET)
 endif
+ifeq ($(findstring $(ARCH), $(x64_ARCHS)),$(ARCH))
+RUST_TARGET = x86_64-unknown-linux-gnu
+endif
+ifeq ($(findstring $(ARCH), $(i686_ARCHS)),$(ARCH))
+RUST_TARGET = i686-unknown-linux-gnu
+endif
+
 ifeq ($(RUST_TARGET),)
 $(error Arch $(ARCH) not supported)
+endif
+
+# By default use the default toolchain if unset
+ifeq ($(TC_RUSTUP_TOOLCHAIN),)
+TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
 endif
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -14,6 +14,10 @@ ifeq ($(RUSTUP_DEFAULT_TOOLCHAIN),)
 RUSTUP_DEFAULT_TOOLCHAIN = stable
 endif
 
+# When building toolchain Tier-3 arch support
+#   https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html#stage-2-the-truly-current-compiler
+RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 1
+
 ifeq ($(TC_RUSTUP_TOOLCHAIN),)
 TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
 endif
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -38,7 +38,7 @@ ifeq ($(findstring $(RUST_ARCH), $(ARMv8_ARCHS)),$(RUST_ARCH))
 RUST_TARGET = aarch64-unknown-linux-gnu
 endif
 ifeq ($(findstring $(RUST_ARCH), $(PPC_ARCHS)),$(RUST_ARCH))
-RUST_BUILD_TOOLCHAIN = 0
+RUST_BUILD_TOOLCHAIN = 1
 RUST_TARGET = powerpc-unknown-linux-gnuspe
 TC_RUSTUP_TOOLCHAIN = stage$(RUSTUP_DEFAULT_TOOLCHAIN_STAGE)-$(RUST_TARGET)
 endif
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -12,6 +12,14 @@ ifeq ($(RUST_TOOLCHAIN),)
 RUST_TOOLCHAIN = stable
 endif
 
+ifeq ($(RUST_TARGET_TOOLCHAIN),)
+ifneq ($(shell rustup target list 2>&1 | grep $(RUST_TARGET) 2>/dev/null),)
+RUST_TARGET_TOOLCHAIN = $(RUST_TOOLCHAIN)
+else
+RUST_TARGET_TOOLCHAIN = $(TC_ARCH)
+endif
+endif
+
 RUST_TARGET =
 # map archs to rust targets
 ifeq ($(findstring $(ARCH), $(x64_ARCHS)),$(ARCH))
@@ -34,7 +42,7 @@ ifeq ($(findstring $(ARCH), $(ARMv8_ARCHS)),$(ARCH))
 RUST_TARGET = aarch64-unknown-linux-gnu
 endif
 ifeq ($(findstring $(ARCH), $(PPC_ARCHS)),$(ARCH))
-RUST_TARGET = powerpc-unknown-linux-gnu
+RUST_TARGET = powerpc-unknown-linux-gnuspe
 endif
 ifeq ($(RUST_TARGET),)
 $(error Arch $(ARCH) not supported)
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -14,6 +14,9 @@ endif
 
 ifeq ($(TC_RUSTUP_TOOLCHAIN),)
 TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
+else
+USE_NATIVE_CMAKE = 1
+include ../../mk/spksrc.native-cmake-env.mk
 endif
 
 RUST_TARGET =
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -8,16 +8,12 @@ export RUSTUP_HOME=$(BASE_DISTRIB_DIR)/rustup
 export PATH:=$(BASE_DISTRIB_DIR)/cargo/bin:$(PATH)
 endif
 
-ifeq ($(RUST_TOOLCHAIN),)
-RUST_TOOLCHAIN = stable
+ifeq ($(RUSTUP_DEFAULT_TOOLCHAIN),)
+RUSTUP_DEFAULT_TOOLCHAIN = stable
 endif
 
-ifeq ($(RUST_TARGET_TOOLCHAIN),)
-ifneq ($(shell rustup target list 2>&1 | grep $(RUST_TARGET) 2>/dev/null),)
-RUST_TARGET_TOOLCHAIN = $(RUST_TOOLCHAIN)
-else
-RUST_TARGET_TOOLCHAIN = $(TC_ARCH)
-endif
+ifeq ($(TC_RUSTUP_TOOLCHAIN),)
+TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
 endif
 
 RUST_TARGET =
@@ -43,6 +39,7 @@ RUST_TARGET = aarch64-unknown-linux-gnu
 endif
 ifeq ($(findstring $(ARCH), $(PPC_ARCHS)),$(ARCH))
 RUST_TARGET = powerpc-unknown-linux-gnuspe
+TC_RUSTUP_TOOLCHAIN = $(TC_ARCH)
 endif
 ifeq ($(RUST_TARGET),)
 $(error Arch $(ARCH) not supported)
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -1,5 +1,8 @@
 # Configuration for rust compiler
-# 
+#
+
+# For newer cmake when building tier-3 toolchains
+include ../../mk/spksrc.native-cmake-env.mk
 
 # Add cargo for rust compiler to default PATH
 ifneq ($(BASE_DISTRIB_DIR),)
@@ -16,7 +19,6 @@ ifeq ($(TC_RUSTUP_TOOLCHAIN),)
 TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
 else
 USE_NATIVE_CMAKE = 1
-include ../../mk/spksrc.native-cmake-env.mk
 endif
 
 RUST_TARGET =
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -16,7 +16,7 @@ endif
 
 # When building toolchain Tier-3 arch support
 #   https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html#stage-2-the-truly-current-compiler
-RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 1
+RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 2
 
 ifeq ($(TC_RUSTUP_TOOLCHAIN),)
 TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -1,14 +1,13 @@
 # Configuration for rust compiler
 #
 
-# For newer cmake when building tier-3 toolchains
-include ../../mk/spksrc.native-cmake-env.mk
-
 # Add cargo for rust compiler to default PATH
 ifneq ($(BASE_DISTRIB_DIR),)
+# Enforce newer cmake when building tier-3 toolchains
+CMAKE_PATH = $(abspath $(WORK_DIR)/../../../native/cmake/work-native/install/usr/local/bin)
 export CARGO_HOME=$(BASE_DISTRIB_DIR)/cargo
 export RUSTUP_HOME=$(BASE_DISTRIB_DIR)/rustup
-export PATH:=$(BASE_DISTRIB_DIR)/cargo/bin:$(PATH)
+export PATH:=$(abspath $(BASE_DISTRIB_DIR)/cargo/bin):$(CMAKE_PATH):$(PATH)
 endif
 
 ifeq ($(RUSTUP_DEFAULT_TOOLCHAIN),)
@@ -17,8 +16,6 @@ endif
 
 ifeq ($(TC_RUSTUP_TOOLCHAIN),)
 TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
-else
-USE_NATIVE_CMAKE = 1
 endif
 
 RUST_TARGET =
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -5,12 +5,10 @@
 ifneq ($(BASE_DISTRIB_DIR),)
 # Enforce newer cmake when building tier-3 toolchains
 CMAKE_PATH = $(abspath $(WORK_DIR)/../../../native/cmake/work-native/install/usr/local/bin)
+LLVM_PATH = $(abspath $(WORK_DIR)/../../../native/llvm-9.0/work-native/install/usr/local/bin)
 export CARGO_HOME=$(BASE_DISTRIB_DIR)/cargo
 export RUSTUP_HOME=$(BASE_DISTRIB_DIR)/rustup
-export PATH:=$(abspath $(BASE_DISTRIB_DIR)/cargo/bin):$(CMAKE_PATH):$(PATH)
-export CFLAGS_$(subst -,_,$(RUST_TARGET))="$(TC_EXTRA_CFLAGS)"
-export CXXFLAGS_$(subst -,_,$(RUST_TARGET))="$(TC_EXTRA_CFLAGS)"
-export LDFLAGS_$(subst -,_,$(RUST_TARGET))="--sysroot=$(WORK_DIR)/$(TC_TARGET)/$(TC_SYSROOT)"
+export PATH:=$(abspath $(BASE_DISTRIB_DIR)/cargo/bin):$(CMAKE_PATH):$(LLVM_PATH):$(PATH)
 endif
 
 ifeq ($(RUSTUP_DEFAULT_TOOLCHAIN),)
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -8,6 +8,9 @@ CMAKE_PATH = $(abspath $(WORK_DIR)/../../../native/cmake/work-native/install/usr
 export CARGO_HOME=$(BASE_DISTRIB_DIR)/cargo
 export RUSTUP_HOME=$(BASE_DISTRIB_DIR)/rustup
 export PATH:=$(abspath $(BASE_DISTRIB_DIR)/cargo/bin):$(CMAKE_PATH):$(PATH)
+export CFLAGS_$(subst -,_,$(RUST_TARGET))="$(TC_EXTRA_CFLAGS)"
+export CXXFLAGS_$(subst -,_,$(RUST_TARGET))="$(TC_EXTRA_CFLAGS)"
+export LDFLAGS_$(subst -,_,$(RUST_TARGET))="--sysroot=$(WORK_DIR)/$(TC_TARGET)/$(TC_SYSROOT)"
 endif
 
 ifeq ($(RUSTUP_DEFAULT_TOOLCHAIN),)
@@ -15,8 +18,9 @@ RUSTUP_DEFAULT_TOOLCHAIN = stable
 endif
 
 # When building toolchain Tier-3 arch support
+#   While stage-2 is the truly current compiler, stage-1 suffice our needs
 #   https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html#stage-2-the-truly-current-compiler
-RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 2
+RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 1
 
 ifeq ($(TC_RUSTUP_TOOLCHAIN),)
 TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
@@ -45,7 +49,7 @@ RUST_TARGET = aarch64-unknown-linux-gnu
 endif
 ifeq ($(findstring $(ARCH), $(PPC_ARCHS)),$(ARCH))
 RUST_TARGET = powerpc-unknown-linux-gnuspe
-TC_RUSTUP_TOOLCHAIN = $(RUST_TARGET)
+TC_RUSTUP_TOOLCHAIN = stage$(RUSTUP_DEFAULT_TOOLCHAIN_STAGE)-$(RUST_TARGET)
 endif
 ifeq ($(RUST_TARGET),)
 $(error Arch $(ARCH) not supported)
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -17,7 +17,7 @@ endif
 # When building toolchain Tier-3 arch support
 #   While stage-2 is the truly current compiler, stage-1 suffice our needs
 #   https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html#stage-2-the-truly-current-compiler
-RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 1
+RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 2
 
 ifeq ($(TC_RUSTUP_TOOLCHAIN),)
 TC_RUSTUP_TOOLCHAIN = $(RUSTUP_DEFAULT_TOOLCHAIN)
--- a/mk/spksrc.cross-rust-env.mk
+++ b/mk/spksrc.cross-rust-env.mk
@@ -19,40 +19,36 @@ endif
 #   https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html#stage-2-the-truly-current-compiler
 RUSTUP_DEFAULT_TOOLCHAIN_STAGE = 2
 
-# When calling directly from
-# toolchain/syno-<arch>-<vertion>
-# ARCH variable is still unset
-ifeq ($(ARCH),)
-ARCH=$(TC_ARCH)
-endif
+# When calling directly from toolchain/syno-<arch>-<version>
+# ARCH variable is still unset thus using $(or $(ARCH),$(TC_ARCH))
 
 # map archs to rust targets
-ifeq ($(findstring $(ARCH), $(ARMv5_ARCHS)),$(ARCH))
+ifeq ($(findstring $(or $(ARCH),$(TC_ARCH)), $(ARMv5_ARCHS)),$(or $(ARCH),$(TC_ARCH)))
 RUST_TARGET = armv5te-unknown-linux-gnueabi
 endif
-ifeq ($(findstring $(ARCH), $(ARMv7_ARCHS)),$(ARCH))
+ifeq ($(findstring $(or $(ARCH),$(TC_ARCH)), $(ARMv7_ARCHS)),$(or $(ARCH),$(TC_ARCH)))
 RUST_TARGET = armv7-unknown-linux-gnueabihf
 endif
-ifeq ($(findstring $(ARCH), $(ARMv7L_ARCHS)),$(ARCH))
+ifeq ($(findstring $(or $(ARCH),$(TC_ARCH)), $(ARMv7L_ARCHS)),$(or $(ARCH),$(TC_ARCH)))
 RUST_TARGET = armv7-unknown-linux-gnueabi
 endif
-ifeq ($(findstring $(ARCH), $(ARMv8_ARCHS)),$(ARCH))
+ifeq ($(findstring $(or $(ARCH),$(TC_ARCH)), $(ARMv8_ARCHS)),$(or $(ARCH),$(TC_ARCH)))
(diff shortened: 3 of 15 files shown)


th0ma7:
@hgy59 Heads-up, this should create packages with full debug_symbols enabled and thus help with gdb to analyze core dumps with a lot more details.  Note that I enabled it on both `qoriq` and `x64`, later to help us out play with it (for instance on my virtual DSM on my syno DS918+)

th0ma7:
@hgy59 I did find a patch to fix building of gdb-7.12.1.  Although what is really really weird is that I can build it from `cross/gdb-7.12` but cannot under `spk` or `diyspk` ?!?!?  Further, after investigating the exact same build issue is also occurring under gdb-13.2 (although from both cross and spk/diyspk).

I was finally able to figure out why (now needs to know how to fix it).  When it goes into the `configure` script under `sim/ppc` at a certain point it tests the build c compiler:
```
checking for i686-pc-linux-gcc... gcc
checking whether the C compiler works... no
configure: error: in `/home/spksrc/qoriq-debug/spksrc/cross/gdb-latest/work-qoriq-6.2.4/gdb-13.2/sim/ppc/build.4119280':
configure: error: C compiler cannot create executables
See `config.log' for more details
mv: cannot stat 'config.h': No such file or directory
```
The end-result is that build-config.h is then missing:
```
fatal error: build-config.h: No such file or directory
   23 | #include "build-config.h"
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
```
As at the end of `sim/ppc/configure` it does a `mv config.h ../build-config.h`.

What I ended-up doing is commenting out the tmpdir removal from that configure script:
```
  CFLAGS="${saved_CFLAGS}"
  mv config.h ../build-config.h
  cd ..
  ###rm -rf $tempdir
```
Which allowed me to keep track of the real error hapening in background:
```
configure:2482: checking whether the C compiler works
configure:2504: gcc -g -O -I/home/spksrc/qoriq-debug/spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/powerpc-e500v2-linux-gnuspe/sysroot/usr/include -mcpu=8548 -mhard-float -mfloat-gprs=double -I/home/spksrc/qoriq-debug/spksrc/cross/gdb-latest/work-qoriq-6.2.4/install/usr/local/gdb/include   conftest.c  >&5
gcc: warning: '-mcpu=' is deprecated; use '-mtune=' or '-march=' instead
gcc: error: unrecognized command-line option '-mfloat-gprs=double'
configure:2508: $? = 1
configure:2546: result: no
```
Now why you ask it does not work?  Simple, while testing the "build" compiler it includes the "target" flags which includes `-mcpu=8548 -mhard-float -mfloat-gprs=double`.  In turns theses flags should never get used on with "build" compiler making it to fail.

Now the question is, how to fix that?

hgy59:
> Although what is really really weird is that I can build it from `cross/gdb-7.12` but cannot under `spk` or `diyspk` ?!?!? Further, after investigating the exact same build issue is also occurring under gdb-13.2 (although from both cross and spk/diyspk).

exactly my findings....

where did you get the patch from? (I created my own).

th0ma7:
> where did you get the patch from? (I created my own).

Backported from https://738272.bugs.gentoo.org/attachment.cgi?id=655808

hgy59:
>Although what is really really weird is that I can build it from `cross/gdb-7.12` but cannot under `spk` or `diyspk`.

This must have something to do with the MAKELEVEL (or include of makefiles), as it fails to build in `cross/gdb` folder too.

the first difference in the configure output is
gdb-7.12 folder:
```
checking for powerpc-e500v2-linux-gnuspe-gcc... powerpc-e500v2-linux-gnuspe-gcc
```

gdb/diyspk/spk folder:
```
checking for powerpc-e500v2-linux-gnuspe-gcc... /spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-gcc
```

The working version finds `powerpc-e500v2-linux-gnuspe-gcc` and the other use the full path. This difference is for the succeeding logs like calling gcc the same. An for other toolchain tools too (ar, as, nm, ranlib, strip, objcopy, objdump, readelf).

Only the evaluation of ld tool differs:
gdb-7.12 folder:
```
checking for ld... (cached) /spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-ld
checking for powerpc-e500v2-linux-gnuspe-ld... (cached) /spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-ld
```

gdb/diyspk/spk folder:
```
checking for ld... (cached) /spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/bin/../lib/gcc/powerpc-e500v2-linux-gnuspe/4.9.3/../../../../powerpc-e500v2-linux-gnuspe/bin/ld
checking for powerpc-e500v2-linux-gnuspe-ld... (cached) /spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/bin/../lib/gcc/powerpc-e500v2-linux-gnuspe/4.9.3/../../../../powerpc-e500v2-linux-gnuspe/bin/ld
```


Does this sound like toolchain specific handling (`PATH=$(TC_PATH):$$PATH`) ?

th0ma7:
@hgy59 I was exactly at that same point until I read this https://scm.linefinity.com/common/openwrt/commit/079d57b0f290a79c9dbc013b6e9c83cebf8a2f99 which at first looked unrelated but it does allow it to build, fully, on both gdb-7.12.1 and gdb-13.2.

I've now created a separete PR for the `libpng` fix https://github.com/SynoCommunity/spksrc/pull/5880 so we can only have the package needed in this one once rebased against master (waiting for the builds to complete, if ever).

What remains to be confirmed is, is it ok to disable sim on ppc?  `sim` stands for simulator which in turns makes sense that it fails to build a simulator on it's own arch?!?  At first glance I belive we can live without this functionality.

hgy59:
> @hgy59 I was exactly at that same point until I read this https://scm.linefinity.com/common/openwrt/commit/079d57b0f290a79c9dbc013b6e9c83cebf8a2f99 which at first looked unrelated but it does allow it to build, fully, on both gdb-7.12.1 and gdb-13.2.


@th0ma7 I just created a patch to build sim/ppc, but don't know whether it is broken anyway.
[002-fix-configure-of-sim-ppc.patch](https://github.com/SynoCommunity/spksrc/files/12596536/002-fix-configure-of-sim-ppc.patch)

th0ma7:
> @th0ma7 I just created a patch to build sim/ppc, but don't know whether it is broken anyway. [002-fix-configure-of-sim-ppc.patch](https://github.com/SynoCommunity/spksrc/files/12596536/002-fix-configure-of-sim-ppc.patch)

If that works out then excellent!  As long as we have a working solution.  Still, more I read and less I believe the ppc simulator is actually needed anyway.

@hgy59 I wonder, shouldn't this be `CPPFLAGS="${CPPFLAGS_FOR_BUILD}"` ?

hgy59:
> @hgy59 I wonder, shouldn't this be `CPPFLAGS="${CPPFLAGS_FOR_BUILD}"` ?


No, `CPPFLAGS_FOR_BUILD` is not a variable of the configure file, only `CFLAGS_FOR_BUILD` and `LDFLAGS_FOR_BUILD` are.
It wouldn't change anything, but would imply that such a variable is available....

hgy59:
@th0ma7 sim/ppc is broken in cross/gdb-latest (i.e. 13.2)  _(but builds fine in cross/gdb-7.12)_
```
make[7]: *** [Makefile:137: icache.o] Error 1
make[7]: Leaving directory '/spksrc/cross/gdb-latest/work-qoriq-6.2.4/gdb-13.2/sim/ppc'
make[6]: *** [Makefile:2556: all-recursive] Error 1
make[6]: Leaving directory '/spksrc/cross/gdb-latest/work-qoriq-6.2.4/gdb-13.2/sim'
make[5]: *** [Makefile:1633: all] Error 2
make[5]: Leaving directory '/spksrc/cross/gdb-latest/work-qoriq-6.2.4/gdb-13.2/sim'
make[4]: *** [Makefile:9950: all-sim] Error 2
make[4]: *** Waiting for unfinished jobs....
```

th0ma7:
> > What remains to be confirmed is, is it ok to disable sim on ppc? `sim` stands for simulator which in turns makes sense that it fails to build a simulator on it's own arch?!? At first glance I belive we can live without this functionality.
> 
> I am totally new on this, but I thought that with the simulator we might try to find the qoriq crash on x64 or other archs.

Indeed, and note that I am relatively new as well to this.

But using this train of thought (and if I get this right), you would need to build gdb for your x86_64 linux that includes the ppc simulator so you can reproduce that locally.  Where you would enable `--enable-sim-powerpc --enable-sim-bitsize=32 --enable-sim-endian=little`.

My theory is that in the current case, having a ppc simulator on a ppc build is most probably useless.  Still, I've checked and when it builds from `cross/gdb-7.12` it does build sim also ... sigh.

th0ma7:
@hgy59 now rebased against master.  resulting github-action results should now reflect our actual needs.

th0ma7:
@hgy59 interestingly, enabling `debug_info` symbols make qoriq fail to build for `python311` which I believe it demonstrates where the bug may lie (don't bother with the `_tkinter` message, but rather on the assembler):
```
.../spksrc/spk/python311/work-qoriq-6.2.4/Python-3.11.5/Modules/_ctypes/stgdict.o -L. -L/home/spksrc/qoriq-debug/spksrc/spk/python311/work-qoriq-6.2.4/install/var/packages/python311/target/lib -L/home/spksrc/qoriq-debug/spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/powerpc-e500v2-linux-gnuspe/sysroot/usr/lib -L/home/spksrc/qoriq-debug/spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/powerpc-e500v2-linux-gnuspe/sysroot/lib -L/home/spksrc/qoriq-debug/spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/powerpc-e500v2-linux-gnuspe/lib -L/home/spksrc/qoriq-debug/spksrc/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/lib/gcc -lffi -ldl -o build/lib.linux-powerpc-3.11/_ctypes.cpython-311-powerpc-linux-gnuspe.so

_tkinter

running build_scripts
creating build/scripts-3.11
copying and adjusting /home/spksrc/qoriq-debug/spksrc/spk/python311/work-qoriq-6.2.4/Python-3.11.5/Tools/scripts/pydoc3 -> build/scripts-3.11
copying and adjusting /home/spksrc/qoriq-debug/spksrc/spk/python311/work-qoriq-6.2.4/Python-3.11.5/Tools/scripts/idle3 -> build/scripts-3.11
copying and adjusting /home/spksrc/qoriq-debug/spksrc/spk/python311/work-qoriq-6.2.4/Python-3.11.5/Tools/scripts/2to3 -> build/scripts-3.11
changing mode of build/scripts-3.11/pydoc3 from 644 to 755
changing mode of build/scripts-3.11/idle3 from 644 to 755
changing mode of build/scripts-3.11/2to3 from 644 to 755
renaming build/scripts-3.11/pydoc3 to build/scripts-3.11/pydoc3.11
renaming build/scripts-3.11/idle3 to build/scripts-3.11/idle3.11
renaming build/scripts-3.11/2to3 to build/scripts-3.11/2to3-3.11
/tmp/ccUd4luC.s: Assembler messages:
/tmp/ccUd4luC.s:1391613: Error: operand out of range (0x0000000000008000 is not between 0xffffffffffff8000 and 0x0000000000007fff)
/tmp/ccUd4luC.s:1391619: Error: operand out of range (0x0000000000008004 is not between 0xffffffffffff8000 and 0x0000000000007fff)
/tmp/ccUd4luC.s:1391626: Error: operand out of range (0x0000000000008008 is not between 0xffffffffffff8000 and 0x0000000000007fff)
/tmp/ccUd4luC.s:1391972: Error: operand out of range (0x000000000000800c is not between 0xffffffffffff8000 and 0x0000000000007fff)
/tmp/ccUd4luC.s:1391989: Error: operand out of range (0x0000000000008010 is not between 0xffffffffffff8000 and 0x0000000000007fff)
/tmp/ccUd4luC.s:1391994: Error: operand out of range (0x0000000000008010 is not between 0xffffffffffff8000 and 0x0000000000007fff)
```

I still presume this is openssl related... perhaps it's missing a few options specifically for ppc?

hgy59:
> /tmp/ccUd4luC.s: Assembler messages:

the final error is:
```
2023-09-14T01:49:01.1161042Z /tmp/ccXENOiA.s:3438448: Error: operand out of range (0x000000000001fc54 is not between 0xffffffffffff8000 and 0x0000000000007fff)
2023-09-14T01:49:01.1161642Z lto-wrapper: /github/workspace/toolchain/syno-qoriq-6.2.4/work/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-gcc returned 1 exit status
```

So this can probably be fixed by disabling optimization (at least the lto optimizer).

hgy59:
> optimization sometimes breaks the code (so far we have such issues for aarch64 only).

just wondering, didn't you have to disable optimization for the debug symbols?

mreid-tt:
Not that I'm an expert in this but this explanation from Bard may assist with debugging:

> The error message "Error: operand out of range (0x000000000001fc54 is not between 0xffffffffffff8000 and 0x0000000000007fff)" means that the program is trying to use an operand (a value that is being operated on) that is outside of the valid range for that operand. In this case, the operand 0x000000000001fc54 is outside of the valid range for the operation that is being performed.
> 
> The specific range that the operand must be within depends on the operation that is being performed. However, in general, the operand must be a signed integer between -32768 and 32767. The range 0xffffffffffff8000 and 0x0000000000007fff is significant because it represents the range of signed integers that can be represented in a 32-bit signed integer variable.
> 
> The range 0xffffffffffff8000 and 0x0000000000007fff is also significant because it is the range of values that can be represented in memory addresses. This means that any memory address in a 32-bit computer can be represented by a value in this range. The (qoriq) Freescale P1022 CPU is a 32-bit computer processor.

hgy59:
> > The range 0xffffffffffff8000 and 0x0000000000007fff is also significant because it is the range of values that can be represented in memory addresses. This means that any memory address in a 32-bit computer can be represented by a value in this range. The (qoriq) Freescale P1022 CPU is a 32-bit computer processor.

IMHO it is eather a 32-bit or endianess problem (or both).
we should avoid the use of an assembler file.

1. build python311 without optimization (`PYTHON_OPTIMIZE = 0`)
2. (if 1. works) build python311 with optimization, but avoid `CONFIGURE_ARGS += --with-lto` for PPC_ARCHS

th0ma7:
> > > The range 0xffffffffffff8000 and 0x0000000000007fff is also significant because it is the range of values that can be represented in memory addresses. This means that any memory address in a 32-bit computer can be represented by a value in this range. The (qoriq) Freescale P1022 CPU is a 32-bit computer processor.
> 
> IMHO it is eather a 32-bit or endianess problem (or both). we should avoid the use of an assembler file.
> 
> 1. build python311 without optimization (`PYTHON_OPTIMIZE = 0`)
> 2. (if 1. works) build python311 with optimization, but avoid `CONFIGURE_ARGS += --with-lto` for PPC_ARCHS

I tried adding `ENV += KERNEL_BITS=32` to openssl3, no luck.  Although removing lto from qoriq build did worked out.  Will push a commit for it.

> just wondering, didn't you have to disable optimization for the debug symbols?

nope I did not... question is, lto breaks qoriq when optimized with debug_info symbols... Could it be broken for PPC with py311.x release?

th0ma7:
@hgy59 I've rebased against master to get rid of the conflict but forgot it would reset the PR state... In any case, for reference, the last sucessfull github-action run is available here https://github.com/SynoCommunity/spksrc/actions/runs/6188300522?pr=5879

hgy59:
@th0ma7 can you enable debug info for all archs?
I would like to run homeassistant with gdb on aarch64. (and maybe on armv7 too).

th0ma7:
> @th0ma7 can you enable debug info for all archs? 

done.

th0ma7:
@hgy59 I believe I have found where the issue really is... into additional missing libraries from the toolchain.

I would need a second pair of eye on commit https://github.com/SynoCommunity/spksrc/pull/5879/commits/530016637c30592c8369462d3bb19c1976d8b77b.  My testing was sucessfull locally on my linux, waiting to see how it behaves online.

Basically what it does:
1. Code was reviewed to accomodate multiple libraries from the toolchain
2. Code also needs to look over shared libraries packaged into wheels... which complicated things significantly (code still looks fine but took quite longer to figure out and debug)
3. interestingly, some libraries (`librt.so`) aren't into `$(TC_LIBRARY_PATH)` but rather into `$(TC_LIBRARY_PATH)/../usr/lib`.  Also, their real name has nothing to do with the link name (`librt-2.20-2014.11.so`) which further complicated things.  Alghough a `find` in conjunction with shell call to `realpath` made that relatively simple in the end.

If you have a few spare cycles to review would be much appreciated.  Much thnx in advance.

hgy59:
> @hgy59 I believe I have found where the issue really is... into additional missing libraries from the toolchain.
> 
> I would need a second pair of eye on commit [5300166](https://github.com/SynoCommunity/spksrc/commit/530016637c30592c8369462d3bb19c1976d8b77b). My testing was sucessfull locally on my linux, waiting to see how it behaves online.
> 
> Basically what it does:
> 
>     1. Code was reviewed to accomodate multiple libraries from the toolchain
> 
>     2. Code also needs to look over shared libraries packaged into wheels... which complicated things significantly (code still looks fine but took quite longer to figure out and debug)
> 
>     3. interestingly, some libraries (`librt.so`) aren't into `$(TC_LIBRARY_PATH)` but rather into `$(TC_LIBRARY_PATH)/../usr/lib`.  Also, their real name has nothing to do with the link name (`librt-2.20-2014.11.so`) which further complicated things.  Alghough a `find` in conjunction with shell call to `realpath` made that relatively simple in the end.
> 
> 
> If you have a few spare cycles to review would be much appreciated. Much thnx in advance.


IMHO this is not the case. The Toolchains contain the libraries that are available on DSM. The `libatomic` is one exception, this is in the toolchain but not in DSM of qoriq.
As I mentioned in https://github.com/SynoCommunity/spksrc/actions/runs/6215356071, libgcc and librt are found on the qoriq system of @marcofranssen.

Regarding the libraries of wheels I am not sure, but as it works on all other systems except qoriq IHMO this is neither an issue.

th0ma7:
@hgy59 I recall having played with core dump files [previously](https://github.com/arvidn/libtorrent/issues/7066) and building a [howto](https://github.com/arvidn/libtorrent/issues/7066#issuecomment-1312262653).  Potentially worth fine tuning and adding to our wiki?

Using gdb locally on your system with this HOWTO would normally allow to trace what's going-on, further if the core was dumped using `debug_info` symbols enabled build (such as packages built from this PR).

I suggest focussing on getting a core dump rather than asking them to run bunch of commands (depending of their level of knowledge in linux).  Core dumps are normally located under `/volume1/` and using a name such as `@<appname>.synology_apollolake_918+.42962.core.gz` (name may vary a bit under DSM6).

th0ma7:
@hgy59 sunday morning theory: we may need to switch to `powerpc-unknown-linux-gnuspe` target instead of `powerpc-unknown-linux-gnu`.  Or better, to properly support qoriq create a `powerpc-e500v2-linux-gnuspe` target.

Issue is, that `powerpc-unknown-linux-gnuspe` is a tier 3 target which means we need to build it first before using it... I've looked into it and it's ain't clear how to do that neither...

th0ma7:
@hgy59 an educated guess found https://github.com/rust-lang/rust/issues/96394 where I using RUSTFLAGS="-Ctarget-cpu=e500" might help...

marcofranssen:
Let me know if there is a new package that I can test and provide feedback.

th0ma7:
@hgy59 things doesn't look too good...  My instinct was right, we need to compile ourselves a compatible rust toolchain for this specific powerpc as it isn't supported otherwise.

After a lot of rust build testing (and I mean a lot) I was able to compile a stage 0, 1 and 2 rust compiler but in order to create the final toolchain it requires to build clang/llvm which in turns needs a "decent" C compiler... which in turns (and similarly to our synocli-devel package) requires >= gcc-5.1 or using clang/llvm...

So, unless I find a relatively "easy" way to build a clang cross-compiler (as I never was able to build one using a newer gcc) our options in supporting the legacy `qoriq` arch is reducing considerably relatively to rust applications (such as `cryptography`).

th0ma7:
@hgy59 a bit of hope... https://github.com/rust-lang/rust/issues/96394#issuecomment-1807222773
So indeed it is mandatory to build a tier 3 gnuspe powerpc arch support.  But it seems "feasible" ... 
Now need to figure out how as all my attempt so far failed.

th0ma7:
@hgy59 potential working solution for fixing `qoriq` arch committed for your review (added this this existing PR - I still have a few commented lines to clean-up once fully tested)...  If all lights are green I'll ask people to try it out to confirm if it does solves things but I'm rather confident...

th0ma7:
@hgy59 (other devs welcomed as well) so I have it all working locally but for an unkown reason I hit an issue with needing a newer CMake with github-action that I can't reproduce.  If you have spare cycles to try it out on your end would be much appreciated.

In the meantime I'll try to publish on my own repository qoriq rebuilt versions of the currently qoriq broken packages so users can give it a shot.

Although relatively to homeassistant I had to slightly bump both `bcrypt` and `cryptography` due to crate build errors fixed in newer versions that build succesfully with the `+nightly` based rust custom built toolchain support for qoriq.  Something potentially needing attention later (or default into marking qoriq as unsupported for this version until next update?) - avenues to be considered once demonstration is done that it can all be built and that gnuspe toolchain really fixes the qoriq issues i guess, my focus being getting all the green lights from github-action, then refining.

hgy59:
unfortunately the packages for qoriq all not available as build artifacts! _(all others are)_
```
Build (qoriq-6.2.4)
No space left on device : '/home/runner/runners/2.311.0/_diag/pages/5d01ad53-1f64-400e-a739-422dc921c365_69bf0e81-3620-54cf-6e4f-fa62cd3e5e0d_1.log'
```

hgy59:
@th0ma7 thanks for your huge work! That is amazing!

Hopefully we get a final confirmation that cryptography wheel now runs on qoriq.

The build of the rust toolchain for `powerpc-unknown-linux-gnuspe` takes pretty long and I hope we can provide a precompiled package somewhere.

log extract of my (not so modern) dev env:
```
===>  Building Tier 3 rust target: powerpc-unknown-linux-gnuspe
Building bootstrap
    Finished dev [unoptimized] target(s) in 1m 39s
Building bootstrap
    Finished dev [unoptimized] target(s) in 0.09s
Building stage0 library artifacts (x86_64-unknown-linux-gnu)
    Finished release [optimized] target(s) in 1m 09s
Building compiler artifacts (stage0 -> stage1, x86_64-unknown-linux-gnu)
    Finished release [optimized] target(s) in 20m 11s
Building stage1 library artifacts (x86_64-unknown-linux-gnu)
    Finished release [optimized] target(s) in 2m 27s
Building compiler artifacts (stage1 -> stage2, x86_64-unknown-linux-gnu)
    Finished release [optimized] target(s) in 20m 49s
Building stage2 library artifacts (x86_64-unknown-linux-gnu -> powerpc-unknown-linux-gnuspe)
    Finished release [optimized] target(s) in 1m 01s
Building stage2 tool rust-analyzer-proc-macro-srv (x86_64-unknown-linux-gnu)
    Finished release [optimized] target(s) in 1m 11s
Building tool rustdoc (stage1 -> stage2, x86_64-unknown-linux-gnu)
    Finished release [optimized] target(s) in 3m 41s
===>  Building Tier 3 rust target: powerpc-unknown-linux-gnuspe - stage2 complete
```

One problem that is boring me is, that the rust installation depends on the build of toolchains.
Similar to the github build action I use a cache for the toolchains.
To get rust updated I have to remove the toolchain cache.

I propose to have isolated targets for rust - or at least dedicated targets and *_done files for rust - and a `make rust-clean` target.

hgy59:
feedback for homeassistant:

The latest build does not run (on DSM 7.2.1 x86_64 VirtualDSM)

var/homessistant.log
```
Wed Nov 29 11:48:09 CET 2023
Starting homeassistant command /volume1/@appstore/homeassistant/env/bin/hass -v --config /volume1/@appdata/homeassistant/config --log-file /volume1/@appdata/homeassistant/homeassistant.log
  File "/volume1/@appstore/homeassistant/env/bin/hass", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/homeassistant/__main__.py", line 181, in main
    args = get_arguments()
           ^^^^^^^^^^^^^^^
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/homeassistant/__main__.py", line 81, in get_arguments
    from . import config as config_util
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/homeassistant/config.py", line 20, in <module>
    from . import auth
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/homeassistant/auth/__init__.py", line 10, in <module>
    import jwt
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/jwt/__init__.py", line 1, in <module>
    from .api_jwk import PyJWK, PyJWKSet
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/jwt/api_jwk.py", line 7, in <module>
    from .algorithms import get_default_algorithms, has_crypto, requires_cryptography
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/jwt/algorithms.py", line 12, in <module>
    from .utils import (
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/jwt/utils.py", line 7, in <module>
    from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurve
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/cryptography/hazmat/primitives/asymmetric/ec.py", line 11, in <module>
    from cryptography.hazmat._oid import ObjectIdentifier
  File "/volume1/@appstore/homeassistant/env/lib/python3.11/site-packages/cryptography/hazmat/_oid.py", line 9, in <module>
    from cryptography.hazmat.bindings._rust import (
ImportError: /lib64/libpthread.so.0: version `GLIBC_PRIVATE' not found (required by /volume1/@appstore/homeassistant/lib/librt.so.1)
```

th0ma7:
> @th0ma7 thanks for your huge work! That is amazing!

Thnx for that... :+1:  being stuborn sometimes pays off (but let's confirm it actually does in this case).

> The build of the rust toolchain for `powerpc-unknown-linux-gnuspe` takes pretty long and I hope we can provide a precompiled package somewhere.

That was my hope as well but I haven't found it anywhere.  An optimization could be of using a `stage1` build instead of a `stage2`, testing required to confirm.  This in turn would reduce by half both build time and storage needs.

> One problem that is boring me is, that the rust installation depends on the build of toolchains.
> Similar to the github build action I use a cache for the toolchains.
> To get rust updated I have to remove the toolchain cache.

Just a thought (and being miles away from there as it doesn't yet work in the first place), we potentially could make that into `native/rust` whereas it only builds it when the corresponding toolchain is unavailable using `work-<target-trippe>` (or directly `cross/rust`)?  This way we could then "re-play" with the rust toolchain version afterwards independently from the actual synology toolchain.  But that would come with it's own caveats as well... 

Or maybie the easiest would be to build it once, package it then use it as needed?  Anyway it won't have to change afterwards besides potentially bumping it's version once in a while?!?

On that, it currently uses defaults meaning it builds from git master being equivalent of `+nightly`.  Haven't yet found how to build against `+stable` which would avoid the need to update bcrypt and cryptography to a minimal version (but that besides the point here)...

Further, making mileage from your recent update I've adapted it to fetch further details when invoking rust builds using `rustc +$(TC_RUSTUP_TOOLCHAIN) -vV`

> I propose to have isolated targets for rust - or at least dedicated targets and *_done files for rust - and a `make rust-clean` target.

Sounds realistic, solution may vary (from comment above).  And btw, there already is a rust cookie file generated by default `.rustc_done` in the toolchain work directory.

> The latest build does not run (on DSM 7.2.1 x86_64 VirtualDSM)
> ```
> ImportError: /lib64/libpthread.so.0: version `GLIBC_PRIVATE' not found (required by /volume1/@appstore/homeassistant/lib/librt.so.1`
> ```

The obvious cause is that it most probably built against the host library instead of the target.  With this in mind I've quicky pushed an obvious "tentative" fix where I pass `--sysroot=` as target `LDFLAGS` using rust target tripple naming scheme.

If that doesn't work then maybie I'll need to review further the flags and align them with all our default when build the gnuspe toolchain ... feel free to give it another shot with this fix once github-action (ever) completes.

th0ma7:
@hgy59 re-reading the error, this has nothing to do with the ppc gnuspe toolchain as being x86_64 related and is clearly a regression (or a bug? stable switched 2-days ago).  The only piece of code I've changed that could have had that effect is the cargo build call whereas I've removed the `+stable` and replaced it with a standard environment variable `RUSTUP_TOOLCHAIN` instead...

I must admit this is odd... although a potential fix could still be passing the sysroot as rust ldflag using the cargo tripplet variable naming scheme.

th0ma7:
@hgy59 ok, I probably found the culprit ... 
> ```
> ImportError: /lib64/libpthread.so.0: version `GLIBC_PRIVATE' not found (required by /volume1/@appstore/homeassistant/lib/librt.so.1`
> ```

While testing things way earlier in the process of finding where the issue was, I had included `librt.so` and `libgcc_s.so` part of the libs to be imported from the toolchain.  I've fixed that with 

Quick testing on DSM-7.2 x64 using a duplicity manually built for DSM-7.1:
```
$ ./duply --version
  duply version 2.4.3
  (https://duply.net)

  Using installed duplicity version 1.2.3, python 3.11.5 (/volume1/@appstore/duplicity/env/bin/python3) 'PYTHONPATH=:/var/packages/python311/target/lib/python311.zip:/var/packages/python311/target/lib/python3.11:/var/packages/python311/target/lib/python3.11/lib-dynload:/volume1/@appstore/duplicity/env/lib/python3.11/site-packages:/var/packages/python311/target/lib/python3.11/site-packages', gpg 2.2.18 (Home: /var/services/homes/th0ma7/.gnupg), awk 'GNU Awk 5.0.0, API: 2.0 (GNU MPFR 4.0.1, GNU MP 6.1.2)', grep 'grep (GNU grep) 3.1', bash '4.4.23(1)-release (x86_64-pc-linux-gnu)'.

$ ./duplicity --version
duplicity 1.2.3
```

hgy59:
@th0ma7 thanks, HA is live again (DSM 7.2.1)

th0ma7:
Boom!  Success at last!

th0ma7:
@hgy59 I've been re-thinking about this, there may be a way to further standardize this whole process.  
1. I believe I could somehow first generate by default a per arch toml config file, always, located under the toolchain work directory.
2. Then custum rust build could be moved to `native` whereas only needed archs ends-up being supported part of the resulting build (i.e. for now only ppc gnuspe)
3. As the process include creating a sort of virtual "link" between the custom resulting binary directory and the default system-wide installation, a link can be created for each needed arch, in this case again only ppc gnuspe

From there the process becomes exactly the same as current from this PR.  Although one thing missing in this picture is the fact that it processes `native` before anything else from `spk.mk` ... That needs some thinking to ensure at the very least toolchains gets processeed first, then native.

But let first see if it works... I've ask people to give it a shot and confirm.  wait & see.

hgy59:
> But let first see if it works... I've ask people to give it a shot and confirm. wait & see.

Congrats and hopefully it works...🤞

marcofranssen:
Happy to test any new release. 🤞

th0ma7:
I haven't had time to pursue this crusade since last attempt.  Although i do have another trick in reserve that i may be able to work on over the holidays.

I'll keep you posted but bare with me, my cycles have been really limited recently.

th0ma7:
Good news, I was able to manually build a `powerpc_unknown_linux_gnuspe` rust toolchain on my native ubuntu using qoriq synology toolchain with locally installed clang+llvm for which `objdump -d` doesn't show any `lwsync` on the resulting `libstd-<ID>.so` library located under `build/host/stage1/lib/rustlib/powerpc-unknown-linux-gnuspe/lib`.

For reproducibility, I've used:
```
CFLAGS_powerpc_unknown_linux_gnuspe="-mcpu=8548 -mhard-float -mfloat-gprs=double" CXXFLAGS_powerpc_unknown_linux_gnuspe="-mcpu=8548 -mhard-float -mfloat-gprs=double" CPPFLAGS_powerpc_unknown_linux_gnuspe="-mcpu=8548 -mhard-float -mfloat-gprs=double" CARGO_TARGET_POWERPC_UNKNOWN_LINUX_GNUSPE_RUSTFLAGS="-Ctarget-cpu=e500" RUST_BACKTRACE=full ./x build --config /home/th0ma7/rust/qoriq.toml
```
Where my `qoriq.toml` configuration file is:
```
profile = "compiler"

[build]
target = ["x86_64-unknown-linux-gnu", "powerpc-unknown-linux-gnuspe"]
build-stage = 1
doc-stage = 2
docs = false
docs-minification = false
compiler-docs = false

[rust]
channel = "stable"
lto = "off"

[llvm]
download-ci-llvm = "if-unchanged"

[install]

[dist]

[target.x86_64-unknown-linux-gnu]

[target.powerpc-unknown-linux-gnuspe]
cc = "/home/th0ma7/rust/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-gcc"
cxx = "/home/th0ma7/rust/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-g++"
ar = "/home/th0ma7/rust/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-ar"
ranlib = "/home/th0ma7/rust/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-ranlib"
linker = "/home/th0ma7/rust/powerpc-e500v2-linux-gnuspe/bin/powerpc-e500v2-linux-gnuspe-gcc"
```

mreid-tt:
I'm sure the community appreciates the work you've put in. Unfortunately, this is definitely out of my expertise so I can't give you more than moral support... :-)

th0ma7:
> I'm sure the community appreciates the work you've put in. Unfortunately, this is definitely outside of my expertise so I can't give you more than moral support... 😀

I finally got it to work!  It does now build locally in my tree! woooooot! :smile: 

Now let's see if it does still build on github-action and proceed with cleanups ...

@marcofranssen if this does build for `qoriq` then there may be something worth trying for you ... let's wait & see https://github.com/SynoCommunity/spksrc/actions/runs/7634289374

marcofranssen:
> > I'm sure the community appreciates the work you've put in. Unfortunately, this is definitely outside of my expertise so I can't give you more than moral support... 😀
> 
> @mreid-tt I must admit that I've always loved theses challenges... but this one got me struggling over the last few monts...
> 
> But good news, I finally got it to work! It does now build locally in my tree! woooooot! (i.e. no more `lwsync` in resulting binaries 😄)
> 
> Now let's see if it does still build on github-action and proceed with cleanups ...
> 
> @marcofranssen if this does build for `qoriq` then there may be something worth trying for you ... let's wait & see [SynoCommunity/spksrc/actions/runs/7634289374](https://github.com/SynoCommunity/spksrc/actions/runs/7634289374)

Awesome work. There seems to be still a minor issue with relation to the config file.
https://github.com/SynoCommunity/spksrc/actions/runs/7634289374/job/20798071040 happy to test the new release and provide my feedback.

th0ma7:
@hgy59 and @mreid-tt I am nearing the point where things will build automagically through github-action.  Once achieved i propose the following:
1. Self host a ppc gnuspe pre-built rust toolchain on our server (or some other location, proposals welcomed)
2. merge all "non-toolchain build code" + including a native/rust-qoriq
3. Keep this PR opened indefinitely for reference in case such functionality is needed later (i.e. needing to build a newer rust for qoriq)

Thoughts?

th0ma7:
> > @marcofranssen if this does build for `qoriq` then there may be something worth trying for you ... let's wait & see [SynoCommunity/spksrc/actions/runs/7634289374](https://github.com/SynoCommunity/spksrc/actions/runs/7634289374)
> 
> Awesome work. There seems to be still a minor issue with relation to the config file.
> https://github.com/SynoCommunity/spksrc/actions/runs/7634289374/job/20798071040 happy to test the new release and provide my feedback.

Indeed there is still an issue i have to figure out for it to build through github-action.  Will let you know once solved but light is in sight and I'll need testers by then.  Thnx.

th0ma7:
@marcofranssen I believe we now have something worth testing 🚀

th0ma7:
@hgy59 using a corresponding toolchain build for `qoriq` would you be able to generate a `native` package available from https://github.com/SynoCommunity/spksrc/releases similar to `llvm-9.0` ?

The rust toolchain is built under `spksrc/toolchain/syno-qoriq-6.2.4/work/rust`.

To enable from a `native` directory freshly extracted, all it needs in theory is :
`rustup toolchain link stage2-powerpc-unknown-linux-gnuspe $(WORK_DIR)/rust/build/host/stage2`

Actual stage value is set throught `$(RUSTUP_DEFAULT_TOOLCHAIN_STAGE)` but i doubt it is available in native.

Whereas `build/host/stage2` contains:
```
host/stage2
├── bin
└── lib
    └── rustlib
        ├── powerpc-unknown-linux-gnuspe
        │   └── lib
        │       └── self-contained
        ├── rustc-src
        │   └── rust -> /home/spksrc/qoriq-debug-update/spksrc/toolchain/syno-qoriq-6.2.4/work/rust
        ├── src
        │   └── rust -> /home/spksrc/qoriq-debug-update/spksrc/toolchain/syno-qoriq-6.2.4/work/rust
        └── x86_64-unknown-linux-gnu
            ├── bin
            ├── codegen-backends
            └── lib
                └── self-contained
```
And note that it is heavy in size but compress relatively well:
```
$ du -sh stage2
446M	stage2
139M	stage2.tar.bz2
```

Then a `rustup show`  to confirm all is properly set ... 

I've never made any releases and noticed you did them all ... (obiwan, you're my only hope :smile:)

hgy59:
> I've never made any releases and noticed you did them all ... (obiwan, you're my only hope 😄)

This is not a great thing. So far I created releases with a dedicated tag. The tag is added to the download path and contains / to define sub dirs.
A release always contains the source code (as zip and tar.gz) of the repo at the given tag. And we can upload any additional archive.
For consistency we should recreate the release after this PR is merged, so the tag will match the source (we can still use the final archive created on this branch to complete the whole integration here).

What do you expect for the final download link to look like?

- https://github.com/SynoCommunity/spksrc/releases/download/toolchains/qoriq6.2.4/rust-1.75.0/rust.txz
or
- https://github.com/SynoCommunity/spksrc/releases/download/toolchains/rust/1.75.0/qoriq6.2.4/rust.txz
or
- we could omit the `toolchains` part of the path

I guess that the archive name must match the target folder `rust` (to be extracted to syno-qoriq-6.2.4/work/) and contain the content of `rust/build/host/stage2`?

We should take the archive format that compresses best (but is not too slow on decompression) out of tar.gz, bz2, txz, ...

I will validate the archive name and directory structure before an upload
If we are not bould to `rust` as archive name, I would prefer a less structured tag (path) and use specific archive names like
- qoriq-6.2.4-rust-1.75.0.tar.gz
- qoriq-6.2.4-rust-1.76.0.tar.gz
- ...

BTW:
We could bundle the synology toolchain and the rust toolchain into one single package to extract to `syno-qoriq-6.2.4/work`. 
This is in the event that Synology would remove the toolchain download when DSM 6 is discontinued in October 2024.
And this would eliminate the need for an additional Rust toolchain.

hgy59:
@th0ma7 while writing the comment above, I was building spk/python311 for qoriq-6.2.4 in a fresh dev environment.

I didn't get the host version of rust.

In `work/rust/build` there are only the following folders
- bootstrap
- cache
- tmp
- x86_64-unknown-linux-gnu

What did I miss?
Is there any Makefile that uses
> To enable from a native directory freshly extracted, all it needs in theory is :
rustup toolchain link stage2-powerpc-unknown-linux-gnuspe $(WORK_DIR)/rust/build/host/stage2

hgy59:
> What did I miss?

Ok, the logs show it exists, ... and it is there (searched for a folder only, but it is a link `host -> /spksrc/toolchain/syno-qoriq-6.2.4/work/rust/build/x86_64-unknown-linux-gnu`).

hgy59:
@th0ma7 I need some help
There are two links in the stage2 folder and I think the rust folder must be included without link
```
t# find stage2 -type l -ls
15611604887      0 lrwxrwxrwx   1 root     root           44 Jan 26 08:00 stage2/lib/rustlib/src/rust -> /spksrc/toolchain/syno-qoriq-6.2.4/work/rust
  128168975      0 lrwxrwxrwx   1 root     root           44 Jan 26 08:00 stage2/lib/rustlib/rustc-src/rust -> /spksrc/toolchain/syno-qoriq-6.2.4/work/rust
```

I think I have to create a script to copy the stage2 folder to expand the linked folders before creating an archive...

hgy59:
@th0ma7 the work/rust folder is very large
```
# du -sh work/rust
4.6G    work/rust
```
we really have to know which parts are finally required.

th0ma7:
Looking at `distrib/rustup/toolchains` a symbolic link got created pointing towards the `stage2` directory and is named refer to the toolchain name from the `rustup toolchain link ` (i.e. `stage2-powerpc-unknown-linux-gnuspe` which is defined in `spksrc.cross-rust-env.mk` such as:
```
ifeq ($(findstring $(ARCH), $(PPC_ARCHS)),$(ARCH))
RUST_TARGET = powerpc-unknown-linux-gnuspe
TC_RUSTUP_TOOLCHAIN = stage$(RUSTUP_DEFAULT_TOOLCHAIN_STAGE)-$(RUST_TARGET)
endif
```

I tested the following:
1. I created a tar.bz2 file from the `stage2` directory (from `host` or `x86_64-unknown-linux-gnu` doesn't matter as one point to the other anyway)
2. Unpacked into a work-native directory such as below (while discarding entirely the `src` and `rust-src` directories :
```
native/rust/
└── work-native
    └── install
        └── usr
            └── local
                ├── bin
                └── lib
                    └── rustlib
                        ├── powerpc-unknown-linux-gnuspe
                        │   └── lib
                        │       └── self-contained
                        └── x86_64-unknown-linux-gnu
                            ├── bin
                            ├── codegen-backends
                            └── lib
                                └── self-contained
```
3. I've re-created the symlink under the rustup directory such as:
```
~/qoriq-debug-update/spksrc/distrib/rustup/toolchains$ ls -la
total 4
drwxr-xr-x 1 spksrc spksrc 132 Jan 26 12:07 .
drwxr-xr-x 1 spksrc spksrc  96 Jan 25 12:10 ..
drwxr-xr-x 1 spksrc spksrc  42 Jan 25 12:10 stable-x86_64-unknown-linux-gnu
lrwxrwxrwx 1 spksrc spksrc  50 Jan 26 12:07 stage2-powerpc-unknown-linux-gnuspe -> ../../../native/rust/work-native/install/usr/local
```
4. Tested building `cross/bat` successfully.

I'd propose something such as the following:
1. include `spksrc.cross-rust-env.mk` in the native `Makefile` to get the `TC_RUSTUP_TOOLCHAIN` value associated with qoriq.
2. extract the package as normal, perhaps using a tar strip to remove the stage2 directory (or whatever may be its parent directory).  If that can help, I recall doing something in `spksrc.toolkit.mk` to call-up tar with `--strip-components=` to remove a pre-defined number or parent directories.
3. Use `rustup toolchain link` such as:
```
rustup toolchain link $(stage2-powerpc-unknown-linux-gnuspe) $(WORK_DIR)/usr/local
```

As for the source, honestly I'd write down that this was generated using `https://github.com/rust-lang/rust.git` as of date ABC.  I could also provide a "howto" for reference?

hgy59:
@th0ma7 I created a test release, available under [rust toolchain for qoriq (beta)](https://github.com/SynoCommunity/spksrc/releases/tag/toolchains/rust/1.77.0-nightly) as
https://github.com/SynoCommunity/spksrc/releases/download/toolchains/rust/1.77.0-nightly/stage2_powerpc-unknown-linux-gnuspe.txz

hgy59:
@th0ma7 what about native/rust/Makefile that directly installs the toolchains to `/spksrc/distrib/rustup/toolchains` ?

hgy59:
I used txz archive to get better compression
```

445M    004/stage2_powerpc-unknown-linux-gnuspe.tar
138M    004/stage2_powerpc-unknown-linux-gnuspe.bz2
98M     004/stage2_powerpc-unknown-linux-gnuspe.txz
```

th0ma7:
> @th0ma7 what about native/rust/Makefile that directly installs the toolchains to `/spksrc/distrib/rustup/toolchains` ?
> 

Why not, if it works out, sure.

th0ma7:
@hgy59 just to get in sync while approaching the finishing line, are you planning to creating a distinct PR relatively to a `native/rust` installation?  as this could be added directly within this PR (although the commits are really polluted due to the vast amount of trials and errors, but will be squashed anyway at merge time)?  Otherwise I may have a few cycles to do that part over the week-end now that the file is online.

Also, before merging this PR there is a need to "disable" the regeneration of the qoriq rust toolchain in favor of the `native/rust`.  I was thinking of adding a variable for that and with a if statement to either install rust qoriq from native (similar to current cmake) or rebuild the toolchain entirely from within `spksrc.tc-rust.mk`.

And lastly, once things becomes relatively settled, I'll need a second pair of eyes to review the code as this been brewing for too long on my end and I may be overlooking at things.  For instance one thing, there is some testing to do as `RUSTUP_TOOLCHAIN` environment variable (mandatory for wheel building as having it's own internal rustc call) supersedes the `rustc +<toolchain>` call - while it works the later should probably be dismissed in favor of the environment variable.

th0ma7:
> BTW: We could bundle the synology toolchain and the rust toolchain into one single package to extract to `syno-qoriq-6.2.4/work`. This is in the event that Synology would remove the toolchain download when DSM 6 is discontinued in October 2024. And this would eliminate the need for an additional Rust toolchain.

I'm wondering (and unrelated to this topic), should we (re)-publish all DSM-6.2.4 toolchains on github, just in case?  Or directly on our synocommunity.com server ... As this is pretty much the last bits standing on sourceforge and note that dsm-6.2.4 is no longer supported got out recently (eol october 2024):
https://www.synology.com/en-global/products/status/eol-dsm62

hgy59:
@th0ma7 I aleady have a local backup of all 6.2.4 toolchains to be prepared to put those to github.
Hosting on synocommunity.com is not an option, since the storage is already on the limit.

Unfortunately I missed some toolchains to backup before January 1 2023 when synology removed all but the DSM 6.2.4 from sf.net.
- northstarplus-1.2 (for R1900ac)
- ppc824x-3.1 (DSM107e)
- 88f618x.3,2 (USB Station 2)

But I have some 4.2 and 5.2 toolchains in my archive.

th0ma7:
Again @hgy59 thnx for the upload of the toolchain, my initial testing seems to be working all-right.

Other than that, I believe I'm now done and would appreciate a code review as I've certainly missed something.
Also, still pending formal user testing to confirm all is really working out Ok.

th0ma7:
There was a tricky part to play with relatively to `ARCH` value depending where you stand when invoking the build.  Now solved and ready for review.

th0ma7:
@hgy59 and all, last round for comments.  I intent to clean-up the package build change to enforce builds and merge to master in no time. Thnx in advance.

toolchain/syno-qoriq-6.2.4/Makefile
```diff
@@ -9,5 +9,16 @@ TC_DIST_SITE_PATH = PowerPC%20QorIQ%20Linux%202.6.32
...
+POST_FIX_TARGET = qoriq_post_fix_target
 
 include ../../mk/spksrc.tc.mk
+
+.PHONY: qoriq_post_fix_target
+qoriq_post_fix_target:
+	@cd $(WORK_DIR)/$(TC_TARGET)/bin ; \
+	for gnutool in $$(ls -1); do \
+	  [ ! -h "powerpc-linux-gnuspe-$${gnutool##*-}" ] && ln -s $${gnutool} "powerpc-linux-gnuspe-$${gnutool##*-}" ; \
+	  [ ! -h "powerpc-unknown-linux-gnuspe-$${gnutool##*-}" ] && ln -s $${gnutool} "powerpc-unknown-linux-gnuspe-$${gnutool##*-}" ; \
```
hgy59, line 23: please use "-L" and avoid the use of "-h" 
regarding this article:
https://stackoverflow.com/questions/1997666/difference-between-test-h-and-test-l
"-h" is legacy version that might be removed some day.

or probably omit the test whether the links not already exist and force the creation (ln -sf )

th0ma7: done

th0ma7:
@SynoCommunity/developers  `qoriq` platform with rust support should now work.

hgy59:
@th0ma7 now on the master branch, I want to recreate the rust stage2 toolchain for qoriq.

Is it correct that I have to replace `RUST_BUILD_TOOLCHAIN = 0` with `RUST_BUILD_TOOLCHAIN = 1` in `mk/spksrc.cross-rust-env.mk` and build e.g.  python311 for qoriq-6.2.4?

th0ma7:
> Is it correct that I have to replace `RUST_BUILD_TOOLCHAIN = 0` with `RUST_BUILD_TOOLCHAIN = 1` in `mk/spksrc.cross-rust-env.mk` and build e.g. python311 for qoriq-6.2.4?

Exactly.  You'll notice that the `RUST_BUILD_TOOLCHAIN` sits under the `qoriq` section as unused on other archs.  And you don't have to build something, you can simply invoke make from `toolchain/syno-qoriq-6.2.4` directory which will install the syno toolchain, rust toolchain + missing qoriq arch which will get either downloaded from native or built.

hgy59:
I have successfully built it. Unfortunatley it is not possible to move the existing tag to create the same release with the related commit on the master branch.
So this is a chicken/egg problem that we will not be able to build a release that is created after the merge. i.e. it is not possilbe to create a release with the source code assets that match the built archive.


The command to create the archive is:
```
/spksrc/toolchain/syno-qoriq-6.2.4# tar -C work/rust/build/x86_64-unknown-linux-gnu/stage2 --exclude=src --exclude=rustc-src -cJf stage2_powerpc-unknown-linux-gnuspe.txz .
```

sorry I'll be offline for some days...

PR #5881, opened 2023-09-14
Framework: add spksrc.main-depends.mk

hgy59:
## Description

This framework cleanup is ahead of the next synocli-develop update.
It simplifies the dummy cross modules that are built for dependency evaluation only.

### Type of change


hgy59:
@th0ma7 can I merge this? 
It is an extract of my work on updating synocli-devel with targets for qoirq and OLD_PPC_ARCHS.

I don't know why the build fails to download nasm (for ffmpeg et. al.) since locally I have no problem to download  https://fossies.org/linux/misc/nasm-2.16.01.tar.xz.

mk/spksrc.configure.mk
```diff
@@ -47,9 +47,8 @@ REAL_CONFIGURE_ARGS += $(CONFIGURE_ARGS)
 
 configure_msg:
 	@$(MSG) "Configuring for $(NAME)"
-	@$(MSG)     - Configure ARGS: $(CONFIGURE_ARGS)
+	@$(MSG)     - Configure ARGS: $(REAL_CONFIGURE_ARGS)
```
th0ma7, line 50: Really?  isn't REAL_CONFIGURE_ARGS related to GNU_CONFIGURE only?

th0ma7 (approved):
Besides the REAL_CONFIGURE_ARGS that I'm not sure of, LGTM.  Overall looks like a neat idea.  If I get this right, as the spksrc.main-depends.mk only has pre-check, cross-env and depend it makes the dependency analysis faster?

Side note, I'd rather use another naming than `main` and `main-depends` but I have no better option to offer... so perhaps for a later time.

mk/spksrc.configure.mk
hgy59: REAL_CONFIGURE_ARGS contains all configure args 

with GNU_CONFIGURE = 1, it contains the --host, --build and --prefix args
and it contains the TC_CONFIGURE_ARGS
plus for DSM 7+ it has --localstatedir defined

without GNU_CONFIGURE it is the same as CONFIGURE_ARGS.


I changed the message to REAL_CONFIGURE_ARGS after I found in cross/gdb-7.12/Makefile there is GNU_CONFIGURE = 1, but finally only CONFIGURE_ARGS are used and the REAL_CONFIGURE_ARGS logic is implemented again.


Is is only confusing with custom configure targets, but this is always the case with a common pre-target message.
Custom configure targets should always add own messages to avoid confusion (when not using REAL_CONDIGURE_ARGS).

hgy59:
> Side note, I'd rather use another naming than `main` and `main-depends` but I have no better option to offer... so perhaps for a later time.

My first approach was spksrc.no-op.mk, then spksrc.evaluate-depends.mk and finally spksrc.main-depends.mk based on your naming (*-main) for the PKG_NAME 😄

th0ma7:
> My first approach was spksrc.no-op.mk, then spksrc.evaluate-depends.mk and finally spksrc.main-depends.mk based on your naming (*-main) for the PKG_NAME 😄

I must be getting old... yet another senior moment :older_man:

PR #5882, opened 2023-09-14
update synocli-devel

hgy59:
## Description

- update strace to v6.5
- add and fix gdb for qoriq arch
- add llvm-9.0 for qoric arch
- add binutils-2.32 for OLD_PPC_ARCHS

## Checklist

### Type of change

- [x] Package update


cross/gdb-7.12/Makefile
```diff
@@ -8,46 +8,40 @@ PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
...
 
 .PHONY: gdb_compile
 gdb_compile:
-	@$(RUN) ; cd build ; PATH=$(TC_PATH):$$PATH make
+	@$(RUN) ; cd build ; PATH=$(abspath $(TC_PATH)):$$PATH make
 
 .PHONY: gdb_install
 gdb_install:
-	@$(RUN) ; cd build ; PATH=$(TC_PATH):$$PATH make install DESTDIR=$(INSTALL_DIR)
+	@$(RUN) ; cd build ; PATH=$(abspath $(TC_PATH)):$$PATH make install DESTDIR=$(INSTALL_DIR)
```
th0ma7, line 47: I suggest using `realpath` as otherwise `abspath` will provide an empty string if not existant.

cross/gdb/Makefile
```diff
@@ -5,8 +5,13 @@ OPTIONAL_DEPENDS += cross/gdb-7.12
 
 include ../../mk/spksrc.main-depends.mk
 
-ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS)),$(ARCH))
+ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS) $(PPC_ARCHS)),$(ARCH))
```
th0ma7, line 8: PPC_ARCHS does build gdb-latest.  Perhaps you meant $(OLD_PPC_ARCHS) ?

spk/synocli-devel/Makefile
```diff
@@ -90,25 +88,27 @@ SPK_COMMANDS += bin/strip
...
 OPTIONAL_DESC += ", gdb 13.2"
 endif
+endif
 SPK_COMMANDS += bin/gcore
 SPK_COMMANDS += bin/gdb
 SPK_COMMANDS += bin/gdb-add-index
 SPK_COMMANDS += bin/gdbserver
-ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS) $(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
+ifeq ($(findstring $(ARCH),$(ARM_ARCHS)),$(ARCH))
 SPK_COMMANDS += bin/run
```
th0ma7, line 105: This would also include aarch64 no?  But I believe it shouldn't.  To be confirmed.

th0ma7 (changes requested):
A few minor adjustments but otherwise looking good!

cross/gdb-latest/Makefile
```diff
@@ -19,10 +19,12 @@ UNSUPPORTED_ARCHS = $(ARMv5_ARCHS) $(PPC_ARCHS)
...
 include ../../mk/spksrc.cross-cc.mk
 
-# Library is not created under armv7
-ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
+# Library is not created for armv5 and armv7
+ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS) $(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
 PLIST_TRANSFORM = sed -e 's?lib:lib/libinproctrace.so??g'
-# binary only created on armv7
-else
+endif
```
th0ma7, line 25: Also missing is :
```
# disable simulator ; it's broken on ppc
# https://scm.linefinity.com/common/openwrt/commit/079d57b0f290a79c9dbc013b6e9c83cebf8a2f99
ifeq ($(findstring $(ARCH),$(PPC_ARCHS)),$(ARCH))
CONFIGURE_ARGS += --disable-sim
endif
```

Further: https://sourceware.org/pipermail/gdb/2023-September/050898.html

th0ma7: Also, this applies to both gdb-7.12.1 and gdb-latest

spk/synocli-devel/Makefile
hgy59: Yes it does. `bin/run` is also built for aarch64. (if not, the build would fail).

cross/gdb-latest/Makefile
hgy59: That was my finding: gdb-latest (13.2) has broken sim for qoriq, but gdb-7.12 has not.
It builds fine with the provided patches.

cross/gdb/Makefile
hgy59: No, see my comment below for building gdb-17.2 for qoriq.

cross/gdb-7.12/Makefile
hgy59: No, it is the other way round.
abspath does not resolve symlinks and does not require the file names to refer to an existing file or directory.

**$(realpath names…)**

For each file name in names return the canonical absolute name. A canonical name does not contain any . or .. components, nor any repeated path separators (/) or symlinks. In case of a failure the empty string is returned. Consult the realpath(3) documentation for a list of possible failure causes.

**$(abspath names…)**

For each file name in names return an absolute name that does not contain any . or .. components, nor any repeated path separators (/). Note that, in contrast to realpath function, abspath does not resolve symlinks and does not require the file names to refer to an existing file or directory. Use the wildcard function to test for existence. 

taken from https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html
(or see https://www.cmcrossroads.com/article/gnu-make-path-handling)

cross/gdb-latest/Makefile
hgy59: see my comment https://github.com/SynoCommunity/spksrc/pull/5879#issuecomment-1717396995

hgy59:
set to WIP:
The spksrc.main-depends.mk is missing the PLIST handling and the generated packages are missing the files of the related packages.

cross/gdb-7.12/Makefile
th0ma7: you are totally right...

cross/gdb-latest/Makefile
th0ma7: From my readings sim isn't needed anyway.  So while it now build for 7.12.1 I still recommend we deflect that and use gdb-latest wth --disable-sim.

Also, more I'm looking at how it's handling the compiler from within `sim/ppc` more I doubt it will ever work correctly due to using "build" compiler instead of "target".  I'd go further and recomment using --disable-sim on both versions for qoriq and use gdb-latest.

Further adding, the test for switching between 7.12 and 3.2 should be if gcc <= 4.9 or similar instead of being armv5 specific.

th0ma7:
> @th0ma7 the only thing that bothers me, is the build of `native/llvm` for all archs. `native/llvm` must only be built for packages depending on cross/llvm-9` but it is built for all packages, even ARMv5 that does not include llvm at all.

Agreed, the nicest would be to either have a workaround to:
1. only build the needed bits (which I haven't found how) or
2. download a version from somewhere like from a debian package
3. or perhaps installing it by default in the debian container (wouldn't that be the easiest?)

> This is an issue with github actions, since every package is built in its own environment (locally it builds native llvm only once and takes up to 3 h on my older build env). Even on the github action environment the build of native/llvm takes more than one hour.
> 
> For ARMv5 archs the cause is the `pre-build-native` target that is called before the UNSUPPORTED_ARCHS check (guessing). 

It is.

> For archs using cross/llvm-latest the cause is the OPTIONAL_DEPENDS of cross/llvm-9 in cross/llvm. Optionally it always depends on native/llvm and that is built with the `pre-build-native` target.

A work-around that would be to add a new "phase" within the build process to do the `pre-build-native` for all and use that resulting build image for every archs.  This would avoid building native portions for every archs, but from a build duration perspective that wouldn't change much.

> Probably you can remeber why you introduced `pre-build-native` target in #4200? If we let `make` resolve the dependencies, this would not happen, but if we use the dependency-list target that resolves all dependencies (including optional deps.) we do not find the real - arch or gcc version specific - dependencies.

That is mandatory for parallel building as otherwise every archs kicks-in at the same time and tries to build the same native dependencies one over the other causing it to fail miserably.  Work-around could be:
1. using flock (unsure how I would do that though but probably feasible) similarly to the download code.  But when running on your local environment, all arch builds except one will be pending for a really long duration for things like llvm (or legacy cmake for that matter that takes a good 10+ minutes if I recall).  Adjusting timer would be a hassle.  On the other hand that would solve things for github-action as being independant for every archs.
2. ensuring native is used by all builds on github-action and being cached - eventually even llvm would be cached in there potentially, thus expediting the build.  But unsure how to do that neither.

> I cannot imagine a solution (exept removing the `pre-build-native` target), so this will probably not be part of this PR.

Such a change would be relatively intrusive and benefit from being standalone.... Unless we install it into the docker image or find binaries to download from.

hgy59:
> That is mandatory for parallel building as otherwise every archs kicks-in at the same time and tries to build the same native dependencies one over the other causing it to fail miserably. Work-around could be:

IMHO this is not the case. Every build in the matrix runs in its own VM and does not share the native build with other arch-TCVERSION build. The only concurrency problem was, when all VMs wanted to download the same sources, but this we fixed with the prepare job and the caching of downloaded files.
please correct me, if I am wrong...

cross/gdb-latest/Makefile
hgy59: > From my readings sim isn't needed anyway. So while it now build for 7.12.1 I still recommend we deflect that and use gdb-latest wth --disable-sim.

So I will disable sim for all archs?

hgy59: > Further adding, the test for switching between 7.12 and 3.2 should be if gcc <= 4.9 or similar instead of being armv5 specific.

_GCC 4.8.1 was the first feature-complete implementation of the 2011 C++ standard..._

So I will check for gcc < 4.8.1 (not <= 4.9)

hgy59:
> EDIT2: But it may well be available in a linux compatible binary form from the llvm download page https://releases.llvm.org/download.html

the only available prebuilt version for x64 (amd64) is for freebsd11
https://github.com/llvm/llvm-project/releases/download/llvmorg-9.0.1/clang+llvm-9.0.1-amd64-unknown-freebsd11.tar.xz

But this archive is missing `clang-tblgen` too.


What about self hosting the `native/llvm-9`, once built locally?

hgy59:
> > EDIT2: But it may well be available in a linux compatible binary form from the llvm download page https://releases.llvm.org/download.html
> 
> the only available prebuilt version for x64 (amd64) is for freebsd11 https://github.com/llvm/llvm-project/releases/download/llvmorg-9.0.1/clang+llvm-9.0.1-amd64-unknown-freebsd11.tar.xz
> 
> But this archive is missing `clang-tblgen` too.
> 
> What about self hosting the `native/llvm-9`, once built locally?

@th0ma7 I started a test to try selfhosted prebuilt native llvm-9.0 (local builds have a slightly different build path (`/spksrc/native/llvm-9.0/` vs `/github/workspace/native/llvm-9.0/`) but library dependencies are relative, so this should not be a problem. 🤞 

The prebuilt package is only 55MB tar.gz, so this will be a fast install of native/llvm-9.0.

cross/gdb-latest/Makefile
th0ma7: > > From my readings sim isn't needed anyway. So while it now build for 7.12.1 I still recommend we deflect that and use gdb-latest wth --disable-sim.
> 
> So I will disable sim for all archs?

No, I'd disable it only for qoriq as it works ok for all others.  Unless we want to go through an optimization of gdb and disable everything that is uneeded.

th0ma7:
> > That is mandatory for parallel building as otherwise every archs kicks-in at the same time and tries to build the same native dependencies one over the other causing it to fail miserably. Work-around could be:
> 
> IMHO this is not the case. Every build in the matrix runs in its own VM and does not share the native build with other arch-TCVERSION build. The only concurrency problem was, when all VMs wanted to download the same sources, but this we fixed with the prepare job and the caching of downloaded files. please correct me, if I am wrong...

You are right in a github-action environment... But when building locally using `make all-supported` it's totally different and the concurrency problem hits you real hard.  For that the download problem was fixed using `flock`.  On github-action as it spawn multiple VMs we pre-download and mount for each VM afterwards to avoid duplicating data (i.e. different use-case).

So again, from a concurrency standpoint when building locally with `PARALLEL_MAKE=max make -jX all-supported` you will face issues with native which are only built once for all archs.  That is why it's being done first before invoking parallel make.

th0ma7:
> > > EDIT2: But it may well be available in a linux compatible binary form from the llvm download page https://releases.llvm.org/download.html
> > 
> > 
> > the only available prebuilt version for x64 (amd64) is for freebsd11 https://github.com/llvm/llvm-project/releases/download/llvmorg-9.0.1/clang+llvm-9.0.1-amd64-unknown-freebsd11.tar.xz
> > But this archive is missing `clang-tblgen` too.
> > What about self hosting the `native/llvm-9`, once built locally?
> 
> @th0ma7 I started a test to try selfhosted prebuilt native llvm-9.0 (local builds have a slightly different build path (`/spksrc/native/llvm-9.0/` vs `/github/workspace/native/llvm-9.0/`) but library dependencies are relative, so this should not be a problem. 🤞
> 
> The prebuilt package is only 55MB tar.gz, so this will be a fast install of native/llvm-9.0.

Could we self-host it on synocommunity.com directly?  Another option could be https://rpmfind.net/linux/rpm2html/search.php?query=clang-tblgen&submit=Search+...&system=&arch=x86_64

hgy59:
> @hgy59 can you merge to master [19b2a78](https://github.com/SynoCommunity/spksrc/commit/19b2a780bd7059128b257517135bea0c58bb90c0) if it does fix builds? I'll rebase #5879 accordingly.

done

hgy59:
> You are right in a github-action environment... But when building locally using `make all-supported` it's totally different and the concurrency problem hits you real hard. For that the download problem was fixed using `flock`. On github-action as it spawn multiple VMs we pre-download and mount for each VM afterwards to avoid duplicating data (i.e. different use-case).

with this in mind, the pre-build-native target should only be used in the context of `make all-supported`.

BTW I never use `make all-supported`, since I always want to build all and not terminate on the first error.

Anyway, with the use of prebuilt native/llvm-9 this is not a priority issue anymore.

cross/gdb/Makefile
th0ma7: This is now irrelevant, marking as resolved.

th0ma7:
> > You are right in a github-action environment... But when building locally using `make all-supported` it's totally different and the concurrency problem hits you real hard. For that the download problem was fixed using `flock`. On github-action as it spawn multiple VMs we pre-download and mount for each VM afterwards to avoid duplicating data (i.e. different use-case).
> 
> with this in mind, the pre-build-native target should only be used in the context of `make all-supported`.

Probably right indeed.

> BTW I never use `make all-supported`, since I always want to build all and not terminate on the first error.

Well it the arch that breaks stops, remaining parallel jobs continues (i.e. `PARALLEL_MAKE=max make -j4 all-supported`).  Note that you absolutely need to use both `PARALLEL_MAKE` and `-jX` flag.

> Anyway, with the use of prebuilt native/llvm-9 this is not a priority issue anymore.

Even better!  Still, I'll check later on if I can enforce building native only from `all-supported`.

th0ma7 (approved):
LGTM

PR #5213, opened 2022-04-19
[WIP] add Urbackup

hgy59:
## Description

This is a followup of #3983 as it was closed by @josef109

Closes  #3819 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

### Remarks
- currently [WIP] as DSM 7 fails to build package with to level var folder
- DSM 7 support needs further investigation, as currently service must be started as root user

PR #5912, opened 2023-10-15
ownCloud: Update to v10.13.2

mreid-tt:
## Description

1. Update ownCloud to version 10.13.2
2. New backup and restore functions
3. Migrate to new shared folder handling
4. Update wizards to simplify maintenance (using mustache)
5. Remove synthetic service daemon (using webservice)
6. Fix for #5819 to remove default uninstall template if shell script present

Fixes # 

## Checklist

### Type of change

- [x] Bug fix
- [x] Package update


spk/owncloud/src/wizard_templates/install_uifile.sh.yml
```diff
@@ -0,0 +1,14 @@
+OWNCLOUD_ADMIN_CONFIGURATION_STEP_TITLE: "ownCloud admin configuration"
```
smaarn, line 1: I believe this file should be named without the .sh part.
Sorry for not documenting that !
I will update the doc today.

smaarn (commented):
I think the only issue here with the wizards are the fact that they are using ".sh.yml" suffixes when it should only be ".yml"
Apologies for not updating the docs

spk/owncloud/src/wizard_templates/install_uifile.sh.yml
mreid-tt: Yes, that did the trick. Much thanks!

mreid-tt:
> I think the only issue here with the wizards are the fact that they are using ".sh.yml" suffixes when it should only be ".yml" Apologies for not updating the docs

Thanks for the explanation, was simple to resolve. Appreciate your PR to add this as it does greatly simplify multilingual wizards. Look forward to the updated docs.

hgy59:
@mreid-tt any feedback on how to migrate to the new shared folder handling is appreciated. I didn't have time to dig into it...

mreid-tt:
> @mreid-tt any feedback on how to migrate to the new shared folder handling is appreciated. I didn't have time to dig into it...

I tried to install onto a volume2 but it didn't complete. The log showed the following:

```
2023/10/15 10:56:14	install owncloud 10.13.2-12 Begin preinst
2023/10/15 10:56:15	SHARE_PATH from share [], variable [wizard_data_share=owncloud]
2023/10/15 10:56:15	Begin initialize_variables
2023/10/15 10:56:15	End initialize_variables
2023/10/15 10:56:15	===> Step preinst. STATUS=INSTALL USER=owncloud GROUP= SHARE_PATH=
2023/10/15 10:56:15	install owncloud 10.13.2-12 End preinst ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/mkdir -p /volume2/@appstore/owncloud -m 755
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/mkdir -p /volume2/@appstore/owncloud -m 755 ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/rm -rf /volume2/@appstore/owncloud
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/rm -rf /volume2/@appstore/owncloud ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/package /volume2/@appstore/owncloud
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/package /volume2/@appstore/owncloud ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/rm -rf /var/packages/owncloud
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/rm -rf /var/packages/owncloud ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/mkdir -p /var/packages/owncloud -m 755
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/mkdir -p /var/packages/owncloud -m 755 ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/touch /var/packages/owncloud/installing
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/touch /var/packages/owncloud/installing ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/INFO /var/packages/owncloud/INFO
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/INFO /var/packages/owncloud/INFO ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/rm -rf /var/packages/owncloud/scripts
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/rm -rf /var/packages/owncloud/scripts ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/scripts /var/packages/owncloud/scripts
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/scripts /var/packages/owncloud/scripts ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/rm -rf /var/packages/owncloud/WIZARD_UIFILES
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/rm -rf /var/packages/owncloud/WIZARD_UIFILES ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/WIZARD_UIFILES /var/packages/owncloud/WIZARD_UIFILES
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/WIZARD_UIFILES /var/packages/owncloud/WIZARD_UIFILES ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/rm -rf /var/packages/owncloud/conf
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/rm -rf /var/packages/owncloud/conf ret=[0]
2023/10/15 10:56:16	install owncloud 10.13.2-12 Begin /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/conf /var/packages/owncloud/conf
2023/10/15 10:56:16	install owncloud 10.13.2-12 End /bin/mv -f /volume2/@tmp/synopkg/install.qr0gr3/conf /var/packages/owncloud/conf ret=[0]
2023/10/15 10:57:08	install owncloud 10.13.2-12 Begin postinst
2023/10/15 10:57:08	SHARE_PATH from share [/volume2/owncloud], variable [wizard_data_share=owncloud]
2023/10/15 10:57:08	Begin initialize_variables
2023/10/15 10:57:08	Evaluate full path for SHARE_NAME [owncloud] and SHARE_PATH [/volume2/owncloud]
2023/10/15 10:57:08	SHARE_PATH is absolute path [/volume2/owncloud].
2023/10/15 10:57:08	End initialize_variables
2023/10/15 10:57:08	===> Step postinst. STATUS=INSTALL USER=owncloud GROUP= SHARE_PATH=/volume2/owncloud
2023/10/15 10:57:08	Begin save_wizard_variables
2023/10/15 10:57:08	End save_wizard_variables
2023/10/15 10:57:08	Begin syno_sync_var_folder
2023/10/15 10:57:08	End syno_sync_var_folder
2023/10/15 10:57:08	Begin service_postinst
2023/10/15 10:57:08	/bin/mkdir: cannot create directory '/volume1/owncloud': Permission denied
2023/10/15 10:57:14	Can't create or write into the data directory /volume1/owncloud/data
2023/10/15 10:57:14	                                                                   
2023/10/15 10:57:14	  There are no commands defined in the "config:system" namespace.  
2023/10/15 10:57:14	                                                                   
2023/10/15 10:57:14	End service_postinst
2023/10/15 10:57:14	install owncloud 10.13.2-12 End postinst ret=[0]
2023/10/15 10:57:20	install owncloud 10.13.2-12 Begin start-stop-status start
2023/10/15 10:57:20	install owncloud 10.13.2-12 End start-stop-status start ret=[0]
```

Any ideas?

mreid-tt:
> @mreid-tt any feedback on how to migrate to the new shared folder handling is appreciated. I didn't have time to dig into it...

The new shared folder handling seems to work well. Tested this package with a `/volume2` installation on both DSM 7 and DSM 6 without any issues.

mreid-tt:
@smaarn, I don't know if your PR catered for excluding the default `uninstall_uifile` if the wizard template generates its own file `uninstall_uifile.sh`. Currently the built package contains both. While it doesn't appear to affect uninstalls it would be great if the default uninstall wizard were not present.

@hgy59 I know you are a master at the framework and may have a quick solution to the above (to amend #5819). Any thoughts come to mind?

hgy59:
> The new shared folder handling seems to work well. Tested this package with a `/volume2` installation on both DSM 7 and DSM 6 without any issues.

Found an issue when migrating #5213 to new shared folders:
On DSM 7 a package update from the old to the new shares works as expected, but on DSM 6 it fails.

In `mk/spksrc.service.installer.functions` in the function `initialize_variables ()` the command `abspath` must be replaced by `realpath` since `abspath` is not a valid command on DSM.
```
        if [ -z "${SHARE_NAME}" ]; then
            SHARE_NAME=$(echo $(realpath ${SHARE_PATH}) | awk -F/ '{print $3}')
        fi
```

But the real problem under DSM 6 is, that the shared folder cannot be created from the resource file on package update when the wizard variable with the share name is not defined.

I fear that only on DSM 7, where the shared folders are (always) created by resource files, a package update without an upgrade wizard works. 
For DSM 6 and migration from manually created share to resource worker by package upgrade, we need one of 
- an upgrade wizard to reinitialize the wizard variable for the share name from etc/installer-variables of the installed package
- patch the resource file to replace the placeholder (wizard-variable) with the share name taken from etc/installer-variables of the installed package
- initialize the wizard-variable from etc/installer-variables of the installed package without an upgarde wizard

I will have to investigate, whether it can be fixed by initialization of the wizard-variable or a patch of the resource file (or a wizard) is required.

mreid-tt:
> @smaarn, I don't know if your PR catered for excluding the default `uninstall_uifile` if the wizard template generates its own file `uninstall_uifile.sh`. Currently the built package contains both. While it doesn't appear to affect uninstalls it would be great if the default uninstall wizard were not present.

I was looking at some of the code again and I see this section:

https://github.com/SynoCommunity/spksrc/blob/ecce16fe7b5f97f553cd57c7822f896cef08bc7f/mk/spksrc.spk.mk#L388-L391

This seems to just take the stock `uninstall_uifile` and copy it into the `DSM_WIZARDS_DIR`.

Then we have this section:

https://github.com/SynoCommunity/spksrc/blob/ecce16fe7b5f97f553cd57c7822f896cef08bc7f/mk/spksrc.spk.mk#L396-L398

This seems to be the new bits that buid the wizards from templates using mustache and outputs them into `WIZARDS_DIR`.

Then we have this bit that seems to be putting it all together:

https://github.com/SynoCommunity/spksrc/blob/ecce16fe7b5f97f553cd57c7822f896cef08bc7f/mk/spksrc.spk.mk#L436-L440

The last line seems to be copying the generated files in `WIZARDS_DIR` (which is the same as `SPKSRC_WIZARDS_DIR`) and placing them into `DSM_WIZARDS_DIR` to join what is already there.

Would it make sense then that to fix this issue that we add a line right after this to look for a file named `uninstall_uifile.sh` and if present, remove the file `uninstall_uifile` that was copied in the first set of code above? Perhaps something like:

```
@if [ -e "$(DSM_WIZARDS_DIR)/uninstall_uifile.sh" ]; then \
    rm "$(DSM_WIZARDS_DIR)/uninstall_uifile"; \
fi
```

mreid-tt:
@hgy59 based on your comment in #5913, I've rebased this PR to incorporate your fix. Should I test by installing the current ownCloud release in DSM 6 and then attempting a package update using the output from this build? I only checked a package update in DSM 7 before.

hgy59:
> @hgy59 based on your comment in #5913, I've rebased this PR to incorporate your fix. Should I test by installing the current ownCloud release in DSM 6 and then attempting a package update using the output from this build? I only checked a package update in DSM 7 before.

As I wrote elsewhere it is not so easy to migrate to the new shared folder handling for DSM 6 (and the same might apply when migrating from DSM 6 to 7).
On PR #5213 the only solution I found so far is, to present an upgrade wizard with the shared folder name as default. This can be hidden when the resource worker is already used for the share.
The related code in `upgrade_uifile.sh` looks like:
```sh
#!/bin/sh

# provide an upgrade wizard if the shared folder is not already migrated to resource worker
RESOURCE_OWN="/var/packages/${SYNOPKG_PKGNAME}/conf/resource.own"
if [ -r ${RESOURCE_OWN} ]; then
  if [ $(cat ${RESOURCE_OWN} | jq '."data-share".shares[0].created') == "true" ]; then
    # shared folder already migrated to resource worker
    exit 0;
  fi
else
  # shared folder not supported by resource worker (DSM 5, SRM)
  exit 0;
fi

# Migration of shared folders requires wizard variable referenced by resource file
INST_ETC="/var/packages/${SYNOPKG_PKGNAME}/etc"
INST_VARIABLES="${INST_ETC}/installer-variables"

# Recreate wizard variable from SHARE_NAME or SHARE_PATH
   while read -r _line; do
      _key=${_line%%=*}
      _value=${_line#*=}
      declare -g "${_key}=${_value}"
   done < ${INST_VARIABLES}
fi

if [ -z "${SHARE_NAME}" ]; then
  if [ -z "${SHARE_PATH}" ]; then
    exit 0;
  fi
  declare -g SHARE_NAME=$(basename ${SHARE_PATH})
fi

cat <<EOF > $SYNOPKG_TEMP_LOGFILE
[
...
```

the final `cat` has to create the json for the wizard with the default value for the shared folder name defined as:
```
             "defaultValue": "${SHARE_NAME}",
```

I didn't find any other solution yet.
Such solution could be
- patch `conf/resource`  in the package so that the share name from installer-variables is used by DSM installer
- avoid the need of an upgrade wizard by definition of the wizard variable for the share name so that it is used by DSM installer

Another solution would be to force the user to uninstall the package first on DSM 6 (or not to migrate to the new shared folder handling)...

mreid-tt:
> As I wrote elsewhere it is not so easy to migrate to the new shared folder handling for DSM 6 (and the same might apply when migrating from DSM 6 to 7).

Ah, I understand now. In this case, I don't believe this will be needed for this package. Since I first modernised this package under https://github.com/SynoCommunity/spksrc/pull/5619, it was migrated to shared folder handling.

Additionally, based on the constraints from the original repo, it was not upgradeable from v8.1.1-7 and required the user to uninstall and re-install per this part of the wizard:

https://github.com/SynoCommunity/spksrc/blob/d254a91bcf474f82ad6a0935eaef22b82fbbab1a/spk/owncloud/src/wizard_templates/upgrade_uifile.sh#L32

As such, I believe we are good to go with this PR. Can you give it a final review?

hgy59:
> Ah, I understand now. In this case, I don't believe this will be needed for this package. Since I first modernised this package under #5619, it was migrated to shared folder handling.

👍 Yes, owncloud already had the resource worker for DSM 6 (with `USE_DATA_SHARE_WORKER = yes`). The upgrade issue is related to migration from manually created shares on DSM 6 to resource worker.

mreid-tt:
@hgy59 nice, are we good to merge?

spk/owncloud/src/service-setup.sh
```diff
@@ -5,6 +5,9 @@ WEB_DIR="/var/services/web_packages"
 if [ $SYNOPKG_DSM_VERSION_MAJOR -lt 7 ];then
     WEB_DIR="/var/services/web"
 fi
+if [ -z "${SYNOPKG_PKGDEST_VOL}" ]; then
+	SYNOPKG_PKGDEST_VOL="/volume1"
+fi
```
hgy59, line 10: variable SYNOPKG_PKGDEST_VOL is always predefined by DSM

spk/owncloud/src/service-setup.sh
```diff
@@ -145,7 +139,7 @@ service_postinst ()
 
     if [ "${SYNOPKG_PKG_STATUS}" = "INSTALL" ]; then
         # Parse data directory
-        DATA_DIR="/volume1/${wizard_data_share}/data"
+        DATA_DIR="${SYNOPKG_PKGDEST_VOL}/${wizard_data_share}/data"
```
hgy59, line 142: with the shared folder redesign you can simply use:
```suggestion
        DATA_DIR="${SHARE_PATH}/data"
```

spk/owncloud/src/service-setup.sh
hgy59: the same applies to the wizard *.sh files

hgy59 (commented):
General comment:

I highly recommend not to evaluate web folders etc. in the wizard *.sh files.

Use as less logic in the wizards as possible.
Synology's intention is still to drop support for the wizards, so it is better to leave such logic in the service_* functions in service-setup.sh

Synology prefers the unattended installation and update of packages. Even this is not possible for 3rd party packages not signed/published by Synology, they removed the wizard documentation in the DSM Developer Guide since DSM 7 beta.

Ok, there is some code to create backups before uninstall.
But one thing to consider is, the configuration data located under the web folder will be lost, even when the var folder is not removed und DSM 7 uninstall.
So I propose to add an uninstall feature to keep the data (on DSM 7) and to use the var folder for the configuration files.

...

spk/owncloud/src/service-setup.sh
mreid-tt: > the same applies to the wizard *.sh files

@hgy59 Unfortunately, I don't believe this is true in reality. I had previously shown from my own tests under https://github.com/SynoCommunity/spksrc/pull/5619#discussion_r1134516271 that the `SYNOPKG_PKGDEST_VOL` variable is only present during upgrade on DSM 6 and under upgrade and uninstall under DSM 7. This is why these variables were hard coded with fallback values if not defined.

spk/owncloud/src/service-setup.sh
mreid-tt: Thanks, will try that approach.

mreid-tt:
@hgy59 thanks for your detailed review, I really appreciate it. Some notes on your comments...

> I highly recommend not to evaluate web folders etc. in the wizard *.sh files.
> 
> Use as less logic in the wizards as possible. Synology's intention is still to drop support for the wizards, so it is better to leave such logic in the service_* functions in service-setup.sh

Given that I do need to know where the web app resides, how would I move that logic to service_* functions? Are we considering something like a service-wizard.sh that would evaluate all the needed variables and set them as needed globally using something like an environment variable export? If so, I would call this logic at the start of the wizard once and then use the variables in memory?

> Synology prefers the unattended installation and update of packages. Even this is not possible for 3rd party packages not signed/published by Synology, they removed the wizard documentation in the DSM Developer Guide since DSM 7 beta.

Hmm, okay. A lot of the inspiration for the wizard logic was drawn from official packages like phpMyAdmin so if they are just going toward unattended only this means that a lot of apps will need to move core configuration elements to be editable within the app.

> Ok, there is some code to create backups before uninstall. But one thing to consider is, the configuration data located under the web folder will be lost, even when the var folder is not removed und DSM 7 uninstall. So I propose to add an uninstall feature to keep the data (on DSM 7) and to use the var folder for the configuration files.

I'm not sure what you mean here. With ownCloud pretty much all of the configuration is within the web folder which is in the share. So in some of the tests I've done where I just uninstall and don't keep a backup, on re-install if it connects to the same share the app pretty much just picks up where it left off. I didn't test this with multiple user accounts but I can look more into this.

spk/owncloud/src/service-setup.sh
hgy59: my analysis on 30.05.2021 with DSM 6.2-25556 and DSM 7.0-41222 showed that `SYNOPKG_PKGDEST_VOL` is defined the in the following context (without difference between DSM 6 and DSM 7):

install:
- validate_preinst
- service_preinst
- service_postinst
- service_prestart

uninstall:
- service_poststop
- validate_preuninst
- service_preuninst
- service_postuninst

upgrade:
- service_poststop
- validate_preupgrade
- service_preupgrade
- service_postupgrade
- validate_preuninst
- validate_preinst
- service_preuninst
- service_postuninst
- service_preinst
- service_postinst
- service_prestart

_I cannot imagine that this behaviour is different with the actual versions_

mreid-tt:
@hgy59, I've implemented a new backup and restore function which was missing before but now complete. I've also experimented with some workarounds for different journeys based on the installation type (new or restore), inspired by the MariaDB package scripts. It's a bit of a hack but works well. This should be the final version for review.

PR #5920, opened 2023-11-08
ownCloud: Update PHP FPM in DSM 6

mreid-tt:
## Description

This is a follow-on from #5912 which fixes the PHP FPM handling in DSM 6 by making all functions run as the `http` user. Code has also been added to revert modification to PHP template defaults from previous versions.

Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/spk/owncloud/Makefile
+++ b/spk/owncloud/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = owncloud
 SPK_VERS = 10.13.2
-SPK_REV = 12
+SPK_REV = 13
 SPK_ICON = src/owncloud.png
 
 DEPENDS = cross/owncloud
@@ -18,7 +18,7 @@ HOMEPAGE = https://owncloud.org/
 
 LICENSE = AGPL
 
-WIZARDS_TEMPLATES_DIR = src/wizard_templates/
+WIZARDS_TEMPLATES_DIR = src/wizard_templates
 SERVICE_WIZARD_SHARENAME = wizard_data_share
 USE_DATA_SHARE_WORKER = yes
 
@@ -31,6 +31,13 @@ ADMIN_PROTOCOL = https
 DSM_UI_DIR = app
 DSM_UI_CONFIG = src/app/config
 CONF_DIR = src/conf/
+
+include ../../mk/spksrc.common.mk
+
+# Alternate conf dir for DSM 6
+ifeq ($(call version_lt, ${TCVERSION}, 7.0),1)
+CONF_DIR = src/conf_6/
+endif
 
 SERVICE_SETUP = src/service-setup.sh
 STARTABLE = no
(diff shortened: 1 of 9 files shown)


mreid-tt:
@hgy59, the main intent of this PR is to convert the behaviour of the DSM 6 deployment of the web interface. Previously, my original workaround tried to mirror the DSM 7 behaviour where we have all the files owned by the package user `sc-owncloud`. This unfortunately also required the modification of the PHP FPM to also run as this user.

When considering a review of the work in https://github.com/SynoCommunity/spksrc/pull/5862 I realised that having other packages follow this same path would create a big issue since the PHP FPM can only run as one user account. Given that by default it runs as `http` I did a bunch of work to reconfigure both the data share and the executables and the files to all run as this user under DSM 6.

What made the process more complex is the handling of previous deployments which ran as the package user account. This required the implementation of a migration strategy from the old to the new user structure (including at the data share level), as well as mitigations for any backups previously executed with the old account.

Finally, what became challenging is that after making changes to the web interface in DSM 6, I would usually trigger a restart of Apache to refresh the server including the PHP FPM configurations. I found that once there instances of PHP FPM configurations outside of the one for this package, the installer would wait endlessly for that other package to stop before proceeding. As a rough workaround, I've added a check for this and the installer log includes a warning that the DSM needs to be rebooted for the new configs to be fully loaded.

The end user experience if this occurs is that the web interface either does not load or throws a PHP extension error. I admit this is not ideal and perhaps we could have something in the wizard to warn about this but I know this would add even more bulk to the wizard which I was hesitant to do. I also considered that there may not be many users that run combinations of ownCloud and Selfoss so it may not come up that often.

Let me know your thoughts on the approach.

mreid-tt:
@hgy59, I'd like to publish this version and then deactivate version 10.13.2-12. let me know if you have any objections moving forward given the summary above.

mreid-tt:
@hgy59, I've successfully published the latest update, and it's visible in my NAS. Could you please remove version 10.13.2-12 from the repository entirely? I've already deactivated it, but I'd appreciate it if you could take care of the removal when you have some free time. Thanks!

hgy59:
@mreid-tt I prefer to wait another day or two until the current version is proved to be visible in the package center of DSM on the devices.
This helps to avoid discussions when someone wants to installal the version `10.13.2-12` as newest and gets download errors (my devices currently still show this version as newest in package center).

hgy59:
@mreid-tt deleted owncloud 10.13.2-12 since all of my devices show 10.13.2-13 in the package center.

mreid-tt:
> @mreid-tt deleted owncloud 10.13.2-12 since all of my devices show 10.13.2-13 in the package center.

Thanks @hgy59, appreciate it.

PR #5923, opened 2023-11-13
tt-rss: upgrade 20221230 to 20230828

mreid-tt:
## Description

1. Upgrade tt-rss to version 20230828
2. Migrate UI wizard to wizard templates
3. Modernise package scripts
4. Fix DSM 6 app startup permissions
5. Add PHP profiles for DSM 6

Amends #5862

## Checklist

### Type of change

- [x] Package update


smaarn:
Have you checked the behaviour of the daemon script ? I find it suspicious that you didn't need to change the installer logic since it specifies virtual php script as php entry point.

mreid-tt:
> Have you checked the behaviour of the daemon script ? I find it suspicious that you didn't need to change the installer logic since it specifies virtual php script as php entry point.

The daemon script works perfectly as far as I can see. Attached is a log of me installing a fresh tt-rss install, restarting DSM and then subscribing to a feed. Everything seems works just fine as far as I can tell. What concerns you have exactly?

[daemon.log](https://github.com/SynoCommunity/spksrc/files/13355882/daemon.log)

smaarn:
> Have you checked the behaviour of the daemon script ? I find it suspicious that you didn't need to change the installer logic since it specifies virtual php script as php entry point.

To be more specific the daemon script was the very reason for the virtual-php script because it required some extensions to be enabled.

mreid-tt:
> To be more specific the daemon script was the very reason for the virtual-php script because it required some extensions to be enabled.

I could find no issues running the simplified daemon. I did see in the legacy files that it mainly did two things (a) identify the default PHP version and (b) extract the enabled extensions.

My thought process for the simplified version was that (a) we don't need to find the default PHP because the new config code assigns this so it is known and (b) the [Update daemon](https://tt-rss.org/wiki/UpdatingFeeds#update-daemon) documentation does not seem to require any extensions.

From what I've observed, the new feed was added and it also downloaded the icon for the feed and correctly displayed news articles.

smaarn:
> > To be more specific the daemon script was the very reason for the virtual-php script because it required some extensions to be enabled.
> 
> 
> 
> I could find no issues running the simplified daemon. I did see in the legacy scripts that they mainly did two things: (a) identify the default PHP version and (b) extract the enabled extensions.
> 
> 
> 
> My thought process for the simplified daemon was that (a) we don't need to find the default PHP because the new config code assigns this so it is known and (b) the [Update daemon](https://tt-rss.org/wiki/UpdatingFeeds#update-daemon) documentation does not seem to require any extensions.
> 
> 
> 
> From what I've observed, the new feed was added and it also downloaded the icon for the feed and correctly displayed news articles.
> 
> 
> 
> EDIT: Are there any specific tests you would like me to run to validate the setup further?

To validate the setup further you would need to validate the "previous version" migration case: you may need to handle the change of the PHP configuration (or remove it altogether).
Ideally validation Maria DB5 upgrade could be worth checking (since this bit is slightly impacted)

mreid-tt:
> To validate the setup further you would need to validate the "previous version" migration case: you may need to handle the change of the PHP configuration (or remove it altogether).

I can try a clean install of version 20221230-16 and then upgrade with this build. I don't think we'll need to do anything with the PHP configuration since it would have been user configured before. Will let you know the result.

> Ideally validation Maria DB5 upgrade could be worth checking (since this bit is slightly impacted)

From what I see in version 20210902-15 it says "_Migrated to Maria DB 10 and add support for DSM 7_". As such, I'd have to go back to version 20200829-12 which presumably is running Maria DB 5? That seems like a lot to test.

mreid-tt:
Okay, here's the upgrade test results:

1. Did a fresh install of 20221230-16 along with clean installs of all dependencies
2. Found that the install generated errors until I set PHP default profile then the install completed with only one error
   - The error appeared non-critical (`ERROR 1045 (28000): Access denied for user 'ttrss'@'localhost' (using password: YES)`)
3. After installation, the app would not start since the daemon was trying to run as root. After changing the ownership of the `daemon.log` file, I was manually able to start the app using `sudo -u http /var/packages/tt-rss/target/bin/tt-rss-daemon`
4. I configured a feed and ensured that it was updating via the daemon
5. I then updated the app using the current build. The `service_restore` reported the same two file move errors I reported earlier but upgraded okay
6. The new daemon log at `/var/services/web/tt-rss/logs/daemon.log` reported an error `sh: /volume1/@appstore/tt-rss/bin/virtual-php: No such file or directory`
7. After investigating, I found that the restored `config.php` file had the following: `putenv('TTRSS_PHP_EXECUTABLE=/volume1/@appstore/tt-rss/bin/virtual-php');`. Changing this to `putenv('TTRSS_PHP_EXECUTABLE=/usr/local/bin/php74');` fixed it
8. Upgraded app functions correctly but as reported earlier, configured feed is missing icon

So all in all the upgrade seemed to go well. I think I'll need to add some logic to the `service_restore` to change the `TTRSS_PHP_EXECUTABLE` variable in the configuration if incorrect.

mreid-tt:
@smaarn, I have pushed a commit which addresses the issue identified with the upgrade from previous version.

mreid-tt:
@smaarn, if you are okay with the changes I'd be happy to push the new commits to your original PR so I can delete this one.

smaarn:
> @smaarn, if you are okay with the changes I'd be happy to push the new commits to your original PR so I can delete this one.

Actually I don't mind if you wish to merge your PR instead since you've been doing all the cleanup phase.
To be perfectly transparent I don't have anyway to test this currently (computer dying right now) so I'm not sure I'll be of any help for the coming week either.

spk/tt-rss/src/service-setup.sh
```diff
@@ -26,145 +28,298 @@ SVC_KEEP_LOG=y
...
+      "${WEB_DIR}/${PACKAGE}/config.php"
+    echo "putenv('TTRSS_PHP_EXECUTABLE=${PHP}');">>"${WEB_DIR}/${PACKAGE}/config.php"
+  fi
+  if [ "${SPK_REV}" -lt "15" ]
+  then
+    sed -i -e "s|putenv('TTRSS_DB_PASS=.*');|putenv('TTRSS_DB_PASS=${wizard_mysql_password_ttrss}');|" \
+      "${WEB_DIR}/${PACKAGE}/config.php"
+    echo "putenv('TTRSS_MYSQL_DB_SOCKET=/run/mysqld/mysqld10.sock');">>"${WEB_DIR}/${PACKAGE}/config.php"
+  fi
+  # Check config file for legacy PHP exec to migrate
```
smaarn, line 302: You could restrict this check to only installation older than the one being published

spk/tt-rss/src/web/tt-rss.conf
```diff
@@ -0,0 +1,16 @@
+Alias "/tt-rss" "/var/services/web/tt-rss"
+
+<Proxy "unix:/run/php-fpm/php-com-synocommunity-packages-tt-rss.sock|fcgi://tt-rss">
+    ProxySet timeout=3600
+</Proxy>
+
+<Directory "/var/services/web/tt-rss">
+
+    <FilesMatch "\.(php[345]?|phtml)$">
```
smaarn, line 9: I'm not sure there is a need for supporting php3, php4 and php5 files... From what I recall I've only seen php files

spk/tt-rss/Makefile
```diff
@@ -1,43 +1,47 @@
...
+CONF_DIR = src/conf_6/
+endif
 
-# Due to not obvious WebStation handling requirements
-REQUIRED_MIN_DSM = 6.0
-INSTUNINST_RESTART_SERVICES = apache-web
+POST_STRIP_TARGET = tt-rss_extra_install
 
-SPK_DEPENDS = "WebStation:PHP7.4:MariaDB10:Apache2.4"
+PHP_EXTENSIONS := $(shell jq -r '.webservice.services[0].php.extensions[]' src/conf/resource)
```
smaarn, line 42: I don't believe you are still using this are you ?
(the point of this was to not have to redefine this list of PHP extensions)

smaarn (approved):
Only a couple of minor comments.
No objection on the principle, not too fond of reformatting files (makes it harder to review changes) but that's another topic.
As indicated I don't have access to any working personal computer as of now so I wouldn't be able to test nor elaborate for the coming week(s).

spk/tt-rss/src/service-setup.sh
mreid-tt: I'm not sure i understand. This appears in a code block which checks SPK_REV < 15. Is it that this condition is changed? Do you prefer that I use something like the SYNOPKG_OLD_PKGVER variable to check the package version instead?

spk/tt-rss/src/web/tt-rss.conf
mreid-tt: This is for the Apache config. The format was taken from configs generated by DSM so I'd prefer to leave them as-is

spk/tt-rss/Makefile
mreid-tt: Agreed, removed...

spk/tt-rss/src/service-setup.sh
smaarn: I mean that any existing installation (16 at least) would be using the virtual-php binary.
Either you ensure the setting is updated for those or you add a basic redirection.

mreid-tt: I think I get you now, the quoted lines were 299-302 and you are probably referring to the lines 302 onwards for the check for the virtual-php binary. The way it is constructed currently is not based on package version. It will check for any version the contents of the `config.php` file to see if the line for the variable `TTRSS_PHP_EXECUTABLE` exists. If it does, it checks that it matches the declared PHP binary. If not, it will replace the line to match.

mreid-tt: @smaarn, for consistency with your previous logic, I've adjusted it. Let me know if this works for you.

smaarn:
LGTM

smaarn (approved):

PR #5927, opened 2023-11-15
[Framework] Add error detection to python pip calls

### LINKED ISSUE #5926: [Framework] Install failures for Python packages
I've observed some recent issues with Python-based package installations, wherein the installation process fails, but erroneously signals successful completion to users (refer to issues #5884 and #5915). Looking into this, we use wheels to distribute Python packages and based on the [wiki page](https://github.com/SynoCommunity/spksrc/wiki/Using-wheels-to-distribute-Python-packages) there are two steps in the `postinstall` which are executed:

```
    # Create a Python virtualenv
    install_python_virtualenv

    # Install the wheels
    install_python_wheels
```

Digging into each of these functions in turn we find them constructed as follows:

https://github.com/SynoCommunity/spksrc/blob/6d0935e9c7bd16cf0b3a912f80091e5f0306bae3/mk/spksrc.service.installer.functions#L74-L78

I've seen some [evidence](https://github.com/SynoCommunity/spksrc/issues/5915#issuecomment-1805196889) that the pip upgrade fails and this can have a cascading effect on the rest of the install. It may be prudent to add some code to replace the last line:

```
if ! python3 -m pip install --upgrade pip; then
    echo "Error: Python upgrade failed"
    exit 1
fi
```

For the second function we see the following:

https://github.com/SynoCommunity/spksrc/blob/6d0935e9c7bd16cf0b3a912f80091e5f0306bae3/mk/spksrc.service.installer.functions#L93-L102

As with the above recommendation, some error checking may be prudent so that the user is not incorrectly informed of a successful installation.

```
if [ -s ${requirements} ]; then 
    echo "Install packages from wheels [${requirements}]"
    pip install $([ $# -gt 0 ] && echo $*) \
                --force-reinstall \
                --cache-dir ${cachedir} \
                --find-links ${wheelhouse} \
                --requirement ${requirements} \
    || { echo "Error: Python package installation failed"; exit 1; }
fi
```

@hgy59, what do you think of this approach?

mreid-tt:
## Description

This proposed framework update aims to rectify issues associated with Python-based package installations. Specifically, it addresses a situation where, in the presence of installation errors, the system erroneously reports a successful installation to the user.

Fixes #5926

## Checklist

### Type of change

- [ ] Bug fix

### DIFF
--- a/mk/spksrc.service.installer.functions
+++ b/mk/spksrc.service.installer.functions
@@ -75,9 +75,16 @@ install_python_virtualenv ()
     python3 -m venv --system-site-packages ${SYNOPKG_PKGDEST}/env
 
     # Update pip to match default used by python package
-    # If none found, update pip to latest version
+    # If none found, update pip to latest version online
     read -r -a PIP_VERSION 2>/dev/null <<< $(python3 -m pip --version)
-    [ "${PIP_VERSION[1]}" ] && PIP="pip==${PIP_VERSION[1]}" || PIP="pip"
+    if [ "${PIP_VERSION[1]}" ]; then
+        PIP="pip==${PIP_VERSION[1]}"
+        echo "INFO: Default pip version found ${PIP}"
+    else
+        PIP="pip"
+        echo "WARNING: No pip found, using default online version"
+    fi
+
     if ! python3 -m pip install --upgrade ${PIP}; then
     echo "Error: Python upgrade failed" 1>&2
         exit 1
--- a/mk/spksrc.service.installer.functions
+++ b/mk/spksrc.service.installer.functions
@@ -68,25 +68,15 @@ initialize_variables ()
 
 install_python_virtualenv ()
 {
-    # Output Python version (confirming PATH)
+    # Print out default python version found in PATH
     python3 --version
+
+    # Print out default pip version
+    read -r -a PIP_VERSION 2>/dev/null <<< $(python3 -m pip --version)
+    echo "INFO: Default pip version found ${PIP_VERSION[1]}"
 
     # Create a Python virtualenv
     python3 -m venv --system-site-packages ${SYNOPKG_PKGDEST}/env
-
-    # Update pip to match default used by python package
-    # If none found, update pip to latest version online
-    read -r -a PIP_VERSION 2>/dev/null <<< $(python3 -m pip --version)
-    if [ "${PIP_VERSION[1]}" ]; then
-        PIP="pip==${PIP_VERSION[1]}"
-        echo "INFO: Default pip version found ${PIP}"
-    else
-        PIP="pip"
-        echo "WARNING: No pip found, using default online version"
-    fi
-
-    python3 -m pip install --upgrade ${PIP} \
-       || echo "ERROR: Python upgrade failed" 1>&2
 
     if [ ${SYNOPKG_DSM_VERSION_MAJOR} -lt 7 ]; then
         set_unix_permissions ${SYNOPKG_PKGDEST}/env
--- a/mk/spksrc.service.installer.functions
+++ b/mk/spksrc.service.installer.functions
@@ -74,8 +74,11 @@ install_python_virtualenv ()
     # Create a Python virtualenv
     python3 -m venv --system-site-packages ${SYNOPKG_PKGDEST}/env
 
-    # Update to latest pip package installer
-    if ! python3 -m pip install --upgrade pip; then
+    # Update pip to match default used by python package
+    # If none found, update pip to latest version
+    read -r -a PIP_VERSION 2>/dev/null <<< $(python3 -m pip --version)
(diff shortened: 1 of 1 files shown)


mk/spksrc.service.installer.functions
```diff
@@ -75,7 +75,10 @@ install_python_virtualenv ()
     python3 -m venv --system-site-packages ${SYNOPKG_PKGDEST}/env
 
     # Update to latest pip package installer
-    python3 -m pip install --upgrade pip
+    if ! python3 -m pip install --upgrade pip; then
+    echo "Error: Python upgrade failed" 1>&2
+        exit 1
+    fi
```
hgy59, line 81: we should only log, but not exit on error.
the following set_unix_permissions is mandatory for DSM 6.

th0ma7: Suggesting we always refer back to python packaged pip version... such as:
```
read -r -a PIP_VERSION 2>/dev/null <<< $(python3 -m pip --version)
[ "${PIP_VERSION[1]}" ] && PIP="pip==${PIP_VERSION[1]}" || PIP="pip"
if ! python3 -m pip install --upgrade ${PIP}; then
   echo "Error: Python upgrade failed" 1>&2
fi
```

mk/spksrc.service.installer.functions
```diff
@@ -98,7 +101,8 @@ install_python_wheels ()
                    --force-reinstall \
                    --cache-dir ${cachedir} \
                    --find-links ${wheelhouse} \
-                   --requirement ${requirements}
+                   --requirement ${requirements} \
+       || { echo "Error: Python package installation failed" 1>&2; exit 1; }
```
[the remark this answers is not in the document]
th0ma7: So I did a bit of testing and the `exit 1` makes the package "looks" like installed correctly... so it ends-up failing silently.  Similarly, removing it turns-out to have the exact same behavior...

mreid-tt:
@th0ma7, is this good to merge now?

th0ma7:
I believe so.  @hgy59 are you fine with comments above and merging?

th0ma7:
Catching-up... @hgy59 any issues with merging this?  (see comments above)

hgy59:
I would drop the whole pip check/update/downgrade stuff in `install_python_virtualenv ()`

and add `--disable-pip-version-check` to pip install command in `install_python_wheels ()`

th0ma7:
> and add `--disable-pip-version-check` to pip install command in `install_python_wheels ()`

forgot, good call, done

th0ma7:
> I would drop the whole pip check/update/downgrade stuff in `install_python_virtualenv ()`

I've been re-reading the code and indeed it`s useless:  its been using default `python3` from PATH and checking its `pip` version, thus updating if not matching default `python3` package (being called in the first place from PATH.

Now only printing it out.

(untested yet but change is trivial) - feel free to test in the meantime.

mk/spksrc.service.installer.functions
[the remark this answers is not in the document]
th0ma7: > Oh! So this is not a solution after all. As an alternative, am I correct in that if we use this argument the wheels components will be included in the package and not downloaded?
> 
> ```
> WHEELS_PURE_PYTHON_PACKAGING_ENABLE = 1
> ```

(delayed answer) Yes you are correct.  This was introduce a while ago in order to reduce the package size by only providing by default the cross-compiled wheels and leaving everything else to be dealt with locally at install time.

th0ma7 (approved):
Tested this morning and working well.  It actually expedite the installation a bit.

mk/spksrc.service.installer.functions
```diff
@@ -68,15 +68,16 @@ initialize_variables ()
...
 install_python_virtualenv ()
 {
-    # Output Python version (confirming PATH)
+    # Print out default python version found in PATH
     python3 --version
 
+    # Print out default pip version
+    read -r -a PIP_VERSION 2>/dev/null <<< $(python3 -m pip --version)
+    echo "INFO: Default pip version found ${PIP_VERSION[1]}"
+
```
hgy59, line 77: @mreid-tt (/cc @th0ma7) This change breaks the installer on all SRM and DSM 5.2 systems

This file must be `/bin/sh` (i.e. ash) compatible.
At least `${PIP_VERSION[1]}` is valid for `/bin/bash` only and breaks the installer in `/bin/sh` environments for all kind of packages - not only packages installing a python virtualenv.

**Please regard lines 5-6 at the top of this file.**

th0ma7: The culprit is most probably:
```
read -r -a PIP_VERSION 2>/dev/null <<< $(python3 -m pip --version)
```
A potential fix would be:
```
read -r -a PIP_VERSION 2>/dev/null << EOF
$(python3 -m pip --version)
EOF
```

mreid-tt: Resolved by #6028.

PR #5957, opened 2023-12-12
COPS: Update to version 1.5.4

mreid-tt:
## Description

This PR contains the update to the last published version of COPS which is compatible with PHP 7.4. The subsequent versions split off resources in preparation of 2.x release. The 2.x.x releases will support PHP >= 8.1 and requires a newer version of PHP.

This pull request introduces enhancements for DSM 6 installations, streamlining the user experience by automatically configuring necessary PHP settings. Additionally, it incorporates a data share worker to automate the configuration of permissions, further improving the overall quality of life for users.

Fixes # 

## Checklist

### Type of change

- [x] Package update


hgy59 (commented):
please keep changes minimal 

- why did you change the order of the lines in app/config?
- why did you add variables for jq, sed and chmod?
- why did you replace `cp`by `${CP}`? (this is not the same, CP = cp -rfp)
- why did you merge the comments regarding config and .htaccess file together?
- ...

this makes it difficult to review the real changes.


_if you do this for the sake of adjusting different packages with web apps, then please do not do this on regular package updates, but create a dedicated PR with all affected packages - but without changing the SPK_REV of SPK_VERS nor publishing the packages._

But there is more to do for web app packages:
I hope we can create generic helper functions for packages with apache_php configuration.

mreid-tt:
> please keep changes minimal

apologies, I didn't consider the challenge I would create by making a bunch of tweaks in addition to substantive changes. I will try to be more considerate in the future.

> * why did you change the order of the lines in app/config?

This was cosmetic to more closely match the order presented in your `demoservice` and `demowebservice` examples which I modelled most of the packages I've worked on recently.

> * why did you add variables for jq, sed and chmod?

This was also cosmetic similar to those introduced in the ownCloud modernisation. I liked the idea of using variables for executables so that I could easily identify when an execution is made with any special parameters.

> * why did you replace `cp`by `${CP}`? (this is not the same, CP = cp -rfp)

Agreed that it is not the same but I found the added parameters in the framework useful to preserve attributes. This was also my preference for using variables for executables.

> * why did you merge the comments regarding config and .htaccess file together?

I didn't like original code which included separate variables for the paths and the filenames, so I collapsed these down to make more readable and similar to other `service_save` and `service_restore` functions I worked on recently.

> * ...
> 
> this makes it difficult to review the real changes.

I do appreciate that some of these cosmetic items are more about style and consistency with other packages so I'll try to be more considerate in the future.

> _if you do this for the sake of adjusting different packages with web apps, then please do not do this on regular package updates, but create a dedicated PR with all affected packages - but without changing the SPK_REV of SPK_VERS nor publishing the packages._

Ah! This makes sense. I was taking on modernising these older packages one at a time so I didn't take a holistic view of similar packages but will definitely consider this in the future.

> But there is more to do for web app packages: I hope we can create generic helper functions for packages with apache_php configuration.

Is that related to the work I did for the DSM 6 workarounds to support the PHP configs?

mreid-tt:
@hgy59, to streamline the service setup script, I've removed my cosmetic preference regarding variables for executables.

PR #5931, opened 2023-11-19
Update Gitea to v1.20.6

wkobiela:
## Checklist

- [x] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #5962, opened 2023-12-19
COPS: Enhance Upgrade Process for Legacy Versions

### LINKED ISSUE #5961: COPS fails to update
### Is this a new Bug?

- [X] I checked that the bug hasn't been reported before

### Package Name

COPS

### Package Version

1.5.4-8

### Device Model

VirtualDSM

### Device Architecture

x86_64

### Firmware Version

DSM 7.2.1-69057 Update 3

### What happened?

installation fails

### Reproduction steps

1. 
2. 
3. 
...

### Install Log

```log
2023/12/18 21:48:17     ===> Step postinst. STATUS=UPGRADE USER=cops GROUP=http SHARE_PATH=
2023/12/18 21:48:17     Begin save_wizard_variables
2023/12/18 21:48:17     End save_wizard_variables
2023/12/18 21:48:17     Begin syno_sync_var_folder
2023/12/18 21:48:17     End syno_sync_var_folder
2023/12/18 21:48:17     Begin service_postinst
2023/12/18 21:48:17     End service_postinst
2023/12/18 21:48:17     upgrade cops 1.5.4-8 End postinst ret=[0]
2023/12/18 21:48:17     upgrade cops 1.5.4-8 Begin postupgrade
2023/12/18 21:48:17     Begin initialize_variables
2023/12/18 21:48:17     End initialize_variables
2023/12/18 21:48:17     ===> Step postupgrade. STATUS=UPGRADE USER=cops GROUP=http SHARE_PATH=
2023/12/18 21:48:17     Begin service_restore
2023/12/18 21:48:17     Restore previous data from /volume1/@tmp/synopkg.tmp/cops/cops
2023/12/18 21:48:17     rsync: change_dir#3 "/var/services/web_packages/cops" failed: No such file or directory (2)
2023/12/18 21:48:17     rsync error: errors selecting input/output files, dirs (code 3) at main.c(729) [Receiver=3.1.2]
2023/12/18 21:48:17     rsync: change_dir#3 "/var/services/web_packages/cops" failed: No such file or directory (2)
2023/12/18 21:48:17     rsync error: errors selecting input/output files, dirs (code 3) at main.c(729) [Receiver=3.1.2]
2023/12/18 21:48:17     End service_restore
2023/12/18 21:48:17     Begin syno_sync_var_folder
2023/12/18 21:48:17     End syno_sync_var_folder
2023/12/18 21:48:17     upgrade cops 1.5.4-8 End postupgrade ret=[0]
```

### Service Log

_No response_

### Other Logs

_No response_

mreid-tt:
## Description

This builds upon the developments in PR #5957 by introducing an upgrade wizard. The wizard is designed to identify the presence of a data share; if none is found, it extracts the library folder name to propose it as the designated data share for the upgrade. This enhancement addresses issues in previous packages where the absence of a defined data share led to upgrade failures.

Fixes #5961

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/spk/cops/src/conf/resource
+++ b/spk/cops/src/conf/resource
@@ -1,57 +1,55 @@
 {
-  "data-share": {
-    "shares": [{
-      "name": "{{wizard_calibre_share}}",
-      "once": true,
-      "permission": {
-        "rw": ["sc-cops"]
-      }
-    }]
-  },
-  "webservice": {
-    "migrate": {
-      "root": [{
-        "new": "cops",
-        "old": "cops"
-      }]
-    },
-    "pkg_dir_prepare": [{
-      "group": "http",
-      "mode": "0755",
-      "source": "/var/packages/cops/target/share/cops",
-      "target": "cops",
-      "user": "sc-cops"
-      }],
-    "portals": [
-      {
-        "alias": "cops",
-        "app": "com.synocommunity.packages.cops",
-        "name": "COPS",
-        "service": "cops",
-        "type": "alias"
-      }
-    ],
-    "services": [{
-      "backend": 2,
-      "display_name": "COPS",
-      "icon": "app/images/cops-{0}.png",
-      "php": {
-        "backend": 8,
-        "extensions": [
-          "gd",
-          "intl",
-          "pdo_sqlite",
-          "sqlite3",
-          "zlib"
-        ],
-        "group": "http",
-        "profile_desc": "PHP profile used by cops",
-        "profile_name": "cops",
-        "user": "sc-cops"
-      },
-      "root": "cops",
-      "service": "cops",
-      "type": "apache_php"
-    }]
-  }
+	"data-share": {
+		"shares": [{
+			"name": "{{wizard_calibre_share}}",
+			"once": true,
+			"permission": {
+				"rw": ["sc-cops"]
+			}
+		}]
+	},
+	"webservice": {
+		"migrate": {
+			"root": [{
+				"new": "cops",
+				"old": "cops"
+			}]
+		},
+		"pkg_dir_prepare": [{
+			"group": "http",
+			"mode": "0755",
+			"source": "/var/packages/cops/target/share/cops",
+			"target": "cops",
+			"user": "sc-cops"
+		}],
+		"portals": [{
+			"alias": "cops",
+			"app": "com.synocommunity.packages.cops",
+			"name": "COPS",
+			"service": "cops",
+			"type": "alias"
+		}],
+		"services": [{
+			"backend": 2,
+			"display_name": "COPS",
+			"icon": "app/images/cops-{0}.png",
+			"php": {
+				"backend": 8,
+				"extensions": [
+					"gd",
+					"intl",
+					"pdo_sqlite",
+					"sqlite3",
+					"zlib"
+				],
+				"group": "http",
+				"profile_desc": "PHP profile used by cops",
+				"profile_name": "cops",
+				"user": "sc-cops"
+			},
+			"root": "cops",
+			"service": "cops",
+			"type": "apache_php"
+		}]
+	}
 }
--- a/spk/cops/src/conf/resource
+++ b/spk/cops/src/conf/resource
@@ -1,4 +1,13 @@
 {
+  "data-share": {
+    "shares": [{
+      "name": "{{wizard_calibre_share}}",
+      "once": true,
+      "permission": {
+        "rw": ["sc-cops"]
+      }
+    }]
+  },
   "webservice": {
     "migrate": {
       "root": [{
@@ -30,9 +39,9 @@
         "backend": 8,
         "extensions": [
           "gd",
+          "intl",
           "pdo_sqlite",
           "sqlite3",
-          "intl",
           "zlib"
         ],
         "group": "http",
@@ -43,15 +52,6 @@
       "root": "cops",
       "service": "cops",
       "type": "apache_php"
-    }]
-  },
-  "data-share": {
-    "shares": [{
-      "name": "{{wizard_calibre_share}}",
-      "permission": {
-        "rw": ["sc-cops"]
-      },
-      "once": true
     }]
   }
 }
--- a/spk/cops/src/wizard_templates/upgrade_uifile_fre.yml
+++ b/spk/cops/src/wizard_templates/upgrade_uifile_fre.yml
@@ -1,6 +1,6 @@
 COPS_CONFIGURATION_UPGRADE_STEP_TITLE: "Mise à jour de la bibliothèque COPS"
 
-UPGRADE_CALIBRE_DIRECTORY_DESCRIPTION: "Le répertoire de la bibliothèque Calibre doit désormais résider dans un partage de données. Sur la base de la configuration précédente détectée, veuillez confirmer que le partage de données spécifié contient votre bibliothèque."
-UPGRADE_CALIBRE_DIRECTORY_LABEL: "Partager le nom"
+UPGRADE_CALIBRE_DIRECTORY_DESCRIPTION: "Le répertoire de la bibliothèque Calibre doit désormais résider dans un dossier partagé. Sur la base de la configuration précédente détectée, veuillez confirmer que le dossier partagé spécifié contient votre bibliothèque."
+UPGRADE_CALIBRE_DIRECTORY_LABEL: "Dossier partagé"
 UPGRADE_CALIBRE_DIRECTORY_VALIDATION_ERROR_TEXT: "Les sous-répertoires ne sont pas pris en charge."
-UPGRADE_NOTE_CALIBRE_DIRECTORY_DESCRIPTION: "REMARQUE: si votre bibliothèque Calibre ne se trouve pas dans le partage spécifié, un nouveau partage sera créé. Après la mise à niveau, déplacez manuellement la bibliothèque vers ce nouvel emplacement."
+UPGRADE_NOTE_CALIBRE_DIRECTORY_DESCRIPTION: "REMARQUE: si votre bibliothèque Calibre ne se trouve pas dans le dossier partagé spécifié, un nouveau sera créé. Après la mise à niveau, déplacez manuellement la bibliothèque vers ce nouvel emplacement."
--- a/spk/cops/src/wizard_templates/upgrade_uifile_fre.yml
+++ b/spk/cops/src/wizard_templates/upgrade_uifile_fre.yml
@@ -1,0 +1,6 @@
+COPS_CONFIGURATION_UPGRADE_STEP_TITLE: "Mise à jour de la bibliothèque COPS"
+
+UPGRADE_CALIBRE_DIRECTORY_DESCRIPTION: "Le répertoire de la bibliothèque Calibre doit désormais résider dans un partage de données. Sur la base de la configuration précédente détectée, veuillez confirmer que le partage de données spécifié contient votre bibliothèque."
+UPGRADE_CALIBRE_DIRECTORY_LABEL: "Partager le nom"
+UPGRADE_CALIBRE_DIRECTORY_VALIDATION_ERROR_TEXT: "Les sous-répertoires ne sont pas pris en charge."
+UPGRADE_NOTE_CALIBRE_DIRECTORY_DESCRIPTION: "REMARQUE: si votre bibliothèque Calibre ne se trouve pas dans le partage spécifié, un nouveau partage sera créé. Après la mise à niveau, déplacez manuellement la bibliothèque vers ce nouvel emplacement."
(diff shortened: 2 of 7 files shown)


hgy59:
@mreid-tt with the shared folder handling, my hope was, that we do not need an upgrade wizard package just for the sake of restoring the wizard variable for the share.

If it is not possible to restore (i.e. migrate) the installer-variables on package upgrade, we probably can make the upgrade page invisible in this case (just to provide the variable for the resource worker).

I doubt that your initial approach to add the upgrade page only when the share does not exist really works. 

You should try to avoid an upgrade wizard and fix the service-setup.sh for the upgrade.

PS: in #5961 the app was running before the upgrade, i.e. the shared folder exists and the app was working.

hgy59:
further info for #5961

The issue seems to be the missing SHARE_PATH and SHARE_NAME variables in `etc/installer-variables`.
Before the upgarde the installer-variables file contains only `GROUP=http`.

But at least on DSM 7, the variables could be recreated in preupgrade from the realpath of the existing link `/var/packages/cops/share` (in my case -> /volume1/@appshare/cops, i.e. SHARE_PATH = /volume1/cops, SHARE_NAME = cops)

mreid-tt:
@hgy59 Just a small tweak and the upgrade successfully works from version 1.4.2-7 to version 1.5.4-9. This was tested both on DSM 6 and DSM 7.

hgy59:
@mreid-tt you find code how to read the installer-variables in `spk/minio/src/wizard/ugrade_uifile.sh`
If defined, use SHARE_NAME for the wizard variable, if not, then extract the name from SHARE_PATH variable.

When implementing the shared folder redesign, I ignored the errors for missing variable in data share config and focused on keeping the SHARE_PATH and SHARE_NAME in the installer-variables.

mreid-tt:
> @mreid-tt you find code how to read the installer-variables in `spk/minio/src/wizard/ugrade_uifile.sh` If defined, use SHARE_NAME for the wizard variable, if not, then extract the name from SHARE_PATH variable.

Thanks for that reference. I assume you are speaking about this block:

https://github.com/SynoCommunity/spksrc/blob/a30bd0aabd1153cc3e00673470ecaf1c4946a5c1/spk/minio/src/wizard/upgrade_uifile.sh#L6-L15

If so, I am reading the SHARE_NAME variable in this line of my code:

https://github.com/SynoCommunity/spksrc/blob/70fd690d7a5183326095eb0c200110d6ab6285ab/spk/cops/src/wizard_templates/upgrade_uifile.sh#L27

I am then checking if it is not empty and if so, skip asking for user input. If it is empty, try to extract a share name from the PHP conf file and set it as the share name default. Is this is not a good approach?

> When implementing the shared folder redesign, I ignored the errors for missing variable in data share config and focused on keeping the SHARE_PATH and SHARE_NAME in the installer-variables.

Hmm, okay. Do you have any specific changes to my implementation to recommend?

mreid-tt:
hey @hgy59, I saw your https://github.com/SynoCommunity/spksrc/pull/5931#issuecomment-1865714588 where you said:

> @mreid-tt can you please have a look here at the update of the shared folder (with the findings of #5962).

... and I took a look at the code there. In this package there is a shared folder used in the Makefile but there is no conf_dir defined. Instead it relies on the spksrc to create one in the package. The generated file looks like this:

```
{
  "port-config": {
    "protocol-file": "app/gitea.sc"
  },
  "data-share": {
    "shares": [
      {
        "name": "{{wizard_shared_folder_name}}",
        "permission": {
          "rw": [
            "sc-gitea"
          ]
        }
      }
    ]
  }
}
```

On upgrade the wizard does not insert a value for `wizard_shared_folder_name`, but unlike with COPS there is no issue with the acquisition of a data share for gitea on package upgrade.

I believe thsi succeeds because of the fact that it was previously using the `SERVICE_WIZARD_SHARE` variable in the Makefile before migrating to `SERVICE_WIZARD_SHARENAME`. As a result the installer-variables file prior to upgrade looked like this:

```
SHARE_PATH=/volume1/git
```

Following the package upgrade it looks like this:

```
SHARE_PATH=/volume1/git
SHARE_NAME=git
```

This is very different than the COPS upgrade which newly introduced the data share. As a result the installer-variables prior to upgrade looks like this:

```
GROUP=http
```

In this scenario, the data share acquisition fails because it cannot determine the `SHARE_PATH`. As such, I believe I do need to have this checked in the upgrade wizard for COPS to upgrade successfully.

Am I interpreting your request to review #5931 correctly?

mreid-tt:
@hgy59 are we good to merge this?

hgy59:
@smaarn in #5817 you added the "migration of `.htaccess` file", but I could not find any reason/description what this is for.

What is the drawback when this file gets overwritten with the (new) one at package update?
IMHO it is not required to backup and restore `.htaccess` on package updates, since an installation does not modify this file, 
It might be breaking the package if an updated `.htaccess` is overwritten by an older version.

At least an update of the package must backup the new file, if the old one is restored (typically we rename new files *.new).
But I suppose we should remove the restore of `.htaccess` at all.

smaarn:
> @smaarn in #5817 you added the "migration of `.htaccess` file", but I could not find any reason/description what this is for.
> 
> 
> 
> What is the drawback when this file gets overwritten with the (new) one at package update?
> 
> IMHO it is not required to backup and restore `.htaccess` on package updates, since an installation does not modify this file, 
> 
> It might be breaking the package if an updated `.htaccess` is overwritten by an older version.
> 
> 
> 
> At least an update of the package must backup the new file, if the old one is restored (typically we rename new files *.new).
> 
> But I suppose we should remove the restore of `.htaccess` at all.
> 
> 

@hgy59 that's in this file that you can configure password and access settings. If we have a way to have it tailored to use synology users I'm guessing it would be fine to drop it but otherwise I would be keen on keeping it unless there are drawbacks which I am not aware of.

hgy59:
> @hgy59 that's in this file that you can configure password and access settings. If we have a way to have it tailored to use synology users I'm guessing it would be fine to drop it but otherwise I would be keen on keeping it unless there are drawbacks which I am not aware of.

So this was a manual change by the user and not by package installation. On Updates, the package should check, whether the `.htaccess` file does not match the one that is in the update package.
If it does not match we could either
- overwrite the file and warn the user, that manual modifications have to be reapplied
- keep the original file, create a copy of the new file and warn the user, that the file might have essential changes that must be manually applied after the update
- If it is posible to make a diff of the two files and we can ensure that there is no diff except additional lines, we could keep the original file and avoid any warnings

I vote for the first

smaarn:
> > @hgy59 that's in this file that you can configure password and access settings. If we have a way to have it tailored to use synology users I'm guessing it would be fine to drop it but otherwise I would be keen on keeping it unless there are drawbacks which I am not aware of.
> 
> 
> 
> So this was a manual change by the user and not by package installation. On Updates, the package should check, whether the `.htaccess` file does not match the one that is in the update package.
> 
> If it does not match we could either
> 
> - overwrite the file and warn the user, that manual modifications have to be reapplied
> 
> - keep the original file, create a copy of the new file and warn the user, that the file might have essential changes that must be manually applied after the update
> 
> - If it is posible to make a diff of the two files and we can ensure that there is no diff except additional lines, we could keep the original file and avoid any warnings
> 
> 
> 
> I vote for the first
> 
> 

I would vote for the third option (last time I checked this file hadn't been altered for some time) but then I ain't the maintainer and the minimum is to provide with a warning and providing with a backup.

mreid-tt:
@hgy59, I'm not clear on the challenge with regard to the `.htaccess` file. If the file is modified by the user we will carry those modifications forward on upgrade. Am I missing something here?

mreid-tt:
@hgy59 are we good to merge this?

spk/cops/src/conf/resource
```diff
@@ -1,57 +1,55 @@
...
-  "data-share": {
-    "shares": [{
-      "name": "{{wizard_calibre_share}}",
-      "permission": {
-        "rw": ["sc-cops"]
-      },
-      "once": true
-    }]
-  }
+	"data-share": {
```
smaarn, line 2: There doesn't seem to be any change on this file, right ? (Seems to be only formatting from what I see, just want to make sure)

spk/cops/src/wizard_templates/upgrade_uifile_fre.yml
```diff
@@ -0,0 +1,6 @@
+COPS_CONFIGURATION_UPGRADE_STEP_TITLE: "Mise à jour de la bibliothèque COPS"
+
+UPGRADE_CALIBRE_DIRECTORY_DESCRIPTION: "Le répertoire de la bibliothèque Calibre doit désormais résider dans un partage de données. Sur la base de la configuration précédente détectée, veuillez confirmer que le partage de données spécifié contient votre bibliothèque."
```
smaarn, line 3: Based on existing DSM translations that should be:

```suggestion
UPGRADE_CALIBRE_DIRECTORY_DESCRIPTION: "Le répertoire de la bibliothèque Calibre doit désormais résider dans un dossier partagé. Sur la base de la configuration précédente détectée, veuillez confirmer que le dossier partagé spécifié contient votre bibliothèque."
```

spk/cops/src/wizard_templates/upgrade_uifile_fre.yml
```diff
@@ -0,0 +1,6 @@
...
+UPGRADE_CALIBRE_DIRECTORY_LABEL: "Partager le nom"
```
smaarn, line 4:
```suggestion
UPGRADE_CALIBRE_DIRECTORY_LABEL: "Dossier partagé"
```

spk/cops/src/wizard_templates/upgrade_uifile_fre.yml
```diff
@@ -0,0 +1,6 @@
...
+UPGRADE_CALIBRE_DIRECTORY_VALIDATION_ERROR_TEXT: "Les sous-répertoires ne sont pas pris en charge."
+UPGRADE_NOTE_CALIBRE_DIRECTORY_DESCRIPTION: "REMARQUE: si votre bibliothèque Calibre ne se trouve pas dans le partage spécifié, un nouveau partage sera créé. Après la mise à niveau, déplacez manuellement la bibliothèque vers ce nouvel emplacement."
```
smaarn, line 6:
```suggestion
UPGRADE_NOTE_CALIBRE_DIRECTORY_DESCRIPTION: "REMARQUE: si votre bibliothèque Calibre ne se trouve pas dans le dossier partagé spécifié, un nouveau sera créé. Après la mise à niveau, déplacez manuellement la bibliothèque vers ce nouvel emplacement."
```

smaarn (commented):
I'm just not sure about this whole logic around the restore where it has two different scenarios depending on whether backup files exist or not. From what I gathered those files should always be present in an upgrade process, right ?

spk/cops/src/wizard_templates/upgrade_uifile_fre.yml
mreid-tt: Thanks for the translation assistance. These have been incorporated.

spk/cops/src/conf/resource
mreid-tt: Yes, this is formatting to be consistent with the other file.

mreid-tt:
> I'm just not sure about this whole logic around the restore where it has two different scenarios depending on whether backup files exist or not. From what I gathered those files should always be present in an upgrade process, right ?

The main challenge was with the data share name (see #5961). When the name is absent the data share doesn't get created and the backup file is not copied from the temp folder (and is lost). The new wizard checks if the share name is known and if not, prompts the user to confirm the best guess. The service setup also checks the scenario for those users who broke their installations with the previous release and attempts to repair by recreating the missing file.

smaarn:
> > I'm just not sure about this whole logic around the restore where it has two different scenarios depending on whether backup files exist or not. From what I gathered those files should always be present in an upgrade process, right ?
> 
> The main challenge was with the data share name (see #5961). When the name is absent the data share doesn't get created and the backup file is not copied from the temp folder (and is lost). The new wizard checks if the share name is known and if not, prompts the user to confirm the best guess. The service setup also checks the scenario for those users who broke their installations with the previous release and attempts to repair by recreating the missing file.

Would be worth documenting (issueing a message for example) that they had to be recreated (so that user doesn't think it got "fixed" when there could be other issues with his installation having lost its original settings).

LGTM otherwise (got my install messed up by previous upgrade but it wasn't much complicated to restore TBH)

mreid-tt:
> Would be worth documenting (issueing a message for example) that they had to be recreated (so that user doesn't think it got "fixed" when there could be other issues with his installation having lost its original settings).

There is a message written to the log. I don't think we need to do more than that as (a) there are limited options to set and (b) only a few users would have experienced this issue as the buggy version was active for only a few days.

> LGTM otherwise (got my install messed up by previous upgrade but it wasn't much complicated to restore TBH)

Thanks for the review. I'll commit tomorrow unless there are further comments.

PR #5956, opened 2023-12-11
Transmission: Update to 4.0.5

mreid-tt:
## Description

This pull request includes the following:

1. Update to Transmission v4.0.5 (bugfix-only release)
2. Migrate to new shared folder handling
3. Remove patch for empty blocklist files on 32-bit ARM (resolved upstream)

Fixes #5867

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #5963, opened 2023-12-19
Transmission: Resolve Port Configuration Issue

mreid-tt:
## Description

This pull request builds upon the changes introduced in #5956, specifically focusing on enhancing the new shared folder handling functionality. The previous implementation identified certain shortcomings, including:

1. The resource worker lacked the assignment of the protocol file, causing the package to fail in releasing port configurations for transmission. This issue, in turn, hindered the re-installation process.

Fixes # 

## Checklist

### Type of change

- [x] Bug fix


mreid-tt:
@hgy59, taking into account the developments in #5962, I conducted an analysis of the data share and port configuration based on feedback received on Discord. While the upgrade process for the data share did not encounter a failure, the install log displayed errors indicating that the `SHARE_NAME` referenced was not an existing share. To rectify this, an adjustment was necessary in the `installer-variables` to accommodate the changes introduced with the new shared folder handling. I am aware of your preference to avoid incorporating these changes in the wizard, but I have been unable to identify a method for updating the file before the variables are loaded. Please provide your insights if you have ideas on seamlessly transitioning this to the service setup script.

Regarding the port configuration, my investigation revealed that the previous package had a `port-config` definition, whereas the newer package did not. This disparity arises from the fact that the new package defines a resource file for the data share. When a resource file is defined, the default assignment of `port-config` is not automatically added and must be manually specified. This manual definition ensures the package correctly obtains a port configuration for transmission, allowing the reserved port in the firewall to be established and subsequently removed during uninstallation.

mreid-tt:
@hgy59, based on your comments in #5962, I realise that I was over-complicating things. Given that the spksrc automatically creates the `resource` file I don't need to worry about the wizard or the `conf` directory. I've removed all of this and just left a preinstall check to remove the installer variables file if it has an invalid entry for `SHARE_PATH` as I encountered when I upgraded. In most of my upgrade tests this file doesn't exist so I believe that removing it would be the safest action for an invalid entry.

PR #5649, opened 2023-02-27
redesign shared folder handling

hgy59:
## Description

The volume specified in the install wizard for shared folders is not used on DSM 7 (and on DSM 6, when using the resource worker with `USE_DATA_SHARE_WORKER = yes`)

By default the shared folders created by DSM are located in the same volume as the package is installed to.
It is possible to create shared folders on a different volume and use an existing share in the installation wizard.
This is currently not supported by our framework (at least not for DSM 6 when using the resource worker)

There is some code in the framework, that handles the volume chosen in the wizard, but only when the variable is named `wizard_volume`.
This is undocumented and bogus, and will be removed here. It is replaced by evaluation of the full path of the share by the system function synoshare on DSM 6 and by using the provided link under shares on DSM 7 (non root user is denied to call synoshare)

Related to #5481
Fixes #5041

## Remarks
This got more framework fixes than originally planned. 
   
Finally this PR updates the `demoservice` package only and introduces the Makefile variable `SERVICE_WIZARD_SHARENAME` to not affect existing packages that use shared folders the old way (i.e. `SERVICE_WIZARD_SHARE` and optionally `USE_DATA_SHARE_WORKER`).

While testing the new shared folders with demoservice, an issue with `SERVICE_COMMAND` on SRM and DSM5 popped up. Therefore a dedicated start script without parameters is used in demoservice now. `SERVICE_COMMAND` with parameters does not work together with `SVC_BACKGROUND=y` and `SVC_CREATE_PID=y` on platforms not supporting bash shell (i.e. ash).

Another issue newly discoverd is, that `load_variables_from_file` does not work to load installer variables. This is fixed by direct call of the function (to avoid a sub shell by use of `call_func`).
And `load_variables_from_file` had an issue on SRM and DSM5 (cut with parameter --delimiter is not supported there)

Another issue affects the installer variables persisted in `etc/installer-variables`. This file was removed on package updates without an upgrade wizard. Without an upgrade wizard that activly reloads the variables into the wizard before the upgrade, those variables are not defined anymore. The fix avoids the deletion of the installer-variables file in such cases.

Additionally this PR completly removes the (deprecated) support for `SERVICE_EXE` and `SERVICE_WIZARD_GROUP`.

And one more issue fixed. The DSM 7 uninstaller was not able to delete the files in the package `etc` folder. This was also a security issue, since the `installer-variables` file might contain sensitive information.

[...]

PR #5968, opened 2024-01-03
Update SABnzbd to 4.2.1

Safihre:
## Description

Update SABnzbd to 4.2.1.

@mreid-tt @hgy59 Do I need to change anything about the shared folder handling? It's a bit unclear to me.
I look at Transmission, but it seems that is more complicated than SABnzbd needs?

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update

### DIFF
--- a/cross/sabnzbd/Makefile
+++ b/cross/sabnzbd/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = SABnzbd
-PKG_VERS = 4.1.0
+PKG_VERS = 4.2.0
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS)-src.$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/sabnzbd/sabnzbd/releases/download/$(PKG_VERS)
--- a/cross/sabnzbd/Makefile
+++ b/cross/sabnzbd/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = SABnzbd
-PKG_VERS = 4.2.0
+PKG_VERS = 4.2.1
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS)-src.$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/sabnzbd/sabnzbd/releases/download/$(PKG_VERS)
--- a/cross/sabnzbd/digests
+++ b/cross/sabnzbd/digests
@@ -1,3 +1,3 @@
-SABnzbd-4.1.0-src.tar.gz SHA1 7ed45efca18ee589ad24a6fa43ae22a47661aa6e
-SABnzbd-4.1.0-src.tar.gz SHA256 24cdc711a9a9425b65b53dd5c084f78cc0f6d978c5c712481b7031751d569588
-SABnzbd-4.1.0-src.tar.gz MD5 f1033e949168250aef5500ab919938e6
+SABnzbd-4.2.0-src.tar.gz SHA1 f088e6d771ef53163d7bca6dc0cff8be70abd1e9
+SABnzbd-4.2.0-src.tar.gz SHA256 ed21b2bf1fb8f2e12da61bd2f31c24b03bbe44869268047fa695ccbdd201bf46
+SABnzbd-4.2.0-src.tar.gz MD5 7b13a952daaaa2b41b8f1b4a6dc099ea
--- a/cross/sabnzbd/digests
+++ b/cross/sabnzbd/digests
@@ -1,3 +1,3 @@
-SABnzbd-4.2.0-src.tar.gz SHA1 f088e6d771ef53163d7bca6dc0cff8be70abd1e9
-SABnzbd-4.2.0-src.tar.gz SHA256 ed21b2bf1fb8f2e12da61bd2f31c24b03bbe44869268047fa695ccbdd201bf46
-SABnzbd-4.2.0-src.tar.gz MD5 7b13a952daaaa2b41b8f1b4a6dc099ea
+SABnzbd-4.2.1-src.tar.gz SHA1 642a1d8fbeea42b7f7273675ab5e4f7f1a0d921e
+SABnzbd-4.2.1-src.tar.gz SHA256 0d59eefd883774fcb9e3f58a073f4e53467341b34026db08dce506625052ab49
+SABnzbd-4.2.1-src.tar.gz MD5 b5489f668a5e528d4176930e9bead006
--- a/spk/sabnzbd/Makefile
+++ b/spk/sabnzbd/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = sabnzbd
-SPK_VERS = 4.1.0
-SPK_REV = 66
+SPK_VERS = 4.2.0
+SPK_REV = 67
 SPK_ICON = src/sabnzbd.png
 
 PYTHON_PACKAGE = python311
@@ -15,7 +15,7 @@ DESCRIPTION_FRE = SABnzbd rend Usenet aussi simple et automatisé que possible. 
 DESCRIPTION_SPN = SABnzbd hace que Usenet sea lo más simple posible, automatizando todo lo que se puede. Todo lo que tienes que hacer es agregar un archivo .nzb. SABnzbd empieza desde ahí. Tus archivos serán automáticamente descargados, verificados, reparados, descomprimidos y archivados.
 DISPLAY_NAME = SABnzbd
 STARTABLE = yes
-CHANGELOG = "1. Update SABnzbd to 4.1.0.<br/>2. Use par2cmdline-turbo instead of par2cmdline.<br/>2. Update UnRar to 6.23 (security fixes)."
+CHANGELOG = "Update SABnzbd to 4.2.0."
 
 HOMEPAGE  = https://sabnzbd.org
 LICENSE   = GPL
--- a/spk/sabnzbd/Makefile
+++ b/spk/sabnzbd/Makefile
@@ -23,7 +23,7 @@ LICENSE   = GPL
 WIZARDS_DIR = src/wizard/
 
 SERVICE_USER = auto
-SERVICE_WIZARD_SHARE = wizard_download_dir
+SERVICE_WIZARD_SHARENAME = wizard_shared_folder_name
 SERVICE_SETUP = src/service-setup.sh
 SERVICE_PORT = 8080
 SERVICE_PORT_TITLE = $(DISPLAY_NAME)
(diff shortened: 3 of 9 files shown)


mreid-tt:
> @mreid-tt @hgy59 Do I need to change anything about the shared folder handling? It's a bit unclear to me. I look at Transmission, but it seems that is more complicated than SABnzbd needs?

If you are up for it, the approach would be similar to Transmission as follows:

1. Move away from `SERVICE_WIZARD_SHARE` in your Makefile and use `SERVICE_WIZARD_SHARENAME`
2. Simplify your install wizard to just use the share name since the volume selector doesn't really do anything. You should only have one key like `wizard_shared_folder_name` for the shared folder in your wizard
3. In your service setup, you can just use `${SHARE_PATH}` to assign to your `shared_folder` variable since the concatenation of volume and folder won't be required anymore

That's pretty much it to use the new shared folder handling.

hgy59:
@mreid-tt since this packages did not use shared resources for DSM 6 before this update (`USE_DATA_SHARE_WORKER` not defined), it needs some extra testing (of the installer-variables) for updates under DSM 6.

mreid-tt:
> @mreid-tt since this packages did not use shared resources for DSM 6 before this update (`USE_DATA_SHARE_WORKER` not defined), it needs some extra testing (of the installer-variables) for updates under DSM 6.

Hmm, okay. To refresh my memory, is this like what we did for COPS and read the config file on upgrade to have the user confirm the name of the shared folder?

mreid-tt:
@Safihre, I've mocked up a `upgrade_uifile.sh` to replace your upgrade wizard file. This should address the concern raised by @hgy59 for DSM6:

```
#!/bin/bash

page_append ()
{
    if [ -z "$1" ]; then
        echo "$2"
    elif [ -z "$2" ]; then
        echo "$1"
    else
        echo "$1,$2"
    fi
}

CFG_FILE="/var/packages/${SYNOPKG_PKGNAME}/var/config.ini"
CONFIGURED_SHARE_PATH=$(awk -F' = ' '/^download_dir/{split($2, path, "/"); print "/" path[2] "/" path[3]}' "${CFG_FILE}")
CONFIGURED_SHARE_NAME=$(echo "${CONFIGURED_SHARE_PATH}" | awk -F'/' '{print $NF}')

VAR_FILE="/var/packages/${SYNOPKG_PKGNAME}/etc/installer-variables"
PACKAGE_SHARE_PATH=$(awk -F'=' '/^SHARE_PATH=/{print $2}' "${VAR_FILE}" 2>/dev/null || echo "")
if [ -z "$PACKAGE_SHARE_PATH" ] && [ -d "/var/packages/${SYNOPKG_PKGNAME}/shares" ]; then
    PACKAGE_SHARE_PATH=$(realpath "/var/packages/${SYNOPKG_PKGNAME}/shares/${CONFIGURED_SHARE_NAME}" 2>/dev/null || echo "")
fi

WARNING_TEXT="IMPORTANT: If your download folder is not currently located in the specified share, a new share will be created. After the upgrade, you may need to manually update your configuration to reflect this new location."

# Check for consistent data share path
if [ -n "$CONFIGURED_SHARE_PATH" ] && [ -n "$PACKAGE_SHARE_PATH" ] && [ "$CONFIGURED_SHARE_PATH" = "$PACKAGE_SHARE_PATH" ]; then
    WARNING_TEXT=""
fi

PAGE_SHARE_UPGRADE=$(/bin/cat<<EOF
{
    "step_title": "Shared Folder Upgrade",
    "invalid_next_disabled_v2": true,
    "items": [{
        "type": "textfield",
        "desc": "The download folder for this package must now be located within a data share. As per the existing configuration, the identified data share for your downloads is:",
        "subitems": [{
            "key": "wizard_shared_folder_name",
            "desc": "Shared Folder",
            "defaultValue": "${CONFIGURED_SHARE_NAME}",
            "validator": {
                "allowBlank": false,
                "regex": {
                    "expr": "/^[\\\w _-]+$/",
                    "errorText": "Subdirectories are not supported."
                }
            }
        }]
    },{
        "desc": "${WARNING_TEXT}"
    }]
}
EOF
)

PAGE_DSM_PERMISSIONS=$(/bin/cat<<EOF
{
    "step_title": "DSM Permissions",
    "items": [{
        "desc": "Please read <a target=\"_blank\" href=\"https://github.com/SynoCommunity/spksrc/wiki/Permission-Management\">Permission Management</a> for details."
    }]
}
EOF
)

main () {
    local upgrade_page=""
    upgrade_page=$(page_append "$upgrade_page" "$PAGE_SHARE_UPGRADE")
    upgrade_page=$(page_append "$upgrade_page" "$PAGE_DSM_PERMISSIONS")
    echo "[$upgrade_page]" > "${SYNOPKG_TEMP_LOGFILE}"
}

main "$@"

```

ivd-git:
Since this has not been merged yet. Would it be possible to bump to 4.2.1 directly?

https://github.com/sabnzbd/sabnzbd/releases/tag/4.2.1

Safihre:
> Since this has not been merged yet. Would it be possible to bump to 4.2.1 directly?
> 
> https://github.com/sabnzbd/sabnzbd/releases/tag/4.2.1

Will do!

Safihre:
> @Safihre, I've mocked up an `upgrade_uifile.sh` to replace your upgrade wizard file and added a commit ([48b315a](https://github.com/SynoCommunity/spksrc/commit/48b315a9a0925bc911008bab77ddf651f559cb88)) that should address the concern raised by @hgy59 for DSM6.

I saw.. But seeing this makes me wonder if we should even do this at all? What's wrong with keeping the "old" way? 
Didn't have any complaints before.

mreid-tt:
> I saw.. But seeing this makes me wonder if we should even do this at all? What's wrong with keeping the "old" way?

The motivations for this I believe were documented under #5649. Were there specific concerns you had?

hgy59:
> > I saw.. But seeing this makes me wonder if we should even do this at all? What's wrong with keeping the "old" way?
> 
> The motivations for this I believe were documented under #5649. Were there specific concerns you had?

The main issue was, that the volume selected in the wizard is not used by DSM 7+. DSM 7 only supports data share resources to use shared folders and those allow the name of the share only.

The redesign of shared folders makes DSM 6 shared folder handling compatible with DSM 7.
As a side effect, it is now possible to use a different volume for shared folders, when the shared folder is created before the installation of a package.

mreid-tt:
> But do we need the extra steps during upgrade? Since it's already working, why do we need any changes there? Just don't touch the user's config and it just keeps working?

The upgrade wizard file may seem bulky but it's written to minimise the impact to users. The logic is simply:

1. Check that a share name exists (either via variable [DSM6] or folder [DSM7]). If present do nothing
2. Check that a share path exists (either via variable or folder as above)
3. Extract the share path and name based on the package config file (config.ini)
4. Present a screen asking the user to confirm that the share name is correct for the upgrade
5. If the package config and the share path do not match, add an additional warning

These steps are essentially only seen once by users since the process establishes a share name. Once a share name exists the confirmation screen will not be shown for any future updates.

Safihre:
But why is it needed to show anything at all to users that are upgrading? They already have a working share and path.

mreid-tt:
> But why is it needed to show anything at all to users that are upgrading? They already have a working share and path.

I appreciate you bringing this to my attention. One crucial aspect that I failed to clarify earlier involves two primary considerations:

1. The `wizard_shared_folder_name` resource must be populated by the wizard to ensure accurate provisioning of the data share under DSM 6. Failure to do so could result in an improper setup.
2. While highly improbable, there exists a slim chance of misinterpretation of the configuration, making it imperative for the user to verify the extracted value.

However, there are potential counter arguments for each:

1. One could explore setting the `wizard_shared_folder_name` in the DSM Permissions screen as a hidden variable, mitigating the need for explicit user intervention during the wizard process.
2. The second point, albeit unlikely, becomes less concerning if the package configuration and share path maintain consistency.

My intention was to exercise caution by providing users with visibility into the value set during the upgrade process. However, if you differ in your risk assessment, I am open to adjusting the wizard to display a confirmation screen only when discrepancies arise between the package config and the share path.

mreid-tt:
@Safihre I've pushed a commit to take a less conservative approach to the wizard as proposed above.

Safihre:
Interesting! Will test myself tonight.
Although I only have 1 volume in my own NAS.

Safihre:
Thank you @mreid-tt, it seems to work for me! I didn't have to confirm my share, and it worked fine after the upgrade.
Should this then also work for DSM 6 or needs more testing?

mreid-tt:
> Should this then also work for DSM 6 or needs more testing?

I've done some tests on my DSM 6 virtual environment as an upgrade from the previous version and an in-place upgrade replacement and it seems to work well. @hgy59, should we be good to go here?

Safihre:
How do we trigger the upgrade step to be shown? I should modify the sabnzbd Config?

mreid-tt:
> How do we trigger the upgrade step to be shown? I should modify the sabnzbd Config?

Yes, I did this experiment under my DSM 7 virtual environment.

1. Downgrade to the published version of SABnzbd v4.1.0-66
2. Manually add read/write permissions to a new folder (e.g. /volume1/git/)
3. In SABnzbd change the Temporary Download Folder (e.g. to /volume1/git/incomplete)
4. Perform manual install of new package to show wizard step

Now running this experiment a second time over the upgraded package will not work because the `/var/packages/sabnzbd/etc/installer-variables` file was updated with a `SHARE_NAME` variable. With this present the first condition is met even though the second condition of a mismatched share path is true.

Safihre:
So, seems good to go then? :)

Safihre:
Thank you!

mreid-tt:
I've downloaded the package and manually deployed to my DS916+ and everything seems to be working well.

gohakn:
Hello just updated to 4.2.1 at my Synology DS1817+ and it`s working well except that my script for forcing Downloads smaller 50MB is no longer working.

Here`s the code which was working since Version 4.0

`#!/usr/bin/python3
import sys

try:
    # Parse the 18 input variables for SABnzbd version >= 4.0.0
    (scriptname, nzbname, postprocflags, category, script, prio, downloadsize, grouplist, showname, season, episodenumber, episodename, is_proper, resolution, decade, year, month, day, job_type) = sys.argv
except ValueError:
    # ...or 11 variables for earlier versions
    (scriptname, nzbname, postprocflags, category, script, prio, downloadsize, grouplist, showname, season, episodenumber, episodename) = sys.argv
except Exception:
    sys.exit(1)  # a non-zero exit status causes SABnzbd to ignore the output of this script

prio = -100  # Default
if int(downloadsize) < 50111222:
    prio = 2

print("1")  # Accept the job
print()
print()
print()
print()
print(prio)
print()

 
sys.exit(0)`

Is there any change in 4.2.1??

Safihre:
Yes. Read the Release Notes / Changelog on https://sabnzbd.org/downloads

PR #5978, opened 2024-01-11
[New Package] Kavita

### LINKED ISSUE #5967: [Package Request] Kavita Book Reader
### Software Name

Kavita

### Brief Description

its a cleaner more modern version of Calibre which in my opinion is out dated. It makes it look and act more like plex. 

### Website

https://www.kavitareader.com/

### Documentation

https://wiki.kavitareader.com/en

https://wiki.kavitareader.com/en/install/linux-install

### Source code

https://github.com/Kareadita/Kavita

### OpenSource License

GPL-3.0

### Other License

_No response_

mreid-tt:
## Description

Kavita is a fast, feature rich, cross platform reading server. Built with a focus for being a full solution for all your reading needs.

Closes #5967

## Checklist

### Type of change

- [x] New Package


mreid-tt:
@hgy59, I've created this new package based on #5967. The source is built on dotnet so I leveraged the templates from the *arr family of packages. There were a few exceptions:

1. I could not use the standard `SERVICE_COMMAND` because each time the package launched I got the following error:
```
Error reading app settings: Could not find file '/config/appsettings.json'.
```
 - Trying to resolve it by adding environment variables for `$HOME` and others did not help. This app also does not seem to support command line arguments based on its documentation
 - To work around the problem I made a manual `service_prestart` based on the wiki page and added an explicit change to the working directory and this resolved the issue
2. The package does not use the latest v0.7.12 from source because under DSM 6, I get the following errors:
```
System.TypeInitializationException: The type initializer for 'Microsoft.Data.Sqlite.SqliteConnection' threw an exception.
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.DllNotFoundException: Unable to load shared library 'e_sqlite3' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable: 
/volume1/@appstore/kavita/share/e_sqlite3.so: cannot open shared object file: No such file or directory
/lib/libc.so.6: version `GLIBC_2.28' not found (required by /volume1/@appstore/kavita/share/libe_sqlite3.so)
libe_sqlite3.so: cannot open shared object file: No such file or directory
/volume1/@appstore/kavita/share/e_sqlite3: cannot open shared object file: No such file or directory
/volume1/@appstore/kavita/share/libe_sqlite3: cannot open shared object file: No such file or directory
```
 - Downgrading to the v0.7.11.2 release resolved the issue

mreid-tt:
hey @hgy59, I was wondering if you had any thoughts on the error I posted:

```
/lib/libc.so.6: version `GLIBC_2.28' not found (required by /volume1/@appstore/kavita/share/libe_sqlite3.so)
```

As this happens only with DSM 6, I would imagine it has something to do with this dependency:

https://github.com/SynoCommunity/spksrc/blob/61b509edfbae1ae182da99aa56d9b64368e70e3c/spk/kavita/Makefile#L35-L39

Or are we missing a dependency?

hgy59:
@mreid-tt for this package prebuilt binaries are used. Those binaries are not compiled with the DSM toolchains.

So the binaries are built with a dependency of gligc that is newer than what is available under DSM 6.

The only solution would be to build the package from source with the toolchains and native/dotnet8, but it could be a hard requirement of the sqlite library in the package and fail to build.
I guess you have to drop support for DSM < 7.

mreid-tt:
> @mreid-tt for this package prebuilt binaries are used. Those binaries are not compiled with the DSM toolchains.
> 
> So the binaries are built with a dependency of gligc that is newer than what is available under DSM 6.
> 
> The only solution would be to build the package from source with the toolchains and native/dotnet8, but it could be a hard requirement of the sqlite library in the package and fail to build. I guess you have to drop support for DSM < 7.

Thanks for the analysis. The package has been updated to remove DSM 6 support. Your final review and approval would be appreciated.

spk/kavita/src/service-setup.sh
```diff
@@ -0,0 +1,32 @@
+
+# Kavita service setup
+KAVITA="${SYNOPKG_PKGDEST}/share/Kavita"
+
+# Kavita uses custom Config and PID directories
+HOME_DIR="${SYNOPKG_PKGDEST}/share"
+CONFIG_DIR="${SYNOPKG_PKGVAR}/config"
+PID_FILE="${SYNOPKG_PKGVAR}/kavita.pid"
```
hgy59, line 8: this is the default and must not be defined here
```suggestion
```

spk/kavita/src/service-setup.sh
```diff
@@ -0,0 +1,32 @@
...
+SVC_WAIT_TIMEOUT=90
+
+service_prestart ()
+{
+    # Replace generic service startup, fork process in background
+    echo "Starting Kavita at ${HOME_DIR}" >> ${LOG_FILE}
+    cd ${HOME_DIR} || exit 1;
+    ${KAVITA} >> ${LOG_FILE} 2>&1 &
+    echo "$!" > "${PID_FILE}"
+}
```
hgy59, line 20: IMO this can also be achieved with
```
SVC_CWD=${HOME_DIR}
SVC_WRITE_PID=y
SERVICE_COMMAND="${KAVITA} >> ${LOG_FILE}"
```

this requires also `SVC_BACKGROUND=y` that is already defined above.

spk/kavita/src/service-setup.sh
mreid-tt: Thanks for this feedback.

spk/kavita/src/service-setup.sh
mreid-tt: I'll try that and see if it works.

spk/kavita/src/appsettings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "TokenKey": "super secret unguessable key that is longer because we require it",
```
hgy59, line 2: please generate a unique TokenKey at installation time.

If the user must know this token, please show it once in the installation wizard.

mreid-tt: When the app is first run it generates that key. This is the default file that's present in the source repo. I just used it to change the port number from 5000 to 5500 to avoid a conflict with DSM.

spk/kavita/Makefile
```diff
@@ -0,0 +1,43 @@
...
+
+POST_STRIP_TARGET = kavita_extra_install
+
+include ../../mk/spksrc.spk.mk
+
+.PHONY: kavita_extra_install
+kavita_extra_install:
+	@$(MSG) "Install appsettings.json."
+	@install -m 755 -d $(STAGING_DIR)/var/config
+	@install -m 644 src/appsettings.json $(STAGING_DIR)/var/config/appsettings.json
```
hgy59, line 43: why not installing appsettings.json to the var folder?
this would avoid the need of awkward link

BTW: isn't there a parameter to tell kavita where to find the appsettings?

mreid-tt: I wanted to ensure that the config folder is stored away from the binaries since it is the only folder that must be retained on upgrade. I would have preferred to use a launch parameter but I haven't been able to find any defined in the source documentation for Kavita.

spk/kavita/src/service-setup.sh
mreid-tt: Should we work over time to remove these in the *arr family of apps? Many of them have the `PID_FILE` defined similarly.

spk/kavita/Makefile
mreid-tt: Actually I thought I was placing the config folder in the var folder. Based on other examples I see I thought this was the way to do it.

spk/kavita/src/service-setup.sh
mreid-tt: It works! The package can now look a lot more like the others. Thanks for that.

mreid-tt: I'm guessing that it can be simplified even further by removing the `>> ${LOG_FILE}` since this would be a default as well?

mreid-tt: Yup, logging still works without this.

spk/kavita/Makefile
mreid-tt: @hgy59, I reached out to the Kavita developers and they confirmed that they don't have configurable parameters for the location of the config folder. One said:

> no, Kavita doesn't allow moving the config folder. Having it in the install folder is ideal to prevent users from editing anything and ensuring it's always there

When asked if this could be a future feature request (given that hard/soft links can be messy for updates), they replied:

> Unfortunately that's not something I'd be open to.

mreid-tt:
@hgy59, any final thoughts on this or are we good to merge?

hgy59 (commented):
I am missing some documentation on how to configure the folders in kavita.

We should document that the _internal system user_ sc-kavita must have read access to the folders (or does it even need write access too?)

mreid-tt:
> I am missing some documentation on how to configure the folders in kavita.
> 
> We should document that the _internal system user_ sc-kavita must have read access to the folders (or does it even need write access too?)

I've amended the wizard. The Kavita [First Time Setup](https://wiki.kavitareader.com/en/guides/first-time-setup) has no guidance on permissions so I'd play it safe with assumed full permissions.

hgy59 (approved):
LGTM

borisrorsvort:
@mreid-tt would you have time to upgrade the package to the latest version?

mreid-tt:
> @mreid-tt would you have time to upgrade the package to the latest version?

hi @borisrorsvort, yes, I'll try to make some time over the weekend to publish an update.

PR #5958, opened 2023-12-13
[Framework] Update arch exclusions for dotnet

mreid-tt:
## Description

This pull request enhances the clarity of dotnet 6.x exclusions by revising existing definitions, streamlines the documentation by consolidating notes and history, and introduces specific exclusions for ARMv7 under DSM 6.2.

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Includes small framework changes

PR #5980, opened 2024-01-13
Sonarr v4: Update to v4.0.0.748

### LINKED ISSUE #5945: sonarr v4 version bump
### Is this a new Bug?

- [X] I checked that the bug hasn't been reported before

### Package Name

Sonarr (v4 beta)

### Package Version

n/a

### Device Model

n/a

### Device Architecture

x86_64

### Firmware Version

n/a

### What happened?

https://forums.sonarr.tv/t/error-parsing-column-when-i-try-to-open-sonarr/32959

due to database migrations the current package is too old and needs a bump

### Reproduction steps

n/a

### Install Log

```log
n/a
```

### Service Log

```log
n/a
```

### Other Logs

```log
n/a
```

mreid-tt:
## Description

1. Update Sonarr to v4.0.0.748 (on main branch)
2. Update SQLite3 dependency
3. Increase service start/stop timeout to 120 seconds

Fixes #5945

## Checklist

### Type of change

- [x] Package update


hgy59:
@mreid-tt I propose add cross/sqlite for all DSM versions. This will ensure we know the sqlite version used by sonarr.

And you need to add the target/bin folder of the package to the path. Otherwise the package will still use sqlite provided by DSM.

mreid-tt:
> @mreid-tt I propose add cross/sqlite for all DSM versions. This will ensure we know the sqlite version used by sonarr.

Sure, I can make that change.

> And you need to add the target/bin folder of the package to the path. Otherwise the package will still use sqlite provided by DSM.

I just did testing under DSM 6 to resolve the issue recently discussed in #5945. This issue seems to be resolved with my recent commit alone. I will defer to your expertise however. Where would this path be? Is it in the arguments of the `SERVICE_COMMAND` or somewhere else?

mreid-tt:
@hgy59, I've noted an issue in DSM 7 for clean installs now which I didn't notice before. Since version 20231129-3 (Sonarr v4.0.0.733) a clean install does not seem to actually complete. With version 20230426-2 (Sonarr v4.0.0.460) the installation ran and the installer just completed even though the service log was still performing migrations. This is also the case in DSM 6 for all versions. However from Sonarr v4.0.0.733 (likely from about v4.0.0.610) and beyond the installer in DSM 7 seems to wait until the app is fully setup. If it does not complete, the process seems to time out and gets terminated. The service log in DSM 7 looks like this:

```
[Info] NzbDrone.Core.Datastore.Migration.Framework.NzbDroneSQLiteProcessor: DROP TABLE "LanguageProfiles" 
[Info] FluentMigrator.Runner.MigrationRunner: => 0.0006998s 
[Info] NzbDrone.Core.Datastore.Migration.Framework.NzbDroneSQLiteProcessor: INSERT INTO "VersionInfo" ("Version", "AppliedOn", "Description") VALUES (175, '2024-01-13T19:34:43', 'language_profiles_to_custom_formats') 
[Info] NzbDrone.Core.Datastore.Migration.Framework.NzbDroneSQLiteProcessor: Committing Transaction 
sonarr is not running
```

This is compared to DSM 6 where it just continues until completion. Any thoughts on how we can mitigate this changed behaviour?

mreid-tt:
@hgy59 after speaking a bit more to the sonarr folks, it seems that this was a change around v4.0.0.708 to resolve a DB lock issue (https://github.com/Sonarr/Sonarr/pull/6120). This results in the initial migration having a `BusyTimeout = 100` setting which causes DSM 7 to wait for that period. Given that this is above our default 90 seconds for this package, it would result in an incomplete migration on new installations.

mreid-tt:
@hgy59, in the publishing of this package I noted once again the presence of a ARMv7 package for DSM 6.2. I know we supposedly addressed this in #5958 so I can't understand why a package was still published. The code related to this is:

https://github.com/SynoCommunity/spksrc/blob/0c40a5455fc2327d55f7b8c1c378af8cf26f77bf/mk/spksrc.pre-check.mk#L53-L60

I can't determine if there is a bug in there but perhaps you can spot something I can't.

I also noted you proposed in https://github.com/SynoCommunity/spksrc/pull/5958#issuecomment-1854758305 that we disable `comcerto2k` since we didn't have hard test data to show it was supposed. Apologies that I didn't respond at the time. I don't disagree with this concept and have disabled the generated package for this build as well.

PR #6028, opened 2024-03-04
fix generic installers

hgy59:
## Description

This fixes the generic installers that are broken with #5927 for all SRM and DSM < 6.0 packages.

### Type of change

- [x] Includes framework fixes

### DIFF
--- a/mk/spksrc.service.installer.functions
+++ b/mk/spksrc.service.installer.functions
@@ -72,7 +72,8 @@ install_python_virtualenv ()
     python3 --version
 
     # Print out default pip version
-    python3 -m pip --version
+    PIP_VERSION=$(python3 -m pip --version | awk '{print $2}')
+    echo "Default pip version ${PIP_VERSION}"
 
     # Create a Python virtualenv
     python3 -m venv --system-site-packages ${SYNOPKG_PKGDEST}/env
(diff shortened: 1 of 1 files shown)


hgy59:
@th0ma7, @mreid-tt as commented in #5927

This is a fast fix that prints pip version like:
```
pip 23.1.2 from /volume1/@appstore/python310/lib/python3.10/site-packages/pip (python 3.10)
```

The original attempt was to put the pip version into a variable for further processing.

IMO there is no need to log the pip version at all.

hgy59:
> Am I correct in thinking that we should not need to re-publish anything as a result of this fix?

No, that is not correct.
The installer fails for all packages, unrelated to python, since the functions file is included in the installer script (service-setup).

The error on DSM 5.2 (in /var/log/synopkg.log) was:
```
/volume1/@tmp/pkginstall/scripts/preupgrade: /volume1/@tmp/pkginstall/scripts/functions: line 75: syntax error: unexpected redirection
```

It fails when loading the functions file, not when executing a function.


So the answer is yes, we need to recreate and publish affected DSM 5.x and SRM packages.

hgy59:
> I found zsh, zsh-static, go2rtc and ruby.

I can recreate those packages for SRM and DSM 5.x since I have published those, I only need a `make spkclean` plus `make` and must not rebuild the whole packages.

First I will delete the affected packages in the repository.

PR #6037, opened 2024-03-14
Fix build disk space - System.IO.IOException: No space left on device

th0ma7:
## Description

Fix build disk space - System.IO.IOException: No space left on device
- Create a new cleanup call `make clean-source` which only removes the `work-<package>*` source directories
- Call both `make clean-source` and `make spkclean` after builds through github-action

Fixes # 

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/.github/actions/build.sh
+++ b/.github/actions/build.sh
@@ -72,6 +72,8 @@ do
         else
             echo "$ make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package}" >>build.log
             make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package} |& tee >(tail -15 >>build.log)
+            make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package} clean-source |& tee >(tail -15 >>build.log)
+            make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package} spkclean |& tee >(tail -15 >>build.log)
         fi
     else
         if [ "${GH_ARCH}" = "noarch" ]; then
--- a/mk/spksrc.spk.mk
+++ b/mk/spksrc.spk.mk
@@ -496,6 +496,22 @@ endif
 clean:
 	rm -fr work work-* build-*.log publish-*.log
 
+# Remove work-*/<pkgname>* directories while keeping
+# work-*/.<pkgname>*|<pkgname>.plist status files
+# This is in order to resolve: 
+#    System.IO.IOException: No space left on device
+# when building online thru github-action
+clean-source:
+	@make --no-print-directory dependency-flat | sort -u | grep cross/ | while read depend ; do \
+	   makefile="../../$${depend}/Makefile" ; \
+	   pkgstr=$$(grep ^PKG_NAME $${makefile}) ; \
+	   pkgname=$$(echo $${pkgstr#*=} | xargs) ; \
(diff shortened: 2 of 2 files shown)


th0ma7:
@SynoCommunity/developers to all developers, let me know if this creates build issue so I revert and fix as needed!

hgy59:
IMHO clean-source target should delete all response files (.*_done) not only the source folders and copy/depend response files.

After `make spkclean` you can build a package without building all dependencies again. (you can manually delete source folders and response files to rebuild specific dependencies, I often use this while creating or updating packages),
After `make clean-source` a `make` must rebuild the packages with all dependencies. For this, `clean-source` must delete all .*_done files and spkclean is obsolete in this situation.

.github/actions/build.sh
```diff
@@ -72,6 +72,8 @@ do
         else
             echo "$ make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package}" >>build.log
             make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package} |& tee >(tail -15 >>build.log)
+            make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package} clean-source |& tee >(tail -15 >>build.log)
+            make arch-${GH_ARCH%%-*}-${GH_ARCH##*-} -C ./spk/${package} spkclean |& tee >(tail -15 >>build.log)
```
hgy59, line 76: why not below, when cleaning is done for all packages except the packages to keep?

```
    if [ "$(echo ${PACKAGES_TO_KEEP} | grep -ow ${package})" = "" ]; then
        # free disk space (but not for packages to keep)
        make -C ./spk/${package} clean
    else
        # delete source folders only for packages to keep
        make -C ./spk/${package} source-clean
    fi
```

th0ma7:
> @th0ma7 what is the size of ffmpeg4/5/6 and python310/311 source folders?
> 
> Is this really the root cause for disk full errors?
> 
> _sorry, I am offline for some days an cannot look myself_

I'll check over the weekend but yesterday's pr with clean-source solved it and now all flags are green :)

th0ma7:
> IMHO clean-source target should delete all response files (.*_done) not only the source folders and copy/depend response files.

To me it serves a different purppose here as it is targeted towards limiting disk space specifically.  Also, by cleaning only the source directories you can still play with `spkclean` and generate a functional package, not that it matters that much from a github perspective.  Also, packages that depends on them won't be affected as the `work-*/install` directory isn't affected thus includes and libraries are still available to be refered to.

But, maybie this is what it should be: `make clean-github-action` or similar, targetted to this in particular?

> what is the size of ffmpeg4/5/6 and python310/311 source folders?

To give you an idea:
```
spksrc@spksrc:~/all-supported-fix/spksrc/spk/ffmpeg6$ du -sh work-x64-7.1/
5.9G	work-x64-7.1/
spksrc@spksrc:~/all-supported-fix/spksrc/spk/python311$ du -sh work-x64-7.1/
1.9G	work-x64-7.1/
```

Long-story short: it demonstrated that it does fix the issue.  Now that the demonstration is over, we can probably fine-tune it futher.

.github/actions/build.sh
th0ma7: only seeing your comment now, started a new PR to change this over here as it makes much more sense indeed.  Ref. https://github.com/SynoCommunity/spksrc/pull/6038

PR #6040, opened 2024-03-18
Python 3.11.8 update

th0ma7:
## Description

Python 3.11.8 update

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6058, opened 2024-04-01
Misc updates in reference to #6040

th0ma7:
## Description

Misc updates in reference to #6040 to minimize rebuild impacts.

IMPORTANT: This includes an update to `xz-utils` using patched Debian security fixed version.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


cross/xz/Makefile
th0ma7, line 1: @hgy59 and @SynoCommunity/developers , this update includes the security "fixed" version of `xz-utils` with backdoor issue being solved.

More info: https://arstechnica.com/security/2024/04/what-we-know-about-the-xz-utils-backdoor-that-almost-infected-the-world/

hgy59: Don't panic, Only versions 5.6.0 and 5.6.1 are affected, but not v5.2.6.

th0ma7: Indeed, that's why I wanted to point-out that this 5.6.1 version actually is the Debian "fixed" version of it.  It looks like they downgraded it to 5.4.5 and only applied meaninful patches from 5.6.1 over it.  Also, this provide an alternate source as official one is currently down.

PR #6061, opened 2024-04-03
Update SABnzbd to 4.2.3

Safihre:
## Description

Update SABnzbd to 4.2.3.

Simple update. If no remarks I'll merge and publish end of the week.

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6002, opened 2024-02-01
Add generic spksrc.supported.mk

th0ma7:
### TL;DR;
Rework, optimize and simplify generic `mk` calls involving a major refactory but most importantly major clean-up of the code in hope to simplify future maintenance:
- [x] remove tons of duplicated code by using `spksrc.common-rules.mk`.  This invovles `*clean`, `dependency-*`, `changelog` and a new `rustup` such as in getting the rustc info using `make rustup show` for instance.
- [x] allow using `make all-supported` and `make all-latest` from anywhere applicable (i.e. `cross`, `spk`  and `diyspk`)
- [x] add a framework self-testing facility (to be incorporated into github-action later to trigger on any changes to `mk/*`)
- [x] allow `DEPENDS` from `toolchain`

### Self-test facility calls (as of March 6th, more to be included):
- `make testall`
- `make test clean`
- `make test depend`
- `make test digests`
- `make test download`
- `make test toolchain`
- `make test test-dependency-*`

TODO:
- `make test rust`
- `make test cmake`
- `make test gnu-configure`
- `make test meson`
- `make test native`
- ...

Superseeds #5998

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

### TODO
- [x] `cc`
- [x] `cmake`
- [x] `dotnet`
- [x] `go`
- [x] `install-resources`
- [x] `meson`
- [x] `rust`
- [ ] `spk`

#### Toolchain -> COMPLETE
- [x] `~/spksrc$ make -C toolchain/syno-x64-7.1`
- [x] `~/spksrc$ make -C toolchain/syno-x64-7.1 clean`
- [x] `~/spksrc/toolchain/syno-x64-7.1$ make`
- [x] `~/spksrc/toolchain/syno-x64-7.1$ make clean`
- [x] `~/spksrc/toolchain/syno-x64-7.1$ make changelog`
```
[...]

PR #6065, opened 2024-04-06
spksrc.publish.mk: add checks for empty variables

hgy59:
## Description

This is another attempt to fix the publishing of packages by github action (broken by #6002)
- ~make setup-synocommunity triggers "make setup" and this fails when mk/spksrc.test-rules.mk is included~ 
   fixed in #6066
- add checks for empty PUBLISH_URL and PUBLISH_API_KEY in spksrc.publish.mk
- print PUBLISH_URL variable when publishing

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Includes small framework changes

PR #6066, opened 2024-04-06
common.mk: Fix BASEDIR to manage github-action workspace directory

th0ma7:
## Description

common.mk: Fix BASEDIR to manage github-action workspace directory.

When using github-action the `spksrc` repository sits under `/github/workspace`.  Therefore the `BASEDIR` value set in `spksrc.common.mk` is not taking this into account, making publishing to file as `make setup-synocommunity` had failed previously with the following:
```
2024-04-06T10:42:28.7394231Z ##[group] ---- initialize build
2024-04-06T10:42:29.5904113Z mk/spksrc.common.mk:59: ../../mk/spksrc.archs.mk: No such file or directory
2024-04-06T10:42:29.5905666Z make: *** No rule to make target '../../mk/spksrc.archs.mk'.  Stop.
2024-04-06T10:42:29.7350751Z sed: can't read local.mk: No such file or directory
2024-04-06T10:42:29.7510857Z ##[endgroup]
```
Above that error can be seen the actual path being used by github-action:
```
"/github/file_commands" -v "/home/runner/work/spksrc/spksrc":"/github/workspace" ghcr.io/synocommunity/spksrc:latest
```

Fixes: #6061
Follow-up to: #6064 (thnx @hgy59) and broken by #6002

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -4,7 +4,7 @@
 
 # Set basedir in case called from spkrc/ or from normal sub-dir
 ifeq ($(BASEDIR),)
-ifneq ($(shell basename $(CURDIR)),spksrc)
+ifeq ($(filter spksrc workspace,$(shell basename $(CURDIR))),)
 BASEDIR = ../../
 endif
 endif


th0ma7:
@hgy59 @mreid-tt  and @Safihre : I believe this should solve the remaining issue.  Tested locally by symlinking `workspace -> spksrc` and `local.mk` files gets generated as expected.

hgy59:
I hate such hard coded values, but can't imagine another solution...

But for the test targets, we should not use `mk/spksrc.test-rules.mk`

th0ma7:
> I hate such hard coded values, but can't imagine another solution...

Likewise.

> But for the test targets, we should not use `mk/spksrc.test-rules.mk`

Why so?  Any other option that comes to mind?  Besides merging it into the `spksrc/Makefile` which I would hate to see there.

mreid-tt:
This is a bit over my head but seems sensible.

hgy59:
> Why so? Any other option that comes to mind? Besides merging it into the `spksrc/Makefile` which I would hate to see there.

Because `spksrc-test-rules.mk` is designed with another scope. It must be included by spksrc/Makefile only and not by make files for cross, spk and diyspk packages.

It's the ugly `include mk/spksrc.common.mk` that I'm complaining about in `spksrc-test-rules.mk`.
Even when you can remove this include, I would recommend to move `spksrc-test-rules.mk` to a different folder like spksrc/test.

But why not creating a new make file like `/spksrc/Makefile.test` ?

th0ma7:
Ok, coming up with something as https://github.com/SynoCommunity/spksrc/pull/6065#issuecomment-2041100735 will take me a little longer.  I'll have a look at it  over the coming days and create a PR (or reuse yours @hgy59) so it uses the default publish errror callout.

In the meantime I belive this simple fix fixes the immediate issue.  I'll merge this as-is and let's confirm it works ok.

I suggest we look into adding the other enhancements elsewhere, possibly into your PR https://github.com/SynoCommunity/spksrc/pull/6065 directly as this would keep the discussion thread:
1. Potentially using a `Makefile.test` or similar
2. make publish-arch* standard error

Are we ok with this proposal?

hgy59:
> Are we ok with this proposal?

I will merge #6065 with changes of spksrc.publish.mk only.

The other tasks shold be done in dedicated PRs.

hgy59 (approved):
LGTM

th0ma7:
yes, let me do, yet another pr... sigh... and good catch.

PR #6144, opened 2024-06-14
fix output of make dependency-list

hgy59:
## Description

- avoid the make command and the list of all packages in the output
- this is a leftover of PR #6002

Added some other fixes while on this topic
- fix znc package to work with `make dependency-list` and `make dependency-tree`
- fix package rename in prepare.sh
- remove call of `dependency-kernel-list`

### Type of change


hgy59:
@th0ma7 this is another side effect of #6002

I assume you were working on the test targets when you temporarily added some outputs.

Found this while comparing the ffmpeg and imagemagick dependencies for #6143

th0ma7:
@hgy59 sounds about right.  But why are you also removing `dependency-flat` ?

hgy59:
> @hgy59 sounds about right. But why are you also removing `dependency-flat` ?

`dependency-flat` in the global Makefile does not produce an output like `dependency-tree` and `dependency-list`
`dependency-flat` is internally used by those (and other) targets.

th0ma7:
I just noticed what level of makefile this refered to, gotcha.

th0ma7 (approved):
LGTM

hgy59:
@th0ma7 just verified, `dependency-flat` indeed has output. It is similar to `dependency-list` but every entry on a single line. 
This is harder to read then one line per package.

And found that `dependency-flat` contains duplicates. 
It needs to be investigated whether it still leads to long-running prepare jobs on GitHub.
I thought this was fixed in #5807

hgy59:
> And found that `dependency-flat` contains duplicates.
> It needs to be investigated whether it still leads to long-running prepare jobs on GitHub.

This is not an issue in prepare.sh, just with former `make dependency-flat` in the top folder.

PR #6148, opened 2024-06-16
ffmpeg 6.0.1 and tvheadend 4.3~366e562 update

th0ma7:
## Description

ffmpeg 6.1 update

Follow-up to https://github.com/SynoCommunity/spksrc/pull/6118#issuecomment-2170428183
Fixes https://github.com/SynoCommunity/spksrc/issues/5729

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6156, opened 2024-06-24
tvheadend: Fix SPK_DEPEND content parsing with '>'

th0ma7:
## Description

tvheadend:
- [x] Fix SPK_DEPEND content parsing with '>'
- [x] Update to latest git hash from June 23rd

Fixes #

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update


spk/tvheadend/Makefile
```diff
@@ -13,7 +13,7 @@ DSM_UI_DIR = app
 # 000-fix-version.patch from cross/tvheadend
 export TVH_VERS
 
-SPK_DEPENDS = "python311>=3.11.4-7"
+SPK_DEPENDS = "python311\>=3.11.4-7"
```
hgy59, line 16: so far this has not been an issue.
the install_dep_packages entry in the INFO file was correct for values like `"python311>=3.11.5-8"`

where have you encountered parsing problems?

th0ma7: Surprisingly, resulting packages from #6148 has the following in their INFO file `install_dep_packages="python311` whereas with the change it's now `install_dep_packages="python311>=3.11.4-7:ffmpeg6>6.0"` from this resulting build.

Still I wonder, I did not change the subsequent `SPK_DEPENDS := "$(SPK_DEPENDS):ffmpeg-bla>something".  So is it due to this following call where the `\>` gets translated OK?  Why haven't we hit this before? not sure really.

mreid-tt: I believe it may be the combination of the two since before the current release candidate you only had `SPK_DEPENDS = "python310"` so there were no special characters as input for the assignment operator. I wonder if a simple fix would be to revert everything back to how it was and remove the double quotes around the assignment value? So the two lines would be:

```
SPK_DEPENDS = "python311>=3.11.4-7"
```
and
```
SPK_DEPENDS := $(SPK_DEPENDS):ffmpeg$(subst 4,,$(FFMPEG_VERSION))>$(shell sed -n 's/^SPK_VERS = \(.*\)/\1/p' $(shell pwd)/../ffmpeg$(FFMPEG_VERSION)/Makefile | cut -c -3)
```

th0ma7: Indeed probably a combination of both where double-quotes does the interpretation at the second `SPK_DEPENDS`.  Tested with `\>` to both and end-result now is `install_dep_packages="python311>=3.11.4-7:ffmpeg6\>6.0"`, probably getting interpreted when being concatenated during the second call due to the `"`

th0ma7:
Pending build & testing on my home cut-the-cord system.

th0ma7:
Tested and migrated my own TVH home instance succesfully.  Also tested ffmpeg6, things looks ok for publishing.

PR #6183, opened 2024-07-28
Update default spksrc image to Debian12

th0ma7:
## Description

Update default spksrc image to Debian12

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes <s>small</s> framework changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ RUN apt update && apt install --no-install-recommends -y \
 	expect \
 	fakeroot \
 	flex \
+	gh \
 	g++-multilib \
 	gawk \
 	gettext \
@@ -89,12 +90,6 @@ RUN apt install --no-install-recommends -y \
 	python3-pip \
 	python3-virtualenv \
 	python3-yaml
-
-# Install hg github tool
-RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg && \
-	echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null && \
-	apt update && \
-	apt install gh
 
 # Clean-up apt db
 RUN apt clean && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN dpkg --add-architecture i386
 
 # Install required packages (in sync with README.rst instructions)
 # ATTENTION: the total length of the following RUN command must not exceed 1024 characters
-RUN apt-get update && apt-get install --no-install-recommends -y \
+RUN apt update && apt install --no-install-recommends -y \
 	autoconf-archive \
 	autogen \
 	automake \
@@ -78,7 +78,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 	zlib1g-dev
 
 # Python based apps
-RUN apt-get install --no-install-recommends -y \
+RUN apt install --no-install-recommends -y \
 	httpie \
 	mercurial \
 	meson \
@@ -97,7 +97,7 @@ RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | d
 	apt install gh
 
 # Clean-up apt db
-RUN apt-get clean && \
+RUN apt clean && \
 	rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Update locate db
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,7 +78,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 	zlib1g-dev
 
 # Python based apps
-RUN apt-get update && apt-get install --no-install-recommends -y \
+RUN apt-get install --no-install-recommends -y \
 	httpie \
 	mercurial \
 	meson \
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 	gh \
 	git \
 	gperf \
-	httpie \
 	imagemagick \
 	intltool \
 	jq \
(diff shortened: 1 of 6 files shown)


th0ma7:
I've been struggling with my newer PR I've been working on due to the old Debian 11 package set, which also doesn't work well with DSM 7.2 due to older libc version.  Above an update that I'll begin testing more in depth on my end and check if there's anything breaking elsewhere.

One thing that will require testing is the publishing that still relied on older python2 which is no longer available.

hgy59:
@th0ma7 I already had a look at debian 12

what about using nasm (and yasm) from debian instead of `native/nasm` and `native/yasm`?

And I added  `python3-mako` and `python3-virtualenv` but you probably have another solution for those...

And I would ommit man-db and manpages-dev since we never provide man pages in packages (due lacking man in DSM)

IMHO python2 is not available in debian 12 isn't it.

th0ma7:
> @th0ma7 I already had a look at debian 12
> 
> what about using nasm (and yasm) from debian instead of `native/nasm` and `native/yasm`?

good idea, let's add them by default.  for the time being framework could check if available in path, else use native.

> And I added `python3-mako` and `python3-virtualenv` but you probably have another solution for those...

yup, `python3-virtualenv` already in, and I just hit mako on one of my PR so I'll definitively add it as well.

> And I would ommit man-db and manpages-dev since we never provide man pages in packages (due lacking man in DSM)

well, as i am often running in my container as spksrc it is quite useful to have the man pages close by... 

> IMHO python2 is not available in debian 12 isn't it.

indeed, now gone.

hgy59:
@th0ma7 IMHO you will get errors, since the install RUN command is too long (currently 1184 characters)

```
# ATTENTION: the total length of the following RUN command must not exceed 1024 characters
```

if we really need to install all those components, we have to split the installation into two RUN commands
e.g use a new RUN für apt-get clean...

hgy59:
@th0ma7 As already mentioned somewhere, I suspected problems with old ppc archs under debian 12.

With recent tests I found that only a few packages fail to build for arch-ppc853x-5.2 under debian 12:

- ffmpeg(4)
- ffmpeg5
- ffmpeg6
- chromaprint
- synocli-file
- java-11-openjdk       [is not supported, should use REQUIRED_MIN_DSM = 6.0]
- synocli-devel
- comskip


Further analysis of ffmpeg revealed an error in the Makefiles of ffmpeg

The following configure arg must not be defined for OLD_PPC_ARCHS, but only for qoriq
```
CONFIGURE_ARGS += --extra-libs=-latomic
```

otherwise configure fails with
```
/spksrc/toolchain/syno-ppc853x-5.2/work/powerpc-none-linux-gnuspe/bin/powerpc-none-linux-gnuspe-gcc is unable to create an executable file.
```

th0ma7:
> IMHO you will get errors, since the install RUN command is too long (currently 1184 characters)

In theory fixed now with the RUN division I've setup.

> The following configure arg must not be defined for OLD_PPC_ARCHS, but only for qoriq
> ```
> CONFIGURE_ARGS += --extra-libs=-latomic
> ```

I adjusted ffmpeg accordingly, now to be confirmed if this will run properly as ? if I recall ? changing the `Dockerfile` isn't replicated immediately to github-action but rather on subsequent PR after merge.

hgy59:
@th0ma7 you didn't update to debian 12 yet.
you have to replace bullseye by bookworm 😆

hgy59:
@th0ma7 some more issues:

- with bookworm `gh` is not a valid package
- the third `RUN` must not end with `&& \`

th0ma7:
@hgy59 sigh... clearly an oversight, thnx.  now fixed (and tested locally) - should now be ok.

todo: adjust framework to make usage of local `nasm` and `yasm` by default if available.

hgy59:
> * with bookworm `gh` is not a valid package

This was a false comment
with buster the `gh` package is not available, with bookworm it is.
But why should we have github command line tool?

Dockerfile
```diff
@@ -51,44 +52,61 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
-# Install setuptools, wheel and pip for Python2
-RUN wget https://bootstrap.pypa.io/pip/2.7/get-pip.py -O - | python2
-# Install virtualenv and httpie for Python2
-# Use pip2 as default pip -> python3
-RUN pip2 install virtualenv httpie
+# Install hg github tool
+RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg && \
+	echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null && \
+	apt update && \
+	apt install gh
```
hgy59, line 97: as commented, gh is installable in debian-bookworm without additional prerequisites...

th0ma7: Thnx, now fixed and using default debian repository.

> But why should we have github command line tool?

I figured, when things goes sideways `gh` can be quite handy to play with your repository.  Totally optionnal, agreed, but as our project entirely sits under github, and its nice to have it around once in a while, it doesn't cost too much, further as now part of default debian repository.

I must admit, that I didn't had to use it in a while as I have a strong preference on `git`.  Although there may be something to do with it for future download options ... Unless you disagree I'd keep it included for now on?

th0ma7:
BTW, thnx for the review @hgy59 as I've been pursuing this during the too litle spare time I have currently and while simple it definitively needs another pair of eye :)

th0ma7:
@hgy59 I've updated `cmake` rules to switch to host nasm when available... Looking at the `cross/*` directory there would be other use cases to disable using `native/nasm`:
- `dav1d`
- `ffmpeg4-5-6`
- `lame`
- `libvpx`
- `openh264`
- `x264`

Wonder if it would be worth adding a `GNU_CONFIGURE_USE_NASM` similar to `cmake` ... ?  That could handle the `PATH` and `AS` environment variable handling... 

Any thoughts?

th0ma7:
@hgy59 I've been testing this new container image sucessfully since a few weeks already.  Besides proposal above https://github.com/SynoCommunity/spksrc/pull/6183#issuecomment-2259446120 which can wait later, anything else missing before merging?

hgy59:
> Wonder if it would be worth adding a `GNU_CONFIGURE_USE_NASM` similar to `cmake` ... ? That could handle the `PATH` and `AS` environment variable handling...
> 
> Any thoughts?

we don't need any nasm path nor native/nasm if nasm is a dev env prerequisite.

PR #6210, opened 2024-08-29
xz: Update from version 5.6.1+really5.4.5 to 5.6.2

th0ma7:
## Description

1. `xz`: Update from version 5.6.1+really5.4.5 to 5.6.2
2. Fix python 3.10-3.11 builds.
3. Fix build issue for hi3535 related to fd-8.7.0 with newer rustc >= 1.80.1 (now version fixed to 1.77.2 like ARMv5)

Superseeds part of https://github.com/SynoCommunity/spksrc/pull/6200

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

### TODO
- [ ] publish `jackett` https://github.com/SynoCommunity/spksrc/pull/6211
- [ ] publish `rutorrent` https://github.com/SynoCommunity/spksrc/pull/6215

PR #6216, opened 2024-08-30
Update ntopng to version 6.2

th0ma7:
## Description

Update ntopng to version 6.2

Relates to https://github.com/SynoCommunity/spksrc/pull/6210

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [x] Package update


hgy59:
@th0ma7 I would like to fix the patches for ntopng 4.2 first (working on it).

hgy59:
@th0ma7 to bypass the need of patching for _rw_locks_supported_, I modified the ntopng 4.2 cross Makefile like this:

[cross_ntopng_Makefile.patch](https://github.com/user-attachments/files/16821873/cross_ntopng_Makefile.patch)

But since the 6.2 build succeeds, the _rw_locks_supported_ patches seem not to be required anymore...

For your PR here, my proposal is limitted to:
1. remove the pre_patch target and run autogen.sh as first step in the pre_configure target

---

both the qoriq and comcerto2k-7.1 build failures seem to have the same cause.
So the libatomic patch for qoriq might be obsolete...

th0ma7:
In parallel I was looking at options to provide necessary privileges to ntopng, I tried the following:
```
{
        "defaults":{
                "run-as": "package"
        },
        "username": "sc-ntopng",
        "groupname": "synocommunity",
        "tool": [{
                "relpath": "bin/ntopng",
                "user": "package",
                "group": "package",
                "capabilities": "cap_net_raw,cap_net_admin",
                "permission": "0700"
        }]
}
```
It ends-up blocked at installation time as needing root privileges... not sure there's any way around that besides possibly a wiki info to set that up manually post install, such as
```
$ sudo /sbin/setcap 'CAP_NET_RAW+eip,CAP_NET_ADMIN+eip' "${SYNOPKG_PKGDEST}/bin/ntopng"
```

hgy59:
> - now having an issue with password at first login 🤷

I have the same issue with password at first login, but I suppose it is an issue with the webserver/web app or the browser.

Then I modified version 4.2 for DSM 7 (adjusting the paths in start-stop-service script and config file, installing config file into var instead of etc folder, and defining package user in privileges)
Patching the privilege file for root after installation made ntopng 4.2 working on DSM 7.2.

The working login dialog of ntpng 4.2 looks like this (default login admin/admin works and password must be changed before login succeeds):

grafik

But with ntopng 6.2 it looks like this (and default login admin/admin does not work `Login failed: invalid username or password.`):

initial page:
grafik

login page:
grafik


I guess that something is wrong with the ntopng web ui.
- it looks like the css/scripting does not work
- the password might not be handled correctly by the web ui...

Same error on DSM 6 and DSM 7

hgy59:
with ntopng 6.2 build of this PR, the `share/ntopng/httpdocs/dist` folder is empty
but it should have (at least) the following files
- third-party.css
- white-mode.css
- ntopng.css
- third-party.js
- login.js

th0ma7:
Well this is really good news in a sense that this is heading is a right direction, and most importantly, it buils (at least from my perspective).  I would propose the following:
1. Can you merge your `ntopng` 4.2 changes + confirm `jackett` PR is ?ok?
2. With this I could then rebase the `xz` and let it re-run
3. Then I could focus on the python build errors, which should be the last thing standing - actually it looks like failing on 32-bit archs only, I recall facing a similar issue but no clue what i did then?!?
4. then in parallel get the domino of other non-python pending PR waiting for `xz`

On a side note, we can then rebase this PR and pursue the work later on.  Feel free to push commits as you see fit while I try to solve the python issue by then?  Thoughts?

hgy59:
> 1. Can you merge your `ntopng` 4.2 changes + confirm `jackett` PR is ?ok?

This is not as easy as it seems, I first have to find a more actual commit first, that works with ntopng 4.2 on DSM 7.
My working test is based on commit 5f8739599f3f4b6c7f60233b467f3b09444a9b94 (10.08.2023) and I used the debian 10 based spksrc image.

hgy59:
@th0ma7 for the ntopng 2.4 branch I have two options, but only debian 11 spksrc image successfully builds.

1. based on commit 03af560a70687f045d6f5d1a59d26792997996d8 ntopng 4.2 successfully builds with debian 11 spksrc image
2. based on current master (commit 9af59c062b2ade32f5edf2a3a2a9959ee0ab505e) ntopng 4.2 build fails with debian 12 spksrc image. The failure is, that while linking ntopng, the curl, mysql and lua libraries can't be found.

th0ma7:
> 2. based on current master (commit [9af59c0](https://github.com/SynoCommunity/spksrc/commit/9af59c062b2ade32f5edf2a3a2a9959ee0ab505e)) ntopng 4.2 build fails with debian 12 spksrc image. The failure is, that while linking ntopng, the curl, mysql and lua libraries can't be found.

@hgy59 have a look at the 6.2 Makefile I revised, I believe the solution may be somewhat close to that.  I do the following at the end:
```
sed -i.orig -e "/^LIBS/s/[[:space:]]\+/ /g" \
                            -e "/^LIBS/s/ -l -l/ -l/g" \
                            -e "/^LIBS/s/ -lssl -lssl / -lssl /g" \
                            -e "/^LIBS/s/ -lssl / -lssl -llua -lmysqlclient /g" \
                            -e '/^LIBS/s/$$(LIBPCAP) $$(LUA_LIB) $$(LIBRRDTOOL_LIB)/$$(LIBPCAP) $$(LIBRRDTOOL_LIB)/g' \
                            ./Makefile
```
Basically it : 
1. removes the multiple space (which facilitates later sed exec)
2. removes the `-l -l` (I recall there where also empty `-l` followed by `-L` in 4.2 version needing attention
3. removes the duplicate `-lssl`
4. most importantly, removes the `$$(LUA_LIB)` and replaces that with `-llua -lmysqlclient` (you probably have to add `-lcurl` for 4.2.  To do so I use the `-lssl` and add to it.

Hopefully this helps, worst case, create a PR and I can have a look at it.  If I recall corectly, he reason is that it messes around with local libraries instead of only using the ones from our install prefix.

th0ma7:
@hgy59 if that is of interest, I did a quick rebase of my changes.  I got it to build for (`aarch64-7.1` is probably a false-positive):
```
-rw-r--r-- 1 spksrc spksrc 61235200 Sep  3 10:30 ntopng_aarch64-6.2.4_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 61040640 Sep  3 10:30 ntopng_armv7-6.2.4_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 61603840 Sep  3 10:32 ntopng_armv7-7.1_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 60631040 Sep  3 10:46 ntopng_comcerto2k-7.1_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 63334400 Sep  3 10:46 ntopng_evansport-6.2.4_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 63989760 Sep  3 10:48 ntopng_evansport-7.1_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 60231680 Sep  3 11:02 ntopng_hi3535-6.2.4_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 62423040 Sep  3 11:01 ntopng_qoriq-6.2.4_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 63324160 Sep  3 11:03 ntopng_x64-6.2.4_6.2-5.spk
-rw-r--r-- 1 spksrc spksrc 64102400 Sep  3 11:01 ntopng_x64-7.1_6.2-5.spk
```

th0ma7:
closed. superseeded with https://github.com/SynoCommunity/spksrc/pull/6220

Issue #6224, opened 2024-09-07
how to get log output during development of wizard pages?


hgy59:
When developing wizard pages, I can write log output to a log file from within the wizard shell script files.

But I can't figure out where I can find the output of `{ console.log(...); }` called by java script code.

Another solution, instead of using `console.log`, would be to write to a log file from java script. Any suggestions on how to write to a (server local) file from java script code of a package installation wizard?

smaarn:
```
But I can't figure out where I can find the output of { console.log(...); } called by java script code.
```

From what I recall this goes into your actual web browser logs.

mreid-tt:
@hgy59 this would usually be in your browser's developer console... see: https://javascript.info/devtools for some further information. Happy to collaborate on this if you are trying to resolve a particular issue as I've gone down the rabbit hole a few times when developing the wizards for packages like ownCloud and a few others.

hgy59:
@mreid-tt I want to check the version of existing minio data format in existing data shares in the upgrade wizard of minio (#5555).
I think it is best to get this information in the activate function of a wizard page (or deactivate of the previous page).

Since you have used this kind of functions in the wizards of the owncloud package, you probably have some hints for me?

hgy59:
> ```
> But I can't figure out where I can find the output of { console.log(...); } called by java script code.
> ```
> 
> From what I recall this goes into your actual web browser logs.

@smaarn thank you!  Yes, it is there 😛

hgy59:
So I see that java script is on the client, but I need to make checks on the server when navigating to a wizard page...

mreid-tt:
Well you won't be able to dynamically get details on the fly however you can run some specific code and store the results in variables that can then be used by JavaScript. See the checks that I helped with in SABnzbd: https://github.com/Safihre/spksrc/blob/master/spk/sabnzbd/src/wizard/upgrade_uifile.sh

hgy59:
I think in the wizard I can't make checks on the server and have to postpone checks to a validate_* installer function.

@mreid-tt compared with the owncloud install wizard: 
The wizard does not validate whether the provided restore path exists on the server.
This validation is done in `service_postinst` (to be honest, it is not validated, but the restore is skipped when the backup file is not found or has unexpected name).

hgy59:
[...]

PR #6225, opened 2024-09-10
ownCloud: Update to v10.15.0

mreid-tt:
## Description

1. Update ownCloud to version 10.15.0
2. Use recommended MariaDB database ([migration guide](https://github.com/SynoCommunity/spksrc/wiki/ownCloud-Database-Migration))
3. Optimise initial ownCloud setup
4. Refine scripts for backup restoration
5. Fix scripts for uninstall validation

Fixes # 

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/cross/owncloud/Makefile
+++ b/cross/owncloud/Makefile
@@ -1,6 +1,6 @@
 PKG_NAME = owncloud
-PKG_VERS = 10.14.0
-PKG_DATE = 20240226
+PKG_VERS = 10.15.0
+PKG_DATE = 20240724
 PKG_EXT = tar.bz2
 PKG_DIST_NAME = $(PKG_NAME)-complete-$(PKG_DATE).$(PKG_EXT)
 PKG_DIST_SITE = https://download.owncloud.com/server/stable
--- a/cross/owncloud/digests
+++ b/cross/owncloud/digests
@@ -1,3 +1,3 @@
-owncloud-complete-20240226.tar.bz2 SHA1 a2b54593da3caebdcf2212480c8e046b7cbad40b
-owncloud-complete-20240226.tar.bz2 SHA256 7b23e7f189f5ceb606fb95e96b5caa202299b3cea90e524fc59bdd3c9c30853c
-owncloud-complete-20240226.tar.bz2 MD5 dc135c062cb9c21656e38309306b6067
+owncloud-complete-20240724.tar.bz2 SHA1 9f514c298d16abd96f807522e5c134c1dbcf23bd
+owncloud-complete-20240724.tar.bz2 SHA256 f7ae9c9c00dee744306e4be058a9de20246730a4210f41e0c86ce9da60562ad5
+owncloud-complete-20240724.tar.bz2 MD5 9862606219b636f0c791d2efd3dbae97
--- a/spk/owncloud/Makefile
+++ b/spk/owncloud/Makefile
@@ -13,7 +13,7 @@ SPK_DEPENDS = WebStation:PHP7.4:Apache2.4
 MAINTAINER = SynoCommunity
 DESCRIPTION = ownCloud is a personal cloud which runs on your own server and gives you freedom and control over your own data.
 DISPLAY_NAME = ownCloud
-CHANGELOG = "1. Update ownCloud to 10.15.0.<br/>2. Various script fixes and refinements."
+CHANGELOG = "1. Update ownCloud to 10.15.0.<br/>2. Various script fixes and refinements.<br/>3. Optimise initial ownCloud setup."
 HOMEPAGE = https://owncloud.com/
 
 LICENSE = AGPLv3
--- a/spk/owncloud/Makefile
+++ b/spk/owncloud/Makefile
@@ -4,8 +4,8 @@ SPK_REV = 18
 SPK_ICON = src/owncloud.png
 
 DEPENDS = cross/owncloud
-# Pure PHP package, make sure ARCH is not defined
-override ARCH=
+# Pure PHP package, make sure ARCH is noarch
+override ARCH=noarch
 
 REQUIRED_MIN_DSM = 6.0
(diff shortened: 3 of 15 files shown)


hgy59:
> hey @hgy59, recently you mentioned a validation check that was missing in [#6224 (comment)](https://github.com/SynoCommunity/spksrc/issues/6224#issuecomment-2336403740). With this new build I'm wondering what's the best way to solve this issue.

@mreid-tt you need a `validate_preinst ()` function in `service-setup.sh`
Here you implement a part of the code in `service_postinst` titled with `# Check backup file path`.
If the user wants to restore from database, but the `$filename` does not match the expected prefix (i.e. the path is missing or the file does not exist or the file has not the expected name) then you echo an error message (to stdout) and stop the execution with "exit 1".
This will show the message to the user and prevent the installation (i.e. the wizard is still open and the user can navigate back to fix the path or select other options)
You find commented samples of validate_* in the service-setup.sh script in spk/demoservice/src

hgy59:
BTW the demoservice package is a good playground for development.

mreid-tt:
@hgy59 thanks for the feedback. It's been a while since I worked on this so I forgot about these functions. Will make the suggested changes in a new commit and test.

mreid-tt:
@hgy59 I believe the scripts are good now. I also took the time to optimise them to use `#!/bin/sh` commands. Let me know if you have any further suggestions before I merge it.

mreid-tt:
@hgy59, I revisited the package and noticed some warnings in the settings screen that have been present since I resumed publishing it. After some investigation, I managed to resolve three of them: a cron warning, an HSTS warning, and a memory cache warning.

However, two warnings remain: one regarding file locking and another concerning the use of SQLite for the database. With a better understanding of how to back up and restore MySQL/MariaDB databases, I'm considering switching the package back to MariaDB. I'd love to hear your thoughts on this.

For the file locking issue, we might be able to address it by using Redis. We already have a Redis package that users could manually configure, or I could try integrating it into the setup wizard.

Overall, the current configuration seems aligned with the "Private Home Server with Low Access Rates" recommendation from the [ownCloud configuration guide on caching](https://doc.owncloud.com/server/next/admin_manual/configuration/server/caching_configuration.html#configuration-recommendations-based-on-type-of-deployment). To move to the next tier, supporting a "Small Server," we'd need to resolve these two remaining issues. I'd appreciate your input on whether you consider this effort to be worthwhile.

mreid-tt:
@hgy59 I've made a major commit that transitions the backend database to MySQL. Unfortunately, automating the process through a script isn’t reliable at this time, so I’ve created a Wiki page with instructions for users to manually complete the steps before updating to this version. You can review it here: https://github.com/SynoCommunity/spksrc/wiki/ownCloud-Database-Migration. I'd appreciate your feedback on it.

mreid-tt:
> @hgy59 I've made a major commit that transitions the backend database to MySQL. Unfortunately, automating the process through a script isn’t reliable at this time, so I’ve created a Wiki page with instructions for users to manually complete the steps before updating to this version. You can review it here: https://github.com/SynoCommunity/spksrc/wiki/ownCloud-Database-Migration. I'd appreciate your feedback on it.

Hi @th0ma7, I'd appreciate your input on this. I initially considered including this as part of an upgrade script, but there are too many potential failure points—such as updating active PHP extensions, managing TCP/IP connections to MariaDB, and the migration itself potentially failing. I'm not sure if users would be open to handling the migration manually, but I wanted to get your perspective.

mreid-tt:
@hgy59, @th0ma7, I'm ready to proceed with this but would appreciate your input on the database migration approach before moving forward. If there's no feedback, I'll plan to publish later this week and address any end-user issues as they arise.

th0ma7:
This is really well documented ...

A few comments:
- I would suggest offering the user to install mariadb from the repository first (otherwise probably done at package installation time) - explaining that would be helpful.  Also a link to MariaDB wiki or something so people can have something that explain what this really is (for newbies)
- I haven't checked your code but I would assume there is a major disclaimer at install time? ... I just checked and now unsure, at upgrade time you state it only support MySQL although this speak to migrating to MariaDB?
- `Set a random password for the database user.` - most probably best practice (I haven't worked with DB in years), but I'd leave an option to set a user-defined passwd.
- `Note: If an error occurs, ` - unclear if this applies to DSM6 or DSM7 or both due to the vertical bar to the side?

yup, well document overall, nice job!

mreid-tt:
> * I would suggest offering the user to install mariadb from the repository first (otherwise probably done at package installation time) - explaining that would be helpful.  Also a link to MariaDB wiki or something so people can have something that explain what this really is (for newbies)

Well it's listed as a pre-requisite in the Makefile so I'm not sure there is a way to show anything before, other than in the release notes or package description.

> * I haven't checked your code but I would assume there is a major disclaimer at install time? ... I just checked and now unsure, at upgrade time you state it only support MySQL although this speak to migrating to MariaDB?

If it's a fresh install it will default to MariaDB. The main challenge is someone moving from 10.14.0 to 10.15.0 who would be running on SQLite that would need to migrate their database to MySQL. Ah! I see your point about mentioning MySQL when I mean MariaDB, yes that language can be made more consistent.

> * `Set a random password for the database user.` - most probably best practice (I haven't worked with DB in years), but I'd leave an option to set a user-defined passwd.

This is the password for the application to connect to the database which is something the user will not be interacting with. The random password is a recommendation from the upstream repo's installation guide which I agree with. It cuts down on admin accounts to remember and manage for the user.

> * `Note: If an error occurs, ` - unclear if this applies to DSM6 or DSM7 or both due to the vertical bar to the side?

The error could potentially occur with either DSM 6 or 7. So one thing which I saw consistently on both is if you don't turn maintenance mode off in the previous step the migration will fail and will leave a partially constructed database in MariaDB. This is why the easiest way to recover would be to just dump it and start the migration again.

> yup, well document overall, nice job!

Yeah, I tried to make it as copy and paste as possible but my main concern was for users who may not be comfortable with command line.

PR #6229, opened 2024-09-14
BicBucStriim: Update to v3.4.2

mreid-tt:
## Description

This updates to the latest BicBucStriim v3.4.2 which now requires PHP 8.2 and DSM 7.2.

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #6249, opened 2024-10-03
Framework: Prevent noarch builds without DSM 5.2

mreid-tt:
## Description

This update ensures `noarch` builds only occur when DSM 5.2 is selected. Previously, `noarch` builds were triggered even when DSM 5.2 wasn't chosen. Now, `noarch`, `noarch-6.1` and `noarch-7.0` are properly tied to DSM 5.2, DSM 6.2 and DSM 7.x, respectively, improving build accuracy.

Follow-up on #6247 and https://github.com/SynoCommunity/spksrc/pull/6229#issuecomment-2351049498.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,19 +126,18 @@ jobs:
     steps:
       - id: set-matrix
         run: |
-          # Start with the noarch architectures at the top
+          # Create a matrix as a JSON object
           matrix='{"include": ['
-          matrix+='{"arch": "noarch"},'
-          matrix+='{"arch": "noarch-6.1"},'
-          matrix+='{"arch": "noarch-7.0"},'
 
-          # Add other architectures based on input flags
+          # Add architectures based on input flags
           if [ "${{ github.event.inputs.add_dsm52_builds }}" == "true" ]; then
+            matrix+='{"arch": "noarch"},'
             matrix+='{"arch": "x86-5.2"},'
             matrix+='{"arch": "88f6281-5.2"},'
             matrix+='{"arch": "ppc853x-5.2"},'
           fi
           if [ "${{ github.event.inputs.add_dsm62_builds }}" == "true" ]; then
+            matrix+='{"arch": "noarch-6.1"},'
             matrix+='{"arch": "x64-6.2.4"},'
             matrix+='{"arch": "aarch64-6.2.4"},'
             matrix+='{"arch": "evansport-6.2.4"},'
@@ -148,6 +147,7 @@ jobs:
             matrix+='{"arch": "qoriq-6.2.4"},'
           fi
           if [ "${{ github.event.inputs.add_dsm71_builds }}" == "true" ]; then
+            matrix+='{"arch": "noarch-7.0"},'
             matrix+='{"arch": "x64-7.1"},'
             matrix+='{"arch": "aarch64-7.1"},'
             matrix+='{"arch": "evansport-7.1"},'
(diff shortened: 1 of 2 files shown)


mreid-tt:
@hgy59, after publishing several `noarch` packages with `REQUIRED_MIN_DSM = 6.0` or higher, I noticed that builds for firmware 3.1-1594 were still being included, requiring manual removal from the repo. Since we no longer support this firmware, I'm proposing this PR to exclude it from the build process entirely.

mreid-tt:
> we need another solution, noarch is still needed for DSM 5.2

Is it possible to rename `noarch` to `noarch-3.1` within the framework? I suspect the lack of a version number might be causing issues with the `REQUIRED_MIN_DSM` rules.

Do you have any other thoughts on what might be causing the issue?

mreid-tt:
@hgy59, I've reconsidered my approach assuming that the reason the block fails is that the `TCVERSION` isn't set. I'm going to test the theory in my local fork to confirm if this is the case.

hgy59:
> EDIT: So that didn't work. I'm open to other ideas on this.

I guess that we need to make TCVERSION mandanory for noach builds.
The related code is in `spksrc.spk.mk`.

currently to rebuild all noarch packages use (in the spk/{package} folder):
```
make clean
make
make TCVERSION=7.0
make TCVERSION=6.1
make TCVERSION=1.1
```

this should be changed to 
```
make clean
make TCVERSION=3.1
make TCVERSION=7.0
make TCVERSION=6.1
make TCVERSION=1.1
```

mreid-tt:
> I guess that we need to make TCVERSION mandanory for noach builds.
> The related code is in `spksrc.spk.mk`.

Unfortunately, I've tried these two variations but neither work:

**Original**
https://github.com/SynoCommunity/spksrc/blob/711444bddb9da3cb4c2147dcfa46a7366d77014a/mk/spksrc.spk.mk#L71-L76

**Option 1**
```
else
# Set default TCVERSION to 3.1 for noarch if not provided
TCVERSION = 3.1
SPK_TCVERS = all
TC_OS_MIN_VER = 3.1-1594
endif
ARCH_SUFFIX = -$(SPK_TCVERS)
endif
```

**Option 2**
```
else
# Set default TCVERSION to 3.1 for noarch if not provided
export TCVERSION=3.1
SPK_TCVERS = all
TC_OS_MIN_VER = 3.1-1594
endif
ARCH_SUFFIX = -$(SPK_TCVERS)
endif
```

hgy59:
One problem is the `ARCH=` assignment to force "noarch" builds.
There are some targets that do not need a definition of ARCH, like `make clean`, `make digests`, ... 

It would be easier to use something like `ARCH = noarch` instead, but then `noarch` must be handled to not expect a toolchain.

mreid-tt:
I'm not entirely sure how to address that suggestion, as it's getting a bit beyond my current understanding.

However, an alternative approach could be to segment the `noarch` versions by DSM as follows:

- Include `noarch` in the DSM 5.2 architectures
- `noarch-6.1` in the DSM 6.2 architectures
- `noarch-7.0` in both DSM 7.x architectures

This way, the builds will align with the appropriate DSM version, and we can avoid the need for additional handling for `noarch`. What do you think?

mreid-tt:
@hgy59, I've tested the simplified approach in my fork, and the build works as expected. Do you think this solution is ready for merging?

hgy59:
> @hgy59, I've tested the simplified approach in my fork, and the build works as expected. Do you think this solution is ready for merging?

sorry, I have worked on an other solution, using `override ARCH=noarch` for `noarch` packages.
will create a dedictated PR...

mreid-tt:
> sorry, I have worked on an other solution, using `override ARCH=noarch` for `noarch` packages.
> will create a dedictated PR...

No problem, once the issue with `noarch` is resolved I'm happy either way. Will close this one.

mreid-tt:
Superseded by #6250.

PR #6250, opened 2024-10-03
Update noarch build

hgy59:
## Description

redesign of noarch build.
- noarch packages should only be build for specific TCVERSION, otherwise REQUIRED_MIN_DSM is not evaluated
- noarch packages must now define `override ARCH=noarch` instead of `override ARCH=`

Replacement for #6249.
Contains a bugfix for #6247 that removed all arch specific builds in github build action.

## Checklist

### Type of change

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
       package:
         description: 'Package to build'
         required: true
-        default: 'syno-magnet'
+        default: 'adminer'
       publish:
         description: 'Publish to repository'
         required: false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,14 @@ on:
         options:
           - 'true'
           - 'false'
+      add_noarch_builds:
+        description: 'Include noarch packages'
+        required: false
+        default: 'true'
+        type: choice
+        options:
+          - 'true'
+          - 'false'
       add_dsm72_builds:
         description: 'Include DSM 7.2 archs'
         required: false
@@ -126,17 +134,25 @@ jobs:
     steps:
       - id: set-matrix
         run: |
-          # Start with the noarch architectures at the top
+          # Create a matrix as a JSON object
           matrix='{"include": ['
-          matrix+='{"arch": "noarch"},'
-          matrix+='{"arch": "noarch-6.1"},'
-          matrix+='{"arch": "noarch-7.0"},'
-
-          # Add other architectures based on input flags
-          if [ "${{ github.event.inputs.add_dsm52_builds }}" == "true" ]; then
-            matrix+='{"arch": "x86-5.2"},'
-            matrix+='{"arch": "88f6281-5.2"},'
-            matrix+='{"arch": "ppc853x-5.2"},'
+          
+          if [ "${{ github.event.inputs.add_noarch_builds }}" == "true" ]; then
+            matrix+='{"arch": "noarch-1.1"},'
+            matrix+='{"arch": "noarch-3.1"},'
+            matrix+='{"arch": "noarch-6.1"},'
+            matrix+='{"arch": "noarch-7.0"},'
+          fi
+          if [ "${{ github.event.inputs.add_dsm72_builds }}" == "true" ]; then
+            matrix+='{"arch": "x64-7.2"},'
+            matrix+='{"arch": "aarch64-7.2"},'
+          fi
+          if [ "${{ github.event.inputs.add_dsm71_builds }}" == "true" ]; then
+            matrix+='{"arch": "x64-7.1"},'
+            matrix+='{"arch": "aarch64-7.1"},'
+            matrix+='{"arch": "evansport-7.1"},'
+            matrix+='{"arch": "armv7-7.1"},'
+            matrix+='{"arch": "comcerto2k-7.1"},'
           fi
           if [ "${{ github.event.inputs.add_dsm62_builds }}" == "true" ]; then
             matrix+='{"arch": "x64-6.2.4"},'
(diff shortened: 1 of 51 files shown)


hgy59:
@mreid-tt, @th0ma7 I have added a workaround for non interactive builds.

but I can't run interactive builds:
grafik

th0ma7:
I can provide you assistance but I probably won't be able to look at this before Sunday?  Maybe @mreid-tt can help?

.github/workflows/build.yml
```diff
@@ -126,19 +134,27 @@ jobs:
...
-          matrix+='{"arch": "noarch-6.1"},'
-          matrix+='{"arch": "noarch-7.0"},'
-
-          # Add other architectures based on input flags
-          if [ "${{ github.event.inputs.add_dsm52_builds }}" == "true" ]; then
-            matrix+='{"arch": "x86-5.2"},'
-            matrix+='{"arch": "88f6281-5.2"},'
-            matrix+='{"arch": "ppc853x-5.2"},'
+          
+          if [ "${{ github.event.inputs.add_noarch_builds }}" == "true" -o "${{ github.event.inputs.add_noarch_builds }}" == "" ]; then
```
mreid-tt, line 140: I don't think this syntax is correct. Logical OR in GitHub actions is || not -o as in bash. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions

mreid-tt:
I'll take a closer look later tonight and perhaps suggest a better fix than the workaround presented.

hgy59:
> EDIT: @hgy59, I've taken the liberty of adding a commit to this PR (hope you don't mind) which should resolve the issue you were trying to workaround. Have not done much testing on it.

@th0ma7 thanks, that makes sense (and it works).

hgy59:
Some thoughts for improvement:

- In interactive builds we could optimize the prepare step by evaluating the dependencies of the requested package only. 
- Since we already have a list of arch and noarch packages to build in automated builds, we could disable the noarch builds at all when no noarch packages are found to build (and disable the arch builds, if only noarch packages are to build).

mreid-tt:
> * Since we already have a list of arch and noarch packages to build in automated builds, we could disable the noarch builds at all when no noarch packages are found to build (and disable the arch builds, if only noarch packages are to build).

It's too bad that the `prepare.sh` doesn't also know the required version of DSM needed for the `arch` packages. If it did, you could extend your approach to just enable those required DSM versions for the automated builds. As it is, when we move the default from DSM 7.1 to DSM 7.2 in the future, the `build.yml` changes may get a bit awkward.

hgy59:
> It's too bad that the `prepare.sh` doesn't also know the required version of DSM needed for the `arch` packages. If it did, you could extend your approach to just enable those required DSM versions for the automated builds. As it is, when we move the default from DSM 7.1 to DSM 7.2 in the future, the `build.yml` changes may get a bit awkward.

I will have a look, and guess we can evaluate the min/max dsm requirements in the prepare script.

th0ma7:
Would there be a way to detect that a 7.2 build is needed when min_dsm is set?  i.e. jellyfin

mreid-tt:
@hgy59, I love the direction of the last change. Would this check for DSM 7.2 be on a per-package basis? So for example if Jellyfin had an external dependency on a .NET package and that package had a minimum of DSM 7.1 would it evaluate both packages and build for DSM 7.1 and DSM 7.2?

hgy59:
> @hgy59, I love the direction of the last change. Would this check for DSM 7.2 be on a per-package basis? So for example if Jellyfin had an external dependency on a .NET package and that package had a minimum of DSM 7.1 would it evaluate both packages and build for DSM 7.1 and DSM 7.2?

Yes, if any package requires dsm 7.2 (i.e. currently jellyfin only) all packages that are built are built for dsm 7.2 too.
- I didn't want to built for DSM 7.2 in the arch-7.1 matrix variant in this case, since the artifacts have the name of the matrix-arch
- in the build job it is unknown whether the DSM 7.2 built is manually triggered or comes from the min dsm requirement

mreid-tt:
@hgy59, I noticed that in the last build, Jellyfin did compile, but I expected the DSM 7.2 build to be additive to the default. My understanding was that once an architecture-specific build (arch) is triggered, the default builds would still be included. I thought the original logic was set to build `noarch` when needed, and similarly, build `arch` if required. Could you clarify if this was the intent?

mreid-tt:
@hgy59, looks like things built well and as expected. The only suggestion I would have is perhaps the naming of the variables. `min_dsm72_packages` perhaps could be `inc_dsm72_packages` or `req_dsm72_packages` and `has_min_dsm72_packages` could just be `has_dsm72_packages`. But that is minor so I'm not pushing for that change.

hgy59:
> @hgy59, looks like things built well and as expected. The only suggestion I would have is perhaps the naming of the variables. `min_dsm72_packages` perhaps could be `inc_dsm72_packages` or `req_dsm72_packages` and `has_min_dsm72_packages` could just be `has_dsm72_packages`. But that is minor so I'm not pushing for that change.

I am happy with  `has_min_dsm72_packages`, it is abbr. of `has_required_min_dsm72_packages`.
The `min_dsm72_packages` is just the list of such packages.
I will revert the definiion of the related action variable as it is used for logging only.

hgy59:
@mreid-tt all noarch packages needs an update. I have to push the remaining.

Since you have an open PR for OwnCloud #6225, you will get a merge conflict. I can omit the update of spk/owncloud/Makefile. This will break the build of owncloud on the master branch and you will have to adjust override ARCH in your PR after updating with upstream when this redesign is integrated.

mreid-tt:
> Since you have an open PR for OwnCloud #6225, you will get a merge conflict. I can omit the update of spk/owncloud/Makefile. This will break the build of owncloud on the master branch and you will have to adjust override ARCH in your PR after updating with upstream when this redesign is integrated.

I can push an update to the ownCloud PR to change the override ARCH in the Makefile once this PR is merged and then do a rebase. The PR is just about done, just needed feedback from you guys on the proposed migration strategy.

th0ma7:
@mreid-tt and @hgy59 i'm guessing changes are still on-going.

My short PR here didn't kicked-off the build as expected https://github.com/SynoCommunity/spksrc/pull/6253

actual build: https://github.com/SynoCommunity/spksrc/actions/runs/11194152942/job/31120397567

hgy59:
@th0ma7 none of the builds work since #6247 was merged. only noarch works.
It will be fixed with b29627208c02a0a6a09f668fe40f26e3303b9685 of this PR, but I suppose we need an immediate fix...

hgy59:
@mreid-tt @th0ma7 shall we merge this?

it has some issues that could be fixed later
- the syno-magnet package is built successfully but it is listed under builds with ERROR in the build status
- the following noarch packages fail to build:
   - plexivity
   - saltpad
   - subliminal

th0ma7:
One way or the other, I don't mind much as i won't have cycles to work on things for a week or two.  So I'm ok at waiting a little longer so build fixes gets in.

hgy59:
all issues fixed, ready to merge.

BTW the dependency-list optimization will come in a future PR.

mreid-tt (approved):

mreid-tt:
LGTM

mreid-tt:
> @mreid-tt and @hgy59 i'm guessing changes are still on-going.
> 
> My short PR here didn't kicked-off the build as expected #6253

@th0ma7, you should now be able to re-base your PR and it should build correctly.

PR #6255, opened 2024-10-05
Redesign dependency-list creation for github prepare action

hgy59:
## Description

This redesign speedsup the evaluation of dependent packages in github prepare action.
Instead of 3 to 5 minutes it now takes only a few seconds to create the dependency list that is required.

- add `dependency-list.sh` script to evaluate the dependencies
- use new script in global Makefile for the dependency-list target
- use global Makefile to create dependency-list in prepare.sh
- avoid variables in `DEPENDS` and `BUILD_DEPENDS` definitions

This script is designed to be used for the github build action only, to evaluate the packages to build and download sources for.
We still need the `dependency-list` target to find real dependencies for packages to build. The script uses all dependencies (it doesn't care about conditional dependencies) but when building packages we need the real dependencies and must not built dependencies not supported for specific archs or DSM versions.

## Mandatory framework changes

- makefile variables like `$(SPK_NAME)` are not allowed in dependency definitions anymore

## Checklist

### Type of change

- [x] Includes framework changes


hgy59:
@th0ma7 I wand to remove some obsolete dependencies.

Can't we remove `native/nasm` and `native/yasm` since those are prerequisites for the development environment (and included in the spksrc docker image)?

---

The same applies to `native/cmake` but here you have recent updates to `intel-vc-intrinsics` that needs `USE_NATIVE_CMAKE`, so I guess we must keep native/cmake.

Another one is native/cmake-legacy. This one is used with `USE_NATIVE_CMAKE_LEGACY` and is used for `mysql-connector-c` only.
Probably we can drop the mysql-connector and use mariadb-connector with mysql compatibility (`CMAKE_ARGS += -DWITH_MYSQLCOMPAT=ON`) instead?

th0ma7:
> Can't we remove `native/nasm` and `native/yasm` since those are prerequisites for the development environment (and included in the spksrc docker image)?

yes certainly.  although a fair bit of testing may be needed along with adapting the framework.  Just to manage expectations, I can certainly review code but my next focus really needs to be python... 
 
> The same applies to `native/cmake` but here you have recent updates to `intel-vc-intrinsics` that needs `USE_NATIVE_CMAKE`, so I guess we must keep native/cmake.

That's probably an oversight as I updated our debian image during the time I was workding on intel opencl and this may have be there while using the older image.  I'll check if that can be removed.

> Another one is native/cmake-legacy. This one is used with `USE_NATIVE_CMAKE_LEGACY` and is used for `mysql-connector-c` only. Probably we can drop the mysql-connector and use mariadb-connector with mysql compatibility (`CMAKE_ARGS += -DWITH_MYSQLCOMPAT=ON`) instead?

Feel free to remove that altogether.  I've been trying to keep `mysql-connector-c` alive as prerequisite for other packages that i ended-up migrating, mainly related to python if i recall.  There used to be `chromaprint` but I believe it's now solved.

Although my 2 cents are, cmake native and legacy have proved to be quite useful. I'd be tempted to keep the code for newer cmake as i'm sure we'll hit this again later in a year or two.

Also, I believe it would be time to start stripping off broken packages that never received any love since pre-DSM6, along with deleting them from our repository online.  As well, I would remove all the older toolchains as no longer needed and clean-up the code related to tc and tk accordingly.  This would reduce the framework weight for a bit and accelerated archs evaluation.

hgy59:
@th0ma7 thank you for the feedback.

I think I will hold back this PR for prior creating some smaller PRs for:
- removal of variables in DEPENDS definitions (multiple PRs to touch less packages in each)
- removal of native/nasm native/yasm
- removal of python2
- ...

Then, later on, in this PR:
- use scripts for all dependency lists (dependency-list and dependency-flat)
- remove dependency-tree. It is not very usefull (not readable) with such a lot of dependencies. It was the only source for dependency evaluation before we introduced dependency-list.

Finally, to remove OPTIONAL_DEPENDS, I would prefer separated PRs again.

hgy59:
@th0ma7, @mreid-tt  This is now ready to merge.

Are there any objections?

mreid-tt:
@hgy59, no objections from me (but I haven't reviewed it in great detail).

With regard to the failures for `lirc`, I assume this is because the Makefile is too old? Should the package be marked as broken?

hgy59:
> With regard to the failures for `lirc`, I assume this is because the Makefile is too old? Should the package be marked as broken?

`lirc` is a package that depends on kernel sources (REQUIRE_KERNEL = 1),  so it is expected to fail when building for generic archs.
And it was never mirgated to DSM 7 (still using the deprecated installer script).
And I guess it will never run on DSM 7 due to restrictions for USB devices.

It successfully builds for hi3535 and evansport for DSM 6.2.4.
qoriq is not supported and ARMv5 (88f6281) is lacking the kernel sources (this might be an issue).

`licr` has no other dependencies and is only triggerd to build when cross/lirc* or spk/lirc was changed, so no need to declare the package as broken.

th0ma7 (approved):
Lgtm, script code is nice and clean, well done.  Maybe missing in your sort is a `-u` in case there are duplicates such as eventually depending on both ffmpeg and python.

Also, i would suggest modifying our wiki (no clue where?) for the cross/appname to make that mandatory that they must match and no makefile variables are to be used.

Lastly, having any timing difference between run with your new script vs old method?  (On my cell phone so maybe you already provided in your pr description).

hgy59:
> Lastly, having any timing difference between run with your new script vs old method? (On my cell phone so maybe you already provided in your pr description).

Yes it is in the description 
Dependency evaluation is now typ. 8-10 sec (vs 5-6 min)

hgy59:
> Lgtm, script code is nice and clean, well done. Maybe missing in your sort is a `-u` in case there are duplicates such as eventually depending on both ffmpeg and python.

The functions `get_python_dependency` and `get_ffmpeg_dependency` add only cross/python* and cross/ffmpeg* to the depencencies. 
Those are added for futher evaluation and duplicates will be removed (`get_dependencies` always removes duplicates and is called until no additional deps are found)

hgy59:
> Also, i would suggest modifying our wiki (no clue where?) for the cross/appname to make that mandatory that they must match and no makefile variables are to be used.

now added to makefile variables wiki page at https://github.com/SynoCommunity/spksrc/wiki/Makefile-variables

PR #6263, opened 2024-10-07
remove native nasm

hgy59:
## Description

- remove native nasm, it is a prerequisite in the dev. environment since update to debian 12
- remove native yasm
- update related Makefiles

Fixes # 

## Checklist

### Type of change


cross/openh264/Makefile
```diff
@@ -18,10 +18,11 @@ CONFIGURE_ARGS += -Dtests=disabled
...
 
-BUILD_DEPENDS = native/nasm
-NASM_PATH = $(realpath $(WORK_DIR)/../../../native/nasm/work-native/install/usr/local/bin)
-ENV += PATH=$(NASM_PATH):$$PATH
-ENV += AS=$(NASM_PATH)/nasm
+NASM_BINARY = $(shell which nasm)
+ifeq ($(NASM_BINARY),)
+$(error nasm not found. Please install NASM assembler)
+endif
+ENV += AS=$(NASM_BINARY)
```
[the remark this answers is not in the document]
hgy59: Found another issue in cross/zstd with ninja compile
When `CMAKE_USE_NASM = 1` is ***not defined*** it uses nasm to create `huf_decompress_amd64.S.o`, but if it is defined, it uses gcc to create the file.

without `CMAKE_USE_NASM = 1` 
```
[24/103] Building ASM object lib/CMakeFiles/libzstd_shared.dir/spksrc/cross/zstd-1.5.6_cmake/work-x64-7.1/zstd-1.5.6/lib/decompress/huf_decompress_amd64.S.o
```
with `CMAKE_USE_NASM = 1` 
```
[24/103] Building C object lib/CMakeFiles/libzstd_shared.dir/spksrc/cross/zstd-1.5.6_cmake_nasm/work-x64-7.1/zstd-1.5.6/lib/decompress/huf_decompress_amd64.S.o
```

hgy59:
Special NASM definitions for meson and cmake build systems are obsolete and removed now.
The auto configuration for nasm works as expected.

further analysis for gcc builds (spksrc.cross-cc.mk) is WIP.

cross/libvpx/Makefile
```diff
@@ -28,40 +30,33 @@ include ../../mk/spksrc.archs.mk
...
-ENV += PATH=$(NASM_PATH):$$PATH
-CONFIGURE_ARGS += --as=nasm
-CONFIGURE_ARGS += --disable-sse4_1
-endif
-
 ifeq ($(findstring $(ARCH),$(i686_ARCHS)),$(ARCH))
 CONFIGURE_ARGS += --target=x86-linux-gcc
+# --as=auto: prefers yasm over nasm
+CONFIGURE_ARGS += --as=auto
+CONFIGURE_ARGS += --disable-sse4_1
```
hgy59, line 52: @th0ma7 can you remember why you disabled sse4_1 for intel archs?
_it builds just fine without disabling it_

hgy59:
@th0ma7 the libraries and *.mk files are ready for review.

On the ffmpeg* packages there is some pending work...

1. OLD_PPC_ARCHS fail to build ffmpeg since 4.4.x (4.3.x are the last working releases, newer have issue with stdatomic)

2. There are some conditional dependencies for `alpine` `comcerto2k` and `monaco` archs.

   example:
   ```
   ifeq ($(findstring $(ARCH),alpine comcerto2k monaco $(ARMv8_ARCHS) $(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
   DEPENDS += cross/libaom
   CONFIGURE_ARGS += --enable-libaom
   endif
   ```

   Except for `arch-comcerto2k-7.1` we do not build additional packages for `alpine` and `monaco`
   If we would build dedicated packages for `alpine` and `monaco` we would need to exclude those archs in the `armv7` generic packages.
   Despite we have an option to exclude dedicated archs from generic ones with `REMOVE_FROM_GENERIC_ARCHS`, we cannot use github build to create and publish dedicated archs.
   And we never planned to created a single package for a list of dedicated archs...

hgy59:
and the third issue is with `cross/lame`:

even we configure with `--enable-asm` (and configure finds /usr/bin/nasm) it does not use the assembler.

cross/dav1d/Makefile
```diff
@@ -18,21 +18,4 @@ CONFIGURE_ARGS += -Denable_examples=false
...
-NASM_PATH = $(abspath $(PWD)/../../native/nasm/work-native/install/usr/local/bin)
-ENV += AS=$(NASM_PATH)/nasm
-ENV += PATH=$(NASM_PATH):$$PATH
-CONFIGURE_ARGS += -Denable_asm=true
-
-# Allow ASM on aarch64, disable on all others
-else ifeq ($(findstring $(ARCH),$(ARMv8_ARCHS)),$(ARCH))
-CONFIGURE_ARGS += -Denable_asm=true
-else
-CONFIGURE_ARGS += -Denable_asm=false
```
th0ma7, line 35: nasm is only used for x86/x86_64, while for aarch64 it uses the gcc assembler (`as`).  Considering the code above, you should have kept something such as:
```
ifeq ($(findstring $(ARCH),$(ARMv8_ARCHS) $(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
CONFIGURE_ARGS += -Denable_asm=true
else
CONFIGURE_ARGS += -Denable_asm=false
endif
```
While it probably builds ok, I recall code to fail on armv7 in multiple cases which led to enable assembly optimizations only on aarch64 and x86 archs.  I would suggest keeping this ifeq else usecase (and maybe elsewhere, haven't looked at all the code changes).

EDIT: On another comment I realized that issues with `as` on armv7 related to gcc <= 4.9, rather suggesting:

```

ifeq ($(findstring $(ARCH),$(ARMv8_ARCHS) $(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
CONFIGURE_ARGS += -Denable_asm=true

# only enable `asm` on armv7 when using gcc > 4.9.* (i.e. DSM7+)
else if ($(findstring $(ARCH),$(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
ifeq ($(call version_lt, ${TC_GCC}, 5),1)
CONFIGURE_ARGS += -Denable_asm=true
endif
else
CONFIGURE_ARGS += -Denable_asm=false
endif
```

cross/ffmpeg4/Makefile
```diff
@@ -59,10 +58,11 @@ OPTIONAL_DEPENDS += cross/shine
...
 ifeq ($(findstring $(ARCH),$(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
-BUILD_DEPENDS = native/nasm
-NASM_PATH = $(abspath $(PWD)/../../native/nasm/work-native/install/usr/local/bin)
-ENV += AS=$(NASM_PATH)/nasm
-ENV += PATH=$(NASM_PATH):$$PATH
+NASM_BINARY = $(shell which nasm)
+ifeq ($(NASM_BINARY),)
+$(error nasm not found. Please intall NASM assembler)
+endif
+ENV += AS=$(NASM_BINARY)
```
th0ma7, line 65: Is that test even needed?  I did not bother with that for ffmpeg5-6-7, maybe worth either using it everywhere or I would suggest not at all.

cross/libvpx/Makefile
```diff
@@ -28,40 +30,33 @@ include ../../mk/spksrc.archs.mk
 
 ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS)),$(ARCH))
 CONFIGURE_ARGS += --target=generic-gnu
-ADDITIONAL_CPPFLAGS = -std=c++0x
 ADDITIONAL_CXXFLAGS = -std=c++0x
 endif
 
 ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
 CONFIGURE_ARGS += --target=armv7-linux-gcc
-CONFIGURE_ARGS += --disable-neon
```
th0ma7, line 37: Disabling of neon was done in this PR https://github.com/SynoCommunity/spksrc/pull/5620
This was the comit: https://github.com/SynoCommunity/spksrc/pull/5620/commits/6ad7f5f927d86a7c56c5c7402614d33e3d642545

So from the commit log it states disabling neon (i.e. which includes `as` optimizations) on armv7 when using a compiler <= 4.9, basically on dsm6.  I suggest the following change:
```

ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
CONFIGURE_ARGS += --target=armv7-linux-gcc
ifeq ($(call version_lt, ${TC_GCC}, 5),1)
CONFIGURE_ARGS += --disable-neon
endif
endif

```

cross/libvpx/Makefile
th0ma7: This change comes from this PR: https://github.com/SynoCommunity/spksrc/pull/6148
Specifically from this commit: https://github.com/SynoCommunity/spksrc/pull/6148/commits/c4c0d44f0066f7ff0fffb1516e35f7d2f9032260

But I haven't left any descriptions as to why I disabled it?  Maybe it failed on DSM6, again using gcc <= 4.9.x for that specific libvpx version (i.e 1.14.1)?

Looking a bit further this may have been a mistake from my part as SSE4.1 (Streaming SIMD Extensions 4.1) has nothing to do with newer AVX extensions and may have mixed them somehow.  A bit more reading on it makes it that pretty much all "recent" intel based nas from synology (i.e. less than 10-12 years old) should support this extension...

I believe it is safe to remove that.

cross/openh264/Makefile
```diff
@@ -15,14 +15,10 @@ LICENSE  = BSD
 UNSUPPORTED_ARCHS = $(PPC_ARCHS)
 
 CONFIGURE_ARGS += -Dtests=disabled
+ADDITIONAL_CPPFLAGS = -O
```
th0ma7, line 18: Why? is it to diable optimizations (i.e. `-O0`)

cross/x264/Makefile
```diff
@@ -3,56 +3,52 @@ PKG_VERS = 20240512-stable
...
-ENV += x264_ARCH=ARM
-CONFIGURE_ARGS += --host=aarch64-linux
+# Define x86asm or use gcc as assembler
+ifeq ($(findstring $(ARCH),$(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
+NASM_BINARY = $(shell which nasm)
+ifeq ($(NASM_BINARY),)
+$(error nasm not found. Please install NASM assembler)
+endif
+ENV += AS=$(NASM_BINARY)
+else
```
th0ma7, line 35: Similarly to ffmpeg, isn't this too much as nasm is always installed?
If anything, I'd rather have that test being done outside from individual cross but within the framework to avoid code duplication if feasible.

mk/spksrc.cross-cmake.mk
```diff
@@ -114,14 +114,6 @@ $(CMAKE_TOOLCHAIN_PKG):
...
 	@cat $(CMAKE_TOOLCHAIN_WRK) ; \
 	echo
-ifeq ($(strip $(CMAKE_USE_NASM)),1)
-ifeq ($(findstring $(ARCH),$(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
-	@echo "# set assembly compiler" ; \
-	echo "set(ENABLE_ASSEMBLY $(ENABLE_ASSEMBLY))" ; \
-	echo "set(CMAKE_ASM_COMPILER $(CMAKE_ASM_COMPILER))" ; \
-	echo
-endif
-endif
```
th0ma7, line 124: This is rather interesting... indeed cmake should auto-detect `nasm` as being in default PATH.

th0ma7 (changes requested):

th0ma7:
> @th0ma7 the libraries and *.mk files are ready for review.

I did a first pass, overall looking good.  However I believe it does require minor adjustments for armv7 with gcc <= 4.9.x pertaining to assembly optimizations (i.e. neon).

> On the ffmpeg* packages there is some pending work...
> 
> 1. OLD_PPC_ARCHS fail to build ffmpeg since 4.4.x (4.3.x are the last working releases, newer have issue with stdatomic)

I would suggest removing `OLD_PPC_ARCHS` from all ffmpeg makefiles and declaring it as unsupported.

> 2. There are some conditional dependencies for `alpine` `comcerto2k` and `monaco` archs.
>    Except for `arch-comcerto2k-7.1` we do not build additional packages for `alpine` and `monaco`
>    If we would build dedicated packages for `alpine` and `monaco` we would need to exclude those archs in the `armv7` generic packages.
>    Despite we have an option to exclude dedicated archs from generic ones with `REMOVE_FROM_GENERIC_ARCHS`, we cannot use github build to create and publish dedicated archs.
>    And we never planned to created a single package for a list of dedicated archs...

Wow, that brings me a while back... when I took over the ffmpeg maintenance (which I didn't new what I was up against), I thought we absolutely needed arch specific packages for acceleration purposes... Not totally wrong but unsustainable as we a) don't have the manpower to support that (and not sure of the benefits) and b) don't have the storage space to do so.

Honestly this can be cleaned-up, it's just crap left-over.  Good catch thoughg :shrug:

th0ma7:
> and the third issue is with `cross/lame`:
> 
> even we configure with `--enable-asm` (and configure finds /usr/bin/nasm) it does not use the assembler.

I tried enabling the following options to check if nasm would be used somewhere, but nothing (btw, nevertless, mp3rtp could be enabled with the addition of `bin/mp3rtp` to the PLIST):
```
CONFIGURE_ARGS  = --enable-mp3rtp
CONFIGURE_ARGS += --enable-shared
CONFIGURE_ARGS += --enable-dynamic-frontends
```
Looking at arch linux and gentoo they both enable nasm...?  Even default debian rules files enables it.  btw `lame-3.100/debian/rules` uses the following:
```
                --with-fileio=sndfile \
                --enable-nasm \
                --with-pic \
                --disable-mp3x \
                --disable-mp3rtp \
                --disable-gtktest \
                --enable-dynamic-frontends \
                --enable-expopt=full
```
Tried them all, nothing shows up from a nasm usage ... So I tried compiling on the host to see what hapens, and it does enter into `libmp3lame/i386` where the assembly files are but does nothing:
```
Making all in libmp3lame
make[2]: Entering directory '/home/spksrc/python-wheels4/spksrc/distrib/lame-3.100/libmp3lame'
Making all in i386 
make[3]: Entering directory '/home/spksrc/python-wheels4/spksrc/distrib/lame-3.100/libmp3lame/i386'
make[3]: Nothing to be done for 'all'.
make[3]: Leaving directory '/home/spksrc/python-wheels4/spksrc/distrib/lame-3.100/libmp3lame/i386'
```
Honestly, I don't believe it ever worked or I'm missing something?

hgy59:
> Tried them all, nothing shows up from a nasm usage ... So I tried compiling on the host to see what hapens, and it does enter into `libmp3lame/i386` where the assembly files are but does nothing:

@th0ma7 I have investigated further for `cross/lame`, see [72cdfb9](https://github.com/SynoCommunity/spksrc/pull/6263/commits/72cdfb96c21abe24e9c11d9a44c4ddca122461b8)
it looks like a bug, or it is limitted to `i386` - the configure file has hardcoded `ASM_FOR_ARCH="i386"` and the only asm files (*.nas) are in the folder libmp3lame/i386 
latest update was 2017 and it seems it was never ported to support nasm for 64-bit cpu...

cross/openh264/Makefile
hgy59: > Why? is it to diable optimizations (i.e. `-O0`)

no `-O` is the same as `-O1` = optimization for code size and execution time

a kind of _default_ optimization.

my favorite for cross packages is `-O2`


<html><body>
<h2>gcc -O option flag</h2>
		<p>Set the compiler's optimization level.</p>
		
option | optimization level | execution time | code size | memory usage | compile time
-- | -- | -- | -- | -- | --
-O0 | optimization for compilation time (default) | + | + | - | -
-O1 or -O | optimization for code size and execution time | - | - | + | +
-O2 | optimization more for code size and execution time | -- |   | + | ++
-O3 | optimization more for code size and execution time | --- |   | + | +++
-Os | optimization for code size |   | -- |   | ++
-Ofast | O3 with fast none accurate math calculations | --- |   | + | +++


		<p>+increase ++increase more +++increase even more -reduce --reduce more 
		---reduce even more</p>
</body>
</html>

cross/x264/Makefile
hgy59: `cross/x264/Makefile` is the only file that needs nasm binary to define `AS`.

hgy59:
@th0ma7 I guess it is time now to remove native/nasm,
https://fossies.org/linux/misc/nasm-2.16.03.tar.xz is not available anymore.

th0ma7:
Yeah, i noticed that with the ffmpeg upgrade...

Issue #6297, opened 2024-10-30
rust builds for qoriq do not work anymore


hgy59:
Since some days the build of qoriq with (our own built) rust toolchain does not work anymore

There are errors like this:
```
2024-10-19T08:13:37.9481043Z    Compiling libc v0.2.161
2024-10-19T08:13:37.9481692Z    Compiling shlex v1.3.0
2024-10-19T08:13:37.9482436Z    Compiling proc-macro2 v1.0.88
2024-10-19T08:13:37.9496362Z    Compiling unicode-ident v1.0.13
2024-10-19T08:13:37.9650369Z error: the `-Z unstable-options` flag must also be passed to enable the flag `check-cfg`
2024-10-19T08:13:37.9651859Z error: the `-Z unstable-options` flag must also be passed to enable the flag `check-cfg`
2024-10-19T08:13:37.9653251Z error: the `-Z unstable-options` flag must also be passed to enable the flag `check-cfg`
2024-10-19T08:13:37.9654018Z 
2024-10-19T08:13:37.9654025Z 
2024-10-19T08:13:37.9655932Z 
2024-10-19T08:13:37.9662166Z error: the `-Z unstable-options` flag must also be passed to enable the flag `check-cfg`
2024-10-19T08:13:37.9663504Z 
2024-10-19T08:13:37.9672032Z error: could not compile `libc` (build script)
2024-10-19T08:13:37.9672877Z warning: build failed, waiting for other jobs to finish...
2024-10-19T08:13:37.9683523Z error: could not compile `proc-macro2` (build script)
2024-10-19T08:13:37.9685652Z error: could not compile `shlex` (lib)
2024-10-19T08:13:37.9687186Z error: could not compile `unicode-ident` (lib)
2024-10-19T08:13:37.9965512Z error: failed to compile `bat v0.24.0 (/github/workspace/spk/synocli-file/work-qoriq-6.2.4/bat-0.24.0)`, intermediate artifacts can be found at `/github/workspace/spk/synocli-file/work-qoriq-6.2.4/bat-0.24.0/target`.
2024-10-19T08:13:37.9967368Z To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

hgy59:
@th0ma7 is there any chance we get this fixed - or shall we drop all packages/tools depending on rust for qoriq models?

hgy59:
[...]

PR #6307, opened 2024-11-05
rust-qoriq: Update to 1.82.0 rust toolchain

th0ma7:
## Description

rust-qoriq: Update generation and use of 1.82.0 rust toolchain

Fixes https://github.com/SynoCommunity/spksrc/issues/6297

Versions available: https://releases.rs/docs/

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/cross/python310/Makefile
+++ b/cross/python310/Makefile
@@ -144,8 +144,8 @@ CROSSENV_WHEELS  = cffi==1.17.0
 # It also breaks numpy at crossenv creation time using github-action
 # https://numpy.org/devdocs/reference/distutils_status_migration.html
 # Python 3.12 considerations: scikit-build-core, meson-python
-CROSSENV_WHEELS += Cython==3.0.2
-#CROSSENV_WHEELS += Cython==0.29.37
+#CROSSENV_WHEELS += Cython==3.0.2
+CROSSENV_WHEELS += Cython==0.29.37
 CROSSENV_WHEELS += flit==3.9.0
 CROSSENV_WHEELS += scikit-build==0.18.1
 CROSSENV_WHEELS += setuptools-rust==1.10.2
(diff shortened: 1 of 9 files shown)


th0ma7:
@hgy59 I haven't tested, but that may also work on 1.82/83 but fails on nightly being next targetted version (i.e. 1.84).  Tested is 1.80.1 and ability to build `cross/bat` afterwards.

Todo, package this resulting rust toolchain + publish on line to then update the `toolchain/syno-qoriq-6.2.4-rust`.

hgy59:
@th0ma7 I did (almost) the same yesterday 😆 

small differences:
Instead of hard coding the version, I named the variable `RUST_BUILD_TAG` instead of `RUSTUP_DEFAULT_TOOLCHAIN` and made it optional instead of hard coded (and I prefere the long version of git arguments here).

mk/spksrc.rust-env.mk:
under `@$(MSG) "Building Tier-3 rust target: $(RUST_TARGET)"`
```
ifneq ($(RUST_BUILD_TAG),)
	@$(MSG) "Checkout rust tag $(RUST_BUILD_TAG)"
	@(cd $(WORK_DIR) && [ ! -d rust ] && git clone --branch $(RUST_BUILD_TAG) --depth 1  https://github.com/rust-lang/rust.git)
else        
	@$(MSG) "Build latests release (current HEAD)"
	@(cd $(WORK_DIR) && [ ! -d rust ] && git clone --depth 1 https://github.com/rust-lang/rust.git || true)
endif
```

so to create a toolchain call (in `toolchain/syno-qoriq-6.2.4-rust`):
```
make RUST_BUILD_TOOLCHAIN=1 RUST_TARGET=1.80.1
```

The name of the variable is subject to change, but I dislike the `DEFAULT` part in your version (RUSTUP_DEFAULT_TOOLCHAIN).
If we must set RUSTUP_DEFAULT_TOOLCHAIN for qoriq for other reason, it should not be hard coded in `mk/spksrc.cross-rust-env.mk`
We have to find a way to use the same value as `PKG_VERS` defined in `toolchain/syno-qoriq-6.2.4-rust/Makefile`.

----

Further I would like to include the toolchain version in the final archive name to omit the need of a release tag for every version and create all further releases under something like https://github.com/SynoCommunity/spksrc/releases/download/toolchains/rust

But my final tests did not work and I planned to validate the whole cycle with a plain environment (later).
_And I didn't want create a PR before everything works as expected..._

Open questions:
- is it possible to name the qoriq toolchain like `1.80.1-stage2_powerpc-unknown-linux-gnuspe` (or even omit the stage2 part) or must it be called exactly `stage2_powerpc-unknown-linux-gnuspe`?
- can we name the archive with the version like `1.80.1-powerpc-unknown-linux-gnuspe.tar.xz` independent of the answer to the question above?

th0ma7:
@hgy59 I've updated the version to 1.82.0 and it works well.  I've also implemented pretty much all of your suggested changes.  `toolchain/syno-qoriq-6.2.4-rust/Makefile` is obviously untested and will probably require a bit more love once you upload an updated qoriq rust package online.

Let me know what you think of the changes (and feel free to chime in additional subtilities you feel would be worth it).

Lastly, reminder, we need to change back `RUST_BUILD_TOOLCHAIN = 0` before merging.

hgy59:
@th0ma7 congrats: nice work, and good news!

everything works as expected.

successfully built all diyspk of tools created with rust (bandwhich bat bottom dua dutree eza fd lsd procs ripgrep sd pngquant) for qoriq-6.2.4.

and python311 including all wheels succeeded.

for python310 I had to adjust setuptools for numpy and to add crossenv/bin to the path for cryptography and other wheels using maturin.

only ffsync (i.e. syncstorage-rs) failed to build for qoriq-6.2.4 (so I left it unsupported)


PS
github failed when creating the tag `toolchains/rust` (500) so I had to create a different tag (`rust/qoriq`).
_probably a conflict with the existing tag `toolchains/rust/1.77.0-nightly`_

hgy59:
Lastly, reminder, we need to change back `WHEELS_TEST_ALL = 0` in spk/python310/Makefile before merging.

hgy59:
@th0ma7 as long as non of the packages (tools) require a different rust toolchain version, this solution works.

So far we need a different version only for the hi3535 arch, this is not a problem either.

Only when different tools require different toolchains for the same arch, this will be a problem.
Hopefully this will never apply for qoirq...

hgy59 (approved):
LGTM

th0ma7:
> @th0ma7 as long as non of the packages (tools) require a different rust toolchain version, this solution works.
> 
> So far we need a different version only for the hi3535 arch, this is not a problem either.
> 
> Only when different tools require different toolchains for the same arch, this will be a problem. Hopefully this will never apply for qoirq...

I have a feeling that armv5 and armv7l may require a home build as well before long... wait to be seen.  But as long as it works we're good!

PR #6288, opened 2024-10-18
synocli-file: update tools and add pixz

hgy59:
## Description
- update detox to v2.0.0
- update eza v0.20.7
- update fd to v10.2.0
- update fdupes to v2.3.2 and add sqlite for optional caches
- update fzf to v0.56.0
- update jdupes to v1.28.0 and libjodycode to v3.1.1
- update less to v668
- update lsd to v1.1.5
- update lzip to v1.24.1
- update mc to v4.8.32
- update micro to v2.0.14
- update nano to v8.2
- update pcre2 to v10.44
- update rhash to v1.4.5
- update ripgrep to v14.1.1
- update tree to v2.1.3

Fixes #6270

Additionaly add pixz from #6149
- Add pixz

Closes #6149, closes #6075, closes #6073

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #6312, opened 2024-11-09
local.mk: add option to disable github api call for maintainer

hgy59:
## Description

- spksrc.spk.mk uses the github api to get the realname and the url of the package maintainer
- if you build packages too often (e.g. small diyspk packages) you can reach the api rate limit and the build fails
- for this case you can define DISABLE_GITHUB_MAINTAINER with any non empty value (1, true, ...) to avoid the api call
- if DISABLE_GITHUB_MAINTAINER is defined the MAINTAINER is used as defined and the MAINTERNER_URL is set to "https://github.com/SynoCommunity" if not defined in the package

### Type of change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,6 +170,9 @@ local.mk:
 	@echo "DISTRIBUTOR_URL =" >> $@
 	@echo "REPORT_URL =" >> $@
 	@echo "DEFAULT_TC =" >> $@
+	@echo "# Option to disable the use of github API to get the real name and url of the maintainer" >> $@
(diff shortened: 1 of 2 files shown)


hgy59:
@th0ma7 I faced this issue when updating a lot of tools of synocli-file in #6288.

see [github: Rate limits for the REST API](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28)

th0ma7 (approved):
LGTM

PR #6329, opened 2024-11-23
Update libxml2 and libxslt

th0ma7:
## Description

- libxml2: Update libxml2 from 2.9.14 to 2.12.9 using cmake & github
- libxslt: Update from 1.1.39 to 1.1.42 and use cmake + github

Fixes: Current download source is failing and dating.

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [x] Package update


hgy59:
@th0ma7 there was a temp hack to redirect downloads from `download.gnome.org/sources` to `cdimage.debian.org/mirror/gnome.org/sources` in #6288.

now this PR shows that synocli-devel and synocli-net fail to download glib-2.78.4.tar.xz from gnome.org.

the reason for this hack was an invalid mirror used to download from gnome.org.

you could try to revert the redirection in spksrc.download.mk...

th0ma7:
@hgy59 ah! that explains!  Still, I'll use that as opportunity to update glib as well while at it.... and remove the redirect.  Thnx.

hgy59:
> @hgy59 ah! that explains! Still, I'll use that as opportunity to update glib as well while at it.... and remove the redirect. Thnx.

the mirror error occurred mostly on older glib versions and not on latest, so this seems not an option.

th0ma7:
> the mirror error occurred mostly on older glib versions and not on latest, so this seems not an option.

Ths github source looks good and stable... I'll work on using that as source instead.

hgy59:
BTW can we remove the xml and xslt tools? I can't imagine a package requires those.

in PLIST remove xmlcatalog and xmllint for libxml and xsltproc for libxsl.
there might be even an option to disable the build of such tools...

th0ma7:
humm... no fun https://gitlab.gnome.org/GNOME/glib/-/issues/2716#note_1534454
So github source doesn't have the `gvdb` submodule which requires additional steps.

Let's see if removing the download.gnome.org redirect suffice.

hgy59:
> Let's see if removing the download.gnome.org redirect suffice.

you find additional info in https://github.com/SynoCommunity/spksrc/pull/6288#issuecomment-2464702998

The faulty mirror “mirror2.sandyriver.net” is chosen randomly, so successful downloads are not proof that it will not be used another time.

PR #6340, opened 2024-11-30
cops, bicbucstriim: update to depend on DSM 7.2

### LINKED ISSUE #6339: COPS package must be restricted to DSM 7.2
COPS v3.2.2-11 was published for DSM 7.0 but it can't be installed on DSM 7.1.1 (lacking PHP 8.2)

The framework must be adjusted to create (and publish) noarch packages for DSM 7.2 when defined as `REQUIRED_MIN_DSM = 7.2`

spk/COPS/Makefile still defines `REQUIRED_MIN_DSM = 7.0` that must be adjusted, but the issue is, that `REQUIRED_MIN_DSM = 7.2` is not handled as expected yet.

hgy59:
## Description

noarch packages depending on PHP 8.2 require DSM 7.2 and packages must include this dependency.
So far we have those packages:
- cops
- bicbucstriim

bugfixes:
- handle REQUIRED_MIN_DSM = 7.2 for noarch packages
- add option to include noarch-7.2 packages in build workflow

Fixes #6339

## Checklist

### Type of change

- [x] Bug fix


mreid-tt:
@hgy59, guess this can be modified to also treat with BicBucStriim?

mreid-tt:
@hgy59 the build failure is because the source file was changed. It now has the suffix of '-php82' in the filename so the whole file is now called: cops-3.2.2-php82.zip. Adjusting this in the cross/ should fix it.

hgy59:
> @hgy59 the build failure is because the source file was changed. It now has the suffix of '-php82' in the filename so the whole file is now called: cops-3.2.2-php82.zip. Adjusting this in the cross/ should fix it.

already found this for bicbucstriim but for cops I had the old one in the distrib folder...

hgy59:
> @hgy59, guess this can be modified to also treat with BicBucStriim?

already done locally, but want to validate the new build matrix first.

mreid-tt (approved):

hgy59:
Both packages published for DSM 7.2.

Existing packages with PHP 8.2 and DSM 7.0 deactivated new (to allow manual installation, for the rare case a user wants to install cops or bicbucstriim on DSM 7.2 before those are visible in Package Center).

hgy59:
@mreid-tt there is a small issue in regular builds, when changed packages require DSM 7.2.
In such a case the arch specific dsm72 build are added to the matrix, even when only noarch packages have changed.
And probably dsm72 noarch packages are added, when only arch specific packages have changed.

For manually triggered builds this is handled by separated options.
For the regular builds there is only one flag yet. The respective evaluation in prepare.sh must be enhanced.

mreid-tt:
@hgy59 a possible solution would be to add checks for the presence of packages by adding these lines:
```
          arch_packages=${{ needs.prepare.outputs.arch_packages }}
          noarch_packages=${{ needs.prepare.outputs.noarch_packages }}
```
right after this line:
https://github.com/SynoCommunity/spksrc/blob/e145e7206dbad83e7d91a97c8d7ea7f9c834a9fc/.github/workflows/build.yml#L172

Then, you could modify this line:
https://github.com/SynoCommunity/spksrc/blob/e145e7206dbad83e7d91a97c8d7ea7f9c834a9fc/.github/workflows/build.yml#L186

to say something like this:
```
          if [ "$add_dsm72_builds" == "true" ] && [ -n "$arch_packages" ]; then
```

This could be extended to do a similar check for the other conditionals like:

```
          if [ "$add_dsm71_builds" == "true" ] && [ -n "$arch_packages" ]; then
```
```
          if [ "$add_dsm62_builds" == "true" ] && [ -n "$arch_packages" ]; then
```
```
          if [ "$add_dsm52_builds" == "true" ] && [ -n "$arch_packages" ]; then
```
As well as earlier:
```
          if [ "$add_noarch_builds" == "true" ] && [ -n "$noarch_packages" ]; then
```

PR #6356, opened 2024-12-23
SABnzbd 4.4.1

Safihre:
## Description

Update SABnzbd to 4.4.1

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


mreid-tt:
hey @Safihre, I noted in the release notes that "_macOS and Windows: Use Python 3.13 for release_". Would this be a requirement for Linux as well in the near future?

Safihre:
> hey @Safihre, I noted in the release notes that "_macOS and Windows: Use Python 3.13 for release_". Would this be a requirement for Linux as well in the near future?

No, we support all version in maintained by Python itself, so 3.9 and above.

Safihre:
@mreid-tt @th0ma7 Unrar needs an update, but it's beyond my very limited makefile knowledge.
I gave it a shot, but failed: https://github.com/Safihre/spksrc/actions
Additionally, for 7zip we still use p7zip which hasn't been updated since 2016. 
The regular 7zip also now supports linux, so we could switch to that. But again, beyond my knowledge. https://7-zip.org/download.html

I'd be happy to donate ($200?) to anyone who can do it for me, if anyone is open to that 🙂🎄

hgy59:
@Safihre  we can update unrar to v7.1.2, but it will support DSM 7 only (excl. comcerto2k).

And I have succesfully built 7zz of 7zip v24.09 for DSM 7.1 (excl. comcerto2k, excl. armv7).

The limitation to DSM 7.0 (unrar) and DSM 7.1 (7zip) comes from compiler version and AES HW accel. functions.
The build of 7zip for ARMv7-7.1 will probably need a patch.

mreid-tt:
@hgy59, what would that mean for SABnzbd? Would versions for the older DSM releases just use the older dependencies or would that limit building only for the new DSM releases?

Safihre:
Ohh, in that case we might want to stick to the older versions for a while..
I don't think we want to drop DSM 6 yet, right?

mreid-tt:
Well, no we do have some `cross/` packages which will build different versions depending on the DSM release required for compatibility reasons. I was just trying to confirm if we would be taking a similar approach with `unrar` and `7zz` or if there was some other constraint which would limit things to DSM 7.

Safihre:
Or we could make them into each their own spk. This way they can first have a DSM 6 compatible release, and all successive ones are DSM 7 only.

PR #6282, opened 2024-10-14
dynamic crossenv + python3*-wheels + python310-313 updates

th0ma7:
## Description

Intent is to:
1. Remove wheel build testing from `python310`/`python311` base `Makefiles` to ease updates
2. Add `python312` and `python313` packages
3. Dissociate `crossenv` creation from `python3*/Makefile` -> Now uses `spksrc.crossenv.mk`
```
$ ls -la ~/spksrc/mk/crossenv/requirements-default.txt
$ make crossenvclean
$ make crossenv-<arch>-<version>
```
4. Allow wheel specific `crossenv` enablement using a `mk/crossenv/` directory containing wheel crossenv definitions:
```
$ ls -la ~/spksrc/mk/crossenv/requirements-aiohttp-3.8.5.txt
$ make crossenvclean
$ WHEEL="aiohttp-3.8.5" make crossenv-x64-7.1
```

Fixes #6284

## Checklist

- [x] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [x] Package update
- [x] Includes <s>small</s> framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

### TODO
- [x] Move usage of pip cache through `$HOME/.cache/pip` to use `$(WORK_DIR)/pip`
```
2024-11-14T02:37:23.1044954Z WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
```
- [x] Fix handling of `requirements.txt` entry with trailing comment `test==1234   # This is a test wheel`
- [x] Have crossenv wheels use #$(WORK_DIR) caching directory instead of default `$(HOME)/.cache/pip`
- [x] Review `OPENSSL_*_DIR` variables and logic usage throughout python related mk files
- [x] Use status cookie for crossenv creation
- [ ] Use status cookie for wheel building -->> LEFT FOR SUBSEQUENT PR
- [x] Have `crossenv` add status info to `status-build.log` file
[...]

PR #6389, opened 2025-01-13
wheel.mk: Migrate to using status cookie

th0ma7:
## Description

Migrate to using status cookie

Follow-up to https://github.com/SynoCommunity/spksrc/pull/6282, in particular https://github.com/SynoCommunity/spksrc/pull/6282#issuecomment-2584475292

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes <s>small</s> framework changes

### TODO
- [x] Allow using `WHEELS="<name>-<version>" make wheel-<arch>-<tcversion>`
- [x] Manage end-of-line comments in requirements files
- [x] Allow downloading to `distrib/pip` all types (`abi3`, `crossenv`, `pure`)
- [x] Allow managing `#egg=` portion on URL based wheel requirements (for newer `pip` compatibility)
- [x] Create new `download-wheels` sub-routines to pre-download only
- [x] Allow downloading wheels part of the github-action preparation phase
- [x] Update python wheel wiki documentation

### DIFF
--- a/mk/spksrc.wheel-compile.mk
+++ b/mk/spksrc.wheel-compile.mk
@@ -1,0 +1,148 @@
+### Wheel rules
+# Compile wheels for modules listed in WHEELS. 
+#
+# Targets are executed in the following order:
+#  wheel_compile_msg_target
+#  pre_wheel_compile_target   (override with PRE_WHEEL_COMPILE_TARGET)
+#  wheel_compile_target       (override with WHEEL_COMPILE_TARGET)
+#  post_wheel_compile_target  (override with POST_WHEEL_COMPILE_TARGET)
+# Variables:
+#  REQUIREMENT             Requirement formatted wheel information
+#  WHEEL_NAME              Name of wheel to process
+#  WHEEL_VERSION           Version of wheel to process (can be empty)
+#  WHEEL_TYPE              Type of wheel to process (abi3, crossenv, pure)
+
+ifeq ($(WHEEL_VERSION),)
+WHEEL_COMPILE_COOKIE = $(WORK_DIR)/.$(COOKIE_PREFIX)wheel_compile-$(WHEEL_NAME)_done
+else
+WHEEL_COMPILE_COOKIE = $(WORK_DIR)/.$(COOKIE_PREFIX)wheel_compile-$(WHEEL_NAME)-$(WHEEL_VERSION)_done
+endif
+
+##
+
+ifeq ($(strip $(PRE_WHEEL_COMPILE_TARGET)),)
+PRE_WHEEL_COMPILE_TARGET = pre_wheel_compile_target
+else
+$(PRE_WHEEL_COMPILE_TARGET): wheel_compile_msg_target
+endif
+ifeq ($(strip $(WHEEL_COMPILE_TARGET)),)
+WHEEL_COMPILE_TARGET = wheel_compile_target
+else
+$(WHEEL_COMPILE_TARGET): $(BUILD_WHEEL_COMPILE_TARGET)
+endif
+ifeq ($(strip $(POST_WHEEL_COMPILE_TARGET)),)
+POST_WHEEL_COMPILE_TARGET = post_wheel_compile_target
+else
+$(POST_WHEEL_COMPILE_TARGET): $(WHEEL_COMPILE_TARGET)
+endif
+
+wheel_compile_msg_target:
+	@$(MSG) "Processing wheels of $(NAME)"
+
+pre_wheel_compile_target: wheel_compile_msg_target
+
+wheel_compile_target: SHELL:=/bin/bash
+wheel_compile_target:
+ifeq ($(wildcard $(WHEELHOUSE)),)
+	@$(MSG) Creating wheelhouse directory: $(WHEELHOUSE)
+	@mkdir -p $(WHEELHOUSE)
+endif
+	@$(MSG) Compiling wheel [$(WHEEL_NAME)], version [$(WHEEL_VERSION)], type [$(WHEEL_TYPE)]
+ifneq ($(WHEEL_TYPE),pure)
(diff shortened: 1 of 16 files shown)


th0ma7:
Functionality for `wheel_download` now migrated to status cookie, next to be `wheel_compile` then `wheel_install` which shouldn't be too long by the look of things.

hgy59:
@th0ma7 it would be convenient having a `wheelclean-%` target to rebuilt a specific wheel like:
```
make wheelclean-numpy-1.26.4
```

just my 5 cents 🪙

th0ma7:
@hgy59 `wheelclean-<wheel>-<version>` functionality added.  Also the code is now fully migrated to using status cookies for all wheel builds which should have a significant impact on build time (if not from a github-action it will from a debugging and testing perspective).

If you have a moment to test things up would be great.  Similarly to `crossenv` I've now added a `make wheel-<arch>-<tcversion>` call.  It can be extended to:
```
make WHEELS="<name>-<version>" wheel-<arch>-<tcversion>
```
This will add that "wheel" to the list of wheels to process.  Noting that `Makefile` first `WHEELS` entry, if using `=` instead of `+=` will discard argument passed in parameter.  Although for the sake of testing, commenting out `WHEELS` entries from your makefile will allow you to test-build a wheel you are struggling with.

I'm sure there will be a few rough-edges and missing is the ability to download only for using with github-action (next on my TODO).

mk/spksrc.wheel-compile.mk
```diff
@@ -0,0 +1,148 @@
...
+# Compile wheels for modules listed in WHEELS. 
+#
+# Targets are executed in the following order:
+#  wheel_compile_msg_target
+#  pre_wheel_compile_target   (override with PRE_WHEEL_COMPILE_TARGET)
+#  wheel_compile_target       (override with WHEEL_COMPILE_TARGET)
+#  post_wheel_compile_target  (override with POST_WHEEL_COMPILE_TARGET)
+# Variables:
+#  REQUIREMENT             Requirement formatted wheel information
+#  WHEEL_NAME              Name of wheel to process
```
hgy59, line 11: how to document, that WHEEL_NAME must not contain `-`?
Since name and version are separated by `-` the name must not contain `-`.

can we add a check here 
or fail if name has `-` instead of `_` 
or implement an auto renaming?

or can we still use `=` as version separator? (can't remeber which make file is affected).

mk/spksrc.wheel-download.mk
```diff
@@ -0,0 +1,92 @@
...
+	@$(MSG) Creating pip caching directory: $(PIP_CACHE_DIR)
+	@mkdir -p $(PIP_CACHE_DIR)
+endif
+	@$(MSG) Downloading wheel [$(WHEEL_NAME)], version [$(WHEEL_VERSION)] ; \
+#	$(MSG) requirement: [$(REQUIREMENT)] ; \
+#	$(MSG) requirement-grep-egg: [$$(grep -s egg <<< $(REQUIREMENT))] ; \
+#	$(MSG) name: [$(WHEEL_NAME)] ; \
+#	$(MSG) type: [$(WHEEL_TYPE)] ; \
+#	$(MSG) version: [$(WHEEL_VERSION)] ; \
+#	$(MSG) type: [$(WHEEL_TYPE)] ; \
```
hgy59, line 60: If this does not work, it might all be commented out (`\` wraps all on a single line, and all following the first # is comment)

_just a guess_

mk/spksrc.wheel-compile.mk
th0ma7: Unless I don't understand your comment: 
`WHEEL_NAME` really is just the wheel name.
`WHEEL_VERSION` really is just the version
`REQUIREMENT` is `WHEEL_NAME==WHEEL_VERSION` most of the time, although sometime it can be a `git+https://` type url with a `#egg=WHEEL_NAME==WHEL_VERSION`.

Although it did made me think of something else... said parameter is "WHEELS" similarly to variable in usage in our makefile, which is fine.  But for the crossenv I'm using "WHEEL" with an expectation that it takes a `<name>-<version>` - version being optional in this case. 

So, potentially:
1. allow using `-` in WHEELS?
2. make crossenv use the same variable name? or something completely different because `WHEEL` and `WHEELS` could be confusing^

hgy59: OK, found the related code in `spksrc.crossenv.mk`
lines 62 and 63:
```
else ifneq ($(wildcard $(CROSSENV_CONFIG_PATH)/requirements-$(firstword $(subst -, ,$(WHEEL))).txt),)
CROSSENV_BUILD_WHEEL = $(firstword $(subst -, ,$(WHEEL)))
```

The expression ` $(firstword $(subst -, ,$(WHEEL)))` does not match the wheel name, when it contains `-`, but it works when `_` is used instead.

we could use `=` or `==` to separate name and version in `WHEEL`-variables or we ensure that we split name and version at the most right `- `.
But even we do this, the wheel specific cross-env definition file name must match the wheel name in the requirements files of the packages.

th0ma7: I suggest the following to simplify things: with this new code `spksrc.wheel.mk` will always manage `WHEEL_NAME` and `WHEEL_VERSION` variables.  I suggest replacing the confusing `WHEEL` only used with the newly introduces dynamic `crossenv` code for using those two variables instead.

I'll update the code for this.  Also this way both wheel and crossenv will be aligned.

mk/spksrc.wheel-download.mk
th0ma7: yeah, well, i intend to remove that prior to merging.  These are left over debug traces.

mk/spksrc.wheel-compile.mk
th0ma7: code now adapted for this.

mk/spksrc.wheel-download.mk
th0ma7: now removed

th0ma7:
I believe this is now ready for merging (conditional to github-action green flags + disabling package bumped verions)

PR #6398, opened 2025-01-17
vim: update to v9.1

hgy59:
## Description

Fixes # 

## Checklist

### Type of change

- [x] Package update


hgy59:
@thoma7 There are several reasons to drop the python feature in vim:

- with the redesing of python packages, the python feature was already missing in `Vim v9.0-6` package
- because it is a build time dependecy only, the python package build does not work (configure --with-python is ignored, because libpython*.so is not found)
- it would need further effort to use the python feature in vim:
   - it must use python of the python311 package
   - but python in path is python 3.8 of Synology
   - we once had a start script for vim (vim82) to start vim of the SynoCommunity package with UTF8 locale in the environment, but this didn't work as expected and I don't want to reinvent it to set custom path when installed python311 is available
- nobody missed the python feature so far (I guess it was just my idea to implement all possible features)

PR #6406, opened 2025-01-18
ffsync: Update syncstorage-rs to v0.18.2

mreid-tt:
## Description

1. Update syncstorage-rs to v0.18.2

Fixes # 

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/cross/syncstorage-rs/Makefile
+++ b/cross/syncstorage-rs/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = syncstorage-rs
-PKG_VERS = 0.17.15
+PKG_VERS = 0.18.2
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/mozilla-services/$(PKG_NAME)/archive/${PKG_VERS}
--- a/cross/syncstorage-rs/digests
+++ b/cross/syncstorage-rs/digests
@@ -1,3 +1,3 @@
-syncstorage-rs-0.17.15.tar.gz SHA1 8ad56f7cf5d26bfcb8f02160bb420ea815e614cb
-syncstorage-rs-0.17.15.tar.gz SHA256 eee17d5b69b59415d83bf3957e3f92da1a4eb0e443e74c39ecac9292be00943f
-syncstorage-rs-0.17.15.tar.gz MD5 f1b5a9a54a54d0aefcc9dd472e3a7143
+syncstorage-rs-0.18.2.tar.gz SHA1 b5eda7e8d2168ece628336d2a2e321916a048c41
+syncstorage-rs-0.18.2.tar.gz SHA256 901c3ef6296ba36300f44d56317769c81c87f756ab0f8e5c2ebb7049135a5f4c
+syncstorage-rs-0.18.2.tar.gz MD5 fc6e691358a468ab1f4f2567ea78b1cf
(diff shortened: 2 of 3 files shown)


hgy59:
I can confirm that this is realted to python packages.

I reproduced the publish error with ffsync, bazarr and python311 packages.

But I guess it is not (only) related to #6389. This PR updated only mk/*.mk files (exchept download.sh for github workflow).
When I took all *.mk file in the version before #6389, the error still exists.

So it might be caused by #6282

@th0ma7 published python310 on 10.01.2025 and that was just about 3 minutes after #6282 was merged (if the timestamp in github and sc repo are the same timezone).
If python310 was published before the merge, then the publish was broken by #6282.
Otherwise we need further investigation.
It might be that python310 is not broken, but python311 is...

hgy59:
@th0ma7 just found that the publish for python310 is still working
The publish is broken for python311 and related packages.

th0ma7:
I did not yet had a chance to respond but this is really odd as the difference between 310 and 311's makefiles are marginal?

I'll be over probably tomorrow with my pr that fixes builds issues, perhaps this will be magically addressed by it?

Still, what i don't get is why my pr do run properly then?  Could it be a caching issue?

hgy59:
@th0ma7 just found the root cause

The publish target uses `http` to upload packages.
`http` is a python script located at `/usr/bin/http`

while publishing, python the variable `PYTHONPATH` is set.
when I remove the definition of `PYTHONPATH` in `spksrc.crossenv.mk` the publish works.

otherwise the call of `http` generates this error:
```
  File "/usr/bin/http", line 33, in <module>
    sys.exit(load_entry_point('httpie==3.2.1', 'console_scripts', 'http')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/httpie/__main__.py", line 8, in main
    from httpie.core import main
  File "/usr/lib/python3/dist-packages/httpie/core.py", line 15, in <module>
    from .client import collect_messages
  File "/usr/lib/python3/dist-packages/httpie/client.py", line 22, in <module>
    from .ssl_ import AVAILABLE_SSL_VERSION_ARG_MAPPING, HTTPieCertificate, HTTPieHTTPSAdapter
  File "/usr/lib/python3/dist-packages/httpie/ssl_.py", line 1, in <module>
    import ssl
  File "/usr/lib/python3.11/ssl.py", line 100, in <module>
    import _ssl             # if we can't import it, let the error propagate
    ^^^^^^^^^^^
```

with python311:
```
PYTHONPATH='/spksrc/spk/python311/work-x64-7.1/Python-3.11.11/build/lib.linux-x86_64-3.11:/spksrc/spk/python311/work-x64-7.1/install/var/packages/python311/target/lib/python3.11/site-packages/'
```
with python310:
```
export PYTHONPATH='/spksrc/spk/python310/work-x64-7.1/Python-3.10.16/build/lib.linux-x86_64-3.10:/spksrc/spk/python310/work-x64-7.1/install/var/packages/python310/target/lib/python3.10/site-packages/'
```

So we must ensure that python3 on the host is used to publish packages with http.

hgy59:
> Still, what i don't get is why my pr do run properly then? Could it be a caching issue?

Building always succeeded, only publish is broken.
IMHO publishing will (must) always be a manual task.

mreid-tt:
@hgy59, @th0ma7, appreciate the quick turnaround on this. The package has now been successfully published.

PR #6403, opened 2025-01-18
FlexGet: Update to v3.13.16

mreid-tt:
## Description

This PR includes the following:

1. Update FlexGet to version 3.13.16

Fixes #6395

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

Issue #5043, opened 2022-01-03
META: Python 3.10 / 3.11 package status


th0ma7:
# Python 3.10 package status

## High-level python PR
- [x] python310 (3.10.1) #5044
- [ ] pre-download wheels #4971
- [x] mysqlclient wheel #5068
- [x] avoid error for package without wheel #5092
- [x] python310 (3.10.2) #5101
- [x] #5169
- [x] python310 (3.10.4) #5171
- [x] #5175
- [x] Refactor crossenv spksrc.wheel.mk #5180
- [x] rust wheel support & update to python 3.10.8 #5435

## Packages formally using python 3.x
| PACKAGE | Python 3.10 migration | Published | Python 3.11 migration | Published 3.11 |COMMENT |
|----------------|-------------------------------|--------------|--------------|--------------|-----------------|
| bazarr | ✔️ | ✔️ | | | Follow-up update #5139 (@smaarn) | 
| beets | ✔️ | ✔️ | | | #4946 |
| borgbackup | ✔️ | ✔️ | | | #5048 (co-author @sunlei)<br>Fixed: https://github.com/borgbackup/borg/issues/6105 |
| domoticz |     |      | | |  Version online uses `python38` | 
| duplicity | ✔️ | ✔️ | | | #5054 |
| fishnet | ❌️ | ❌️ | | | @mayweed Rewritten in Rust (removal #5216) |
| flexget | ✔️ | ✔️ | | | #5176 |
| homeassistant | ✔️ | ✔️ | | | #5579, #5096, ARMv5 support pending: https://github.com/numpy/numpy/issues/20664<br>Looks like working based on testing https://github.com/SynoCommunity/spksrc/pull/5095/commits/fbd98191653d2138140c697454d41e82edbd64ab<br>Confirmed, fixed https://github.com/SynoCommunity/spksrc/pull/5155|
| mercurial | ✔️ | ✔️ | | | #5045 |
| octoprint | ✔️ | ✔️ | | | #5217 |
| plexpy-custom | ✔️ |  | | | Ready for migration but BUG with the package install, to remove? |
| rdiff-backup | ✔️ | ✔️ |  | | #5447 |
| rutorrent | ✔️ | ✔️ | | | Migration to DSM7 #5455, #5669, #5617 |
| sabnzbd | ✔️ | ✔️ | ✔️ |✔️ | #5109 (@Safihre)<br>#5137<br>#5527 |
[...]

PR #6409, opened 2025-01-19
wheel building: Pile of small fixes to recent changes

th0ma7:
## Description

Fixes: 
- [x] https://github.com/SynoCommunity/spksrc/pull/6403#issuecomment-2600908263
- [x] `python-wheel.mk` fix for maturin built based wheels
- [x] standardise `crossenv` calls
- [x] standardise `PATH` definitions and call to `cross-python3`
- [x] updates to `python311`
- [x] updates to `python312` and `python312-wheels`
- [x] updates to `python313` and create `python313-wheels`
- [x] does not halt on errors while preparing the `crossenv`
- [x] gcc-4.9.x archs requires `msgpack==1.0.5` or it fails to install the crossenv.
- [x] force break when crossenv or wheel build fails
- [x] manage compile, install, plist cookies for `spksrc.python-wheel.mk`
- [x] allow building newer `pillow` versions

Follows-up:
- https://github.com/SynoCommunity/spksrc/pull/6282
- https://github.com/SynoCommunity/spksrc/pull/6389

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update


hgy59:
The build of pendulum as python wheel succeeded locally for armv7-7.1 and aarch64-7.1 so far.

th0ma7:
@mreid-tt and @hgy59 FYI I have found a false positive whereas compilation would continue although errors may have been trigerred.  This is due to `-` in front of `-@MAKEFLAGS= $(MAKE) ...` which makes it to ignore errors.

Currently working on figuring out exactly a) how to ensure it does halt on errors and b) where the errors are.  At first glance it looks relatively easy to solve and limited in scope.  I'll keep you posted once solved (fix to be in this PR).

th0ma7:
I may have now solved the issue where wheel builds should halt properly on errors.  This brought light to other wheel build issues such as with `pydantic-core`, now also fixed (had to migrate from pip using requirements file to `python-wheel.mk` using cross).

Remaining there may still be a weird case involving `numpy` with `python311` only... next on my todo list.  Beyond that I believe with this PR that code is becoming quite stable (fingers-crossed).

th0ma7:
@hgy59 This PR is now ready for review.  Many small changes and fixed many small'ish long-standing issues.  Been testing it for a few days and cannot find anything left.

th0ma7:
@hgy59 and @mreid-tt with the completion of the PR I believe time ma be appropriate to initiate work on migrating to `python313`.  We'd need an issue to track updates like done previously https://github.com/SynoCommunity/spksrc/issues/5043 and I would suggest releasing easy ones first to shake out any remaining issues.  Thoughts?

th0ma7:
pending: `python313-wheels` requires a bit more love and got a fix pending for `pillow` with older gcc <= 4.9.x

mreid-tt:
I'm guessing that based on the [cadence](https://devguide.python.org/versions/) of Python development it would be prudent to go to 3.13. I agree with you that the finalisation of `python313-wheels` would be very useful for migration of existing packages. Your wheels for 3.11 was very helpful for me when I attempted a package update recently.

th0ma7:
There's also a need to be strategic with regards to impacts... 

python311:
- newer `pillow` now finally compile!!!  Although gcc < 4.9 (armv5 and probably hi3535, to be confirmed) no longer works unless refering back to version 9.x of `pillow` as workaround for older platforms.
 
python312:
- newer `numpy` finally compile to the detriment that older compatible with gcc-4.9.x no longer works (now DSM7 only)
- impacted packages: `bazarr`, `homeassistant`

python313:
- `greenlet` only work with newer gcc (now DSM7 only)
- impacted packages: `bazarr`, `ffsync`, `flexget`, `homeassistant`, `sickchill`

Migration directly to `python313` would be nice, but more python packages will stop supporting DSM6.  While `python312` would be less impactful.

cross/pillow/Makefile
```diff
@@ -0,0 +1,34 @@
...
+PKG_DIST_SITE = https://files.pythonhosted.org/packages/source/p/pillow
+PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
+
+DEPENDS  = cross/freetype cross/libimagequant cross/libjpeg
+DEPENDS += cross/libtiff cross/libwebp cross/zlib
+
+HOMEPAGE = https://github.com/pyca/cryptography
+COMMENT  = Provide cryptographic recipes and primitives to Python developers
+LICENSE  = BSD/ASL2
+
```
hgy59, line 14: copied from cross/cryptography?
```suggestion
HOMEPAGE = https://python-pillow.github.io/
COMMENT  = Python Imaging Library (Fork).
LICENSE  = MIT-CMU

```

hgy59:
we have some python-wheel packages with (empty) PLIST file and some without PLIST.
since there is no warning printed about missing PLIST file we can safly remove those.
And we have some "version selector" Makefiles (using spksrc.main-depends.mk) that have no PLIST files but a warning ""No PLIST for $(NAME)"" is printed out.

we can cleanup this in another PR:
- remove empty PLIST files for spksrc.python-wheel.mk packages
- suppress the warning "NO PLIST for ..." for main.depends packages.

hgy59:
@th0ma7 still an issue with wheel status cookies.

When I run make on a package after a successful build, or after successful build and running `make spkclean`, I do not expect the `===>  Processing wheel for {package-name}` and all the `===>  Processing requirement [...]` logs.

It should only create the INFO file, the scripts, wizard, ... (files outside package.tgz) and finally update the spk package.
Similar to skipping the build of all the dependencies when the cookie `.depend_done` exists, it must not trigger the wheel building in this case. I guess we are lacking a cookie like `.wheel_done`.

Otherwise it takes too long and overwhelms with logs, when working on the service-setup or on the wizard files for packages that successfully build so far....

Remarks:
It does not build the wheels again and it logs `Nothing to be done for 'wheel'.` for all the wheels, but it should not trigger those targets.

th0ma7:
indeed, i had that previously but it was a pain to play with when wanting to redo/debug/test some other wheels.  But that is dam easy to bring back, further what your suggesting is changing the wheel level so it can be cleaned-up at `spkclean` which is clever.

It doesn't avoid python migration to occur though, and I'll submit a PR for that in no time, easy to tackle.

mreid-tt:
hey @th0ma7, looking forward to this PR. Looking at the build errors it seems that based on this https://github.com/python-greenlet/greenlet/issues/392#issuecomment-2446961814, they are caused by:

> That's because you're trying to install greenlet < 3.1, which doesn't support Python 3.13. You must use greenlet >= 3.1.x.

As such, I believe if you comment out this line your build should succeed: https://github.com/SynoCommunity/spksrc/blob/17ece54326e21633ea101ba73b12b8b6d848a26a/spk/python313-wheels/src/requirements-crossenv-greenlet-v3.txt#L9

PR #6411, opened 2025-01-21
Fix dependency list

hgy59:
## Description

follow up to #6255
not all dependencies were found.
also the former implementation with make dependecy-list did not find all dependencies

### Remarks
`cross/berkeley` is such a dependency that was lacking in the dependency list (it was lacking in the pre #6255 version too).

Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/mk/dependency-list.sh
+++ b/mk/dependency-list.sh
@@ -91,7 +91,7 @@ function contains ()
 function get_dependencies ()
 {
    local dependencies=
-   for dep in "${1}"; do
+   for dep in ${1}; do
       dependencies+="${dep} "
       dependencies+="$(get_file_dependencies ${dep}) "
    done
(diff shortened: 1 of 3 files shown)


mk/dependency-list.sh
```diff
@@ -106,7 +106,7 @@ function get_spk_dependencies ()
    local cumulated_dependencies=$(get_dependencies "${dependencies}")
    while [ "${cumulated_dependencies}" != "${dependencies}" ]; do
       dependencies=${cumulated_dependencies}
-      cumulated_dependencies=$(get_dependencies "${dependencies}")
+      cumulated_dependencies=$(get_dependencies "${dependencies}" | tr ' ' '\n' | sort -u | tr '\n' ' ')
```
smaarn, line 109: Naive question: isn't this already done by the `get_dependencies` function ?

hgy59: yes for recursive calls it is, but not im `get_file_dependencies` when python and or ffmpeg dependencies are added.

(maybe the _unique sort filter_ should be added at line 106 too)

hgy59: @smaarn your question was justified. 
This is now updated, and the test for python and ffmpeg packages is optimized to evaluate on top level (spk folder) only.

PR #6418, opened 2025-01-25
fix publish

hgy59:
## Description

related to https://github.com/SynoCommunity/spksrc/pull/6406#issuecomment-2613837511

This is a hotfix since publish python311 and related packages is broken

- avoid PYTHONPATH definition for http tool
- PYTHONPATH is defined spksrc.crossenv.mk and was introduced with #6282

Fixes # 

## Checklist

### Type of change

### DIFF
--- a/mk/spksrc.publish.mk
+++ b/mk/spksrc.publish.mk
@@ -57,7 +57,7 @@ endif
 ifeq ($(PUBLISH_API_KEY),)
 	$(error Set PUBLISH_API_KEY in local.mk)
 endif
-	@response=$$(http --verify=no --ignore-stdin --auth $(PUBLISH_API_KEY): POST $(PUBLISH_URL)/packages @$(SPK_FILE_NAME) --print=hb) ; \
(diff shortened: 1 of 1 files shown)


hgy59:
@th0ma7 I guess we can't omit the defintion of `PYTHONPATH` in `spksrc.crossenv.mk`

since the publish target depends on the build targets, IMHO the overwrite of the variable when using `http` command is a reasonable solution.

hgy59:
@th0ma7 the `http` python script is part of `httpie` that is installed in the development environment.

We might use an alternate tool like curl.

We use the following python based apps in the environment.
can you please veryfy whether none of the others is affected by the `PYTHONPATH` variable?
```
RUN apt install --no-install-recommends -y \
	httpie \
	mercurial \
	meson \
	ninja-build \
	python3 \
	python3-distutils \
	python3-mako \
	python3-pip \
	python3-virtualenv \
	python3-yaml
```

candidates are mercurial, meson and ninja-build when used for non python dependencies but run in a python dependent package.

hgy59:
> We might use an alternate tool like curl.

almost 10 years ago the download tool was changed from curl to http/httpie in https://github.com/SynoCommunity/spksrc/commit/abcf15306adddda85b5657f757d874a210ed45d2

th0ma7 (approved):
Nice catch!  Lgtm

PR #6422, opened 2025-01-26
mercurial: Update hg from 6.5 to version 6.9.1 and python 3.12

th0ma7:
## Description

mercurial: Update hg from 6.5 to version 7.0 and python 3.12

Fixes n/a

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6428, opened 2025-01-26
Add wheel top-level status cookie and mark armv5 as unsupported >= python 3.12

th0ma7:
## Description

Add wheel top-level status cookie and mark armv5 as unsupported >= python 3.12

Relates to comments:
- Mark armv5 as unsupported: https://github.com/SynoCommunity/spksrc/pull/6422#issuecomment-2614601077
- Add top-level wheel status cookie https://github.com/SynoCommunity/spksrc/pull/6409#issuecomment-2614492794

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.python.mk
+++ b/mk/spksrc.python.mk
@@ -7,7 +7,12 @@
 # set default spk/python* path to use
 PYTHON_PACKAGE_WORK_DIR = $(realpath $(CURDIR)/../../spk/$(PYTHON_PACKAGE)/work-$(ARCH)-$(TCVERSION))
 
-include ../../mk/spksrc.archs.mk
+include ../../mk/spksrc.common.mk
+
+# armv5 no longer supported with python >= 3.12
+ifeq ($(call version_ge, $(subst python,,$(PYTHON_PACKAGE)), 312), 1)
+UNSUPPORTED_ARCHS += $(ARMv5_ARCHS)
+endif
 
 ifneq ($(wildcard $(PYTHON_PACKAGE_WORK_DIR)),)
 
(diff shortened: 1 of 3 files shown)


th0ma7:
@hgy59 I believe this would be addressing your two concerns.   Would you mind giving it a shot?  thnx.

mreid-tt:
@th0ma7, with the `mk/spksrc.python.mk` now including `mk/spksrc.common.mk`, I am guessing that we shouldn't have Makefiles with both anymore?

For example in my current #6431: [spk/sabnzbd/Makefile](https://github.com/SynoCommunity/spksrc/blob/24eefbd21362affa5032e697b5cf68a2e6b0900f/spk/sabnzbd/Makefile)

th0ma7:
Yes and no... It's a question of timing as if you require a call to version_* for instance you'll need to source common.mk prior to it, or calling it after python.mk

mreid-tt:
Okay, I just moved the include call earlier in the file and it seems to be okay. Here's the updated [spk/sabnzbd/Makefile](https://github.com/SynoCommunity/spksrc/blob/ce6e4b165f54f00c0aeb06ab4fc34d45bcaadc66/spk/sabnzbd/Makefile).

th0ma7:
Rule of thumbs include of python.mk similarly to cross-cc.mk or spk.mk should always go at the end.

Only locally declared recipes would go after.

You could place call to version* after it as well if it's a one liner but i suggest calling common.mk right before instead.

Most important, all *TARGET déclarations must go above it as otherwise it becomes a redeclaration of an already existing recipe.

Exemple :
```
POST_INSTALL_TARGET = bla

include spksrc.python.mk

.PHONY: bla
bla:
     Myrecipe abcd
```

PR #6437, opened 2025-02-02
Review content of crossenv to respect build vs cross.

th0ma7:
## Description

Review content of crossenv to respect build vs cross.
- [x] spksrc.requirement.mk: Unified method to process requirements files
- [x] Migrate pure python wheels to build using crossenv/build instead of native
- [x] Manage distinct installation of `crossenv/build` and `crossenv/cross`
- [x] Duplicate log output under `crossenv|wheel-<arch>-<tcversion>.log`
- [x] Upgrade to python 3.12.9
- [x] Upgrade to python 3.13.2

Closes #6430

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6449, opened 2025-02-12
cross/libsodium: update to v1.0.20

mreid-tt:
## Description

1. Update to libsodium v1.0.20

Fixes # 

## Checklist

### Type of change

- [x] Library update

### DIFF
--- a/cross/dtv-scan-tables/Makefile
+++ b/cross/dtv-scan-tables/Makefile
@@ -1,9 +1,11 @@
 PKG_NAME = dtv-scan-tables
-PKG_VERS = n/a
+# There are no releases available. We take the current master in the github repository.
+# PKG_VERS: date of the latest commit, adjust it when master got updates.
+PKG_VERS = 2025.02.11
 PKG_EXT = tar.gz
 PKG_DIST_NAME = master.$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/crazycat69/dtv-scan-tables/archive
-PKG_DIST_FILE = $(PKG_NAME).$(PKG_EXT)
+PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIR = $(PKG_NAME)-master
 
 DEPENDS =
@@ -12,6 +14,6 @@ HOMEPAGE = https://github.com/crazycat69/dtv-scan-tables
 COMMENT  = Digital TV scan tables.
 LICENSE  = GPLv2
 
-INSTALL_MAKE_OPTIONS = DATADIR=$(INSTALL_DIR)/$(INSTALL_PREFIX)/share/tvheadend/data DVBV5DIR=dvb-scan install
+INSTALL_MAKE_OPTIONS = DATADIR=$(STAGING_INSTALL_PREFIX)/share/tvheadend/data DVBV5DIR=dvb-scan install
 
 include ../../mk/spksrc.install-resources.mk
--- a/cross/dtv-scan-tables/PLIST
+++ b/cross/dtv-scan-tables/PLIST
@@ -1,0 +1,1 @@
+rsc:share/tvheadend/data/dvb-scan/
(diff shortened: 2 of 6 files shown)


hgy59:
@mreid-tt you forgot to update the PLIST file

mreid-tt:
@hgy59, I added this yesterday because the source repo was down, and I noticed they also had a GitHub source. However, I later realized that the GitHub source doesn’t always include a "stable" version for all releases, unlike the original repo. Now that the source repo is back up, I’m considering canceling this — what do you think?

hgy59:
> the GitHub source doesn’t always include a "stable" version for all releases

If the original source repo has the update too, I recommend to proceed with this PR. If you don't know the exact version for PLIST (I know, you are lacking a dev env), I can evaluate those for you.

had the same finding for `cross/libxml2` so switched back to original sources (while downgrading in #6440).

mreid-tt:
> If the original source repo has the update too, I recommend to proceed with this PR. If you don't know the exact version for PLIST (I know, you are lacking a dev env), I can evaluate those for you.

I've updated the changes based on the original source, incorporating their new naming convention with an explicit "stable" in the filename. If you could assist with updating the PLIST, I'd appreciate it.

hgy59:
> I've updated the changes based on the original source, incorporating their new naming convention with an explicit "stable" in the filename. If you could assist with updating the PLIST, I'd appreciate it.

Just curious why you didn't take the stable release anymore.

there is a release 1.20.0 from 25-May-2024 12:22  
and a release 1.20.0-stable from 16-Jan-2025 13:42

mreid-tt:
> Just curious why you didn't take the stable release anymore.

I thought I was. When you extract `libsodium-1.0.19.tar.gz` you get a folder `libsodium-stable` but when you extract `libsodium-1.0.20.tar.gz` you get a folder `libsodium-1.0.20`. However when you extract `libsodium-1.0.20-stable.tar.gz` you get a folder `libsodium-stable`. As such, I thought they changed their naming. But if the file without the name "stable" in the title is in fact the stable version then you would need to update the `PKG_DIR`.

mreid-tt:
@hgy59, I think I got an understanding of this from this https://github.com/jedisct1/libsodium/issues/1373#issuecomment-2135172301:

> Recently, a new cycle started with the RELEASE-1.0.20 tag. stable now represents version 1.0.20 with minor fixes and improvements (but no breaking changes nor new features). And since a new cycle started, tag 1.0.19 represents the last stable 1.0.19 version, identical to the last 1.0.19-stable tarball. If your application requires 1.0.19 and can't use 1.0.20, you can use this instead of 1.0.19-RELEASE. It will, for example, ensure that it can be compiled with recent compiler versions while 1.0.19 as originally release can't.
>
> You can stick to the -RELEASE tags if you want. Or use stable to always get the latest release with backported fixes and improvements before the next point release.

It looks like I had it backward. The filename without "stable" refers to the first release of that version, while the one with "stable" represents the latest version within that point release. If that’s the case, our digests will periodically fail as the upstream team updates their fixes. How do you feel about using the latest version moving forward, or should we stick with the initial release for better stability in our build system?

hgy59:
> It looks like I had it backward. The filename without "stable" refers to the first release of that version, while the one with "stable" represents the latest version within that point release. If that’s the case, our digests will periodically fail as the upstream team updates their fixes. How do you feel about using the latest version moving forward, or should we stick with the initial release for better stability in our build system?

@mreid-tt you are right (and the comment in the referenced issue is absolutly corret: ***The moving tags (used like a branch) are quite confusing, and difficult to carry through different trackers.***)

It is very bad practice to release *-stable versions that get continues updates without changing the version number (If that is the case here).

So we go better back to the real _stable_ version (i.e. the one that does not change) the one without `stable` in the name.

@mreid-tt sorry for not digging into the details yesterday.

hgy59 (approved):
LGTM

mreid-tt:
@hgy59, thanks for the approval. Before merging, I wanted to check — since we're sticking with the initial release on this repo, should I re-implement the source switch to GitHub? The first release is the identical file, and using GitHub as the source might help reduce potential build failures. Let me know your thoughts.

hgy59:
@mreid-tt I have cancelled the workflow since it already succeeded on the branch (and we have a known issue since #6437)

To answer your latest question:
we could have switched back to github repo, but we do not know, how future releases are provided. So far we just should avoid the so called "stable" versions of libsodium...

PR #6453, opened 2025-02-15
homeassistant: update to v2024.12.5 and python312

hgy59:
## Description

- homeassistant is now limited to DSM 7 (all ARMv7 are not supported)
- add cross/dtlssocket_ha and cross/pillow_ha with HA specific versions
- update homeassistant to v2024.12.5  (originally planned 2025.1.4, but this version has a mandatory issue: it forces encrypted backups that can only be used for restore).
- update HACS to v2.0.5
- update python wheels (requirements)
- remove postinst_components_requirements.txt as those requirements are installed on demand
- add requirements-cross_from_index.txt to install cross built wheels from index
- add python312 custom crossenv requirements for propcach, pymicro_vad and pyseex_noise
- avoid building pure-python wheels and download at installation time only

Fixes #6332, fixes #6008, closes #5849

## Remarks

- Homeassistant with Python 3.12 is supported on DSM 7 only (ARMv7 models are not supported anymore).
- 2025.1.4 is the latest version that runs with Python 3.12. An update to v2025.1.5 is not planned and further updates will require Python 3.13 (the system log shows a related warning)
- wheels that can't be built with spksrc framework are downloaded from the index at package installation (numpy, av and others)
- DSM 7 with x64, aarch64 and i686 (evansport) are supported

## TODO:
- [x] include numpy==2.1.3 (mandatory requirement, not available in index for evansport and ARMv7)
- [ ] document limited functionality for cross wheels not available in index for aarch64, evansport, ARMv7

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #6454, opened 2025-02-16
[WIP] Handle meson based python-wheel

th0ma7:
## Description

Handle meson based python-wheel

Fixes n/a (add reference to numpy+ comments here)

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


th0ma7:
Note that this is not ready for testing yet... still early WIP, simply ported in a new PR my pending changes from my local branch.

hgy59:
additional info:

1. Supprted DSM

   building for comcerto2k-7.1 shows error:
   `meson.build:28:4: ERROR: Problem encountered: NumPy requires GCC >= 8.4`

   DSM < 7 will not be supported and we have to add to numpy Makefile something like:

   ```
   # numpy requires gcc >= 8.4
   REQUIRED_MIN_DSM = 7.0
   UNSUPPORTED_ARCHS = comcerto2k
   ```

2. cython not found
   when adding this to cross/numpy/Makefile:
   `ENV += PATH=$(WORK_DIR)/crossenv-default/build/bin:$(PATH)`
   it can find cython and sucessfully builds the wheel for x64

   it still fails for evansport, aarch64 and armv7:
   IMHO it should not use native/python312 but python in the build crossenv
   and additionally it must use python header files of cross python
   currently evansport and armv7 have errors like
    `/spksrc/native/python312/work-native/install/usr/local/include/python3.12/pyport.h:586:2: error: #error "LONG_BIT definition appears wrong for platform (bad gcc/glibc config?)." ` 
    (`pyport.h` does no match the 32-bit target platform)
    and aarch64 fails with:
    `../numpy/_core/src/umath/loops_autovec.dispatch.c.src:107:43: internal compiler error: Segmentation fault`

th0ma7:
> additional info:
> 
> 1. Supprted DSM
>    building for comcerto2k-7.1 shows error:
>    `meson.build:28:4: ERROR: Problem encountered: NumPy requires GCC >= 8.4`
>    DSM < 7 will not be supported and we have to add to numpy Makefile something like:
>    ```
>    # numpy requires gcc >= 8.4
>    REQUIRED_MIN_DSM = 7.0
>    UNSUPPORTED_ARCHS = comcerto2k
>    ```

Yup, on my radar, will add that indeed.  Although I wonder if numpy 1.26.x might still work with older DSM considering the new meson I'm trying to build-up.

> 2. cython not found
>    when adding this to cross/numpy/Makefile:
>    `ENV += PATH=$(WORK_DIR)/crossenv-default/build/bin:$(PATH)`
>    it can find cython and sucessfully builds the wheel for x64
>    it still fails for evansport, aarch64 and armv7:
>    IMHO it should not use native/python312 but python in the build crossenv

That should be the case from `spksrc.python-module-meson.mk`... Although up to now I was still on the per depend fully-generated meson cross-file (which looks like working now).  So I haven't reconvene on that part just yet.

>    and additionally it must use python header files of cross python

The per depend fully-generated meson cross-file should now fix that.  I had encountered that same issue with cmake long ago where the `$(WORK_DIR)/tc-vars.cmake` would be use in conjunction of per depend `C|CPP|CXX|LDFLAGS`.  Issue was that:
1. The generic `$(WORK_DIR)/tc-vars.cmake` cannot handle per depends `*FLAGS` as there may be additional ones defined using `ADDITIONAL_*FLAGS` variables in any of the dependencies in `cross/<meh>`.  And cmake cross file does not handle ${var} of that kind.
2. So to fix that when cmake is called it fetches the content of the generic `$(WORK_DIR)/tc-vars.cmake` and create a `$(WORK_DIR)/$(PKG_DIR)/<arch>.cmake` file and add all FLAGS information to it so it is fully complete and no longer dependent on env flags.
3. To that effect, with a fully defined per dependency cmake cross file the build process is called with an empty environment similarly as native.  This ensures that the host is set to the targeted architecture while the build remains unchanged (i.e. the docker arch we're building on which gets auto-detected).

Long story short, the meson had never received such enhancement as things we're working just fine.  But that's no longer true with python wheels whereas with the python virtual environment things gets totally confused for meson.  Thus the need to have a fully functional meson cross-file defining all library and include path properly.

have a look at current `tc-vars.cmake|meson` and you'll notice no flags being set.  When invoking a cmake build you'll have a fully compliant cross-file within the corresponding `$(PKG_DIR)`.

This is now mostly working with meson, still have a few things to go through but getting there.

>    currently evansport and armv7 have errors like
>    `/spksrc/native/python312/work-native/install/usr/local/include/python3.12/pyport.h:586:2: error: #error "LONG_BIT definition appears wrong for platform (bad gcc/glibc config?)." `
>    (`pyport.h` does no match the 32-bit target platform)
>    and aarch64 fails with:
>    `../numpy/_core/src/umath/loops_autovec.dispatch.c.src:107:43: internal compiler error: Segmentation fault`

That's a current known issue with numpy.  We need to force settin the long bit accordingly for `big` or `little` endian into the meson cross-file such as:
```
[properties]
longdouble_format = 'IEEE_DOUBLE_BE'
```

th0ma7:
@hgy59 having a proof of concept that does build sucessfully for both aarch64 and x64 using latest numpy 2.2.3.

Although struggling with armv7 and evansport... I'll check if I can make 1.26.4 to work instead for the moment.

mreid-tt:
@th0ma7, was looking at the errors which remain:
```
error: #error "LONG_BIT definition appears wrong for platform (bad gcc/glibc config?)."
```
And found the following which may be useful if you've not already considered:
* [Building cython based module with numpy dependency in SDK](https://lists.yoctoproject.org/g/yocto/topic/building_cython_based_module/73974417)
* https://github.com/pyodide/pyodide/issues/4118

Hope they can assist...

th0ma7:
I believe I now have something functional, but unmaintainable as-is.

## The good
Using normal `python -m pip wheel ... ` and passing proper meson-python flags:
```
--config-settings=setup-args="--cross-file=$(MESON_CROSS_TOOLCHAIN_PKG)" \
--config-settings=setup-args="--native-file=$(MESON_NATIVE_FILE)" \
--config-settings=install-args="--tags=runtime,python-runtime" \
--config-settings=build-dir="$(MESON_BUILD_DIR)" \
--no-build-isolation
```
I can now sucesfully cross-compile for arm7, evansport and x64 for DSM-7.1

## The bad
There is a know bug in gcc<=10 with aarch64 that makes the compiler to segfault.  I tried pretty much every possible alternatives of flags/disabling in the code but I wasn't able to workaround it.
Bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97696
Potential workaround that I was not able to make use of: https://github.com/google/highway/pull/1932/files
@hgy59 maybe you can help on this which would be really nice to have a working patch.

## The ugly
Part of my crusade at making meson-python build to work, I ended-up at one point to reproduce the normal meson+ninja build.  Surprisingly, this ended-up allowing me to sucesfully build numpy for aarch64... Missing is then the (re)packaging in wheel format, which hapens to be the exact same process as "The good" as long as I re-use the exact same builddir (which I ended-up figuring out tonight).

This really is ugly, but does work.  This last commit https://github.com/SynoCommunity/spksrc/pull/6454/commits/a2068f5ef66ff654688e611fa974c73eafd56e3a was not tested on previously working x64, evansport and armv7.  I'll let this rest for tonight.  Good news is, we're probably much closer now... just need to tie-up the remaining loose-ends.

mk/spksrc.cross-meson-env.mk
```diff
@@ -65,3 +69,53 @@ ifeq ($(findstring $(ARCH),$(x64_ARCHS)),$(ARCH))
...
+meson_pkg_toolchain:
+	@cat $(MESON_CROSS_TOOLCHAIN_WRK)
+	@echo
+	@echo "[properties]" ; \
+	echo "needs_exe_wrapper = false"
+ifeq ($(findstring $(ARCH),$(ARMv8_ARCHS)),$(ARCH))
+	@echo "longdouble_format = 'IEEE_DOUBLE_BE'"
+else ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS) $(i686_ARCHS) $(x64_ARCHS)),$(ARCH))
+	@echo "longdouble_format = 'IEEE_DOUBLE_LE'"
+endif
```
hgy59, line 89: I suppose that all those archs have to use `IEEE_DOUBLE_LE` (i.e. liddle endian)
Only PPC_ARCHS have big endian byte ordering.

th0ma7: thnx, an oversight of me part of my crusade.  Code is not ready for prime use just yet.

hgy59:
@th0ma7 the wheels created from python/*/Makefile are not yet added to `wheelhouse/requirements.txt`.
And I guess it should be added to `wheelhouse/requirements-cross.txt` too.

The `wheelhouse/requirements.txt` is used by the `install_python_wheels` function in `spksrc.service.installer.functions`

th0ma7:
Thnx for catching this up, will include this.

I'm also looking at how to install numpy in the crossenv... I got an idea on how i could reuse the newly cross-compiled numpy wheel so it gets installed into the cross portion of the crossenv so it can then be made available for other wheels that depends on it.

Lastly, also looking at adding flexibility to have different vendor managed meson (other than numpy use case where the source package provides its own modified meson.py) and skipping that meson+ninja part when no vendor managed meson is provided (i.e. being the default use case)

All in all, taking shape but will require a few more spare cycles before reaching the finishing line...

hgy59:
@th0ma7 another small issue popped up:
The wheel for charset-normalizer is generated as `charset_normalizer-3.4.1-py3-none-any.whl` and also the WHEEL file shows, it is pure python:
```
Wheel-Version: 1.0
Generator: setuptools (75.8.0)
Root-Is-Purelib: true
Tag: py3-none-any
```

The original wheels in the index (pypi) are cross compiled (like `
charset_normalizer-3.4.1-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`).
maybe our crossenv build needs a flag or something to create a cross compiled wheel for charset-normalizer.

hgy59:
@th0ma7 I have successfully built python311-wheels with added python/numpy and python/numpy_1.26 for aarch64-7.1 and armv7-7.1.

It would be interresting to validate whether such wheels created with gcc 8.5 will run under DSM 6. I guess if the *.so files within the wheels do not reference GLBC > 2.20 functions, it might work.

My background: I am trying to build a final homeassistant package with support for DSM 6. This will be homeassistant 2024.3.3 that depends on numpy 1.26.0. This version is available in the index for x86_64 and aarch64 only, and I will have to build it at least for armv7 and evansport (i686).
To support armv7, I will have to cross build additionally `ha-av==10.1.1` (this is not so easy since it depends on ffmpeg libraries).

_To support armv7 in homeassistant 2025.1.4, it will be `av==13.1.0` that depends on ffmpeg libraries too_

th0ma7:
> @th0ma7 another small issue popped up:
> The wheel for charset-normalizer is generated as `charset_normalizer-3.4.1-py3-none-any.whl` and also the WHEEL file shows, it is pure python:
> ```
> Wheel-Version: 1.0
> Generator: setuptools (75.8.0)
> Root-Is-Purelib: true
> Tag: py3-none-any
> ```
> 
> The original wheels in the index (pypi) are cross compiled (like `
> charset_normalizer-3.4.1-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`).
> maybe our crossenv build needs a flag or something to create a cross compiled wheel for charset-normalizer.
> 

Maybe this is similar to msgpack where it can fit in both?

th0ma7:
I got some pretty cool code locally that allows installing cross-compiled numpy wheel into the crossenv to allow building scipy and others... I got one major major major problem, gcc version.

For `numpy`, with aarch64, it mandatory requires gcc>=10 otherwise it segfaults.  I was able to workaround that up until needing to include `OpenBLAS` which re-trigerred it under `numpy`.  And without it I cannot build `scipy`, thus I cannot build `scikit-learn`.

All in all, this would require bumping our minimal version to DSM-7.2.

th0ma7:
Good news, I was able to create a workaround patch for aarch64 ... a few loose ends but looking much better now.

th0ma7:
@hgy59 and @mreid-tt It may look like stagnating but after spending numerous hours on this I finally made a major leap forward which now allows using default `pip wheel` methodology to build meson type wheels.  Hopefully this will now allow me to include `lapack` and open the way to `scipy`, and thus things should start rolling at that point.

This has definitively been taking way longer than anticipated but I believe things will now start to shape nicely :crossed_fingers:

toolchain/syno-qoriq-6.2.4-rust/Makefile
```diff
@@ -2,7 +2,7 @@ PKG_ARCH = qoriq
 PKG_NAME = rust-$(PKG_ARCH)
 PKG_VERS = 1.82.0
 PKG_EXT = txz
-PKG_DIST_TARGET = $(PKG_VERS)-powerpc-unknown-linux-gnuspe
+PKG_DIST_TARGET = $(PKG_VERS)-powerpc-unknown-linux-gnuspe-spksrc
```
mreid-tt, line 5: I don't believe that this is the correct rename strategy:

1. The original filename didn't change so this will result in a file not found.
2. The issue is the downloaded file. Once downloaded it should be renamed to match the syntax of https://rust-lang.github.io/rustup/overrides.html#toolchain-file-settings. In theory, it needs a prefix before the version number (e.g. `stable|beta|nightly`)

hgy59: @th0ma7 what was the reason for this renaming?
- is it related to #6476?
- your rename can't work, because the PKG_DIST_TARGET is the name of the file to download
- but we could create a `rustup toolchain link` with a different name if this helps (see rust-qoriq-postinstall target in this file)

mreid-tt: @th0ma7, I've proposed a fix for this in #6477, and it seems to work with my test builds for `memcached`. Once I get approval, we can merge and rebase.

th0ma7: > @th0ma7 I have reverted your change since `powerpc-unknown-linux-gnuspe` (and `1.82.0-powerpc-unknown-linux-gnuspe`) is a valid rust toolchain name.

This was just a really quick test I had pushed online as I thought the issue was dead simple to fix (which looks like not).

th0ma7:
@hgy59 and @mreid-tt progress being made:
- [x] `numpy` (and thnx @hgy59 for your few additions to enable lapack from openblas)
- [x] `scipy`
- [ ] `scikit-learn` ... next.  Any other missing using meson-python that also depends on `numpy` + `scipy` ?

Afterwards, missing are:
- [ ] adding fortran to cmake portion of the framework
- [ ] cleanups & testing
Then things should be good for merging :)

hgy59:
@th0ma7 another wheel that requires numpy is pandas (HA 2025.1.4 depends on pandas==2.2.3)

`python312/crossenv/requirements-pandas.txt` would look like:
```
pip==24.3.1
setuptools==75.8.0
wheel==0.45.1
#
build:build==1.2.2
build:Cython==3.0.11
build:meson==1.2.1
build:meson-python==0.13.1
build:pip-tools==7.4.1
#build:versioneer[toml]
build:versioneer==0.29
#
wheelhouse:numpy==2.2.3
```

hgy59:
Some pending issues (maybe for another PR)
- When a requirement-file is missing in `WHEELS = missing-requirements.txt` it is silently ignored (I expect an error like `file ... not found`)
- If a package depends on cross compiled wheels (python/* or cross/*) and does not have other requirements (i.e. does not have `WHEELS = ...` defined) then the wheels are not added to the package (those are in wheelhouse but not in install/..../shared/wheelhouse, and whl names are not adjusted)
- when using package names like `numpy_v2` (in python/numpy/Makefile) a specificy crossenv is not used (I suggest that the file must be named requirements-numpy_v2.txt instead of requirements-numpy.txt ) If this is by intention it should be documented somewhere... or we need another solution to include multiple versions of such wheels

hgy59:
> @th0ma7 another wheel that requires numpy is pandas (HA 2025.1.4 depends on pandas==2.2.3)
> 
> `python312/crossenv/requirements-pandas.txt` would look like:
> 
> ```
> pip==24.3.1
> setuptools==75.8.0
> wheel==0.45.1
> #
> build:build==1.2.2
> build:Cython==3.0.11
> build:meson==1.2.1
> build:meson-python==0.13.1
> build:pip-tools==7.4.1
> #build:versioneer[toml]
> build:versioneer==0.29
> #
> wheelhouse:numpy==2.2.3
> ```

I tried this, and the crossenv is created as expected 🎉 -  but it fails to build:
_I use numpy==2.2.0 as required for HA_

```
===>  Package list for /spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/build:
Package            Version
------------------ -------
build              1.2.2
click              8.1.8
Cython             3.0.11
meson              1.2.1
meson-python       0.13.1
packaging          24.2
pip                23.2.1
pip-tools          7.4.1
pyproject_hooks    1.2.0
pyproject-metadata 0.9.0
setuptools         75.8.0
versioneer         0.29
wheel              0.45.1
===>  Package list for /spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/cross:
Package    Version
---------- -------
numpy      2.2.0
pip        23.2.1
setuptools 75.8.0
wheel      0.45.1
===>  Generating
env make --no-print-directory meson_pkg_toolchain > /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3/x64-toolchain.meson 2>/dev/null;
===>  Compiling for pandas
===>  Installing for pandas
find /spksrc/diyspk/pandas_ha/work-x64-7.1/install//var/packages/pandas-wheel/target/../ \! -type d -printf '%P\n' | sed 's?^target/??g' | sort > /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas.plist.tmp
===>  - crossenv: [/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/bin/activate]
===>  - meson: [/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/build/bin/meson]
===>  - python: [/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/bin/cross-python]
===>  _PYTHON_HOST_PLATFORM="x86_64-pc-linux-gnu" PATH=/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/cross/bin:/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/build/bin:/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/bin:/spksrc/distrib/cargo/bin:/home/.local/share/pnpm:/scripts:/home/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/games /spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/bin/cross-python -m pip wheel . --config-settings=setup-args="-Dbuildtype=release" --config-settings=setup-args="--cross-file=/spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3/x64-toolchain.meson" --config-settings=builddir="/spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3/builddir" --no-build-isolation --wheel-dir /spksrc/diyspk/pandas_ha/work-x64-7.1/wheelhouse --verbose
Processing /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3
  Preparing metadata (pyproject.toml): started
  Running command Preparing metadata (pyproject.toml)
  + meson setup /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3 /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3/builddir -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md --vsenv -Dbuildtype=release --cross-file=/spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3/x64-toolchain.meson --native-file=/spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3/builddir/meson-python-native-file.ini
  The Meson build system
  Version: 1.2.1
  Source dir: /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3
  Build dir: /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3/builddir
  Build type: cross build
  Project name: pandas
  Project version: 2.2.3
  C compiler for the host machine: /spksrc/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu-gcc (gcc 8.5.0 "x86_64-pc-linux-gnu-gcc (GCC) 8.5.0")
  C linker for the host machine: /spksrc/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu-gcc ld.bfd 2.30
  C++ compiler for the host machine: /spksrc/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu-g++ (gcc 8.5.0 "x86_64-pc-linux-gnu-g++ (GCC) 8.5.0")
  C++ linker for the host machine: /spksrc/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu-g++ ld.bfd 2.30
  Cython compiler for the host machine: cython (cython 3.0.11)
  C compiler for the build machine: cc (gcc 12.2.0 "cc (Debian 12.2.0-14) 12.2.0")
  C linker for the build machine: cc ld.bfd 2.40
  C++ compiler for the build machine: c++ (gcc 12.2.0 "c++ (Debian 12.2.0-14) 12.2.0")
  C++ linker for the build machine: c++ ld.bfd 2.40
  Cython compiler for the build machine: cython (cython 3.0.11)
  Build machine cpu family: x86_64
  Build machine cpu: x86_64
  Host machine cpu family: x86_64
  Host machine cpu: x86_64
  Target machine cpu family: x86_64
  Target machine cpu: x86_64
  Program python found: YES (/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/bin/cross-python)

  ../pandas/meson.build:1:15: ERROR: Command `/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/bin/cross-python -c '
  import os
  import numpy as np
  try:
      # Check if include directory is inside the pandas dir
      # e.g. a venv created inside the pandas dir
      # If so, convert it to a relative path
      incdir = os.path.relpath(np.get_include())
  except Exception:
      incdir = np.get_include()
  print(incdir)
       '` failed with status 1.

  A full log can be found at /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3/builddir/meson-logs/meson-log.txt
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> See above for output.

  note: This error originates from a subprocess, and is likely not a problem with pip.
  full command: /spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/cross/bin/python /spksrc/spk/python312/work-x64-7.1/install/var/packages/python312/target/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py prepare_metadata_for_build_wheel /tmp/tmp1bmpbgw8
  cwd: /spksrc/diyspk/pandas_ha/work-x64-7.1/pandas-2.2.3
  Preparing metadata (pyproject.toml): finished with status 'error'
error: metadata-generation-failed


× Encountered error while generating package metadata.
╰─> See above for output.

```

with further details in meson-log.txt:
```
We have compiled some common reasons and troubleshooting tips at:

    https://numpy.org/devdocs/user/troubleshooting-importerror.html

Please note and check the following:

  * The Python version is: Python3.12 from "/spksrc/diyspk/pandas_ha/work-x64-7.1/crossenv-pandas/cross/bin/python"
  * The NumPy version is: "2.2.0"

and make sure that they are the versions you expect.
Please carefully study the documentation linked above for further help.

Original error was: libgfortran.so.5: cannot open shared object file: No such file or directory

```

I guess this is an known issue (numpy is lacking fortran libraries) and numpy wheels are not yet fully working.

th0ma7:
I had a similar issue with `scipy` wheras I ended-up generating a second cross-file that I also pass.  Have a look at the post function in its makefile whereas it generates the following:
```
[properties]
numpy-include-dir = '/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-aarch64-7.1/crossenv-scipy/cross/lib/python3.12/site-packages/numpy/_core/include'
pythran-include-dir = '/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-aarch64-7.1/crossenv-scipy/build/lib/python3.12/site-packages/pythran'
build-python-path = '/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-aarch64-7.1/crossenv-scipy/bin/build-python'
build-python-version = '3.12'
host-python-path = '/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-aarch64-7.1/crossenv-scipy/bin/cross-python'
host-python-version = '3.12'
```

you may have something similar in panda's `meson.build` file whereas you can enforce giving the full path in that second meson cross-file so it ends-up skipping the loading tests that will never work out, as loading numpy from cross tries to load aarch64 binaries over x86_64

th0ma7:
Another thought, maybe you can explicitely add a skip-dependency-check to the `python -m pip` in `spksrc.python-wheel-meson.mk` and/or try replacing that for `python -m build` (and add build to cross in your crossenv) such as:
```
-          $$(which cross-python) -m pip wheel . \
-          $(foreach arg,$(CONFIGURE_ARGS),--config-settings=setup-args="$(arg)" ) \
-          $(foreach arg,$(INSTALL_ARGS),--config-settings=install-args="$(arg)" ) \
-          --config-settings=builddir="$(MESON_BUILD_DIR)" \
-          --no-build-isolation \
-          --wheel-dir $(WHEELHOUSE) \
+          $$(which cross-python) -m build -w -n -x . \
+          $(foreach arg,$(CONFIGURE_ARGS),-Csetup-args="$(arg)" ) \
+          $(foreach arg,$(INSTALL_ARGS),-Cinstall-args="$(arg)" ) \
+          -Cbuilddir="$(MESON_BUILD_DIR)" \
+          --outdir $(WHEELHOUSE) \
```

hgy59:
> I had a similar issue with `scipy` wheras I ended-up generating a second cross-file that I also pass. Have a look at the post function in its makefile whereas it generates the following:
> 
> ```
> [properties]
> numpy-include-dir = '/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-aarch64-7.1/crossenv-scipy/cross/lib/python3.12/site-packages/numpy/_core/include'
> pythran-include-dir = '/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-aarch64-7.1/crossenv-scipy/build/lib/python3.12/site-packages/pythran'
> build-python-path = '/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-aarch64-7.1/crossenv-scipy/bin/build-python'
> build-python-version = '3.12'
> host-python-path = '/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-aarch64-7.1/crossenv-scipy/bin/cross-python'
> host-python-version = '3.12'
> ```
> 
> you may have something similar in panda's `meson.build` file whereas you can enforce giving the full path in that second meson cross-file so it ends-up skipping the loading tests that will never work out, as loading numpy from cross tries to load aarch64 binaries over x86_64

It is not so easy!
the meson.build script of pandas looks like this:
```
incdir_numpy = run_command(py,
  [
    '-c',
    '''
import os
import numpy as np
try:
    # Check if include directory is inside the pandas dir
    # e.g. a venv created inside the pandas dir
    # If so, convert it to a relative path
    incdir = os.path.relpath(np.get_include())
except Exception:
    incdir = np.get_include()
print(incdir)
     '''
  ],
  check: true
).stdout().strip()

inc_np = include_directories(incdir_numpy)
inc_pd = include_directories('_libs/include')

fs.copyfile('__init__.py')

subdir('_libs')

subdirs_list = [
    '_config',
    '_testing',
    'api',
    'arrays',
    'compat',
    'core',
    'errors',
    'io',
    'plotting',
    'tests',
    'tseries',
    'util'
]
foreach subdir: subdirs_list
    install_subdir(subdir, install_dir: py.get_install_dir() / 'pandas')
endforeach

top_level_py_list = [
    '__init__.py',
    '_typing.py',
    '_version.py',
    'conftest.py',
    'testing.py'
]
py.install_sources(top_level_py_list, subdir: 'pandas')
```

It imports numpy to evaluate the numpy include directory. And, as you mentioned, this might not work in cross builds. 
But even when building for x64 (i.e. compatible target platform), it fails to load numpy.
Since there is no libgfortran* file in the working folder, I assume that numpy wheel is not working yet because the shared fortran libraries are missing.

We could try to patch meson.build script of pandas to overwrite the numpy include dir, but we have to fix the fortran dependency of numpy anyway.

hgy59:
> We could try to patch meson.build script of pandas to overwrite the numpy include dir, but we have to fix the fortran dependency of numpy anyway.

our numpy wheel is lackaging the fortran and other libs in the whl file

The whl file of `numpy-2.2.3-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` contains a folder `numpy.libs` with the files:
```
 2833617 Feb 12 20:03 libgfortran-040039e1-0352e75f.so.5.0.0
  250985 Feb 12 20:03 libquadmath-96973f99-934c22de.so.0.0.0
22468673 Feb 12 20:03 libscipy_openblas64_-6bb31eeb.so
```

hgy59:
BTW numpy wheel on the master branch does not contain the numpy.libs, but works for homeassistant (base system) in #6453 

numpy_test.py:
```python
import os
import numpy as np
try:
    # Check if include directory is inside the pandas dir
    # e.g. a venv created inside the pandas dir
    # If so, convert it to a relative path
    incdir = os.path.relpath(np.get_include())
except Exception:
    incdir = np.get_include()
print(incdir)
```
successfully runs:
```
 /var/packages/homeassistant/target/env/bin/python3  numpy_test.py
../../../var/packages/homeassistant/target/env/lib/python3.12/site-packages/numpy/_core/include
```

hgy59:
> At least, resulting wheel works from your testing so its already a win.

No, this is a missunderstanding. It works on the master branch only, not when built with this PR as spksrc/python/numpy*.
i.e. the former crossenv build works, the python-meson build does not.

th0ma7:
@hgy59 would that hapen to be on x86_64?  I noticed weird cpu detection at meson along with shorter build steps involved for that arch in particular (I usually use aarch64 by default to detect cross-compiling issues).

hgy59:
> @hgy59 would that hapen to be on x86_64? I noticed weird cpu detection at meson along with shorter build steps involved for that arch in particular (I usually use aarch64 by default to detect cross-compiling issues).

Sorry, I have only x86_64 for homeassistant 2025.1.4.
But I don't understand your question.
We have numpy built from src/requirements-numpy.txt that builds without native libraries and is known to work.
And we have numpy in this PR created with python-meson, resulting in wheel that does not include native libraries but links to native libraries, so it will only work when native libraries are included.
IMHO it is not a cross target issue, it is a basic issue that the native libraries are not included in the wheel.

th0ma7:
> > would that hapen to be on x86_64? I noticed weird cpu detection at meson along with shorter build steps involved for that arch in particular (I usually use aarch64 by default to detect cross-compiling issues).
> 
> Sorry, I have only x86_64 for homeassistant 2025.1.4. But I don't understand your question. 

Sorry for being cryptic... I meant, on what arch are you testing numpy wheels onto?  I was presuming x86_64... and its the case.  I noticed something when cross-compiling numpy for x86_64 on x86_64 is that it doesn't detect the CPU type properly and enables wrong optimizations.  You'll notice errors such as: 
```
WARNING: Unknown CPU family 'x86_64-pc-linux-gnu', please report this at https://github.com/mesonbuild/meson/issues/new with the output of `uname -a` and `cat /proc/cpuinfo`
Generating multi-targets for "_simd.dispatch.h" 
  Enabled targets: SSE42, AVX2, FMA3, FMA3__AVX2, AVX512F, AVX512_SKX, baseline
WARNING: Unknown CPU family 'x86_64-pc-linux-gnu', please report this at https://github.com/mesonbuild/meson/issues/new with the output of `uname -a` and `cat /proc/cpuinfo`
...
Build targets in project: 100
```

I've been looking into working around that and a way to do so is adding this:
```
# numpy badly detects CPU on x86_64
# Options are: HASWELL (recent), SANDYBRIDGE (older), NEHALEM (oldest and most compatible)
# CPU flags below are "sandybridge" compatible
ifeq ($(findstring $(ARCH),$(x64_ARCHS)),$(ARCH))
CONFIGURE_ARGS += -Dcpu-baseline="sse2,sse3,ssse3,sse41,sse42,avx"
CONFIGURE_ARGS += -Dcpu-dispatch="popcnt"
endif
```
Which will then output the following:
```
Generating multi-targets for "_simd.dispatch.h" 
  Enabled targets: baseline
WARNING: Unknown CPU family 'x86_64-pc-linux-gnu', please report this at https://github.com/mesonbuild/meson/issues/new with the output of `uname -a` and `cat /proc/cpuinfo`
Build targets in project: 59

NumPy 1.26.4

  User defined options
    Cross files : /home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-x64-7.1/numpy-1.26.4/numpy.meson
                  /home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-x64-7.1/numpy-1.26.4/x64-toolchain.meson
    Native files: /home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-x64-7.1/numpy-1.26.4/builddir/meson-python-native-file.ini
    buildtype   : release
    b_ndebug    : if-release
    b_vscrt     : md
    blas        : OpenBLAS
    cpu-baseline: sse2,sse3,ssse3,sse41,sse42,avx
    cpu-dispatch: popcnt
```

> We have numpy built from src/requirements-numpy.txt that builds without native libraries and is known to work.

@hgy59 sorry, what do you mean without "native" libraries?  Are you referring to `spksrc/native/python3.12/..../lib..` ?

> And we have numpy in this PR created with python-meson, resulting in wheel that does not include native libraries but links to native libraries, so it will only work when native libraries are included.

Unless you refer to `openblas` ?

> IMHO it is not a cross target issue, it is a basic issue that the native libraries are not included in the wheel.

Sorry but I need a bit more explanation...  Unless you are referring to these from comment above https://github.com/SynoCommunity/spksrc/pull/6454#issuecomment-2709732800: 
```
 2833617 Feb 12 20:03 libgfortran-040039e1-0352e75f.so.5.0.0
  250985 Feb 12 20:03 libquadmath-96973f99-934c22de.so.0.0.0
22468673 Feb 12 20:03 libscipy_openblas64_-6bb31eeb.so
```

hgy59:
@th0ma7 found some useful information at https://numpy.org/doc/stable/building/redistributable_binaries.html

IHMO we have to build the native libraries separately and use `auditwheel` to include those in the wheel.

hgy59:
@th0ma7 there other build informations for numpy, like that a numpy specific version of meson must be used (on the next page https://numpy.org/doc/stable/building/understanding_meson.html).

th0ma7:
@hgy59 I'll definitively have a look at those links.  If you don't mind reading my previous post https://github.com/SynoCommunity/spksrc/pull/6454#issuecomment-2711983646 to confirm we're referring to the same things.

As far as meson, numpy's internal build system uses the vendor meson by default so in theory that is not an issue.

As for said files, assuming when talking about "native" you refer to:
```
 2833617 Feb 12 20:03 libgfortran-040039e1-0352e75f.so.5.0.0
  250985 Feb 12 20:03 libquadmath-96973f99-934c22de.so.0.0.0
22468673 Feb 12 20:03 libscipy_openblas64_-6bb31eeb.so
```
Both `libgfortran` and `libquadmath` are part of gcc and included in our DSM7 toolchains:
```
spksrc@spksrc:~/wheel-meson2/spksrc/toolchain/syno-x64-7.1$ find . -name "libquadmath*"
./work/x86_64-pc-linux-gnu/share/licenses/gcc/libquadmath
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libquadmath.so.0.0.0
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libquadmath.so.0
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libquadmath.a
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libquadmath.so
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libquadmath.so
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libquadmath.a
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libquadmath.so.0
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libquadmath.so.0.0.0
spksrc@spksrc:~/wheel-meson2/spksrc/toolchain/syno-x64-7.1$ find . -name "libgfortran*"
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libgfortran.so.5.0.0
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libgfortran.so.5
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libgfortran.so
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libgfortran.a
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib64/libgfortran.spec
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libgfortran.spec
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libgfortran.a
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libgfortran.so
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libgfortran.so.5
./work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib/libgfortran.so.5.0.0
```
I assume they are available from our sysroot at compile time with the regular lib directories passed:
```
c_link_args = [
        '-L/home/spksrc/wheel-meson2/spksrc/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sys-root/lib',
        '-L/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-x64-7.1/install/var/packages/python312-wheels/target/lib',
        '-Wl,--rpath-link,/home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-x64-7.1/install/var/packages/python312-wheels/target/lib',
        '-Wl,--rpath,/var/packages/python312-wheels/target/lib',
        '-L/home/spksrc/wheel-meson2/spksrc/spk/python312/work-x64-7.1/install/var/packages/python312/target/lib',
        '-Wl,--rpath-link,/home/spksrc/wheel-meson2/spksrc/spk/python312/work-x64-7.1/install/var/packages/python312/target/lib',
        '-Wl,--rpath,/var/packages/python312/target/lib',
        ]
```
The only thing may be that they are missing on the default synology image and would require us to package them somehow... and part of the numpy wheel could indeed be a way to go.

Leaving only `libscipy_openblas64` to handle... which again is part of our default package library path:
```
spksrc@spksrc:~/wheel-meson2/spksrc/spk/python312-wheels/work-x64-7.1/install$ find . -name "*openblas*.so*"
./var/packages/python312-wheels/target/lib/libopenblas.so.0.3
./var/packages/python312-wheels/target/lib/libopenblas.so.0
./var/packages/python312-wheels/target/lib/libopenblas.so
```
All in all I'm not sure anything else is missing?!?!

## TL;DR
One thing we may have to look at are `libgfortran.so*` and `libquadmath.so*` taken from the toolchain and either include them into the numpy wheel or part of the package like we do for `libatomic.so` part of `spksrc.strip.mk`.

th0ma7:
Theory to preliminary conclusions:

1. I've added a service_setup.sh on my local branch into `python312-wheels` package in order to automatically create the virtualenv at package installation time (although it avoid installing the wheels).
2. I then manually installed the wheels such as:
```
$ sudo su -s /bin/bash sc-python312-wheels -c '/var/packages/python312/target/bin/pip3 install numpy==2.2.3 --no-index --force-reinstall --find-links=/var/packages/python312-wheels/target/share/wheelhouse'
$ sudo su -s /bin/bash sc-python312-wheels -c '/var/packages/python312/target/bin/pip3 install scipy==1.15.2 --no-index --force-reinstall --find-links=/var/packages/python312-wheels/target/share/wheelhouse'
```
3. Validated that said libraries are available:
```
root@th0ma7-nas:/var/packages/python312-wheels/target/lib# find . -name "libopenblas.so*"
./libopenblas.so
./libopenblas.so.0
./libopenblas.so.0.3
root@th0ma7-nas:/var/packages/python312-wheels/target/lib# find . -name "libgfortran.so*"
./libgfortran.so.5.0.0
./libgfortran.so
./libgfortran.so.5
```
4. Then I test-bed your script numpy test script:
```
  File "/var/packages/python312-wheels/home/.local/lib/python3.12/site-packages/numpy/_core/__init__.py", line 23, in <module>
    from . import multiarray
  File "/var/packages/python312-wheels/home/.local/lib/python3.12/site-packages/numpy/_core/multiarray.py", line 10, in <module>
    from . import overrides
  File "/var/packages/python312-wheels/home/.local/lib/python3.12/site-packages/numpy/_core/overrides.py", line 7, in <module>
    from numpy._core._multiarray_umath import (
ImportError: libopenblas.so.0: cannot open shared object file: No such file or directory
...
```
And... not boom but rather bang.. ?

## Next?
Either a) ensure the `rpath` are functioning properly into numpy/* wheel libraries OR b) include the libraries into the wheel package.  I have a strong preference with a) ...

hgy59:
@th0ma7 I use a diyspk/nump-wheel package and include the numpy_test.py shown above.

Added a service-setup.sh with
```
service_postinst ()
{
    install_python_virtualenv
    
    pip install --disable-pip-version-check --no-deps --no-input --no-index ${SYNOPKG_PKGDEST}/share/wheelhouse/*.whl
}
```

running on virtualdsm:
```
./bin/python3 ../numpy_test.py
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/_core/__init__.py", line 23, in <module>
    from . import multiarray
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/_core/multiarray.py", line 10, in <module>
    from . import overrides
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/_core/overrides.py", line 7, in <module>
    from numpy._core._multiarray_umath import (
ImportError: libopenblas.so.0: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/__init__.py", line 114, in <module>
    from numpy.__config__ import show_config
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/__config__.py", line 4, in <module>
    from numpy._core._multiarray_umath import (
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/_core/__init__.py", line 49, in <module>
    raise ImportError(msg)
ImportError:

IMPORTANT: PLEASE READ THIS FOR ADVICE ON HOW TO SOLVE THIS ISSUE!

Importing the numpy C-extensions failed. This error can happen for
many reasons, often due to issues with your setup or how NumPy was
installed.

We have compiled some common reasons and troubleshooting tips at:

    https://numpy.org/devdocs/user/troubleshooting-importerror.html

Please note and check the following:

  * The Python version is: Python3.12 from "/volume1/@appstore/numpy-wheel/env/bin/python3"
  * The NumPy version is: "2.2.0"

and make sure that they are the versions you expect.
Please carefully study the documentation linked above for further help.

Original error was: libopenblas.so.0: cannot open shared object file: No such file or directory


The above exception was the direct cause of the following exception:

  File "/volume1/@appstore/numpy-wheel/env/../numpy_test.py", line 2, in <module>
    import numpy as np
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/__init__.py", line 119, in <module>
    raise ImportError(msg) from e
ImportError: Error importing numpy: you should not try to import numpy from
        its source directory; please exit the numpy source tree, and relaunch
        your python interpreter from there.

```

I can't find any binary that depends on libopenblas within the wheel, so I guess it is dynamically loaded and might have a specific search order.

running with explicit library path ` LD_LIBRARY_PATH=/var/packages/numpy-wheel/target/lib  ./bin/python3 ../numpy_test.py`:
```
# LD_LIBRARY_PATH=/var/packages/numpy-wheel/target/lib  ./bin/python3 ../numpy_test.py
  File "/volume1/@appstore/numpy-wheel/env/../numpy_test.py", line 2, in <module>
    import numpy as np
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/__init__.py", line 114, in <module>
    from numpy.__config__ import show_config
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/__config__.py", line 4, in <module>
    from numpy._core._multiarray_umath import (
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/_core/__init__.py", line 23, in <module>
    from . import multiarray
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/_core/multiarray.py", line 10, in <module>
    from . import overrides
  File "/volume1/@appstore/numpy-wheel/env/lib/python3.12/site-packages/numpy/_core/overrides.py", line 7, in <module>
    from numpy._core._multiarray_umath import (
RuntimeError: NumPy was built with baseline optimizations:
(SSE SSE2 SSE3 SSSE3 SSE41 POPCNT SSE42 AVX) but your machine doesn't support:
(AVX).
```

the AVX optimization seems not supported in virtualdsm.

The definition of LD_LIBRARY_PATH is not a problem for the HA package (it already has it).
but a self-contained solution is preferable.

to fix this, the rpath must be fixed/adjusted in the so files of the numpy wheel.
there are two libraries in the installed numpy package that depend on libopenblas
- numpy/_core/_multiarray_umath.cpython-312-x86_64-linux-gnu.so
- numpy/linalg/lapack_lite.cpython-312-x86_64-linux-gnu.so

and both have `Library rpath: [/spksrc/diyspk/numpy_ha/work-x64-7.1/install/var/packages/numpy-wheel/target/lib]`

hgy59:
@th0ma7 the above test works on DS-115j (armada370 - armv7) with DSM 7.1 when using LD_LIBRARY_PATH.

th0ma7:
@hgy59 would you mind commenting out the cpu-dispatch and cpu-baseline definition in numy's baseline I've added yesterday and retrying on you x86_64? and removing avx and re-test?  I'll have to read further on this to get a proper understanding on how to using that (away from my build system atm).

Also, I'm almost certain that the rpath is not functional atm, not only for meson-python wheels but probably overall for all meson builds.  And that issue is new with this PR.  Once that is fixed the LD_LIBRARY_PATH should not be needed.

hgy59:
@th0ma7 my test even works on DS-218 (aarch64) with DSM 6.2.4 when patching `os_min_version` in the INFO file of the package from DSM 7.1 to 6.2.

BTW the `libquadmath.so` is not contained in armv7 and aarch64 packages (seems x64 only).

hgy59:
> @hgy59 would you mind commenting out the cpu-dispatch and cpu-baseline definition in numy's baseline I've added yesterday and retrying on you x86_64? and removing avx and re-test? I'll have to read further on this to get a proper understanding on how to using that (away from my build system atm).

working on it locally...

th0ma7:
> Update: when I first removed the non apollolake compatible cpu flags (sse3 and avx) in cpu-baseline, I got core dumps. This was fixed with the correct TARGET for openblas and removal of cpu-baseline and cpu-dispatch. Finally I reactivated cpu-baseline (and cpu-dispatch) with apollolake compatible flags and it still worked.

This is really nice!  Good work!  What would be even better is the ability to completely avoid that altogether... but I doubt we can.  Now let's find the issue with rpath...

hgy59:
@th0ma7 my findings are, that we have to setup meson with --prefix for correct rpath, but I can't find where (or how) meson setup is called for our python-meson builds.

```
meson setup build --prefix=$(INSTALL_PREFIX)
```

it is also documented for numpy on https://numpy.org/doc/stable/building/understanding_meson.html

th0ma7:
Aaahhh!  Thnx for the pointer, i believe i know how to fix this later tonight!

hgy59:
> Aaahhh! Thnx for the pointer, i believe i know how to fix this later tonight!

I did some more tests but didn't succeed. (added `install_rpath` and `build_rpath` to [properties], added `CONFIGURE_ARGS += --prefix=$(INSTALL_PREFIX)` and set `prefix = $(INSTALL_PREFIX)`  under [built-in options]).

The so files in the builddir are binary the same as in the whl file.
My suggestion is that the meson install step (that would adjust the rapth) does not happen.

th0ma7:
I may have something that works now.  The `libtree -vvv -p` now include our rpath as depth 1 as below:
```
│   └── libgfortran.so.5 not found
│       ┊ Paths considered in this order:
│       ┊ 1. rpath:
│       ┊    depth 2
│       ┊    /var/packages/python312-wheels/target/lib
│       ┊    /var/packages/python312/target/lib
│       ┊    depth 1
│       ┊    /var/packages/python312-wheels/target/lib
│       ┊    /var/packages/python312/target/lib
│       ┊    /home/spksrc/wheel-meson2/spksrc/spk/python312-wheels/work-x64-7.1/install/var/packages/python312-wheels/target/lib
│       ┊ 2. LD_LIBRARY_PATH was not set
│       ┊ 3. runpath was not set
│       ┊ 4. ld config files:
│       ┊    /usr/lib/x86_64-linux-gnu/libfakeroot
│       ┊    /usr/local/lib
│       ┊    /usr/local/lib/x86_64-linux-gnu
│       ┊    /lib/x86_64-linux-gnu
│       ┊    /usr/lib/x86_64-linux-gnu
│       ┊    /lib32
│       ┊    /usr/lib32
│       ┊    /libx32
│       ┊    /usr/libx32
│       ┊ 5. Standard paths:
│       ┊    /lib
│       ┊    /lib64
│       ┊    /usr/lib
│       ┊    /usr/lib64
```
Still, I don't get the rpath depth 2 neither the equivalent to `$(STAGING_INSTALL_PREFIX)` part of rpath depth 1 which is totally useless.

As expected, the `$(ENV)` environment has a strong influence over the build process.  Although when using cmake or meson with a "full-featured" cross-file, the environment needs to be fully cleaned-up of any autoconf-type env variables.  Something I already did with CMAKE as it already had a `full-featured` cross-file, although I wasn't with meson before this PR as it's crossenv was still lagging per `cross/<depend>` cross-file generation and therefore dependent of `*FLAGS` from `$(ENV)`.

Now, I did try going all-in with a systematic `env -i` instead of unsetting specific variables (using `env -u`) but if fails on pkg-config in some cases.  If I could find why that may remove the uneeded rpath in our librairies.  On the other hand, it would then be missing other non-autoconf variables such as CARGO / RUSTC and others.

For fun, have a look at `libtree -vvv -p ./numpy/core/_multiarray_umath.cpython-312-x86_64-linux-gnu.so` before then after changing line 81 of `spksrc.python-wheel-meson.mk` such as:
```
cd $(MESON_BASE_DIR) && env $(ENV) \
```
I recall struggling on this when adding cmake & meson long ago... Now I recall why a bit more.

mk/spksrc.cross-meson-env.mk
```diff
@@ -84,21 +87,24 @@ $(MESON_CROSS_TOOLCHAIN_PKG):
 meson_pkg_toolchain: SHELL:=/bin/bash
 meson_pkg_toolchain:
 	@cat $(MESON_CROSS_TOOLCHAIN_WRK)
+	@echo "pkgconfig = '$$(which pkg-config)'"
```
hgy59, line 90:
```suggestion
	@echo "pkgconfig = '$$(which pkgconfig)'"
```

hgy59:
BTW the build log of openblas is flooded with warnings when building numpy only in diyspk:
```
f951: Warning: Nonexistent include directory ‘/spksrc/diyspk/numpy_ha/work-x64-7.1/install/var/packages/numpy-wheel/target/include’ [-Wmissing-include-dirs]
```

This comes from `-I $(STAGING_INSTALL_PREFIX)/include` somewhere.

It occurs while `Building Fortran preprocessed CMakeFiles/LAPACK_OVERRIDES.dir/lapack-netlib/SRC/*`when no other dependency is built before.

A pragmatic solution would be to create the include folder in pre_compile, since modules with dependencies require it.

```Makefile
.PHONY: openblas_pre_compile
openblas_pre_compile:
	install -d -m 755 $(STAGING_INSTALL_PREFIX)/include
```

A better solution would be to create this folder in the Makefile that defines the include path.
For openblas it refers to cross-cmake, but other cross-envs could also be affected.

hgy59:
> I may have something that works now. The `libtree -vvv -p` now include our rpath as depth 1 as below:

It now works now without LD_LIBRARY_PATH 🎉 .
Alas it still has the full library install path in it, that does not exist on the target system...
```
ash-4.4# readelf -d  lib/python3.12/site-packages/numpy/_core/_multiarray_umath.cpython-312-x86_64-linux-gnu.so

Dynamic section at offset 0x3cf978 contains 25 entries:
 0x0000000000000001 (NEEDED)             Shared library: [libopenblas.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x000000000000000f (RPATH)              Library rpath: [/var/packages/numpy-wheel/target/lib:/var/packages/python312/target/lib:/spksrc/diyspk/numpy_ha/work-x64-7.1/install/var/packages/numpy-wheel/target/lib]
```

th0ma7:
Yes finally!  I'm starting to wonder if this isn't an issue specific to numpy's vendor meson....  While functional it still needs to be fixed somewhat though.

th0ma7:
@hgy59 I haven't count how many times I've build numpy (and it's way too many), but I'm now almost certain this is a bug with meson.  I've documented my findings in https://github.com/mesonbuild/meson/issues/14354 and believe this relates to a long-standing bug at https://github.com/mesonbuild/meson/issues/6541

I do have code to workaround that, enforcing to have `LDFLAGS` set in our environment on top of the cross-file (to be pushed soon).  This ensures to have proper rpath as depth 1 but still, build rpath shows in that depth 1 and our host rpath are being duplicated in depth 2 and there is nothing I can do except unpack the wheel and edit every `.so` files to delete that from it (which I don't intent to do).

Slowly pursuing...

mreid-tt:
@th0ma7, I really appreciate all your effort in resolving this. From the long-standing bug you mentioned, it appears that many attempts have been made across various Meson versions to address this issue. I also noticed references from the university HPC community about different patches to Meson that alter RPATH handling, especially in environments with wrappers or cross compilation. Though some of this is a bit over my head, I’m curious how you’d describe our current situation.

In your new ticket, you mention using `rpath` and `rpath-link`. However, in an older ticket, one of the authors stated:

> If you use just a single `-rpath` or `-rpath-link` option, it's completely discarded.

That made me wonder about our specific scenario with `spksrc`. Sorry if you’ve already explained this — I'm just trying to understand the workaround you’re implementing and how it compares to what was discussed in the older ticket.

th0ma7:
@mreid-tt My assumption is rather that the rpath is not being discarded but actually hidden under the rpath depth.  When using `readlink -d` it only shows rpath depth 1 whereas when using `libtree -vvv -d` it actually shows depth 1+2.  Note that I didn't knew that rpath depth even existed before now but when using the cross-compiled binary later on our destination host, it seems to only refers to rpath depth 1.

From what I gathered from our use-case, when setting `PKG_CONFIG_LIBDIR`, mandatory needed in our cross-compile case to reset the default location of pkgconfig directory, it somehow ends-up enforcing a "build-environment" rpath as depth 1 (sort of rpath-link), leaving our rpath settings from `install_rpath` variable in our generated meson cross-file as rpath depth 2.

A bit more background is needed here: when using a cross-file with meson or cmake, you need to empty your shell environment from any other duplicate flags (i.e. mostly `*FLAGS` variables) that are now already managed from your cross-file to avoid any conflicts.  Work-around I've found is keeping the `LDFLAGS` with our rpath definitions makes meson to also include those rpath as depth 1 (and duplicated also in depth 2).  My assumption is that `install_rpath` is handled as depth 2 while `CFLAGS` rpath are handled as depth 1.  No clue as to why.

I still have a few more tests to go through to confirm our final meson build environment state, but this is starting to stabilize with my latest commit https://github.com/SynoCommunity/spksrc/pull/6454/commits/e325eb1f5648a810347ede727fd89a6c2740b1b3.

Lastly, I still have to double-check its behavior outside of meson-python to ensure this isn't affecting our regular meson builds.

mreid-tt:
@th0ma7, thanks for sharing. To clarify my understanding about "rpath depth," Google AI describes it as:

> "In software compilation and linking, **rpath depth** refers to the number of directory levels specified in an RPATH (runtime path) when linking an executable. RPATH directs the linker to locations of shared libraries at runtime, and the depth represents how many subdirectory levels are searched."

Given this definition, what is the practical implication of specifying more directory levels than necessary (e.g., depth of 1+2) for our builds? Would this result in larger binaries, or merely a slight increase in build time? I'm trying to fully understand any potential impact of your proposed workaround.

spk/python312-wheels/Makefile
```diff
@@ -129,8 +139,15 @@ WHEELS_CFLAGS += [msgpack] -std=c11
 endif
 
 # [numpy]
-ifeq ($(call version_gt, $(TC_GCC), 5.0),1)
-WHEELS += src/requirements-crossenv-numpy.txt
+# gcc-8.4 is the minimal version to build numpy
+# gcc-8.5 hapens to be the default version for DSM-7.1
```
mreid-tt, line 143: So, it seems we can no longer build older NumPy versions for DSM versions below 7.1, correct?

mk/spksrc.cross-meson-env.mk
th0ma7: yes, but no.  There are other use-cases that requires `pkgconfig` or `pkg-config` definitions.  Including both solves that.

spk/python312-wheels/Makefile
th0ma7: close enough, we can no longer build older numpy with python 3.12 as no longer compatible.
And newer numpy requires gcc >= 8.4 therefore only supporting DSM-7.1 (as even DSM7 was using gcc7)

mreid-tt:
@th0ma7, anything remains to be done for this PR before it can be merged?

th0ma7:
A bit of code cleanups and adding fortran to cmake for dsm7 and above.  This has been taking way more cycles than expected, cycles that are limited.

th0ma7:
I'll do my best to get this merged over the coming days, hopefully within a week.

Note that i did try to get panda and scikit-learn to build but i wasn't able to.  Other changes may have to follow for those in particular.

th0ma7:
@hgy59 and @mreid-tt I've now cleaned-up the python-meson code and am pretty happy with it.  I'm testing the latest changes to fix any remaining bugs to be found.

Once that is settled I'll then make the necessary adjustments to the `cmake` code which should be "relatively" straight-forward at that point.

Overall the changes to the framework will be invsasive and there may be a bit of bug hunting todo afterwards.  On that I've updated this PR description to provide additional background for later-on.

th0ma7:
@hgy59 and @mreid-tt I'll let github-action run and if no remaining issues are found I'll merge.  I'll do the remaining changes mentioned in the description from another PR helping tracking issues between PR if any.

Feel free to let me know if you see anything needing adjustments prior from me merging things.  thnx.

spk/python311-wheels/Makefile
```diff
@@ -100,6 +107,15 @@ endif
...
 include ../../mk/spksrc.python.mk
 
+# [atom]
+# Requires
+#  - build:setuptools-scm
+#  - cross:cppy
+ifeq ($(call version_ge, $(TC_GCC), 4.8),1)
+WHEELS += atom==0.10.5
+WHEELS += atom==0.11.0
+endif
```
th0ma7, line 117: @hgy59 and @mreid-tt just briging your attention that the ability to include wheel requirement definition directly within the WHEELS variable is now functional once again.  This can allow avoiding using multiple requirement files for every gcc versions and ease code maintenance.  Although, I don't suggest moving away from using requirement files, but use this for special cases.

Lastly, note that you cal add the `pure:`, `abi3:` or `cross:` prefix.  With none defaults to `cross`.

hgy59:
@th0ma7 it looks like it failed to download https://download.savannah.gnu.org/releases/acl/acl-2.3.1.tar.xz

will cancel, delete the download cache an rerun all jobs (not only failed) should fix it.

hgy59:
@th0ma7 thanks for the huge work!

shall we now publish Python 3.13 (and update of Python 3.12)?

th0ma7:
> @th0ma7 thanks for the huge work!

A pleasure, and sorry for taking that long!  This was more complex than anticipated.

> shall we now publish Python 3.13 (and update of Python 3.12)?

Sure, although i would not suggest migrating to 3.13 now (feels like we had that chat but don't recall the endind)

Also, if you want to help out on this effort, feel free to move all python module/wheel based entries from cross to the new python folder :)

hgy59:
> Sure, although i would not suggest not migrating to 3.13 now (feels like we had that chat but don't recall the ending)

already tried py313 with homeassistant >= 2025.2.x, but several wheels fail to cross compile...

hgy59:
> Also, if you want to help out on this effort, feel free to move all python module/wheel based entries from cross to the new python folder :)

sure, but my cycles are limitted too, in the next few weeks...

th0ma7:
thnx @hgy59 for uploading py312-313.  I just noticed the py313 changelog, not that it matter too much:
```
Version 3.13.2-2
1. Initial Python 3.13.1 package release
```
and likewise for py312:
```
Version 3.12.9-2
1. Initial Python 3.12.8 package release
```

PR #6490, opened 2025-03-24
make dsm7 uninstall less verbose

hgy59:
## Description

- do not print deleted files to installer log when removing all files on uninstall

### Motivation
The generic uninstall function on dsm7 supports the removal of all installed data.
When removing all files, every file is logged to the installer log of the packag.

This floods the installer log with (IMHO) unimportant information.
When a python package is uninstalled with this option, all files in the pip cache (in the var folder) are printed to the installer log For Home Assistant with a lot of wheels and package data there are > 60'000 lines printed and the log file gets immediatly rotated.

## Checklist

### Type of change


publicarray:
Thanks, I think this was useful once upon a time when the DSM failed to uninstall a package. But it probably isn't necessary anymore

Issue #3633, opened 2019-02-17
Python 2.7 EOL


hgy59:
When creating the new docker image for the build pipeline I got a lot of warnings telling the End of Life of python 2.x at the end of this year (2019)!.

Therefore this issue should initiate a discussion about the timeline and need to remove python 2.7 from the synocommunity build environment and the kick out of all packages depending on python 2.

I know that python 2.7 is the base of a lot of packages here, but let give me some facts:

- I suppose that synology will drop python 2 from the current DSM version this year
- Or may be that synology will introduce a new DSM version with python 3.x only
- Any diskstation running with python 2.x are considered to be vulnerable as of January 1, 2020

Therefore
- We need a schedule to migrate all packages depending on python 2.x to python 3.x
- All the existing packages depending on python 2.x (and python 2.7 itself) should be removed from the repository by end of 2019

let the discussion begin...

ymartin59:
I would even say it question the need to re-publish 2.7.15 and all python2-based packages... Really a large effort which sounds useless if we decide to invest on moving applications to python3 - when supported upstream already.

BenjV:
I very much disagree with this proposal to remove Python 2.
The fact that development  of Python 2 is stopped, does not mean that it is absolute.
Development of python 2 has been stopped already from the day they release the first 2.7 version and that contained only backports from Python 3 to make migration easier.
There are thousands of python scripts that depend on python 2 so it will be around for many, many more years.

And converting all python based package from python 2 to python 3 is not something the SynoCommunity should even try.
It is not an easy task and would require al lot of effort, especially in testing the application.
That can only be done by the developers of the application.

Synology itself stopped distributing the python 2 packages as of DSM 6 but still distribute Python 2 with DSM because DSM 6 uses a lot of python 2 scripts.

Instead of this poorly thought out proposol, I propose to remove the Python 3 package from the SynoCommunity.
Python 3 is distributed by Synology themselves and two Python 3 packages are only creating confusion.
So lets change the packages that use the python 3 of the SynoCommunity to use the Synology version and remove the python 3 package of the SynoCommunity.
We could also create a package that could easily add site-packages to the Synology distribution from the PyPi website so no application should need to distribute those site-packages themselves.

hgy59:
python 2.x usage will be a risk from January 1, 2020 an will stop working some time later.

The message when building the current docker image for this repo ist definite:

[...]

PR #6504, opened 2025-04-01
drop python2

hgy59:
## Description

- five years after EOL of python 2 it is time to drop it

The single remaining package that depends on python2 is pyload.
This is deprecated too (the latest release of pyload depends on python 3.6). 
So we should drop pyload too (released 2016 for DSM 5 only, and still beta)

related to #3633

## TODO

- [ ] Drop python2 and pyload in the package repository.

### Type of change


hgy59:
@th0ma7 before the migration of all python-module and python-wheel dependencies from cross to python folder, I propose to kick off any python2 only references.

th0ma7:
> @th0ma7 before the migration of all python-module and python-wheel dependencies from cross to python folder, I propose to kick off any python2 only references.

Amen!

mreid-tt:
I'm guessing these lines are no longer needed as well:
https://github.com/SynoCommunity/spksrc/blob/2d0addbf82ba102c3d569488136479f65b47f5ee/.github/actions/prepare.sh#L43-L45

PR #6518, opened 2025-04-07
Python module migration

mreid-tt:
## Description

Aligning with the new module structure in https://github.com/SynoCommunity/spksrc/issues/6421, this PR includes the following:

1. Re-home `cross/cryptography` to `python/cryptography` and update `salt-master`, `salt-minion`.
2. Re-home `cross/pip` to `python/pip` and update `python31[0-3]`.

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6519, opened 2025-04-07
Build Script: Add strict mode, error trapping, and checksum verification

mreid-tt:
## Description

This change refactors the existing `download.sh` script to run in strict mode (`set -euo pipefail`), adds a global `ERR` trap for clear error and line reporting, and verifies that `make` is installed before proceeding. It preserves the two loops that handle cross-package (`download-all`) and Python-wheel (`download-wheels`) downloads — now iterating with clearer variable names and inlined progress echoes — while appending a `checksum` invocation to the cross-package download step. Finally, it combines `NOARCH_PACKAGES` and `ARCH_PACKAGES` into a Bash array for reliable empty checks and prints a succinct “All downloads completed successfully.” message, ensuring full parity with the old behavior but with much stronger error visibility and integrity verification.

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/.github/actions/download.sh
+++ b/.github/actions/download.sh
@@ -37,9 +37,9 @@ else
     echo "===> Downloading wheels: ${build_pkgs[*]}"
     for pkg in "${build_pkgs[@]}"; do
         current="spk/${pkg}"
-        echo "  → ${current}: download-wheels then checksum"
-        # download-wheels grabs all needed .whl files; checksum verifies them.
-        make -C "${current}" download-wheels checksum
+        echo "  → ${current}: download-wheels"
+        # download-wheels grabs all needed .whl files.
+        make -C "${current}" download-wheels
     done
 fi
 
--- a/.github/actions/download.sh
+++ b/.github/actions/download.sh
@@ -22,15 +22,12 @@ download_package() {
         echo "Error: Failed to download ${package_dir}. Exiting." >&2
         exit 1
     fi
-    echo "===> Successfully downloaded: ${package_dir}"
-
     # Verify checksum after download
-    echo "===> Verifying checksum for: ${package_dir}"
     if ! make -C "${package_dir}" checksum; then
         echo "Error: Checksum verification failed for ${package_dir}. Exiting." >&2
         exit 1
     fi
-    echo "===> Checksum verified: ${package_dir}"
+    echo "===> Successfully downloaded: ${package_dir}"
 }
 
 # Download regular cross/* sources
--- a/.github/actions/download.sh
+++ b/.github/actions/download.sh
@@ -43,6 +43,7 @@ build_packages="${NOARCH_PACKAGES} ${ARCH_PACKAGES}"
 if [ -z "${build_packages}" ]; then
     echo "===> No wheels to download. <==="
 else
+    echo "===> Download wheels: ${build_packages}"
     for package in ${build_packages}; do
         download_package "spk/${package}" "download-wheels"
     done
--- a/.github/actions/download.sh
+++ b/.github/actions/download.sh
@@ -1,61 +1,47 @@
 #!/bin/bash
-#
+
 # Part of GitHub build action.
 # This script downloads source files for cross and python packages as part of our build process.
 #
 # Functions:
-# - Download all referenced native and cross source files for packages.
-# - Download all referenced python wheels needed to build.
-# - Use "download-all" target to get all files when a package has multiple (architecture-specific) files.
-
+#  - Download all referenced native and cross source files for packages.
+#  - Download all referenced python wheels needed to build.
+#  - Use the “download-all” target when a package has multiple (arch-specific) files.
+#
 set -euo pipefail
+# Report any error (with line and package context) and exit.
+trap 'echo "::error::Error on line ${LINENO} while processing ${current:-<none>}"; exit 1' ERR
 
-# Function to print an error message and exit.
-fail() {
-    echo "::error::$*" >&2
-    exit 1
-}
+# Ensure required tooling is present.
+command -v make >/dev/null 2>&1 || { echo "::error::make is not installed"; exit 1; }
 
-# Check for required commands.
-command -v make >/dev/null 2>&1 || fail "make is not installed"
-
-# Function to download a package and verify its checksum.
-download_package() {
-    local package_dir="$1"
-    local target="$2"
-
-    echo "===> Attempting to download: ${package_dir}"
-    if ! make -C "${package_dir}" "${target}"; then
-        fail "Failed to download ${package_dir}."
-    fi
-    # Verify checksum after download.
-    if ! make -C "${package_dir}" checksum; then
-        fail "Checksum verification failed for ${package_dir}."
-    fi
-    echo "===> Successfully downloaded: ${package_dir}"
-}
-
-# Download regular cross/* sources.
+echo ""
+# 1) Download native / cross-compiled sources.
 if [ -z "${DOWNLOAD_PACKAGES:-}" ]; then
-    echo "===> No packages to download. <==="
+  echo "===> No packages to download. <==="
 else
-    echo "===> Download packages: ${DOWNLOAD_PACKAGES}"
-    for download in ${DOWNLOAD_PACKAGES}; do
-        download_package "${download}" "download-all"
-    done
+  echo "===> Downloading packages: ${DOWNLOAD_PACKAGES}"
+  for current in ${DOWNLOAD_PACKAGES}; do
+    echo "  → ${current}: download-all then checksum"
+    # download-all pulls down all sources; checksum verifies them.
+    make -C "${current}" download-all checksum
+  done
 fi
 
 echo ""
-
-# Download python wheel source files.
-build_packages="${NOARCH_PACKAGES:-} ${ARCH_PACKAGES:-}"
-if [ -z "${build_packages}" ]; then
-    echo "===> No wheels to download. <==="
+# 2) Download Python wheel source files.
+build_pkgs=( ${NOARCH_PACKAGES:-} ${ARCH_PACKAGES:-} )
+if [ ${#build_pkgs[@]} -eq 0 ]; then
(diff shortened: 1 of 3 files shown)


hgy59:
@mreid-tt another improvement (to fail-fast) is to validate the checksums after download

this additional code in `download_package()` would do it (must be done after the download):
```
    if ! make -C "${package_dir}" checksum; then
        echo "Error: Download verification failed ${package_dir}. Exiting." >&2
        exit 1
    fi
```

This will capture checksum errors;
- when url leads to an error page and the content of the page is downloaded as package source.
- when origin source is changed (same name, different content).

.github/actions/download.sh
```diff
@@ -10,16 +10,28 @@
...
 
 set -o pipefail
+set -e  # Exit on any command failure
+
+# Function to download a package and handle errors
+download_package() {
+    local package_dir="$1"
+    local target="$2"
+    
+    echo "===> Attempting to download: ${package_dir}"
```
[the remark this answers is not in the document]
mreid-tt: Thanks, went with the latter proposal.

hgy59:
@mreid-tt for testing you can modify a package that triggers a lot of packages (like cross/libxml).
And to force an error, you could modify (or remove) an entry in the digests file (when added the checksum test suggested above).

But I have no idea how to simulate connection errors.

.github/actions/download.sh
```diff
@@ -10,16 +10,36 @@
...
+    echo "===> Successfully downloaded: ${package_dir}"
+
+    # Verify checksum after download
+    echo "===> Verifying checksum for: ${package_dir}"
+    if ! make -C "${package_dir}" checksum; then
+        echo "Error: Checksum verification failed for ${package_dir}. Exiting." >&2
+        exit 1
+    fi
+    echo "===> Checksum verified: ${package_dir}"
+}
```
hgy59, line 34: one success message is enough (IMHO)

```suggestion
    fi
    # Verify checksum after download
    if ! make -C "${package_dir}" checksum; then
        echo "Error: Checksum verification failed for ${package_dir}. Exiting." >&2
        exit 1
    fi
    echo "===> Successfully downloaded: ${package_dir}"
}
```

mreid-tt:
> @mreid-tt for testing you can modify a package that triggers a lot of packages (like cross/libxml). And to force an error, you could modify (or remove) an entry in the digests file (when added the checksum test suggested above).

I cloned the changes to a test branch so I can trigger it for myself only and changed one line in the `cross/libxml2` as suggested. I think the error sensitivity is revealing something else though (see: https://github.com/mreid-tt/spksrc/actions/runs/14318954031/job/40131504605)

```
make: Entering directory '/home/runner/work/spksrc/spksrc/cross/bind_9.16'
===>  Verifying files for bind
===>    Checking sha1sum of file bind-9.16.50.tar.xz
===>    Checking sha256sum of file bind-9.16.50.tar.xz
===>    Checking md5sum of file bind-9.16.50.tar.xz
/home/runner/work/spksrc/spksrc/cross/bind_9.16/../../distrib/bind-9.16.50.tar.xz
make: Leaving directory '/home/runner/work/spksrc/spksrc/cross/bind_9.16'
===> Checksum verified: cross/bind_9.16
===> Attempting to download: cross/binutils
make: Entering directory '/home/runner/work/spksrc/spksrc/cross/binutils'
make: *** No rule to make target 'download-all'.  Stop.
make: Leaving directory '/home/runner/work/spksrc/spksrc/cross/binutils'
Error: Failed to download cross/binutils. Exiting.
Error: Process completed with exit code 1.
```

Any thoughts on how to address this?

hgy59:
> Any thoughts on how to address this?

cross/binutils is a dummy package (binutils-main) that splits up different package versions, depending on gcc version.
we have some more of this (evaluating gcc version or package archs).

It seems that such errors occur since introduction of `spksrc.main-depends.mk`.

needs some investigation (@th0ma7 any idea?)

hgy59:
@mreid-tt you can try to add this to line 8 in spksrc.main-dependency.mk

```
# noting to download
download:
download-all:

```

this works for me in cross/binutil folder.

th0ma7:
Exactly, probably needs extra include such as spkarc.download.mk

hgy59:
@mreid-tt can you try to format an error for github workflow like
```
echo "::error::Failed to download ${package_dir}. Exiting."
```
instead of 
```
echo "Error: Failed to download ${package_dir}. Exiting." >&2
```

I suppose with this we get a colored error log
https://stackoverflow.com/questions/71087094/how-to-make-a-github-workflow-error-message-show-up-in-log-and-annotations

hgy59:
finally found the reference doc on this:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions

@mreid-tt can you test whether it terminates without the "exit 1" command?

mreid-tt:
> @mreid-tt can you test whether it terminates without the "exit 1" command?

I just did with https://github.com/mreid-tt/spksrc/actions/runs/14319904919/job/40134473026 and it doesn't terminate. Still needs the "exit 1".

It does however reveal a lot of errors in downloading packages including a path error for `cross/chromaprint` which results in a 404.

mreid-tt:
For the path errors, I wouldn't want to address them in this PR. Any other changes or are we good to merge?

hgy59:
@th0ma7 I guess we have an issue with chromaprint

We have `cross/chromaprint` and `cross/chromaprint-fftw`.
both have the same source file (and the same PKG_NAME !).

All ffmpeg* depend on `cross/chromaprint-fftw`

chromaprint depends on ffmpeg7 and on cross/chromaprint.

So when building `spk/chromaprint` it will use `cross/chromaprint-fftw` (from ffmpeg7) and never build `cross/chromaprint`.
Am I missing something?

hgy59:
> For the path errors, I wouldn't want to address them in this PR. Any other changes or are we good to merge?

Are the path errors related to chromaprint only?

I suppose to remove `--waitretry=0 --tries=5` in spksrc.download.mk (https://github.com/SynoCommunity/spksrc/pull/6518#issuecomment-2784293106)

hgy59:
@mreid-tt can we have different exit codes for download and checksum errors (like 101 and 102)?

_my final proposal, time for sleep and vacation is coming..._

mreid-tt:
> @mreid-tt can we have different exit codes for download and checksum errors (like 101 and 102)?

Understand the idea but the implementation would be a challenge. Adding a numeric code would not be very unique given all the randomness that could be in download URLs. A simple search for the string "Exiting." would be far more effective.

mreid-tt:
As a final sanity check, I synced the test branch with this PR again to include the removal of the extra wget arguments and re-ran the test: https://github.com/mreid-tt/spksrc/actions/runs/14322938273/job/40143172391

The only difference in the test branch is that it includes the `cross/chromaprint` fix I mentioned above. However, I don't know if it's the time of day but this time each and every download was successful!

th0ma7:
Nicely done!  I didn't had time to look at all the details but that may help solving false positive errors we've been getting.

mreid-tt:
Is this goof to merge?

.github/actions/download.sh
```diff
@@ -10,16 +10,33 @@
...
 if [ -z "${DOWNLOAD_PACKAGES}" ]; then
     echo "===> No packages to download. <==="
 else
     echo "===> Download packages: ${DOWNLOAD_PACKAGES}"
-    for download in ${DOWNLOAD_PACKAGES}
-    do
-        echo "$ make -c ${download} download-all"
-        make -C ${download} download-all
+    for download in ${DOWNLOAD_PACKAGES}; do
+        download_package "${download}" "download-all"
```
th0ma7, line 39: I'm wondering why calling a download sub-function?  Maybe the the set -o pipefail or similar could be added within the download makefile recipe and simply be capturing `$?` upon exit it to validate it did succeed?  That would avoid having yet another download function that calls download functions...

mreid-tt: I'm not sure what you mean by this. Can you elaborate on your solution?

mreid-tt:
I’ve simplified the script to mirror the original more closely while preserving all new functionality. I tested it with `cross/libxml2`, `spk/deluge`, and even a forced checksum failure—which was correctly caught. Please review and let me know if it’s ready to merge.

mreid-tt:
When reviewing the test logs, I wasn’t able to trigger the “Download Python wheel source files” section—it always prints “No wheels to download,” even when I set what seem to be valid Python package targets. Is this section still applicable, or am I missing something in the configuration?

th0ma7:
It was still valid but I haven't tested myself in a while.  It may have gone broken with all the python framework fixes...?

mreid-tt:
> It was still valid but I haven't tested myself in a while. It may have gone broken with all the python framework fixes...?

This function may no longer be in use, but I’m not familiar with any package that previously relied on it to confirm. Another concern is that, if it’s intended to download wheel files directly, that doesn’t appear to align with our current setup—we don’t seem to reference any `*.whl` files directly in the repo, as they’re all built from source. My main question is whether this function is simply downloading wheels without performing a checksum. I’ve never seen a `*.whl` file in the repo accompanied by a digest, which would be necessary for checksum verification.

th0ma7:
Correction, not downloading compiled wheels but rather its source package from pypi.  I'm pretty sure i broke that part with the python updates i did to the framework but the fix is probably dead easy once identified as by default it calls that same function when building wheels as it's not using spksrc.download.mk but rather its wheel related mk.

mreid-tt:
@th0ma7, @hgy59, this PR should be ready to merge.

mreid-tt:
@th0ma7, regarding the download-wheels possibly not working, I was working with ChatGPT and it suggested that this line may contain a bug:

https://github.com/SynoCommunity/spksrc/blob/f46d1e78b697cf32b3028a8d50c4dd16fb79d308/mk/spksrc.wheel.mk#L46

I don't know if this could be a possible cause.

PR #6526, opened 2025-04-11
Python: Update to v3.12.10 and v3.13.3

mreid-tt:
## Description

1. Update python312 to v3.12.10
2. Update python313 to v3.13.3

Fixes # 

## Checklist

### Type of change

- [x] Package update


th0ma7:
There was also a 3.10 update... Although could be into another pr to reduce build time.

mreid-tt:
@th0ma7 Agreed, we can push the updates to 3.10 and 3.11 via a separate PR. Anything else needed before this can be merged and published?

th0ma7:
I haven't looked at the pr content just yet, but normally i take the opportunity to update all wheels from python3xx-wheels packages as well as all crossenv definition files, well at least the default ones.

This ensures keeping up with newer releases.  Likewise, i also update native wheels used in related python.

mreid-tt:
Hey @th0ma7, I've attempted to update the default wheels for python312 as a test — let me know if this looks correct.

One question came up for me: If our goal is always to use the latest package, is there a particular reason we specify explicit wheel versions? Should we follow the approach used with `certifi`, where we only specify the wheel name, allowing it to always fetch the latest available version?

native/python312/Makefile
```diff
@@ -1,5 +1,5 @@
 PKG_NAME = python312
-PKG_VERS = 3.12.9
+PKG_VERS = 3.12.10
```
th0ma7, line 2: Also worth updating `setuptools==78.1.0` although let pip stay at current version as v25.x as a lot more changes that we're not ready yet to handle entirely. That would apply for pyth312+313 (and the other PR for 311)

spk/python312/crossenv/requirements-atom.txt
```diff
@@ -1,5 +1,5 @@
-pip==24.3.1
-setuptools==75.8.0
+pip==25.0.1
```
th0ma7, line 1: I should have thought of this but pip 25.x has a lot of changes in it.  Maybe worth waiting a bit for this one in particular.  Applicable in all crossenv definitions.

th0ma7 (changes requested):
Overall looks good, things missing:
1. updating the wheels in the native makefile, only setuptools being in sight
2. downgrading pip to 24.3.1 for now as version 25.x implies other changes

th0ma7:
> One question came up for me: If our goal is always to use the latest package, is there a particular reason we specify explicit wheel versions? Should we follow the approach used with `certifi`, where we only specify the wheel name, allowing it to always fetch the latest available version?

That is for reproducibility... There are a lot of moving parts but knowing exatly how it was built last time it worked is crucial.  indeed `certifi` is letft using latest version by default, but its the only one we've kept like that and I'd argue should return to a fixed version (although it doesn't have any impact on anything otherwise).

mreid-tt:
@th0ma7 would this PR be ready to merge with the last set of changes?

th0ma7 (approved):
LGTM

PR #6529, opened 2025-04-13
fish: update to version 4.0.2

etcusrvar:
## Description

Update fish to ~4.0.1~ 4.0.2

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


etcusrvar:
I did `make all-supported` locally, and all but one of the targets completed successfully. x64-6.2.4 failed with a link error. I'm opening the PR to see if it might be a local build issue.

hgy59:
@etcusrvar this is not an easy task.

Despite the targets (except x64-6.2.4) seem to build, only the evansport targets have correct binaries.
The binaries of all packages (as in github build artifacts), are built as `ELF 32-bit LSB shared object, Intel 80386`.
i.e. the cross compilation does not work at all!

The linker error `unrecognized relocation (0x2a)` is lucky coincidence (IMHO the linkers for arm7, aarch64 and qoriq should fail too).
Another hint is shown while stripping the binaries:
```
Stripping bin/fish... failed!
Stripping bin/fish_indent... failed!
Stripping bin/fish_key_reader... failed!
```
i.e. the (cross) strip of the toolchains fail when the binaries are not of the expected/supported arch. Unfortunately stripping errors are ignored by spksrc...


There is a hint in the build fish log:
```
# CMake Warning at cmake/FindRust.cmake:719 (message):
#   CMake is in cross-compiling mode, but the cargo target-triple could not be
#   inferred. Falling back to the default host target.  Please consider manually
#   setting `Rust_CARGO_TARGET`.
```

Tried to fix this by definition of:
```
CMAKE_ARGS += -DRust_CARGO_TARGET=$(RUST_TARGET)
```
but this didn't fix the fish binaries.


The only fix that works is using cargo/rust instead of cmake to build fish.
But this will build fish as _experimental static_ version (without any data files), see comment in fish releases:
_The files called fish-static-* are experimental, statically linked builds, for any Linux with the given architecture. Extract and run, and they will extract their data files as needed._

ATM it is not possible to build rust targets by cmake with spksrc framework (/cc @th0ma7 ).

@etcusrvar probably we have to wait until static rust builds are official releases. What do you think?

hgy59:
> Despite the targets (except x64-6.2.4) seem to build, only the evansport targets have correct binaries.
> The binaries of all packages (as in github build artifacts), are built as `ELF 32-bit LSB shared object, Intel 80386`.
> i.e. the cross compilation does not work at all!

corrigenda:
the gibhub build artifacts have binaries of type `ELF 64-bit LSB pie executable, x86-64` 
```
ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=55e3b8726f4816a795cdf4d0b08b5c0362d09d70, for GNU/Linux 3.2.0, not stripped
```

the local builds have `ELF 32-bit LSB shared object, Intel 80386`.
```
ELF 32-bit LSB pie executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, BuildID[sha1]=61af5e683f93ed76d407f0038d455a13735e47ac, for GNU/Linux 3.2.0, not stripped
```

th0ma7:
> ATM it is not possible to build rust targets by cmake with spksrc framework (/cc @th0ma7 ).

From reading on this https://github.com/fish-shell/fish-shell#building it requires setting additional cmake variables:
```
Rust_COMPILER=path - the path to rustc. If not set, cmake will check $PATH and ~/.cargo/bin
Rust_CARGO=path - the path to cargo. If not set, cmake will check $PATH and ~/.cargo/bin
Rust_CARGO_TARGET=target - the target to pass to cargo. Set this for cross-compilation.
```
In our use-case that would be:
```
CMAKE_ARGS += -DRust_COMPILER=$(CARGO_HOME)/bin
CMAKE_ARGS += -DRust_CARGO=$(CARGO_HOME)/bin
CMAKE_ARGS += -DRust_CARGO_TARGET=$(CARGO_BUILD_TARGET)
```
Note that `$(RUST_TARGET)` and `$(CARGO_HOME)` might not be available early and may require to position that line after `spksrc.cross-cmake.mk`.

Lastly, @hgy59 reading on this, there isn't any way to support rust in cmake.  It's rather the contrary to support cmake in rust builds using `cmake-rs` crate.  Therefore the way fish is doing things isn't "standard" and could not be added to our default cmake toolchain file as-is.  I suppose this will get standardized before long by either rust/cmake/meson folks?

etcusrvar:
> i.e. the cross compilation does not work at all!
> ...
> The only fix that works is using cargo/rust instead of cmake to build fish.

I discovered the same not long after making this PR. In fact, I got more correct results using `cross-rust.mk` in place of `cross-cmake.mk` [here](https://github.com/SynoCommunity/spksrc/pull/6529/commits/46eb17bf74e743f37a6bb921f55b52adbb99ba90#diff-58564328fce86ad94d3aa9d3359e51070c14629e85a682b460dae89206673dc9L17). At least with `cross-rust.mk`, the binaries that are built are for the correct arch-- (from my local x86_64 cross-rust build)--
```
$ file work-aarch64-7.1/install/var/packages/fish/target/bin/fish
work-aarch64-7.1/install/var/packages/fish/target/bin/fish: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, not stripped
```

> But this will build fish as experimental static version (without any data files), see comment in fish releases:

This file (from above) doesn't appear to me to be the same as their new experimental static releases-- running `ldd` on the ones for my x64 arch shows they have several system dependencies, and running this on my system does use the external data files installed from the spk.

> ATM it is not possible to build rust targets by cmake with spksrc framework...
Well, at least cross-builds may not be possible ATM. I do have fish 4.0.1 built, installed, and running fine on my x64 using exactly what is included in this PR. (And now 4.0.2 also.)

I've been experimenting with taking pieces of `cross-rust.mk` and `cross-cmake.mk` and trying to combine them together. But not with much success so far. I'll continue trying various suggestions you've each made.

hgy59:
@etcusrvar finally got it working (except for qoriq).

cross/fish/Makefile
```diff
@@ -1,20 +1,76 @@
...
+CMAKE_ARGS = -DWITH_GETTEXT=OFF
+
+# fish >= 4.0.0 requires rust
+# IMPORTANT:
+# fish must still be built with cmake
+# direct builds with cargo/rust (using spksrc.cross-rust.mk) create
+# experimental static versions without data files
+
+# keep cross tools in ENV for cargo to find CC and LN of toolchain
+CMAKE_USE_TOOLCHAIN_FILE = OFF
```
th0ma7, line 41: @hgy59 guessing you mean `LD`?
noting that I have additional changes to do to `CMAKE` to sync with meson and fortran's extra details. Maybe I could add an option to keep `CC` and `LD` defined when setting something such as `CMAKE_USE_RUST` or similar?

hgy59: yes, without this it was using `/usr/bin/ld`
`/usr/bin/gcc` could be overwritten with `ENV += TARGET_CC=$(TC_PATH)$(TC_PREFIX)gcc` but there is only `LD` for the linker.
This was required for pcre2/pcre2-sys and fish-printf.
IMO there is no need for special handling of _CMAKE_USE_RUST_ ...

What bothers me more is the different behaviour of the builds in cross/fish and spk/fish...

mreid-tt:
@th0ma7, is this good to merge and publish? Saw it still pending your review.

PR #6555, opened 2025-05-10
Python: Disable PEP 517 build isolation for the cryptography wheel

mreid-tt:
## Description

**Background**

* **cryptography <45.0.0** declares a PEP 517 build backend (`maturin`) in its `pyproject.toml` with `requires = ["maturin>=1,<2", …]`.
* Under default build isolation, pip creates a throw-away venv and resolves `maturin>=1,<2`, picking up maturin 1.8.4 (the latest <2) along with its bundled `setuptools-rust`.
* That freshly installed `setuptools-rust` no longer accepts the `env=` argument, so building cryptography fails with:

  ```
  TypeError: RustBin.__init__() got an unexpected keyword argument 'env'
  ```

**Change**

```diff
+ # Disable build isolation for cryptography < 45.x
+ ifeq ($(WHEEL_NAME),cryptography)
+ EXTRA_PIP_ARGS += --no-build-isolation
+ endif

include ../../mk/spksrc.python-module.mk
```

**Why**
Disabling build isolation for just the `cryptography` wheel forces pip to install it *into* our outer crossenv — where we have pinned `maturin==1.8.1` and `setuptools-rust==1.10.2` — so the build uses those known-good versions and avoids the `env=` mismatch.

**Testing**

* Confirmed that `cryptography==44.0.0` now compiles without the `RustBin.__init__()` error.
* Verified that all other wheels continue to install under normal PEP 517 isolation.

Fixes https://github.com/SynoCommunity/spksrc/issues/6421#issuecomment-2868571309

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/python/pip/Makefile
+++ b/python/pip/Makefile
@@ -15,7 +15,7 @@ LICENSE  = MIT
 
 # Disable build isolation for cryptography < 45.x
 ifeq ($(WHEEL_NAME),cryptography)
-  EXTRA_PIP_ARGS += --no-build-isolation
+EXTRA_PIP_ARGS += --no-build-isolation
 endif
 
 include ../../mk/spksrc.python-module.mk
--- a/python/pip/Makefile
+++ b/python/pip/Makefile
@@ -13,4 +13,9 @@ HOMEPAGE = https://pip.pypa.io
 COMMENT  = The PyPA recommended tool for installing Python packages
 LICENSE  = MIT
 
+# Disable build isolation for cryptography < 45.x
+ifeq ($(WHEEL_NAME),cryptography)
+  EXTRA_PIP_ARGS += --no-build-isolation
+endif
+
 include ../../mk/spksrc.python-module.mk
--- a/spk/python311/Makefile
+++ b/spk/python311/Makefile
@@ -4,6 +4,8 @@ SPK_VERS_MAJOR_MINOR = $(word 1,$(subst ., ,$(SPK_VERS))).$(word 2,$(subst ., ,$
 SPK_REV = 10
 SPK_ICON = src/python3.png
 
+# change to trigger build
+
 # Platform "powerpc-none-linux-gnuspe" with compiler "gcc" is not supported by the
 # CPython core team, see https://peps.python.org/pep-0011/ for more information.
 # And compiler must support std=c++11 (OlD_PPC_ARCHS fail, but ARMv5_ARCHS have no issue).
--- a/spk/python311/Makefile
+++ b/spk/python311/Makefile
@@ -4,8 +4,6 @@ SPK_VERS_MAJOR_MINOR = $(word 1,$(subst ., ,$(SPK_VERS))).$(word 2,$(subst ., ,$
 SPK_REV = 10
 SPK_ICON = src/python3.png
 
-# change to trigger build
-
 # Platform "powerpc-none-linux-gnuspe" with compiler "gcc" is not supported by the
 # CPython core team, see https://peps.python.org/pep-0011/ for more information.
 # And compiler must support std=c++11 (OlD_PPC_ARCHS fail, but ARMv5_ARCHS have no issue).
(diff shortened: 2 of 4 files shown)


mreid-tt:
@hgy59, @th0ma7 this patch for the cryptography wheel is ready to merge. I've removed the test code which showed that it works in https://github.com/SynoCommunity/spksrc/actions/runs/14946144427.

th0ma7:
@mreid-tt worry to say but this solution does not work - you've tested python builds but it should have been python-wheels builds which includes building cryptography.  Adding that option to the `pip` cross file won't change the build process but rather include that `--no-build-isolation` only when "building" pip... which isn't applicable.

The `--no-build-isolation` is setup at multiple places and definitively needs alignment and cleanups in our mk directory, but not this way sadly.  For the moment I'd rather go back in previous commits and try to identify if anything changed since causign it to fail - or it is par of the autodected dependencies in the crossenv where one changed version and screwed things up?

Also note, I'm traveling for work and won't be much of help over the next week or so.

mreid-tt:
Hey @th0ma7, thanks for taking a look and providing more clarity on the testing approach. I’ll close this PR and explore an alternative solution.

hgy59:
> @mreid-tt worry to say but this solution does not work - you've tested python builds but it should have been python-wheels builds which includes building cryptography.

My reported issue **was** for python builds (when cryptography is built for the default cross env) and not for python-wheels builds.
Because it failed to build the crossenv, it is not possible to build any wheel (for python, python-wheels, homeassistant...).

So this PR fixed the original issue, and the build of cryptography wheels would be a different story.

th0ma7:
ah! that's interesting... so when `pyhton/pip` kicks-off it triggers a crossenv creation, using default config.  During the default crossenv build it eventually fails on `cryptography`.

So, since `python/pip` has `EXTRA_PIP_ARGS += --no-build-isolation` the `crossenv-install-%:` recipe in `spksrc.crossenv.mk`  does make use of it:
```
$(RUN) \
PATH=$${PATH} \
$$(which $(if $(filter wheelhouse,$(WHEEL_TYPE)),cross,$(WHEEL_TYPE))-python) -m pip install \
--cache-dir $(PIP_CACHE_DIR) \
$(EXTRA_PIP_ARGS) \
--disable-pip-version-check \
$(WHEEL_NAME)==$(WHEEL_VERSION) ;
```
I would then try to add `--no-build-isolation` by default to `crossenv-install-%:`  recipe in `spksrc.crossenv.mk` and see if that works out?  such as:
```
$(RUN) \
PATH=$${PATH} \
$$(which $(if $(filter wheelhouse,$(WHEEL_TYPE)),cross,$(WHEEL_TYPE))-python) -m pip install \
--cache-dir $(PIP_CACHE_DIR) \
$(EXTRA_PIP_ARGS) \
--no-build-isolation \
--disable-pip-version-check \
$(WHEEL_NAME)==$(WHEEL_VERSION) ;
```
Still I wonder, why now?  what made it to break?

hgy59:
> I would then try to add `--no-build-isolation` by default to `crossenv-install-%:` recipe in `spksrc.crossenv.mk` and see if that works out? such as:
> 

Good catch: this fixes the build of python/cryptography (version 41.0.3).

But the next wheel in python313-wheels is python/pendulum and this failed with the same error (`TypeError: RustBin.__init__() got an unexpected keyword argument 'env'`)

Then I added `--no-isolation` to `build_python_wheel_target` in `spksrc-python-wheel.mk` 
This fixes python/pendulum (and all other python/* where built too: numpy*, pillow, pydantic_core, scipy)
Also the wheels defined in requirements-crossenv.txt are built (cryptography==44.0.0, ...)
=> python313-wheels for arch-x64-7.1 sucessfully built.

hgy59:
> Still I wonder, why now? what made it to break?

As @mreid-tt found, it is related to `maturin==1.8.4` that was released yesterday.

AFAICT "with isolation", pip creates an isolated venv and installs all required dependencies for the build of wheels.
With "no isolation" pip does not install build dependencies and takes what is available (i.e. the user has to prepare the correct environment).
So "with isolation" `maturin==1.8.4` is used for the build (that fails, not only for cryptography) and "without isolation" it takes `maturin==1.8.1` of our crossenv.

@th0ma7 I guess we don't need isolation anymore, since you introduced the default and specific crossenv to build wheels.

PR #6556, opened 2025-05-11
Python: Disable PEP 517 build isolation

mreid-tt:
## Description

This change builds on #6555 by adding flags to disable PEP 517 build isolation for all wheel builds. It has been verified when building both the core Python interpreter and the python-wheels package.

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.crossenv.mk
+++ b/mk/spksrc.crossenv.mk
@@ -230,6 +230,7 @@ crossenv-install-%:
 	   $$(which $(if $(filter wheelhouse,$(WHEEL_TYPE)),cross,$(WHEEL_TYPE))-python) -m pip install \
 	   --cache-dir $(PIP_CACHE_DIR) \
 	   $(EXTRA_PIP_ARGS) \
+	   --no-build-isolation \
 	   --disable-pip-version-check \
 	   $(WHEEL_NAME)==$(WHEEL_VERSION) ; \
 	$(RUN) \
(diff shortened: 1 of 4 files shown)


mreid-tt:
@hgy59, @th0ma7 based on both your feedback, this PR was created to address the Python wheel building issue raised in https://github.com/SynoCommunity/spksrc/issues/6421#issuecomment-2868571309. The build was confirmed to work in https://github.com/mreid-tt/spksrc/actions/runs/14950765542/job/41999897301 before rolling back the test code.

th0ma7 (approved):
This LGTM

hgy59 (approved):
LGTM

I already tested this on an own branch 😄

hgy59:
@th0ma7, @mreid-tt it seems that python/pip is not needed at all. I am drafing a PR to remove python/pip.

th0ma7:
That's been a long standing issue where pip gets installed as a module embedded into python blocking crossenv pip to get installed.

When i started working on refreshing python wheels there was also setuptools and wheel.  I was able to get those out.  I'm glad pip can be next from a package building standpoint.

But be careful of one thing, pip as a module makes it to provide a pip in python by default at package install time.  Removing it will require changing the setup script to download a pip.  Or, to find a way to setup the system to initiate pip from the pip wheel that we provide.

You'll notice that we provide a pip wheel but it never gets installed on the nas as it finds the packaged pip module already available (like in crossenv when building).  So the setup script will have to work around that and initiate a pip install without pip available.

Further thinking, there may be one, obsolete but that could suffice provided by dsm that we could use for that, just to install the initial pip wheel?

hgy59:
> Python comes with the ensurepip module that already contains the pip-wheel, so nothing to download at installation time.

ensure pip does not create `/target/bin/pip` it only creates `/target/bin/pip3`  and for python312 `/target/bin/pip3.12` .

@th0ma7 shall I adjust the installer scripts to use pip3 or copy pip3 to pip after installation with ensurepip?
since the package installs python3 and python3.12 but not python (and pydoc3, pydoc3.12 but not pydoc) I would prefer pip3 too.

th0ma7:
I'm ok with pip3, and probably something where a symlink could suffice...?

hgy59:
> I'm ok with pip3, and probably something where a symlink could suffice...?

For backwards compatibility (already published packages using pip in `install_python_wheels`) we need both.
but it is easy to get `pip` too:
```
${SYNOPKG_PKGDEST}/bin/python3 -m ensurepip --default-pip
```

PR #6557, opened 2025-05-11
duplicity: fix build

hgy59:
## Description

Finally this PR provides a redesign of all python packages to install pip with python module `ensurepip` instead of prebuilt pip module.
The main reason to drop the pip module is, that it is not possible to built the module for pip >= 24.0.
The ignition for this was duplicity, that requires pip >= 24.0 to build and this fails since we have #6556.

follow up to #6556.

- drop obsolete python/pip
- use custom crossenv to build duplicity
- install pip with `python3 -m ensurepip --default-pip`
- use pip3 in `install_python_wheels()` 

comments:
- https://github.com/SynoCommunity/spksrc/pull/6556#issuecomment-2870262435

Fixes # 

## Checklist

### Type of change

- [x] Bug fix


spk/python312/crossenv/requirements-duplicity.txt
```diff
@@ -0,0 +1,10 @@
+pip==24.3.1
+setuptools==78.1.0
+wheel==0.45.1
+#
+build:Cython==3.0.11
+build:pip-tools==7.4.1
+#
+cross:build==1.2.2
+cross:cffi==1.17.1
+cross:pipx==1.7.1
```
mreid-tt, line 10: I'm curious about how you derive this file. I don't believe we have this documented fully. If I look at the "_pyproject.toml_" for duplicity, I see the following:

> [build-system]
> requires = [
>     "build>=1.1",
>     "pip>=24.0",
>     "pipx>=1.5",
>     "packaging>=20.0",
>     "setuptools>=78.1.0",
>     "wheel>=0.42.0",
> ]
> build-backend = "setuptools.build_meta"

Would this be used to create the above? If so, how do you determine what should be inserted as-is and what should have a `build:` or a `cross:` prefix?

th0ma7: I was thinking of that and i believe @hgy59 also proposed that.  Issue i foud is that would genereate a crossenv for every wheel builds if we were to automate the use of that dependency from the toml file.

And we cannot update a crossenv on demand as some wheels will have conflicting dependencies.

th0ma7:
@hgy59 please read my post on the refered issue with regards to pip implications at package install time.

mreid-tt:
Hey @hgy59, although not an exhaustive test, I downloaded the `python312` package from this build and successfully installed it on a clean DSM 6.2 environment. I was then able to install the existing `duplicity` package from our SynoCommunity catalog without any issues. Overall, the changes to the Python build appear to maintain existing functionality.

th0ma7:
@hgy59 and @mreid-tt good job!  I'm really glad that this python portion of the framework continues to mature progressively with your help :)  There are still a bit more cleanups to do but things are moving along nicely :rocket:

spk/python312/src/requirements-pure.txt
```diff
@@ -1,28 +1,28 @@
...
-pip==24.3.1
-setuptools==78.1.0
+#pip==25.0.1    # we use module ensurepip to install latest version included
+setuptools==80.4.0
 wheel==0.45.1
 
 # Always use latest version available
-certifi==2025.1.31
+#certifi==2025.4.26
+certifi
```
mreid-tt, line 9: I thought @th0ma7 indicated that we should hardcode versions for the build so that the output is predictable? This is why I started changing to specific versions of `certifi` with the last update.

th0ma7: certify is an exception as it is really the one that doesn't add any value to enforce a version since we always want the latest version no matter what.

PR #6579, opened 2025-05-20
Py310 311 update

th0ma7:
## Description

Update python 3.10 and 3.11 to latest point releases.

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


hgy59:
@th0ma7 I already startet with this updates shortly...

At least I fixed OLD_PPC_ARCHS for 3.10 (does not support lto).

All my changes in the following patch file.
[update-python-3.10.15_to_3.10.17.patch](https://github.com/user-attachments/files/20325720/update-python-3.10.15_to_3.10.17.patch)

hgy59:
And this is my update of requirements-crossenv.txt of python310
- fix the comment on how to activate the build
- update example requirements
- add remarks that Pillow==10.4.0 fails to build (former version is vulearable, see CVE-2023-44271)

[update_python310_requirements-crossenv.patch](https://github.com/user-attachments/files/20325812/update_python310_requirements-crossenv.patch)

th0ma7:
@hgy59 I believe I have included your code into python310 and also decided to standardize it by creating a `python310-wheels` package.  If you have a moment to try things up would be appreciated, in particular for old ppc arch (`ppc853x-5.2`).

EDIT: I've started testing things up.  python310 builds fine for old ppc.  python310-wheels fails on rust.  Would the qoriq rustc work for old ppc as it can't find a rust compiler.

Further looked at the logs, no rust toolchain is defined for this old arch therefore it takes x64 default:
```
info: downloading installer
info: profile set to 'default'
info: default host triple is x86_64-unknown-linux-gnu
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
info: latest update on 2025-05-15, rust version 1.87.0 (17067e9ac 2025-05-09)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
info: default toolchain set to 'stable-x86_64-unknown-linux-gnu'
```
Then later when building cryptography:
```
error: override toolchain 'stable-1.82.0-powerpc-unknown-linux-gnuspe' is not installed: the RUSTUP_TOOLCHAIN environment variable specifies an uninstalled toolchain

    =============================DEBUG ASSISTANCE=============================
    If you are seeing a compilation error please try the following steps to
    successfully install cryptography:
    1) Upgrade to the latest pip and try again. This will fix errors for most
       users. See: https://pip.pypa.io/en/stable/installing/#upgrading-pip
    2) Read https://cryptography.io/en/latest/installation/ for specific
       instructions for your platform.
    3) Check our frequently asked questions for more information:
       https://cryptography.io/en/latest/faq/
    4) Ensure you have a recent Rust toolchain installed:
       https://cryptography.io/en/latest/installation/#rust
    
    Python: 3.10.17
    platform: Linux-powerpc-none-linux-gnuspe-with
    pip: 24.3.1
    setuptools: 75.4.0
    setuptools_rust: n/a
    rustc: n/a
    =============================DEBUG ASSISTANCE=============================
    
error: can't find Rust compiler
```
Thus wonder if qoriq rust would work?

hgy59:
> Thus wonder if qoriq rust would work?

It does not. 
Can't remeber the details (currently out of my dev env).

hgy59:
@th0ma7 while working on home assistant update, I found, that the requirements-abi3.txt are not needed anymore.
The related wheels can be defined in requirements-crossenv.txt and are correctly built as abi3 wheels. This looks like a fix by the redesign of the wheel build process.
For python312/313-wheels I prepared a dedicated PR (including consolidation of python313/crossenv/requirements*).
But I guess for python 310/311 we could eliminate the requirements-abi3.txt now.

Some more suggestions:
- For the python*-wheel packages we should only create one version of each wheel (regarding /python/* and requirements-crossenv.txt). Otherwise it is not possilbe to install the wheels (and it shortens the build process).
- When all python*-wheel packages are merged without requirements-abi3.txt we can remove the related code in *.mk file(s)
- There is an issue with wheelhouse:* dependencies in crossenv/requirements-*. For home assistant I need python/scipy_ha with python/numpy_ha dependency, but numpy in crossenv/requirements-scipy.txt is a different version (just a hint for future redesigns like package specific python crossenv overrides).

th0ma7:
> while working on home assistant update, I found, that the requirements-abi3.txt are not needed anymore. The related wheels can be defined in requirements-crossenv.txt and are correctly built as abi3 wheels. This looks like a fix by the redesign of the wheel build process.

@hgy59 unsurprising, `pip` is a fast moving target and it isn't easy to keep track of all evolutions while keeping the ability to build older wheels.  Definitively something to address which will simplify the code greatly (and related wiki page).  Why don't we create an issue for the few tidbits we'd like to address?

> For python312/313-wheels I prepared a dedicated PR (including consolidation of python313/crossenv/requirements*). But I guess for python 310/311 we could eliminate the requirements-abi3.txt now.

I actually wonder if we could not have a `spksrc/python/crossenv/requirement*`.  Actually we could have both such as, it looks for `python313/crossenv/requirements*` then fallback to `spksrc/python/crossenv/requirement*`.  With luck the python version specific ones won't be needed at all, but functionality would still exists if we ever need it?

> Some more suggestions:
> 
> * For the python*-wheel packages we should only create one version of each wheel (regarding /python/* and requirements-crossenv.txt). Otherwise it is not possilbe to install the wheels (and it shortens the build process).

Yes, no, maybe... Fair point with regards to the installation, although the intent here with the `python3*-wheels` really is just to confirm the ability that these wheels can still be built on that python version.  Thus I did added a few versions of a same wheel where I had found issues in the past (or such as numpy which is really tricky, but looking like solved now with meson-python builds).

> * When all python*-wheel packages are merged without requirements-abi3.txt we can remove the related code in *.mk file(s)

Totally.  May actually need to be adjusted to have the ability to select abi3 python level.

> * There is an issue with wheelhouse:* dependencies in crossenv/requirements-*. For home assistant I need python/scipy_ha with python/numpy_ha dependency, but numpy in crossenv/requirements-scipy.txt is a different version (just a hint for future redesigns like package specific python crossenv overrides).

Interesting... unsure how to deal with that simply while avoiding code/configuration duplications... Wonder (and just a thought) if we could have a `wheelhouse/dependency` superseeding a `cross|python/dependency`... all in all, interesting challenge... probably for early fall timeframe.

I'll add as well, currently we are using old pip arguments which will be superseeded soon with the following:
```
# configure part of the wheel process uses:
#     pip: --config-settings=setup-args='<value>'
#   build: -Csetup-args='<value>'
#
# install using python wheel process uses:
#     pip: --config-settings=install-args='<value>'
#   build: -Cinstall-args='<value>'
```
We need to migrate to those before long.  On that I do have reusable code in `spksrc.python-wheel-meson.mk` made for using with `build` rather than `pip` but the core principle remains the same.  On that matter, for already downloaded wheel sources I wonder if we shouldn't migrate to `build` and only use `pip` for URL based sources.

PR #6447, opened 2025-02-10
Beets: Update to v2.3.1 and migrate to Python 3.12

mreid-tt:
## Description

This PR includes the following:

1. Update to beets 2.3.1
2. Migrate to Python 3.12

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #6603, opened 2025-06-08
glib-latest: Update version from 2.78.4 to 2.84.2

th0ma7:
## Description

glib-latest: Update version from 2.78.4 to 2.84.2

Relates to https://github.com/SynoCommunity/spksrc/pull/6447 in particular https://github.com/SynoCommunity/spksrc/pull/6447#discussion_r2134270892

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


mreid-tt:
@th0ma7, thanks for this. I suspect that the code in `cross/glib-latest/patches/hi3535/001-fix-build-SYS_sched_getattr.patch` is probably incompatible with this update.

th0ma7:
Issue is interesting, basically it requires a newer meson... current is `1.0.1` and needed is `1.2` or `1.4` depending of the 1.8x version of glib tested.  I'd recomment using https://packages.debian.org/bookworm-backports/meson version being 1.7

th0ma7:
@hgy59 There's a `synocli-file` build issue on armv5, wonder if you could take a look?  I'll investitage the imagemagick/ffmpeg7 build issue for hi3535 first.

mreid-tt:
Hey @th0ma7, I was looking at the note in https://github.com/SynoCommunity/spksrc/pull/6603/commits/8a95b453fc3694d8c72a8deb97d24127a5f48da8 and put together [this spreadsheet](https://docs.google.com/spreadsheets/d/1qbHgPpjnTVM_8hIRWznHLzMsJVtldltT/edit?usp=sharing&ouid=105344619544753364529) based on the DSM tool chain download folders. Would `glib` on other kernel 3.2 platforms also work?

th0ma7:
> Hey @th0ma7, I was looking at the note in https://github.com/SynoCommunity/spksrc/pull/6603/commits/8a95b453fc3694d8c72a8deb97d24127a5f48da8 and put together [this spreadsheet](https://docs.google.com/spreadsheets/d/1qbHgPpjnTVM_8hIRWznHLzMsJVtldltT/edit?usp=sharing&ouid=105344619544753364529) based on the DSM tool chain download folders. Would `glib` on other kernel 3.2 platforms also work?

Simple answer is i don't know... I recall doing the original split for glib latest vs older... I tried finding the reasoning for the kernel limitations i had setup but i could not find anything.  Actually wonder if this is not rather a glibc limitation rather than kernel, or if this was circumvented since in glib?

... And so i was surprised to confirm it does build for older kernel such as for evansport... Definitely something to figure out before pushing this in and finding issues later on.  There may be hints buried in one of the commits related to the inclusion of this glib split?

hgy59:
> @hgy59 There's a `synocli-file` build issue on armv5, wonder if you could take a look?

Already noticed this. The problem is, that rust builds that previously worked without issues begin to fail. It looks like there is no "lock" or "freeze" and dependencies are updated as soon as available and rust needs an update too (and ARMv5 is limitted to rust 1.77.2)
I only see two solutions
- drop all rust based tools in synocli-file for ARMv5 (and may be qoriq)
- include such tools in binary format 

Currently I ignore such build failures, but for the next synocli-file update we need a solution.
The best solution would be to avoid the dependency updates for rust builds, but I have no idea whether this would be possible.

th0ma7:
> Currently I ignore such build failures, but for the next synocli-file update we need a solution.

@hgy59 Thnx for sharing this, I'll avoid those error completely for this PR.

> The best solution would be to avoid the dependency updates for rust builds, but I have no idea whether this would be possible.

no clue either, can we freeze crates under rust?  or set max crates version or similar, no clue...

hgy59:
@th0ma7 the downloads fail in the build workflow.
The download from https://ftp.gnu.org is very slow, but worked on a second try.
But now the download of libsodium fails (manual downloads fail too, libsodium.org currently not reachable by https).

hgy59:
@mreid-tt related to #6449 I finally changed the source site of libsodium to github since libsodium.org is down.

hgy59:
@th0ma7 I have created a dedicated PR for libsodium source fix (to avoid python* dependencies)

th0ma7:
woot! success at last!   @mreid-tt it did took me longer than expected to fix harfbuzz for hi3535, but's now closed and getting all the green flags!  @hgy59 thnx for fixing the libsodium issue - as previously mentioned I won't bother with synocli-file for armv5 as known to fail.

Merging :rocket:

hgy59:
> Already noticed this. The problem is, that rust builds that previously worked without issues begin to fail. It looks like there is no "lock" or "freeze" and dependencies are updated as soon as available and rust needs an update too (and ARMv5 is limitted to rust 1.77.2)

The same issue exists for bandwhich in synocli-monitor:
```
 $ make arch-88f6281-6.2.4 -C ./spk/synocli-monitor
error: failed to compile `bandwhich v0.22.2 (/github/workspace/spk/synocli-monitor/work-88f6281-6.2.4/bandwhich-0.22.2)`, intermediate artifacts can be found at `/github/workspace/spk/synocli-monitor/work-88f6281-6.2.4/bandwhich-0.22.2/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  package `icu_collections v2.0.0` cannot be built because it requires rustc 1.82 or newer, while the currently active rustc version is 1.77.2
  Either upgrade to rustc 1.82 or newer, or use
  cargo update icu_collections@2.0.0 --precise ver
```

PR #6599, opened 2025-06-05
Multipe Python updates

th0ma7:
## Description

Multipe Python updates

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6534, opened 2025-04-15
Python: Rename numpy package and correct spelling errors for 312/313-wheels

mreid-tt:
## Description

This PR introduces the following changes:

1. Renames the package `numpy_v2` to `numpy` to enable direct inclusion in Python builds.  
2. Fixes minor spelling errors in `python31[2,3]-wheels` to ensure build validation.

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Includes small framework changes

PR #6611, opened 2025-06-12
Allow using PKG_REAL_NAME or PKG_NAME when defining WHEEL_NAME

th0ma7:
## Description

Allow using PKG_REAL_NAME or PKG_NAME when defining WHEEL_NAME.

Issue is that some `python/numpy*` entries define `WHEEL_NAME` within the Makefile.  In some cases it mixup the crossenv installation whereas the name and version of the wheels to be installed gets mixed up.

This proposal avoids this entirely by : 
1. satisfying the need for independent `PKG_NAME` within a build with multiple common dependencies
2. ensuring `WHEEL_NAME` gets define only within mk files

In turns, this will now allow creating crossenv requirement definitions matching python/* naming.

EXTRA: While review `python/numy*` files I've taken care to update to latest versions where applicable.

Fixes :
- https://github.com/SynoCommunity/spksrc/pull/6447#issuecomment-2947384651
- https://github.com/SynoCommunity/spksrc/pull/6447#issuecomment-2948669168
- https://github.com/SynoCommunity/spksrc/pull/6447#issuecomment-2950817539

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


mreid-tt:
@th0ma7, should we also remove the `mk/*` changes introduced in #6534 since we are no longer using `WHEEL_NAME` in the `python/*` files?

th0ma7:
> @th0ma7, should we also remove the `mk/*` changes introduced in #6534 since we are no longer using `WHEEL_NAME` in the `python/*` files?

thnx, I've already changed that, but atm it allows `WHEEL_NAME`, `PKG_REAL_NAME` then `PKG_NAME`.  Although looking at the code I do believe `WHEEL_NAME` can be stripped from that list.

Let me do a few rounds of testing to confirm.

th0ma7:
Note that a solution is needed as I cannot currently generate a `numpy-1.26.4` crossenv based on `python/numpy_v1` as things gets all mixed up due to the `WHEEL_NAME` pre-definition.  So a more permanent solution is needed before I can close https://github.com/SynoCommunity/spksrc/pull/6599

hgy59:
my simple solution would be to avoid PKG_NAME that does not match the WHEEL_NAME and to forbid to depend on different versions of the same wheel.
but that's not what @th0ma7 likes, ...

th0ma7:
> my simple solution would be to avoid PKG_NAME that does not match the WHEEL_NAME 

@hgy59 issue is that using `WHEEL_NAME` in cross bleeds into sub-wheel processing and gets into crossenv wheel processing... 

> and to forbid to depend on different versions of the same wheel. but that's not what @th0ma7 likes, ...

The main reason behind `python31*-wheels` package is exactly that: ability to test multiple versions of the same wheel.  

But open otherwise open to suggestions?

EDIT: And current does not manage creating crossenv to specific `python/*` packages, at least the ones that have `WHEEL_NAME` in it...

th0ma7:
Thnx @hgy59 for fixing that!  I believe this was hidden due to the variable mixup.

hgy59:
> Thnx @hgy59 for fixing that! I believe this was hidden due to the variable mixup.

you merged updated beets into this branch and got new crossenv files (lap, pygobject).

th0ma7:
Exactly, this behavior was not seen previously on the beets branch... While now it pops-up.  Unless there is something else i don't get?

hgy59:
> Exactly, this behavior was not seen previously on the beets branch... While now it pops-up. Unless there is something else i don't get?

It has nothing to do with the beats branch.
You updated `python/numpy` and adjusted the `python312/crossenv/` files at the moment of the update.
Then you got new crossenv files (from the main branch) one with older setuptools and one with outdated numpy version.

Since we do not (or can't) use variables for python/module versions referenced in crossenv files, it will stay fragile when updating python/* versions.

I already mentioned a related issue when HA depends on python/scipy, but needs a different version of numpy (python/numpy_ha) and finally the crossenv/scipy.txt references a numpy version that does not exist when building HA.

spk/python312/crossenv/requirements-pygobject.txt
```diff
@@ -1,5 +1,5 @@
 pip==24.3.1
-setuptools==73.0.1
+setuptools==78.1.0
```
mreid-tt, line 2: I don't know if this will work as the project's [pyproject.toml](https://gitlab.gnome.org/GNOME/pygobject/-/blob/main/pyproject.toml) states the following:

>     # Setuptools 74 removed msvccompiler, required by g-ir-scanner (<1.82.0)
>     # See https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/515
>     "setuptools<74",

This is the reason that I included `setuptools==73.0.1`

th0ma7:
> It has nothing to do with the beats branch. You updated `python/numpy` and adjusted the `python312/crossenv/` files at the moment of the update. Then you got new crossenv files (from the main branch) one with older setuptools and one with outdated numpy version.

I had forgot that numpy bump from 2.2.4 to 2.2.6... Although the setuptools update came in from pillow failing to build on python 3.10 and 3.11 needing a newer version.

> Since we do not (or can't) use variables for python/module versions referenced in crossenv files, it will stay fragile when updating python/* versions.

That is a relly good point.  Wonder how we could do that... I did added the following in numpy's makefile but that's far from being sufficient:
```
# Reminder to update wheelhouse:numpy==<version>
# in python31*/crossenv/requirements-scipy.txt
```
I wonder if we could simply use `wheelhouse:numpy` whtout any version and ensure pip uses what ever version found it that directory?  Or have a `wheelhouse:numpy==<wheelhouse>` which would lead into capturing said version available in wheelhouse then installing it...

In any event doing nothing about it will become a nightmare when trying to update wheels later on.

> I already mentioned a related issue when HA depends on python/scipy, but needs a different version of numpy (python/numpy_ha) and finally the crossenv/scipy.txt references a numpy version that does not exist when building HA.

Yup.

Unrelated, one thing I enforced in my other PR https://github.com/SynoCommunity/spksrc/pull/6599 is making sure every python version uses the exact same `crossenv/*` definitions.  My hope from here is to later on be able to move those definitions elsewhere so they are agnostic to python's version.

spk/python312/crossenv/requirements-pygobject.txt
th0ma7: humm, it did worked... but `mscccompiler` isn't it Microsoft C++ (MSVC) , which in turns would be unrelated to our environment as we're using gcc...  Although there isn't any check for that.

@hgy59 was the setuptools version update also mandatory for pygobject to work?

th0ma7: @mreid-tt I gave it a shot at downgrading back setuptools to its previous version for pygobject and it worked out ok.  I suggest leaving things as-is for now and reconvene later if something comes.

PR #6654, opened 2025-07-19
TVH update to July 17th commit da9fa60 hash

th0ma7:
## Description

TVH update to July 17th commit da9fa60 hash.  Also includes update of zap2epg to version 3.3

Fixes n/a

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


cross/libssh2/Makefile
```diff
@@ -15,5 +15,14 @@ CMAKE_ARGS += -DBUILD_SHARED_LIBS=ON
...
 CMAKE_ARGS += -DBUILD_TESTING=OFF
+ifneq ($(wildcard $(OPENSSL_STAGING_PREFIX)),)
+CMAKE_ARGS += -DOPENSSL_INCLUDE_DIR=$(OPENSSL_STAGING_PREFIX)/include
+CMAKE_ARGS += -DOPENSSL_SSL_LIBRARY=$(OPENSSL_STAGING_PREFIX)/lib/libssl.so
+CMAKE_ARGS += -DOPENSSL_CRYPTO_LIBRARY=$(OPENSSL_STAGING_PREFIX)/lib/libcrypto.so
+else
+CMAKE_ARGS += -DOPENSSL_INCLUDE_DIR=$(STAGING_INSTALL_PREFIX)/include
+CMAKE_ARGS += -DOPENSSL_SSL_LIBRARY=$(STAGING_INSTALL_PREFIX)/lib/libssl.so
+CMAKE_ARGS += -DOPENSSL_CRYPTO_LIBRARY=$(STAGING_INSTALL_PREFIX)/lib/libcrypto.so
+endif
```
th0ma7: @hgy59 there's only a few cases where this is needed (`mariadb-connector-c`, `znc` and now `libssh2`) although wonder if this would not be worth adding to the cmake definition file to avoid case-by-case declarations when used from `spksrc.python.mk`.... Also to note, I'll move this in its own PR.

mk/spksrc.python.mk
```diff
@@ -56,15 +56,19 @@ PYTHON_DEPENDS_EXCLUDE = bzip2 xz zlib
...
 ifneq ($(findstring $(ARCH),$(ARMv5_ARCHS) $(OLD_PPC_ARCHS)),$(ARCH))
 BUILD_DEPENDS += cross/$(PYTHON_PACKAGE)
 endif
 endif
 
+ifneq ($(FFMPEG_PACKAGE),)
+include ../../mk/spksrc.ffmpeg.mk
+else
 include ../../mk/spksrc.spk.mk
+endif
```
th0ma7, line 71: @hgy59 if I recall you had a case, presumably homeassistant, where some python packages needed to bind against both ffmpeg and python libraries as well?  With the few changes in here you can now declare: 
```
FFMPEG_PACKAGE = ffmpeg7
PYTHON_PACKAGE = python312
```
And that should do the trick to provide your environment with the needed `*_FLAGS` and pkgconfig symlinks.

hgy59: @th0ma7 thanks, that was for the av wheel, but it needs its own python build evnironment to build the wheels and I didn't succeed to build it (with the original build environment) for armv7.
For supported archs the HA packages use prebuilt av wheel from index.

BTW home assistent (core) will not be supported anymore from version 2025.12.

PR #6669, opened 2025-07-29
[new package] ddns-go

GrandDuke1106:
## Description

This is a new package. Its upstream is <https://github.com/jeessy2/ddns-go>.

> Automatically obtain your public IPv4 or IPv6 address and resolve it to the corresponding domain name service.

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update


GrandDuke1106:
It's just that Synology seems to be unable to directly change Cloudflare's DNS records. And the logo image is something I "drew" myself.

mreid-tt:
@GrandDuke1106 found a [nicer](https://img.alicdn.com/imgextra/i1/76498269/O1CN01ucUOed2AxG5cqCrUA_!!76498269.webp) logo for you and pushed to your PR.

cross/ddns-go/Makefile
```diff
@@ -0,0 +1,33 @@
...
+LICENSE = MIT
+
+COMPILE_TARGET = ddns-go_compile_target
+INSTALL_TARGET = ddns-go_install
+
+include ../../mk/spksrc.cross-go.mk
+
+.PHONY: ddns-go_compile_target
+ddns-go_compile_target:
+	@$(RUN) $(MAKE) VERSION=$(shell git describe --tags `git rev-list --tags --max-count=1` 2>/dev/null || git rev-parse --short HEAD) BUILD_TIME="$$(date -u +'%Y-%m-%dT%H:%M:%SZ')" build
```
hgy59: We can't take the version from git, because we build from source within the git repository of spksrc.
currently the version looks like this:
```
./ddns-go -v
b8362bd
```

An official release shows
```
 ./ddns-go -v
v6.12.0
```

We could have the same with:

```suggestion
ddns-go_compile_target:
	@$(RUN) $(MAKE) VERSION=v$(PKG_VERS) BUILD_TIME="$$(date -u +'%Y-%m-%dT%H:%M:%SZ')" build
```

hgy59:
@GrandDuke1106  please add line breaks at the end of all new text files.

spk/ddns-go/src/service-setup.sh
```diff
@@ -0,0 +1,6 @@
+DDNS_GO="${SYNOPKG_PKGDEST}/bin/ddns-go -l :9876 -f 60 -cacheTimes 180"
```
hgy59: use port defined in Makefile
```suggestion
DDNS_GO="${SYNOPKG_PKGDEST}/bin/ddns-go -l :${SERIVCE_PORT} -f 60 -cacheTimes 180"
```

hgy59: And I have added the used port to the documentation of [SynoCommunity used Ports](https://github.com/SynoCommunity/spksrc/wiki/SynoCommunity-Used-Ports)

GrandDuke1106: Ah, it's ${SERVICE_PORT} instead of ${SERIVCE_PORT}.

hgy59: I assume you meant `${SERVICE_PORT}` (bash/sh script variable) vs. `$(SERVICE_PORT)` (Makefile variable).

GrandDuke1106: I initially made the modifications according to your suggestions, but found that the software had some issues running. I thought it was because I couldn't use this variable, but later realized it was just a spelling mistake.


image

hgy59: 🤣  didn't see the difference too...

mreid-tt:
@hgy59 anything further before publishing this? I've installed it, logged in and set an admin password, upgraded it and it seems okay. I don't have a DDNS service myself so I can't test the DNS update feature.

GrandDuke1106:
> @hgy59 anything further before publishing this? I've installed it, logged in and set an admin password, upgraded it and it seems okay. I don't have a DDNS service myself so I can't test the DNS update feature.

at least I have started using it normally

hgy59:
@GrandDuke1106 is it possible to make the -f (update frequency) and -changeTime parameters configurable?
why not using the defaults?

It seems that it is not possible to configure these values in the config.yml. So for customizing the values, we would need an installation wizard and write the values to a file like installer-variables and to load the values from this file in service-setup.sh.
If we support changing the values at package update, it needs some additional effort.
It is not mandatory to change the values at package update since the file with the variable definitions could be edited anytime.
We should not force the user to edit the service-setup file to change such values, because the changes would be lost at each update.
We should use the default values or make it configurable.

Packages using such an approach are minio, demoservice, kiwix,

GrandDuke1106:
> @GrandDuke1106 is it possible to make the -f (update frequency) and -changeTime parameters configurable? why not using the defaults?
> 
> It seems that it is not possible to configure these values in the config.yml. So for customizing the values, we would need an installation wizard and write the values to a file like installer-variables and to load the values from this file in service-setup.sh. If we support changing the values at package update, it needs some additional effort. It is not mandatory to change the values at package update since the file with the variable definitions could be edited anytime. We should not force the user to edit the service-setup file to change such values, because the changes would be lost at each update. We should use the default values or make it configurable.
> 
> Packages using such an approach are minio, demoservice, kiwix,

The default value for the -f parameter of this program is 300 (i.e., 5 minutes), and the default value for -cacheTimes is 5 times.

I just want to update the DNS as quickly as possible, even though 5 minutes is already quite short.

I will add an installation wizard so that users can customize -f and -cacheTimes.

GrandDuke1106:
Currently, with this installation method, upgrading or reinstalling with retaining data will generate an `environment-variables.new` file. Does this file need to be processed?

hgy59:
> Currently, with this installation method, upgrading or reinstalling with retaining data will generate an `environment-variables.new` file. Does this file need to be processed?

No, this is for reference of new entries. If someone had modified the environment-variables file, an update must not overwrite it.
But a user could compare it to see whether there are new entries after a package update.

spk/ddns-go/Makefile
```diff
@@ -0,0 +1,36 @@
...
+SPK_ICON = src/ddns-go.png
+
+DEPENDS = cross/ddns-go
+
+UNSUPPORTED_ARCHS = $(ARMv5_ARCHS)
+# not supported by go:
+UNSUPPORTED_ARCHS += $(PPC_ARCHS)
+
+MAINTAINER = GrandDuke1106
+DESCRIPTION = Automatically obtain your public IPv4 or IPv6 address and resolve it to the corresponding domain name service.
```
hgy59: I would expect to document the supported ddns providers like this:
```suggestion
DESCRIPTION = Automatically obtain your public IPv4 or IPv6 address and resolve it to the corresponding domain name service. Supported domain service providers are: Aliyun, Tencent, Dnspod, Cloudflare, Huawei, Callback, Baidu, Porkbun, GoDaddy, Namecheap, NameSilo, Dynadot, DNSLA, Nowcn, Eranet, Gcore.
```

PR #6678, opened 2025-08-11
syncthing: update to support x25 models

hgy59:
## Description

- update syncthing to v1.30.0
- create new package to support archs of x25 models
- add remarks in changelog to document the use of syncthing's autoupdate

Fixes # 

## Checklist

### Type of change

- [x] Package update


hgy59:
@acolomb it seems that the tech-ui does not work anymore (Error Status: 403 Message: Http failure response for http://diskstation:8384/rest/system/config: 403 Forbidden).

I am about to remove the tech-ui from this package. 
The original repo https://github.com/syncthing/tech-ui is archived long ago and the https://github.com/syncthing/next-gen-gui has gone.
In https://forum.syncthing.net/t/building-and-updating-next-gen-gui/23131 it seems that the nextgen ui is now part of the syncthing repo but I can't find it anywhere.
And your PR https://github.com/syncthing/syncthing/pull/9797 was not integrated.

Any objections?

acolomb:
@hgy59 Yes, go ahead and remove it.  It was an experiment but ultimately led nowhere.  We certainly don't need it in the SPK.  Not requiring all the Node-JS stuff for the build is a great simplification anyway.

acolomb (approved):

agottschling:
Question on this package. My synology is already running the v2 branch. Will upgrading to this new version break things? Will I need to hold off on upgrading in package center until one of these drops with v2?

acolomb:
Please be careful when "upgrading" this package with an already more recent Syncthing version installed through the internal updater.  For the v2.0 migration, the database is converted to a new format, which the older version will not see / understand.  Worst case, you could be pushing some files back to other connected Syncthing nodes which had been deleted there, or cause conflicts to be created.

I'm not quite sure what exactly will happen when doing the migration to v2 again after updating this package.  It should probably find the previously migrated v2 database and just continue from there.  Please ask on the Syncthing forum if someone has experience with this.

Either way, I'd recommend _pausing all folders and devices_ **before** each upgrade step.  Then see what happens and proceed to the next step once everything has settled.  If the global state of each folder looks alright, let them sync to the remotes again by unpausing.  This way you avoid pushing any intermediary state to the other nodes.

hgy59:
> Question on this package. My synology is already running the v2 branch. Will upgrading to this new version break things? Will I need to hold off on upgrading in package center until one of these drops with v2?

No, it will not break. The package installer ensures that no downgrade is made. 
You can install the latest package version without affecting the currently installed syncthing executable.

acolomb:
Oh, right. I forgot we put in that safety measure. :wink:

PR #6680, opened 2025-08-14
[Package Request] byobu

tofuHero:
## Description

[Byobu](https://www.byobu.org/) is a GPLv3 open source text-based window manager and terminal multiplexer. It was originally designed to provide elegant enhancements to the otherwise functional, plain, practical GNU Screen, for the Ubuntu server distribution. Byobu now includes an enhanced profiles, convenient keybindings, configuration utilities, and toggle-able system status notifications for both the GNU Screen window manager and the more modern Tmux terminal multiplexer, and works on most Linux, BSD, and Mac distributions.

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update

### Something to need help

I'm trying to compile and install this software package in a DSM virtual environment, but it doesn't work. When I run `byobu`, it exits with code 0. After running `byobu-config`, it shows an error message "ERROR: Could not import the python snack module". How can I compile `snack` into byobu?

图片


tofuHero:
I tried compiling version 5.133 from launchpad.net, and now the `byobu` command works fine, but I still get the error "ERROR: Could not import the python snack module" when running `byobu-config`.

```Makefile
PKG_VERS      = 5.133
PKG_EXT       = tar.gz
PKG_DIST_NAME = $(PKG_NAME)_$(PKG_VERS).orig.$(PKG_EXT)
PKG_DIST_SITE = https://launchpad.net/byobu/trunk/$(PKG_VERS)/+download
```

图片

hgy59:
@tofuHero Thanks for your contribution!

Some thoughts at first sight:
- please create a dedicated branch for your PR (don't use the master branch)
- if you include tmux and screen in the package, it must not depend on synocli-net
- if the package depends on python, it must have e.g. SPK_DEPENDS = python312
- for python dependent packages we normally create a virtual environment (venv) to install python modules on the target system at installation time (and need a prebuilt spk/python312 at build time)

At 2nd sight
- snack must be available as compiled python wheel and depends on python 2.7
- we can't support the python snack module as we don't support python 2 anymore
- if snack is a mandadory requirement, a byobu package will never run

Finally
- If we can get rid of the python dependency, and if the binaries of byobu aren't that big, we might include byobu in the synocli-net package.

tofuHero:
new pull request to #6681

PR #6696, opened 2025-08-27
znc: use alternate download

hgy59:
## Description

Use Synocommunity sources to download as the original site does not work

Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/cross/znc/Makefile
+++ b/cross/znc/Makefile
@@ -2,7 +2,9 @@ PKG_NAME = znc
 PKG_VERS = 1.10.1
 PKG_EXT = tar.gz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
-PKG_DIST_SITE = https://znc.in/releases/archive
+# use synocommunity "abandoned sources" 
+# original site https://znc.in/releases/archive does not work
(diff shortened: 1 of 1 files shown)


th0ma7:
@hgy59 sorry for the late replying but I just recalled that ZNC has a tagged github repository we can use:
https://github.com/znc/znc/tags

Also the link for the latest release probably changed where latest release sits in the top directory, until it get archived:
https://znc.in/releases/

I would suggest switching to github tagged version instead?

hgy59:
> I would suggest switching to github tagged version instead?

I already would have switched if those are compatible.

The github source archives are not suitable. Those are lacking the files of the (third_party) submodules and the modules folder seems incomplete too. Additionally it requires autoconf to create the configure file. The github archives are really just a snapshot of the repository as is.
It is really an issue of the znc.in server. Only the *.tar.gz files can't be downloaded (read error) but the *.sig and *.changelog.txt files can.
The https://znc.in/releases/ contains the latest release only, but https://znc.in/releases/archive/ has all releases. Using the archive folder does not break our builds, when znc release is updated.

hgy59:
@th0ma7 I have created an upstream issue https://github.com/znc/znc/issues/1973

DarthGandalf:
> Additionally it requires autoconf to create the configure file.

ZNC hasn't been using autoconf for several versions already, only cmake

PR #6704, opened 2025-09-07
library update: harfbuzz, libaqm

hgy59:
## Description

- update harfbuzz from v2.9.1 and v5.3.1 to v10.4.0
- drop cross/harfbuzz_2 since ARMv7L is supported with v10.4.0
- update libraqm from v0.8.0 to v0.10.2
- adjust harfbuzz dependent libraries (libass, libraqm, pango)
- pango: add harfbuzz for all supported archs and add fribidi to avoid building as subproject
- pango: for ARMv7L, harfbuzz 2.9.1 was not recognized and built as subproject (current head of harfbuzz was successfully built, but not added to related packages)

Libraries with common dependencies.
Separated from next update of imagemagick.

Fixes # 

## Checklist

### Type of change

- [x] Library update


hgy59:
@th0ma7 again, I need some information.

You kept harfbuzz 2.9.1 for ARMv7L archs as cross/harfbuzz_2.
This version had two patches
- 001-remove-duplicate-hb-gobject-enums-entries.patch
- 002-misc-armv7l-build-fixes.patch

The first patch is missing in cross/harfbuzz. As it is not related to a specific arch, and the patched file `fix_get_types.py` is equal in v2.9.1 and v5.3.1, we either need this patch in all versions or can omit it (since we only use shared libraries of harfbuzz and no python code, I think we don't need this patch).

The second patch (as from the file name) seems to fix the build for ARMv7L.
But the build does not fail on any version without this patch.

So I dropped the v2.9.1 and use updated harfbuzz for ARMv7L too.

The main reason for this is, that when building pango, it does not recognize v2.9.1 and builds a clone of harfbuzz in its subproject folder _(and pango successfully builds HEAD of harfbuzz for ARMv7L)_.
So pango would expect shared libraries of harfbuzz that we don't include in related packages and I am sure, this will break at runtime.


Pango has the same behaviour with fribidi dependency. So I added cross/fribidi, and pango does not build fribidi in its subproject folder anymore.

Do you have any concerns?

th0ma7:
> So I dropped the v2.9.1 and use updated harfbuzz for ARMv7L too.

Excellent!  Maybe the issue was super specific and caused by a number of factor such as the build ordering?  I also noticed this change part of hardfbuzz_v2 which may have solve the issue altogether if that was linked to glib badly detecting things:
```
.PHONY: harfbuzz-post-install-target
harfbuzz-post-install-target:
	$(RUN) sed -i 's|$(INSTALL_PREFIX)|$(STAGING_INSTALL_PREFIX)|g' $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/harfbuzz*.pc
``` 
> Pango has the same behaviour with fribidi dependency. So I added cross/fribidi, and pango does not build fribidi in its subproject folder anymore.

Good to know!

> Do you have any concerns?

None at all.  Further considering the update to a much more recent release.  I'm all in if we can simplify things up.

PR #6706, opened 2025-09-08
Update libraries: x265, x264

hgy59:
## Description

- update x265 from v3.6 to v4.1 and use assembler for aarch64 (DSM >= 7.1)
- x264: update to latest git revision

Fixes # 

## Checklist

### Type of change

- [x] Library update


hgy59:
@th0ma7 I updated the build of x265 for aarch64. To enable assembler (DSM >= 7.1 only works) I had to remove `CROSS_COMPILE_ARM` in the toolchain cmake file for aarch64. This was introduced in #5280.
We might use `CROSS_COMPILE_ARM64`  instead, but it seems uncommon and `CMAKE_SYSTEM_PROCESSOR = aarch64` is enough.

th0ma7:
I'm glad you found a way to make it work properly on aarch64.  There were back then major reworks with the toolchain file and then ensuring x264|265 do work with standard cross-cmake which needed a fair share of rewrite.  Hopefully maintenance won't be too bad now :)

PR #6707, opened 2025-09-08
Ente: Initial package add for the Photo Manager

### LINKED ISSUE #6305: [Package Request] Ente (End to End Encrypted Photos, and Auth)
### Software Name

Ente (Photos, and Auth)

### Brief Description

Fully open source, End to End Encrypted alternative to Google Photos and Apple Photos.

### Website

https://ente.io

### Documentation

_No response_

https://help.ente.io/self-hosting/

### Source code

https://github.com/ente-io/ente

### OpenSource License

Apache

### Other License

_No response_

solarkennedy:
Ente is an encrypted self-hosted photo management app.

This packages it natively.
It depends on Minio object storage and Postgres.

Minio is already a syno package, we can use that.
Synology NASs come with Postgres, so the readme has instructions on how to just use that, but you could connect it to any postgres you wanted to.

Fixes #6305.

Relates to https://github.com/ente-io/ente/discussions/4057.

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update

----

This is my first syno package ever, and this is a hard one.


mreid-tt:
hey @solarkennedy, thanks for the contribution, and for tackling such a complex, multi-component application. There’s clearly a lot of good effort here.

I was looking back at the original request and the earlier suggestion to run this as Docker given the overall scope. Before going further, could you share a bit more about the motivation for packaging this natively rather than via Docker?

From a quick review, the implementation works, but there are a few areas where I’m wondering whether this approach ends up being simpler or more complex than a containerized deployment; especially given some of the manual setup involved (MinIO, database, etc.). A few specific observations and questions:

1. The package appears to rely on Synology’s built-in PostgreSQL, which we generally discourage for third-party apps. Would using MariaDB (or another bundled dependency) be an option?
2. A separate nginx instance is being set up, which adds some overhead. Would it be possible to leverage Synology’s WebStation instead?
3. The build uses `native/nodejs` rather than a versioned dependency like `native/nodejs_22`, which could introduce breakage with future Node.js updates.
4. There’s a hard-coded path in the nginx config: `root /volume1/@appstore/ente/web;`
   This should use `${SYNOPKG_PKGDEST}/web`, since the install volume can vary.
5. A few files are missing a final newline at EOF, which is part of our standard practice.
6. `SPK_DEPENDS = "minio"` is missing a version constraint. If there’s an API dependency, we should specify a minimum version. Also, there’s an upcoming Garage package that may be preferred over MinIO.
7. For Go cross-compilation, the Makefile uses `env $(ENV)`, but `ENV` is already set up by `spksrc.cross-go.mk`, so this could likely be simplified to `$(RUN) go build …`.
8. We don’t usually ship a dedicated README inside packages; content like this is typically added to the wiki and referenced during the install wizard.

These are just initial thoughts from a cursory review. Were you able to build and test the package locally on a DSM system?

Looking forward to your thoughts.

solarkennedy:
My motivation here was just fun.

I don't think Docker is a bad suggestion, I just don't like it, personally.

1. I don't think so. It requires postgres.
2. Hmm. Yea that could be cool.
3. Good point
...

The thing is, while I did build this, and I tried it out, connected my phone, and it _did_ work, in the end I decided not to host this.

I can't in good faith recommend this PR, since I won't be running it myself.
Looking back, I made it in September. I don't remember all the things I did.

At this point I would say we should close it as wont-fix unless someone else is keen to support it.

mreid-tt:
> I can't in good faith recommend this PR, since I won't be running it myself. Looking back, I made it in September. I don't remember all the things I did.
> 
> At this point I would say we should close it as wont-fix unless someone else is keen to support it.

Thanks for the feedback. I think the main sticking point here is the PostgreSQL requirement. It’s a similar reason why we haven’t been able to update Tiny Tiny RSS (`tt-rss`), which introduced the same dependency. Unless we provide our own PostgreSQL server package, it’s unlikely we’d publish something that relies on or modifies Synology’s built-in PostgreSQL.

PR #6646, opened 2025-07-13
Require kernel fixes

th0ma7:
## Description

Require kernel fixes:
1. Solve kernel 5.x builds `epyc7002` and `rtd1619b`
2. Update kernel-module code to simplify log redirection and properly exit on errors
3. Solve issue where arch would exist in toolchain but no kernel sources available (now filtered out)

TODO:
4. fix `make download`
5. Update kernels for DSM-7.2 using latest sources 72806

Relates to https://github.com/SynoCommunity/spksrc/issues/6632

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6193, opened 2024-08-13
Create proper debug configuration

icegood:
## Description (original from @icegood)

With this PR it is easier to create more rigorous debug builds for further debugging in gdb.  (for at least spk/ffmpeg)
Not applicable out of the box since other dependant repos need to be fit as well. 

To check against: https://github.com/icegood/spksrc/tree/fix/libpng_debug_builds at it fixes png bug

Applicable for issue #6176

## Additions (from @th0ma7)
1. Dissociate CMake toolchain file generation from default `spksrc.cross-cmake.mk` (similar to meson)
2. Enforce using an "empty" environment for both `CMake` and `meson` (fix issue against 'build' vs 'host' builds)
3. Fix meson crossfile generation - needs `[built-in options]` and `fortran_*` instead of `fc_*`
4. Fix CMake toolchain file generation with regards to ldshared definition
5. Add `CMAKE_*_FOR_BUILD` to CMake toolchain definition
6. Enable enforcing passing proper debug flags throughout CMake and Meson builds
7. Add passing debug flags to `ENV` with autoconf/autotools
8. Fixed a few cross/* Debug compile errors
9. Fixed situation where `spksrc.copy.mk` would not exit on error using tar from pipe
10. Updated cross/* when extra code was needed to allow debug build to work

Relates to https://github.com/SynoCommunity/spksrc/pull/6719

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes <s>small</s> framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

## Tests
Tested in debug mode against:
- [x] `synocli-videodriver`
- [x] `ffmpeg`
- [x] `python312`
- [x] `tvheadend`

PR #6717, opened 2025-09-17
SRM enhancements

hgy59:
## Description

This is mainly a framework update to support  SRM 1.3

1. Fix evaluation of REQUIRED_MIN_SRM 
2. Add Kernels for SRM 1.3
3. Adjust packages that have REQUIRED_MIN_SRM = 1.3

## Checklist

- [ ] Build rule `all-supported` completed successfully (this PR triggers the build of a lot of packages, but only builds for SRM 1.2 and SRM 1.3 are affected)

### Type of change

- [x] Includes framework changes


hgy59:
@th0ma7 can you have a look at this PR please?

The spksrc.tc.mk changes will conflict with your changes in #6193.
Can we merge this PR first and you will resolve the conflicts in #6193 then?

spk/synokernel-usbserial/Makefile
```diff
@@ -18,7 +18,15 @@ MAINTAINER = th0ma7
 DESCRIPTION = "Provides usbserial.ko ch341.ko cp210x.ko pl2303.ko ti_usb3410_5052.ko and ftdi_sio.ko"
 CHANGELOG = "1. Support DSM-7.1 and DSM-7.2"
 
-UNSUPPORTED_ARCHS = $(PPC_ARCHS) epyc7002 rtd1619b
+UNSUPPORTED_ARCHS = $(PPC_ARCHS)
+# archs with kernel 5.x fail to build
+# error: Cannot generate ORC metadata for CONFIG_UNWINDER_ORC=y, please install libelf-dev, libelf-devel or elfutils-libelf-devel
+# with installation of elfutils and libelf-dev to the development environment,
+# it further fails with error: pointer may be used after ‘realloc’ [-Werror=use-after-free]
+UNSUPPORTED_ARCHS += epyc7002 rtd1619b
```
th0ma7, line 26: Yeah I have an on-going PR to solve that https://github.com/SynoCommunity/spksrc/pull/6646

spk/demowebservice/Makefile
```diff
@@ -43,7 +46,7 @@ CONF_DIR = src/conf_php$(PHP_VERSION)/
 # Disable the creation of app/config file:
 NO_SERVICE_SHORTCUT = y
 # create firewall rules, since we add a port based "Web Portal"
-SERVICE_PORT = 8889
+SERVICE_PORT = 18889
 endif
```
th0ma7, line 50: Is there a need to update the wiki?  Friendly reminder just in case.

spk/demoservice/Makefile
```diff
@@ -27,7 +27,7 @@ STARTABLE = yes
 SERVICE_WIZARD_SHARENAME = wizard_shared_folder_name
 
 # Service configuration
-SERVICE_PORT = 8888
+SERVICE_PORT = 18888
```
th0ma7, line 30: similar to other comment, is the wiki updated?

mk/spksrc.tc-flags.mk
```diff
@@ -1,7 +1,17 @@
+# we can't check whether gfortran exists, because toolchain is not yet extracted
 ifeq ($(strip $(firstword $(subst ., ,$(TC_VERS)))),7)
-TOOLS = ld ldshared:"gcc -shared" cpp nm cc:gcc as ranlib cxx:g++ fc:gfortran ar strip objdump objcopy readelf
-else
+TC_HAS_FORTRAN = 1
+else ifeq ($(strip $(TC_VERS)),1.3)
+TC_HAS_FORTRAN = 1
+else ifeq ($(strip $(TC_VERS)),6.2.4)
+ifeq ($(findstring $(ARCH),$(x64_ARCHS)),$(ARCH))
+TC_HAS_FORTRAN = 1
```
th0ma7, line 8: Good catch!

mk/spksrc.pre-check.mk
```diff
@@ -94,7 +94,7 @@ endif
 
 # Check minimum SRM requirements of package
 ifneq ($(REQUIRED_MIN_SRM),)
-  ifeq ($(ARCH),$(findstring $(ARCH),$(SRM_ARCHS)))
+  ifeq ($(call version_lt, ${TCVERSION}, 3.0),1)
```
th0ma7, line 97: Your overall changes in pre-check.mk makes me reconsider why was the code like this in the first place?  Maybe I'm just tired tonight but unsure that this is the best method to confirm DSM requirements based on SRM min level and so on.  I do not have a better idea and mabye best to leave that for another time to revisit but something is odd from my perspective.

mk/spksrc.kernel-flags.mk
```diff
@@ -124,22 +124,23 @@ endif
 else ifeq ($(strip $(TC_TYPE)),SRM)
 
 ifeq ($(strip $(KERNEL_DIST_SITE)),)
-KERNEL_DIST_SITE = https://sourceforge.net/projects/dsgpl/files/Synology%20Router%20GPL%20Source/$(KERNEL_BUILD)branch/$(KERNEL_URL_DIR)
+KERNEL_DIST_SITE = https://github.com/SynoCommunity/spksrc/releases/download/kernels/srm$(KERNEL_VERS)
 endif
```
th0ma7, line 128: If I get this right, that mean that all SRM (e.g. now only SRM-1.3) are now hosted on our github repo?  Not now but food for thought may be to retire entirely older inexistent SRM+DSM references...

hgy59: > If I get this right, that mean that all SRM (e.g. now only SRM-1.3) are now hosted on our github repo? Not now but food for thought may be to retire entirely older inexistent SRM+DSM references...

That is, because Synology does not provide a separate linux-* archive for SRM kernels but only the full gplsources that are 1.7..1.8 GB.
I extracted the linux-* folders from the gpl sources to get archives of ~ 80 MB.
How to get those archives is documented in the related release comments.

We already did this for the SRM 1.3 toolchains to extract those from the dev-env archives.

spk/demowebservice/Makefile
hgy59: > Is there a need to update the wiki? Friendly reminder just in case.

I added a comment that it conflicts with a port used by Synology SRM and will update it when this is merged.
I updated the wiki with ports additionally used by Synology SRM
https://github.com/SynoCommunity/spksrc/wiki/Synology-Used-Ports

hgy59:
> Secondly, somewhere was mentioned that piece of code is missing the kernel sources? Should we instead consider adding a patch directory in the appropriate kernel directory tree instead? Or maybe I misread.

There are not the kernel sources but the toolchains that are missing for nortstarplus-1.2 and dakota-1.2.
We have the armv7-1.2 toolchain to create packages for all SRM 1.2 models, but we can't build kernel moduls for those, since we missed to download the toolchains before Synology's removal in 2023.

May be we can build kernel modules using the armv7-1.2 toolchain, but this is out of scope of this PR...

hgy59:
@th0ma7 thanks for taking a look (or two)...

I have added a list with all packages that can be built for SRM.
There are some compile errors for SRM 1.3, that have to be analyzed later (lacking time ATM).
And I didn't want to define REQUIRED_MAX_SRM since SRM packages are manually built only (not with regular gibhub build action).

The tcp_wrapper error affects DSM 7.2 too so it will be worth to dive deeper...

hgy59:
> First thing that crossed my mind is, could we use `REQUIRED_MIN_SRM = nop` instead of `REQUIRED_MIN_SRM = 3.0` to invalidate support for SRM, similar to other `nop` in used elsewhere?

 _other `nop` elsewhere_ is used for targets only, REQUIRED_MIN_* and REQUIRED_MAX_* sould only be defined with version numbers.

An other solution would be to define `UNSUPPORTED_ARCHS = SRM_ARCHS`.
But this would be ugly for noarch packages (and most of the packages not supported for SRM are noarch packages).
This could be solved with  `UNSUPPORTED_ARCHS_TCVERSION = noarch-1.1 noarch-1.2 noarch-1.3` (but this would need and update when SRM 1.4+ are released).

Or we introduce a Makefile variable like `UNSUPPORTED_PRODUCT_TYPES` that could be used as `UNSUPPORTED_PRODUCT_TYPES = SRM` (or, if we ever have packages supported on SRM only, with `UNSUPPORTED_PRODUCT_TYPES = DSM`).

@th0ma7 what's your preferred solution?

BTW we still use the TCVERSION to distinct SRM and different geneartions of DSM like
- \>= 1.0, < 3.0 for SRM
- \>= 3.0, < 6.0 for DSM 3 to 5 compatibility
- \>= 6.0, < 7.0 for DSM 6
- \>= 7.0 for DSM 7

th0ma7:
I'm wondering how hard would it be to have a generic UNSUPPORTED variable with catchall rules for srm|dsm|><=|x.y .... Avoiding all together the burden of multiple variable names for basically the same thing.  Something somehow like the python requirement ruleset that uses something like this.  Question is, how hard would it be to implement it?  What do you think?

But i may just be shoveling clouds... I'll think further about this one...

hgy59:
> I'm wondering how hard would it be to have a generic UNSUPPORTED variable with catchall rules for srm|dsm|><=|x.y .... Avoiding all together the burden of multiple variable names for basically the same thing. Something somehow like the python requirement ruleset that uses something like this. Question is, how hard would it be to implement it? What do you think?
> 
> But i may just be shoveling clouds... I'll think further about this one...

It's worth considering, but too much for this PR. 

If we do a redesign, I propose to include a solution for gcc version dependency et.al. (i.e. we need to have gcc and glibc version available for checks prior to build any dependency like the toolchain).

PR #6718, opened 2025-09-20
bazarr: bump bazarr 1.5.2 to 1.5.3

smaarn:
## Description

This is a mere bump of bazarr from 1.5.2 to 1.5.3

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


mreid-tt:
@smaarn, may I suggest moving to `lxml==6.0.2` in the `requirements-crossenv.txt`.

smaarn:
> @smaarn, may I suggest moving to `lxml==6.0.2` in the `requirements-crossenv.txt`.

@mreid-tt will be checking the impact of doing so. Any particular reason for the move ?

mreid-tt:
> @mreid-tt will be checking the impact of doing so. Any particular reason for the move ?

I was just checking this against the standard requirements generator, and that was the only component available for update. I also noticed that this version has already been tested with the wheels package for Python 3.12, so it appears to be build-tested already.

smaarn:
> > @mreid-tt will be checking the impact of doing so. Any particular reason for the move ?
> 
> I was just checking this against the standard requirements generator, and that was the only component available for update. I also noticed that this version has already been tested with the wheels package for Python 3.12, so it appears to be build-tested already.

Fair enough. My only worry here is the functional impacts which could need being investigated. Building locally to see how it goes.

smaarn:
Did a little bit of digging and the features which were actually dropped (according to the changelog) are not used in bazarr (direct compressed access or direct access through URLs) so we should be good I guess (ran a few tests and all signals are green...).

mreid-tt:
Looks good. Don't think it's necessary to include individual package updates in the changelog as this was not the case in previous updates. Otherwise, seems good to go.

smaarn:
> Don't think it's necessary to include individual package updates in the changelog as this was not the case in previous updates.

Too bad we don't have a more advanced "changelog" section allowing to include such details 😅 

My main purpose here was to ensure that if some issues are found by users "in the wild" and they check the logs they would be aware of the technical changes which were done and thus may be able to refine their search results and help out potentially... But that's a lot of ifs, I give you that. Will remove.

PR #6344, opened 2024-12-06
[New package] .NET 9

brice-fromentin:
## Description

Initial release of .NET 9 package for DSM 7.0+ (and DSM 6).

Closes #6343

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6741, opened 2025-10-16
October 2025 dotnet updates

brice-fromentin:
## Description

Update dotnet8-runtime -> 8.0.21
Update dotnet9-runtime -> 9.0.10
Update dotnet-sdk-8 -> 8.0.415

Tested on 6.2.4 and 7.1 (x64)
Tested with https://github.com/SynoCommunity/spksrc/pull/6344#issuecomment-2525101266

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update


spk/dotnet10-runtime/Makefile
```diff
@@ -0,0 +1,40 @@
+SPK_NAME = dotnet10-runtime
+SPK_VERS = 10.0.0-rc.2
```
hgy59: version number must contain only dots and digits, dash and alpha are not supported.

but we should not deploy release candiate versions anyway.

My intention is to wait for the official dotnet 10 release and drop dotnet 9 when dotnet 10 is released.

brice-fromentin: OK, I removed dotnet 10 rc. However, I think we will need to keep and maintain .NET 9 until it reaches the end of its support.

Issue #6757, opened 2025-10-23
[Prowlarr v2.0.5.5160] Unable to connect on DSM 6.2 and DSM 7.1


mreid-tt:
### Is this a new Bug?

- [x] I checked that the bug hasn't been reported before

### Package Name

Prowlarr

### Package Version

20251012-9

### Device Model

VirtualDSM

### Device Architecture

x86_64

### Firmware Version

DSM 6.2.4-25556 Update 8

### What happened?

Performing a clean install of the latest version of Prowlar completes however the app does not show anything when opened. The message "Unable to connect" is shown in the web browser.

### Reproduction steps

1. Install Prowlarr v20251012-9 from the Package Center
2. Open application webpage
3. No webpage appears

### Install Log

```log
2025/10/23 08:47:02	install prowlarr 20251012-9 Begin preinst
2025/10/23 08:47:02	Begin initialize_variables
2025/10/23 08:47:02	End initialize_variables
2025/10/23 08:47:02	===> Step preinst. STATUS=INSTALL USER=prowlarr GROUP=sc-download SHARE_PATH=
2025/10/23 08:47:02	install prowlarr 20251012-9 End preinst ret=[0]
2025/10/23 08:47:02	install prowlarr 20251012-9 Begin /bin/rm -rf /volume1/@appstore/prowlarr
2025/10/23 08:47:02	install prowlarr 20251012-9 End /bin/rm -rf /volume1/@appstore/prowlarr ret=[0]
2025/10/23 08:47:02	install prowlarr 20251012-9 Begin /bin/mv -f /volume1/@tmp/pkginstall/package /volume1/@appstore/prowlarr
2025/10/23 08:47:02	install prowlarr 20251012-9 End /bin/mv -f /volume1/@tmp/pkginstall/package /volume1/@appstore/prowlarr ret=[0]
2025/10/23 08:47:02	install prowlarr 20251012-9 Begin /bin/rm -rf /var/packages/prowlarr
2025/10/23 08:47:02	install prowlarr 20251012-9 End /bin/rm -rf /var/packages/prowlarr ret=[0]
2025/10/23 08:47:02	install prowlarr 20251012-9 Begin /bin/mkdir -p /var/packages/prowlarr
2025/10/23 08:47:02	install prowlarr 20251012-9 End /bin/mkdir -p /var/packages/prowlarr ret=[0]
[...]

PR #6755, opened 2025-10-22
Kavita: Update to v0.8.8

mreid-tt:
## Description

This PR includes the following changes:

1. Update Kavita to v0.8.8

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #6613, opened 2025-06-14
Kavita: Add fix for SQLite

mreid-tt:
## Description

This PR includes the following changes:

1. Patch Kavita v0.8.6.2 with fix for SQLite

Fixes #6612

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix

PR #6758, opened 2025-10-25
Kavita: Update to v0.8.8.3

mreid-tt:
## Description

1. Update Kavita to v0.8.8.3

Fixes # 

## Checklist

### Type of change

- [x] Package update


mreid-tt:
Hey @hgy59, I was looking at this most recent update and reflecting on the error you noted in the last one in https://github.com/SynoCommunity/spksrc/pull/6755#issuecomment-3435119389. Given what was discovered in the latest version of Prowlarr where the database migration seems to fail with an older version of SQLite (https://github.com/SynoCommunity/spksrc/issues/6757#issuecomment-3436948467), it may be that the same thing may have happened to your migration.

Looking back at when this package was first published we had this same `GLIBC` dependency issue with SQLite (https://github.com/SynoCommunity/spksrc/pull/5978#issuecomment-1886887208) but at the time we didn't realise that the issue also affected DSM 7.1. We eventually worked around the issue in https://github.com/SynoCommunity/spksrc/pull/6613 with the patch developed in https://github.com/SynoCommunity/spksrc/pull/6348. It was initially shipped with SQLite 3.46.1 and while this seems to work for most scenarios there have been more recent updates upstream (https://github.com/ericsink/cb/commit/b0f90da0f67a530dd89f9df90b4034b1693acdfb) to sqlite3 and sqlite3mc v3.49.1.

Now the challenge I am having is trying to determine the version of `libe_sqlite3.so` that ships with Kavita but given what I saw in Prowlarr, the version of SQLite we replace it with may have been the cause of the database migration you experienced but I was unable to reproduce. If so, we may have to include the upstream version and restrict Kavita to DSM 7.2 and up. If your Kavita library is still in the state where you can reproduce the issue it would be useful to know if you included the upstream `libe_sqlite3.so` would your update succeed?

ZyorTaelon:
> Now the challenge I am having is trying to determine the version of `libe_sqlite3.so` that ships with Kavita but given what I saw in Prowlarr, the version of SQLite we replace it with may have been the cause of the database migration you experienced but I was unable to reproduce. If so, we may have to include the upstream version and restrict Kavita to DSM 7.2 and up. If your Kavita library is still in the state where you can reproduce the issue it would be useful to know if you included the upstream `libe_sqlite3.so` would your update succeed?

In Kavita/API/API.csproj the following mentions of SQLite are made:
```
<PackageReference Include="Hangfire.Storage.SQLite" Version="0.4.2" />
<PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="9.0.10" />
```
Does that answer your version question?

mreid-tt:
> Does that answer your version question?

Not exactly but it did give me an idea to take the version of `libe_sqlite3.so` that comes with Kavita and manually replace the file into the Prowlarr I was working on. It reported back the same version as the one I inserted:

```
[Info] DatabaseEngineVersionCheck: SQLite 3.46.1 
```

So in that regard we should be in line with what is expected by the upstream provider.

PR #6759, opened 2025-10-26
Prowlarr: Update to v2.1.5.5216

### LINKED ISSUE #6757: [Prowlarr v2.0.5.5160] Unable to connect on DSM 6.2 and DSM 7.1

mreid-tt:
## Description

1. Update Prowlarr to v2.1.5.5216
2. Add custom `libe_sqlite3` for DSM < 7.2
3. Disable auto-update for DSM < 7.2

Fixes #6757

## Checklist

### Type of change

- [x] Bug fix
- [x] Package update

### DIFF
--- a/spk/prowlarr/src/service-setup.sh
+++ b/spk/prowlarr/src/service-setup.sh
@@ -27,9 +27,9 @@ internal_update_supported() {
 
     # If the file contains "UpdateMethod=External", updates are NOT supported
     if grep -q '^UpdateMethod=External' "$info_file" 2>/dev/null; then
-        return 1    # DSM < 7.2; Not supported
+        return 1    # Not supported
     else
-        return 0    # DSM ≥ 7.2; Supported
+        return 0    # Supported
     fi
 }
 
@@ -37,7 +37,7 @@ service_postinst ()
 {
     if [ "${SYNOPKG_PKG_STATUS}" = "INSTALL" ]; then
         if internal_update_supported; then
-            # DSM ≥ 7.2; make Prowlarr do an update check on start
+            # make Prowlarr do an update check on start
             echo "Set update required"
             touch "${PROWLARR_CONFIG_DIR}/update_required" 2>&1
         fi
@@ -51,7 +51,7 @@ service_postinst ()
 service_preupgrade ()
 {
     if internal_update_supported; then
-        # DSM ≥ 7.2; don't update Prowlarr distribution, use internal updater only
+        # don't update Prowlarr distribution, use internal updater
         [ -d "${SYNOPKG_TEMP_UPGRADE_FOLDER}/backup" ] && rm -rf "${SYNOPKG_TEMP_UPGRADE_FOLDER}/backup"
         echo "Backup existing distribution to ${SYNOPKG_TEMP_UPGRADE_FOLDER}/backup"
         mkdir -p "${SYNOPKG_TEMP_UPGRADE_FOLDER}/backup" 2>&1


mreid-tt:
Hey @hgy59, following up on my note in https://github.com/SynoCommunity/spksrc/issues/6757#issuecomment-3447850092, I was able to generate both DSM 7.1 and DSM 7.2 builds (https://github.com/mreid-tt/spksrc/actions/runs/18809872858).

When installed, each package correctly reflected the expected SQLite version in the logs:

For `prowlarr_x64-7.1_20251025-10.spk`:
```
[Info] DatabaseEngineVersionCheck: SQLite 3.46.1 
```

For `prowlarr_x64-7.2_20251025-10.spk`:
```
[Info] DatabaseEngineVersionCheck: SQLite 3.50.4 
```

Both packages appear to be working well so far. However, I’m still a bit concerned that the built-in updater might overwrite the custom `libe_sqlite3` in future updates. I’m running a few more tests on other Prowlarr versions and will share my findings once complete.

mreid-tt:
Hey @hgy59, as I suspected, once the v2.1.5.5216 of Prowlarr showed up as an available update, the automatic updater did its job and replaced all the binary files. For DSM 7.2 and up this was not an issue but for older versions like DSM 6, we see the following:

```
[Info] Microsoft.Hosting.Lifetime: Content root path: / 
[Info] InstallUpdateService: Downloading update 2.1.5.5216 
[Info] InstallUpdateService: Verifying update package 
[Info] InstallUpdateService: Update package verified successfully 
[Info] InstallUpdateService: Extracting Update package 
[Info] InstallUpdateService: Update package extracted successfully 
[Info] BackupService: Starting Backup 
[Info] InstallUpdateService: Preparing client 
[Info] InstallUpdateService: Starting update client /volume1/@appstore/prowlarr/tmp/prowlarr_update/Prowlarr.Update 
[Info] InstallUpdateService: Prowlarr will restart shortly. 
prowlarr is not running
Starting prowlarr ...
tail: /var/packages/prowlarr/var/prowlarr.log: file truncated
Sun Oct 26 05:04:48 -04 2025
Starting prowlarr command env HOME=/volume1/@appstore/prowlarr/var LD_LIBRARY_PATH=/volume1/@appstore/prowlarr/lib /volume1/@appstore/prowlarr/share/Prowlarr/bin/Prowlarr -nobrowser -data=/volume1/@appstore/prowlarr/var/.config/Prowlarr
[Info] Bootstrap: Starting Prowlarr - /volume1/@appstore/prowlarr/share/Prowlarr/bin/Prowlarr - Version 2.1.5.5216 
[Info] AppFolderInfo: Data directory is being overridden to [/volume1/@appstore/prowlarr/var/.config/Prowlarr] 
[Debug] Bootstrap: Console selected 
[Info] AppFolderInfo: Data directory is being overridden to [/volume1/@appstore/prowlarr/var/.config/Prowlarr] 
[Info] AppFolderInfo: Data directory is being overridden to [/volume1/@appstore/prowlarr/var/.config/Prowlarr] 
[Info] MigrationController: *** Migrating data source=/volume1/@appstore/prowlarr/var/.config/Prowlarr/prowlarr.db;cache size=-20000;datetimekind=Utc;journal mode=Wal;pooling=True;version=3;busytimeout=100 *** 
[Info] FluentMigrator.Runner.MigrationRunner: DatabaseEngineVersionCheck migrating 
[Info] FluentMigrator.Runner.MigrationRunner: PerformDBOperation  
[Info] NzbDrone.Core.Datastore.Migration.Framework.NzbDroneSQLiteProcessor: Performing DB Operation 
[Error] FluentMigrator.Runner.MigrationRunner: Unable to load the driver. Attempted to load: System.Data.SQLite, Mono.Data.Sqlite, Version=4.0.0.0, Culture=neutral, PublicKeyToken=(removed) Microsoft.Data.Sqlite, with System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.TypeInitializationException: The type initializer for 'System.Data.SQLite.SQLiteFactory' threw an exception.
 ---> System.DllNotFoundException: Unable to load shared library 'e_sqlite3' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable: 
/volume1/@appstore/prowlarr/share/Prowlarr/bin/e_sqlite3.so: cannot open shared object file: No such file or directory
/lib/libc.so.6: version `GLIBC_2.28' not found (required by /volume1/@appstore/prowlarr/share/Prowlarr/bin/libe_sqlite3.so)
libe_sqlite3.so: cannot open shared object file: No such file or directory
/volume1/@appstore/prowlarr/share/Prowlarr/bin/e_sqlite3: cannot open shared object file: No such file or directory
/volume1/@appstore/prowlarr/share/Prowlarr/bin/libe_sqlite3: cannot open shared object file: No such file or directory
```

This  breaks the instance of Prowlarr and essentially re-introduces the problem in #6757. Now I can perhaps disable the auto-update for versions on DSM < 7.2 but this will be a fundamental change for our ***arr** apps and I'm not sure I'll be able to prevent users from manually executing the update function. Let me know your thoughts on the sustainability of this approach.

hgy59:
@mreid-tt I have built the libe_sqlite3.so as included in the current Prowlarr builds (SQLite 3.50.4)

But it depends on glibc 2.33 and 2.34 even when built with DSM 6 toolchain.
As long as it works with the current libe_sqlite3.so (cross/libe_sqlite) we can deliver DSM 6 packages, but we must disable the automatic update to ensure to use the custom built library.

Starting with SQLitePCLRaw 3.x or System.Data.SQLite 2.x it depends on libe_sqlite3.so that depends on glibc >= 2.34.

```
$ readelf -sW work-aarch64-6.2.4/install/usr/local/sqlite-amalgamation/lib/libe_sqlite3.so  | grep GLIBC_2.[23][0-9]               2: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND dlerror@GLIBC_2.34 (3)
    26: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND lstat64@GLIBC_2.33 (4)
    39: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND stat64@GLIBC_2.33 (4)
    47: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_mutexattr_settype@GLIBC_2.34 (3)
    62: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_mutex_trylock@GLIBC_2.34 (3)
    72: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND dlopen@GLIBC_2.34 (3)
   101: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_create@GLIBC_2.34 (3)
   115: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND dlsym@GLIBC_2.34 (3)
   123: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND fstat64@GLIBC_2.33 (4)
   125: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_mutexattr_init@GLIBC_2.34 (3)
   126: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_join@GLIBC_2.34 (3)
   127: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND dlclose@GLIBC_2.34 (3)
   130: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_mutexattr_destroy@GLIBC_2.34 (3)
  2993: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND dlerror@GLIBC_2.34
  3061: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND lstat64@GLIBC_2.33
  3101: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND stat64@GLIBC_2.33
  3114: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_mutexattr_settype@GLIBC_2.34
  3168: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_mutex_trylock@GLIBC_2.34
  3204: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND dlopen@GLIBC_2.34
  3326: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_create@GLIBC_2.34
  3371: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND dlsym@GLIBC_2.34
  3414: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND fstat64@GLIBC_2.33
  3424: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_mutexattr_init@GLIBC_2.34
  3425: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_join@GLIBC_2.34
  3432: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND dlclose@GLIBC_2.34
  3439: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND pthread_mutexattr_destroy@GLIBC_2.34
```

mreid-tt:
@hgy59, thanks for looking into this. I actually just completed another approach. Instead of locking down to a specific version, we can continue using the in-app updates but using a custom script. This script will detect the DSM version and preserve the current `libe_sqlite3.so` we ship with it if running on DSM < 7.2. Early testing looks promising so far on DSM 6.

The generated update log looks like this:
```
2025-10-26 12:42:33 === Prowlarr upgrade | DSM 6.2 ===
2025-10-26 12:42:33 SRC=/volume1/@appstore/prowlarr/tmp/prowlarr_update/Prowlarr | BIN=/volume1/@appstore/prowlarr/share/Prowlarr/bin | DATA=/volume1/@appstore/prowlarr/var/.config/Prowlarr | RUN_LOG=/volume1/@appstore/prowlarr/var/prowlarr.log
2025-10-26 12:42:33 Stopping PID 30038 (TERM → up to 30s → KILL)
2025-10-26 12:42:34 Cleared stale PID: /volume1/@appstore/prowlarr/var/.config/Prowlarr/prowlarr.pid
2025-10-26 12:42:34 Preserved libe_sqlite3.so
2025-10-26 12:42:34 Wipe /volume1/@appstore/prowlarr/share/Prowlarr/bin
2025-10-26 12:42:34 Copy /volume1/@appstore/prowlarr/tmp/prowlarr_update/Prowlarr -> /volume1/@appstore/prowlarr/share/Prowlarr/bin
2025-10-26 12:42:34 Restored libe_sqlite3.so
2025-10-26 12:42:34 Start prowlarr cmd: env HOME=/volume1/@appstore/prowlarr/var LD_LIBRARY_PATH=/volume1/@appstore/prowlarr/lib /volume1/@appstore/prowlarr/share/Prowlarr/bin/Prowlarr -nobrowser -data=/volume1/@appstore/prowlarr/var/.config/Prowlarr
2025-10-26 12:42:35 Started via PID file: 30448
2025-10-26 12:42:35 === done ===
```

The service log looks like this:
```
2025-10-26 12:41:52.5|Info|Microsoft.Hosting.Lifetime|Now listening on: http://[::]:9696
2025-10-26 12:41:52.6|Info|AppSyncProfileService|Setting up default app profile
2025-10-26 12:41:52.6|Info|CommandExecutor|Starting 3 threads for tasks.
2025-10-26 12:41:52.8|Info|ManagedHttpDispatcher|IPv4 is available: True, IPv6 will be disabled
2025-10-26 12:41:53.0|Info|Microsoft.Hosting.Lifetime|Application started. Press Ctrl+C to shut down.
2025-10-26 12:41:53.0|Info|Microsoft.Hosting.Lifetime|Hosting environment: Production
2025-10-26 12:41:53.0|Info|Microsoft.Hosting.Lifetime|Content root path: /
2025-10-26 12:42:00.2|Info|Prowlarr.Http.Authentication.ApiKeyAuthenticationHandler|AuthenticationScheme: SignalR was challenged.
2025-10-26 12:42:10.3|Info|Prowlarr.Http.Authentication.ApiKeyAuthenticationHandler|AuthenticationScheme: SignalR was challenged.
2025-10-26 12:42:20.4|Info|Prowlarr.Http.Authentication.ApiKeyAuthenticationHandler|AuthenticationScheme: SignalR was challenged.
2025-10-26 12:42:26.0|Info|InstallUpdateService|Downloading update 2.1.5.5216
2025-10-26 12:42:30.9|Info|Prowlarr.Http.Authentication.ApiKeyAuthenticationHandler|AuthenticationScheme: SignalR was challenged.
2025-10-26 12:42:32.2|Info|InstallUpdateService|Verifying update package
2025-10-26 12:42:32.3|Info|InstallUpdateService|Update package verified successfully
2025-10-26 12:42:32.3|Info|InstallUpdateService|Extracting Update package
2025-10-26 12:42:33.7|Info|InstallUpdateService|Update package extracted successfully
2025-10-26 12:42:33.7|Info|BackupService|Starting Backup
2025-10-26 12:42:33.7|Info|InstallUpdateService|Removing Prowlarr.Update
2025-10-26 12:42:33.7|Info|InstallUpdateService|Starting update script: /var/packages/prowlarr/target/share/Prowlarr/servarr_spk_upgrade.sh
2025-10-26 12:42:33.8|Info|Microsoft.Hosting.Lifetime|Application is shutting down...
2025-10-26 12:42:33.8|Info|ConsoleApp|Exiting main.
2025-10-26 12:42:35.1|Info|Bootstrap|Starting Prowlarr - /volume1/@appstore/prowlarr/share/Prowlarr/bin/Prowlarr - Version 2.1.5.5216
2025-10-26 12:42:35.2|Info|AppFolderInfo|Data directory is being overridden to [/volume1/@appstore/prowlarr/var/.config/Prowlarr]
2025-10-26 12:42:35.2|Info|AppFolderInfo|Data directory is being overridden to [/volume1/@appstore/prowlarr/var/.config/Prowlarr]
2025-10-26 12:42:35.4|Info|AppFolderInfo|Data directory is being overridden to [/volume1/@appstore/prowlarr/var/.config/Prowlarr]
2025-10-26 12:42:35.5|Info|MigrationController|*** Migrating data source=/volume1/@appstore/prowlarr/var/.config/Prowlarr/prowlarr.db;cache size=-20000;datetimekind=Utc;journal mode=Wal;pooling=True;version=3;busytimeout=100 ***
2025-10-26 12:42:35.5|Info|FluentMigrator.Runner.MigrationRunner|DatabaseEngineVersionCheck migrating
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|PerformDBOperation 
2025-10-26 12:42:35.6|Info|NzbDrone.Core.Datastore.Migration.Framework.NzbDroneSQLiteProcessor|Performing DB Operation
2025-10-26 12:42:35.6|Info|DatabaseEngineVersionCheck|SQLite 3.46.1
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|=> 0.0306166s
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|DatabaseEngineVersionCheck migrated
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|=> 0.0340187s
2025-10-26 12:42:35.6|Info|MigrationController|*** Migrating data source=/volume1/@appstore/prowlarr/var/.config/Prowlarr/logs.db;cache size=-20000;datetimekind=Utc;journal mode=Wal;pooling=True;version=3;busytimeout=100 ***
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|DatabaseEngineVersionCheck migrating
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|PerformDBOperation 
2025-10-26 12:42:35.6|Info|NzbDrone.Core.Datastore.Migration.Framework.NzbDroneSQLiteProcessor|Performing DB Operation
2025-10-26 12:42:35.6|Info|DatabaseEngineVersionCheck|SQLite 3.46.1
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|=> 0.0037521s
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|DatabaseEngineVersionCheck migrated
2025-10-26 12:42:35.6|Info|FluentMigrator.Runner.MigrationRunner|=> 0.0039704s
2025-10-26 12:42:35.8|Info|Microsoft.Hosting.Lifetime|Now listening on: http://[::]:9696
2025-10-26 12:42:35.9|Info|CommandExecutor|Starting 3 threads for tasks.
2025-10-26 12:42:36.0|Info|ManagedHttpDispatcher|IPv4 is available: True, IPv6 will be disabled
2025-10-26 12:42:36.2|Info|Microsoft.Hosting.Lifetime|Application started. Press Ctrl+C to shut down.
2025-10-26 12:42:36.3|Info|Microsoft.Hosting.Lifetime|Hosting environment: Production
2025-10-26 12:42:36.3|Info|Microsoft.Hosting.Lifetime|Content root path: /volume1/@appstore/prowlarr/share/Prowlarr/bin
```

hgy59:
> This script will detect the DSM version and preserve the current `libe_sqlite3.so` we ship with it if running on DSM < 7.2. Early testing looks promising so far on DSM 6.

But this works only as long as Prowlarr is compatible with this libe_sqlite3.so.

When the first incomatible version of Prowlarr is released, we need to publish a package for DSM <= 7.1 with the latest compatible version and disabled autoupdate.
But it will be too late for the installations that did an autoupdate before updating to the package with the latest compatible version.

A more secure solution would be to disable the autoupdate and build packages for DSM <= 7.1 for every new release. So we can test for comatibiliy before publishing.

EDIT:
and keep the autoupdate for DSM >= 7.2 only

mreid-tt:
@hgy59 thanks for the feedback, this was going to be the other approach if the script proved unworkable. I was having some conversations with the Servarr people this morning on Discord (https://discord.com/channels/264387956343570434/816637845397307392/1431950932194627664).

I guess I'll work with the more secure solution as you recommend. I think we'll have to fundamentally change to have the package replace the binaries though. If we have a scenario where autoupdates are disabled on DSM 7.1 and Prowlarr is installed then the user updates their DSM to 7.2 they will need to reinstall the version for DSM 7.2 otherwise autoupdate will not enable. We will also need to have a separate 7.2 version in the repo where it is enabled.

EDIT: This can make things a bit tricky with one version allowing autoupdate and another not. I think we need to standardise on one which is why I liked the script route. The Servarr folks said "_it needs something newer than 3.10 which was released in 2016_". As such, I thought we would be safe with the internal updates for a while.

hgy59:
> then the user updates their DSM to 7.2 they will need to reinstall the version for DSM 7.2 otherwise autoupdate will not enable.

No need to reinstall, we only have to create a new package for DSM 7.2 after the final package for DSM <= 7.1

Some other thoughts:
- I suppose most of the users already updated to DSM 7.1 or DSM 7.2 when supported (there are models that do not support DSM 7 or DSM 7.2)
- a user can manually patch the INFO file to update to the installed version without prior uninstall
- When we create DSM 7.2 packages with a higher PKG_REV than the DSM <= 7.1 packages, a package upgrade will be available after update to DSM 7.2
- When we published the final package for DSM <= 7.1 we should publish a package for DSM 7.2 with at least a higher PKG_REV

mreid-tt:
> No need to reinstall, we only have to create a new package for DSM 7.2 after the final package for DSM <= 7.1

I guess you may be right. The thing that blocks autoupdate is a line in the `package_info` that says `UpdateMethod=External`. This file is replaced with a new one from the package on install so if they install the DSM 7.2 version it should remove this line and the configuration in the `config.xml` should take over.

hgy59:
@mreid-tt another proposal:

we can create a final package of Prowlarr [v1.37.0.5076](https://github.com/Prowlarr/Prowlarr/releases/tag/v1.37.0.5076) for DSM <= 7.1 with disabled autoupdate (do not publish for DSM 7.2 and don't disable autoupdate when installed on DSM 7.2)

Than we will create packages of Prowlarr 2.x for DSM 7.2 only (and keep the autoupdate).
i.e. 
- use REQUIRED_MIN_DSM = 7.2 for Prowlarr >= 2.x
- get rid of cross/libstdc++ and cross/libe_sqlite3
- keep autoupdate


The release plan will be like this
1. disable all packages already published as 20251012-9
2. create package of Prowlarr v1.37.0.5076 and publish for DSM 6.2.4 only as (date of build)-10
3. recreate package of Prowlarr v2.1.5.5216 and publish for DSM 7.2 only as (date of build)-11
4. create package of Prowlarr v1.37.0.5076 and publish for DSM 7.1 only as (same date as DSM 6.2.4 package)-10
The step 4 requires manually defining the SPK_VERS to the same value we created the DSM 6.2.4 packages in step 2)

Or a variant with less effort
1. disable all packages already published as 20251012-9
2. recreate package of Prowlarr v2.1.5.5216 and publish for DSM 7.2 only as (date of build)-11
3. create package of Prowlarr v1.37.0.5076 and publish for DSM 6.2.4 and DSM 7.1 only as 20250603-10
The step 3 requires defining SPK_VERS = 20250603  (must be between 20250601 and 20251012, 20250603 is the date 1.37.0.5076 was created)

finally we can drop the 20251012-9 releases in the repo.

hgy59:
> This file is replaced with a new one from the package on install so if they install the DSM 7.2 version it should remove this line and the configuration in the `config.xml` should take over.

No, this is not the case. None of the files in the `/var/packages/{package-name}/var` is overwritten at package update. The related files contained in the package will be renamed as *.new and the former files are unchanged.
This is the standard behaviour for all DSM 7+ packages (on DSM <= 6 this might need extra code AFAICR).

This means the package with disabled autoupdate must not only install a config.xml with UpdateAutomatically=False, but on package upgrade it must patch the installed config.xml to replace UpdateAutomatically=True with UpdateAutomatically=False (and may be change UpdateMechanism from BuiltIn to External).

Further the package with disabled autoupdate should contain a note in the package description that this package is the final release for DSM < 7.2 (yes, in the package description, not only in the changelog).

EDIT
oops, changed comfig.toml to config.xml

mreid-tt:
> No, this is not the case. None of the files in the `/var/packages/{package-name}/var` is overwritten at package update. The related files contained in the package will be renamed as *.new and the former files are unchanged. This is the standard behaviour for all DSM 7+ packages (on DSM <= 6 this might need extra code AFAICR).

The file in question is not in that path and this code explicitly excludes that file from being restored from the backup:

https://github.com/SynoCommunity/spksrc/blob/905f1d3ccf0fce46f773ffd5953618dd51d3293e/spk/prowlarr/src/service-setup.sh#L42-L51

I have confirmed this by installing the current version of Prowlarr for DSM 7.1 in the repo which gives the following output:

```
$ cat /var/packages/prowlarr/target/share/Prowlarr/package_info 
PackageVersion=x64-7.1_20251012-9
PackageAuthor=[Team Prowlarr](https://prowlarr.com)
```

Following the installation of Prowlarr for DSM 7.1 from this PR, it shows the following:

```
$ cat /var/packages/prowlarr/target/share/Prowlarr/package_info 
PackageVersion=x64-7.1_20251026-10
PackageAuthor=[Team Prowlarr](https://prowlarr.com)
UpdateMethod=External
```

This in turn blocks the user from applying any updates as shown below:

Screenshot 2025-10-26 at 5 41 21 PM

mreid-tt:
> we can create a final package of Prowlarr [v1.37.0.5076](https://github.com/Prowlarr/Prowlarr/releases/tag/v1.37.0.5076) for DSM <= 7.1 with disabled autoupdate (do not publish for DSM 7.2 and don't disable autoupdate when installed on DSM 7.2)
> 
> Than we will create packages of Prowlarr 2.x for DSM 7.2 only (and keep the autoupdate).

I am unclear on the reason for this proposal. We already have a working solution for Prowlarr v2.1.5.5216 for DSM <= 7.1. Why are we restricting users from Prowlarr 2.x? My suggestion is that we make Prowlarr v2.1.5.5216 the new baseline. That way we can disable autoupdate for the non-DSM 7.2 versions.

hgy59:
> Why are we restricting users from Prowlarr 2.x?

My proposal was based on the fact that the DSM 6.2 packages are disabled in the repo and the guess that it does not work on DSM 7.1 too.

mynameisbogdan:
> $ cat /var/packages/prowlarr/target/share/Prowlarr/package_info 
> PackageVersion=x64-7.1_20251026-10
> PackageAuthor=[Team Prowlarr](https://prowlarr.com)
> UpdateMethod=External

I recommend using something like `PackageAuthor=[SynoCommunity](https://github.com/SynoCommunity/spksrc)` since you guys are actually the packagers, not the Prowlarr team.

mreid-tt:
> I recommend using something like `PackageAuthor=[SynoCommunity](https://github.com/SynoCommunity/spksrc)` since you guys are actually the packagers, not the Prowlarr team.

Good point. Done.

mreid-tt:
> My proposal was based on the fact that the DSM 6.2 packages are disabled in the repo and the guess that it does not work on DSM 7.1 too.

I like the general concept, but I think we should designate v2.1.5.5216 as the point where updates via the internal updater are blocked for DSM < 7.2. The reasoning is that users with automatic updates enabled would likely have already received this version, which would either fail to start (on DSM < 7.2) or install successfully (on DSM ≥ 7.2).

I agree that the existing versions in the repo should be disabled to prevent users from initiating internal updates that could fail. This PR is positioned as follows:

1. For DSM ≥ 7.2: Automatic updates continue to function normally, with no binary replacement during install.
2. For DSM < 7.2: Automatic updates are disabled, but installing the package will replace the binaries from the package itself.

My suggested next steps:

1. Publish this version for DSM 6.2, 7.1, and 7.2.
   - For DSM < 7.2, the update will replace binaries, disable auto-updates, and repair any installs broken by previous updates.
   - For DSM ≥ 7.2, users should see no change — automatic updates will continue as normal.
2. After a few days, disable older versions in the repo, with the eventual goal of removing them entirely.

Note: DSM < 7.2 users will lose the ability to switch branches (e.g., master to develop) from the Prowlarr UI, since updates will no longer be possible there.

Let me know if I’ve missed anything or if you have further feedback on this approach.

hgy59:
> Let me know if I’ve missed anything or if you have further feedback on this approach.

I am still wondering, whether the following options in var/config.xml are completely ignored?
```
  <UpdateAutomatically>True</UpdateAutomatically>
  <UpdateMechanism>BuiltIn</UpdateMechanism>
```

Isn't it possible to avoid the `package_info` file and only use config.xml for such options?

mreid-tt:
> > Let me know if I’ve missed anything or if you have further feedback on this approach.
> 
> I am still wondering, whether the following options in var/config.xml are completely ignored?
> 
> ```
>   <UpdateAutomatically>True</UpdateAutomatically>
>   <UpdateMechanism>BuiltIn</UpdateMechanism>
> ```

I’ve confirmed this with the Servarr team on Discord — once you set it to external, all automatic updates are disabled (see [source code](https://github.com/Prowlarr/Prowlarr/blob/8aad1ac55497579c1afdc2d918f7ebe490fb4013/src/NzbDrone.Core/Update/InstallUpdateService.cs#L270-L274)).

> Isn't it possible to avoid the `package_info` file and only use config.xml for such options?

It is, but it’s not foolproof. We’d need to implement logic to modify existing config.xml files already in use, and even then, users could still visit the update page, see an available update, and manually tap Install Update (since the update mechanism would still be Built-In rather than External).

mreid-tt:
Hey @hgy59, we are having a number of users reporting problems with Prowlarr now. Are we good to go ahead with this PR?

EDIT: Sorry for not responding to these earlier:

> This means the package with disabled autoupdate must not only install a config.xml with UpdateAutomatically=False, but on package upgrade it must patch the installed config.xml to replace UpdateAutomatically=True with UpdateAutomatically=False (and may be change UpdateMechanism from BuiltIn to External).

This would not be required. As noted above once the UpdateMechanism is set to external, it will override what's in the `config.xml` (you see this overriding in the service log) and automatic updates will not function.

> Further the package with disabled autoupdate should contain a note in the package description that this package is the final release for DSM < 7.2 (yes, in the package description, not only in the changelog).

My intention is that this would not be the final release for DSM < 7.2 and that future updates will continue with periodic manual releases until we hit an incompatibility with our version of SQLite. At that point we can update the release notes to reflect that.

mreid-tt:
@hgy59, any final thoughts before I merge this?

spk/prowlarr/src/service-setup.sh
```diff
@@ -19,37 +19,52 @@ fi
...
 
+dsm_lt_7_2() {
+    if [ "${SYNOPKG_DSM_VERSION_MAJOR}" -lt 7 ] || \
+       { [ "${SYNOPKG_DSM_VERSION_MAJOR}" -eq 7 ] && [ "${SYNOPKG_DSM_VERSION_MINOR}" -lt 2 ]; }; then
+        return 0    # DSM < 7.2
+    else
+        return 1    # DSM ≥ 7.2
+    fi
+}
+
```
hgy59: In my opinion I would rename this function to something like `update_supported` and inverse the logic.
I don't like conditions like _if ! dsm_lt_7_2; then_.
it is more readable like ` if update_supported; then` ...

And the comments starting with `# DSM ≥ 7.2; ` are not needed anymore or, moved to the function.
And if the condition is ever changed, you don't need to adjust the function name...

mreid-tt: Done.

hgy59: I don't like the checks distributed into `Makefile` and `service-setup.sh`.
There are two different suggestions to change this.

1. Approach
   Read the UpdateMethod in `${SYNOPKG_PKG_DIR}/share/Prowlarr/package_info` instead of evaluation of DSM version again. But I am not aware, whether this file is already updated at the time of use.
   Or do you have another idea, how the `service-setup.sh` can use the `PROWLARR_UPDATE` variable in the Makefile?

2. Approach
   Omit the Makefile variable `PROWLARR_UPDATE` and adjust the `package_info` in `service-setup.sh` too.

hgy59: @mreid-tt sorry, my new proposal crossed your changes....

mreid-tt: Done, using approach 1(a).

mreid-tt:
@hgy59 I’ve updated the code so the previously evaluated condition is now only checked once for consistency. If there are no further concerns, I’ll proceed with merging this later today.

hgy59:
> @hgy59 I’ve updated the code so the previously evaluated condition is now only checked once for consistency. If there are no further concerns, I’ll proceed with merging this later today.

I suggest to remove any `DSM ≥ 7.2;` from the comments in `service-setup.sh`.
It now depends on `UpdateMethod=External` and does not care about the affected DSM versions.

mreid-tt:
> I suggest to remove any `DSM ≥ 7.2;` from the comments in `service-setup.sh`.
> It now depends on `UpdateMethod=External` and does not care about the affected DSM versions.

Done. All good now.

PR #6415, opened 2025-01-23
Transmission: Add v4.1.0-beta.1

mreid-tt:
## Description

This pull request includes the following:

1. Adds Transmission version 4.1.0-beta.1

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Package update

PR #6784, opened 2025-11-09
cmake-env.mk: Allow using BUILD_SHARED_LIBS and fix legacy behavior

th0ma7:
## Description

cmake-env.mk: Allow using `BUILD_SHARED_LIBS` and fix legacy behavior to not override shared library definition.

Relates to https://github.com/SynoCommunity/spksrc/pull/6415

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update


hgy59:
@th0ma7 This solution is only half the truth.

Good, you fixed the use of `BUILD_SHARED_LIBS` for builds without toolchain files.

But what, if you must suppress the definition of `-DBUILD_SHARED_LIBS` at all?

The definition of `-DBUILD_SHARED_LIBS=OFF` for transmission is only a workaround.
The real solution would be to omit the definition.

IMHO the `-DBUILD_SHARED_LIBS` should be used when building libraries only.
transmission is not a library but an executable that contains several third-party libraries in its own source code.

So I miss an option like `BUILD_SHARED_LIBS = DISABLED` to avoid adding `-DBUILD_SHARED_LIBS` to `CMAKE_ARGS`.


In the transmission package this popped up, because of an upstream _issue_.
For other third-party libraries transmission has dedicated options to create static libs (like `-DUPNPC_BUILD_STATIC=ON` or `-DLIBDEFLATE_BUILD_SHARED_LIB=OFF`), but this is missing for the crc32c lib, that was added recently.
transmission has also third-party libraries that are used as shared libraries and it's pure chance that our workaround works.

th0ma7:
When I first added CMake functionality years ago I somehow enforced `BUILD_SHARED_LIBS` to be always `ON` unless already set (no clue why, probably thought it made sense).  Later I added meson and I did <u>not</u> enforce building as shared library by default (neither was the original framework using GNU configure).

Issue may be twofold:
1. It is feasible to add a workaround to disable this flag altogether, but that would really just be a workaround.  I am rather presuming `transmission` has an issue within its CMake configuration files.
2. We could remove that shared libs option from the default CMake configuration altogether, assuming a similar behavior than GNU configure and meson.  Although this "might" break many packages (or not) where the fix would be to re-add it to their respective makefile... 

@hgy59 Also, to return to your point above, I don't believe there is any way to set shared library building only for libraries per-say.  For transmission specifically, couldn't we simply patch the code to add a `-DCRC32C_BUILD_STATIC=ON` or similar?

PR #6785, opened 2025-11-11
add new package: pigz for parallel gzip compression

Musicminion:
## Description

This pr add a new cmd tool `pigz`, for parallel gzip compression. And I add this tool to `spk/synocli-file` because it seems more like a file tool.

Nowadays, most Synology NAS have multi-core processor (like j4125, v1500b...). When using tar gz, we can add pigz to perform parallel compression to improve speed.

Btw, this is my first attempt at packaging, please help me fix the bugs I don't know that I've created yet, much thanks~

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update


hgy59:
@Musicminion  thanks for this.
Can you please create a dedicated branch? You will get serious troubles using the master branch for PRs.

Otherwise it looks good to me.
We are used to create something like `spksrc/diyspk/pigz` for tools that are part of larger synocli-* packages.
This helps adding/updating tools, since it does not require to build all dependent tools.
With this you can even install a single tool as a package (when you have a local dev env).

hgy59:
@Musicminion some additional comments (after looking into the code)

The source does not provide a configure script, so it is ok to skip this step.
And the Makefile does not support an install target, so the custom install is required.

But you shouldn't need to modify the `COMPILE_MAKE_OPTIONS` variable as you did.
But I guess the Makefile should be patched to make the `CC`, `CFLAGS` and `LDFLAGS` variables optional and use the variables of the spksrc dev env.

So I propose to add 
```
ADDITONAL_CFLAGS = -O3 -Wall -Wextra -Wno-unknown-pragmas -Wcast-qual
```
and to patch the Makefile like this:

```patch
--- Makefile.orig	2019-04-11 08:39:47.000000000 +0000
+++ Makefile	2025-11-11 11:55:30.276311464 +0000
@@ -1,6 +1,6 @@
-CC=gcc
-CFLAGS=-O3 -Wall -Wextra -Wno-unknown-pragmas -Wcast-qual
-LDFLAGS=
+CC?=gcc
+CFLAGS?=-O3 -Wall -Wextra -Wno-unknown-pragmas -Wcast-qual
+LDFLAGS?=
 # CFLAGS=-O3 -Wall -Wextra -Wno-unknown-pragmas -Wcast-qual -g -fsanitize=thread
 # LDFLAGS=-g -fsanitize=thread
 # CFLAGS=-O3 -Wall -Wextra -Wno-unknown-pragmas -Wcast-qual -g -fsanitize=address
```


BTW:
The install target does not need to install the man file(s) since there is no man command under DSM.
(and remove it in PLIST)

Musicminion:
> hgy59

I have made a new Pr: #6796, thanks~

hgy59:
replaced by #6796

PR #6083, opened 2024-04-25
mono: update to v6.12.0.199

hgy59:
## Description

- update mono
- try to fix BTLS for ARMv5

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Package update

PR #6810, opened 2025-11-22
ruTorrent: Update to 5.2.10

### LINKED ISSUE #6724: [Package Update] rutorrent
### Software Name

Update to last stable 

### Brief Description

Please can you update rutorrent to last stable version? 

### Website

https://github.com/Novik/ruTorrent

### Documentation

_No response_

https://github.com/Novik/ruTorrent

### Source code

https://github.com/Novik/ruTorrent

### OpenSource License

MIT

### Other License

_No response_

mreid-tt:
## Description

1. Bump ruTorrent to 5.2.10
2. Bump rtorrent/libtorrent to 0.16.4
3. Align PHP dependency with DSM version (PHP 8.0 on DSM 7.0/7.1, 8.2 on DSM 7.2+)

Fixes #6724

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/spk/rutorrent/Makefile
+++ b/spk/rutorrent/Makefile
@@ -65,6 +65,7 @@ ifeq ($(call version_ge, ${TCVERSION}, 7.2),1)
 # Default to PHP 8.2 on DSM 7.2+ (PHP 8.0 not supported)
 SPK_DEPENDS = "WebStation:PHP8.2:Apache2.4:${PYTHON_PACKAGE}:mediainfo"
 CONF_DIR = src/conf_72
+OS_MAX_VER = 7.2-79999
 else
 # Default to PHP 8.0 on DSM 7.0/7.1
 SPK_DEPENDS = "WebStation:PHP8.0:Apache2.4:${PYTHON_PACKAGE}:mediainfo"


mreid-tt:
hey @smaarn, I've gone ahead and updated ruTorrent in a number of areas. Most importantly we are changing the PHP back-end to support newer DSM versions. Let me know if you have time for a review.

cross/libmediainfo-25.04/Makefile
```diff
@@ -0,0 +1,23 @@
+PKG_NAME = libmediainfo
+PKG_VERS = 25.04
```
smaarn: Maybe a naive question but why the two different versions of mediainfo libraries ?

spk/rutorrent/src/service-setup.sh
```diff
@@ -32,7 +26,7 @@ SERVICE_COMMAND="env RUTORRENT_WEB_DIR=${RUTORRENT_WEB_DIR} SYNOPKG_PKGVAR=${SYN
 
 fix_shared_folders_rights()
 {
-    local folder=$1
+    folder=$1
```
smaarn, line 35: Naive question: why removing the `local` ?

spk/rutorrent/src/service-setup.sh
```diff
@@ -96,7 +94,7 @@ service_postinst ()
                "${RTORRENT_RC}"
 
         if [ -n "${wizard_watch_dir}" ]; then
-            local effective_watch_dir="${wizard_download_dir}${wizard_watch_dir}"
+            effective_watch_dir="${wizard_download_dir}${wizard_watch_dir}"
```
smaarn: Same here (why removing `local` ?)

spk/rutorrent/src/service-setup.sh
```diff
@@ -138,19 +134,19 @@ service_postuninst ()
 
 service_save ()
 {
-    local source_directory="${RUTORRENT_WEB_DIR}"
-    if [ -d "${DSM6_WEB_DIR}/${PACKAGE}" ] && [ ! -f "${SYNOPKG_PKGVAR}/.dsm7_migrated" ]; then
+    source_directory="${RUTORRENT_WEB_DIR}"
+    if [ -d "${DSM6_WEB_DIR}/${PACKAGE}" ]; then
```
smaarn: I have a feeling that the upgrade procedure is not purging the old DSM6 resources. Won't this mean that the logic will be done over and over again until the resources are actually cleared ?

smaarn (commented):
There are a few changes which seem to be some kind of convention changes (this thing migrating `\`` to `$(` which is beyond my expertise I must confess.

I only have a few comments / questions (would need to test it, will try to see if I can do that by EOD)

cross/libmediainfo-25.04/Makefile
mreid-tt: We hit an inflection point when upgrading past v25.04 and encountered build incompatibilities with DSM 6, which also affected other packages. Since this package targets DSM 7, it can use the newer version without any issues. However, additional code was needed to support other packages that cannot run the newer version — which is why two versions now exist in the repo.

spk/rutorrent/src/service-setup.sh
mreid-tt: `local` was removed because the scripts are intended to target `#!/bin/sh`. According to https://www.shellcheck.net/wiki/SC3043, `local` is not defined in POSIX `sh`.

spk/rutorrent/src/service-setup.sh
mreid-tt: Same as above

spk/rutorrent/src/service-setup.sh
mreid-tt: This part was somewhat of an educated guess. It appeared that the `.dsm7_migrated` file was always created, even when the system wasn’t actually upgrading from DSM 6. I assumed that the presence of a folder in the old path would be enough to detect this and trigger moving the content.

If that assumption is incorrect, I’m open to suggestions on a better approach for cleaning up DSM 6 data.

hgy59:
@mreid-tt please omit the php version in the profile description for all php profiles.
The php profile list in DSM shows the PHP version as a dedicated column. 
But the main reason is, that the description in the php profile is not updated on package updates.
i.e. when Synology provides PHP 8.4 and we update the package for PHP 8.4 the profile description will still stay "PHP 8.2 profile for ruTorrent".

cross/libmediainfo-25.04/Makefile
smaarn: Fair enough.

spk/rutorrent/src/service-setup.sh
smaarn: To be fair `/bin/sh` seems to be a redirect to `bash` on my NAS.

Either we should be forcing `/bin/bash` or indeed doing the change.

Up yo you.

spk/rutorrent/src/service-setup.sh
smaarn: Well, that's the point, the sole folder existence wasn't sufficient enough (since it's not purged and given the "wild side of the upgrade process" I wasn't under the impression there was a "silver bullet" option).

> If that assumption is incorrect, I’m open to suggestions on a better approach for cleaning up DSM 6 data.

Well, I wasn't under the impression that there was an initiative around it and it seemed to work fine until now.
I have no suggestion as of now for any better approach (would need to take some time to think about it, which I don't have right now).

There could also be the option of saying "hey we no longer support that" (migrating from previous DSM 6 compliant version) and only support migrating from a DSM 7 only supporting "previous version" and documenting how to proceed (like here we could be saying "we only support upgrading from 18 version" for example).

spk/rutorrent/Makefile
```diff
@@ -1,13 +1,13 @@
 SPK_NAME = rutorrent
-SPK_VERS = 5.1.12
-SPK_REV = 18
+SPK_VERS = 5.2.10
+SPK_REV = 19
 SPK_ICON = src/rutorrent.png
 DSM_UI_DIR = app
 
 MAINTAINER = smaarn
```
smaarn: TBH I'm being listed as a maintainer but I really don't use it nor have I been monitoring its evolution and since you've been doing all the heavy lifting lately on this it only make sense to update this.

```suggestion
MAINTAINER = mreid-tt
```

spk/rutorrent/src/service-setup.sh
smaarn: 🙀 

```
eisbear@mallinas2:~$ sh --version
GNU bash, version 4.4.23(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

spk/rutorrent/src/service-setup.sh
mreid-tt: > There could also be the option of saying "hey we no longer support that" (migrating from previous DSM 6 compliant version) and only support migrating from a DSM 7 only supporting "previous version" and documenting how to proceed (like here we could be saying "we only support upgrading from 18 version" for example).

I think I prefer that option, this app has been on DSM 7 for a while now. That way I could also remove the bunch of code that is there to fix earlier `rtorrent.rc` files. Will look into it.

spk/rutorrent/Makefile
mreid-tt: Okay, I'll just change it to `synocommunity` then.

spk/rutorrent/src/service-setup.sh
hgy59: Only DSM < 6 and SRM are lacking bash. Those have ash (of busybox) instead.

So only packages that might be published for DSM 5.2 and for SRM must be /bin/sh (i.e. ash) compatible.

hgy59:
regarding libmediainfo:
I propose to stay on version 25.04, since it seems that the latest version is not required by rutorrent.

If you really want to use 25.10 for newer compilers, you must update the mediainfo package too and drop support for DSM < 7.x.

I recently updated the mediainfo package and have chosen to use the latest version that builds for DSM 6.
The former version was 21.09 so an update was not often demanded.

mreid-tt:
> regarding libmediainfo: I propose to stay on version 25.04, since it seems that the latest version is not required by rutorrent.
> 
> If you really want to use 25.10 for newer compilers, you must update the mediainfo package too and drop support for DSM < 7.x.

Yes, both `libmediainfo` and `mediainfo` were updated to v25.10, and since this package already doesn’t support DSM < 7, I didn’t see any issue with moving to the latest version. I also updated the framework to handle different versions based on their GCC requirements, similar to how it’s done in other examples. That should keep things consistent for the other packages as well.

mreid-tt:
Hey @smaarn, I came across an interesting discussion here: [https://github.com/Novik/ruTorrent/issues/2983](https://github.com/Novik/ruTorrent/issues/2983) about the long-standing issue where newer rTorrent versions break the ratio plugin. I downloaded the latest rTorrent, applied one of the suggested patches plus one of my own, and it seems to build and install correctly.

Here’s the current run if you’d like to test it:
`---removed---`

I’m not planning to push this to the PR yet, since the latest rTorrent release has an open bug ([https://github.com/rakshasa/rtorrent/issues/1650](https://github.com/rakshasa/rtorrent/issues/1650)). I’ll wait for the next version and try again once it’s out.

My limited tests (on DSM 7.1) so far show:

* clean install
* ratio plugin working
* able to download a test torrent

—but that’s as far as I’ve tested for now.

hgy59:
> Yes, both `libmediainfo` and `mediainfo` were updated to v25.10, and since this package already doesn’t support DSM < 7, I didn’t see any issue with moving to the latest version. I also updated the framework to handle different versions based on their GCC requirements, similar to how it’s done in other examples. That should keep things consistent for the other packages as well.

This is a missunderstanding.
you splitted mediainfo to cross/mediainfo, cross/mediainfo-25.04 and cross/mediainfo-latests
but for spk/mediainfo we only need cross/mediainfo depending on cross/libmediainfo-latest.
We shouldn't provide mutliple versions if not required.

OK, now I can see that spk/rutorrent/Makefile depends on cross/mediainfo
I thought that it depends on cross/libmediainfo...
Why does rutorrent depend on cross/mediainfo? does it need the mediainfo tool at runtime or does it need the libraries at build time?

If rutorrent only needs mediainfo at runtime, we should remove the mediainfo dependency completely and add SPK_DEPENDS  = mediainfo instead.

If rutorrent requires  mediainfo at build time, I guess it can depend on cross/libmediainfo instead of cross/mediainfo.

The least required change is to modify spk/mediainfo/Makefile to depend on either cross/mediainfo-25.04 OR on cross/mediainfo-latest but not on a gcc version dependent cross package. Otherwise we will build packages for DSM 6 with invalid version number.

---

further thoughts:
also nzbdrone (aka sonarr3) depends on cross/mediainfo and I guess it is the same story here (should depend on cross/libmediainfo or SPK_DEPENDS on mediainfo)
**BUT this brings a different topic to discussion.**
IMHO the sonarr3 package is outdated and should be removed from spksrc and from the synocommunity repo.
this includes the drop of the mono sources and packages (including the open PR #6083).
_(and this will free 2.7 GB of the repo storage)_

/CC @th0ma7 @publicarray

mreid-tt:
> This is a missunderstanding. you splitted mediainfo to cross/mediainfo, cross/mediainfo-25.04 and cross/mediainfo-latests but for spk/mediainfo we only need cross/mediainfo depending on cross/libmediainfo-latest. We shouldn't provide mutliple versions if not required.

Hmm, the reason I split both mediainfo and libmediainfo is because they have different releases upstream and I assumed that they needed to have matching versions. i.e. We shouldn't have a cross/mediainfo that may depend on cross/libmediainfo-latest when it really needs cross/libmediainfo-25.04 as I assumed it needed a matching version to build correctly.

> OK, now I can see that spk/rutorrent/Makefile depends on cross/mediainfo I thought that it depends on cross/libmediainfo... Why does rutorrent depend on cross/mediainfo? does it need the mediainfo tool at runtime or does it need the libraries at build time?
> 
> If rutorrent only needs mediainfo at runtime, we should remove the mediainfo dependency completely and add SPK_DEPENDS = mediainfo instead.
> 
> If rutorrent requires mediainfo at build time, I guess it can depend on cross/libmediainfo instead of cross/mediainfo.

I'm not exactly sure what type of dependency it needs but I assume it has something to do with the mediainfo plugin (https://github.com/Novik/ruTorrent/wiki/PluginMediainfo). This is not a package I'm intimate with and it has many layers so I was trying to make minimal changes.

> The least required change is to modify spk/mediainfo/Makefile to depend on either cross/mediainfo-25.04 OR on cross/mediainfo-latest but not on a gcc version dependent cross package. Otherwise we will build packages for DSM 6 with invalid version number.

If I"m understanding you correctly, you are concerned about the spk/mediainfo package possibly building from the cross/mediainfo based on the GCC version which could result in the 25.04 version building on a 25.10 cross if it's a newer GCC? If so that is a fair call and we would probably need to change spk/mediainfo to just depend on latest and restrict the spk/mediainfo to DSM 7 only. If I'm not understanding you correctly please let me know.

hgy59:
looking deeper:

rutorrent needs the mediainfo binary for the mediainfo plugin. The path to mediainfo is defined in config.php and the code to patch this file is in service-setup.sh. So it is easy to remove mediainfo from the rutorrent package and depend on the mediainfo package and adjust the path.

Analysis of service-setup.sh showed another issue.
The path to php is patched to /bin/php in config.php and this must be fixed to use the correct php binary of version 8.0 or 8.2.

ATM the config.php file is updated on package install but not on upgrade, so this must be fixed for upgrades too,


Additionally I propose to remove the DSM 6 to DSM 7 migration in this release and, for the rare case that someone still tries to migrate from DSM 6, to display an error that to update rutorrent a manuall update with package version 5.1.12-18 has to be done first (and for some time we must not remove rutorrent 5.1.12-18 in the repo).
It is now time to get rid of the extra migration code in service-setup.sh.
If we support package update only from 5.1.12-18, we can simplify code in service_save and service_restore in service-setup.sh.

Another proposal is to use pgrep of cross/procps-ng to get rid of the busybox stuff (dependency and configuration for build, and initialization in service-setup).

cross/xmlrpc-c/Makefile
```diff
@@ -14,7 +14,19 @@ LICENSE  = BSD
...
 
+POST_INSTALL_TARGET = xmlrpc_c_post_install
+
 include ../../mk/spksrc.cross-cc.mk
+
+.PHONY: xmlrpc_c_post_install
+xmlrpc_c_post_install:
+	# make xmlrpc-c-config use xml2-config from the same dir
+	sed -i -e 's|LIBXML=`xml2-config --libs`|LIBXML=`"$$\(dirname "$$0")/xml2-config" --libs`|' \
+		$(STAGING_INSTALL_PREFIX)/bin/xmlrpc-c-config
```
hgy59, line 32: just wondering whether this does not work as it did before:

```
# requires xml2-config in path to build libxml2-backend
ENV += PATH=$(INSTALL_DIR)/$(INSTALL_PREFIX)/bin:$$PATH
```

mreid-tt: As far as I can see, this hasn't worked for some time. In the logs you would see these lines:

```
checking for XMLRPC-C... /spksrc/spk/rutorrent/work-x64-7.1/install/var/packages/rutorrent/target/bin/xmlrpc-c-config: 1: xml2-config: not found /spksrc/spk/rutorrent/work-x64-7.1/install/var/packages/rutorrent/target/bin/xmlrpc-c-config: 1: xml2-config: not found
ok
ignored
```

It would not stop the build but this was part of me trying to apply fixes for the newer rtorrent versions to work smoothly.

With this change the logs now show:

```
checking for XMLRPC-C... ok
ignored
```

spk/rutorrent/Makefile
```diff
@@ -1,13 +1,13 @@
 SPK_NAME = rutorrent
-SPK_VERS = 5.1.12
-SPK_REV = 18
+SPK_VERS = 5.2.10
+SPK_REV = 19
 SPK_ICON = src/rutorrent.png
 DSM_UI_DIR = app
 
-MAINTAINER = smaarn
+MAINTAINER = SynoCommunity
```
hgy59, line 7: @mreid-tt I would recommend that you take the ownership, because this package is very special and has a lot of sofisticated code and therefore needs a dedicated maintainer.

mreid-tt:
> rutorrent needs the mediainfo binary for the mediainfo plugin. The path to mediainfo is defined in config.php and the code to patch this file is in service-setup.sh. So it is easy to remove mediainfo from the rutorrent package and depend on the mediainfo package and adjust the path.

Okay so we can switch the package to have mediainfo as a package dependency rather than build dependency. I'm okay with that. @smaarn any objections?

> Analysis of service-setup.sh showed another issue. The path to php is patched to /bin/php in config.php and this must be fixed to use the correct php binary of version 8.0 or 8.2.

Good catch. I've tested on a DSM 7.3 box and you're right that the existing path with pick up the OS version as follows:

```
$ /bin/php --version
PHP 8.1.32 (cli) (built: Jun 20 2025 08:09:01) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.32, Copyright (c) Zend Technologies
```

Whereas a specific path to the installed expected version would report this:

```
$ /usr/local/bin/php82 --version
PHP 8.2.28 (cli) (built: Jun 16 2025 16:09:43) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.28, Copyright (c) Zend Technologies
```

> ATM the config.php file is updated on package install but not on upgrade, so this must be fixed for upgrades too,

Agreed, as this could create a potential edge-case where we would need to re-evaluate this if the user upgrades their DSM version and they then need a package upgrade but the package dependencies may no longer match what's in the config.php. Will think on this some more.

> Additionally I propose to remove the DSM 6 to DSM 7 migration in this release and, for the rare case that someone still tries to migrate from DSM 6, to display an error that to update rutorrent a manuall update with package version 5.1.12-18 has to be done first (and for some time we must not remove rutorrent 5.1.12-18 in the repo). It is now time to get rid of the extra migration code in service-setup.sh. If we support package update only from 5.1.12-18, we can simplify code in service_save and service_restore in service-setup.sh.

I thought I removed some of it but will look again.

> Another proposal is to use pgrep of cross/procps-ng to get rid of the busybox stuff (dependency and configuration for build, and initialization in service-setup).

I'm not familiar with this so will need to do some research.

hgy59:
> I'm not familiar with this so will need to do some research.

only this is needed:
1. replace `DEPENDS += cross/busybox` by `DEPENDS += cross/procps-ng`  in spk/rutorrent/Makefile (the comment still matches)
2. remove `export BUSYBOX_CONFIG=procutils` (in the same file)
3. remove these lines in service-setup.sh
   ```
    # Install busybox stuff
    "${SYNOPKG_PKGDEST}/bin/busybox" --install "${SYNOPKG_PKGDEST}/bin"
    ```

smaarn:
> > rutorrent needs the mediainfo binary for the mediainfo plugin. The path to mediainfo is defined in config.php and the code to patch this file is in service-setup.sh. So it is easy to remove mediainfo from the rutorrent package and depend on the mediainfo package and adjust the path.
> 
> Okay so we can switch the package to have mediainfo as a package dependency rather than build dependency. I'm okay with that. @smaarn any objections?
> 

No objection. I would even recommend it. Personally I think it's easier to maintain this way since it makes maintenance decorrelated.

> > ATM the config.php file is updated on package install but not on upgrade, so this must be fixed for upgrades too,
>
> Agreed, as this could create a potential edge-case where we would need to re-evaluate this if the user upgrades their DSM version and they then need a package upgrade but the package dependencies may no longer match what's in the config.php. Will think on this some more.

Actually those are defined by the below kind of lines:

```
    if is_not_defined_external_program 'php'; then
        define_external_program 'php' '/bin/php' '/usr/bin/php'
    fi
```

If you wish to force the PHP binary path you can use those sections but I'm not sure it's even relevant (last time I checked it didn't cause an issue).

mreid-tt:
Hey @hgy59 @smaarn, I think I’ve made some solid progress on the update. Now that the latest rTorrent/libtorrent (0.16.4) releases have landed, I’ve integrated them into the package. I also removed the version bumps and the mediainfo split, switching to the existing external package instead. As suggested, I’ve dropped BusyBox and refactored a good portion of the service-setup script to address DSM 7 concerns, removing the older DSM 6 logic entirely.

I’ve tested the build successfully on DSM 7.1 and 7.2, but DSM 7.3 is proving problematic. There’s very little in the logs to point to the issue. The `diskspace` plugin fails to load from the plugins tab, and on DSM 7.3 the watch folder never picks anything up. Torrents added manually remain stuck in a “Pausing” state and never download.

I’ve kicked off a workflow run here so you can try the DSM 7.2+ build on DSM 7.3 if you have a moment:
`---removed---`

If we can’t resolve the DSM 7.3 issues, I’ll exclude it from the supported versions for now.

At this point, the only remaining cleanup I can see is the installation wizard. Let me know if your reviews highlight anything else.

smaarn (approved):

th0ma7 (approved):
Really nice work, LGTM!

smaarn:
I get the below error when starting the UI, not sure of the impact though:

```
[29.11.2025 19:41:02] JS error: [http://mallinas2.local/rutorrent/ : 98] ReferenceError: Can't find variable: theOptionsWindow
```

spk/rutorrent/src/service-setup.sh
```diff
@@ -34,57 +41,98 @@ fix_shared_folders_rights()
...
+            }
+            END {
+                exit(inserted ? 0 : 1)
+            }
+        ' "${sep_config_file}" > "${tmp_file}"; then
+            cat "${tmp_file}" > "${sep_config_file}"
+            rm -f "${tmp_file}"
+            chown "${EFF_USER}:${APACHE_USER}" "${sep_config_file}" 2>/dev/null || true
+            chmod 0664 "${sep_config_file}" 2>/dev/null || true
+        fi
```
smaarn, line 102: Though impressive this can be I would only be wary of the potential maintenance cost 😅

smaarn:
> I get the below error when starting the UI, not sure of the impact though:
> 
> ```
> [29.11.2025 19:41:02] JS error: [http://mallinas2.local/rutorrent/ : 98] ReferenceError: Can't find variable: theOptionsWindow
> ```

When opening the console it seems that it's related to the `autotools` plugin which fails loading:

```
Plugin "autotools" failed to load:
ReferenceError: Can't find variable: theOptionsWindow
(fonction anonyme) — rutorrent:98:583
(fonction anonyme) — rutorrent:131:3854
(fonction anonyme) — plugins.js:164
Code général — rutorrent:29
appendChild
m — jquery.js:2:892
globalEval — jquery.js:2:3033
text script — jquery.js:2:81005
(fonction anonyme) — jquery.js:2:77351
l — jquery.js:2:77467
(fonction anonyme) — jquery.js:2:80267
```

spk/rutorrent/src/service-setup.sh
mreid-tt: This was just a nitpick of mine with the adding of the external applications. I liked to see it within the block with the others and not at the end of the config file. That is its only purpose. It can revert to your original approach if preferred.

mreid-tt:
> > I get the below error when starting the UI, not sure of the impact though:
> > ```
> > [29.11.2025 19:41:02] JS error: [http://mallinas2.local/rutorrent/ : 98] ReferenceError: Can't find variable: theOptionsWindow
> > ```
> 
> When opening the console it seems that it's related to the `autotools` plugin which fails loading:
> 
> ```
> Plugin "autotools" failed to load:
> ReferenceError: Can't find variable: theOptionsWindow
> (fonction anonyme) — rutorrent:98:583
> (fonction anonyme) — rutorrent:131:3854
> (fonction anonyme) — plugins.js:164
> Code général — rutorrent:29
> appendChild
> m — jquery.js:2:892
> globalEval — jquery.js:2:3033
> text script — jquery.js:2:81005
> (fonction anonyme) — jquery.js:2:77351
> l — jquery.js:2:77467
> (fonction anonyme) — jquery.js:2:80267
> ```

Can you tell me a bit more about your setup? Running in DSM 7.1 or 7.2 I have no entries in my console and the plugins all look clean:

Screenshot 2025-11-29 at 3 23 14 PM

EDIT: I've found ruTorrent to be finicky in the past with the UI. Does it still happen if you open in a new private window?

smaarn:
> > > I get the below error when starting the UI, not sure of the impact though:
> > > ```
> > > [29.11.2025 19:41:02] JS error: [http://mallinas2.local/rutorrent/ : 98] ReferenceError: Can't find variable: theOptionsWindow
> > > ```
> > 
> > 
> > When opening the console it seems that it's related to the `autotools` plugin which fails loading:
> > ```
> > Plugin "autotools" failed to load:
> > ReferenceError: Can't find variable: theOptionsWindow
> > (fonction anonyme) — rutorrent:98:583
> > (fonction anonyme) — rutorrent:131:3854
> > (fonction anonyme) — plugins.js:164
> > Code général — rutorrent:29
> > appendChild
> > m — jquery.js:2:892
> > globalEval — jquery.js:2:3033
> > text script — jquery.js:2:81005
> > (fonction anonyme) — jquery.js:2:77351
> > l — jquery.js:2:77467
> > (fonction anonyme) — jquery.js:2:80267
> > ```
> 
> Can you tell me a bit more about your setup? Running in DSM 7.1 or 7.2 I have no entries in my console and the plugins all look clean:
> 
> Screenshot 2025-11-29 at 3 23 14 PM
> EDIT: I've found ruTorrent to be finicky in the past with the UI. Does it still happen if you open in a new private window?

Indeed reloading seems to make the error go away. 🤷 

(I'm using DSM 7.3 and Safari)

spk/rutorrent/src/service-setup.sh
smaarn: No worries, really. To be honest the first implementation was maybe a little bit fragile so fair enough.

mreid-tt:
> Indeed reloading seems to make the error go away. 🤷
> 
> (I'm using DSM 7.3 and Safari)

You have it working successfully on DSM 7.3?!? I couldn't get it to work at all. Do you have any of the errors I noted above?

EDIT: Because I couldn't get DSM 7.3 to work I included a [commit](https://github.com/SynoCommunity/spksrc/pull/6810/commits/f929502fafe432bb9fd1e183b607a56a61be89f2) to limit its installation to DSM 7.2.

EDIT: Merging as-is and we can re-visit DSM 7.3 support at a later date.

spk/rutorrent/src/service-setup.sh
mreid-tt: cool, just as a quick explainer on the function:

`set_external_program_path()` is intentionally small and defensive: it first tries the simple case—if an entry already exists in `$pathToExternals`, a busybox-friendly `sed` just swaps the value.

Only when the key is missing do we take the slower path: `awk` copies the file, injects a single `"program" => 'path'` line right before the array’s closing `);`, then writes it back while restoring ownership/perms. The script never rewrites the whole config or touches anything outside that array, so user comments/ordering stay intact; if it can’t find the array, it bails without altering the file.

In practice this keeps `config.php` aligned with the binaries we ship (e.g. moving from PHP 8.0 to 8.2) without asking users to edit by hand.

PR #6814, opened 2025-11-28
Nextcloud 32: Initial package release

### LINKED ISSUE #2403: [Package Request] NextCloud
**Name:** NextCloud
**Description:** This is the fork of OwnCloud. "Access & share your files, calendars, contacts, mail & more from any device, on your terms"
**Website:** https://nextcloud.com/
**Software documentation:** https://docs.nextcloud.org/server/9/user_manual/
**Build/Installation documentation:** https://docs.nextcloud.org/server/9/developer_manual/
**Source code:** https://github.com/nextcloud
**License:** GNU AFFERO GENERAL PUBLIC LICENSE (if i'm not wrong, from server/COPYING-AGPL in github)

mreid-tt:
## Description

- Add a DSM 7.2+ Nextcloud 32 package that consumes PHP 8.2 and aligns profiles/resources with Synology WebStation backends.
- Replace the inherited ownCloud logic with a DSM 7–only service script handling install/restore/upgrade, OCC maintenance tasks, cron emulation, and backup export.
- Refresh metadata (Makefile, wizard, language strings, icon, WebStation profiles) so the package installs with the right externals, performs trusted-domain setup, and enforces HTTPS headers.

Fixes #2403

## Checklist

### Type of change

- [x] New Package

### DIFF
--- a/spk/nextcloud/src/conf_72/resource
+++ b/spk/nextcloud/src/conf_72/resource
@@ -60,7 +60,9 @@
           "php_settings": {
             "error_log": "/var/services/web_packages/nextcloud/php_errors.log",
             "memory_limit": "512M",
+            "post_max_size": "16G",
             "opcache.interned_strings_buffer": "32",
+            "upload_max_filesize": "16G",
             "pdo_mysql.default_socket": "/run/mysqld/mysqld10.sock"
           },
           "profile_desc": "PHP profile for Nextcloud",


mreid-tt:
hey @th0ma7, while working on this I found a limitation in the package where the `$SHARE_PATH` variable isn't defined yet in the pre-install stage. As such, if I need to refer to it (as in this check) I have to use a workaround to derive my own:

https://github.com/SynoCommunity/spksrc/blob/87c5a3afc7ccadb4002be41d0f00b33777a6ab76/spk/nextcloud/src/service-setup.sh#L137-L148

I don't know if this was something I needed to do or if this was the best way but do let me know your thoughts on it.

th0ma7:
> hey @th0ma7, while working on this I found a limitation in the package where the `$SHARE_PATH` variable isn't defined yet in the pre-install stage. As such, if I need to refer to it (as in this check) I have to use a workaround to derive my own:
> 
> I don't know if this was something I needed to do or if this was the best way but do let me know your thoughts on it.

No clue really, would need to review this in depth.  Although I believe this refers to https://github.com/SynoCommunity/spksrc/pull/6163 - @hgy59 does this ring any bell to you?

spk/nextcloud/Makefile
```diff
@@ -0,0 +1,45 @@
...
+REQUIRED_MIN_SRM = 3.0
+
+MAINTAINER = SynoCommunity
+DESCRIPTION = Nextcloud is a self-hosted productivity platform giving you control over your data with web, desktop and mobile access.
+DISPLAY_NAME = Nextcloud
+CHANGELOG = "Initial packaging for Nextcloud 32.0.2."
+HOMEPAGE = https://nextcloud.com/
+
+LICENSE = AGPLv3
+LICENSE_FILE = $(WORK_DIR)/nextcloud/COPYING
```
hgy59, line 24: Do you really want the user to confirmat the license?

hgy59:
> > hey @th0ma7, while working on this I found a limitation in the package where the `$SHARE_PATH` variable isn't defined yet in the pre-install stage. As such, if I need to refer to it (as in this check) I have to use a workaround to derive my own:
> > I don't know if this was something I needed to do or if this was the best way but do let me know your thoughts on it.
> 
> No clue really, would need to review this in depth. Although I believe this refers to #6163 - @hgy59 does this ring any bell to you?

@mreid-tt you shouldn't use ${SHARE_PATH} in pre-install.
The path might not yet be created by DSM package installation.

hgy59:
@mreid-tt I see a problem in the following line
```
         echo "Cannot install because ${share_path}/data already contains existing Nextcloud data (.ncdata). Please rename or remove the folder and try again." 
```

If the package can't be installed when .ncdata already exists how can you do a package update?
OR
If the package can't be installed when .ncdata already exists, the .ncdata folder must not be created in the share but in SYNOPG_PKGVAR.

spk/nextcloud/Makefile
mreid-tt: I don't know what is the standard but from the other packages I've worked on, if there is a license file I typically point to it.

mreid-tt:
> @mreid-tt you shouldn't use ${SHARE_PATH} in pre-install.
> The path might not yet be created by DSM package installation.

@hgy59, I wasn't able to, and that was the point of raising this in the above https://github.com/SynoCommunity/spksrc/pull/6814#issuecomment-3590756420.

> If the package can't be installed when .ncdata already exists how can you do a package update?
> OR
> If the package can't be installed when .ncdata already exists, the .ncdata folder must not be created in the share but in SYNOPG_PKGVAR.

During testing I found that unlike ownCloud, Nextcloud has a challenge creating users with the same identification as part of the installation. This does not affect upgrades as we are not triggering the install function for those.

This is only for the specific edge case where Nextcloud was uninstalled and then you try to re-install it using the same share. All the user needs to do is rename the data folder within the share before proceeding. I can do this for the user if you think this would be better but I believed that the user needs to take this action themselves.

Installing the data folder in `SYNOPG_PKGVAR` is not an option as that breaks the purpose of having a separate share for data storage.

farfromrefug:
@mreid-tt does php have acpu with you package? that s one issue i have with nextcloud on latest php in synology is that synology php8+ does not have acpu/memcached

mreid-tt:
> @mreid-tt does php have acpu with you package? that s one issue i have with nextcloud on latest php in synology is that synology php8+ does not have acpu/memcached

Yes, memcached is already included as you can see from the resource file:

https://github.com/SynoCommunity/spksrc/blob/5396863fecc4c1bb625db957f8ab49e3ea1f32ab/spk/nextcloud/src/conf_72/resource#L48

From what I found, acpu does not need to be explicitly enabled as it is already present. Doing so just triggers log entries in PHP saying acpu is already present.

EDIT: PHP already has acpu and memcached but not all of them are shown in the UI or enabled by default. You can check with the command line to see which extensions are present.

farfromrefug:
@mreid-tt thanks. but i don't understand, you are using synology provided php packages?(8.2). Because the issue is on the php package side. They removed acpu/memcached so libs

mreid-tt:
> @mreid-tt thanks. but i don't understand, you are using synology provided php packages?(8.2). Because the issue is on the php package side. They removed acpu/memcached so libs

I don't believe we are on the same page. This is the output for my PHP 8.2 on DSM 7.3:

```
$ /usr/local/bin/php82 -m | grep -i memcached
memcached
```
```
$ /usr/local/bin/php82 -v
PHP 8.2.28 (cli) (built: Jun 16 2025 16:09:43) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.28, Copyright (c) Zend Technologies
```
```
$ ls /volume1/@appstore/PHP8.2/usr/local/lib/php82/modules/
apcu.so      ftp.so      intl.so       pdo_dblib.so   redis.so    sysvmsg.so
bcmath.so    gd.so       ldap.so       pdo_mysql.so   shmop.so    sysvsem.so
bz2.so       gettext.so  mailparse.so  pdo_pgsql.so   soap.so     sysvshm.so
calendar.so  gmp.so      memcached.so  pdo_sqlite.so  sockets.so  xdebug.so
curl.so      iconv.so    mysqli.so     pgsql.so       sodium.so   xsl.so
dba.so       imagick.so  opcache.so    phar.so        sqlite3.so  zip.so
exif.so      imap.so     openssl.so    posix.so       ssh2.so     zlib.so
```

As you can see from the above, both extensions are present.

farfromrefug:
@mreid-tt actually i was certain i looked before and it was not there. But it is! 
Did you manage to activate it? i cant enable it from php settings and i never managed to have nextcloud see it

mreid-tt:
> @mreid-tt actually i was certain i looked before and it was not there. But it is! Did you manage to activate it? i cant enable it from php settings and i never managed to have nextcloud see it

Yes, we've been using acpu/memcached with ownCloud for sometime now and it works the same with Nextcloud. Main difference is the newer PHP version which allows for DSM 7.3 support.

farfromrefug:
@mreid-tt ok  i ll wait for this package and test. With direct installation in /volume1/web and nextcloud config.php never had it to use acpu. Always reports error. And in web station i cant "enable" acpu/memcached as the modules do not appear in the settings
image

mreid-tt:
> @mreid-tt ok i ll wait for this package and test. With direct installation in /volume1/web and nextcloud config.php never had it to use acpu. Always reports error. And in web station i cant "enable" acpu/memcached as the modules do not appear in the settings

Understood. The Synology implementation of PHP can be annoying. You can test this build now actually as it's ready to merge. Just download from the `Artifacts` under the `Checks` tab. Once you get the file, unzip and then perform a manual install.

farfromrefug:
@mreid-tt awesome will do. Can i make your package my existing nextcloud data from /volume1/nextcloud/data and /volume1/nextcloud/config?

mreid-tt:
> @mreid-tt awesome will do. Can i make your package my existing nextcloud data from /volume1/nextcloud/data and /volume1/nextcloud/config?

Hmm, that may get a bit tricky. For this package the config is actually at `/var/services/web_packages/nextcloud/config`, the data is in the same location but the database details even if it's the same name (`nextcloud`) may be a bit different. Was this a package from another source or was it something you deployed yourself from scratch?

farfromrefug:
@mreid-tt i deployed from scratch years ago. It uses local mariadb and redis. I cant migrate to the package if i loose all my data.
Would also be nice to be able to decide config location so that it does not get removed upon package deletion.

mreid-tt:
> @mreid-tt i deployed from scratch years ago. It uses local mariadb and redis. I cant migrate to the package if i loose all my data. Would also be nice to be able to decide config location so that it does not get removed upon package deletion.

Thanks for the detail. I'll not be able to test a migration from your setup to this package so there is always the risk of data loss. However, the package includes a built-in backup and restore function for all configuration and data. One approach if you wish to transition would be to backup your existing configuration and data into an archive of the same structure, move/rename/remove your existing deployment and then on package install, restore from the archive you hand-crafted.

I know this sounds like a lot but if you are willing to try I could write up some steps for you which would take you though the process.

EDIT: Ideally if you are on a DSM that can run Virtual Machine Manager, you can simulate this entire process in VirtualDSM to minimise the risk of data loss.

EDIT: I had an AI generate the overall structure of the method I described above in this wiki: [Nextcloud → SynoCommunity Package Migration Backup](https://github.com/mreid-tt/spksrc/wiki/Nextcloud-%E2%86%92-SynoCommunity-Package-Migration-Backup). If you’re using VirtualDSM, you can follow it up to step 9 to generate the archive, then test a fresh install and restore from that file to confirm the migration works before performing it in your production environment.

farfromrefug:
@mreid-tt nice! i ll try to take a look, but my nextcloud data is 46gb. Will make a big zip. On question still. Where is nextcloud data with that package? i am scared of using it in the sense that if it is a "package" path, then removing the package would remove my data? I would really like it to be a shared folder like /volume1/nextcloud

mreid-tt:
@farfromrefug, the package does use a shared folder with a default name which you can change (so `/volume1/nextcloud` by default) and within it a data folder is created.

On uninstall, the data folder is not removed. However for safety I would recommend a full backup.

mreid-tt:
> i ll try to take a look, but my nextcloud data is 46gb. Will make a big zip.

I would recommend first installing the base package in your VirtualDSM so you get a feel of how the install and uninstall works as a test. Examine the various concerns you had about acpu/memcached and then decide if you want to go through with your full production backup.

mreid-tt:
Hey @hgy59 @th0ma7, unless there are further comments I'll be merging this today.

th0ma7:
None on my end.

mreid-tt:
Looks like this is too big to publish... help!

```
<html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx/1.25.4</center>
</body>
</html>
```

hgy59:
> Looks like this is too big to publish... help!
> 
> ```
> <html>
> <head><title>413 Request Entity Too Large</title></head>
> <body>
> <center><h1>413 Request Entity Too Large</h1></center>
> <hr><center>nginx/1.25.4</center>
> </body>
> </html>
> ```

The limit is 200MB, but nextcloud package is 282 MB
we increased the limit from 170MB to 200MB in https://github.com/SynoCommunity/spkrepo/issues/82 and from 128MB to 170MB in https://github.com/SynoCommunity/spkrepo/pull/30

hgy59:
maybe we can create a separate package with the nextcloud apps?
The apps folder is 450MB of 760MB (of unkompressed package)

hgy59:
Since cross/nextcloud does not build any code, it would be better, to download nextcloud by the package installer at installation time. This would help to keep the package small (and the repo disk space too).

mreid-tt:
> Since cross/nextcloud does not build any code, it would be better, to download nextcloud by the package installer at installation time. This would help to keep the package small (and the repo disk space too).

Hmm, open to suggestions on how to do this.

EDIT: Solution proposed in #6834

Issue #6773, opened 2025-11-03
Framework bugs since recent update


th0ma7:
Since the recent framework changes there are a few outstanding issues that needs fixing.

Original PR: #6738 

Issues:
- [x] status log output missing : #6768
- [ ] kernel build should stop on exit: https://github.com/SynoCommunity/spksrc/pull/6772#issuecomment-3482457300
- [ ] new delay in the build process https://github.com/SynoCommunity/spksrc/pull/6771#issuecomment-3480347287

PR #6826, opened 2025-12-04
Framework: Fix install-resources staging for patched payloads

mreid-tt:
## Description

- record the patched install-resources tree (source + destination) during cross builds so dependents reuse the modified payload
- merge install-resources artefacts from dependencies while copying, preferring the dependency’s workdir and skipping directories already present to avoid “file changed” errors

## Testing
- `make -C cross/tt-rss`
- `make -C spk/tt-rss arch-noarch-6.1`
- unpacked package to confirm patched classes/config.php contains `MYSQL_DB_SOCKET`

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix

### DIFF
--- a/mk/spksrc.install-resources.mk
+++ b/mk/spksrc.install-resources.mk
@@ -49,11 +49,15 @@ COMPILE_TARGET = nop
 # Include base cross-cc makefile for common functionality
 include ../../mk/spksrc.cross-cc.mk
 
-# Record the patched payload location so dependent packages can merge it during
-# their copy step.
 INSTALL_RESOURCES_INFO = $(WORK_DIR)/install-resources-$(PKG_NAME).info
 INSTALL_RESOURCES_DEST ?= share/$(PKG_NAME)
-INSTALL_RESOURCES_SRC  ?= $(WORK_DIR)/$(PKG_DIR)
+# Source used for the default copy helper (normally the patched
+# $(WORK_DIR)/$(PKG_DIR) tree).
+INSTALL_RESOURCES_COPY_SRC ?= $(WORK_DIR)/$(PKG_DIR)
+# Location recorded for downstream packages to merge. Default to the staging
+# install output so consumers always get the post-install (i.e. patched)
+# payload.
+INSTALL_RESOURCES_STAGE_SRC ?= $(STAGING_INSTALL_PREFIX)/$(INSTALL_RESOURCES_DEST)
 
 post_install_target: install_resources_info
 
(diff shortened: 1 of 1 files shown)


mreid-tt:
hey @th0ma7, this was an issue I came across recently. When building tt-rss, the package would build but it would not include the patch which is present in `cross/tt-rss`. If I build from that path directly it is included. It appears that patches in dependencies were being ignored when building from `spk/tt-rss`. I don't know if this may qualify as an item for https://github.com/SynoCommunity/spksrc/issues/6773 but would appreciate your review.

th0ma7:
Can you provide the exact working vs non working command lines you're using?  Which is which?

mreid-tt:
> Can you provide the exact working vs non working command lines you're using? Which is which?

So essentially, if you build the `tt-rss` package in your local environment with this:
```
root@ghcr-io-synocommunity-spksrc-1:/spksrc/spk/tt-rss# make arch-noarch-6.1
```
During the build there is this section:
```
===>  Extracting for tt-rss                                                      
tar -xzpf /spksrc/cross/tt-rss/../../distrib/tt-rss-gitafd04d141c7bd8a1d6732a2a6a
4c91c0a9b4cd6b.tar.gz -C /spksrc/spk/tt-rss/work-noarch-6.1                      
===>  Patching for tt-rss                                                        
===>  Configuring for tt-rss                                                     
```
And that is supposed to apply the [patch](https://github.com/SynoCommunity/spksrc/blob/master/cross/tt-rss/patches/001-Add_possibility_to_customize_socket_used_to_connect_to_MySQL.patch) to the config files.

However when you extract the spk and then the `package.tgz` inside you don't see the patch applied. This is a search of both the published package from GitHub (`tt-rss.v19.f15047[noarch]`), and the locally built one (`tt-rss_noarch-dsm6_20230828-19`):
```
mreid@iMac Desktop % find . -name "config.php" -exec grep -l "MYSQL_CHARSET" {} \;
./tt-rss_noarch-dsm6_20230828-19/package/share/tt-rss/classes/config.php
./tt-rss.v19.f15047[noarch]/package/share/tt-rss/classes/config.php
```
As you can see, searching for a string we know should be in the file finds both config files. However when we search for the string that should be there from the patch we only see the one from GitHub containing it:
```
mreid@iMac Desktop % find . -name "config.php" -exec grep -l "MYSQL_DB_SOCKET" {} \;
./tt-rss.v19.f15047[noarch]/package/share/tt-rss/classes/config.php
```
The equivalent action in GitHub shows the following:
```
  ===>  Extracting for tt-rss
  tar -xzpf /github/workspace/cross/tt-rss/../../distrib/tt-rss-gitafd04d141c7bd8a1d6732a2a6a4c91c0a9b4cd6b.tar.gz -C /github/workspace/spk/tt-rss/work-noarch-6.1  
  ===>  Patching for tt-rss
  patch -p1 < patches/001-Add_possibility_to_customize_socket_used_to_connect_to_MySQL.patch
  patching file classes/db.php
  Hunk #1 succeeded at 37 (offset 2 lines).
  patching file classes/config.php
  Hunk #2 succeeded at 210 (offset 25 lines).
  ===>  Configuring for tt-rss
```
Based on this, it appears the local environment isn't actually picking up the patch at all. I may have to refactor this PR since the original premise was that it was patching but that it was being overridden.

mreid-tt:
> I may have to refactor this PR since the original premise was that it was patching but that it was being overridden.

Digging deeper into this, I found a much simpler solution by tweaking the "mk/spksrc.patch.mk" file instead. Will close this and make a new PR.

EDIT: PR #6829

PR #6829, opened 2025-12-06
framework: normalize ARCH when resolving patch directories

mreid-tt:
## Description

- strip the toolchain suffix from ARCH before selecting patch directories so builds such as x64-7.0 and noarch-7.0 reuse the expected patch buckets
- keeps the original lookup order and generic fallback intact; no other framework changes required

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix

### DIFF
--- a/mk/spksrc.patch.mk
+++ b/mk/spksrc.patch.mk
@@ -25,15 +25,24 @@ endif
 #    patches/$(arch)-$(TCVERSION)/*.patch
 # supported groups: arm, armv5, armv7, armv7l, armv8, ppc, i686, x64
 ifeq ($(strip $(PATCHES)),)
-ifeq ($(strip $(ARCH)),)
+PATCH_ARCH = $(strip $(ARCH))
+ifneq ($(PATCH_ARCH),)
+PATCH_ARCH := $(firstword $(subst -, ,$(PATCH_ARCH)))
+endif
+
+ifeq ($(PATCH_ARCH),)
 PATCHES = $(wildcard patches/*.patch)
 else ifneq ($(filter cross diyspk python spk,$(shell basename $(dir $(abspath $(dir $$PWD))))),)
 PATCHES = $(sort $(foreach group,ARM_ARCHS ARMv5_ARCHS ARMv7_ARCHS ARMv7L_ARCHS ARMv8_ARCHS PPC_ARCHS i686_ARCHS x64_ARCHS, \
 	$(foreach arch,$($(group)), \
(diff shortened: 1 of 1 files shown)


mk/spksrc.patch.mk
```diff
@@ -25,15 +25,24 @@ endif
 #    patches/$(arch)-$(TCVERSION)/*.patch
 # supported groups: arm, armv5, armv7, armv7l, armv8, ppc, i686, x64
 ifeq ($(strip $(PATCHES)),)
-ifeq ($(strip $(ARCH)),)
+PATCH_ARCH = $(strip $(ARCH))
+ifneq ($(PATCH_ARCH),)
+PATCH_ARCH := $(firstword $(subst -, ,$(PATCH_ARCH)))
+endif
```
th0ma7: Why not just doing:
```
ifeq ($(strip $(PATCHES)),)
PATCH_ARCH := $(firstword $(subst -, ,$(strip $(or $(PATCH_ARCH),$(ARCH))))
```

mk/spksrc.patch.mk
th0ma7, line 19: @mreid-tt what are you trying to fix exacly?  I do not understand the workflow you're trying to address.

This comment tells that by default `noarch` is currently meant to only be using the default `patches/*.patch`.  Are you trying to add sub-directory management for noarch based on DSM versions?  If so you can use `patches/DSM-$(TCVERSION)/*.patch` which should normally works for noarch use-case.

It also relates to how is your noarch build initiated, revisiting the github-action call it should normally be using something such as: `make TCVERSION=${TCVERSION} ARCH= -C ./spk/${package}`

mreid-tt: I am trying to fix this scenario... https://github.com/SynoCommunity/spksrc/pull/6826#issuecomment-3618900308

th0ma7: ok, I'll have another look at it.  I suppose that using `make TCVERSION=${TCVERSION} ARCH= -C ./spk/${package}` works, but make arch-noarch-x.y doesn't, which is a different problem on its own.

th0ma7: Found where the issue is, it's in spksrc.supported.mk where it never get into the `build-noarch-%` recipe.  This is seen by the lack of printing: `@$(MSG) BUILDING noarch package for TCVERSION $*` ... 

So fix needed is elsewhere, I'll fix this later this PM.

mreid-tt:
Superseded by #6830.

mk/spksrc.patch.mk
mreid-tt: @th0ma7, I don't know if it would be worth reexamining this earlier approach instead.

PR #6684, opened 2025-08-18
framework: fix arch-suffix for noarch builds

hgy59:
## Description

- fix work folders for noarch packages
- define ARCH_SUFFIX before include of spksrc.directories.mk for noarch packages too

follow up to #6627, similar to #6673

Without this fix, builds for noarch packages WORK_DIR = work instead of work-noarch-$(TCVERSION) and builds were lacking build*.log files.
This behaviour fix fixed for builds in cross and spk folders.

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [x] Includes small framework changes

PR #6830, opened 2025-12-06
Fix call to make arch-noarch-x.y

th0ma7:
## Description

Fix call to make `arch-noarch-<tcversion>` as currently is enters the regular `arch-<arch>-<tcversion>`.

Relates to : #6826 #6829 and in particular https://github.com/SynoCommunity/spksrc/pull/6826#issuecomment-3618900308

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.supported.mk
+++ b/mk/spksrc.supported.mk
@@ -78,7 +78,7 @@ build-arch-%:
 build-noarch-%: SHELL:=/bin/bash
 build-noarch-%: 
 	@$(MSG) BUILDING noarch package for TCVERSION $*
-	@MAKEFLAGS= $(MAKE) TCVERSION=$* ARCH= 2>&1 ; \
+	@MAKEFLAGS= $(MAKE) TCVERSION=$* ARCH=noarch 2>&1 ; \
 	status=$${PIPESTATUS[0]} ; \
 	$(MSG) $$(printf "%s MAKELEVEL: %02d, PARALLEL_MAKE: %s, TCVERSION: %s, NAME: %s [END]\n" "$$(date +%Y%m%d-%H%M%S)" $(MAKELEVEL) "$(PARALLEL_MAKE)" "$*" "$(NAME)") | tee --append $(STATUS_LOG) ; \
 	[ $${status[0]} -eq 0 ] || false


th0ma7:
@mreid-tt Can you confirm this does fix the issue for you?

mreid-tt:
> @mreid-tt Can you confirm this does fix the issue for you?

Yes, for the specific use-case cited it works!

hgy59:
@th0ma7 I can confirm that the patches are now applied.

BUT the build log file is now lacking the arch and tcversion in the name.

The log file is called build.log instead of build-noarch-{TCVERSION}.log


EDIT:
with `make ARCH=noarch TCVERSION={TCVERSION}` the build log is correct, but the patches are not applied.

with `make arch-noarch-{TCVERSION}`  the build log is `build.log` and the patches are applied.

hgy59:
This might be a side effect of #6250 and/or  #6684 (or a PR related to those).

There shouldn't exist any target for noarch in spksrc.supported.mk.
_supported_ in this context means archs supported (and noarch isn't an arch).

When building noarch packages, the `build-noarch-%` target in spksrc.supported.mk is not used (neigher with `make arch-noarch-{TCVERSION}` nor with `make arch=noarch TCVERSION={TCVERSION}`)

th0ma7:
> BUT the build log file is now lacking the arch and tcversion in the name.
>
> The log file is called build.log instead of build-noarch-{TCVERSION}.log

This is now fixed
 
> EDIT: with `make ARCH=noarch TCVERSION={TCVERSION}` the build log is correct, but the patches are not applied.

That should never be used, unsure if I could ever workaround that.  A noarch must either use `make arch-noarch-<tcversion>` or `make TCVERSION=<tcversion> ARCH=` (ARCH must be empty or unset).

th0ma7:
> There shouldn't exist any target for noarch in spksrc.supported.mk. _supported_ in this context means archs supported (and noarch isn't an arch).

It always existed, but clearly when calling `make all-supported` this should just make a noarch 6 and 7 packages.  It is not doing that currently.  Although this could be fixed through a different PR as the immediate problem is now solved.

hgy59:
> That should never be used, unsure if I could ever workaround that. A noarch must either use `make arch-noarch-<tcversion>` or `make TCVERSION=<tcversion> ARCH=` (ARCH must be empty or unset).

The redesign in #6250 changed in that `ARCH=noarch` is now used instead of `ARCH=`.

th0ma7:
@hgy59 and @mreid-tt there are two use cases still not working:
1. when calling `make ARCH=aarch64|x64|etc TCVERSION=6.2|7.1|7.2` it downloads the toolchain while unecessary
2. when calling `make ARCH=noarch TCVERSION=6.2|7.1|7.2` is doesn't apply patches - although works if using `ARCH=`

th0ma7:
Ok, item 2 (`ARCH=noarch` patching issue) now fixed, only remaining 1 which is sort of estetic more than anything else.

th0ma7:
@hgy59 and @mreid-tt mind giving this a try?

mreid-tt:
> @hgy59 and @mreid-tt mind giving this a try?

Still works from my use case so I'm happy.

mreid-tt:
@th0ma7 @hgy59, I don't know if it's this PR or an earlier one but current builds are failing. See https://github.com/mreid-tt/spksrc/actions/runs/20023803740

th0ma7:
We absolutely need to find where the issue comes from .. indeed it may well be from here.  To be confirmed.  I'll check this out later tonight using a dummy pr and see...

mreid-tt:
@th0ma7 this does seem patch related since in the logs I see this:

```
2025-12-08T10:12:30.4834204Z ===>  Extracting for libtorrent
2025-12-08T10:12:30.4856251Z tar -xzpf /github/workspace/cross/libtorrent-rakshasa/../../distrib/libtorrent-0.16.4.tar.gz -C /github/workspace/spk/rutorrent/work-x64-7.2  
2025-12-08T10:12:30.5209695Z ===>  Patching for libtorrent
2025-12-08T10:12:30.5262133Z ===>  Configuring for libtorrent
```

whereas in previous builds I saw this:

```
2025-12-01T03:28:59.9680699Z ===>  Extracting for libtorrent
2025-12-01T03:28:59.9700407Z tar -xzpf /github/workspace/cross/libtorrent-rakshasa/../../distrib/libtorrent-0.16.4.tar.gz -C /github/workspace/spk/rutorrent/work-x64-7.2  
2025-12-01T03:29:00.0043182Z ===>  Patching for libtorrent
2025-12-01T03:29:00.0092195Z patch -p0 < patches/001-fix-cross-compile.patch
2025-12-01T03:29:00.0137262Z patching file configure
2025-12-01T03:29:00.0146982Z Hunk #1 succeeded at 18980 (offset 605 lines).
2025-12-01T03:29:00.0204023Z ===>  Configuring for libtorrent
```

mreid-tt:
@th0ma7, I've confirmed in my test environment that if I revert this PR, the build succeeds. Seems like while it fixes calls like `make arch-noarch-6.1`, it breaks `make arch-x64-7.2`.

hgy59:
my latest run confirms, that the patches are not applied anymore
https://github.com/SynoCommunity/spksrc/actions/runs/20027803338

th0ma7:
What I find odd is that I started 3 different builds, none of which have this fix applied and they all failed:
https://github.com/SynoCommunity/spksrc/pull/6836
https://github.com/SynoCommunity/spksrc/pull/6837
https://github.com/SynoCommunity/spksrc/pull/6835

There clearly is something I don't understand ... and/or I need way more ☕ ☕ ☕

hgy59:
#6835 failed because cross/bzip2 patch was not applied (x64-7.1 succeeded, but python312 failed for all others due missing bzip2)

BTW each build of a PR uses a merged version (i.e. the master branch is applied before the build).
And a PR can't be built nor merged when conflicts against the master exist.

So the "fix" was applied in your three builds.

hgy59:
The builds in the upstream repo (here) might be different than the builds in your fork...

th0ma7:
> BTW each build of a PR uses a merged version (i.e. the master branch is applied before the build). And a PR can't be built nor merged when conflicts against the master exist.
> 
> So the "fix" was applied in your three builds.

OMG! this explains a lot!  Then, if this PR is the issue it's probably trivial to fix... I faced similar issues with prioritization of recipes in the early-stages of the testing portion of the framework.  I can have a look at it tonight.

mreid-tt:
> Then, if this PR is the issue it's probably trivial to fix... I faced similar issues with prioritization of recipes in the early-stages of the testing portion of the framework. I can have a look at it tonight.

Thanks. If too troublesome, we can revert this PR and re-engage at a future date.

th0ma7:
Issue should now be fixed, thxn for you patience :shrug:

PR #6832, opened 2025-12-07
Add pelican_panel package

Kalletal:
## Summary

Add **Pelican Panel** package - a free, open-source game server management panel built with PHP, React, and Go.

### Features
- Pelican Panel web interface (Docker container)
- Wings daemon for server management (Docker container)
- DSM integration with "Open" button and Wings configuration UI
- Installation wizard for easy setup on Synology NAS
- Automatic Docker-in-Docker fixes for Synology compatibility

### Requirements
- DSM 7.0+
- Container Manager (Docker)
- x86_64 architecture

### Package Details
- **Name**: pelican_panel
- **Version**: 1.0.0
- **Architecture**: noarch (Docker-based)
- **Homepage**: https://pelican.dev
- **License**: MIT

### Screenshots / Testing
Tested on DS920+ with DSM 7.2

## Checklist
- [x] Package follows spksrc conventions
- [x] Makefile properly configured
- [x] Service scripts (service-setup.sh, dsm-control.sh)
- [x] Installation wizard
- [x] DSM UI integration (Open button, Wings config)
- [x] Documentation in package


hgy59:
Some remarks

In this repository (spksrc) we mainly provide packages with cross compiled code.
There are some noarch packages that do not need compilation and contain arch independent code (mainly php, but also html, perl, etc..)
If we ever provide packages like this, to run with synology docker-manager, we should create a new repository, because the installation, configuration and package update have less in common.

IMHO it is not worth to create packages for docker-manager.
The docker way is to create a docker-compose file for the deployment.
With this you have full control over the ports (port mappings) the volumes, environments, users, etc.
If you want to have a better deployment, you could use portainer to manage your containers.
But a spk is not the right solution.

Kalletal:
Personally, I find it simpler to use a package manager (SPK) than the container manager. So if it's simpler for me, I think it can be for others too.
Everything happens in the package manager; I install the SPK, it's seamless, everything runs in the background, and it works.

PR #6833, opened 2025-12-07
spk/php84: add PHP 8.4.15 with Extension Manager

Kalletal:
## Summary

- Add PHP 8.4.15 package for Synology NAS (DSM 7.2+)
- Extension Manager with ExtJS UI for graphical extension management
- 163 PHP extensions compiled for x86_64 architecture
- Installation wizard with profile selection (Minimal/Standard/Complete)

## Features

- **Extension Manager UI**: Native DSM application to enable/disable PHP extensions
- **Profile-based installation**: Choose from Minimal (7 extensions), Standard (11 extensions), or Complete (163 extensions)
- **Extension dependencies**: Automatic load order management
- **PHP-FPM service**: Unix socket configuration with dynamic process management

## Architecture

- Generic `x86_64` package compatible with all Intel/AMD Synology NAS
- Pre-compiled binaries from cross/php84 (geminilake toolchain)

## Testing

- Tested on DSM 7.3.1 (geminilake architecture)
- Service starts correctly
- Extension Manager UI accessible via "Open" button
- Extensions can be enabled/disabled via UI

## Screenshots

Extension Manager allows toggling extensions and viewing PHP info.

---


hgy59:
@mreid-tt This is an interesting extension (if it works with WebStation/Apache).

hgy59:
@Kalletal please rework this PR (and do not use AI generated code!)
- use cross/php84 to cross compile binary code
- avoid version updates by scripts (this is always a manual task)
- use english as default language (wizards, conf) you can optionally add other languages for package description and wizards
- remove conf/privilege, this file is generated by the spksrc framework (and the package user will be sc-php84, not php84)
- remove conf/resources, this file is genearted by the spksrc framework, by definition of SPK_COMMANDS
- remove images/php84_*.png, those are created by the framework (but the name is php84-*.png - with dash, not underscore)
- remove scripts/*  (legacy/* contradicts with the spksrc framework, other scripts go to service-setup.sh
- and service-setup.sh needs a lot of rework...

Please try to cross compile the binaries.
Packages that support only x64 tend to be candidates for manual installation with docker (container-manager).

Kalletal:
No, unfortunately it doesn't work via Web Station. It's a bit complicated to set up, but it works very well this way. And it's PHP 8.4.15.

I am French by origin and I would like to be able to distribute packages in French by default.

Initially I had made the package for my geminilake architecture but I was advised to make it generic x84_64 so that it would be pushed to your git.

Everything is done with Claude Code. The whole package. If it's a problem, it's not a big deal; I'll keep my packages to myself.

But I can try to rework the points you raised. If that allows for integration with the Synocommunity and makes it shareable with others.

What should be changed in service-setup.sh?

th0ma7:
@Kalletal packages allows providing description in different languages, including french like I do myself for packages such as tvheadend using `DESCRIPTION_FRE`.  You will find here the list of fields that allows using multiple languages and can be set in your spk makefile https://help.synology.com/developer-guide/synology_package/INFO_optional_fields.html

@mreid-tt this is the initiative I was referring you to with regards of supporting a synocommunity based php instead of depending of Synology's version that dates or is no longer supported on older archs.  Maybe we could package a cross/php with said applications that requires it or better, find a way to install it side-by-side with synology's default?

mreid-tt:
> @mreid-tt this is the initiative I was referring you to with regards of supporting a synocommunity based php instead of depending of Synology's version that dates or is no longer supported on older archs. Maybe we could package a cross/php with said applications that requires it or better, find a way to install it side-by-side with synology's default?

Is this now superseded by https://www.synology.com/en-global/releaseNote/PHP8.4?

th0ma7:
@mreid-tt I guess to some degree... I was mostly thinking of providing our own PHP so that packages from DSM6-7 can all benefit from using the same version, thus simplifying its support.  There could be a generic php84 to rule them all.  Unclear how this would work out now ... but could still be applicable for DSM6, 7 and 7.1 as if I get this right php84 will only be for dsm-7.2 and above?

mreid-tt:
> but could still be applicable for DSM6, 7 and 7.1 as if I get this right php84 will only be for dsm-7.2 and above?

Not from what I see here:

https://github.com/SynoCommunity/spksrc/blob/52d8344e7c394d3db52ff4b658a19bc540b73296/spk/php84/Makefile#L15-L16

th0ma7:
I believe this is a WIP, to confirm whether or not it compiles using an older gcc...

hgy59:
No need for this, since Synology provides PHP 8.4 (for DSM >= 7.2).
We don't want to have packages depending on a community php 8.4 package, due to the lack of integration into DSM webstation/apache/nginx.

PR #6834, opened 2025-12-08
Nextcloud: Download application archive at install time

mreid-tt:
## Description

- Drop the `cross/nextcloud` package and ship only metadata plus licensing in the SPK.
- Extend `service-setup.sh` to fetch, verify, and unpack the Nextcloud release during install/upgrade, reusing the staged copy for restores.
- Add installer/upgrade wizard notices explaining the large download so users expect a longer runtime.

Follow-on from #6814

## Checklist

### Type of change

- [x] Package refresh


hgy59:
@mreid-tt great, you found a way for this!
I missed to give a hint, that wie have a similar approach in nzbget.

I still would avoid the use of the LICENSE_FILE, especially for a package with a UI that (I assume) has an about or other dialog showing the license.

mreid-tt:
> I still would avoid the use of the LICENSE_FILE, especially for a package with a UI that (I assume) has an about or other dialog showing the license.

Noted but I've found no clear way to access this in the UI so I'd like to leave it in.

Open for reviews but I'll be merging later today unless there are any other comments.

PR #6845, opened 2025-12-11
Borg: bump to v1.4.3

mreid-tt:
## Description

This PR includes the following:

1. Update borg to v1.4.3
2. Update borgmatic to v2.0.12

Fixes #6788

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #6848, opened 2025-12-12
borgbackup: update of python311 compatible package

hgy59:
## Description

Follow up to #6845

Fixes # 

## Checklist

### Type of change

- [ ] Package update

### DIFF
--- a/spk/borgbackup/Makefile
+++ b/spk/borgbackup/Makefile
@@ -3,6 +3,12 @@ SPK_VERS = 1.4.3
 SPK_REV = 20
 SPK_ICON = src/borgbackup.png
 
+# Remarks:
+# keep spk/borgbackup_py311 in sync.
+# update and validate borgbackup for python311 too.
+# publish python312 version first, then publish py311 version for DSM 6,
+# for priority of py312 over py311 for archs that support both.
+
 PYTHON_PACKAGE = python312
 SPK_DEPENDS = "$(PYTHON_PACKAGE)"
 UNSUPPORTED_ARCHS = $(OLD_PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS)
--- a/spk/borgbackup_py311/Makefile
+++ b/spk/borgbackup_py311/Makefile
@@ -1,11 +1,10 @@
 SPK_NAME = borgbackup
-SPK_VERS = 1.4.1
-# reuse former SPK_REV to separate from python312 dependent packages
-SPK_REV = 17
+SPK_VERS = 1.4.3
+SPK_REV = 21
 SPK_ICON = src/borgbackup.png
 
 # REMARKS
-# Python 3.11 compatible version for ARMv5 archs
+# Python 3.11 compatible version for ARMv5 and ARMv7L archs
 
 PYTHON_PACKAGE = python311
 SPK_DEPENDS = $(PYTHON_PACKAGE)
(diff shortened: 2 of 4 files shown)


hgy59:
@th0ma7, @mreid-tt 
When building both borgbackup and borgbackup_py311 the python312 version is built first. When both have the same SPK_REV, the py311 version will overwirte the py312 artifacts, because the spk files have identical names.

We could rename spk/borgbackup to spk/borgbackup_py312 to build borgbackup_py311 first and have the 312 version overwriting the 311 artifact.
For publishing this is not an issue (when we publish for py312 first),

Or we could use different SPK_REV for py312 and py311.
To test py311 versions for other archs than ARMv5 and ARMv7L, we need a local build or a manual github build if the SPK_REV are the same.

What shall we do?

mreid-tt:
@hgy59, my view on this hasn’t changed. Maintaining two separate `spk/*` paths that produce the same package isn’t a sustainable best practice.

A cleaner approach would be similar to the `cross/*` packages, using a routing mechanism that selects the appropriate build based on the target architecture, with no overlap. Alternatively, if overlapping architecture combinations are truly required, we should produce a distinct package that explicitly declares a different Python dependency.

hgy59:
> @hgy59, my view on this hasn’t changed. Maintaining two separate `spk/*` paths that produce the same package isn’t a sustainable best practice.
> 
> A cleaner approach would be similar to the `cross/*` packages, using a routing mechanism that selects the appropriate build based on the target architecture, with no overlap. Alternatively, if overlapping architecture combinations are truly required, we should produce a distinct package that explicitly declares a different Python dependency.

And my argument for separate packages is that we need a python 311 version for all archs for testing.
Or do you have an ARMv5 or ARMv7L model?

mreid-tt:
> And my argument for separate packages is that we need a python 311 version for all archs for testing.
> Or do you have an ARMv5 or ARMv7L model?

Was this a regular practice? I typically only test for one arch on multiple DSM versions. I don't have any ARM based models so testing is limited to x64 for me.

th0ma7:
> > And my argument for separate packages is that we need a python 311 version for all archs for testing.
> > Or do you have an ARMv5 or ARMv7L model?

@hgy59 i do have an armv5 and an armv7 (not 7L)  for testing in my basement - actually i wonder if i could not make it available for testing to all when needed (maybe a christmas pet project).

> Was this a regular practice? I typically only test for one arch on multiple DSM versions. I don't have any ARM based models so testing is limited to x64 for me.

@mreid-tt no.  but previously we where not facing this issue neither ... and unsupported archs will just grow over time.  I mean, let's face it, gcc<=4.x is a little off by now.

Looking at the python support chart (ai-generated):
| Version | Release | End of Life | Time Remaining |
|---------|---------|-------------|----------------|
| Python 3.9 | Oct 2020 | Oct 2025 | ❌ **Expired** |
| Python 3.10 | Oct 2021 | Oct 2026 | ⚠️ **~10 months** |
| Python 3.11 | Oct 2022 | Oct 2027 | ✅ **~2 years 10 months** |
| Python 3.12 | Oct 2023 | Oct 2028 | ✅ **~3 years 10 months** |
| Python 3.13 | Oct 2024 | Oct 2029 | ✅ **~4 years 10 months** |
| Python 3.14 | Oct 2025 | Oct 2030 | ✅ **~5 years 10 months** |

Considering python 3.11 is still supported for almost 3-years it may be worthwhile to do something about it... I mean we could have `PYTHON_PACKAGE = python311:DSB6,comcerto2k,hi3535 python312:default` (or what ever) which would allow building using the right python, for the right arch, and the right DSM... and resulting packages could even have by default the `_py311|_py312` suffix auto-added or similar.

It might even not be that difficult considering all is now within a one-stop shop under `spksrc.python.mk` ...

th0ma7:
Could be as simple as changing this in `spksrc.python.mk` (totally untested):
```
# armv5 and hi3535 no longer supported with python >= 3.12
ifeq ($(call version_ge, $(subst python,,$(PYTHON_PACKAGE)), 312), 1)
ifeq ($(findstring $(ARCH), $(ARMv5_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
PYTHON_PACKAGE = python311
endif
```

mreid-tt:
@th0ma7, I don't think it needs to be that complex. Couldn't we just adjust it in the Makefile as I suggested before in https://github.com/SynoCommunity/spksrc/pull/6845#issuecomment-3647998658?

th0ma7:
Actually i believe this may be the opposite and real easy with a slight change to python.mk

mreid-tt:
hey @hgy59, I re-combined Borg into a single package that builds for both Python 3.12 and Python 3.11 in #6936. As such, I'll close off this PR.

PR #6865, opened 2025-12-20
Syncthing: Update to v2.0.12

### LINKED ISSUE #6721: Syncthing ver 2.x for DS214+ MARVELL Armada XP MV78230
### Software Name

syncthing

### Brief Description

Please add syncthing version 2.x on the Synology Community packages for DS214+ armv7l arch... ver 1.3 is essentially deprecated ..

### Website

https://synocommunity.com/packages

### Documentation

https://synocommunity.com/

https://docs.syncthing.net/dev/building.html

### Source code

https://github.com/syncthing/syncthing

### OpenSource License

MIT

### Other License

_No response_

mreid-tt:
## Description

1. Update Syncthing to v2.0.12 (built with Go 1.25).
2. Drop legacy ARMv5 models that required the Go 1.23 compiler.

Fixes #6721

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/spk/syncthing/src/service-setup.sh
+++ b/spk/syncthing/src/service-setup.sh
@@ -29,33 +29,6 @@ service_postinst() {
     if [ "${SYNOPKG_PKG_STATUS}" == "INSTALL" ]; then
         set_credentials
     fi
-    
-    if [ "${SYNOPKG_DSM_ARCH}" == "88f628x" ]; then
-        # disable auto update for AMRv5 archs
-        comment="disable auto update for ${SYNOPKG_DSM_ARCH} arch"
-        if [ -r ${SYNCTHING_OPTIONS_FILE} ]; then
-            if grep -q "^SYNCTHING_OPTIONS=" "${SYNCTHING_OPTIONS_FILE}"; then
-                options=$(grep '^SYNCTHING_OPTIONS=' ${SYNCTHING_OPTIONS_FILE} | cut -d= -f2 | tr -d '"')
-                if [[ "${options}" == *"--no-upgrade"* ]]; then
-                    echo "- 'auto update' option is already disabled for ${SYNOPKG_DSM_ARCH} arch"
-                else
-                    echo "- Patch ${SYNCTHING_OPTIONS_FILE} to ${comment}"
-                    sed "s/^SYNCTHING_OPTIONS=.*/SYNCTHING_OPTIONS=\"${options} --no-upgrade\"/" -i ${SYNCTHING_OPTIONS_FILE}
-                fi
-            elif grep -q "#SYNCTHING_OPTIONS=" "${SYNCTHING_OPTIONS_FILE}"; then
-                # commented SYNCTHING_OPTIONS defined, append line to disable auto update
-                echo "- Patch ${SYNCTHING_OPTIONS_FILE} to add option to ${comment}"
-                sed '/#SYNCTHING_OPTIONS=.*/a SYNCTHING_OPTIONS="--no-upgrade"' -i ${SYNCTHING_OPTIONS_FILE}
-            else
-                echo "- Add option in ${SYNCTHING_OPTIONS_FILE} to ${comment}"
-                echo "SYNCTHING_OPTIONS=\"--no-upgrade\"" >>  ${SYNCTHING_OPTIONS_FILE}
-            fi
-        else
-            # create options file (should not occur, since such a file was just installed)
-            echo "- Create ${SYNCTHING_OPTIONS_FILE} with option to ${comment}"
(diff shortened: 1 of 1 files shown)


hgy59:
@mreid-tt please let @acolomb join in.

see https://github.com/SynoCommunity/spksrc/issues/6721#issuecomment-3453095770

AFAICR a package update has to ensure that syncthing was autoupdated to 2.x before the update.

hgy59:
@mreid-tt there is no need for such "staging sqlite"
only depend on cross/sqlite
and use
```
	  CGO_CFLAGS="-I$(STAGING_INSTALL_PREFIX)/include" \
	  CGO_LDFLAGS="-L$(STAGING_INSTALL_PREFIX)/lib -Wl,-rpath,\$$ORIGIN/../lib -Wl,--rpath-link,$(STAGING_INSTALL_PREFIX)/lib -lsqlite3 -lz" \
```
No need to install sqlite libraries, those are already installed into workdir/install/...

The real issue here is, that cgo (the go c compiler) defines an argument `-m64` that is invalid for all cross compilers (except x64).

_I faced a similar issue in cross/helix, but this was with rust and I had to patch the compiler crate to fix this. I have no idea how the cgo works and didn't find a way so far to make the build verbose to find the cause..._

hgy59:
@mreid-tt we can't build syncthing with dependency to libsqlite.so

The syncthing build does not support builds with CGO_ENABLED (despite the dev docs say so).
_I tried several approaches like build build.go for host and run it to build the target, patching build.go, ..._

BUT even if we could cross compile, it is worth nothing, because of the autoupdate.
The packages installed by autoupdate do not depend on libsqlite either.

BTW we need to create syncthing v2.0.11 to test the autoupdate at first installation (and dont't forget to test a package update with an installation that already is auto updated to v2.0.12 to verify that it does not downgrade).

@acolomb is there really an sqlite performance issue with the official syncthing binaries?

acolomb:
> @acolomb is there really an sqlite performance issue with the official syncthing binaries?

Not that I know of, quite the contrary.  The official builds have been configured to use the best-performing variant.  I've read reports about some Docker images which were produced and distributed by third parties, where the performance was less than optimal.  IIRC they ended up with the "modernc" build, see e.g. [here](https://forum.syncthing.net/t/syncthing-db-migration-errors/25122/6?u=acolomb).

I don't see any need to make sure the installed version has previously auto-updated to v2 before installing the new package.  What we do need to avoid is "upgrading" to an SPK containing an older version than the one already obtained through auto-upgrades.  But that is already handled for a few years, by simply not replacing the binary if it's already newer.

We should check again how that version check works though, as some early v2.0.x versions didn't accept `--version`, see https://github.com/syncthing/syncthing/pull/10269.

I think we're fine building the way we did so far.  As a "trust anchor", the SPK contains a version built from verified source.  That one should use the cross-compiled libsqlite, which probably won't require installation on the system though, as the Go binaries are statically linked IIRC.  The internal upgrader then replaces that trusted binary with a newer one, while keeping the trust chain intact through its internal signature verification.  The trust basically shifts from Syncthing's source code + SynoCommunity's build scripts to solely the digitally signed binaries from the Syncthing Foundation.

What part in that scenario are you having build troubles with?

hgy59:
> I don't see any need to make sure the installed version has previously auto-updated to v2 before installing the new package.

This is to avoid an update from 1.x to 2.x version because there is no real autoupdate. The user needs to manually enable the "internal update" from 1.x to 2.x and is encouraged to do a database backup before, because the db is irreversible mirgrated.
The idea is to avoid the update from 1.x to 2.x by the spk installer and to force the user to trigger the update manually before.

sure we could investigate into the package update and create a backup in this case, but this would need a lot of documentation (like how to restore the 1.x version and where to find the backup) and testing, that I would like to avoid, for the rare case that a user didn't already update to 2.x.

acolomb:
> The idea is to avoid the update from 1.x to 2.x by the spk installer and to force the user to trigger the update manually before.

I don't see that need. Not sure about DSM itself, but can it be set to automatically install SPK updates unattended?

If you want to follow the (sensible) upstream decision of not surprising users with the v2 upgrade, and assuming that the answer above is yes, why not simply add a step to the install wizard which explicitly asks for consent and otherwise refrain from overwriting the installed binary?

Adding a separate v2 SPK is probably out of the question, as it cannot share config files with the old package. 

During the upgrade, a complete backup of the old DB is kept for recovery by simply renaming its directory after migration. So the recovery path is quite simple:

- Stop Syncthing v2. 
- Rename the backup to its original name.
- Install the older v1 SPK.

I suppose the data is kept during a package downgrade as well? The time window is limited though, as the backup gets deleted by v2 after some time (not sure from the top of my head).

mreid-tt:
Superseded by #6926

PR #6884, opened 2026-01-06
Add Playerr package structure

Maikboarder:
### Package details
Adds **Playerr**, a self-hosted game library manager (similar to Sonarr/Radarr but for games).

- **Source:** https://github.com/Maikboarder/Playerr
- **License:** MIT

### Implementation
This is a standard Docker wrapper package. It relies on the generic Docker installation method.

### Note to Maintainers
I am the developer of Playerr. I do not currently own Synology hardware, so this package is submitted based on the standard Docker template structure. 

I have verified the file structure and syntax, but I request a maintainer or community member with a DSM 7 device to verify the installation and startup process.


mreid-tt:
This is a novel approach (having a package that only automates a docker deployment). I am not sure our framework supports this. Are there any more typical ways to deploy this other than with Docker?

GrandDuke1106:
> I am the developer of Playerr. I do not currently own Synology hardware, so this package is submitted based on the standard Docker template structure.

If you don't have the hardware, you can use [rr](https://github.com/RROrg/rr) to launch a DSM in a virtual machine to test the software package.

Maikboarder:
@mreid-tt Thank you for the clarification. I understand that the Docker wrapper approach isn't suitable and that a native spksrc integration is the standard here.

@GrandDuke1106 Thanks for the tip on rr! That will be incredibly useful for verifying the native package locally without dedicated hardware.

Regarding the next steps: I am currently working hard on the core features and making significant progress with every release. However, given that Playerr is still in beta, I believe it would be best to hold off on this submission until I can deliver a more mature, feature-complete version to the SynoCommunity.

I will close this PR for now to avoid cluttering your queue. I will return with a proper native package structure once Playerr is closer to a stable v1.0 release and I have fully tested the native build process on a virtualized DSM.

Thank you both for your guidance!

PR #6890, opened 2026-01-07
garage: initial package

fgma:
## Description

Provide a supported alternative to the now unsupported minio S3 storage provider. For details see [Garage](https://garagehq.deuxfleurs.fr/).

## Checklist

### Type of change

- [x] New Package


mreid-tt:
Will take a more detailed review later today but in the meantime I notice some of your files don't have a trailing carriage return. Just a minor issue to correct.

Glad to see more rust packages! While working on some others I found that builds often don't target hi3535 (ARMv7L_ARCHS); is your build actually compatible with this arch?

EDIT: Also the `spk/garage/src/garage.sc` could add a blank line between each block for readability.

spk/garage/Makefile
```diff
@@ -0,0 +1,29 @@
...
+SPK_ICON = src/garage.png
+DSM_UI_DIR = app
+
+DEPENDS = cross/garage
+
+UNSUPPORTED_ARCHS = $(ARMv5_ARCHS) $(PPC_ARCHS)
+
+MAINTAINER = fgma
+DESCRIPTION = Garage is an S3-compatible distributed object storage service designed for self-hosting at a small-to-medium scale.
+DISPLAY_NAME = garage
```
mreid-tt:
```suggestion
DISPLAY_NAME = Garage
```

fgma:
I've added the trailing newlines and cleaned up garage.sc. I'm not sure about ARMv7L_ARCHS because I can't test against it myself.

mreid-tt:
hey @fgma, thanks for the changes. Apologies if my comment was confusing. I meant it should have a newline character at the end of the last line with content. Not that there should be an extra blank line. Anyway, I've cleaned them up for you.

As for the `ARMv7L_ARCHS`, I don't know if one of the other devs can help with testing this. I'll do some more research in the meantime.

EDIT: Do you have any direct way of testing the actual function of the app itself?

mreid-tt:
hey @fgma, one other design pattern I noted was the generation of the config file from the `service-setup.sh` script. Ideally it would preferred if you have a template file for the `garage.toml` in the `src/` folder which is installed via the `Makefile`. This can then be used to substitute any specific values if needed from the `service-setup.sh`. This makes the scripts much easier to review and maintain.

mreid-tt:
hey @fgma, I've installed the test builds on DSM 6 and DSM 7 and it generally installs clean. On DSM 6 we have these warnings in the install log:
```
2026/01/20 10:47:00	Creating default Garage config at /volume1/garage/garage/garage.toml
2026/01/20 10:47:00	unable to write 'random state'
2026/01/20 10:47:01	unable to write 'random state'
2026/01/20 10:47:01	unable to write 'random state'
```

These however seem to be cosmetic since the file at `/volume1/garage/garage/garage.toml` does have the random strings included correctly. This can be resolved by adding the following just before the `openssl` calls:
```
export RANDFILE="/tmp/.rnd"
```

The service log in both deployments seem to be showing no errors and listening on ports (3900-3904). One minor observation is that you have a data share created for `garage` but within it you have a folder for `garage` which seems like an unnecessary additional folder. Just writing out the files in to the share path root should be sufficient. Again, just an observation.

As for the earlier comment on ARMv7L, this does not seem to be a constraint in this case as the binary seems to be statically linked so it should in theory run just fine.

fgma:
@mreid-tt You can test garage with any S3-compatible client software: https://garagehq.deuxfleurs.fr/documentation/connect/cli/

The config file is now generated from a template file and I added a workaround for openssl warnings.

Using a subfolder in the share for all files was an arbitrary decision to make testing easier.

mreid-tt:
hey @fgma, thanks for the updates. Last observation: why does it appear you have the install wizard repeated in the upgrade wizard? If you are concerned about the `service_postinst()` firing during the upgrade, you can wrap the commands within a code block like this:
```
if [ "${SYNOPKG_PKG_STATUS}" = "INSTALL" ]; then
    ...
fi
```
As it is the content of the upgrade wizard is redundant and could be removed since the share name is retained by the rest of the service support framework.

fgma:
I did so because other packages do it in a similar way? What would happen if i remove the upgrade wizard? Will it be skipped or will the install wizard be used by default?

mreid-tt:
If we have other packages that duplicate the install wizard in the upgrade wizard then please let me know as this would be a bug. The wizards should only be used to get input from the user. In the case of this upgrade all variables presented are already known. If for example the folder creation should only be run on install only then the code block wrap I proposed above is the proper solution and is recommended since the `service_postinst()` also runs on upgrade.

fgma:
@mreid-tt I've removed the upgrade file.

PR #6893, opened 2026-01-08
homeassistant: update to v2025.11.3

hgy59:
## Description

This is the first package that requires Python 3.13
It has a new approach to download all available python packages at installation time and the package contains only cross compiled wheels that are not available in the index.
With the old approach, that includes all cross wheels, the package size would be > 130MB. 
With this new approach, that also downloads HACS at installation time, the package sizes shrink to less than 5MB.

- update homeassistant to the last officially supported version of Home Assistant Core
- update to Python 3.13
- include only a minimal set of cross compiled wheels
- this release is restricted to x64 and aarch64 and requires DSM 7.2
- download and install HACS at installation time
- patch several installed wheels to fix dependencies and to find native libraries

Fixes # 

## Checklist

### Type of change

- [x] Package update


azphys:
Was able to use the [Build (x64-7.2)](https://github.com/SynoCommunity/spksrc/actions/runs/20830475183/job/59843209076?pr=6893#logs) artefact to update my existing v2024.12.5-24 installation on a DS220+ with DSM 7.3.2-86009.

Worked without issues so far, thank you so much!

PR #6897, opened 2026-01-11
framework: add standardized boolean parsing helper

mreid-tt:
## Description

**Changes**

- `mk/spksrc.bool.mk` - New helper with `bool()`, `is_true()`, `is_false()` functions
- `mk/spksrc.common.mk` - Includes the new helper
- `mk/spksrc.service.mk` - Updated `NO_SERVICE_SHORTCUT` and `USE_DATA_SHARE_WORKER` checks

**Analysis Findings**

- 35 boolean checks in mk files use `,1)` pattern
- All packages consistently use 1 for those variables
- Only `NO_SERVICE_SHORTCUT` has mixed usage (`yes`, `true`, `y`)
- Phase 2 can extend usage to other variables if needed

Fixes # 

## Checklist

### Type of change

### DIFF
--- a/mk/spksrc.bool.mk
+++ b/mk/spksrc.bool.mk
@@ -1,0 +1,36 @@
+# Standardized boolean parsing helper
+#
+# This file provides consistent boolean evaluation across the build system.
+# It normalizes various boolean representations (yes/no, true/false, 1/0, y/n)
+# into a consistent format for Make conditionals.
+#
+# Usage:
+#   include ../../mk/spksrc.bool.mk
+#   ifeq ($(call bool,$(MY_VAR)),true)
+#     # do something when MY_VAR is truthy
+#   endif
+#
+# Truthy values: yes, true, 1, y, YES, TRUE, Y (case insensitive)
+# Falsy values:  no, false, 0, n, NO, FALSE, N, empty (case insensitive)
+#
+
+# Convert any boolean-like value to lowercase 'true' or 'false'
+# Empty values are considered false
+define bool
+$(strip \
+  $(if $(1), \
+    $(if $(filter yes YES Yes true TRUE True 1 y Y,$(1)),true,false), \
+    false))
+endef
+
+# Check if a value is truthy
+# Returns non-empty string if true, empty if false
+define is_true
+$(strip $(filter true,$(call bool,$(1))))
+endef
+
+# Check if a value is falsy
+# Returns non-empty string if false, empty if true
+define is_false
+$(strip $(if $(call is_true,$(1)),,false))
+endef
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -85,6 +85,9 @@ AVAILABLE_KERNEL_VERSIONS = $(sort $(foreach arch,$(AVAILABLE_KERNEL),$(shell ec
 # Global arch definitions
 include $(BASEDIR)mk/spksrc.archs.mk
 
+# Standardized boolean parsing helpers
+include $(BASEDIR)mk/spksrc.bool.mk
+
 # Load local configuration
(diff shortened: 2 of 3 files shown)


mk/spksrc.bool.mk
```diff
@@ -0,0 +1,36 @@
...
+#
+
+# Convert any boolean-like value to lowercase 'true' or 'false'
+# Empty values are considered false
+define bool
+$(strip \
+  $(if $(1), \
+    $(if $(filter yes YES Yes true TRUE True 1 y Y,$(1)),true,false), \
+    false))
+endef
```
th0ma7, line 24: interesting approach, although shouldn't there be somewhere all valid falsy values defined?

Suggest instead something such as:
```
TRUE_VALUES  := yes true 1 y YES TRUE Y
FALSE_VALUES := no false 0 n NO FALSE N

define is_true
$(filter $(TRUE_VALUES),$(call lower,$(call strip,$(1))))
endef

# Also accept empty
define is_false
$(or \
  $(filter $(FALSE_VALUES),$(call  lower,$(call strip,$(1)))), \
  $(if $(strip $(1)),,1) \
)
endef

# assert as in case of error exit immediately
define assert_bool
$(if $(or $(call is_true,$(1)),$(call is_false,$(1))),,\
$(error $(if $(2),$(2): ,)invalid boolean value '$(1)'))
endef

# standardize the exit to TRUE or FALSE always
define bool
$(call assert_bool,$(1),$(2))
$(if $(call is_true,$(1)),TRUE,FALSE)
endef
```
Call would be such as, name of the variable (optional) help identifying which variable is in error:
```
# call bool, variable to check for, name of the variable (optional)
$(call bool,$(ENABLE_GPU),ENABLE_GPU)
```

mk/spksrc.common.mk
```diff
@@ -85,6 +85,9 @@ AVAILABLE_KERNEL_VERSIONS = $(sort $(foreach arch,$(AVAILABLE_KERNEL),$(shell ec
 # Global arch definitions
 include $(BASEDIR)mk/spksrc.archs.mk
 
+# Standardized boolean parsing helpers
+include $(BASEDIR)mk/spksrc.bool.mk
+
```
th0ma7, line 90: Pertaining to https://github.com/SynoCommunity/spksrc/pull/6906#issuecomment-3805240317 I'd be moving `spksrc.bool.mk` into `spksrc.common/bool.mk` and include it in `spksrc.common.mk` with the others.

mk/spksrc.service.mk
```diff
@@ -256,7 +256,7 @@ ifneq ($(strip $(SPK_USR_LOCAL_LINKS)),)
 endif
 ifneq ($(strip $(SERVICE_WIZARD_SHARE)),)
 # e.g. SERVICE_WIZARD_SHARE=wizard_download_dir, for DSM 6 with USE_DATA_SHARE_WORKER = yes
-ifeq ($(strip $(USE_DATA_SHARE_WORKER)),yes)
+ifeq ($(call is_true,$(USE_DATA_SHARE_WORKER)),true)
```
th0ma7, line 259: These would need to be callging `bool` and using `TRUE` or `FALSE` in capital to follow spkrc standard practice, along with including the variable name as additional (optional) parameter.  There's a few more of these in meson / cmake if I recall.

th0ma7 (changes requested):
A few suggestions, this and #6906 and #6914 benefits from being separate to track errors later on based from individual commits.

th0ma7:
from #6906 changes I suggest to simply at these bool related macros to the spksrc.common/macros.mk

mreid-tt:
Not sure this is relevant anymore. Closing for now.

PR #6877, opened 2025-12-30
synocli-devel updates and republish for x25 nas

th0ma7:
## Description

This PR introduces:
1. infrastructure fixes to prevent build environment leakage in spksrc, and updates the synocli-devel package to ensure correct builds and packaging.
2. Fix runpath so `native` binaries are relocatable (specific to CMake)
3. `cross-cmake.mk`: Disable `CMAKE_BUILD_WITH_INSTALL_RPATH` to avoid build-time executables to use the same RPATH as the cross-compiled `bin`|`lib`
4. update various `synocli-devel` dependencies targetting notably x25 platforms as package was never re-released.

### 1. Prevent build environment leakage

Several build failures were traced back to toolchain and build variables leaking across dependency boundaries, especially when mixing different build systems (autotools, CMake or Meson builds).

In particular:

Variables such as `CC`, `CXX`, `CFLAGS`, `LDFLAGS`, `STRIP`, etc. were implicitly exported even when they should not be.  This is an historic behavior of spksrc as it used to only be supporting autotools and manual build types until I added CMake and meson much later on. This caused target-built binaries to be executed as native tools, leading to errors such as Exec format error (e.g. `llvm-min-tblgen` being built for the target instead of the host).

Changes:
- Introduce an explicit `DEFAULT_ENV` selector to control which build environment is initialized.
- Prevent automatic inclusion of autotools-related variables unless explicitly required.
- Ensure that build environments do not propagate into dependent packages unless intended.
- Fix meta packages (e.g. cross/llvm) to select dependencies without initializing any build environment.

This finally allows complex compilation such as `llvm` to avoid using `CC` and `CXX` content (as now inexistant in CMake) for compiling `NATIVE` in-build tools, and in turn avoid the need of keeping around `native/llvm-9.0` for instance.

Worth noting that some leakage might still occur when going through the dependencies as the code is still as follow:
```
        @set -e; \
        for depend in $(filter-out native/%,$(TOOLKIT_DEPEND) $(BUILD_DEPENDS) $(DEPENDS)); \
        do                          \ 
          env $(ENV) $(MAKE) -C ../../$$depend ; \
        done
```
Although impact should be minimal considering:
1. most of the depencies now uses CMake or Meson
5. only a fraction (if any) might end-up being called as a dependency of autotools type build primary source
6. from this fraction only exceptions requires to compile `NATIVE` in-build targets

[...]

PR #6839, opened 2025-12-09
Noarch additional fix + early start logging printing

th0ma7:
## Description

Noarch additional fix + early start logging printing

In continuation of https://github.com/SynoCommunity/spksrc/pull/6835

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6898, opened 2026-01-11
syncstorage-rs, diesel: Fix MYSQLCLIENT_VERSION timing issue

mreid-tt:
## Description

Fixes `MYSQLCLIENT_VERSION` being empty when building `syncstorage-rs` and `diesel` for cross-compilation targets (e.g., `aarch64-7.1`), which caused the `mysqlclient-sys` crate to fail with `"" is not supported`.

**Two issues addressed**:
1. **Wildcard timing**: The original `$(wildcard $(WORK_DIR)/mariadb-connector-c-[0-9]*)` evaluated before the dependency directory exists. Moving it after the include was tested but did not resolve the issue.
2. **TC_GCC availability**: The version selection needs `TC_GCC` to choose between `mariadb-connector-c-latest` (GCC ≥ 5) and `mariadb-connector-c-3.3.11` (GCC < 5), but `TC_GCC` isn't defined until after `spksrc.cross-rust.mk` is processed.

**Solution**: Move the assignment after the include and use `$(shell sed ...)` to read the version directly from the source Makefiles, which always exist on disk. The `TC_GCC` conditional mirrors the logic in `cross/mariadb-connector-c/Makefile` to select the correct version.

Fixes https://github.com/SynoCommunity/spksrc/pull/6877#issuecomment-3734884542

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/cross/diesel/Makefile
+++ b/cross/diesel/Makefile
@@ -29,11 +29,15 @@ export MYSQL_DB_SOCKET=/run/mysqld/mysqld10.sock
 ENV += MYSQLCLIENT_LIB_DIR=$(STAGING_INSTALL_PREFIX)/lib
 ENV += MYSQLCLIENT_INCLUDE_DIR=$(STAGING_INSTALL_PREFIX)/include/mariadb
 
-# pass mysql client version (needed by mysqlclient-sys crate)
-# Read from mariadb-connector-c-latest Makefile which exists at parse time
-ENV += MYSQLCLIENT_VERSION=$(shell sed -n 's/^PKG_VERS = //p' $(BASEDIR)cross/mariadb-connector-c-latest/Makefile)
-
 # let ./configure find mysql_config (it is a script and works for cross compile)
 ENV += "PATH=$(PATH):$(STAGING_INSTALL_PREFIX)/bin"
 
 include ../../mk/spksrc.cross-rust.mk
+
+# pass mysql client version (needed by mysqlclient-sys crate)
+# Use same conditional as mariadb-connector-c to select correct version
+ifeq ($(call version_ge, $(TC_GCC), 5),1)
+ENV += MYSQLCLIENT_VERSION=$(shell sed -n 's/^PKG_VERS = //p' $(BASEDIR)cross/mariadb-connector-c-latest/Makefile)
+else
+ENV += MYSQLCLIENT_VERSION=$(shell sed -n 's/^PKG_VERS = //p' $(BASEDIR)cross/mariadb-connector-c-3.3.11/Makefile)
+endif
--- a/cross/syncstorage-rs/Makefile
+++ b/cross/syncstorage-rs/Makefile
@@ -33,13 +33,17 @@ ENV += MYSQLCLIENT_LIB_DIR=$(STAGING_INSTALL_PREFIX)/lib
 # to find dependencies of libmysqlclient (libz)
 ENV += RUSTFLAGS="-Clink-arg=-Wl,--rpath,$(INSTALL_PREFIX)/lib -Clink-arg=-Wl,--rpath-link,$(STAGING_INSTALL_PREFIX)/lib"
 
-# pass mysql client version (needed by mysqlclient-sys crate)
-# Read from mariadb-connector-c-latest Makefile which exists at parse time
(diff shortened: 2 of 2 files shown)


th0ma7:
The reasoning above is right, the method could be done differently which could serve other means...

Throwing out an idea, what if we could invoke a `make DEPENDENCY=mariadb-connector-c dependency-version`.  Such method could be added to `spksrc.dependency-tree.mk`.  This could even be further adapted to print the version along with the dependency name into a new method such as `make dependency-tree|list|flat-version` or similar.

mreid-tt:
hey @th0ma7, I've simplified the fix without going into the framework. Let me know if this works for you.

mreid-tt:
hey @th0ma7, looking into this more I realise we also have a build problem for `cross/mariadb-connector-c` itself. If you try to build it, it fails for something like `make arch-comcerto2k-7.1`. I did a little digging in the recent commits and realised that it did work before PRs #6839 and #6857. Putting all this into the AI for testing it came back with this:

## Summary

**PR #6857's rationale was partially correct but incomplete:**

1. **Valid concern**: Packages that set `ADDITIONAL_CFLAGS` **before** the include (like `mosquitto`) do get flag duplication because:
   - The flag is embedded in `CFLAGS` via `tc_vars.mk`
   - The cmake toolchain file adds `$(ADDITIONAL_CFLAGS)` again

2. **What PR #6857 broke**: Packages that set `ADDITIONAL_CFLAGS` **after** the include using TC_GCC conditionals (like `mariadb-connector-c-3.3.11`) lose their flags entirely because:
   - TC_GCC is only available after the include
   - So these conditionals must run after the include
   - By then, tc_vars.mk has already expanded `$(ADDITIONAL_CFLAGS)` to empty

**Affected packages by category:**

| Package | Flags set | Build system | Impact of PR #6857 | Impact of my fix |
|---------|-----------|--------------|-------------------|------------------|
| mosquitto | Before include | cmake | No duplication ✓ | Duplication (harmless) |
| lz4, zstd | Before include | meson | No duplication ✓ | Duplication (harmless) |
| mariadb-connector-c-3.3.11 | After include (TC_GCC conditional) | cmake | **BROKEN** ✗ | Works ✓ |
| `fish`, `intel-*`, `libaom-*`, `libwebp` | After include (TC_GCC conditional) | cmake | **BROKEN** ✗ | Works ✓ |

**Recommendation:**
1. My fix (restoring `$(ADDITIONAL_*FLAGS)`) is the correct approach - duplication is harmless, missing flags breaks builds
2. Optionally, packages like `mosquitto` that set flags before the include could be cleaned up to avoid duplication (move the flag to the include point or document the duplication is expected)

---

I have the branch that it created to restore `$(ADDITIONAL_*FLAGS)` but before I did a PR, I wanted to get your thoughts on this.

th0ma7:
this fixes the issue https://github.com/SynoCommunity/spksrc/pull/6899
I double checked and Meson isn't affected, the change only impacted CMake.
Thnx for finding this out :)

mreid-tt:
> this fixes the issue #6899 I double checked and Meson isn't affected, the change only impacted CMake. Thnx for finding this out :)

You're welcome. There was an echo in line 34 that could also be restored:

```diff
-	 echo $(RUSTFLAGS) | tr ' ' '\n' | sed -e "s/^/  \"/" -e "s/$$/\"/" ; \
+ 	 echo $(RUSTFLAGS) $(ADDITIONAL_RUSTFLAGS) | tr ' ' '\n' | sed -e "s/^/  \"/" -e "s/$$/\"/" ; \
```

Not critical. Just sharing what was in the AI's recommendation.

th0ma7:
> > this fixes the issue #6899 I double checked and Meson isn't affected, the change only impacted CMake. Thnx for finding this out :)
> 
> You're welcome. There was an echo in line 34 that could also be restored:
> 
> ```diff
> -	 echo $(RUSTFLAGS) | tr ' ' '\n' | sed -e "s/^/  \"/" -e "s/$$/\"/" ; \
> + 	 echo $(RUSTFLAGS) $(ADDITIONAL_RUSTFLAGS) | tr ' ' '\n' | sed -e "s/^/  \"/" -e "s/$$/\"/" ; \
> ```
> 
> Not critical. Just sharing what was in the AI's recommendation.

Fair point, I'll try to reproduce as we don't have any CMake based rust project just yet...

PR #6900, opened 2026-01-13
CMake + Meson ADDITIONAL_*FLAGS fixes

th0ma7:
## Description

This PR addresses several long-standing issues in the handling of compiler flags for cross-compilation in spksrc, especially when generating CMake toolchain files and Meson cross files.

### 1. Introduce an order-preserving uniq macro for compiler flags
A new GNU make macro uniq is introduced to remove duplicate flags while preserving their original order. This is critical for cross-compilation, where flags such as:
- `-I`, `-L`
- `-Wl,…`
- ABI / tuning flags
- optimization flags (`-O*`, `-fPIC`)

are order-sensitive and cannot be safely deduplicated using $(sort ...).

The macro is now used when emitting compiler and linker flags into both the CMake toolchain file and the Meson cross file.  This guarantees:
- no duplicated flags
- no semantic changes caused by flag reordering
- deterministic output regardless of variable expansion order

### 2. Correct and re-integrate `ADDITIONAL_*FLAGS` in CMake and Meson cross files
This PR fixes a subtle but important issue related to how `ADDITIONAL_*FLAGS` are propagated. The root problem is that `ADDITIONAL_*FLAGS` behave differently depending on where they are defined relative to:
```
include ../../mk/spksrc.cross-meson.mk
```
If `ADDITIONAL_*FLAGS` are defined before the include, they are already appended to `CFLAGS` / `CXXFLAGS` / etc. definitions as interpreted at that stage.  If they are defined after the include, they are not yet present in `CFLAGS` and must be explicitly re-added.

This leads to two problematic cases:

#### Case 1 – flag silently missing
```

ifeq ($(call version_gt,$(TC_GCC),4.9),1)
ADDITIONAL_CFLAGS += -Wno-error=maybe-uninitialized
ADDITIONAL_CFLAGS += -Wno-error=strict-aliasing
endif
```
In this common scenario, the flags are added after the include (because the GCC version is only known then), but are not automatically present in the generated CMake or Meson configuration unless explicitly re-integrated.

#### Case 2 – duplicated flags
```
ADDITIONAL_CFLAGS += -Wno-error
```
Here, the flag is:
- already included indirectly via `CFLAGS`
- then appended again explicitly

Result: duplicated flags in the generated toolchain file.

#### The fix
This PR explicitly re-integrates `ADDITIONAL_*FLAGS` when generating CMake and Meson configuration files and applies the uniq macro at emission time.  This ensures that:
- flags are always present, regardless of where they are defined
- flags appear exactly once
- the behavior is deterministic and order-independent

### 3. Add missing Fortran flags to the CMake toolchain
While `CMAKE_Fortran_COMPILER` can be defined in spksrc depending of the GCC in use (e.g. DSM>=7 using gcc>=8.5), the corresponding flag variables were missing from the generated CMake toolchain:
- `CMAKE_Fortran_FLAGS`
- `CMAKE_Fortran_FLAGS_DEBUG`

This PR adds proper support for Fortran flags, aligned with C and C++ handling:
- Uses FFLAGS, TC_EXTRA_FFLAGS, and ADDITIONAL_FFLAGS
- Applies the same uniq logic for deduplication
- Ensures ABI, sysroot, PIC, and optimization flags are consistent across languages

This fixes potential issues when building packages that enable Fortran conditionally during cross-compilation.

Fixes https://github.com/SynoCommunity/spksrc/pull/6898#issuecomment-3740818411

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.cross-meson-crossfile.mk
+++ b/mk/spksrc.cross-meson-crossfile.mk
@@ -65,9 +65,9 @@ ifneq ($(strip $(MESON_BUILTIN_CPP_ARGS)),)
 	@echo -ne "\t'$(MESON_BUILTIN_CPP_ARGS)',\n"
 endif
 ifeq ($(GCC_DEBUG_INFO),1)
-	@echo $(call uniq,$(patsubst -O%,,$(CPPFLAGS) $(GCC_DEBUG_FLAGS) $(ADDITIONAL_CPPLAGS))) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
+	@echo $(call uniq,$(patsubst -O%,,$(CPPFLAGS) $(GCC_DEBUG_FLAGS) $(ADDITIONAL_CPPFLAGS))) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
 else
-	@echo $(call uniq,$(CPPFLAGS) $(ADDITIONAL_CPPLAGS)) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
+	@echo $(call uniq,$(CPPFLAGS) $(ADDITIONAL_CPPFLAGS)) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
 endif
 	@echo -ne "\t]\n"
 	@echo


th0ma7:
@mreid-tt this was waaaaaaaaaayyyyyyyyyy more subtile than expected... I tried documenting at best the root-cause of this issue where depending of when `ADDITIONAL_*FLAGS` is being defined it will (or not) appear in the resulting meson crossfile or CMake toolchain file.

This explains why those flags where duplicate in some cases and not appearing in other depending of the inclusion (or not) of those extra flags.  I now also enforced the re-introduction of `TC_EXTRA_*FLAGS` (special arch specific flags defined at toolchain level that normally are included in `CFLAGS`/`RUSTFLAGS`) which will end-up being discarded by the `uniq` macro as needed if being duplicate, but this ensure that every type of build flags are properly introduced, always, period.

You might notice commented `#ifneq ($(strip $(FFLAGS)),)`, these will be re-enabled in #6877 part of variable leaking that is being fixed.  Hopefully after all that meson + cmake will be more solid (although they've been not too bad since their inclusion into this now old framework).

I'll let this sit until tomorrow so I can further test and review before merging (although preliminary testing looked good).  Feel free to test it as well on you end, that can only help :)

Noting, cmake toolchain file reside under `$(WORK_DIR)/$(APP_DIR)/build` and meson crossfile under `$(WORK_DIR)/$(APP_DIR)` in case you want to look at auto-generated files.

mreid-tt:
hey @th0ma7, great work. ran it though the AI and only found one significant issue as follows:

---

Based on my analysis:

**Issues Found:**
1. **TYPO**: Lines 68 and 70 in `mk/spksrc.cross-meson-crossfile.mk` use `ADDITIONAL_CPPLAGS` instead of `ADDITIONAL_CPPFLAGS` (missing 'F')

**Strengths:**
1. The `uniq` macro is order-preserving - critical for flag ordering
2. Addresses both the duplication issue AND missing flags issue simultaneously
3. Fortran support is added for future-proofing
4. Clean removal of the comcerto2k-specific patch since the framework fix makes it unnecessary

Now let me compose and share the detailed review comment for the user to post:

Based on my detailed review of PR #6900, here's my analysis:

**Summary**

PR #6900 introduces an elegant solution to the `ADDITIONAL_*FLAGS` timing issue using a new order-preserving `uniq` macro. This solves both problems simultaneously:
- **Missing flags**: Packages setting flags AFTER the include (using `TC_GCC` conditionals) now work correctly
- **Duplicate flags**: Packages setting flags BEFORE the include no longer cause duplicates

**Commits in PR**

1. **`64618911` - common.mk: Add uniq macro definition** - Adds an order-preserving `uniq` macro that removes duplicate words while maintaining the first occurrence's position
   
2. **`4efb0f6a` - CMake+Meson: Re-integrated `ADDITIONAL_*FLAGS` and use uniq macro** - Wraps all flag variables with `$(call uniq,...)` in both toolchain files
   
3. **`3f617c94` - mariadb-connector-c-3.3.11: Remove comcerto2k specific patch** - Cleanup since the framework fix makes the package-level workaround unnecessary

**Issues Found**

🐛 **Typo in `mk/spksrc.cross-meson-crossfile.mk`** (lines 68 and 70):
```makefile
$(ADDITIONAL_CPPLAGS)  # ← Missing 'F'
```
Should be:
```makefile
$(ADDITIONAL_CPPFLAGS)
```

**Strengths**

- **Order preservation** in `uniq` is critical - flag ordering matters for `-I`, `-L`, ABI flags
- **Clean solution** that handles both pre-include and post-include flag assignments
- **Fortran support** added for future-proofing (commented conditionals pending #6877)
- **Proper cleanup** of the temporary comcerto2k patch

mk/spksrc.cross-meson-crossfile.mk
```diff
@@ -47,60 +47,62 @@ ifneq ($(strip $(MESON_BUILTIN_C_ARGS)),)
...
+	@echo $(call uniq,$(LDFLAGS) $(ADDITIONAL_LDFLAGS)) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/" ; \
 	echo -ne "\t]\n"
 	@echo
 	@echo "cpp_args = ["
 ifneq ($(strip $(MESON_BUILTIN_CPP_ARGS)),)
 	@echo -ne "\t'$(MESON_BUILTIN_CPP_ARGS)',\n"
 endif
 ifeq ($(GCC_DEBUG_INFO),1)
-	@echo $(patsubst -O%,,$(CPPFLAGS) $(GCC_DEBUG_FLAGS)) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
+	@echo $(call uniq,$(patsubst -O%,,$(CPPFLAGS) $(GCC_DEBUG_FLAGS) $(ADDITIONAL_CPPLAGS))) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
```
mreid-tt:
```suggestion
	@echo $(call uniq,$(patsubst -O%,,$(CPPFLAGS) $(GCC_DEBUG_FLAGS) $(ADDITIONAL_CPPFLAGS))) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
```

mk/spksrc.cross-meson-crossfile.mk
```diff
@@ -47,60 +47,62 @@ ifneq ($(strip $(MESON_BUILTIN_C_ARGS)),)
...
 else
-	@echo $(CPPFLAGS) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
+	@echo $(call uniq,$(CPPFLAGS) $(ADDITIONAL_CPPLAGS)) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
```
mreid-tt:
```suggestion
	@echo $(call uniq,$(CPPFLAGS) $(ADDITIONAL_CPPFLAGS)) | tr ' ' '\n' | sed -e "s/^/\t'/" -e "s/$$/',/"
```

th0ma7:
@mreid-tt do you nind applying the fix and merging?  At the office and won't have time to reconvene before tonight or tomorrow - in case you're waiting for it.

PR #6901, opened 2026-01-13
framework: fix version_ge dependency in directories.mk

mreid-tt:
## Description

Fixes the circular dependency issue in `mk/spksrc.directories.mk` where `version_ge` is called before it's defined.

### Problem

In files like `spksrc.cross-cc.mk`, the include order is:

1. Line 25: `include ../../mk/spksrc.directories.mk`
2. Line 28: `include ../../mk/spksrc.common.mk`

The `version_ge` function is defined in `spksrc.common.mk` (line 160), but `spksrc.directories.mk` tries to use it at line 64 to determine `INSTALL_PREFIX_VAR`. This causes the conditional to silently fail, resulting in incorrect var directory paths for DSM 7+ builds.

### Solution

Replace the `$(call version_ge, ${TCVERSION}, 7.0)` with an inline shell comparison using the same logic that `version_ge` uses internally:
```
$(shell if printf '%s\n' "$(TCVERSION)" "7.0" | sort -VCr ; then echo 1; fi)
```
This is the simplest fix because:

- Single-line change with minimal footprint
- Preserves exact DSM 6 vs 7 behavior
- No include reordering or new files required
- Uses identical shell logic to `version_ge`

### Testing

| DSM Version | INSTALL_PREFIX_VAR | Expected | Status |
|------------|-------------------|----------|--------|
| 7.1 | /var/packages/demoservice/var | /var/packages/SPK_NAME/var | ✅ |
| 6.2.4 | /var/packages/demoservice/target/var | /var/packages/SPK_NAME/target/var | ✅ |

### Related

- Ref: PR #5829 discussion
- Alternative to commit e0bcf980 which removed the conditional entirely

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/mk/spksrc.directories.mk
+++ b/mk/spksrc.directories.mk
@@ -61,7 +61,8 @@ endif
 # when invoking make from under spk/*.  Setting var when
 # test-building dependencies from under cross/* is unecessary.
 #
-ifeq ($(call version_ge, ${TCVERSION}, 7.0),1)
+# Using inline shell comparison to avoid dependency on version_ge from spksrc.common.mk
+ifeq ($(shell if printf '%s\n' "$(TCVERSION)" "7.0" | sort -VCr ; then echo 1; fi),1)
 ifeq ($(lastword $(subst /, ,$(INSTALL_PREFIX))),target)
 INSTALL_PREFIX_VAR = $(shell dirname $(INSTALL_PREFIX))/var
 endif


mreid-tt:
@th0ma7 @hgy59, is this a good solution to the issue raised in https://github.com/SynoCommunity/spksrc/pull/5829#issuecomment-3733332942?

th0ma7:
In theory yes, but wouldn't inverting the order of the include suffice?

mreid-tt:
> In theory yes, but wouldn't inverting the order of the include suffice?

Unfortunately, inverting the include order won't work because `spksrc.common.mk` depends on variables defined in `spksrc.directories.mk` (`WORK_DIR`, `PIP_DISTRIB_DIR`, `PIP_CACHE_DIR`, `WHEELHOUSE`). Swapping them would leave those variables undefined when `common.mk` tries to use them.

The inline shell approach avoids this circular dependency with a single-line change using the same logic as `version_ge`.

th0ma7:
Indeed but you can delay their allocation if not mandatory for a ifeq type testing in either  common.mk or directory.mk by using := vs = type thing

th0ma7:
Indeed but you can delay their allocation if not mandatory for a ifeq type testing in either  common.mk or directory.mk by using := vs = type thing

hgy59:
> In theory yes, but wouldn't inverting the order of the include suffice?

no, we need to split it.
we need a minimal base.mk (or global.mk, defines.mk, ....) that does not depend on any directory but contains only plain macros or defines
then include directory.mk
and after this a _common.mk_ with the rest of the current common.mk


candidates for base.mk

```

# Stop on first error
SHELL := $(SHELL) -e

# Define $(empty) and $(space)
empty :=
space := $(empty) $(empty)

# Display message in a consistent way
MSG = echo "===> "

# Available languages
LANGUAGES = chs cht csy dan enu fre ger hun ita jpn krn nld nor plk ptb ptg rus spn sve trk

# Terminal colors
RED=$$(tput setaf 1)
GREEN=$$(tput setaf 2)
NC=$$(tput sgr0)

# Macro: Version Comparison
version_le = $(shell if printf '%s\n' "$(1)" "$(2)" | sort -VC ; then echo 1; fi)
version_ge = $(shell if printf '%s\n' "$(1)" "$(2)" | sort -VCr ; then echo 1; fi)
version_lt = $(shell if [ "$(1)" != "$(2)" ] && printf "%s\n" "$(1)" "$(2)" | sort -VC ; then echo 1; fi)
version_gt = $(shell if [ "$(1)" != "$(2)" ] && printf "%s\n" "$(1)" "$(2)" | sort -VCr ; then echo 1; fi)

# Remove duplicate words within string while preserving order
define uniq
$(strip \
  $(eval __seen :=) \
  $(foreach f,$1, \
    $(if $(filter $f,$(__seen)),, \
      $(eval __seen += $f)$(f) \
    ) \
  ) \
)
endef

# Macro: dedup
#        removes duplicate entries from a specified delimiter,
#        preserving the order of unique elements.
dedup = $(shell /bin/bash -c '\
    input="$$(echo "$1" | xargs)"; \
    delimiter="$$(echo "$2" | xargs)"; \
    echo "$$input" | \
    tr "$$delimiter" "\n" | \
    awk '\''!seen[$$0]++ {print $$0}'\'' | \
    tr "\n" "$$delimiter" | \
    sed "s/$$delimiter$$//" \
')

# Macro: merge
#        merges multiple environment variable values from a given input string,
#        inverting their order and separating them with a specified delimiter
merge = $(shell /bin/bash -c '\
    input="$$(echo "$1" | xargs)"; \
    var_name="$$(echo "$2" | xargs)"; \
    delimiter="$$(echo "$3" | xargs)"; \
    echo "$$input" | \
    grep -o "$$var_name=[^ ]*" | \
    tac | \
    sed "s/^$$var_name=//" | \
    tr "\n" "$$delimiter" | \
    sed "s/$$delimiter$$//" \
')
```

All other definitions are related to directories, pip and archs due include of archs.mk, and must be defined in the current/regular common.mk file.

th0ma7:
Ah!  Now i get it, fair point and worth adding to the todo by list...

mreid-tt:
Superseded by #6906

PR #6908, opened 2026-01-18
framework: Add CC/CXX and other build tools as Make variables in tc_flags

mreid-tt:
## Description

This is a follow-on from #6877 to address a regression in 1ebe52b affecting the cross-compilation of boost libraries.

Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/mk/spksrc.tc.mk
+++ b/mk/spksrc.tc.mk
@@ -332,7 +332,15 @@ autotools_vars:
 
 .PHONY: tc_flags
 tc_flags:
-	@echo CFLAGS := $(CFLAGS) $$\(GCC_DEBUG_FLAGS\) $$\(ADDITIONAL_CFLAGS\) ; \
+	@echo CC := $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)gcc ; \
+	echo CXX := $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)g++ ; \
+	echo LD := $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)ld ; \
+	echo AR := $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)ar ; \
+	echo AS := $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)as ; \
+	echo NM := $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)nm ; \
+	echo RANLIB := $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)ranlib ; \
+	echo STRIP := $(WORK_DIR)/$(TC_TARGET)/bin/$(TC_PREFIX)strip ; \
+	echo CFLAGS := $(CFLAGS) $$\(GCC_DEBUG_FLAGS\) $$\(ADDITIONAL_CFLAGS\) ; \
 	echo CPPFLAGS := $(CPPFLAGS) $$\(GCC_DEBUG_FLAGS\) $$\(ADDITIONAL_CPPFLAGS\) ; \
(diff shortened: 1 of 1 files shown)


th0ma7:
I don't have access to my build system atm but nope, the problem lies elsewhere as autotools.mk has those declarations and probably needs to be included instead.

Once my system is available I'll have a look at it.

th0ma7:
@mreid-tt So I've been playing with this issue quite a bit to figure out how to best resolve this... The simple and immediate fix is as follow in boost makefile:
```
@$(RUN) sh -c 'echo "using gcc : ${TC_GCC} : $${CXX} : <address-model>\"$(ADDRESS_MODEL)\" <cflags>\"$(CFLAGS)\" <cxxflags>\"$(CXXFLAGS) $(ADDITIONAL_CXXFLAGS)\" <linkflags>\"$(LDFLAGS)\" <link>\"shared\" ; " > $(WORK_DIR)/$(PKG_DIR)/user-config.jam'
ifneq ($(strip $(WITH_PYTHON_LIBRARY)),)
```
By invoking a sub-shell the variables incuded in `env $(ENV)` are becoming avaiable, as they should.

Question is, why this change of behavior?

It hapens that  `env $(ENV)` ended-up leaking from the shell environment to the make environment under certain conditions, making variables embeded in `$(ENV)` to become directly available in make.  When calling `${CXX}` this is exactly the same thing as calling `$(CXX)`, which means that these have leaked in make.

I've tested this from an old commit to have the previous behavior, using this code:
```
.PHONY: boost_pre_compile
boost_pre_compile:
        @$(RUN) echo "pseudo shell CXX: ${CXX}"
        @$(RUN) echo "make CXX: $(CXX)"
        @$(RUN) echo "shell no sh CXX: $${CXX}"
        @$(RUN) sh -c 'echo "shell whith sh CXX: $${CXX}"'
        $(info CXX origin: $(origin CXX))
        exit 2
```
Which returns the following:
```
===>  Compiling for boost
CXX origin: default
pseudo shell CXX: g++
make CXX: g++
shell no sh CXX: 
shell whith sh CXX: /home/spksrc/spksrc/toolchain/syno-x64-7.1/work/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu-g++
exit 2
```
So even prior to my recent changes you could not get access to `CXX` value unless invoking a sub-shell as otherwise it uses the default make, in this case the host `g++`.  Interestingly this will end-up working when boost is called as a dependency because in `spksrc.depend.mk` we use the following:
```
env $(ENV) $(MAKE) -C ../../$$depend ; \
```
Which in turns make the variable in `$(ENV)` to leak into the default make environment when building a dependency.... Exactly what we try to prohibit to avoid messing around with CMake and Meson environments.  My recent changes blocked this.

To demonstrate, if you try to build boost "as-is" on current master or from an older git commit it will fail miserably as `${CXX}` is always worth g++.  Although on an older commit, if boost is a dependency of something else it will work.  I always wondered why I could not build it from cross, now I get it.

Going back to the proper fix, using `@$(RUN) sh -c 'echo "using gcc : ${TC_GCC} : $${CXX} :  ...` in boost solves things.

Last question is: do we want to have CC, CXX, etc. directly available in make or not?  I believe this wasn't a planned thing but rather an unexpected behavior that we tried to make some sense of... Still, I'll see what can be done as this whole environment variable thing can clearly be simplified a little furhter.

So I suggest you simply fix boost makefiles to use `sh -c` and things will work with no changes to the framework, fixing a long standing issue with boost makefile itself.

mreid-tt:
Hey @th0ma7, thanks for the analysis. Other AI reviews initially pointed toward changes in the Boost makefile as the solution, but I pushed for a deeper look at the framework since we had successful cross-compiled updates for both ZNC and Deluge about 6–7 months ago, with the last Boost changes predating those updates.

Your solution is much more elegant than the more complex approaches that were suggested. I’ll open a new PR to address Boost directly and close this one. Thanks again for the solid analysis.

mreid-tt:
Superseded by a proper solution in #6911.

PR #6909, opened 2026-01-19
Node Exporter: Update to v1.10.2

neutralalice:
## Description

Updating the node exporter version to include fixes and features

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [x] Package update

### DIFF
--- a/cross/node-exporter/Makefile
+++ b/cross/node-exporter/Makefile
@@ -1,5 +1,5 @@
 PKG_NAME = node_exporter
-PKG_VERS = 1.8.2
+PKG_VERS = 1.10.2
 PKG_EXT = tar.gz
 PKG_DIST_NAME = v$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://github.com/prometheus/node_exporter/archive
--- a/cross/node-exporter/digests
+++ b/cross/node-exporter/digests
@@ -1,3 +1,3 @@
-node_exporter-1.8.2.tar.gz SHA1 243512a1691787ec799ba92cdd67e394bab5e881
-node_exporter-1.8.2.tar.gz SHA256 f615c70be816550498dd6a505391dbed1a896705eff842628de13a1fa7654e8f
-node_exporter-1.8.2.tar.gz MD5 096bfaf7a902b105288b616cc3215d63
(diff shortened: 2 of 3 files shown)


neutralalice:
This is working on my synology.  I check my metrics endpoint, and it looks good, so that to me is "functionality working". Is there anything else I need to test after manually upgrading?

```node_exporter_build_info{branch="master",goarch="arm64",goos="linux",goversion="go1.25.4",revision="4de15094207a6d2e7a4c5d3245fdfc1c5380f78a",tags="unknown",version="1.10.2"} 1```

mreid-tt:
hey @neutralalice — thanks for your submission, and apologies for the delayed review. Based on the package structure and the similar changes in the last PR, everything looks good to me. We typically ask that all items in the checklist be completed before approval, but I’ll go ahead and approve this one for now.

neutralalice:
No problem, it just wasn't clear to me  "new installation of package" needed to be ticked/tested as well given that it's not a new package to the repository;  I will keep in mind to do all of the checklist items next time regardless;

I imagine generally the workflow for version changes then would be "upgrade current version to see if it works -> uninstall -> reinstall" so that I can cover both bases.

Is there anything else with respect to architecture testing that should normally happen or is building all-supported essentially good enough for that? I imagine most independant pr's don't have multiple architectures sitting around to test the install/upgrade setup of each version.

mreid-tt:
> I imagine generally the workflow for version changes then would be "upgrade current version to see if it works -> uninstall -> reinstall" so that I can cover both bases.

Yes, that would be an ideal sequence.

> Is there anything else with respect to architecture testing that should normally happen or is building all-supported essentially good enough for that? I imagine most independant pr's don't have multiple architectures sitting around to test the install/upgrade setup of each version.

Architecture testing would be difficult and would usually be called for when there are specific architectural build differences. So for example, I only have x64 platforms at my disposal but I can examine binaries built using the standard `file` command in linux to confirm that something was compiled for aarch64 (ARM) correctly.

It really depends on the package itself to determine the level of testing appropriate. As another example some other apps I will test for both DSM 6 and DSM 7 whereas some only have one target (DSM 7.2).

As for functional testing, ideally those with intimate knowledge of the app could identify how it should be tested (and we could eventually update our wiki with this so that others could be guided).

PR #6911, opened 2026-01-19
cross/boost, cross/znc: fix cross-env handling during build

mreid-tt:
## Description

This PR resolves cross-compilation issues by addressing two related but independent package-level problems uncovered during testing.

First, it corrects how Boost’s `user-config.jam` is generated during the build. Previously, the `using gcc` line was written directly via `echo`, which caused certain toolchain variables (notably `CXX`) to be expanded outside the cross-build environment. This resulted in incorrect compiler and linker selection in some cases. The fix wraps the `echo` in a `$(RUN)` subshell, ensuring all compiler variables are expanded within the proper cross environment and restoring reliable Boost builds (as discussed in https://github.com/SynoCommunity/spksrc/pull/6908#issuecomment-3768173690).

Second, this PR fixes cross-compilation of ZNC extra modules, which are built via `znc-buildmod` rather than the main CMake flow. Unlike standard CMake targets, `znc-buildmod` depends on compiler, autotools, and flag variables being present directly in the execution environment. The updated ZNC Makefile explicitly enables the required environments (`cmake`, `autotools`, and `flags`) and invokes `znc-buildmod` within a subshell that preserves those variables. This ensures the correct cross compiler and linker are used when building modules and prevents host tool leakage during the link stage.

Together, these changes developed by @th0ma7, restore successful cross-compilation for Boost-dependent packages and ZNC, while keeping the fixes narrowly scoped to the affected packages.

Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/cross/mpd/Makefile
+++ b/cross/mpd/Makefile
@@ -6,14 +6,16 @@ PKG_DIST_SITE = https://github.com/MusicPlayerDaemon/MPD/archive
 PKG_DIST_FILE = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIR = MPD-$(PKG_VERS)
 
+HOMEPAGE = https://www.musicpd.org/
+COMMENT  = Music Player Daemon (MPD) is a flexible, powerful, server-side application for playing music.
+LICENSE  = GPLv2
+
 # with mpd ver 0.22 (2020/09/23)
 # - switch to C++17 (GCC 8 or clang 5)
 # GCC version is too old. You need at least version 8.
 REQUIRED_MIN_DSM = 7.0
 UNSUPPORTED_ARCHS = comcerto2k
 REQUIRED_MIN_SRM = 1.3
-
-include ../../mk/spksrc.archs.mk
 
 DEPENDS  = cross/glib
 DEPENDS += cross/alsa-lib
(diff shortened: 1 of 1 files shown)


th0ma7 (approved):
LGTM although i suggest leaving the check running to ensure there isn't any other issue pending.

mreid-tt:
hey @th0ma7, there seems to be a separate cross-compile issue with the `znc-buildmod` component which may also need a fix?

th0ma7:
Probably the modules / plugins.. was tricky to setup properly in the first place so i bet its there.  I'll have a look at it, shouldn't be too hard to fix.

th0ma7:
Now fixed with https://github.com/SynoCommunity/spksrc/pull/6911/changes/245d5c2d452ad69de654d0a54211a113ed69c834 - its actually interesting how one can control the build environment using `DEFAULT_ENV` variable... 

Something I noted though (and unrelated), the (re)generation of the toolchain `*.mk` files is not following the `$(WORK_DIR)` resulting in having an unecessary `spksrc/cross/znc/work-<arch>-<version>/tc_vars*.mk` while everything should be within the `spk/znc/work-<arch>-<version>`.  I'll check this out as well.

th0ma7:
The issue I found is unrelated, is not a blocker and will require a different PR... moving this framework from being autotools centric to multi-build environment is not an easy thing... sigh.

mreid-tt:
hey @th0ma7, looks like we are almost there. Now we have an `mpd` build failure on `evansport-7.1`. seems like this block isn't being processed:

https://github.com/SynoCommunity/spksrc/blob/155a80d9ecc3d1e341e7390b80512f1128bf15ca/cross/mpd/Makefile#L48-L51

Because we are getting this error:
```
error: ‘SO_REUSEPORT’ was not declared in this scope
```

th0ma7:
So in meson `cpp` defaults to `g++` whereas the notion of `cxx` nor `CXXFLAGS` doesn't really exists.  With the new enforced clean meson environment there is no more `CXXFLAGS` having leaked in the environment whereas `ADDITIONAL_CXXFLAGS` would have been included... 

Long-story-short: with meson use `ADDITIONAL_CPPFLAGS`.  And problem solved.

PR #6915, opened 2026-01-21
framework: support of ARM64 architecture for the Dockerfile

brice-fromentin:
## Description

When building using the amd64 docker container, for dotnetX-runtime the build is really slow. Building jellyfin has a coredump while building de dotnet server using dotnet 9 SDK, and nodejs build is also really slow.

The changes are mainly done to use the current CPU architecture used by the container (tested on Mac with Apple Silicon).

Currently, "native/dotnet-sdk-9.0" and "native/nodejs_22" are modified to support "make digests" with multiple architecture and "Dockerfile" is modified to enable ARM64 support. I also add a script to build and start the container.

I currently use "ifneq ($(strip $(CARGO_HOME)),)" to detect if a build is running and make both digests and build work but it is not, IMHO, a clean way to this. I will be gratefull if you can explain me how to create a env var for example to specify that a "make digests" is currently running.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update


mreid-tt:
hey @brice-fromentin, I would love to be able to build natively on my Mac but isn't the limiting factor the Synology toolchains which are built for x64 only? When you say "_tested on Mac with Apple Silicon_", what exactly do you mean? were you able to complete a full package build?

brice-fromentin:
hello @mreid-tt , currently I am able with thoses modification to successfully build dotnet8, dotnet9 and dotnet10 runtimes and also jellyfin.

dotnetX runtimes is the easy part because they repackage files in some way. Jellyfin relies on dotnet9 sdk and nodejs 22, the trick is to use the ARM64 compilers from dotnet sdk and node JS.

hgy59:
This is a very special use case that is out of scope of the spksrc dev environment.

brice-fromentin:
@hgy59 , it depends if we want to enhance the development experience for those who have arm machine, Mac on Apple Silicon are slow and on some project, they just not work.

PR #6926, opened 2026-01-24
Syncthing: Update to v2.0.13

### LINKED ISSUE #6721: Syncthing ver 2.x for DS214+ MARVELL Armada XP MV78230

mreid-tt:
## Description

1. Update Syncthing to v2.0.13 (built with Go 1.25).
2. Drop legacy ARMv5 models that required the Go 1.23 compiler.

Fixes #6721

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/spk/syncthing/Makefile
+++ b/spk/syncthing/Makefile
@@ -13,7 +13,7 @@ MAINTAINER = acolomb
 DESCRIPTION = Automatically sync files via secure, distributed technology.
 DESCRIPTION_FRE = Synchronisation automatique de fichiers via une technologie sécurisée et distribuée.
 DISPLAY_NAME = Syncthing
-CHANGELOG = "1. Update Syncthing to v2.0.13 (built with Go 1.25). <br/>2. Drop legacy ARMv5 models that required the Go 1.23 compiler. <br/><br/>ATTENTION: This is a major version upgrade requiring a database migration on first launch. Please read the <a target=_blank href=https://github.com/syncthing/syncthing/releases/tag/v2.0.0>official release notes</a> and be prepared for some service downtime. <br/>REMARKS: New packages are rarely created because Syncthing's in-app updater is used."
+CHANGELOG = "1. Update Syncthing to v2.0.13 (built with Go 1.25). <br/>2. Drop legacy ARMv5 models that required the Go 1.23 compiler. <br/><br/>ATTENTION: This is a major version upgrade requiring a database migration on first launch. Please read the <a target=_blank href=https://github.com/syncthing/syncthing/releases/tag/v2.0.13>official release notes</a> and be prepared for some service downtime. <br/>REMARKS: New packages are rarely created because Syncthing's in-app updater is used."
 HOMEPAGE = https://www.syncthing.net
 LICENSE = MPLv2.0
 STARTABLE = yes


mreid-tt:
hey @hgy59, I've cleaned up and tested this with the latest Syncthing and is ready to go. Given the comments in #6865, I'm not sure there is anything further to fix.

acolomb:
The release notes should definitely mention that this is a major upgrade. Depending on the database size, it could take a considerable time until the program is serving its purpose after migrating to the new database. And it requires additional space for the database storage volume, as the old one is kept for a month IIRC after migration.

I'd like to hear if the concerns from @hgy59 regarding upgrades have been addressed. My last questions in #6865 were not fully answered. Especially whether Synology could upgrade the package unattended, in that case we should block it by introducing a confirmation page in the upgrade process. There will be some angry users if they run into unexpected service downtime during the upgrade.

hgy59:
> Especially whether Synology could upgrade the package unattended,

Thirdparty packages do not support unattended updates.

acolomb:
Then I think a simple notice in the package release notes will be enough. But we should definitely try to draw people's attention to the possible downtime.

hgy59:
> The release notes should definitely mention that this is a major upgrade. Depending on the database size, it could take a considerable time until the program is serving its purpose after migrating to the new database. And it requires additional space for the database storage volume, as the old one is kept for a month IIRC after migration.

That is why I wanted to avoid the major update by package update and wanted to force the users to do the update and db migration by themself before (this is not mandatory, but I'm out of resources to validate and test without this restritcion).
The syncthing package still uses the internal updater and the main reason for a package update is to have a v2 package for new installations to avoid the need of db migration.

acolomb:
For those who have already migrated without a new SPK, there will be no problem. But forcing everyone to do it before they can upgrade the SPK is not the right way IMHO. So we're good to go with this approach, except for the clear statements in the package release notes.

mreid-tt:
hey @acolomb, @hgy59 — I’ve added an upgrade warning page to the wizard that appears only when upgrading from v1.x and is skipped if the upgrade was already performed internally; it clearly calls out the LevelDB → SQLite database migration on first launch, strongly warns users not to interrupt the process, and recommends backing up the configuration and database before proceeding.

spk/syncthing/Makefile
```diff
@@ -1,19 +1,19 @@
...
-UNSUPPORTED_ARCHS = $(PPC_ARCHS)
+# archs not supported by go (ARMv5 requires legacy toolchain no longer provided)
+UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS)
 
 MAINTAINER = acolomb
 DESCRIPTION = Automatically sync files via secure, distributed technology.
 DESCRIPTION_FRE = Synchronisation automatique de fichiers via une technologie sécurisée et distribuée.
 DISPLAY_NAME = Syncthing
-CHANGELOG = "Build syncthing v1.30.0 with go v1.23 for final ARMv5 compatibility. <br/><br/>REMARKS: New packages are rarely created because syncthing's automatic package updater is used. <br/>On ARMv5 archs (i.e. DS213air, DS213, DS413j, DS112, DS112+, DS212, DS212+, RS212, RS812, DS212j, DS112j, DS111, DS211, DS211+, DS411slim, DS411, RS411, DS211j and DS411j) the automatic update is disabled. Further updates will not work, and manual use of the internal updater will result in a broken installation."
+CHANGELOG = "1. Update Syncthing to v2.0.13 (built with Go 1.25). <br/>2. Drop legacy ARMv5 models that required the Go 1.23 compiler. <br/><br/>REMARKS: New packages are rarely created because Syncthing's in-app updater is used."
```
acolomb:
```suggestion
CHANGELOG = "1. Update Syncthing to v2.0.13 (built with Go 1.25). <br/>2. Drop legacy ARMv5 models that required the Go 1.23 compiler.<br/><br/>ATTENTION: This is a new major version of Syncthing, requiring a possibly time-consuming database migration after the upgrade. Please read the official release notes and be prepared for some service downtime.<br/><br/>REMARKS: New packages are rarely created because Syncthing's in-app updater is used."
```

spk/syncthing/src/wizard/upgrade_uifile.sh
```diff
@@ -0,0 +1,75 @@
...
+{
+    "step_title": "Important: Major Version Upgrade",
+    "items": [{
+        "desc": "<b style=\\"color: red\\">Warning: You are upgrading from Syncthing v1.x to v2.x.</b>"
+    },{
+        "desc": "<b>Database Migration:</b> The database backend has switched from LevelDB to SQLite. On first launch after the upgrade, there will be a migration process which can be lengthy for larger setups."
+    },{
+        "desc": "<b style=\\"color: red\\">Do not interrupt the migration process.</b>"
+    },{
+        "desc": "<b>Recommendation:</b> Backup your Syncthing configuration and database before proceeding."
```
acolomb: This isn't worth much without instructions on how to actually reach the data to back it up.  Is there some generic DSM mechanism for making a backup of the app's "var" folder?

As long as the migration is running, the old database is only read, not written.  When done, it leaves behind the old database folder named `index-v0.14.0.db-migrated`.  After two weeks, the new version deletes it automatically.  So in general, making a backup is not really vital.  If the migration fails, one can simply revert to the old binary.  And of course the database can always just be nuked completely, without major consequences if the device was in sync with others.  It will only take time, CPU and IO load to recreate.

spk/syncthing/src/wizard/upgrade_uifile.sh
```diff
@@ -0,0 +1,75 @@
...
+
+PAGE_MAJOR_UPGRADE=$(/bin/cat<<EOF
+{
+    "step_title": "Important: Major Version Upgrade",
+    "items": [{
+        "desc": "<b style=\\"color: red\\">Warning: You are upgrading from Syncthing v1.x to v2.x.</b>"
+    },{
+        "desc": "<b>Database Migration:</b> The database backend has switched from LevelDB to SQLite. On first launch after the upgrade, there will be a migration process which can be lengthy for larger setups."
+    },{
+        "desc": "<b style=\\"color: red\\">Do not interrupt the migration process.</b>"
```
acolomb: This is not very important, as the migration routine will simply pick up the process where it stopped.  As long as the process is terminated cleanly, as in no hard power-off.

spk/syncthing/src/wizard/upgrade_uifile.sh
mreid-tt: Removed

spk/syncthing/src/wizard/upgrade_uifile.sh
mreid-tt: Removed

mreid-tt:
hey @acolomb, thanks for your suggestions. These have been implemented.

spk/syncthing/Makefile
```diff
@@ -13,7 +13,7 @@ MAINTAINER = acolomb
 DESCRIPTION = Automatically sync files via secure, distributed technology.
 DESCRIPTION_FRE = Synchronisation automatique de fichiers via une technologie sécurisée et distribuée.
 DISPLAY_NAME = Syncthing
-CHANGELOG = "1. Update Syncthing to v2.0.13 (built with Go 1.25). <br/>2. Drop legacy ARMv5 models that required the Go 1.23 compiler. <br/><br/>REMARKS: New packages are rarely created because Syncthing's in-app updater is used."
+CHANGELOG = "1. Update Syncthing to v2.0.13 (built with Go 1.25). <br/>2. Drop legacy ARMv5 models that required the Go 1.23 compiler. <br/><br/>ATTENTION: This is a major version upgrade requiring a database migration on first launch. Please read the <a target=_blank href=https://github.com/syncthing/syncthing/releases/tag/v2.0.0>official release notes</a> and be prepared for some service downtime. <br/>REMARKS: New packages are rarely created because Syncthing's in-app updater is used."
```
acolomb:
```suggestion
CHANGELOG = "1. Update Syncthing to v2.0.13 (built with Go 1.25). <br/>2. Drop legacy ARMv5 models that required the Go 1.23 compiler. <br/><br/>ATTENTION: This is a major version upgrade requiring a database migration on first launch. Please read the <a target=_blank href=https://github.com/syncthing/syncthing/releases/tag/v2.0.13>official release notes</a> and be prepared for some service downtime. <br/>REMARKS: New packages are rarely created because Syncthing's in-app updater is used."
```

Since we're already using mentioning the latest release before, why not link to the corresponding notes.

PR #6932, opened 2026-01-27
myMPD: Update to v24.0.2

mreid-tt:
## Description

Updates myMPD to v24.0.2 and MPD to v0.23.17 with improved integration between the packages.

### myMPD v24.0.2

- Major version update with fuzzy search, improved MPD protocol handling, and UTF-8 support
- Lua scripting and FLAC metadata support enabled
- Installation wizard prompts for music folder to grant cover art/lyrics/metadata access
- Automatic MPD socket connection for music and playlist directory detection

### MPD v0.23.17

- Version bump with improved service setup
- Socket connection configured for myMPD/ympd integration

### ympd

- Fixed cross-compilation and DSM 7 compatibility
- Socket connection configured for MPD integration

### Supporting Changes

- cross/libmpdclient: Updated to match upstream v2.25
- cross/libid3tag: Added pkg-config file for Meson consumers
- cross/utf8proc: New dependency for UTF-8 support
- native/lua: New build dependency for Lua scripting

Fixes # 

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/cross/libid3tag/Makefile
+++ b/cross/libid3tag/Makefile
@@ -20,13 +20,15 @@ include ../../mk/spksrc.cross-cc.mk
 # Required for Meson-based consumers (e.g. MPD)
 libid3tag_pc:
 	@mkdir -p $(STAGING_INSTALL_PREFIX)/lib/pkgconfig
-	@echo 'prefix=$(INSTALL_PREFIX)' >  $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo 'exec_prefix=$${prefix}' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo 'libdir=$${exec_prefix}/lib' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo 'includedir=$${prefix}/include' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo '' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo 'Name: id3tag' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo 'Description: ID3 tag manipulation library' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo 'Version: $(PKG_VERS)' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo 'Libs: -L$${libdir} -lid3tag' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
-	@echo 'Cflags: -I$${includedir}' >> $(STAGING_INSTALL_PREFIX)/lib/pkgconfig/id3tag.pc
+	@( \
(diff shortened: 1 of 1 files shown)


hgy59:
The ympd start.sh is required not only for DSM 5.2 but also for any SRM version (all systems with ash shell).

mreid-tt:
> The ympd start.sh is required not only for DSM 5.2 but also for any SRM version (all systems with ash shell).

Noted. That said, this change doesn’t appear to have ever been published... if it had been, we would have seen that it fails to start on DSM 7.x. Rather than troubleshooting the wrapper script, I chose to remove it, especially since the package was never published for DSM 5.x either.

Regarding SRM, I’ve never been entirely clear on the criteria for publishing packages to devices without a GUI. I understand the case for terminal-based tools, but for a GUI-oriented package the benefit is less obvious. If this startup limitation is indeed correct, then the existing SRM 1.x packages in the repo would not function and should also be removed. In any case, I don’t plan to publish this GUI package for a non-GUI platform.

In my opinion, if DSM 5.x and SRM 1.x require special handling of `SERVICE_COMMAND` launch parameters, those workarounds should live in the framework rather than being implemented ad hoc in individual packages.

hgy59:
> In my opinion, if DSM 5.x and SRM 1.x require special handling of `SERVICE_COMMAND` launch parameters, those workarounds should live in the framework rather than being implemented ad hoc in individual packages.

Yes, this should be handled in the generic service support.
but if you just delete the script, my former analysis and solution is lost.
The current solution here would be to have different service-setup.sh for SRM and DRM 5.2

mreid-tt:
> The current solution here would be to have different service-setup.sh for SRM and DRM 5.2

Fair enough. As I don't have DSM 5.x anymore or SRM 1.x box, can you confirm that this script actually works on those platforms?

hgy59:
> > The current solution here would be to have different service-setup.sh for SRM and DRM 5.2
> 
> Fair enough. As I don't have DSM 5.x anymore or SRM 1.x box, can you confirm that this script actually works on those platforms?

Yes, version 1.3.0-1 runs on a Synology Router RT2600ac with SRM 1.3.1-9346 Update 14.

My DS210+ (ppc853x-5.2) was never supported (libmpdclient requires meson/rust build system that does no support OLD_PPC_ARCHS).

mreid-tt:
> Yes, version 1.3.0-1 runs on a Synology Router RT2600ac with SRM 1.3.1-9346 Update 14.

This is not proof since the v1.3.0-1 revision does not have the wrapper script in it. In fact if you download the [1.2 ipq806x](https://packages.synocommunity.com/ympd/1/ympd.v1.f7742%5Bnorthstarplus-ipq806x-dakota%5D.spk) file from the repo and examine the `scripts/service-setup` file, you will see this:

```
### Package specific variables and functions
### ----------------------------------------


SERVICE_COMMAND="${SYNOPKG_PKGDEST}/bin/ympd -w ${SERVICE_PORT}"
SVC_BACKGROUND=y
SVC_WRITE_PID=y
```

Recall, the changes in https://github.com/SynoCommunity/spksrc/pull/5652 were never published, thus the reason for my request for validation if this wrapper script is even needed.

mreid-tt:
> Yes, this should be handled in the generic service support.
> but if you just delete the script, my former analysis and solution is lost.

So that your research isn't lost, I had an AI translate the findings into a wiki page: https://github.com/SynoCommunity/spksrc/wiki/Ash-Shell-Background-Process-PID-Tracking-Issue

PR #6914, opened 2026-01-21
Toolchain misc fixes

th0ma7:
## Description

Toolchain enhancements:
- [x] Use status cookie to avoid regenerating `tc_vars.*mk` files
- [x] Better handle logging messages
- [x] Properly handle logging in the `toolchain/syno-$(TC_NAME|ARCH)-$(TC_VERS)` context
- [x] Have `spksrc.cross-cc.mk` call toolchain generation instead of clunky `spksrc.cross-env.mk`
- [x] Document each toolchain mk files
- [x] Move all `toolchain/syno-*` no longer available to `toolchain/archive`
- [x] Migrate DSM-6.2.4 to be using our github abandoned source URL
- [x] Re-organize `spksrc.toolchain/tc-version.mk` to define per-DSM-version download URL
- [x] Create a `spksrc.toolchain/tc-base.mk` and `tc-url.mk` to separate from compiler specific flags.
- [x] Re-organize toolchain internal sub-mk files such as:
```
mk/spksrc.toolchain.mk
mk/spksrc.toolchain
├── tc-base.mk
├── tc-flags.mk
├── tc-normalize.mk
├── tc-rust.mk
├── tc-url.mk
├── tc_vars.mk
└── tc-versions.mk
```

Fixes
- [x] Avoid multiple re-iteration of `tc_var*` toolchain generation
- [x] Avoid generation of `$(WORK_DIR)/tc_var*` in dependencies folders
- [x] Fix lgging at spk level that was missing

Other
- [x] Enhanced significantly documentation of `cross-cc|env.mk`
- [x] Create a `$(call LOG_WRAPPED,<recipe>)` default logging recipe

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes <s>small</s> framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #6936, opened 2026-01-29
Borg: Update borgmatic to 2.1.2

### LINKED ISSUE #6910: [Package Update] Update Borgmatic (part of the Borg package) to Latest (currently 2.1.0)
### Software Name

Borg

### Brief Description

When you're able, please update Borgmatic (within the Borg packaged) to the latest (currently [2.1.0](https://github.com/borgmatic-collective/borgmatic/releases)).

And if there are new updates to [Borg](https://www.borgbackup.org/releases/) by the time you get around to this then of course update that as well.

Thank you!

### Website

https://torsion.org/borgmatic/

### Documentation

_No response_

https://torsion.org/borgmatic/how-to/develop-on-borgmatic/

### Source code

https://projects.torsion.org/borgmatic-collective/borgmatic

### OpenSource License

GPLv3

### Other License

_No response_

mreid-tt:
## Description

- Update borgmatic to v2.1.2
- Refresh Python dependencies
- Build for ARMv5 and ARMv7L using Python 3.11

Fixes #6910
Reverts #6781

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/spk/borgbackup/Makefile
+++ b/spk/borgbackup/Makefile
@@ -43,8 +43,16 @@ DEPENDS = cross/acl cross/attr cross/fuse cross/libb2 cross/lz4 cross/openssl3 c
 # # The script will create three files (pure, cross, and combined) for the Borg deployment
 # # with Python 3.12. Use the pure and cross requirement files as the base for your PR.
 
-# Default wheels for most architectures
+# Wheels: pure-python requirements (all architectures)
 WHEELS = src/requirements-pure.txt
+
+# Cross-compiled requirements vary by architecture
+# ARMv5 uses rpds_py pinned to 0.21.0 for Cargo.lock v3 compatibility with Rust 1.77.2
+ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS)),$(ARCH))
+WHEELS += src/requirements-crossenv-armv5.txt
+else
+WHEELS += src/requirements-crossenv.txt
+endif
 
 # Force building pure-python wheels (issue #5204)
 WHEELS_PURE_PYTHON_PACKAGING_ENABLE = 1
@@ -66,14 +74,6 @@ SPK_USR_LOCAL_LINKS = bin:env/bin/borg bin:env/bin/borgmatic bin:env/bin/emborg
 
 include ../../mk/spksrc.python.mk
 
-# ARMv5: Use separate requirements with rpds_py pinned to 0.20.0
-# (Cargo.lock v3 compatibility with Rust 1.77.2)
-ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS)),$(ARCH))
-WHEELS += src/requirements-crossenv-armv5.txt
-else
-WHEELS += src/requirements-crossenv.txt
-endif
-
 # [borgbackup]
 # Use OpenSSL path as defined by either
 # spksrc.wheel-env.mk or spksrc.python.mk
@@ -85,23 +85,29 @@ ENV += BORG_LIBZSTD_PREFIX="$(STAGING_INSTALL_PREFIX)"
 ENV += BORG_OPENSSL_PREFIX="$(OPENSSL_STAGING_PREFIX)"
 
 # [borgbackup]
+# GCC < 5.0 (DSM 6.2.4) requires gnu99 for Cython-generated code.
+# C11 (the original setting) is not fully supported by GCC 4.9.x, and Cython
+# uses GNU extensions (e.g., typeof, statement expressions) in its fallback paths.
 ifeq ($(call version_lt, $(TC_GCC), 5.0),1)
 WHEELS_CFLAGS += [borgbackup] -std=gnu99
 endif
 
 # [llfuse]
-# llfuse uses Cython which generates C11 atomics code that fails
-# due to ATOMIC_INT_LOCK_FREE not being a preprocessor constant in older GCC.
+# GCC < 5.0 requires gnu99 for Cython-generated code (see borgbackup above).
+# Requires CYTHON_ATOMICS=0 because ATOMIC_INT_LOCK_FREE is not a
+# preprocessor constant in GCC < 5.0.
 ifeq ($(call version_lt, $(TC_GCC), 5.0),1)
 WHEELS_CFLAGS += [llfuse] -std=gnu99 -DCYTHON_ATOMICS=0
 endif
 
 # [msgpack]
+# GCC < 5.0 requires gnu99 for Cython-generated code (see borgbackup above).
 ifeq ($(call version_lt, $(TC_GCC), 5.0),1)
 WHEELS_CFLAGS += [msgpack] -std=gnu99
 endif
 
 # [ruamel_yaml_clib]
+# GCC < 5.0 requires gnu99 for Cython-generated code (see borgbackup above).
 ifeq ($(call version_lt, $(TC_GCC), 5.0),1)
 WHEELS_CFLAGS += [ruamel_yaml_clib] -std=gnu99
 endif
--- a/spk/borgbackup/src/requirements-pure.txt
+++ b/spk/borgbackup/src/requirements-pure.txt
@@ -1,4 +1,4 @@
-# Generated requirements for borgbackup with Python 3.12
+# Pure-python requirements for borgbackup (Python 3.11/3.12)
 
 appdirs==1.4.4
 arrow==1.4.0
--- a/spk/borgbackup/src/requirements-crossenv-armv5.txt
+++ b/spk/borgbackup/src/requirements-crossenv-armv5.txt
@@ -1,4 +1,4 @@
-# ARMv5-specific crossenv requirements
+# ARMv5-specific cross-compiled requirements (Python 3.11)
 # rpds_py pinned to 0.21.0 for Cargo.lock v3 compatibility with Rust 1.77.2
 
 borgbackup==1.4.3
--- a/spk/borgbackup/src/requirements-crossenv.txt
+++ b/spk/borgbackup/src/requirements-crossenv.txt
@@ -1,4 +1,4 @@
-# Generated requirements for borgbackup with Python 3.12
+# Cross-compiled requirements for borgbackup (Python 3.11/3.12)
 
 borgbackup==1.4.3
 charset_normalizer==3.4.4


mreid-tt:
Hey @th0ma7, I was trying to combine a python311-based build for older archs (based on https://github.com/SynoCommunity/spksrc/pull/6845#issuecomment-3649092079) using a code block like this:
```
# Build for ARMv5 and ARMv7L using Python 3.11. All other use Python 3.12
ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))
PYTHON_PACKAGE = python311
SERVICE_SETUP = src/service-setup_py311.sh
else
PYTHON_PACKAGE = python312
SERVICE_SETUP = src/service-setup_py312.sh
endif

SPK_DEPENDS = "$(PYTHON_PACKAGE)"
```
But then in the log we see this:

```
make: Entering directory '/home/runner/work/spksrc/spksrc/native/python312'
===>  Downloading files for python312
===>    File Python-3.12.12.tar.xz already downloaded
===>  Verifying files for python312
===>    Checking sha1sum of file Python-3.12.12.tar.xz
===>    Checking sha256sum of file Python-3.12.12.tar.xz
===>    Checking md5sum of file Python-3.12.12.tar.xz
/home/runner/work/spksrc/spksrc/native/python312/../../distrib/Python-3.12.12.tar.xz
make: Leaving directory '/home/runner/work/spksrc/spksrc/native/python312'
  -> python312: download-all then checksum
make: *** python312: No such file or directory.  Stop.
Error: Download/checksum failed for python312
Error: Error on line 69 while processing python312
Error: Process completed with exit code 1.
```
This is I believe because it misinterprets the packages to be built at the start like this:
```
===> Downloading packages:  cross/acl cross/attr cross/berkeleydb cross/bzip2 cross/fuse cross/gdbm cross/libb2 cross/libexpat cross/libffi cross/libuuid cross/lz4 cross/ncursesw cross/openssl3 cross/python311 cross/python312 cross/readline cross/sqlite cross/xxhash cross/xz cross/zlib cross/zstd native/python311 native/python312 python312 
  -> cross/acl: download-all then checksum
```
As compared to other builds, there is an extra package `python312` which isn't present in the other builds. Is my approach wrong or is there a limitation in the framework?

th0ma7:
I'm assuming you've got an include spksrc.common.mk right before your code block or that your code block is after the include for apksrc.python.mk.

It most probably is a bug as all python are present in that line + an extra python312 with no cross/native prefix.

One other thing may be that, if i recall correctly, python.mk defines unsupported archs based on  python being selected thus that code would need to be rethought.

Lastly, wonder if the behavior is the same under my toolchain pr.  In theory it should.

mreid-tt:
> I'm assuming you've got an include spksrc.common.mk right before your code block or that your code block is after the include for apksrc.python.mk.

Yes, you are correct, this block is after the import of `spksrc.python.mk`.

> It most probably is a bug as all python are present in that line + an extra python312 with no cross/native prefix.
> 
> One other thing may be that, if i recall correctly, python.mk defines unsupported archs based on python being selected thus that code would need to be rethought.

Hmm, okay. Open to suggestions, otherwise I'll probably merge this branch as-is and abandon the hybrid approach.

> Lastly, wonder if the behavior is the same under my toolchain pr. In theory it should.

Yes, I did a merge commit of your PR #6914 onto my test branch and it had the same outcome.

th0ma7:
I can look at fixing things up to allow hybrid approach upon my return if you can wait, or for a subsequent release of borg.

th0ma7:
For your consideration, you could add a post install recipe to do a sed to swap the python version in the service setup script thus avoiding to maintain two of them.

In the longer term this could be automated to avoid changing it all together...

mreid-tt:
hey @th0ma7, I believe this package is now complete. Do let me know if you have any other notes before I merge.

spk/borgbackup/Makefile
```diff
@@ -32,7 +43,9 @@ DEPENDS = cross/acl cross/attr cross/fuse cross/libb2 cross/lz4 cross/openssl3 c
 # # The script will create three files (pure, cross, and combined) for the Borg deployment
 # # with Python 3.12. Use the pure and cross requirement files as the base for your PR.
 
-WHEELS = src/requirements-crossenv.txt src/requirements-pure.txt
+# Default wheels for most architectures
+WHEELS_CROSSENV = src/requirements-crossenv.txt
+WHEELS = $(WHEELS_CROSSENV) src/requirements-pure.txt
```
th0ma7: `WHEELS_CROSSENV` does not exists, this should be:
```
WHEELS  = src/requirements-crossenv.txt
WHEELS += src/requirements-pure.txt
```

spk/borgbackup/src/requirements-pure.txt
```diff
@@ -1,25 +1,25 @@
-# Generated using the requirements script on December 11, 2025
+# Generated requirements for borgbackup with Python 3.12
```
th0ma7: cosmetic but this is for both Python 3.11/3.12

spk/borgbackup/Makefile
```diff
@@ -42,17 +55,24 @@ DESCRIPTION = BorgBackup \(short: Borg\) is a deduplicating backup program. Opti
...
+POST_SERVICE_TARGET = borgbackup_extra_install
 
 SPK_USR_LOCAL_LINKS = bin:env/bin/borg bin:env/bin/borgmatic bin:env/bin/emborg
 
 include ../../mk/spksrc.python.mk
 
+# ARMv5: Use separate requirements with rpds_py pinned to 0.20.0
+# (Cargo.lock v3 compatibility with Rust 1.77.2)
+ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS)),$(ARCH))
+WHEELS_CROSSENV = src/requirements-crossenv-armv5.txt
```
th0ma7: similar here, this should be: `WHEELS += src/requirements-crossenv-armv5.txt`

spk/borgbackup/Makefile
```diff
@@ -78,5 +102,10 @@ endif
 
 # [ruamel_yaml_clib]
 ifeq ($(call version_lt, $(TC_GCC), 5.0),1)
-WHEELS_CFLAGS += [ruamel_yaml_clib] -std=c11
+WHEELS_CFLAGS += [ruamel_yaml_clib] -std=gnu99
```
th0ma7, line 112: Unsure why switching from `-std=c11` to `-std=gnu99` unless this is to be compatible with armv5 gcc verstion 4.6.4 where a unique compile option allows to rule them all.

Looking further indeed `c11` standard only appeared after that... I recall experimental was `-std=c1x` (I used to compile gcc on HP-UX at the time).

Anyhow, if that works for all of them, fine by me but I'd add a comment related to that as to why gnu99 instead of c11 OR would consider using (I personally prefer the comment.. simpler):
```
ifeq ($(call version_lt, $(TC_GCC), 4.8),1)
WHEELS_CFLAGS += [ruamel_yaml_clib] -std=gnu99  ### noting c99 might work without the extra gnu extensions
else ifeq ($(call version_lt, $(TC_GCC), 5.0),1)
WHEELS_CFLAGS += [ruamel_yaml_clib] -std=c11
endif

File: spk/borgbackup/Makefile (hunk 3 above)
Comment (th0ma7) [reply spk/borgbackup/Makefile]: I just noticed what you did!  I suggest you do this differently such as:
```
WHEELS = src/requirements-pure.txt
WHEELS += src/requirements-crossenv-armv5.txt
else
WHEELS += src/requirements-crossenv.txt
endif
```

spk/borgbackup/Makefile
mreid-tt: `WHEELS_CROSSENV` in this context is a variable which could be set to another value later if needed. It was not meant to populate an existing keyword from the framework.

spk/borgbackup/Makefile
mreid-tt: `WHEELS_CROSSENV` in this context is a variable which could be set to another value later if needed. It was not meant to populate an existing keyword from the framework.

spk/borgbackup/Makefile
mreid-tt: The move from `-std=c11` was because this function does not exist with the ARMv5 GCC. I will try with the simpler `-std=c99` as suggested to test. Overall, I just wanted to have a simpler makefile to cover older GCCs without a bunch of conditional statements. Will look to add comments as suggested.

th0ma7:
I've tested the build for armv5 and armv7L and with my changes to the `WHEELS` variable it did worked all right.

mreid-tt:
> I've tested the build for armv5 and armv7L and with my changes to the `WHEELS` variable it did worked all right.

Thanks for testing. I'd prefer to consolidate this section as a single block but I'll fix that after I test the `-std=c99` flag.

spk/borgbackup/Makefile
th0ma7: Related to `WHEELS_CROSSENV` issues where:
1. there are similar variable names in the context of creating a crossenv in the code.  I'd rather not have this set as an example to follow as it is not an official variable name, further it could blur things up with other internal framework variable names.
2. `WHEELS_CROSSENV` is set once by default for all, then (re)set once more in the context of ARMv5 - this can lead to weird behavior later one.  While direct assignment from `ifeq` then `else` only using `WHEELS` with += avoids all of that.

spk/borgbackup/Makefile
th0ma7: yeah, that's why i gave it a shot to confirm its behavior.  Works well - I suggest simply adding a comment related to that.

th0ma7:
> > I've tested the build for armv5 and armv7L and with my changes to the `WHEELS` variable it did worked all right.
> 
> Thanks for testing. I'd prefer to consolidate this section as a single block but I'll fix that after I test the `-std=c99` flag.

See my additional comment on that (e.g. `WHEELS_CROSSENV` vs `WHEELS`)

mreid-tt:
hey @th0ma7, all changes made (sorry for the force pushes). It should be ready to merge now.

spk/borgbackup/Makefile
mreid-tt: Done

spk/borgbackup/src/requirements-pure.txt
mreid-tt: Done

PR #6947, opened 2026-02-04
Framework: Auto-orchestrate multi-arch download, checksum, and digests

mreid-tt:
## Description

This PR simplifies working with packages that download architecture-specific binaries (e.g., `cross/btop`). Instead of requiring separate `-all` targets, the standard `download`, `checksum`, and `digests` targets now automatically detect multi-arch packages and orchestrate per-architecture execution.

Credit to @th0ma7 for the original concept and approach.

### Problem

Packages using `PKG_DIST_ARCH_LIST` to define multiple architecture-specific downloads (like prebuilt binaries) required:

- Manual use of `download-all` to fetch all architecture variants
- No equivalent `checksum-all` to validate cached files across architectures
- CI failures when the distrib cache contained stale files for non-default architectures

### Solution

Unified orchestration: When `PKG_DIST_ARCH_LIST` contains 2+ entries, the framework automatically iterates over each architecture:
```
make -C cross/btop download   # Downloads all 4 arch-specific files
make -C cross/btop checksum   # Verifies all 4 arch-specific files
make -C cross/btop digests    # Generates checksums for all 4 files
```
Architecture-specific cookies: Each architecture gets its own completion cookie (e.g., `.btop-x86_64-linux-musl-download_done`) to track progress independently and avoid redundant operations.

Backward compatible: Single-arch packages continue to work unchanged with generic cookies.

### Changes

- **mk/spksrc.download.mk** - Auto-orchestrates when `PKG_DIST_ARCH_LIST` has 2+ entries; removes deprecated `download-all`
- **mk/spksrc.checksum.mk** - Mirrors download orchestration pattern; adds `[arch]` suffix to status messages
- **mk/spksrc.generate-digests.mk** - Simplified to use orchestration pattern; generates combined digests file
- **.github/actions/download.sh** - Uses `download` (auto-orchestrating) instead of removed `download-all`

### Testing

Verified with:

- **cross/btop** (multi-arch: 4 architectures) - download, checksum, digests all work correctly
- **cross/zlib** (single-arch) - regression test passes, no change in behavior

Fixes # 

## Checklist

### Type of change

- [x] Bug fix
- [x] Includes ~small~ framework changes

### DIFF
--- a/mk/spksrc.checksum.mk
+++ b/mk/spksrc.checksum.mk
@@ -67,7 +67,11 @@ endif
 .PHONY: $(PRE_CHECKSUM_TARGET) $(CHECKSUM_TARGET) $(POST_CHECKSUM_TARGET)
 
 checksum_msg:
+ifneq ($(strip $(PKG_DIST_ARCH)),)
+	@$(MSG) "Verifying files for $(NAME) [$(PKG_DIST_ARCH)]"
+else
 	@$(MSG) "Verifying files for $(NAME)"
+endif
 
 pre_checksum_target: checksum_msg
 


th0ma7:
@mreid-tt while looking at the code I thought maybe there's an easier way to workaround that by avoiding to add yet another `make <something>-all` ... What do you think of the following aiming to always use `make checksum` ... A similar code could be used for `download` and `digests` to remove the `*-all` calls... 

What do you think?

Note: made it easy to revert my single commit as needed...

mreid-tt:
hey @th0ma7, this is certainly a cleaner and more elegant approach. Love it! Tested in the CI and the output meets my objective for this PR.

th0ma7:
Then would it make more sense to change download and digest all together?

mreid-tt:
hey @th0ma7, I had the AI look at your improvement and applied it to the download, include backward compatibility and simplify code no longer needed in the digests. Seems to work with my testing but would appreciate a review.

th0ma7:
@mreid-tt I did a few additional adjustments, looks bigger than it is in reality:
1. for `download` and `checksum` it now generate per-arch status cookie providing finer control and avoiding to fail when directly invoking with a single arch using a multi-arch call
2. reworked the logic so when using single arch (classic method) the digests recipe is the usual such as `digests: download` although in multi-arch it avoids pre-calling `download` but rather calls it within its for loop.  This makes the log output being per-arch ordered instead of per-call ordered.

I've now also updated the PR description a bit and enhanced the testing call method you initially provided.

From my point of view this PR is now ready to go - I'll let you re-test things up to confirm and merge, good job :rocket:

th0ma7 (approved):

PR #3866, opened 2020-01-25
Fix shairport

hgy59:
_Motivation:_  Shairport-sync DSM6 compatibiliy is asked a long time ago
_Linked issues:_  #3040 (with links to other issues), fixes #3922

### Checklist
- [x] Build rule `all-supported` completed successfully (except for ppc853x-5.2)
- [ ] Package upgrade completed successfully (hard to test - old package does not install, may be test with upgrade 2.3.11 => 2.8.5)
- [x] New installation of package completed successfully (except DSM <6)
- [ ] Fix installation on DSM<6
- [x] Update shairport to version 2.8.5 (#2529) or later

### Remarks
- update busybox to enable --large-opts (yes I like readable parameters)
- busybox removed from dependencies and use daemonize of shairport

PR #6950, opened 2026-02-04
Shairport Sync: Update to v4.3.7 with AirPlay 2 support

### LINKED ISSUE #3040: DS418j can't find Shairport Sync in Package Center
Hi,
i migrated to new DS418j from DS216j
Unfortunately i can not find  Shairport Sync in Package Center.
Beta is enabled in settings.
Could anyone help, please?

DS418j
CPU Realtek RTD1293 SoC
DSM 6.1.4-15217 Update1

### LINKED ISSUE #3922: Hello, I have compiled the SPK successfully, and installed the sharport-sync-armada37xx-6.2-2.3.11-2.spk. But when I started, I told me that the startup failed. What can I do to solve it??

mreid-tt:
## Description

Major update from v2.3.11 to v4.3.7 with full AirPlay 2 support. Enables multi-room audio, Siri integration, and improved streaming from Apple devices.

**Note:** DSM 6 only - DSM 7 doesn't allow the root access required for ALSA audio devices.

### What's New for Users

- **AirPlay 2** - Multi-room audio, Siri voice control, improved reliability
- **Setup Wizard** - Easy configuration of audio device, mixer, and AirPlay name
- **Metadata Support** - Track info and album art available to other apps
- **High-Quality Audio** - soxr resampling for better sound quality

### Requirements

- DSM 6.x (not compatible with DSM 7)
- AudioStation installed (provides USB audio drivers)
- USB sound card or speakers connected to NAS

### Technical Changes

- Add `native/xxd` - Build tool for AirPlay 2 configuration generation
- Add `cross/nqptp` - NQPTP daemon for AirPlay 2 timing
- Add `cross/ffmpeg6-audio` - Audio-only FFmpeg for AAC decoding
- Add `cross/avahi-headers` - Compile-time avahi (uses DSM system libs at runtime)
- Modernize package structure (service-setup.sh replaces legacy scripts)
- Dynamic wizard with automatic sound card and mixer detection

Fixes #3922
Fixes #3040
Supersedes #3866

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/cross/ffmpeg6-audio/Makefile
+++ b/cross/ffmpeg6-audio/Makefile
@@ -67,10 +67,13 @@ include ../../mk/spksrc.common.mk
 # Minimal dependencies - no video-related packages
 DEPENDS = cross/zlib
 
-# Disable assembly for all architectures to avoid nasm dependency.
+# Disable assembly for most architectures to avoid nasm dependency.
 # This is acceptable for an audio-only build where performance impact is minimal.
-# ARMv8 could use native ASM but we disable it for simplicity.
+# However, ARMv8 MUST have ASM/NEON enabled otherwise FFmpeg's init code
+# references undefined NEON symbols (FFmpeg assumes NEON is always available on ARMv8).
+ifneq ($(findstring $(ARCH),$(ARMv8_ARCHS)),$(ARCH))
 CONFIGURE_ARGS += --disable-asm
+endif
 
 # Architecture-specific settings
 ifeq ($(findstring $(ARCH),$(x64_ARCHS)),$(ARCH))
@@ -79,6 +82,8 @@ endif
 
 ifeq ($(findstring $(ARCH),$(ARMv8_ARCHS)),$(ARCH))
 CONFIGURE_ARGS += --arch=arm64
+# Enable NEON on ARMv8 - required to compile the assembly that the init code references
+CONFIGURE_ARGS += --enable-neon --enable-thumb
 endif
 
 ifeq ($(findstring $(ARCH),$(ARMv7_ARCHS) $(ARMv7L_ARCHS)),$(ARCH))


mreid-tt:
hey @th0ma7, @hgy59 — I’ve built on the earlier work in #3866 and prepared this update to Shairport Sync. It adds AirPlay 2 support and introduces user-friendly wizard steps to select the audio device and set the AirPlay target name. I’ve tested it using a small USB DAC, and it’s working well — including AirPlay 2 multi-room playback.

To keep the package size reasonable and avoid compatibility exclusions on older NAS devices, I’ve included a lightweight audio-only build of `ffmpeg` instead of the larger `ffmpeg7` package. One limitation is that this currently works only on DSM 6, as two components require `root` privileges to function properly.

I’m only able to test on x64 platforms, so I’d really appreciate help testing on other architectures. Even without a physical sound card, I’ve included a dummy audio device option and verbose logging can be enabled in the config file, which should allow others to assist with testing regardless of hardware.

th0ma7:
A few hints, you should use the meta spksrc.ffmpeg.mk instead spksrc.spk.mk.  this will allow you to link directly against ffmpeg7 target destination lib directory without needing to bundle ffmpeg within your package, reducing the size considerably.  That's being done for comskip, chromaprint and tvheadend.  Similarly to python on your development system you build spk/ffmpeg7 once, then it gets reused by your new package builds.

Also related to dsm7 i would suggest a clear explanation using the wizard with an url to the Wiki (eventually its reworked version presumably) on what steps needed to be done post install to making it work...  Question being, what's needed on dsm7 to work aside root?  Otherwise it sadly would be meaningless to provide that package if unable to making it run on dsm7.

th0ma7:
Also, i do have a test armv7 dsm7 system that could be used for testing.

th0ma7:
Adding (sorry for the noise), there might be something similar to the videodriver group allowing access to the audio devices, and thus alsa... Food for thoughts.

mreid-tt:
Thanks for the suggestions @th0ma7!

**Regarding `spksrc.ffmpeg.mk`:** I considered using the shared `ffmpeg7` approach, but opted for the bundled `ffmpeg6-audio` to maintain broader architecture support. The `ffmpeg7` package excludes `ARMv5` (88f6281), `ARMv7L` (hi3535), and `OLD_PPC` architectures, whereas the audio-only build supports all platforms. The ~8MB size increase seemed acceptable to avoid dropping support for older NAS devices. Happy to revisit if the smaller package footprint is preferred.

**Regarding DSM 7 support:** Unfortunately, DSM 7 presents multiple challenges:

1. **No USB audio drivers** - DSM 7 doesn't include USB audio kernel modules (unlike DSM 6 via AudioStation). I previously attempted to create a `synokernel-usbaudio` package ([branch link](https://github.com/mreid-tt/spksrc/tree/synokernel-usbaudio-1.0)) to compile the necessary ALSA/USB modules, but testing was unsuccessful.

2. **Manual module loading** - Even if the kernel modules worked, based on the `synocli-kernel` pattern, they would need to be manually loaded after each DSM reboot (no automatic loading mechanism exists for third-party kernel modules).

3. **Root requirement for NQPTP** - The NQPTP daemon (required for AirPlay 2 timing) needs to bind to privileged UDP ports 319/320, which requires root access that DSM 7 doesn't permit for third-party packages.

Given these constraints, DSM 6 seems to be the practical target for now. If there's interest in pursuing DSM 7 support further, I'd welcome collaboration on the kernel module approach or any insights into audio group permissions that might help.

Would appreciate your help testing on the armv7 DSM 7 system if you want to explore the kernel module path!

th0ma7:
Not that obvious... I could give it a try at building kernel modules but that wouldn't solve the root issue - possibly using a manual startup with sudo...  As for ffmpeg, i get where you're coming from but, armv7l is sooo niche, and armv5 is sooo old that i don't believe it makes sense to support yet another ffmpeg build strictly for these two.

Honestly, unless there's a solution for dsm7 i would not release this package.  Dsm8 will be out soon'ish which means less and less cycles will be spent on dsm6 and older.

But as i said, i can give it a shot at building modules if you can provide me with what's needed.

mreid-tt:
Realistically, unless we see a breakthrough with the DSM 7 limitations mentioned above, I don’t expect to continue supporting this package beyond this release. My primary goal here was to close out the long-standing requests for a DSM 6 version and to clean up older package structures in the repo.

Long term, I’d prefer that the repo only contain packages compatible with the DSM versions we actively support, and that we remove unsupported ones from both the codebase and the website.

That said, based on my local testing, this package is actually quite a solid solution. It might be worth exploring whether Synology would consider publishing it themselves, or whether we could lobby for additional resource handlers (e.g., for kernel modules or privileged ports) to make this viable on newer DSM versions. I’m just not sure what level of engagement we currently have with them — my past attempts to reach developer support haven’t received a response.

PR #6984, opened 2026-02-22
Refactor mk/ layout: reorganize into functional submodules

th0ma7:
## Description
This PR reorganizes the `mk/` directory into clearly separated functional modules.
There are no intended functional changes, but the structure has been redesigned to:
- Clarify include boundaries
- Enforce separation between reusable definitions and rule entry points
- Clean-up and prepare the codebase for upcoming DSM8-related changes

Fixes : n/a

## New `mk` directory structure
```
$ tree -d
├── spksrc.build
├── spksrc.common
├── spksrc.cross
├── spksrc.kernel
├── spksrc.native
├── spksrc.rules
├── spksrc.service
├── spksrc.spk
├── spksrc.toolchain
├── spksrc.toolkit
├── spksrc.wheel
└── wizard
```
This cleans-up the root of the `mk` directory leaving only main entry-points used under `cross`, `native`, `spk`, `toolchain`, `toolkit` and `kernel`.

## Highlights
`spksrc.common.mk` and files under `spksrc.common/` now contain:
- Variables
- Macros
- Pure definitions
- Directory layout
- Logging helpers

Rule definitions are now grouped under `spksrc.rules/` and only included through main entry points such as:
- `spksrc.spk.mk` (indirectly via `spksrc.rules/supported.mk`)
- `spksrc.cross-cc.mk` (indirectly via `spksrc.rules/supported.mk`)
- `spksrc.native-cc.mk`
- `spksrc.kernel.mk`
- `spksrc.toolkit.mk`
- `spksrc.toolchain.mk`

By isolating them, we:
- Avoid accidental rule redefinitions
- Make the include graph explicit
- Clarify which files are actual build entry points

### Why this matters
The previous flat layout mixed of entry points, variables, macros, common rules definitions, etc.  This made it harder to reason about where a rule originated and whether a file could be safely re-included.

The new structure enables a “divide and conquer” approach:
- It is easier to trace where an include comes from
- Responsibilities are grouped by concern
- Cross-cutting logic is easier to identify
- Subsystems are now structurally isolated
Leaving only core entry-points into the root of the `mk` directory, which avoids confution greatly.

### DSM8 Preparation
This restructuring is also hope to help when initiating groundwork for DSM8 compatibility work expected over the next year.  It should becomes significantly easier to target and modify only the relevant subsystem without unintended side effects and reduces risk for large upcoming changes.

### Functional Impact
- No intended behavioral changes
- No rule semantics modified
- Entry points preserved for compatibility

This is a structural refactor only.

[...]

PR #6985, opened 2026-02-22
dnscrypt-proxy: Update to v2.1.15

mreid-tt:
## Description

Updates dnscrypt-proxy from v2.0.44 to v2.1.15.

### User-Facing Changes

- **New Monitoring Dashboard** - Replaces config file editor with real-time DNS query statistics (port 8153, accessible via "Open" button)
- **Improved Setup Wizard** - 2-step wizard: DNS settings, then dashboard credentials
- **Dynamic DHCP Coexistence** - Automatically switches between ports 53/10053 when DHCP server is enabled/disabled (no reinstall needed)
- **Auto Server Selection** - Leave servers blank for automatic fastest-server selection (previously used hardcoded list)
- **Upgrade Wizard** - Users upgrading from v2.0.x see a one-time wizard to configure dashboard credentials

### Bug Fixes

- Fixed undefined function `disable_dhcpd_dns_port` in uninstall
- Fixed insecure permissions on var directory (`0777` → `0755`)
- Fixed duplicate cron job entries on SRM
- Fixed `require_dnssec=true` and `netprobe_timeout=2` settings that caused DNS resolution issues
- Updated URLs from deprecated `jedisct1` repo to current `DNSCrypt` organization

### Other Changes

- Limited to DSM 6.x (`REQUIRED_MAX_DSM = 6.999`)

Fixes # 

## Checklist

### Type of change

- [x] Package update

### DIFF
--- a/spk/dnscrypt-proxy/src/service-setup.sh
+++ b/spk/dnscrypt-proxy/src/service-setup.sh
@@ -18,10 +18,11 @@ CFG_FILE="${SYNOPKG_PKGVAR}/dnscrypt-proxy.toml"
 MONITORING_PORT="8153"  # Web UI for monitoring DNS queries
 BACKUP_PORT="10053"     # Used when DHCP server occupies port 53
 
-# Detect OS type: DSM (DiskStation) vs SRM (Router)
-# SRM devices have model names in uname output
-case "$(uname -a)" in
-    *rt1900ac*|*rt2600ac*|*mr2200ac*) OS="srm" ;;
+# Detect OS type: DSM (DiskStation Manager) vs SRM (Synology Router Manager)
+# SRM uses productversion < 3.0 (e.g., 1.2.x), DSM uses >= 6.0
+productversion=$(grep '^productversion=' /etc.defaults/VERSION | cut -d= -f2 | tr -d '"')
+case "${productversion%%.*}" in
+    [012]) OS="srm" ;;
     *) OS="dsm" ;;
 esac
 


mreid-tt:
hey @publicarray — I’ve done a major refactor of this app and tested both upgrades and new functionality. The previous version had limited usability due to outdated references and the codebase having moved to a new repository. When you have a few spare cycles, I’d appreciate it if you could take a look and review.

th0ma7:
@mreid-tt associated with https://github.com/SynoCommunity/spksrc/pull/6984#issuecomment-3940887630, the AI is right, this package is indeed affected.  Although the way its being built is wrong, the `go build` should never occur at the spk level but rather at the cross level.

Can you move out the `go build` call from the spk down to the cross?  this will allow removing the `include ../../mk/spksrc.cross-go-env.mk` which isn't an entry point and should not being used directly.

mreid-tt:
> Can you move out the `go build` call from the spk down to the cross? this will allow removing the `include ../../mk/spksrc.cross-go-env.mk` which isn't an entry point and should not being used directly.

It's removed as part of this PR. Once merged it won't be there. Unless you are asking to do that as a separate PR in advance of this one being reviewed?

th0ma7:
> It's removed as part of this PR. Once merged it won't be there. Unless you are asking to do that as a separate PR in advance of this one being reviewed?

Great then!  No need for a separate PR, I'll simply won't bother about it on my end considering its fixed here.  thnx.

PR #6969, opened 2026-02-13
gh-action: fix intermittent CI failure due to toolchain cache

mreid-tt:
## Description

Fixes intermittent CI build failures where rust/cargo is not installed, causing `ModuleNotFoundError: No module named 'puccinialin'` errors during Python wheel builds.

### Problem

The CI uses two separate caches:
- **Toolchain cache** - contains compiled toolchain and status cookies (`.toolchain_done`, `.rustc_done`)
- **Distrib cache** - contains downloaded/installed tools including rust/cargo

These caches have different keys and can get out of sync. When:
1. Toolchain cache is restored with `.toolchain_done` present
2. Distrib cache is restored **without** rust/cargo installed

The build fails because:
- The `toolchain` target sees `.toolchain_done` and skips entirely (`toolchain: ;`)
- The `rustc` target never executes (it's a dependency of `_all` which is never reached)
- Later, `maturin` tries to build wheels, can't find `cargo`, attempts auto-install via `puccinialin`, fails

### Solution

The existing workaround removed `.rustc_done` but this is ineffective because the parent `toolchain` target guards the entire dependency chain.

This PR adds removal of `.toolchain_done` to force re-evaluation of all toolchain dependencies including rust installation:

```bash
rm -f toolchain/syno-${GH_ARCH}/work/.rustc_done
rm -f toolchain/syno-${GH_ARCH}/work/.toolchain_done  # NEW
```

### Evidence

Comparing failed vs successful runs:

| Run | Distrib Cache Size | Rust Installed | Result |
|-----|-------------------|----------------|--------|
| Good | ~973 MB | Yes | ✅ Build succeeds |
| Failed | ~842 MB | No | ❌ `puccinialin` error |

Both runs had identical toolchain cache keys, confirming the issue is cache desync.

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [x] Includes gh-action changes

PR #7040, opened 2026-03-30
gh-action fixes

th0ma7:
## Description

gh-action fixes

Relates to #7034 and #6969.
Toolchain fix probably due to `stage0`/`stage1` notions added part of #7031

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/.github/actions/build.sh
+++ b/.github/actions/build.sh
@@ -39,8 +39,8 @@ echo "===> MIN_DSM73_PACKAGES packages: ${MIN_DSM73_PACKAGES}"
 # Remove toolchain status files to enforce re-building toolchain including cargo/rust
 # This fixes issues on github-action where toolchain caching omits the
 # actual installation state of cargo/rust within the distrib folder
-rm -f toolchain/syno-${GH_ARCH}/work/.rustc_done
-rm -f toolchain/syno-${GH_ARCH}/work/.toolchain_done
+rm -f toolchain/syno-${GH_ARCH}/work/.toolchain*_done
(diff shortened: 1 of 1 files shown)


th0ma7:
Testing complete and successfull, merging.

mreid-tt:
hey @th0ma7 — I think I’ve come across a minor edge case in how we segment standard vs. restricted arch packages. With the recent build for #7069, the package doesn’t define a minimum DSM version, but it does build differently for DSM 7.1 vs 7.2. As a result, when I triggered a publish build targeting DSM 7.1 and 7.2, only the DSM 7.1 build was actually produced and published.

Not a critical issue, but something worth looking into. My suspicion is that restricted build targets are only being generated when the Makefile explicitly defines a minimum DSM version, and manual selections for a specific DSM version are being ignored in that logic.

th0ma7:
Edge case indeed, unsure how to manage this i.e. would need rethinking our current MIN_DSM_* or adding new vars to the loop to enforce building for DSM7.2 for testing?

On the other end, we do have the possibility to manually trigger a build for selected DSM in action section... that could suffice..

mreid-tt:
> Edge case indeed, unsure how to manage this i.e. would need rethinking our current MIN_DSM_* or adding new vars to the loop to enforce building for DSM7.2 for testing?

I'm not sure it is too complex. Did some digging and proposed a fix in #7070 which I've manually verified working in the narrow use-case for `radarr`.

PR #7065, opened 2026-04-06
ntopng: Avoid -lmysqlclient LDFLAGS to leak in dependencies

th0ma7:
## Description

The linker flag (`-lmysqlclient`) leaks into dependency builds and cause premature linking failures when the library is not yet available.  This is exactly the issue for `-lmysqlclient` as it does not yet exists when `mariadb-connector-c` gets built, thus fails.

Instead, I created a `mariadb_config` wrapper that is being copied over into the mariadb `build/NATIVE/bin` directory which obeys to regular calls.  This allows adding this to the `ENV` path, and enforcing `ntopng` to make use of it to capture the `--libs`, `--libs_r` and `--include` that it would normally ask for.

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/cross/mariadb-connector-c-3.3.11/Makefile
+++ b/cross/mariadb-connector-c-3.3.11/Makefile
@@ -57,7 +57,7 @@ MYSQL_DB_SOCKET = /run/mysqld/mysqld.sock
 endif
 CMAKE_ARGS += -DMARIADB_UNIX_ADDR=$(MYSQL_DB_SOCKET)
 
-include ../../mk/spksrc.cross-cmake.mk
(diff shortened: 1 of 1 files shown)


th0ma7:
@mreid-tt @hgy59 FYI - you might have been struggling a bit with mariadb and its inability to properly use `mariadb_config` normal calls as the tool is built for the target and unable to run on the host.  With this patch it does solves this issue by simulating a functional `mariadb_config` (same also applied before with `mysql_config`)

Just mentioning, this alternative could prove to be useful elsewhere.

PR #7066, opened 2026-04-06
gh-action optimize disk cleanup

th0ma7:
## Description

Use `easimon/maximize-build-space@v10` standard routine to maximize disk space.  This avoids maintaining our manual disk cleanup routine that takes between 2-5x times to run.

What this does:
1. Cleans-up the host space much faster than current manual method (current 1m45 up to 5m, new < 1m)
2. Take all the remaining storage space and reallocate it under a dedicated workspace such as `/home/runner/work/spksrc/spksrc`.  The first `spksrc` being the project, the second being the git checkout base directory.
3. All paths are now redirected under `path: ${{ github.workspace }}/<something>` which refert workspce created at step above.

## Checklist

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,7 +153,7 @@ jobs:
         id: cache-restore
         uses: actions/cache/restore@v5
         with:
-          path: distrib
+          path: ${{ github.workspace }}/distrib
           # Use a placeholder key that will never match for the exact key,
           # forcing GitHub to use restore-keys to find the best available cache
           key: distrib-placeholder-${{ github.run_id }}
@@ -203,7 +203,7 @@ jobs:
         if: steps.cache-check.outputs.save_cache == 'true'
         uses: actions/cache/save@v5
         with:
-          path: distrib
+          path: ${{ github.workspace }}/distrib
           key: distrib-${{ steps.distrib-hash.outputs.hash }}
 
   set-defaults:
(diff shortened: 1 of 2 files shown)


th0ma7:
While it does simplify the code the gain in time is lost in space - current method for cleaning up the host allows to increase space up to 120G while this method is limited to around 90G.  A combination of the two could be convenient but to the detriment of greater complexity.

Closing.

th0ma7:
@hgy59 i my not be totally crazy afterall, the total storage did increased for our github action runners, by quite a bit actually... 

Documentation still refers to a 14GB storage size as of https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories.

Although current disk space is significantly higher (below is from using `easimon/maximize-build-space@v10`) :
```
Filesystem                   Size  Used Avail Use% Mounted on
/dev/root                    145G  134G   11G  93% /
tmpfs                        7.9G   84K  7.9G   1% /dev/shm
tmpfs                        3.2G  1.1M  3.2G   1% /run
tmpfs                        5.0M     0  5.0M   0% /run/lock
efivarfs                     128M   29K  128M   1% /sys/firmware/efi/efivars
/dev/sda16                   881M   64M  756M   8% /boot
/dev/sda15                   105M  6.2M   99M   6% /boot/efi
tmpfs                        1.6G   12K  1.6G   1% /run/user/1001
/dev/mapper/buildvg-buildlv   88G  4.5G   84G   6% /home/runner/work/spksrc/spksrc
```

th0ma7:
@hgy59 I suggest we reconsider this approach being much faster at disk cleanup.  Further considering we now have much more disk space than we used to.  See comment above: https://github.com/SynoCommunity/spksrc/pull/7066#issuecomment-4229660461

I suggest we use this to replace the current cleanup routines and call it a day.  Further, this makes the disk layout more strucutured using standard gh-action variables.

mreid-tt:
At 93% it does flag a potential low disk space warning in https://github.com/SynoCommunity/spksrc/actions/runs/24058968908.

th0ma7:
> At 93% it does flag a potential low disk space warning in https://github.com/SynoCommunity/spksrc/actions/runs/24058968908.

Fair point, considering the much increased disk space I could probably increase root fs to reach 89% or similar.  Also this give a specific delimited storage for the build.

mreid-tt:
The wording of this seems bizarre though:
```
==========================================
DISK SPACE BEFORE CLEANUP
==========================================

=== FILESYSTEM USAGE (df -h) ===
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       145G   57G   88G  40% /
tmpfs           7.9G   84K  7.9G   1% /dev/shm
tmpfs           3.2G  1.0M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
efivarfs        128M   26K  128M   1% /sys/firmware/efi/efivars
/dev/sda16      881M   64M  756M   8% /boot
/dev/sda15      105M  6.2M   99M   6% /boot/efi
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```
then...
```
==========================================
DISK SPACE AFTER CLEANUP
==========================================

=== FILESYSTEM USAGE (df -h) ===
Filesystem                   Size  Used Avail Use% Mounted on
/dev/root                    145G  135G   10G  94% /
tmpfs                        7.9G   84K  7.9G   1% /dev/shm
tmpfs                        3.2G  1.1M  3.2G   1% /run
tmpfs                        5.0M     0  5.0M   0% /run/lock
efivarfs                     128M   26K  128M   1% /sys/firmware/efi/efivars
/dev/sda16                   881M   64M  756M   8% /boot
/dev/sda15                   105M  6.2M   99M   6% /boot/efi
tmpfs                        1.6G   12K  1.6G   1% /run/user/1001
/dev/mapper/buildvg-buildlv   88G   28K   88G   1% /home/runner/work/spksrc/spksrc
```

Cleanup makes less free space? We shouldn't call it cleanup then?

th0ma7:
> Cleanup makes less free space? We shouldn't call it cleanup then?

@mreid-tt good question... and the answer is yes-no-maybe.  Let me explain, I had put in place a super agressive clean-up mechanism to allow building things such as LLVM.  At that time the storage available was less than 16GB  or similar for the entire build system.

Recently @hgy59 noted that this cleanup eats-up a lot of cycles... it takes around 5 minutes at every run.  Could that be more efficient?

I found out about this `easimon/maximize-build-space@v10` which is pretty decent at cleaning-up disk-space although not as agressive as our current code.  But I then realized that the storage on our build image was increased to around 150GB, which is a really big incrase and solves our disk space problem altogether.

So we could argue of:
1. just removing our cleanup mechanism
2. or still doing a cleanup but using a standard method as total storage is less an issue now.  This is what this PR is now about.

The `easimon/maximize-build-space@v10` also allow having a dedicated mount-point for our workspace, segmenting it further form the build-system which is also a really neat feature.

Thus why I'm proposing this.

mreid-tt:
hey @th0ma7, thanks for the explanation. This makes sense. I was primarily commenting on the wording on the log messages. Cleanup doesn't seem appropriate if we are getting less space than when we start. Perhaps there are other operations in between which are actually consuming more space? If so, then the before cleanup report is being run too early or in an order that isn't quite right? Alternately, we can just call it something other than cleanup. Perhaps "DISK SPACE BEFORE INITIALISATION" ?

th0ma7:
> Perhaps "DISK SPACE BEFORE INITIALISATION" ?

@mreid-tt I have now rebanded it using rather "`DISK SPACE BEFORE MAXIMIZE-BUILD-SPACE`"

@mreid-tt and @hgy59 any opposition from any of you two in switching the cleanup for this more standard `easimon/maximize-build-space@v10`

mreid-tt:
hey @th0ma7, that renaming is much clearer. No objections to implementing this more standardised cleanup approach, especially if it executes faster as you mentioned.

th0ma7:
@hgy59 friendly reminder for a follow-up.

hgy59:
> @hgy59 friendly reminder for a follow-up.

INHO it would be best to keep the former cleanup, but to remove all apt-get commends (clean, remove, autoclean, clean)
- I guess it was the apt-get commands that took up to 20minutes (only saw 1.5 .. 3 minutes now, the maximize action takes < 30 seconds and does not use apt-get).
- with the former cleanup we got 123G, with `easimon/maximize-build-space@v10` only 88GB (and the annoying warning `Disk usage is at 93% - build may have failed due to low disk space`)

What's the amount of freed space with apt-get remove? we might add additional `rm` commands for what was removed with apt.get.

BTW the `easimon/maximize-build-space@v10` seems outdated and has issues with the `/mnt` path


just my 5 🪙 (take the best of each solution)

PR #7070, opened 2026-04-07
gh-action: Fix missing builds for standard packages on higher DSM toolchains

mreid-tt:
## Description

This is a follow-on from #7040

### Problem

Packages that build differently depending on DSM version but do not declare `REQUIRED_MIN_DSM` (e.g. `radarr`) were not being built when a higher DSM toolchain (e.g. `x64-7.2`) was selected.

In `build.sh`, when `GH_ARCH` matches a versioned toolchain like `x64-7.2`, `build_packages` is replaced with the DSM-restricted package list (`ARCH_MIN_DSM72_PACKAGES`). Since `radarr` is not a restricted package it never appears in that list, leaving `build_packages` empty and causing the build to exit early with "No packages to build."

### Fix

Use Bash parameter expansion to fall back to the standard package list when the DSM-version-specific list is empty, in both the arch and noarch branches of `build.sh`:

```bash
build_packages="${!var:-${ARCH_PACKAGES}}"
```

and:

```bash
build_packages="${!var:-${NOARCH_PACKAGES}}"
```

This ensures that standard packages are built on all selected toolchains, while DSM-restricted packages continue to be built only on their designated toolchain.

Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/.github/actions/build.sh
+++ b/.github/actions/build.sh
@@ -7,8 +7,11 @@
 # Functions:
 # - Build all packages depending on files defined in ${ARCH_PACKAGES} or ${NOARCH_PACKAGES}.
 # - Build for arch defined by ${GH_ARCH} (e.g. x64-6.1, noarch, ...).
-# - For DSM versions above the default builds, only packages declared for that
-#   minimum DSM version are built (driven by MIN_DSM<V>_PACKAGES env vars).
+# - For DSM versions above the default builds, packages declared for that minimum
+#   DSM version are built (driven by MIN_DSM<V>_PACKAGES env vars). If no
+#   DSM-restricted packages are present, standard packages are built instead,
+#   allowing packages that behave differently per DSM version (but declare no
+#   minimum) to be built on all selected toolchains.
 # - Successfully built packages are logged to ${BUILD_SUCCESS_FILE}.
 # - Failed builds are logged to ${BUILD_ERROR_FILE} and annotated as error.
 # - For failed builds, the make command and the latest 15 lines of the build output are written to ${BUILD_ERROR_LOGFILE}.
@@ -81,6 +84,7 @@ if [ "${GH_ARCH%%-*}" = "noarch" ]; then
         if [ "${DSM_VERSION}" = "${version}" ]; then
             v=${version//.}
             var="NOARCH_MIN_DSM${v^^}_PACKAGES"
+            # No DSM-restricted packages — fall back to standard noarch packages
             build_packages="${!var:-${NOARCH_PACKAGES}}"
             break
         fi
@@ -91,6 +95,7 @@ else
         if [ "${DSM_VERSION}" = "${version}" ]; then
             v=${version//.}
             var="ARCH_MIN_DSM${v^^}_PACKAGES"
+            # No DSM-restricted packages — fall back to standard arch packages
             build_packages="${!var:-${ARCH_PACKAGES}}"
             break
         fi


mreid-tt:
Hey @th0ma7 further to my comment on this edge case, I've tested with `radarr` on DSM 7.2 and it works fine. Let me know what other tests may be required.

th0ma7:
@mreid-tt would you mind testing in, indeed with `radarr`, but combined with other corner-cases such as `jellyfin`, `nextcloud`, `tvheadend` (although may be disabled atm, need to rm `BROKEN` file), and `borgbackup`.  No need to have it to run, just so to get the prepare action and copy-paste the expected resulting build would suffice, then canceling the build alltogether.

EDIT - Launched testing, here's the output:
```
Run ./.github/actions/prepare.sh
  ./.github/actions/prepare.sh
  shell: /usr/bin/bash -e {0}
  env:
    GH_SPK_PACKAGES:  borgbackup jellyfin nextcloud radarr tvheadend
    GH_DEPENDENCY_FOLDERS:  
    USER_SPK_TO_BUILD: 
 ---- find dependent packages
  Creating local configuration "local.mk"...
  Setting default toolchain version to DSM-6.2.4
  Setting default toolchain version to DSM-7.1
  Building dependency list...

 ---- build summary
  
  STANDARD builds (DSM 6.2.4, 7.1):
    arch    : python311 python312 borgbackup radarr synocli-videodriver ffmpeg7 tvheadend 
    noarch  : none
  
  RESTRICTED builds (min DSM version):
    DSM 7.2 arch: jellyfin
    DSM 7.2 noarch: nextcloud
===> PACKAGES to download references for: python311 python312 borgbackup jellyfin nextcloud radarr synocli-videodriver ffmpeg7 tvheadend
```

th0ma7:
@mreid-tt everything else looks good and unaffected - LGTM.

Just to confirm, in my build run output above (and from reading your description), `radarr` is now properly located in the expected build?  Where previously it would't be listed at all, right?

mreid-tt:
> Just to confirm, in my build run output above (and from reading your description), `radarr` is now properly located in the expected build? Where previously it would't be listed at all, right?

Yes, this is correct. In my initial publishing I got:

```
 ---- build summary
  
  STANDARD builds (DSM 6.2.4, 7.1):
    arch    : radarr 
    noarch  : none
  
  RESTRICTED builds (min DSM version):
    none
===> PACKAGES to download references for: radarr
```

This has not changed in the revised publishing (as expected):

```
 ---- build summary
  
  STANDARD builds (DSM 6.2.4, 7.1):
    arch    : radarr 
    noarch  : none
  
  RESTRICTED builds (min DSM version):
    none
===> PACKAGES to download references for: radarr
```

What is different is the actual build for DSM 7.2 which before showed:

```
 ---- initialize build
===> TARGET: x64-7.2
===> ARCH   packages: radarr 
===> NOARCH packages: 
===> ARCH_MIN_DSM72_PACKAGES: 
===> NOARCH_MIN_DSM72_PACKAGES: 
===> ARCH_MIN_DSM73_PACKAGES: 
===> NOARCH_MIN_DSM73_PACKAGES: 
===> No packages to build. <===
```

And after this PR shows:

```
 ---- initialize build
===> TARGET: x64-7.2
===> ARCH   packages: radarr 
===> NOARCH packages: 
===> ARCH_MIN_DSM72_PACKAGES: 
===> NOARCH_MIN_DSM72_PACKAGES: 
===> ARCH_MIN_DSM73_PACKAGES: 
===> NOARCH_MIN_DSM73_PACKAGES: 
===> PACKAGES to Build: radarr 
```

th0ma7:
Noting that in my mixed use case radarr did not show alongside jellyfin for dsm72 though, that's expected?

mreid-tt:
> Noting that in my mixed use case radarr did not show alongside jellyfin for dsm72 though, that's expected?

That is my understanding. The dsm72 and dsm73 lists are min DSM versions and not the standard ones. As such, by definition packages without min DSM versions for 7.2 or 7.3 will not show up there. So for normal / automatic builds, they would not include packages like radarr. However if the user directs a DSM 7.2 build of radarr the github actions should allow it to be built. This manual specification of DSM version really only happens with GitHub-based publishing.

th0ma7:
Ah, now i get it, 👍

PR #7083, opened 2026-04-14
bazarr: bump to 1.5.6

smaarn:
## Description

Updates bazarr from 1.5.5 to 1.5.6 (https://github.com/morpheus65535/bazarr/releases/tag/v1.5.6)

## Checklist

### Type of change

- [x] Package update


smaarn:
Still testing functionality to validate everything works.

SWu:
Tested the package and it works basically same as previous version. There is an issue with Synology reporting package running state though (has been true of previous versions as well), since the bazarr parent process doesn't exit without direct intervention so when the child is shut down or dies Synology will continue reporting the package as running since it only [checks the parent PID](https://github.com/SynoCommunity/spksrc/blob/319f1af00ba0faf3c928f6637ed79a07aed31aa0/mk/spksrc.service.start-stop-status#L113-L132)

smaarn:
> Tested the package and it works basically same as previous version. There is an issue with Synology reporting package running state though (has been true of previous versions as well), since the bazarr parent process doesn't exit without direct intervention so when the child is shut down or dies Synology will continue reporting the package as running since it only [checks the parent PID](https://github.com/SynoCommunity/spksrc/blob/319f1af00ba0faf3c928f6637ed79a07aed31aa0/mk/spksrc.service.start-stop-status#L113-L132)

Investigating but it seems to me that there was a specific logic implemented at bazarr level to ensure that, when the parent process was killed, the bazarr subprocess would be killed. How did you try shutting down the bazarr service ? via the GUI ?

SWu:
> > Tested the package and it works basically same as previous version. There is an issue with Synology reporting package running state though (has been true of previous versions as well), since the bazarr parent process doesn't exit without direct intervention so when the child is shut down or dies Synology will continue reporting the package as running since it only [checks the parent PID](https://github.com/SynoCommunity/spksrc/blob/319f1af00ba0faf3c928f6637ed79a07aed31aa0/mk/spksrc.service.start-stop-status#L113-L132)
> 
> Investigating but it seems to me that there was a specific logic implemented at bazarr level to ensure that, when the parent process was killed, the bazarr subprocess would be killed. How did you try shutting down the bazarr service ? via the GUI ?

Actually shutdown seems to work, but if there's something in the child that causes an unhandled exception and the child crashes it seems the parent process never terminates so Syno always reports the service as still running.

SWu:
The above would seem to be more an issue with the bazarr codebase though so I think the Syno package is good to go.

smaarn:
> > > Tested the package and it works basically same as previous version. There is an issue with Synology reporting package running state though (has been true of previous versions as well), since the bazarr parent process doesn't exit without direct intervention so when the child is shut down or dies Synology will continue reporting the package as running since it only [checks the parent PID](https://github.com/SynoCommunity/spksrc/blob/319f1af00ba0faf3c928f6637ed79a07aed31aa0/mk/spksrc.service.start-stop-status#L113-L132)
> > 
> > 
> > Investigating but it seems to me that there was a specific logic implemented at bazarr level to ensure that, when the parent process was killed, the bazarr subprocess would be killed. How did you try shutting down the bazarr service ? via the GUI ?
> 
> Actually shutdown seems to work, but if there's something in the child that causes an unhandled exception and the child crashes it seems the parent process never terminates so Syno always reports the service as still running.

Interesting, so here, from what I see in the bazarr codeline there's indeed a glitch where, if the process dies for some unplanned reason (e.g. without generating a stop or restart file) it will indeed keep on running. Seems to be an issue already present in current version though. Raised in https://github.com/morpheus65535/bazarr/issues/3299

mreid-tt:
> Actually shutdown seems to work, but if there's something in the child that causes an unhandled exception and the child crashes it seems the parent process never terminates so Syno always reports the service as still running.

This may be something we can possibly address with a custom `start-stop-status.sh` script. Looking at the existing standard `service-setup.sh` script we have a traditional launcher:

https://github.com/SynoCommunity/spksrc/blob/319f1af00ba0faf3c928f6637ed79a07aed31aa0/spk/bazarr/src/service-setup.sh#L11

But this [`bazarr.py`](https://github.com/morpheus65535/bazarr/blob/a21300b3e1acf1e9e4f58b051ff9204e24459c45/bazarr.py) is a self-contained launcher that includes its own child processes and management functions including:
```
start_bazarr()
check_status()
exit_program(status_code)
terminate_child()
```

There are also a number of literals defined in [`bazarr/literals.py`](https://github.com/morpheus65535/bazarr/blob/master/bazarr/literals.py) which seem to write to files like:
```
FILE_RESTART = 'bazarr.restart'
FILE_STOP = 'bazarr.stop'
```

We can perhaps craft a script to leverage these existing components to determine the true status of the child processes and report upwards to DSM correctly. A simple example would be this one used in [`spk/postgresql/src/start-stop-status.sh`](https://github.com/SynoCommunity/spksrc/blob/master/spk/postgresql/src/start-stop-status.sh).

mreid-tt:
> Interesting, so here, from what I see in the bazarr codeline there's indeed a glitch where, if the process dies for some unplanned reason (e.g. without generating a stop or restart file) it will indeed keep on running.

I don't know if this is exactly correct. Another possibility is that we may not be wired correctly to detect non-standard terminations of child processes. I've suggested a possible solution above.

What would help is if we can have a test case to reliably reproduce child process failures to determine if their existing management shell detects the failure. If so, does it write the output to one of the file literals? If so, then we we could look for this file to determine the actual status of the child process.

Normally the framework just looks for a PID file and if it's not present it will consider the process stopped. Our framework does have the ability to monitor multiple PID files but I don't know enough about what `bazarr` writes out as I've not used the package.

smaarn:
> > Interesting, so here, from what I see in the bazarr codeline there's indeed a glitch where, if the process dies for some unplanned reason (e.g. without generating a stop or restart file) it will indeed keep on running.
> 
> I don't know if this is exactly correct. Another possibility is that we may not be wired correctly to detect non-standard terminations of child processes. I've suggested a possible solution above.

It seems that it actually was correct if I understand correctly https://github.com/morpheus65535/bazarr/issues/3233.

Since it has been fixed in 1.5.7-beta-2 it seems so we should be good now.

Thanks @mreid-tt for your insights though.

PR #7098, opened 2026-04-19
patch.mk: Avoid duplicate patches to be applied

th0ma7:
## Description

patch.mk: Avoid duplicate patches to be applied

In rare case armv7 and x64 patches could be double counted as they are both an 'arch' and a 'group'.  This fix ensure evey patch are unique within the PATCHES variable.

Fixes https://github.com/SynoCommunity/spksrc/pull/6839#issuecomment-4274576828

## Checklist

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update

### DIFF
--- a/mk/spksrc.patch.mk
+++ b/mk/spksrc.patch.mk
@@ -36,7 +36,7 @@ PATCHES += $(sort $(foreach group,ARM_ARCHS ARMv5_ARCHS ARMv7_ARCHS ARMv7L_ARCHS
 	   $(wildcard patches/$(shell echo $(group) | cut -f1 -d '_' | tr 'A-Z' 'a-z')/*.patch \
 	              patches/$(shell echo $(group) | cut -f1 -d '_' | tr 'A-Z' 'a-z')-$(TCVERSION)/*.patch))))
 endif
-PATCHES := $(realpath $(PATCHES))
+PATCHES := $(call uniq,$(realpath $(PATCHES)))
 
 PATCH_COOKIE = $(WORK_DIR)/.$(COOKIE_PREFIX)patch_done
 


th0ma7:
@hgy59 quick testing I was able to build `bind_9.20` succesfully.  Mind giving it a try to confirm?

hgy59:
we have another issue with patches

cross/libvpx can't be built for arch-88f6281-5.2.

There are patches folders `88f6281` and `DSM-5.2`  - it is the same patch file, but again for 88f6281-5.2 both are found (but not in the same folder).

We have the same situation in python311 and this is solved by dedicated folders like
 `88f6281`, `ppc853x-5.2` and `x86-5.2`  (ppc853x is not supported with cross/libvpx)
 
To solve this without dedicated folders we would need to avoid duplicate files in different folders, but I guess this is out of scope of this PR.

hgy59 (approved):

th0ma7:
> To solve this without dedicated folders we would need to avoid duplicate files in different folders, but I guess this is out of scope of this PR.

@hgy59  is this a real issue?  or is this more like an enhancement?  Thinking of how we could do this, whereas if the patch filename is identical to just be keeping one instance (doesn't matter which one in theory).  That would probably "relatively" easy to code... if that is the intent and possibly the right angle of attack?

hgy59:
> > To solve this without dedicated folders we would need to avoid duplicate files in different folders, but I guess this is out of scope of this PR.
> 
> @hgy59 is this a real issue? or is this more like an enhancement? Thinking of how we could do this, whereas if the patch filename is identical to just be keeping one instance (doesn't matter which one in theory). That would probably "relatively" easy to code... if that is the intent and possibly the right angle of attack?

This would actually simplify the patches that are same for 88f6281 and DSM-5.2. And may be other in future.
If it is simple we could benefit, otherwise we have a workaround.

th0ma7:
> This would actually simplify the patches that are same for 88f6281 and DSM-5.2. And may be other in future. If it is simple we could benefit, otherwise we have a workaround.

... and filename would not be sufficient, it would need to be an md5sum of the patch files or similar ... let me think about it, I might be able to suggest something in a foreseable future.

PR #7095, opened 2026-04-18
Update core libs

th0ma7:
## Description

Update core libs:
- `libexpat`: Update from version 2.7.1 to 2.7.5 and migrate to cmake
- `freetype`: Update from version 2.14.1 to 2.14.3
- `fontconfig`: Update from version 2.15.0 to 2.16.0

Note the following was attempted but got reverted - synology kernels are too old and need staying on v2.38
- `libuuid`: Update from version 2.38.1 to 2.42 and migrate to meson

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [ ] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)
- [x] Update to misc core libraries

PR #7101, opened 2026-04-19
update util-linux to v2.41.4

hgy59:
## Description

- update libraries created with util-linux (libblkid, libmount)
- update tools created with util-linux (ionice)

Fixes # 

## Checklist

### Type of change

- [x] Library update


hgy59:
@th0ma7 I will update cross/libuuid in a dedicated PR (due to the many dependencies).
I do not update to v2.42, because cross/misc-util-linux fails to build.
I do not use meson build, because it is not possible to build dedicated libraries (despite the options are available in meson_options.txt, it does not work).

th0ma7 (approved):
Looks pretty straight forward and probably best to avoid switching to meson considering issues I've had with #7095 when attempting that.

Any intent to try migrating `libuuid`?  Keeping things with autoconf might actually work out like the others?

hgy59:
> Any intent to try migrating `libuuid`? Keeping things with autoconf might actually work out like the others?

To minimze the possible build timeout I will create a dedicted PR for libuuid.

PR #7107, opened 2026-04-22
github build action: set specific title for manual builds

hgy59:
## Description

add package name to title of manual build actions
show the publish option in the title

Better distinction of manual builds in Actions history.


th0ma7:
That's a really good idea 🚀

PR #7108, opened 2026-04-24
Add iCloud Photo Sync package

Euphonique:
## Summary
- New SPK package **icloudphotosync** that automatically mirrors iCloud photo libraries to Synology NAS devices
- Features: multi-account sync, incremental downloads, album selection, HEIC-to-JPEG conversion, scheduled operation, DSM 7.2 web UI
- Includes three new cross packages as dependencies:
  -  — HEIC/HEIF conversion tools from libheif 1.18.2
  -  — Limited Error Raster Compression library v4.0.0
  -  — fast deflate/zlib/gzip compression library v1.22

## Details
- **Requires**: DSM 7.2+, Python 3.11
- **Unsupported architectures**: OLD_PPC_ARCHS, ARMv5_ARCHS
- **License**: MIT
- **Homepage**: https://github.com/Euphonique/iCloudPhotoSync

## Test plan
- [x] Successfully built SPK for x64-7.2 architecture (40 MB)
- [x] Verified all cross-compilation dependencies resolve correctly
- [x] Verified HEIC conversion tools are included in the package
- [x] Verified PLIST files match actual build output
- [x] Verified digest checksums for all source archives
- [ ] Install and test on DSM 7.2 NAS (additional architectures)


mreid-tt:
Hey @Euphonique, I've not gone through in detail but some overall comments:

- The folders in `cross/*` are generally lowercase so `cross/lerc` would be preferred
- The `spksrc` framework is typically used to package other open source applications. Looking at your source repo, you seem to have already structured it to create a package
    - Because of this the key components of the code (like the `app.cgi` and a number of `*.js` files), you have effectively replicated into our source code. This would make maintenance a bit unusual since rather than just downloading a new version from an external source you would be editing the source code in this repo which would result in large PRs in the future
    - As we typically don't maintain external source code in this repo, you may wish to consider packaging from your repo the unique components into a release that can then be installed into the packager as part of the build. So for example, if the key application code you provide is in `src/app`, `src/bin` and `src/lib`, then you can download from your source repo in a `cross/icloudphotosync` and use the `Makefile` to install those resources into your package here
- In your `service-setup.sh` you have a function to find the python executable however this would not be necessary since you have a specific Python version in the `Makefile` as a package dependency
- In your  `service-setup.sh`, you have a few instances of `cat >> "$SYNOPKG_TEMP_LOGFILE" <<MSGEOF` but the content seems to be better served by wizard UI files. If not, in general it would be better to leverage an installed template rather than using `cat` to create files from the `service-setup.sh`
- In your `Makefile`, I see you have a package dependency for `python311`. Is this a hard dependency? If not, most of our packages are updated to a unified version of python. We use `python312` at present and will be transitioning to a newer global dependency in time

mreid-tt:
hey @th0ma7, I'd appreciate your thoughts on this submission. Also, I noted an anomaly in the GitHub build action. For the evaluation dependencies I noted:
```
---- build summary
  
  STANDARD builds (DSM 6.2.4, 7.1):
    arch    : icloudphotosync 
    noarch  : none
  
  RESTRICTED builds (min DSM version):
    none
```
However, this should include a DSM 7.2 build since this is the `REQUIRED_MIN_DSM` set in the `Makefile`.

th0ma7:
> However, this should include a DSM 7.2 build since this is the `REQUIRED_MIN_DSM` set in the `Makefile`.

@mreid-tt I woulder if the `UNSUPPORTED_ARCHS = $(OLD_PPC_ARCHS) $(ARMv5_ARCHS)` which is irrelevant as not supported under DSM 7.2 anyway isn't affecting that?  further as https://github.com/SynoCommunity/spksrc/pull/7066 runs just fine for DSM-7.2 only ... 

That could be quickly tested through a PR to confirm with and without, and as needed open up an issue to track it outside from here.

th0ma7:
@Euphonique I agree with @mreid-tt on his feedback, in particular in mapping all of the code directly into the spk diractory - spksrc is meant to integrate multiple external package sources, not to act as maintaining a software source tree per say.

The natural fit would be of using a `cross/icloudphotosync` whereas you would use `spksrc.install-resources.mk` to simply install files/folders in the destination install folder for the spk to bundle.  I presume this code comes from a repository of some sort, if not I suggest you create one in your own github, tag a version, then the `cross/icloudphotosync` would simply refer to it, making things easier later to maintain and update as needed.

PR #7117, opened 2026-04-29
headphones: remove PYTHON_PACKAGE

hgy59:
## Description

- headphones it not a python meta-spk, definition of PYTHON_PACKAGE is misleading

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/spk/headphones/Makefile
+++ b/spk/headphones/Makefile
@@ -4,8 +4,8 @@ SPK_REV = 13
 SPK_ICON = src/headphones.png
 
 DEPENDS =
-PYTHON_PACKAGE = python312
-SPK_DEPENDS = "${PYTHON_PACKAGE}:git"
+
+SPK_DEPENDS = "python312:git"
 
 MAINTAINER = Diaoul
 DESCRIPTION = Headphones is an automatic NZB and torrent downloader for music. Mark albums as wanted and have them downloaded as soon as they are available. Different quality settings are possible, including lossless. With Headphones you can also manage your music library automatically: move, tag, rename and index for your DiskStation.
@@ -21,7 +21,7 @@ STARTABLE  = yes
 SERVICE_USER = auto
 SERVICE_SETUP = src/service-setup.sh
 SERVICE_PORT = 8181
-SERVICE_PORT_TITLE = $(DISPLAY_NAME)
+SERVICE_PORT_TITLE = Headphones Web UI
 
 # Admin link for in DSM UI
 ADMIN_PORT = $(SERVICE_PORT)


mreid-tt:
hey @hgy59, I'm unclear as to the reasoning for this change. Maybe I don't quite get the distinction of a _python meta-spk_, perhaps you can explain more? I see a similar structure in ruTorrent ([Makefile](https://github.com/SynoCommunity/spksrc/blob/master/spk/rutorrent/Makefile)). Does this also apply there?

hgy59:
> hey @hgy59, I'm unclear as to the reasoning for this change. Maybe I don't quite get the distinction of a _python meta-spk_, perhaps you can explain more? I see a similar structure in ruTorrent ([Makefile](https://github.com/SynoCommunity/spksrc/blob/master/spk/rutorrent/Makefile)). Does this also apply there?

This is a noarch package that doesn't require prebuilt python and does not build any wheels. It just depends on python312 package at installation/run time.

A python meta-spk package includes `spksrc.spk-meta.mk` (former `spksrc.python.mk`) but headphones includes `spksrc.spk.mk`.

The definition of `PYTHON_PACKAGE` is required for `spksrc.spk-meta.mk` and used in  `spksrc.python*.mk`.


The same applies to rutorrent, but in rutorrent Makefile  the `PYTHON_PACKAGE` variable is used two times in the Makefile and there is a definition `WHEELS = src/requirements-pure.txt`
Further research is needed to determine whether pure python wheels can be built without pre-compiled python. If pre-compiled python is required, rutorrent must use `spksrc.spk-meta.mk` and `PYTHON_PACKAGE`; otherwise, `PYTHON_PACKAGE` can be omitted (and SPK_DEPENDS must not use it).
@th0ma7 any hint related to rutorrent and spk-meta?

mreid-tt (approved):

hgy59:
> The same applies to rutorrent, but in rutorrent Makefile the `PYTHON_PACKAGE` variable is used two times in the Makefile and there is a definition `WHEELS = src/requirements-pure.txt`

@mreid-tt I did som e analysis of rutorrent package.
You should avoid the definition of `PYTHON_PACKAGE` and the definition of `WHEEL`.
rutorrent does not build any wheel. It only packages `requirements-pure.txt` in the folder `share/wheelhouse`.
Instead of the definition of `WHEELS = src/requirements-pure.txt` it sould install `requirements-pure.txt` in the folder `share/wheelhouse` directly.
Since rutorrent isn't a meta-spk package it is neither possible to include the pure wheels with definition of `WHEELS_PURE_PYTHON_PACKAGING_ENABLE = 1`

th0ma7:
> any hint related to rutorrent and spk-meta?

I believe the code as-is is expected to generated to default `crossenv` which means it requires a `native/python31*` at minimum (untested).  Also, it currently uses `build-python` from the crossenv to build pure python wheels as the crossenv contains all the needed tools that has been installed.

That could probably be done differently, such as using the system level default python and adding the tools in our default image, thus not needing a crossenv for pure python wheels.

th0ma7:
> You should avoid the definition of `PYTHON_PACKAGE` and the definition of `WHEEL`.
> rutorrent does not build any wheel. It only packages `requirements-pure.txt` in the folder `share/wheelhouse`.
> Instead of the definition of `WHEELS = src/requirements-pure.txt` it sould install `requirements-pure.txt` in the folder `share/wheelhouse` directly.
> Since rutorrent isn't a meta-spk package it is neither possible to include the pure wheels with definition of `WHEELS_PURE_PYTHON_PACKAGING_ENABLE = 1`

it would need to be installed as `share/wheelhouse/requirements.txt` which is what is being called at package install time.  Noting that this is to change soonish... https://ichard26.github.io/blog/2026/04/whats-new-in-pip-26.1/ for using `pylock.toml` or `pylock.<name>.toml` instead.

hgy59:
> > You should avoid the definition of `PYTHON_PACKAGE` and the definition of `WHEEL`.
> > rutorrent does not build any wheel. It only packages `requirements-pure.txt` in the folder `share/wheelhouse`.
> > Instead of the definition of `WHEELS = src/requirements-pure.txt` it sould install `requirements-pure.txt` in the folder `share/wheelhouse` directly.
> > Since rutorrent isn't a meta-spk package it is neither possible to include the pure wheels with definition of `WHEELS_PURE_PYTHON_PACKAGING_ENABLE = 1`
> 
> it would need to be installed as `share/wheelhouse/requirements.txt` which is what is being called at package install time. Noting that this is to change soonish... https://ichard26.github.io/blog/2026/04/whats-new-in-pip-26.1/ for using `pylock.toml` or `pylock.<name>.toml` instead.

Ok, my mistake. The function `install_python_wheels` uses the file `share/wheelhouse/requirements.txt` (not requirements-pure.txt). So the file  `src/requirements-pure.txt` must be added to the package as `share/wheelhouse/requirements.txt` when definition of `WHEELS` is removed (currently the rutorrent package contains requirements-pure.txt and requirements.txt with same content).

mreid-tt:
hey @th0ma7, @hgy59 thanks for the advice on ruTorrent. I have incorporated the changes discussed in https://github.com/SynoCommunity/spksrc/pull/7146/changes/646fae7adaab5a5872d15a1d472d273617ea04a0

Issue #6145, opened 2024-06-15
custom rust versions are not supported


hgy59:
As discussed in https://github.com/SynoCommunity/spksrc/pull/6123#issuecomment-2169170879 it is not possible yet, to have a specific rust version (i.e. 1.77.2, 1.78, nightly, ...) for a specific cross package.

A solution is now required, since `/cross/fd_8.7.0` fails to build for hi3535 arch with rust >= 1.79.

th0ma7:
I don't believe this is the issue... The installer tries by default to install a newer version of the rust toolchain using the minimal version needed.... in this case version 1.81 so it looks for a `1.81.0-armv7l-unknown-linux-gnueabi`.  Such version no longer exists for `armv7l` since 1.77.2.

Interestingly I do have a locally compiled and accessible `1.82.0-armv7l-unknown-linux-gnueabi` but it doesn't even try to use it.

The solution is probably to find a way to override this behavior to avoid the cargo call:
````
  ==> Cargo install rust package fd (rustc +1.81.0-armv7l-unknown-linux-gnueabi -vV)
error: toolchain '1.81.0-armv7l-unknown-linux-gnueabi' is not installable
```
So it enforces using the version already available along with using the rust armv7l build from #6364.

hgy59:
> I don't believe this is the issue... The installer tries by default to install a newer version of the rust toolchain using the minimal version needed.... in this case version 1.81 so it looks for a `1.81.0-armv7l-unknown-linux-gnueabi`. Such version no longer exists for `armv7l` since 1.77.2.

But when I clean the syno-hi3535-6.2.4 toolchain and build a package for hi3535-6.2.4 it installs the toolchain and rustc is version 1.83.0

```
===>  Installing rustc toolchain for syno-hi3535
===>  - rustup installation PATH: /spksrc/distrib/rustup
===>  - cargo installation PATH: /spksrc/distrib/cargo
===>  - default PATH: /spksrc/distrib/cargo/bin:/scripts:/usr/local/bin:/usr/bin:/bin:/usr/games
===>  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path
info: downloading installer
warning: it looks like you have an existing installation of Rust at:
warning: /spksrc/distrib/cargo/bin
warning: It is recommended that rustup be the primary Rust installation.
warning: Otherwise you may have confusion unless you are careful with your PATH
[...]

Issue #5283, opened 2022-05-25
Armadaxp toolchain incorrectly using vfpv3-d16 flag?


ta264:
### Is this a new Bug?

- [X] I checkd that the bug hasn't been reported before

### Package Name

syno-armadaxp

### Package Version

All

### Device Model

DS414

### Device Architecture

ARMv7

### Firmware Version

7.0.1

### What happened?

The toolchain for armadaxp is setting `-mfpu=vfpv3-d16`.  

https://github.com/SynoCommunity/spksrc/blob/a6da5d8e312b3755c157ce0ff248289a9a63a773/toolchain/syno-armadaxp-7.0/Makefile#L12

I think this is incorrect based on the output of 
```
$ cat /proc/cpuinfo
Processor       : Marvell PJ4Bv7 Processor rev 2 (v7l)
processor       : 0
BogoMIPS        : 1332.01

processor       : 1
BogoMIPS        : 1332.01

Features        : swp half thumb fastmult vfp edsp vfpv3 tls
CPU implementer : 0x56
CPU architecture: 7
CPU variant     : 0x2
CPU part        : 0x584
CPU revision    : 2

Hardware        : Marvell Armada XP Development Board
Revision        : 0000
Serial          : 0000000000000000
```

Note it reports `vfpv3` but **not** `vfpv3d16`.

Compare with the output on an Armada370 DS115j:
```
# cat /proc/cpuinfo
Processor       : Marvell PJ4Bv7 Processor rev 1 (v7l)
BogoMIPS        : 795.44
Features        : swp half thumb fastmult vfp edsp vfpv3 vfpv3d16 tls
CPU implementer : 0x56
CPU architecture: 7
CPU variant     : 0x1
CPU part        : 0x581
CPU revision    : 1

Hardware        : Marvell Armada-370
Revision        : 0000
Serial          : 0000000000000000
```

This does report `vfpv3d16` and so needs the `-mfpu=vfpv3-d16` flag.

I think the `armadaxp` toolchains should be using `-mfpu=vfpv3` like `armada375` does:
https://github.com/SynoCommunity/spksrc/blob/a6da5d8e312b3755c157ce0ff248289a9a63a773/toolchain/syno-armada375-7.0/Makefile#L12

[...]

PR #7119, opened 2026-05-01
synocli-file: update tools

### LINKED ISSUE #7131: [Package Request] Microsoft Edit
### Software Name

msedit

### Brief Description

A simple editor for simple needs.

This editor pays homage to the classic [MS-DOS Editor](https://en.wikipedia.org/wiki/MS-DOS_Editor), but with a modern interface and input controls similar to VS Code. The goal is to provide an accessible editor that even users largely unfamiliar with terminals can easily use.

Edit v2 adds the Lightweight Syntax Highlighter. It has a ~40kB footprint for a dozen languages plus runtime, barely grows with each language added, and runs at >100MB/s.

The highlighter is based on a simple programming language that combines regular expressions with explicit control flow. It's designed such that the runtime can be easily ported to other languages, including JavaScript.

### Website

https://github.com/microsoft/edit

### Documentation

_No response_

https://github.com/microsoft/edit#linux-build-from-source

### Source code

https://github.com/microsoft/edit

### OpenSource License

MIT

### Other License

_No response_

hgy59:
## Description

- add Microsoft Edit v2.0.0
- update dos2unix from v7.5.3 to v7.5.6
- update eza from v1.23.4 to v1.23.5
- update fd from v10.3.0 to v10.4.2 (keep v10.3.0 for qoriq)
- update file from v5.46 to v5.48
- update fzf from v0.67.0 to v0.74.2
- update jdupes from v1.30.0 to v1.31.1
- update less from v685 to v704
- update lzip from 1.25 to v1.26 and plzip from v1.12 to v1.13
- update micro from v2.0.14 to v2.0.15
- update mg from v3.7 to v4.0
- update nano from v8.7 to v9.2
- update nnn from v5.1 to v5.3
- update ripgrep from v15.1.0 to v15.2.0
- update sd from v1.0.0 to v1.1.0 (keep v1.0.0 for qoriq and ARMv5)
- update tree from v2.2.1 to v2.3.2

Closes #7131

## Checklist

### Type of change

- [x] Package update


hgy59:
@th0ma7 I have errors when building rust tools for ARMv7L (hi3535-6.2.4) in my local environment.

```
  = note: the `armv7-unknown-linux-gnueabi` target may not be installed
  = help: consider downloading the target with `rustup target add armv7-unknown-linux-gnueabi`
```

Although the Rust toolchain `armv7-unknown-linux-gnueabi` is installed, I'm still getting the error above.

I also tried cleaning up the syno-hi3535-6.2.4 toolchain before building to reinstall the Rust toolchain `armv7-unknown-linux-gnueabi`, but the errors persist.

I suspect this is related to other installed Rust toolchains (ARMv7, etc.) and framework changes since January 2026. These errors don't occur in GitHub builds, as they always use a single architecture environment with only host- and one architecture-specific Rust toolchain installed.

It is only ARMv7L, all other archs work just fine.
Any idea how to fix this?

th0ma7:
Which package?  You can enforce DEFAULT_ENV += rust which will engorce adding it up if need be (not on my build system currently, going from memory)

hgy59:
> Which package? You can enforce DEFAULT_ENV += rust which will engorce adding it up if need be (not on my build system currently, going from memory)

It is fixed, sorry for the noise.
All rust packages were affected.
It was related to my local environment where  I share some folders between all containers (distrib, toolchain, kernel, toolkit) and for an unknown reason it was not possible to install the rust toolchain for ARMv7L. After re-population of all shared folders it works again...

th0ma7:
Glad it works!

hgy59:
@th0ma7 @mreid-tt  I have an issue with armada370 cpu and ripgrep (built with rust)
It seems related to #5283, the difference, however, is that ripgrep is built for vfpv3-d32 but supports d16 only (and .NET vs rust).
 
ripgrep (rg) crashes with a core dump (illegal instruction) when calling "rg --help" but not with "rg -h".
Thanks to synocli-devel I was able to find the illegal instructions (there are instr. using d32, but armada370 has d16 only i.e. 32- vs. 16-bit floating point registers) in the code of the textwrap crate. (rg --help shows extended help text that needs textwrap, but rg -h does not).

I tried several rustflags to disable d32 to force cortex-a9 cpu etc. and to disable optimizations, but nothing fixes the code. 
Probably it is related to the rust toolchain version since ripgrep 14.1.1 in synocli-file 4.0.1-28 still works, but when I currently build ripgrep 14.1.1 (with rust 1.95) it has the illegal instructions.

I don't know whether other commands of ripgrep will crash on armada370.
My main issue is that I don't want to remove support armada370 of the synocli-file package and I don't want to remove ripgrep for all armv7-6.2.4 models.

One possible solution (that I considered) is to remove all rust tools from the synocli-file package an create a dedicated synocli-rust-tools (or similar) package.

There are several issues with the current spksrc framework
- it is not possible to build tools with different rust versions into one package (see #6145)
- we can't provide packages for generic archs that exclude some dedicated archs and create separate packages for the dedicated archs (I guess the repo would support that we create an armv7 package excluding armada370 and create an armada370 package with same name, version and revision - but the github build worksflow do not support this).

And since the disk of the repository is limited we can't provide older package versions that still support some older versions.

One argument for a dedicated synocli-rust-tools package is, that almost all rust tools are fully selfcontained and don't share (large) common libraries (this was one of the main argument to create the synocli-* packages.

One older idea (of mine) was to keep the working binaries of the rust tools and package those when it was not possible to build it anymore. This idea could be considered again (so far it got obsolete with the --locked parameter).

What shall we do?

EDIT:
I was able to fix this ARMv7 issue (it does not exist with Rust < 1.94)
There was a change in 1.94 that is the reason for a related issue https://github.com/rust-lang/rust/issues/159973

mreid-tt:
hey @hgy59, regarding your earlier comment, this is a bit beyond my understanding. Perhaps this AI response can assist?

---

Thanks for laying out the situation so clearly. I dug into the current framework state and both crashes are **build-target bugs, not packaging-structure problems**. That framing matters for the fix. My take:

## Context

**qoriq (#7304)** — the framework has mapped PPC/qoriq to `powerpc-unknown-linux-gnuspe` since Feb 2024 (#5879), with a pinned `RUSTUP_QORIQ_TOOLCHAIN = 1.82.0`. **#7304 (July 2026, synocli-file 4.0.2-29) still reports bat/lsd crashing on qoriq — so this is NOT confirmed resolved.**

I verified the actual shipped binary from the published repo index (synocli-file 4.0.2-29 for qoriq, md5-matched): `bat`/`lsd` are Rust binaries built with rust **1.82.0** (the framework's qoriq pin, so the qoriq rust pipeline is in effect). **However, static inspection found no ABI difference between them and the C-built tools** — all binaries in the package share `e_flags=0x0`, the soft-float ABI tag, and contain no SPE (`ev*`/`ef*`) mnemonics at all. So I could **not** confirm the "built with `powerpc-unknown-linux-gnu` instead of `gnuspe`" theory from the binary, nor identify the crashing instruction. The qoriq diagnosis needs runtime data or an A/B rebuild:
- ask the #7304 reporter for the SIGILL address (dmesg / run under gdb) — that pinpoints the offending instruction;
- or build bat/lsd A/B with `gnuspe` vs `gnu` targets and compare, since the ELF markers do not discriminate the two;
- confirm whether the source-built Tier-3 qoriq toolchain (`RUST_BUILD_TOOLCHAIN=1`, opt-in, default 0) is actually built and used in CI.

**armada370** — it sits in `DSM_ARMv7_ARCHS`, so the rust mapping picks the generic `armv7-unknown-linux-gnueabihf` target. The repo itself notes in `mk/spksrc.common/archs.mk` that some ARMv7 chips are "without full vfpv3 support (having only vfpv3-d16)" — armada370 (and alpine) are the ones called out. So the d16 vs d32 mismatch is well supported. **Caveat: the exact mechanism by which the current rust build emits d32 instructions is not confirmed** (standard rust ARMv7 target features specify `+d16`; if d32 code appears it may come from LLVM's default `target-cpu` or a toolchain/LLVM-version regression rather than the target spec itself). Worth confirming with a minimal repro before committing to a fix.

## What I'd advise

**1. Don't let a dedicated `synocli-rust-tools` package be the fix.** It doesn't resolve either crash — it just moves the SIGILL into a differently-named package. And the split loses the one thing synocli-* packages are for (pooling large shared libs) for no benefit to the affected users.

**2. Treat both as build-target/FPU problems, and get a minimal repro first.** The framework already has the plumbing for a target-level fix:
- per-target rustflags via `CARGO_TARGET_<TARGET>_RUSTFLAGS` (wired from `TC_RUSTFLAGS`/`TC_EXTRA_RUSTFLAGS` in `mk/spksrc.toolchain/tc-rust.mk`), and
- a Tier-3 source-build mechanism for targets rustup doesn't distribute (`RUST_BUILD_TOOLCHAIN=1`, used for qoriq).

So the paths are: an armada370-specific target/FPU override (custom target JSON or per-arch rustflags forcing `+d16`/`-d32`), and a genuine qoriq SPE verification. I know rustflags "didn't fix it" before — that was likely flags not propagating through cargo or stale incremental artifacts; the current framework's per-target rustflags plumbing is worth re-testing. But please treat "the fix works" as unverified until it's reproduced and confirmed on the hardware.

**3. If an arch genuinely can't build a tool, exclude just that tool for that arch** using the pattern these Makefiles already use (`ifneq ($(findstring $(ARCH),$(ARMv5_ARCHS) ...))` guards, as done for duf/gdu/tdu on PPC and dua/dutree on OLD_PPC). That gives you exactly what you want — armada370 loses only ripgrep, other ARMv7 models keep it — without the unsupported "same name, different per-arch package" mechanism.

**4. On keeping pre-built old binaries:** last resort only. Binaries you can't reproduce from source are a supply-chain and maintenance liability; prefer a target fix or per-arch exclusion.

**5. Process / open items:**
- qoriq: investigate why the crash persists despite the gnuspe mapping (#5879) — see the three checks above. This is the higher-value question right now.
- #6145 (custom rust versions) is still open; the framework only special-cases a few per-arch pins (armv5 1.77.2, qoriq 1.82.0), so "build with an older rust" is not a general lever yet.
- Confirm the armada370 d32 repro on current master, then land the target/flags fix.

Bottom line: **the split is the wrong tool for this bug; per-arch build-target fixes + per-tool `ifneq` exclusion are the right ones — but neither crash is confirmed fixed yet, so the immediate step is reproduction and diagnosis, not restructuring.**

mreid-tt:
Hey @hgy59, I've asked my AI to compare your work with the work I removed from the #7367 that was in conflict. Here's some suggested improvements to your docs:

---

Thanks for the `synocli-file` doc rework — the per-tool links + licenses and the General/Rust grouping look great. I reviewed it closely (and verified all the URLs resolve). A few small amendments to consider before merge:

### Worth fixing

1. **Typo**: `[lzip/plizp]` → `[lzip/plzip]` — "plizp" is misspelled (the tool is `plzip`).
2. **Stray backslash**: `It\'s` → `It's` in the mc row. An apostrophe isn't an escapable character in Markdown, so `\'` renders literally as `It\'s` with a visible backslash (artifact from the old wiki).
3. **Trailing space in link text**: `[zlib-license ](...)` → `[zlib-license](...)` in the pigz row.

### Minor / cosmetic

4. "Diskstation" → "DiskStation" (Important note).
5. "filename extentions" → "filename extensions" (file row, carried over from the wiki).
6. "Tools built with rust" → "Tools built with Rust".
7. Stray double-spaces in a few table cells (fdupes, nnn, pixz rows) and trailing spaces at line ends.

### Optional

8. License column consistency — most entries are plain text but a few are linked (file, pigz, pixz). Either style is fine; keeping it uniform would be nice (plain text unless the license isn't a familiar SPDX name, which is what the other synocli pages do).
9. The `mg` description repeats "Micro (GNU) emacs-like text editor" in both the link text and the description — could drop the duplicate line.

No blockers — these are all small polish items. Happy to drop the separate synocli-file PR I had prepared and defer to this one.

hgy59:
@mreid-tt done.

except the optional:
8. already kept it as uniform as possible (OK, now I see that one 2-Clause BSD is with a link) - I will review it 
9. already removed it in the Tools column (i.e. moved it from there to the Description to avoid overflow)

Regarding the licenses:
pigz, pixz have only a text that is a copy of the zlib license.
found that the license of `file` is  `2-Clause BSD simplified`
The name `2-Clause BSD` and `2-Clause BSD simplified` have the SPDX-identifier `BSD-2-Clause`, what do you expect in the documentation?

So far I prefered the license as shown in the github repository overview when available.

th0ma7:
Related to the extra rustc build flags i did identify a bug in the current rustc toolchain management.  I have a fix pending in my pr on the go in #7353 not committed yet.  I could then include the extra rustflags properly where needed, just need a few more cycles to close loose ends.  Let me know if this is an avenue you're ok with.

mreid-tt:
Hey @hgy59, thanks for incorporating the suggested documentation changes.

As for the license column, I don't have a strong opinion either way. It was only raised because there appeared to be some inconsistencies, but nothing is ever going to be perfect. At the end of the day, your best judgment on how to represent the licensing information is all we can really ask for.

hgy59:
One of the final changes broke the package (mix of SPK_COMMANDS and SPK_USR_LOCAL_LINKS seems not to work).
Therefore, the released packages have been withdrawn.

PR #7138, opened 2026-05-12
fix: five form fields (frontend_name, frontend_ip, f... in...

orbisai0security:
## Summary
Fix high severity security issue in `spk/haproxy/src/haproxy-dashboard/routes/main_routes.py`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-003 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-003` |
| **File** | `spk/haproxy/src/haproxy-dashboard/routes/main_routes.py:11` |

**Description**: Five form fields (frontend_name, frontend_ip, frontend_port, lb_method, protocol) are read directly from HTTP form submissions without any sanitization or validation before being written into the HAProxy configuration file. An attacker can inject newline characters and HAProxy configuration directives into any of these fields, adding unauthorized frontends, backends, or ACL rules to the load balancer configuration.

## Changes
- `spk/haproxy/src/haproxy-dashboard/routes/main_routes.py`

## Verification
- [x] Build passes
- [x] Scanner re-scan confirms fix
- [x] LLM code review passed

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/spk/haproxy/src/haproxy-dashboard/routes/main_routes.py
+++ b/spk/haproxy/src/haproxy-dashboard/routes/main_routes.py
@@ -8,11 +8,11 @@ main_bp = Blueprint('main', __name__)
 @requires_auth
 def index():
     if request.method == 'POST':
-        frontend_name = request.form['frontend_name']
-        frontend_ip = request.form['frontend_ip']
-        frontend_port = request.form['frontend_port']
-        lb_method = request.form['lb_method']
-        protocol = request.form['protocol']
+        frontend_name = request.form['frontend_name'].replace('\n', '').replace('\r', '')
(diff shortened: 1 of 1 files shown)


orbisai0security:
Thanks, this is a much better placement for the fix.

I agree that sanitising at the HAProxy config write boundary is stronger than doing it only in `main_routes.py`, since it protects all current and future callers of `update_haproxy_config()`. Covering the additional fields and backend server values also addresses the broader class of config-injection cases better than my initial five-field patch.

I’m happy for your commitment to supersede mine, or I can rebase/drop my original commit if that makes the PR history cleaner. Thanks for taking the report seriously and improving the fix.

PR #7148, opened 2026-05-20
HAProxy: Update to v3.0.23

mreid-tt:
## Description

This PR includes the following:

1. Update to HAProxy 3.0.23.
2. Migrate to Python 3.14.
3. Fix missing authentication on dashboard routes.

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [x] Package update

PR #7008, opened 2026-03-04
Standardise and rename meta spk mk files (videodrv, ffmpeg, python)

th0ma7:
## Description

Objective of this PR is to standardise and rename meta spk mk files (videodrv, ffmpeg, python) such as:
```
spksrc.spk.mk
spksrc.spk-meta.mk
spksrc.spk
├── base.mk
├── ffmpeg.mk
├── python.mk
└── videodriver.mk
```
With this new structure:
- any combination of ffmpeg, python and videodriver can occur
- all use exactly the same code for preparing the basis of reusing an existing pre-built package structure
- unique include `spksrc.spk-meta.mk` which in turn includes default `spksrc.spk.mk`
- Reduce number of direct-entry points in `mk/*` and use common-naming for spk related entries.
- Fixes most of the duplicate `RPATH` entries in resulting binaries and libraries.

## Sub-PR

- Relates to restructuring umbrella PR #6984 being broken-up into smaller PR.
- This PR also led into enhancing `spksrc.dependency-tree.mk` functionalities (ref.: #7028)
- This PR was sub-divided in smaller PR, dividing `openssl`, `python`, `videodriver`, `ffmpeg` (ref.: #7041)

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Bug fix
- [ ] New Package
- [ ] Package update
- [x] Includes framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #7152, opened 2026-05-22
Mozilla Sync Server: Update to v0.23.0

mreid-tt:
## Description

1. Update Mozilla Sync Server package to v0.23.0.
2. Migrate to Python 3.14.

Fixes # 

## Checklist

### Type of change

- [x] Package update


mreid-tt:
hey @th0ma7 — as noted in https://github.com/SynoCommunity/spksrc/pull/7148#issuecomment-4508512858, this is now the second package where Python was defined as a build-time dependency but not listed as a dependent package in the Makefile. This pattern seems to trace back to #7008.

My question is whether the framework was intended to automatically inject Python into `SPK_DEPENDS`, or if the removal of Python from the package dependencies was simply an oversight.

th0ma7:
The framework cannot inject of python unless specified currently.  In theory for packages needing only pure python wheels we could adapt the framework to accomodate that, probably relatively easily.

So indeed this is "partially" an oversight..

mreid-tt:
hey @th0ma7 — thanks for the explanation. I’ll keep an eye out for these cases and re-add the dependencies as I work through package updates. I’m also migrating packages to Python 3.14 as part of that process.

For this package specifically, I noticed another oddity: when the only include is

```make
include ../../mk/spksrc.spk-meta.mk
```

the build fails if that line appears below the `.PHONY:` declaration. Moving the include above `.PHONY:` resolved the issue for me. Another small anomaly worth keeping in mind.

mreid-tt:
hey @th0ma7, even more confusing, looking at the build run (https://github.com/SynoCommunity/spksrc/actions/runs/26292520442) from before I re-added the python dependency, the `INFO` file seems to be correct:
```
install_dep_packages="python314:MariaDB10"
```
What am I missing here?

mreid-tt:
hey @th0ma7, further to my last I believe this can be fixed in the framework as analysed in https://github.com/SynoCommunity/spksrc/pull/7154#issuecomment-4521025308.

PR #7126, opened 2026-05-05
SABnzbd: Update to v5.0.3

mreid-tt:
## Description

This PR includes the following:

1. Update to SABnzbd 5.0.3
2. Migrate to Python 3.14

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #7154, opened 2026-05-22
SABnzbd: Add python314 runtime dependency

mreid-tt:
## Description

This is a follow-on from #7126 to fix the missing `python314` runtime dependency.

Fixes # 

## Checklist

### Type of change

- [ ] Bug fix

### DIFF
--- a/spk/sabnzbd/Makefile
+++ b/spk/sabnzbd/Makefile
@@ -1,9 +1,10 @@
 SPK_NAME = sabnzbd
 SPK_VERS = 5.0.3
-SPK_REV = 80
+SPK_REV = 81
 SPK_ICON = src/sabnzbd.png
 
 PYTHON_PACKAGE = python314
+SPK_DEPENDS = "${PYTHON_PACKAGE}"
 
 DEPENDS = cross/ionice cross/coreutils cross/par2cmdline-turbo cross/7zz cross/sabnzbd
 OPTIONAL_DEPENDS = cross/unrar7 cross/unrar


mreid-tt:
hey @th0ma7 — now I’m genuinely confused. I opened this PR to restore the missing `SPK_DEPENDS`, but when I checked the original builds from #7126, they already seem to contain the following in the generated `INFO`:

```
install_dep_packages="python314"
```

Given that, how is this working without `SPK_DEPENDS` being explicitly defined? I would have expected that dependency line to be required for it to appear there.

mreid-tt:
hey @th0ma7, I asked an AI to analyse this. Let me know if this looks correct...

## Analysis: PYTHON_PACKAGE Runtime Dependency in SPK Packages

### Background

When a SPK Makefile sets `PYTHON_PACKAGE = python314`, the question is whether the Python SPK ends up in `install_dep_packages` (the INFO file field that causes DSM to install it at runtime) automatically, or whether it must be declared explicitly in `SPK_DEPENDS`.

### How `install_dep_packages` is Set

In `mk/spksrc.spk.mk` (line 292-293):
```makefile
ifneq ($(strip $(SPK_DEPENDS)),)
    @echo install_dep_packages=\"$(SPK_DEPENDS)\" >> $@
endif
```

`install_dep_packages` is sourced only from `SPK_DEPENDS`.

### Original Auto-Injection (PR #7008 — Unreliable)

PR #7008 introduced a mechanism in `mk/spksrc.spk/python.mk` to auto-add `${PYTHON_PACKAGE}` to `SPK_DEPENDS` during the `spk-stage2` build phase:

```makefile
ifneq ($(and $(wildcard $(PYTHON_PACKAGE_WORK_DIR)),$(filter spk-stage2,$(MAKECMDGOALS))),)
  ...
  $(eval $(call SPK_BASE_TEMPLATE,PYTHON))
```

The `SPK_BASE_TEMPLATE` (in `mk/spksrc.spk/base.mk`, line 72) prepends `${PYTHON_PACKAGE}` to `SPK_DEPENDS`.

However, this only works when **both** conditions are met:
1. `spk-stage2` is in `MAKECMDGOALS` (true during `LOG_WRAPPED` recursive make)
2. `$(PYTHON_PACKAGE_WORK_DIR)` exists (Python was previously built for that arch)

Condition 2 is unreliable. In CI's `all-supported` flow, Python is built first, so the work directory exists before dependent packages. But standalone `make arch-x64-7.2` builds skip this entirely.

### Test Result: HAProxy Without `SPK_DEPENDS`

A clean build of HAProxy **without** `SPK_DEPENDS` produced an INFO file with **no** `install_dep_packages` entry — Python was not listed as a runtime dependency.

### Framework Fix (PR #7155)

PR #7155 adds the same injection to the `else` branch of `python.mk`, which runs during the normal (non-`spk-stage2`) build path:

```makefile
else
  export PYTHON_PACKAGE
  export PYTHON_PACKAGE_DIR
  BUILD_DEPENDS := $(call uniq,$(PYTHON_DEPENDS) $(BUILD_DEPENDS))
  SPK_DEPENDS := $(call dedup,$(PYTHON_PACKAGE):$(SPK_DEPENDS),:)   # <-- added
endif
```

This ensures `${PYTHON_PACKAGE}` is reliably added to `SPK_DEPENDS` for all build scenarios.

### Verified

A clean build of HAProxy without explicit `SPK_DEPENDS` after the framework fix produced:
```
install_dep_packages="python314"
```

### Impact

- **Packages with explicit `${PYTHON_PACKAGE}` in SPK_DEPENDS** (e.g., rutorrent, haproxy pre-fix): The framework now handles this — the explicit entry is redundant but harmless (dedup removes the duplicate).
- **Packages without any SPK_DEPENDS**: Python is now automatically declared as a runtime dependency.
- **Packages with `SPK_DEPENDS` that omit Python**: Python is now injected correctly.

### Redundant Workarounds to Remove

| Package | PR | Redundant `SPK_DEPENDS` Line |
|---------|-----|-----------------------------|
| haproxy | #7148 | `SPK_DEPENDS = "${PYTHON_PACKAGE}"` — removed in PR #7155 |

### Remaining Packages No Longer Needing Individual Fixes

All packages that set `PYTHON_PACKAGE` but don't include `${PYTHON_PACKAGE}` in `SPK_DEPENDS` now get it automatically from the framework. No individual Makefile changes needed.

PR #7155, opened 2026-05-22
Framework: Reliably inject PYTHON_PACKAGE into SPK_DEPENDS

mreid-tt:
### Problem

The `spksrc.spk/python.mk` framework has a mechanism to auto-add `${PYTHON_PACKAGE}` to `SPK_DEPENDS` during the `spk-stage2` build phase. However, this only works when both:
1. `spk-stage2` is in `MAKECMDGOALS` (only true during the recursive `LOG_WRAPPED` make call)
2. `$(PYTHON_PACKAGE_WORK_DIR)` exists (i.e., Python was previously built for that architecture)

Condition 2 is unreliable. In CI's `all-supported` flow, Python is built first, so the work directory exists before dependent packages are built. But standalone `make arch-x64-7.2` builds skip this entirely, leaving packages without Python as a declared runtime dependency in `install_dep_packages`.

This gap was worked around in PR #7148 by adding an explicit `SPK_DEPENDS = "${PYTHON_PACKAGE}"` to the HAProxy Makefile.

### Fix

Added `SPK_DEPENDS := $(call dedup,$(PYTHON_PACKAGE):$(SPK_DEPENDS),:)` to the `else` branch of `python.mk`, which runs during the normal (non-`spk-stage2`) build path. This ensures `${PYTHON_PACKAGE}` is reliably prepended to `SPK_DEPENDS` for all build scenarios, not just when Python happens to have been pre-built.

### Changes

1. **`mk/spksrc.spk/python.mk`** — Add PYTHON_PACKAGE injection to the normal build path
2. **`spk/haproxy/Makefile`** — Remove now-redundant explicit `SPK_DEPENDS` (the framework handles it)

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/mk/spksrc.spk/python.mk
+++ b/mk/spksrc.spk/python.mk
@@ -40,4 +40,5 @@ else
   export PYTHON_PACKAGE
   export PYTHON_PACKAGE_DIR
   BUILD_DEPENDS := $(call uniq,$(PYTHON_DEPENDS) $(BUILD_DEPENDS))
+  SPK_DEPENDS := $(call dedup,$(PYTHON_PACKAGE):$(SPK_DEPENDS),:)
 endif
--- a/spk/haproxy/Makefile
+++ b/spk/haproxy/Makefile
@@ -7,7 +7,6 @@ DEPENDS = cross/haproxy
 
 PYTHON_PACKAGE = python314
 WHEELS = src/requirements-crossenv.txt src/requirements-pure.txt
-SPK_DEPENDS = "${PYTHON_PACKAGE}"
 
 MAINTAINER = Diaoul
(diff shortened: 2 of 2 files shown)


th0ma7 (approved):
LGTM

th0ma7:
Indeed this should resolve the issue.

This is also part of what @hgy59 wanted me to revisit in a subsequent PR:
1. To not allow building python packages outside from meta with a pre-built python (e.g. replacing the else part with a $(error msg)
2. Rework the injection of the `SPK_DEPEND` ... I'll need to revisit that as I don't fully recall the details.

Recently added and may be related (from memory, haven't had much time recently to work on spksrc):
3. allow building packages that only requires native python wheels without needed to pre-build a full python stack.

PR #7195, opened 2026-06-10
NZBGet: Update package to correctly install and update NZBGet on older kernels (< 3.2)

phnzb:
## Description

In v27 (the current testing release), we are dropping support for some older systems. This includes devices running the Linux 2.6 kernel and devices based on certain platforms, such as mipsel, mipseb, and ppc500.

This update changes the package installation behavior on pre-3.2 kernels as follows:

1. Testing release installation is completely blocked.
2. Stable release installation installs a pinned version (26.3 - the last version supporting pre-3.2 kernels) and disables the internal updater to prevent updates to unsupported versions.

This update also replaces the package icon to match the current NZBGet icon.

## Checklist

- [X] Package upgrade completed successfully (Manually install the package again)
- [X] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [X] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [X] Package update


Safihre:
@mreid-tt Want to chime in on this one?

@phnzb instead of having the version in the wizard, which might be confusing for users (at least it was for me), why not just fix the version to specific one you know works on those older devices? 
And what if users on older devices still fill an unsupported version, the install will fail with a generic error, do they know what to do instead?

phnzb:
@Safihre 
In general, we want users to run the latest stable/testing versions, and this change does not affect that.

We can't test the package on all older devices and guarantee that nothing will break. This applies not only to future releases that may drop support for certain architectures, but also to older versions. This update simply gives users the ability to downgrade the NZBGet version if something goes wrong.

Unfortunately, we can't display any messages if the updated version fails to start. In that case, the NZBGet package simply stops, and the user can only downgrade it manually.

Pinning to a specific version is not a universal solution. Users may want to install a particular version for their own reasons and avoid updates through the NZBGet updater.

mreid-tt:
> @mreid-tt Want to chime in on this one?

**Alternative approach: auto-detect the right version**

Instead of adding a version number dropdown (which most users won't know how to pick from), a simpler UX would be to keep the existing stable/testing wizard choice but add automatic kernel detection as an override:

```bash
KERNEL=$(uname -r)
case "$KERNEL" in
    2.6*) NZBGET_VERSION="24.5" ;;
    *)
        # Use the user's wizard choice (stable or testing)
        if [ -n "${wizard_testing_release}" ] && [ "${wizard_testing_release}" = true ]; then
            NZBGET_VERSION="latest-testing"
        else
            NZBGET_VERSION="latest"
        fi
        ;;
esac

wget --quiet --output-document="${NZBGET_INSTALLER}" \
    "https://nzbget.com/download/nzbget-${NZBGET_VERSION}-bin-linux.run"
```

This way:
- Old hardware (Linux 2.6) automatically gets a compatible pinned version
- Modern hardware respects the user's stable/testing preference from the wizard — no change in UX
- No confusing version number dropdown

What do you think?

phnzb:
@mreid-tt 
1. The reference to the kernel version isn’t entirely accurate. While NZBGet 27+ will most likely require Linux kernel 3.2+ - although this has not been finalized yet - it will also drop support for some platforms entirely, so automatically detecting the correct version to install is not a trivial task. In addition, the NZBGet updater does not support updating to a specific version. In other words, if we pin the package to an older version, users will only be able to update directly to the latest release. If the latest version is incompatible, they can only roll back to the pinned version. Rolling back to any other version-for example, the previous release - will not be possible.

2. From a UI perspective, this solution may not be ideal due to certain limitations of the Synology SDK. However, the primary goal is to give advanced users the ability to install any NZBGet release version they choose. Automatic detection does not solve this problem.

mreid-tt:
Typically, when we upgrade packages and certain architectures are no longer supported upstream, we either constrain the application's internal update mechanism for those architectures (as we do with Lidarr, Radarr, and Prowlarr) or exclude them entirely so users do not see update notifications for releases they cannot run.

In the first case, the package remains on the version we provide and users are prevented from updating beyond it through the application's own updater, as doing so would result in a broken installation.

I'm not familiar enough with NZBGet's update mechanism to know whether a similar approach is feasible there, but it's certainly something worth considering.

In general, we should be striving to simplify the end-user experience as much as possible. That's the reasoning behind the alternative I suggested for consideration.

phnzb:
@mreid-tt

That's reasonable, thank you! I have updated the package with the following changes:

1. The stable installation path now relies on the kernel version. If the kernel version is lower than 3.2, a pinned version will be installed (currently 26.2, the latest release that supports all devices). This ensures that, with the default installation, users always get a working package.

3. The testing and specific-version installation paths are unchanged. Users can still install the latest testing release or any specific version they choose.

This change will not affect future default installations while also allowing users to roll back by reinstalling any specific version they choose.

Please take a look.

Safihre:
What if users with older kernels use the auto updater from within nzbget? Or will it not offer the update to them?
Guess that's not SynoCommunity specific, but still.

phnzb:
NZBGet's updater does not support automatic updates. Its built-in updater is designed to prevent silent updates, so every update requires explicit user confirmation.

However, users on older kernels will still be able to upgrade to a package that fails to start, and then have to downgrade it manually.

To prevent this on older kernels, we can:

* Disable update checks (users will still be able to update manually).
* Disable the updater completely (users will still see that a new version is available, but they won't be able to update through the updater).

@mreid-tt

Please let me know if you think either of these changes should be made.
I think that current proposal is great balance for new users and old users and we can keep it like proposed. But I'm open for feedback.

mreid-tt:
Hey @phnzb, I'm not sure what has changed. From what I can see, the install wizard is still asking the user to select a specific version, which I previously mentioned is likely to be confusing for new users.

Regarding the concern about preventing broken installations, disabling the application's internal updater when users cannot safely upgrade is the same approach we already take with the `*arr` applications. In those cases, newer releases bundle a version of SQLite that is incompatible with older DSM versions. Rather than allowing users to upgrade into a broken state, we package a compatible version of SQLite and disable the application's auto-update mechanism so it cannot reintroduce the incompatible SQLite version (see #6759).

My broader point is that we should take a step back and think about the overall user journey. Whenever possible, we should aim to make the experience as simple as possible for new users and avoid requiring them to make technical decisions or recover from situations that we could have prevented through the package design.

phnzb:
@mreid-tt

Thank you for feedback. Based on it - would you be OK if we:

- Remove the Specific Version installation option completely, since it may confuse new users,
- Latest Stable installation path: on older kernels, install the pinned version and disable updates completely,
- Latest Testing installation path: on older kernels, block installation completely, since v27+ will not run on those systems,

With these changes, advanced users still won't be able to roll back or install a specific version (which was the original goal of this PR), but the overall user experience will be preserved.

Safihre:
I like that much better.

phnzb:
@mreid-tt I have implemented everything from the previous post and also updated the PR title and description. Please review.

Safihre:
@mreid-tt to me it looks good, I'd say release? 😊

mreid-tt (approved):

PR #7162, opened 2026-05-25
Borg: Update to v1.4.4

mreid-tt:
## Description

This PR includes the following:

1. Update Borg to v1.4.4.
2. Update borgmatic to v2.1.5.
3. Migrate to Python 3.14.

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #7201, opened 2026-06-12
Borg: Fix lz4 header detection after Python 3.14 migration

mreid-tt:
### Problem
After migrating borgbackup from Python 3.11 to Python 3.14, CI builds fail on all architectures with:

```
src/borg/compress.c:1253:10: fatal error: lz4.h: No such file or directory
```

The `cross/lz4` package is built during python314's stage-1 (installing headers to python314's prefix) and is not rebuilt for borgbackup's stage-2 due to dependency caching. The existing `BORG_LIBLZ4_PREFIX=$(STAGING_INSTALL_PREFIX)` pointed to borgbackup's own staging prefix where `lz4.h` was never installed. Borgbackup's setup.py then fell back to host `pkg-config`, which found the CI image's `liblz4-dev`, but the cross-compiler couldn't locate the host's `/usr/include/lz4.h` because it resolves paths against the cross-sysroot.

### Fix
Point `BORG_LIBLZ4_PREFIX` and `PKG_CONFIG_PATH` to the Python install prefix where `cross/lz4` was actually installed during stage-1:

```makefile
_prefixed := $(abspath $(WORK_DIR)/../../../spk/$(PYTHON_PACKAGE)/work-$(ARCH)-$(TCVERSION)/install/var/packages/$(PYTHON_PACKAGE)/target)
ENV += PKG_CONFIG_PATH=$(_prefixed)/lib/pkgconfig
ENV += BORG_LIBLZ4_PREFIX=$(_prefixed)
```

Also reverts the earlier `cross/lz4/PLIST` header change — headers are not typically tracked in PLIST files for cross packages.

### Files changed
- `spk/borgbackup/Makefile` — point `BORG_LIBLZ4_PREFIX` and `PKG_CONFIG_PATH` to Python install prefix

Fixes #7162

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/cross/lz4/PLIST
+++ b/cross/lz4/PLIST
@@ -1,3 +1,6 @@
+rsc:include/lz4.h
+rsc:include/lz4hc.h
+rsc:include/lz4frame.h
 lnk:lib/liblz4.so
 lnk:lib/liblz4.so.1
 lib:lib/liblz4.so.1.10.0
--- a/cross/lz4/PLIST
+++ b/cross/lz4/PLIST
@@ -1,6 +1,3 @@
-rsc:include/lz4.h
-rsc:include/lz4hc.h
-rsc:include/lz4frame.h
 lnk:lib/liblz4.so
 lnk:lib/liblz4.so.1
 lib:lib/liblz4.so.1.10.0
--- a/spk/borgbackup/Makefile
+++ b/spk/borgbackup/Makefile
@@ -74,9 +74,13 @@ SPK_USR_LOCAL_LINKS = bin:env/bin/borg bin:env/bin/borgmatic bin:env/bin/emborg
 # spksrc.wheel-env.mk or spksrc.spk-meta.mk
 ENV += BORG_LIBACL_PREFIX="$(STAGING_INSTALL_PREFIX)"
 ENV += BORG_LIBB2_PREFIX="$(STAGING_INSTALL_PREFIX)"
-ENV += BORG_LIBLZ4_PREFIX="$(STAGING_INSTALL_PREFIX)"
 ENV += BORG_LIBXXHASH_PREFIX="$(STAGING_INSTALL_PREFIX)"
 ENV += BORG_LIBZSTD_PREFIX="$(STAGING_INSTALL_PREFIX)"
+# BORG_LIBLZ4_PREFIX: prefer the Python install prefix (where lz4 was
+# installed during stage-1), fall back to own staging prefix.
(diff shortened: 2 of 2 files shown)


th0ma7:
Part of the meta package processing exclude some basic libraries that do not always behave properly.  In this case it could be a matter of adding it to your local depend or using $(or puthon staging install prefix, staging install prefix).

There are other cases like this i believe...  Not on my build system so can't confirm for sure.

th0ma7:
Instead of _prefix it should be PYTHON_STAGING_INSTALL_PREFIX

mreid-tt:
> Instead of _prefix it should be PYTHON_STAGING_INSTALL_PREFIX

I asked the AI about this and it said:

## Analysis of `PYTHON_STAGING_INSTALL_PREFIX` vs `_prefixed`

### Question
Can we replace `_prefixed` with `PYTHON_STAGING_INSTALL_PREFIX`?

### Investigation

`PYTHON_STAGING_INSTALL_PREFIX` is defined in two places in the framework:

- `mk/spksrc.crossenv.mk:133`: `ifeq ($(PYTHON_STAGING_INSTALL_PREFIX),) PYTHON_STAGING_INSTALL_PREFIX = $(STAGING_INSTALL_PREFIX) endif`
- `mk/spksrc.wheel-env.mk:38`: `PYTHON_STAGING_INSTALL_PREFIX = $(STAGING_INSTALL_PREFIX)`

In both cases it is unconditionally set to **`$(STAGING_INSTALL_PREFIX)`** — which for the borgbackup SPK build resolves to `/var/packages/borgbackup/target`. It is **never** set to the actual Python package install prefix (e.g., `/var/packages/python314/target`).

This means using `PYTHON_STAGING_INSTALL_PREFIX` directly would have the **same problem** as the original `BORG_LIBLZ4_PREFIX="$(STAGING_INSTALL_PREFIX)"` — it points to borgbackup's staging prefix where lz4 headers were never installed (lz4 was installed during python314's stage-1 and cached, not rebuilt for borgbackup's stage-2).

### How `_prefixed` differs

Our `_prefixed` variable explicitly computes the Python package's install path:

```makefile
_prefixed := $(abspath $(WORK_DIR)/../../../spk/$(PYTHON_PACKAGE)/work-$(ARCH)-$(TCVERSION)/install/var/packages/$(PYTHON_PACKAGE)/target)
```

This resolves to something like:
`.../spk/python314/work-x64-7.1/install/var/packages/python314/target`

Which is where lz4 headers were actually installed during stage-1.

### Conclusion

`PYTHON_STAGING_INSTALL_PREFIX` is not a drop-in replacement. To make it work, the framework would need to explicitly set it to the Python package's install prefix rather than defaulting to `$(STAGING_INSTALL_PREFIX)`. That could be a useful framework improvement, but for this fix the current approach is correct.

PR #7223, opened 2026-06-22
Framework/meta: resolve cmake deps via the toolchain file (OPENSSL_ROOT_DIR + CMAKE_FIND_ROOT_PATH)

th0ma7:
## Summary

Second in the stacked series splitting the meta cross-dependency refactor (#7218). **All the cmake changes** — framework + every impacted cross package — in one PR, and the cmake channel is **fully meta-aware on its own** (no need to wait for the cross-env.mk hoist).

> **Stacked on #7222** (meta env foundation, now merged). CMAKE_FIND_ROOT_PATH and the ordered PKG_CONFIG_LIBDIR consume the `META_PKGCONFIG_DIRS` introduced there.

### Framework (`mk/spksrc.cross-cmake-toolchainfile.mk`)
- **cmake-local ordered `PKG_CONFIG_LIBDIR`** — local staging first, then the meta pkgconfig dirs. Drives both the baked `set(ENV{PKG_CONFIG_LIBDIR})` and the OpenSSL scan, so cmake builds share meta libs immediately. *(PR #3 hoists this ordering into `cross-env.mk` for autotools/meson too and reverts the toolchain file to the shared var.)*
- **`OPENSSL_ROOT_DIR`** baked centrally — `find_package(OpenSSL)` ignores pkg-config, so derive its root from the first ordered `PKG_CONFIG_LIBDIR` dir providing `libssl.so` (local-first), else local staging.
- **`CMAKE_FIND_ROOT_PATH`** gets the meta staging roots — `find_library/find_path/find_package` re-root under it (MODE=ONLY), so shared meta libs (absolute paths) must be listed there.
- colon-list aware `PKG_CONFIG_LIBDIR` baking — `$(abspath)` mangles a `:`-separated list; make each element absolute before re-joining.

### Impacted cross packages (drop per-package `-D` flags, now resolved centrally)
- OpenSSL: `libssh2`, `librabbitmq`, `mariadb-connector-c-latest`, `mariadb-connector-c-3.3.11`, `znc`
- via `CMAKE_FIND_ROOT_PATH`: `chromaprint` (16 `-DFFMPEG_LIBAV*` → 0), `dbus` (`-DEXPAT_*`)

### Out of scope (later PRs)
- autotools openssl consumers (`nmap`, `zabbix-agent`) and `spk/tvheadend` → the "rest" PR
- `boost` → already in #7220
[...]

PR #7226, opened 2026-06-23
CI: build-time budget + checkpoint to survive the 6h job kill

th0ma7:
## Problem

Per-arch package builds can exceed GitHub's hard **6h job limit**, which kills the job before the `if: always()` steps upload the completed `.spk` and diagnostics — so a long build loses everything *and* gives no indication of what was left to do.

## Mechanism

A per-arch **build-time budget** (`BUILD_BUDGET_SECONDS`, default **20700 = 5h45m**, below the 6h kill) enforced as a **hard cap**:

- **`.github/actions/build.sh`** — each package build runs under `timeout -k 60 <remaining-budget>`, so the **in-progress build is killed at the ceiling** instead of overrunning (a boundary check alone would let the current package finish and exceed 6h). The killed package is recorded in `build_remaining.txt`, and `build_timeout.txt` captures `stopped_during`/`stopped_before` + the remaining list.
- **`.github/actions/build_status.sh`** — reports **SUCCESS / UNSUPPORTED / NOT PROCESSED / ERRORS** in one place, names the package **interrupted mid-build**, and `exit 1` on a partial build (or real errors) so the job indicator goes **red** instead of a misleading green.
- **`.github/workflows/build.yml`** — passes the budget + checkpoint file paths, sets a `timeout-minutes: 350` backstop on the build step, and the existing `if: always()` upload steps still run on a budget stop → **completed `.spk` and diagnostics are preserved**, plus a `::warning::` annotation flags the partial build (naming the interrupted package).

## Notes

- `BUILD_BUDGET_SECONDS: 0` disables the budget.
- This PR only touches `.github/` so it does not trigger the heavy build workflow itself; the mechanism is exercised on the next real package build (or a manual dispatch). Extracted from the meta cmake work to land as a standalone CI improvement.

### DIFF
--- a/.github/actions/build.sh
+++ b/.github/actions/build.sh
@@ -134,7 +134,44 @@ fi
 # Initialize remaining packages list for tracking build progress
 remaining_packages="${build_packages}"
 
+# Build-time budget: stop at a package boundary before GitHub's 6h job kill so
+# the completed .spk and diagnostics still upload, and so we can report what was
+# left to build. Tunable via BUILD_BUDGET_SECONDS (default 5h45m); 0 disables it.
+build_budget=${BUILD_BUDGET_SECONDS:-20700}
+build_start=$(date +%s)
+: "${BUILD_REMAINING_FILE:=build_remaining.txt}"
+: "${BUILD_TIMEOUT_FILE:=build_timeout.txt}"
+budget_hit=0
+
 for package in ${build_packages}; do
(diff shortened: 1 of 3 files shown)


th0ma7:
@mreid-tt @hgy59 i was finally able to solve this long standing issue where we would loose artifacts when a build exceed the max duration.  Have a look at #7223 where there is a clean example of this use case.

I'll finish reviewing this over the weekend and merge it, but i thought this was worth letting bringing your attention to this.

PR #7214, opened 2026-06-18
Update GitHub workflows

hgy59:
## Description

- partial revert of https://github.com/SynoCommunity/spksrc/pull/7206
- update github actions to latest versions

Fixes # 

## Checklist

- [ ] Build rule `all-supported` completed successfully
- [ ] New installation of package completed successfully
- [ ] Package upgrade completed successfully (Manually install the package again)
- [ ] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Includes small framework changes

PR #7238, opened 2026-06-26
ci: only cancel superseded runs and skip fork builds while a PR is draft

th0ma7:
## TL;DR

Two build behaviors are now gated on **PR Draft state** — **nothing changes outside draft**:

- **While a PR is a Draft** → superseded runs are cancelled (iterative pushes don't pile up) and the redundant fork build is skipped.
- **Non-draft PRs, `master` pushes, and manual runs** → unchanged: every build runs to completion.

Plus a small fix: the change-detection regex is anchored (`^`) so the new `mk/spksrc.cross/` / `mk/spksrc.native/` dirs aren't mis-detected as `cross/` / `native/` package folders.

---

## Background

While exercising heavy branches, the Build workflow piled up overlapping, multi-hour runs. Two compounding causes:

- **No `concurrency` guard** — successive pushes never cancelled the previous, superseded run, so several full arch matrices ran at once.
- **Fork + upstream builds on every push** — with an open fork PR, each push triggers a `push` build on the fork *and* a `pull_request` build upstream.

## 1. Cancel superseded runs (draft PRs only)

```yaml
concurrency:
  group: build-${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft }}
```

Concurrency is inherently per-repository (a fork run and an upstream run can never cancel each other) and keyed per ref + event. A superseded run is cancelled **only while the pull request is a Draft**, so iterative pushes during development don't pile up. Everything else runs to completion:

- **push to `master`** (each PR merge) → never cancelled, so every merged commit keeps its own visible build result;
- **non-draft PR** builds → never cancelled;
- **`workflow_dispatch`** (manual build & publish) → never cancelled.

## 2. Gate the fork build on PR draft state

The fork's `push` build is **not** a duplicate of the upstream PR build (thanks @hgy59 for the correction): the upstream `pull_request` build covers the full PR diff (`origin/master...HEAD`), while the fork `push` build covers only the latest commit (`git diff-tree HEAD`) — complementary, not redundant. So this PR no longer removes it.

Instead, a small `gate` job decides whether to build:

- **upstream events** (`pull_request`, or push to `SynoCommunity/spksrc`) and **`workflow_dispatch`** → always build;
- a **fork `push`** → skipped **only while the branch has an open *draft* PR** upstream, and builds normally again once the PR is non-draft / ready for review.

A `push` event carries no PR context, so the gate looks it up (fail-open on any error):

```yaml
draft=$(gh api \
  "/repos/SynoCommunity/spksrc/pulls?head=${{ github.repository_owner }}:${{ github.ref_name }}&state=open" \
  --jq '[.[].draft] | first // false' 2>/dev/null || echo false)
```

Net effect:

- when a fork doesn't use draft PRs, **its fork build is unchanged** — it keeps running on every push;
- keeping a PR in draft during development stops the redundant fork build, while the **upstream full build still runs on every push**;
- (1) cancels superseded upstream runs while in draft, keeping the queue clear during development;
- manual *Run workflow* stays available throughout.

### Why this rather than the earlier options

- **(a)** skip fork auto-builds entirely (first version of this PR) — drops the useful fast narrow build; rejected.
- **(b)** trigger `push` only on `master` — same downside for everyone.
- **(c)** do nothing — heavy matrices saturate the Actions queue and waste fork runner time during development.
- **(d, chosen)** gate the fork build on draft state — keeps the narrow fork build (as restored in #7214) for everyone outside draft, and only trades it away while a PR is explicitly marked WIP. No per-fork config variable required.

## 3. Anchored change-detection regex

The `dependency_folders` detection used `grep -oP "(cross|python|native)/[^/]*"`, which matched anywhere in a path. It is now anchored as `^(cross|python|native)/...`, mirroring the existing `^spk/` anchor for `spk_packages`. This drops false positives such as `mk/spksrc.cross/env-default.mk` → `cross/env-default.mk` while keeping every real top-level package folder.

## Validation

- `build.yml` parses as valid YAML; the `gate` job, the `prepare`/`consolidate-diagnostics` wiring, the draft-conditional concurrency and the anchored regex were verified.
- Anchored regex tested locally against this repo's change set: all `mk/spksrc.cross/` / `mk/spksrc.native/` false positives drop, every real `cross/` package folder is retained.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,10 +104,24 @@ env:
   # .spk/diagnostics still upload and the remaining packages are reported.
   BUILD_BUDGET_SECONDS: "20700"
 
+# Cancel a previous, now-superseded run when a new commit lands on the same
+# branch or pull request, instead of letting both full build matrices run to
+# completion. The event_name is part of the group so push, pull_request and
+# manual workflow_dispatch runs never cancel one another; a manual build &
+# publish (workflow_dispatch) is additionally never auto-cancelled.
+concurrency:
+  group: build-${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: ${{ github.event_name != 'workflow_dispatch' }}
+
 jobs:
   prepare:
     name: Prepare for Build
     runs-on: ubuntu-latest
+    # Skip the automatic push/pull_request build on forks: a fork PR is already
+    # built by the pull_request event in the SynoCommunity/spksrc base repo, so
+    # also building it from the fork's push event would duplicate the whole
+    # matrix. workflow_dispatch on a fork is unaffected (different code path).
+    if: github.repository == 'SynoCommunity/spksrc' || github.event_name == 'workflow_dispatch'
     outputs:
       arch_packages: ${{ steps.dependencies.outputs.arch_packages }}
       noarch_packages: ${{ steps.dependencies.outputs.noarch_packages }}
@@ -848,7 +862,7 @@ jobs:
     name: Consolidate Diagnostics
     needs: build
     runs-on: ubuntu-latest
-    if: always()
+    if: always() && (github.repository == 'SynoCommunity/spksrc' || github.event_name == 'workflow_dispatch')
     steps:
       - name: Download all diagnostic artifacts
         uses: actions/download-artifact@v8
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,14 +104,15 @@ env:
   # .spk/diagnostics still upload and the remaining packages are reported.
   BUILD_BUDGET_SECONDS: "20700"
 
-# Cancel a previous, now-superseded run when a new commit lands on the same
-# branch or pull request, instead of letting both full build matrices run to
-# completion. The event_name is part of the group so push, pull_request and
-# manual workflow_dispatch runs never cancel one another; a manual build &
-# publish (workflow_dispatch) is additionally never auto-cancelled.
+# Cancel a previous, now-superseded run only while a pull request is still a
+# Draft, so iterative pushes during development don't pile up. Non-draft PR
+# builds and master/push builds are never cancelled — every merged commit keeps
+# its own build result on master — and workflow_dispatch (manual build &
+# publish) is never cancelled either. The event_name is part of the group so
+# push, pull_request and workflow_dispatch runs never cancel one another.
 concurrency:
   group: build-${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
-  cancel-in-progress: ${{ github.event_name != 'workflow_dispatch' }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft }}
 
 jobs:
   # Decide whether the build should run at all. Upstream events (pull_request,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,14 +114,47 @@ concurrency:
   cancel-in-progress: ${{ github.event_name != 'workflow_dispatch' }}
 
 jobs:
+  # Decide whether the build should run at all. Upstream events (pull_request,
+  # or push to SynoCommunity/spksrc) and manual workflow_dispatch always build.
+  # A fork's own push build is skipped while the branch has an open *draft* PR
+  # upstream: during development the authoritative full build still runs on the
+  # upstream PR, so re-running the fork's narrow build only wastes runner time.
+  # A push event carries no PR context, hence the API lookup here.
+  gate:
+    name: Gate
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


th0ma7:
@hgy59 and @mreid-tt I'd like your opinion on this PR - basically this is trying to avoid generating tons of uneeded gh-action by two methods: 
1. canceling previous running jobs on PR when new commits are added concurrently and
2. limiting fork gh-action builds in addition to master

mreid-tt (approved):

hgy59:
> A fork PR is already built by the pull_request event in the base repo, so building it again from the fork's push event ran the entire arch matrix a second time. Forks now skip the automatic run (instant, zero compute) but can still trigger a build on demand via the manual Run workflow button. The guard is on prepare (it cascades to set-defaults and build through needs:) and on consolidate-diagnostics (which uses if: always()).

That is what I do not agree.
The push of new commits trigger the fulle build by all changes on the PR but on the fork only the changes of the latest commit trigger the build.
If I don't want the build of any of the two, I always can cancel one of those.
Sometimes I want to cancel the build on the fork and sometimes the full build on the upstream.

I like the changes to cancel superseeded runs, but only within the same repository (upstream, fork)
I want to keep the "duplicate fork build" because it is not a duplicate (except for the first push).
That is what I wanted to explain earlier and restored in #7214.

th0ma7:
Thanks @hgy59 — agreed, and you've convinced me the fork build isn't a duplicate.  I've reworked the PR; revised plan:

### TL;DR
1. **kept as-is**: superseded-run cancellation already works only within the same repository (concurrency is per-repo by nature).
2. the fork build is **no longer removed** — it's only **skipped while the PR is a Draft**; mark it ready (or don't use draft) and it runs exactly as today.


### Longer-version
  **1. Cancel superseded runs — kept, scoped per repository.** Concurrency in GitHub is inherently per-repo (a fork run and an upstream run can never cancel each other), keyed per ref + event, so it only cancels an older, superseded run of the *same* kind in the *same* repo — exactly what you asked for.

  **2. Fork build — kept, but skipped only while the PR is a Draft.** You're right it isn't a duplicate: the upstream `pull_request` build covers the full PR diff (`origin/master...HEAD`), while the fork `push` build covers only the latest commit (`git diff-tree HEAD`) — complementary, not redundant. So I dropped the blanket fork guard and the "duplicate" wording.
  
  What I actually need is just to silence the fork build during active development, where it's noise and burns my fork's runner time for nothing. So instead of removing it (or adding a config variable), it's now gated on **Draft** state:
 
  - **upstream PR build → always runs**, draft or not (it stays the authoritative full CI for the PR);
  - **fork push build → skipped only while the branch has an open *Draft* PR upstream**, and runs normally again as soon as the PR is non-draft / marked ready for review.

  A `push` event carries no PR context, so a small `gate` job looks it up:
 
```yaml
  draft=$(gh api \
    "/repos/SynoCommunity/spksrc/pulls?head=${{ github.repository_owner }}:${{ github.ref_name }}&state=open" \
    --jq '[.[].draft] | first // false' 2>/dev/null || echo false)
```
Reads the upstream PR; the fork's GITHUB_TOKEN can read public PRs; any lookup failure falls back to building. If no PR exists yet, nothing changes.

Net effect:
  - when you don't use draft PRs, your fork build is unchanged — it keeps running on every push;
  - I keep my PRs in draft during development → my fork stops auto-building, while the upstream full build still runs on every push;
  - (1) keeps superseded runs from piling up in either repo;
  - manual Run workflow stays available.
 
  This keeps the narrow fork build you wanted (restored in #7214) for everyone outside draft, and only trades it away while a PR is explicitly marked WIP.

This PR now already contains the updated changes.

Does that work for you?

hgy59:
@th0ma7 IMO superseded runs on the master branch should not be cancelled. On the master branch we should see for every commit whether the build succeeded.
i.e. when a PR is merged while the build of the previous merged PR is still running, it should not be terminated.

th0ma7:
@hgy59 I take that's related to item 1.  What about the superseeded commit runs gets canceled only if in draft mode.  That would serve me real much...

And ok with the approach for item 2, wich would align well now the two of them using draft?

th0ma7:
@hgy59 Changes applied and now only impacting draft mode for both items.  I believe this is probably the best of both worlds, fulfilling your preferences along with mine to avoid filling the build queue 🚀 

That and being a little more green by reducing uneeded ghaction builds, thus reducing computational footprint.

hgy59:
@th0ma7 I guess this is not what you wanted:
it has now a concurrency rule for all branches: a new build on a branch waits until the former build is completed.
and you can't run manuel builds in parallel anymore.
May be it works as expected on PRs that are draft (i.e. cancells the former build and starts a new on a push), but for others, the behaviour changed.

th0ma7:
K, I'll have another look at it, thnx for noticing

PR #7248, opened 2026-06-29
LibYAML: Add libyaml-0.so symlink for pkg-config compatibility

mreid-tt:
## Description

`cross/libyaml` builds with `--disable-static`, resulting in `libtool` creating only `libyaml.so` and `libyaml-0.so.2` in the install prefix — but **not** `libyaml-0.so`. cmake targets that use pkg-config to find libyaml (e.g., via `yaml-0.1.pc`) get `-lyaml-0` as the linker flag, which fails because `libyaml-0.so` doesn't exist.

This adds a `POST_INSTALL_TARGET` that creates the missing `libyaml-0.so → libyaml.so` symlink, fixing the link step for downstream consumers without modifying libyaml's source or .pc files.

Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/cross/libyaml/Makefile
+++ b/cross/libyaml/Makefile
@@ -14,4 +14,10 @@ LICENSE  = MIT
 GNU_CONFIGURE = 1
 CONFIGURE_ARGS = --disable-static
 
+POST_INSTALL_TARGET = libyaml_post_install
+
 include ../../mk/spksrc.cross-cc.mk
+
+.PHONY: libyaml_post_install
+libyaml_post_install:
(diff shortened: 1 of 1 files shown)


hgy59:
@mreid-tt I guess you have to update the PLIST file too?

mreid-tt:
> @mreid-tt I guess you have to update the PLIST file too?

@hgy59, for correctness yes. I'll make another PR to fix it.

EDIT: I seem to have pushed the fix directly on `master` in https://github.com/SynoCommunity/spksrc/commit/dcb9f115def636bbf301596b82ded37cf9099675. My apologies.

PR #7254, opened 2026-07-01
download: split per-method macros and add bounded, generic mirror fallback

th0ma7:
## What

Refactor `mk/spksrc.build/download.mk` for maintainability and download resilience.

### Per-method macros
The ~125-line monolithic `download_target` (one `for url` loop with a giant inline `case` for git/svn/hg/wget) is split into `DOWNLOAD_GIT` / `DOWNLOAD_SVN` / `DOWNLOAD_HG` / `DOWNLOAD_HTTP` macros (git/svn/hg bodies moved verbatim). `download_target` just loops over `$(URLS)` and dispatches on `PKG_DOWNLOAD_METHOD`.

### Generic mirror fallback
`ftp.gnu.org`'s `ftpmirror.gnu.org` redirector can resolve to a dead mirror, and `wget` then just retries the same dead host until the build fails. Now a plain (http) download tries the original URL first, then falls back to each mirror in turn until one works, failing only if **all** fail.

A single **mirror table is the sole source of truth** — there is no per-family URL rewrite, so for GNU the geo-redirector `ftpmirror.gnu.org` is just another table entry (only SourceForge project-page URLs are still canonicalised to a downloadable host):

| Family | Bases |
|--------|-------|
| GNU | ftpmirror.gnu.org, ftp.gnu.org, mirrors.kernel.org, mirror.csclub |
| SourceForge | downloads.sourceforge.net, netcologne.dl, phoenixnap.dl |
| GNOME | download.gnome.org, mirror.csclub/gnome |
| X.org | www.x.org, mirror.csclub/x.org |
| kernel.org | cdn.kernel.org, mirrors.kernel.org, www.kernel.org |

A candidate replaces everything up to the family's tree-root marker (`/gnu/`, `/project/`, `/sources/`, `/releases/`, `/pub/`) with each base; the list is de-duplicated; `PKG_DIST_MIRRORS` adds per-package bases; non-matching URLs (e.g. GitHub) use the single original URL.

### Bounded
Each host is tried `DOWNLOAD_TRIES` (default 2) times over a finite, de-duplicated candidate list, so a failing download is bounded (≈ candidates × 2) and can never loop to the CI job limit.

## Validation

- `make -n` + `sh -n` on git / GNU / SourceForge / GNOME / kernel packages; candidate construction and de-dup verified per family.
- The fallback loop was proven end to end in CI (a dead-primary package recovered from `ftp.gnu.org` via `PKG_DIST_MIRRORS`): `==> download failed, trying next mirror` → successful fetch.

> The `demodl-*` demo packages used to exercise this live are kept in a separate PR (#7252) for archiving.

### DIFF
--- a/mk/spksrc.build/download.mk
+++ b/mk/spksrc.build/download.mk
@@ -16,6 +16,8 @@
 #                           (2 or more entries enable multi-arch orchestration)
 #  PKG_DIST_ARCH          : Optional current distribution architecture
 #                           (single element used during leaf execution)
+#  PKG_DIST_MIRRORS       : Optional per-package fallback base URLs; the file
+#                           name is appended to each and tried in turn.
 #
 # Files:
 #  $(WORK_DIR)/.$(COOKIE_PREFIX)download_done
@@ -27,6 +29,11 @@
 #
 # Notes:
 #  - The download target is idempotent and guarded by a completion cookie.
+#  - Per download method, the actual work lives in a DOWNLOAD_<METHOD> macro
+#    (git / svn / hg / http) selected by PKG_DOWNLOAD_METHOD; download_target
+#    only loops over $(URLS) and dispatches to the right macro.
+#  - For plain (http) downloads a failed mirror is retried on the next mirror
+#    (see the mirror table below) before giving up.
 #  - When PKG_DIST_ARCH is unset, a single generic cookie is used, preserving
 #    the classic single-archive behavior.
 #  - When PKG_DIST_ARCH is set, the cookie is architecture-specific, allowing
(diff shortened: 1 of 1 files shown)


th0ma7:
@hgy59 and @mreid-tt for your awareness, I got fedup from failing downloads so I tried updating the `spksrc.download.mk` in order to try-out multiple mirrors when applicable.  At the same time the code has been dividided into sub-recipes for it to be easier to maintain in the longer term.

Please let me know if you hit any issue with this update, likewise if failed download occurences during the prepare stage actually decreases.

hgy59:
@th0ma7 got another failing download:

```
===>  Downloading files for attr
===>    wget -O attr-2.5.2.tar.xz https://download.savannah.gnu.org/releases/attr/attr-2.5.2.tar.xz
https://download.savannah.gnu.org/releases/attr/attr-2.5.2.tar.xz:
Read error (Connection timed out) in headers.
===>    ==> download failed, trying next mirror
===>  *** All mirrors failed for attr-2.5.2.tar.xz. Stop.
```

unfortunately it does not try `https://download-mirror.savannah.gnu.org/releases/attr/attr-2.5.2.tar.xz`

I guess we need to enhance the mirror table?

th0ma7:
I guess so... So the mirror functionality did not kick in at all, or is the mirror not pointing at the right place?

hgy59:
> I guess so... So the mirror functionality did not kick in at all, or is the mirror not pointing at the right place?

I retry worked

savanna is not in the mirror list...
there are only mirrors for gnu, sourceforge, gnome, xorg, kernel

PR #7278, opened 2026-07-12
Immich: Add initial package v3.0.2

### LINKED ISSUE #6991: [Package Request] Immich
### Software Name

Immich

### Brief Description

Congrats for the splendid work!
Would it be an option to consider Immich as a SynoCommunity package?
I've tried running it on a remote VPS but upload time of my data makes it very, very hard to on-board.
Running Immich locally on Synology would be a great plus.
Thank you!

### Website

https://immich.app/

### Documentation

_No response_

https://docs.immich.app/overview/quick-start

### Source code

https://github.com/immich-app/immich

### OpenSource License

GPLv3

### Other License

_No response_

mreid-tt:
## Description

This PR adds Immich (v3.0.2), a Self-hosted photo and video backup solution.

### `spk/immich`
- Three worker processes managed by DSM's service framework:
  - **API server** (`dist/main.js`) — NestJS HTTP server on port 2283
  - **Microservices** — background jobs via `Worker` threads
  - **Machine Learning** (optional) — CLIP smart search, facial recognition, OCR on port 3003
- **Wizard** with three screens: PostgreSQL credentials, shared folder, optional ML
- **Uninstall wizard** with optional database export in Immich-compatible format
- Proper signal handling via `server-launcher.js` wrapper

### `cross/immich`
- Builds the Immich monorepo via pnpm (server, web, CLI)
- Patches hardcoded PostgreSQL binary paths for DSM compatibility
- Dev dependency pruning saves ~125 MB of unnecessary node_modules

### `cross/immich-geodata`
- Downloads geodata files from geonames.org and naturalearthdata.com
- Used by Immich for reverse geocoding (map view)

## Ports

| Port | Protocol | Service |
|------|----------|---------|
| 2283 | TCP | Immich API / Web UI |
| 3003 | TCP | Immich Machine Learning (optional) |

## Build Requirements
- DSM 7.2+, 64-bit only
- Dependencies: Node.js_v22, PostgreSQL >=17.10-4, Redis, Perl, Python 3.14

## Installation Notes
- PostgreSQL and Redis must be installed separately (SynoCommunity packages)
- ML is optional — downloads ~300 MB of Python deps on first install if enabled
- Media storage is a DSM shared folder — use a volume with sufficient space

Fixes #6991

## Checklist

### Type of change

- [x] New Package


mreid-tt:
Hey @th0ma7, I was wondering if you had any thoughts on implementing the ML component as an optional runtime installation rather than following the usual pattern of fully compiling it into the Python package.

From what I can tell, the current approach works as-is. For the architectures we target, the required wheels are available as `manylinux` builds, so installation should also work on ARMv8 platforms without requiring us to build them ourselves.

I initially explored making the ML component a separate package that could be properly cross-compiled, but the implementation quickly became much more complicated than I think the benefit justified.

I'd be interested to hear your thoughts on the trade-offs or whether you see any issues with the runtime installation approach.

EDIT: I targeted ffmpeg7 and was wondering if I should target ffmpeg8 instead if that is our new baseline?

th0ma7:
@mreid-tt It depends the scope of archs you intent to support and manylinux wheel availability.  In theory you could strip out entirely the PYTHON_PACKAGE and add requirement-pure.txt file listing all wheels needed... and they would be downloaded and installed at package installation time.

The thing is, if you require supporting hi3535, qoriq, sometimes even our generic armv7 or i686 you require to build and package the wheels.  So it all comes to the supported archs.

Noting that, if I end-up being able to upgrade dsm-6.2.4 toolchain, this package could be more widely supported, thus at that point requiring wheels to be pre-built and packaged.

As for ffmpeg, I suggest you target ffmpeg8 if compatible.

mreid-tt:
hey @th0ma7, thanks for the detailed feedback!

**On arch support:** Currently targeting x64 and ARMv8 (64-bit only). I'm gating behind `REQUIRED_MIN_DSM = 7.2` and excluding all `$(32bit_ARCHS)`, so manylinux wheel availability should cover those — I verified that onnxruntime, opencv, and the other ML deps all have `manylinux` wheels for both `x86_64` and `aarch64`.

The ML source (`immich_ml/`) itself isn't on PyPI — it ships as part of the Immich monorepo's `machine-learning/` subdirectory — so we can't pip install it regardless. I save it during the build and bundle it in the SPK, then the deps get pip-installed at runtime into a venv if the user opts in via the wizard. That keeps the base package lean (~300 MB savings for users who don't need ML).

**On ffmpeg:** Agreed, I'll switch to ffmpeg8. Checking compatibility now, but fluent-ffmpeg is generally version-agnostic and I don't expect issues.

th0ma7:
> I save it during the build and bundle it in the SPK, then the deps get pip-installed at runtime into a venv if the user opts in via the wizard. That keeps the base package lean (~300 MB savings for users who don't need ML).

not a bad idea, there could be alternate methods to pre-generate a wheel, and presumably it isn't compiled but its rather of a pure python type.

mreid-tt:
hey @th0ma7, I've changed the dependency to `ffmpeg8` and everything seems to be working. I can upload videos and see them transcoding in the log:
```
[Nest] 14941  - 07/14/2026, 2:39:53 PM     LOG [Microservices:MediaService] Transcoding video 47eb0810-402d-42e1-a9dd-2a51e870c0e3 without hardware acceleration
[Nest] 14941  - 07/14/2026, 2:40:28 PM     LOG [Microservices:MediaService] Successfully encoded 47eb0810-402d-42e1-a9dd-2a51e870c0e3
```
I'm guessing I don't get hardware acceleration in a virtual DSM instance so that should be okay.

Any other things to change or additional review you can provide before this is merged?

cross/immich-geodata/Makefile
```diff
@@ -0,0 +1,42 @@
...
+CHECKSUM_TARGET = nop
+CONFIGURE_TARGET = nop
+COMPILE_TARGET = nop
+INSTALL_TARGET = immich_geodata_install
+
+ADMIN1_URL = http://download.geonames.org/export/dump/admin1CodesASCII.txt
+ADMIN2_URL = http://download.geonames.org/export/dump/admin2Codes.txt
+COUNTRIES_URL = https://github.com/nvkelso/natural-earth-vector/raw/master/geojson/ne_10m_admin_0_countries.geojson
+
+include ../../mk/spksrc.cross-cc.mk
```
th0ma7: Couldn't you simply use cross-install.mk and avoid the `nop` above (besides checksum)?

cross/immich/Makefile
```diff
@@ -0,0 +1,150 @@
...
+
+BUILD_DEPENDS = native/nodejs_22 native/pnpm
+
+HOMEPAGE = https://immich.app
+COMMENT  = Self-hosted photo and video backup solution.
+LICENSE  = GPLv3
+
+INSTALL_TARGET = immich_install
+POST_INSTALL_TARGET = immich_post_install
+ENV += PATH=$(WORK_DIR)/../../../native/nodejs_22/work-native/node/bin:$(WORK_DIR)/../../../native/pnpm/work-native/bin:$$PATH
```
th0ma7, line 17: Note for future framework addition: This is something that could eventually be added to the default environment somehow to avoid declaring it all the time.  Also, we may be able to use debian defaults instead of maintaining ours...

cross/immich/Makefile
```diff
@@ -0,0 +1,150 @@
...
+ENV += SKIP_GIT_INSTALL_HOOKS=1
+
+include ../../mk/spksrc.cross-install.mk
+
+# Map spksrc ARCH to Node.js arch for native module downloads
+ifeq ($(findstring $(ARCH),$(ARMv8_ARCHS)),$(ARCH))
+NODE_ARCH = arm64
+else
+NODE_ARCH = x64
+endif
```
th0ma7, line 31: likewise, this could also be pre-populated by the framework...

th0ma7 (changes requested):
Just one thing related to using cross-install.mk unless I'm mistaking.  Otherwise LGTM.

cross/immich-geodata/Makefile
mreid-tt: much thanks. implemented.

MAngel666:
Hi there,

I was looking at the requirements and feel that the Postgres and Redis dependencies might not be mandatory for everyone. For instance, I host all my databases in a single VM on a Proxmox server.

Would it be possible to mark them as optional in the setup?

Thank you in advance for considering this!

mreid-tt:
Hey @MAngel666, looking at the production Docker Compose configuration (https://github.com/immich-app/immich/blob/main/docker/docker-compose.prod.yml), I don't get the impression that Redis or PostgreSQL are optional components.

Additionally, the PostgreSQL deployment relies on a number of extensions, including `pgvector`, which may not be available in a standard PostgreSQL installation. That dependency is one of the reasons the package bundles a compatible PostgreSQL instance rather than relying on an existing system database.

ya4ept:
@mreid-tt 
I'm currently using a Docker-based setup and have specified local Redis usage in my .env file:
```
# CUSTOM Redis:
REDIS_SOCKET=/run/synocached.sock
```

It would be great if your package also used the standard Redis.

mreid-tt:
> It would be great if your package also used the standard Redis.

I am not sure what is the standard Redis. SynoCommunity provides a Redis package, and it is declared as a dependency of the Immich package. During installation, Package Center will automatically prompt users to install any required dependencies, just as it does for other SynoCommunity packages.

MAngel666:
Hi @mreid-tt , yes, they are indeed necessary for the app to run, but they don't have to run as services on the Synology NAS or even on the same host. As mentioned, I host them on a different machine (including the vectoring lib, etc.).

Sure, I could just install the dependent packages on Syno and leave them disabled, but wouldn't it make more sense this way?

Other than that, really great work! This is honestly one of the few packages on Synology that I was personally missing.

mreid-tt:
> Hi @mreid-tt , yes, they are indeed necessary for the app to run, but they don't have to run as services on the Synology NAS or even on the same host. As mentioned, I host them on a different machine (including the vectoring lib, etc.).
> 
> Sure, I could just install the dependent packages on Syno and leave them disabled, but wouldn't it make more sense this way?

Ah, I see what you mean now. Unfortunately, this comes down to a limitation of Synology's package framework. Dependencies are static and must be declared at build time—they can't be made conditional based on user choices during installation.

For example, it isn't possible to present a checkbox in the installation wizard such as *"I have an external PostgreSQL server"* and then make the PostgreSQL package dependency optional. The Package Center requires all dependencies to be defined up front, so we have to declare them explicitly.

If Synology ever adds support for conditional dependencies, I'd be very happy to make the dependency list more minimal for users with external services. Until then, declaring Redis and PostgreSQL as dependencies is the only reliable way to ensure the package has everything it needs to run, even if that means the dependencies are redundant for users such as yourself.

> Other than that, really great work! This is honestly one of the few packages on Synology that I was personally missing.

Thank you! I'm really glad to hear that, and I hope it proves useful for many Synology users.

MAngel666:
I just installed Immich and the setup went smoothly, but I noticed the dependencies are quite strict. The system requires the packages to not only be installed but actively running. This wastes local resources that many users might prefer to save.

Would you mind reconsidering whether Postgres and Redis must be hard local dependencies? If it's possible to decouple them, adding input fields for external Postgres and Redis addresses in the Immich configuration UI would be an amazing solution.

mreid-tt:
> I just installed Immich and the setup went smoothly, but I noticed the dependencies are quite strict. The system requires the packages to not only be installed but actively running. This wastes local resources that many users might prefer to save.

Thank you for taking the time to test the package. As mentioned above, the hard dependency configuration is largely a consequence of how Synology's package framework handles dependencies. It does unfortunately mean that some resources are consumed unnecessarily in scenarios where users already have external services available.

I've seen similar trade-offs on other NAS platforms, where for example multiple applications each end up running their own PostgreSQL instance. It's not ideal, but it's the compromise that comes with making packages as straightforward as possible to install for the majority of users.

> Would you mind reconsidering whether Postgres and Redis must be hard local dependencies? If it's possible to decouple them, adding input fields for external Postgres and Redis addresses in the Immich configuration UI would be an amazing solution.

The alternative would be to require users to manually install and configure those services themselves when needed, which I think would result in a much poorer experience overall.

If a required dependency is missing or misconfigured, the application simply fails to start, with the underlying reason typically buried in the service logs. Diagnosing those failures is well beyond what we'd expect a typical user to troubleshoot on their own. By declaring the dependencies explicitly, Package Center can ensure everything required is installed before Immich is started, significantly reducing the chances of users ending up with a broken installation.

That being said, if this is a critical requirement for your particular setup, there is a workaround. You can manually edit the package's `INFO` file to remove the dependency before installing it. The downside is that this change isn't persistent, so you'll need to reapply the modification each time you upgrade to a newer package release.

MAngel666:
I think that's a good approach, but it seems modifying the INFO file alone is not sufficient. I encounter the error message shown in the attached image during the installation process. Is there another file or configuration I need to modify?

Thanks for your help with this!
Screenshot_2026-07-16_17-56-48

mreid-tt:
> I think that's a good approach, but it seems modifying the INFO file alone is not sufficient. I encounter the error message shown in the attached image during the installation process. Is there another file or configuration I need to modify?

That error would be as a result of the `spk/immich/src/service-setup.sh` script. At the top there are some variables for the connection details for PostgreSQL which you would need to modify since they are not taken as inputs from the wizard.

If you are manually editing the package that would be in `scripts/service-setup`.

ya4ept:
I could be wrong, but when you install DSM, the Redis database is enabled by default, similar to Postgres. While using the standard Postgres is not recommended, Redis can be used without any problems. It doesn't contain any critical data.

mreid-tt:
> I could be wrong, but when you install DSM, the Redis database is enabled by default, similar to Postgres. While using the standard Postgres is not recommended, Redis can be used without any problems. It doesn't contain any critical data.

As mentioned above, I have no evidence that DSM ships with a Redis instance. This is why we include a Redis package in our catalog.

MAngel666:
I was able to customize everything so far and I can see that Immich is trying to communicate with my DB. However, no matter what I do, it fails to connect, giving an error that the password for the user immich is incorrect.

I’ve already tried the following:

- Setting a new password
- Switching between scram-sha-256 and md5
- Removing any problematic special characters from the password
- Confirming that the IP in pg_hba.conf on the DB server is set correctly for the immich user

Connecting via the psql CLI tool directly from the Synology is working.

What else could be wrong here? (The server is running Postgres version 18.4.)"

mreid-tt:
> Connecting via the psql CLI tool directly from the Synology isn't working either; I get the exact same error. However, running Immich in Docker works without any issues.

Troubleshooting PostgreSQL itself is outside the scope of this issue. However, I can confirm that connecting from a Synology NAS to a remote PostgreSQL 18 server is supported. For example, from a DSM 7.2 system I was able to successfully connect to a Docker-hosted PostgreSQL 18.4 instance using the bundled `psql` 11.11 client:

```console
mreid@DSM72:~$ psql --version
psql (PostgreSQL) 11.11
mreid@DSM72:~$ psql "host=192.168.10.160 port=5432 dbname=spkrepo user=spkrepo password=spkrepo"
psql (11.11, server 18.4 (Debian 18.4-1.pgdg13+1))
WARNING: psql major version 11, server major version 18.
         Some psql features might not work.
Type "help" for help.

spkrepo=# \dt
[...]
(20 rows)
```

Since you're unable to connect using `psql` directly from the Synology, this suggests the issue isn't specific to the Immich package. I would recommend resolving the PostgreSQL connectivity first, as Immich will not be able to connect until a standard PostgreSQL client can do so successfully.

The next step would be to focus on why `psql` authentication is failing—for example by reviewing the PostgreSQL server logs, verifying that the connection is reaching the expected server, checking which `pg_hba.conf` rule is being matched, and confirming the credentials and SSL configuration. If needed, there are many PostgreSQL troubleshooting guides available online, and an AI assistant can also help interpret the server logs and identify the root cause.

MAngel666:
Sorry, I've updated my comment - psql is working normally, I must have had a typo. For Immich, it's still not working after checking multiple times, even without any changes.

```
/usr/local/bin/psql -h pgsql.db.mydom -p 5432 -U immich -d immich
Password for user immich: 
psql (17.10, server 18.4)
WARNING: psql major version 17, server major version 18.
         Some psql features might not work.
Type "help" for help.

immich=> 

```

mreid-tt:
@MAngel666, the database connection is constructed in the `spk/immich/src/service-setup.sh` script as follows:

https://github.com/SynoCommunity/spksrc/blob/94535bb79f42c8b82683f55a9d7076d8128a5665/spk/immich/src/service-setup.sh#L30-L40

The key parameters here are:

* `DB_HOSTNAME`
* `DB_PORT`
* `DB_USERNAME`
* `DB_PASSWORD`

If you are connecting to an external PostgreSQL server, these values need to be adjusted accordingly.

Reviewing this conversation, I realised that these values should have been using the global variables defined at the start of the script rather than being hardcoded. I'll correct this in the next version of the package.

This section is likely the reason you're seeing the connection failure. For example, the current defaults assume the SynoCommunity PostgreSQL package, which uses port `5433`, whereas an external PostgreSQL server would typically be using port `5432`.

MAngel666:
done it in my pkg:
```
service_prestart()
{
    DB_PASSWORD="$(cat /var/packages/immich/var/db_password 2>/dev/null)"
    export NODE_OPTIONS=--unhandled-rejections=warn
    export DB_PASSWORD
    export IMMICH_BUILD_DATA=/var/packages/immich/target/share/immich
    export IMMICH_MEDIA_LOCATION="${MEDIA_PATH}"
    export DB_HOSTNAME=pgsql.db.mydom
    export DB_PORT=5432
    export DB_USERNAME=immich
    export REDIS_HOSTNAME=pgsql.db.mydom
    export PATH="/var/packages/ffmpeg8/target/bin:${PATH}"

    SERVICE_COMMAND="${NODE} ${SERVER_LAUNCHER}"

    if [ -x "${ML_VENV}/bin/python3" ]; then
        export IMMICH_MACHINE_LEARNING_URL=http://127.0.0.1:3003
        SERVICE_COMMAND="${SERVICE_COMMAND}
${MLLAUNCHER}"
    else
        export IMMICH_MACHINE_LEARNING_ENABLED=false
    fi
}

```

mreid-tt:
> done it in my pkg

Is it working now? if not you may want to confirm that the database password value in `/var/packages/immich/var/db_password` is correct.

MAngel666:
the file doesnt exist :/
I've created /var/packages/immich/var/db_password with read premissions for all and now it starts

mreid-tt:
@MAngel666, okay, so that confirms the root cause.

That file would have been generated as part of the installation wizard. For the next package update, I think I'll move towards using an environment variable file at:

```
/var/packages/immich/var/immich.conf
```

This file will contain the exported configuration values above and will simply be sourced before launching the application. This should make use cases like yours much easier, as you would be able to directly edit the configuration file rather than modifying the package scripts.

This change will only allow you to override the PostgreSQL and Redis connection details. It will not remove or alter the package dependencies defined in the base package. If you choose to use an external PostgreSQL or Redis instance, you will still need to manually adjust those dependency settings if needed.

MAngel666:
now I have a next problem... The share for immich I've created before install and configured it at immich install as share. It have all rights for user sc-immich but I get this:
```
Nest] 31587  - 07/20/2026, 3:15:59 PM     LOG [Api:StorageService] Verifying system mount folder checks, current state: {"mountChecks":{"thumbs":true,"upload":true,"backups":true,"library":true,"profile":true,"encoded-video":true}}
[Nest] 31587  - 07/20/2026, 3:15:59 PM   ERROR [Api:StorageService] Failed to read (/var/packages/immich/shares/immich/encoded-video/.immich): Error: ENOENT: no such file or directory, open '/var/packages/immich/shares/immich/encoded-video/.immich'
(node:31587) UnhandledPromiseRejectionWarning: Error: Failed to read: "<UPLOAD_LOCATION>/encoded-video/.immich (/var/packages/immich/shares/immich/encoded-video/.immich) - Please see https://docs.immich.app/administration/system-integrity#folder-checks for more information."
    at StorageService.verifyReadAccess (/volume11/@appstore/immich/share/immich/dist/services/storage.service.js:130:19)
    at async /volume11/@appstore/immich/share/immich/dist/services/storage.service.js:57:21
    at async /volume11/@appstore/immich/share/immich/dist/repositories/database.repository.js:376:27
    at async /volume11/@appstore/immich/share/immich/node_modules/.pnpm/kysely@0.28.17/node_modules/kysely/dist/cjs/kysely.js:541:20
    at async DefaultConnectionProvider.provideConnection (/volume11/@appstore/immich/share/immich/node_modules/.pnpm/kysely@0.28.17/node_modules/kysely/dist/cjs/driver/default-connection-provider.js:12:20)
    at async /volume11/@appstore/immich/share/immich/dist/repositories/database.repository.js:373:13
(node:31587) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)

```

mreid-tt:
> now I have a next problem... The share for immich I've created before install and configured it at immich install as share. It have all rights for user sc-immich but I get this:

Immich requires a number of directories to exist under the configured media path, each containing an empty `.immich` marker file. This is handled automatically by the package during installation with the following code:

https://github.com/SynoCommunity/spksrc/blob/94535bb79f42c8b82683f55a9d7076d8128a5665/spk/immich/src/service-setup.sh#L118-L121

So if the shared folder was correctly selected during the installation wizard, these folders and files should already exist.

Can you confirm the exact path configured during installation and compare it with the shared folder you are checking now? If the share was renamed after installation, or a different share/path was selected, the package may have created the structure elsewhere.

MAngel666:
done it manually and now immich works correctly...
I've created an own share "immich" and configured it as share at immich installation. But the directories was not created...

mreid-tt:
> I've created an own share "immich" and configured it as share at immich installation. But the directories was not created...

Hmm, that should not be the case. If you can review the installation log at `/var/log/packages/immich.log` for the `service_postinst` step it may give some insight into what happened or did not happen.

mreid-tt:
@MAngel666, I discovered that the current build wasn't correctly preserving the machine learning virtual environment. I've published a new release in #7316 that fixes this issue and also includes the migration to the new configuration file format I mentioned earlier.

You can manually download the updated package from https://synocommunity.com/package/immich. It should also become available through Package Center within approximately 24 hours.

PR #7285, opened 2026-07-13
radarr: allow built-in auto-updates on DSM 7.2+

ogerardin:
## Motivation

The previous implementation baked `UpdateMethod=External` into `package_info` at build time when `TCVERSION < 7.2`. This meant packages built with TCVERSION=7.1 (the current default) always block Radarr's self-updater, even on DSM 7.2+ systems where the upstream `libe_sqlite3.so` (GLIBC ≥ 2.28) works fine.

This was introduced in commit [a755a5d](https://github.com/SynoCommunity/spksrc/commit/a755a5d) to protect the custom `libe_sqlite3.so` (built for GLIBC 2.20–2.26) on DSM 6–7.1 from being overwritten by Radarr's updater. The protection is necessary but was applied too broadly — it also blocks self-updates on DSM 7.2+ where the upstream library works.

### What changes

- **`spk/radarr/Makefile`**: Remove `RADARR_UPDATE = disabled` and the conditional `UpdateMethod=External` injection from `radarr_extra_install`. The `libe_sqlite3` custom build dependency is kept (still required for DSM < 7.2).
- **`spk/radarr/src/service-setup.sh`**: 
  - `internal_update_supported()` now checks DSM version instead of reading `package_info`
  - New `configure_update_method()` writes `UpdateMethod=External` to `package_info` only on DSM < 7.2
  - Called from `service_postinst` and `service_postupgrade`

### Behavior by DSM version

| Scenario | UpdateMethod in package_info | Radarr self-update |
|---|---|---|
| DSM 7.2+ fresh install | Removed → defaults to BuiltIn | ✅ Enabled |
| DSM 7.2+ upgrade | Removed → defaults to BuiltIn | ✅ Enabled |
| DSM 7.1 fresh install | Added External | ❌ Disabled |
| DSM 7.1 upgrade | Added External | ❌ Disabled |

## Linked issues

None (original issue: Radarr shows "Unable to update Radarr directly" on DSM 7.2.1)

## Checklist

## Related PRs

Same fix applied to the other *arr packages:
- **Prowlarr**: [#7287](https://github.com/SynoCommunity/spksrc/pull/7287)
- **Lidarr**: [#7286](https://github.com/SynoCommunity/spksrc/pull/7286)

### DIFF
--- a/spk/radarr/Makefile
+++ b/spk/radarr/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = radarr
 SPK_VERS = $(shell date +%Y%m%d)
-SPK_REV = 26
+SPK_REV = 27
 SPK_ICON = src/radarr.png
 
 OPTIONAL_DEPENDS = cross/libstdc++ cross/libe_sqlite3
@@ -12,7 +12,7 @@ DOTNET_CORE_ARCHS = 1
 MAINTAINER = Team Radarr
 MAINTAINER_URL = https://radarr.video/
 DESCRIPTION = Radarr is a movie collection manager for Usenet and BitTorrent users. It can monitor multiple RSS feeds for new movies and will interface with clients and indexers to grab, sort, and rename them. It can also be configured to automatically upgrade the quality of existing files in the library when a better quality format becomes available.
-CHANGELOG = "1. Update Radarr to v6.2.1.10461."
+CHANGELOG = "1. Dynamically set UpdateMethod based on DSM version, enabling self-update on DSM 7.2+."
 DISPLAY_NAME = Radarr
 HOMEPAGE = https://radarr.video/
 LICENSE = GPLv3
@@ -45,8 +45,6 @@ ifeq ($(call version_lt, ${TCVERSION}, 7.2),1)
 # Use libe_sqlite3 built for GLIBC 2.20–2.26 (DSM 6–7.1) instead of upstream ≥ 2.28
 DEPENDS += cross/libe_sqlite3
 POST_STRIP_TARGET += radarr_custom_libe_sqlite3
-# Disable auto-update for DSM < 7.2
-RADARR_UPDATE = disabled
 endif
 
 POST_STRIP_TARGET += radarr_extra_install
@@ -67,7 +65,3 @@ radarr_extra_install:
 	@install -m 755 -d $(STAGING_DIR)/var/.config/Radarr
 	@install -m 644 src/config.xml $(STAGING_DIR)/var/.config/Radarr/config.xml
 	@echo "PackageVersion=$(PACKAGE_VERSION)\nPackageAuthor=$(PACKAGE_AUTHOR)" > $(STAGING_DIR)/share/Radarr/package_info
-	@if [ "$(RADARR_UPDATE)" = "disabled" ]; then \
-		$(MSG) "Disable auto-update in package_info." ; \
-		echo "UpdateMethod=External" >> $(STAGING_DIR)/share/Radarr/package_info ; \
-	fi
--- a/spk/radarr/src/service-setup.sh
+++ b/spk/radarr/src/service-setup.sh
@@ -23,22 +23,41 @@ SVC_BACKGROUND=y
 SVC_WAIT_TIMEOUT=90
 
 internal_update_supported() {
-    info_file="${SYNOPKG_PKGINST_TEMP_DIR}/share/Radarr/package_info"
+    # DSM >= 7.2: Radarr can self-update (upstream libe_sqlite3.so is GLIBC 2.28+ compatible)
+    # DSM < 7.2: Radarr cannot self-update (custom libe_sqlite3.so must be preserved)
+    if [ "${SYNOPKG_DSM_VERSION_MAJOR}" -gt 7 ] || \
+       { [ "${SYNOPKG_DSM_VERSION_MAJOR}" -eq 7 ] && [ "${SYNOPKG_DSM_VERSION_MINOR}" -ge 2 ]; }; then
+        return 0    # Supported
+    else
+        return 1    # Not supported
+    fi
+}
 
-    # If the file doesn't exist, assume update is supported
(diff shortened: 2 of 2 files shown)


ogerardin:
CI failure is a pre-existing upstream issue: `Download/checksum failed for cross/zsh` — unrelated to this PR. Upstream PR #7284 is addressing the zsh download issue. Local Docker build completed successfully.

mreid-tt:
@ogerardin, I believe the premise of this fix is based on an incorrect assumption. The `Makefile` logic already accounts for this scenario, which is why the DSM 7.2 build correctly includes auto-updates.

We certainly appreciate new contributors and the time taken to improve the project. However, in cases where the behavior is uncertain, it's generally better to open an issue first and discuss it with the maintainers before concluding that there is a bug requiring a fix. That gives everyone an opportunity to confirm whether the observed behavior is intentional or an actual defect before work is undertaken.

ogerardin:
> @ogerardin, I believe the premise of this fix is based on an incorrect assumption. The `Makefile` logic already accounts for this scenario, which is why the DSM 7.2 build correctly includes auto-updates.
> 
> We certainly appreciate new contributors and the time taken to improve the project. However, in cases where the behavior is uncertain, it's generally better to open an issue first and discuss it with the maintainers before concluding that there is a bug requiring a fix. That gives everyone an opportunity to confirm whether the observed behavior is intentional or an actual defect before work is undertaken.

Hi and thanks for taking time to look at this PR. 

You're right that the Makefile conditionals are correct for compile-time builds, but the issue we hit is that the CI doesn't produce a TCVERSION=7.2 build for Radarr by default (since Radarr doesn't declare REQUIRED_MIN_DSM = 7.2). And since the 7.1 SPK has no os_max_ver, DSM 7.2 systems happily install it, which sets UpdateMethod=External and disables auto-updates. That's what I observed on my DS918+ running DSM 7.2.1.

So the question is: what's the preferred way to fix this? My PRs use runtime detection: at service start, check the actual DSM version and only set UpdateMethod=External if below 7.2. Works with a single SPK for both 7.1 and 7.2.

Another way is to add REQUIRED_MIN_DSM = 7.2: This would make the CI build a separate 7.2 SPK with auto-updates enabled. But it changes the distribution scope, and users who already installed the 7.1 SPK on 7.2 would need to reinstall.

Happy to hear your thoughts.

mreid-tt:
> You're right that the Makefile conditionals are correct for compile-time builds, but the issue we hit is that the CI doesn't produce a TCVERSION=7.2 build for Radarr by default (since Radarr doesn't declare REQUIRED_MIN_DSM = 7.2). And since the 7.1 SPK has no os_max_ver, DSM 7.2 systems happily install it, which sets UpdateMethod=External and disables auto-updates. That's what I observed on my DS918+ running DSM 7.2.1.

For distribution, we already publish separate builds for DSM 7.1 and DSM 7.2, as you can see on our website: https://synocommunity.com/package/radarr. When installing through Package Center, DSM will automatically select and install the appropriate build for the target system.

> So the question is: what's the preferred way to fix this? My PRs use runtime detection: at service start, check the actual DSM version and only set UpdateMethod=External if below 7.2. Works with a single SPK for both 7.1 and 7.2.

While your approach is certainly clever, I don't think it's necessary at this time. In the unlikely event that the installation constraint is not removed during an upgrade from DSM 7.1 to DSM 7.2, the simplest remedy is to manually download the DSM 7.2 package from the website and install it over the existing installation. That will replace the older build and resolve the issue without requiring additional complexity in the package logic.

ogerardin:
> For distribution, we already publish separate builds for DSM 7.1 and DSM 7.2, as you can see on our website: https://synocommunity.com/package/radarr. When installing through Package Center, DSM will automatically select and install the appropriate build for the target system.

If that's the case, I don't understand how I ended up in this sitation. Is it possible that I installed the package while I was still in 7.1, and after I had updated to 7.2 the update did not switch to the 7.2 build? As far as I can tell, automatic updates were always available in Radarr, and I only noticed today that they were not, which prompted me to investigate.

This is unfortunate since I believe I probably won't be the only one in this situation.

mreid-tt:
> If that's the case, I don't understand how I ended up in this sitation. Is it possible that I installed the package while I was still in 7.1, and after I had updated to 7.2 the update did not switch to the 7.2 build? As far as I can tell, automatic updates were always available in Radarr, and I only noticed today that they were not, which prompted me to investigate.

This is one of the main reasons I suggested opening an issue first. A bug report would typically include the installation log, allowing us to review your upgrade history and better understand what actually occurred before proposing a fix.

Either way, as mentioned above, simply reinstalling the appropriate package will resolve the issue if you've encountered this situation.

ogerardin:
I'll go ahead and close the PRs

I still believe my approach was better for handling this edge case, but not my call.
Happy to contribute it separately if there's interest.

Thanks again for taking the time to review.

ogerardin:
One last note: the Radarr 7.2 build requires DSM build 7280. On my DS918+, DSM auto-update stopped at DSM 7.2.1 build 69057. That probably explains why I didn't get the Radarr 7.2 package (and can't install it manually)

mreid-tt:
> One last note: the Radarr 7.2 build requires DSM build 7280. On my DS918+, DSM auto-update stopped at DSM 7.2.1 build 69057. That probably explains why I didn't get the Radarr 7.2 package (and can't install it manually)

Ah, that explains a lot.

Yes, our DSM 7.2 packages require DSM build **72806** or newer. If I remember correctly, the https://www.synology.com/en-global/releaseNote/DSM release notes mention a DSM 7.2 version where automatic updates did not work correctly. Synology's recommended workaround was to manually install the next DSM update, after which subsequent updates would once again install automatically.

When you have a chance, I'd recommend updating your NAS to the latest DSM 7.2.x release to stay current with the latest security updates and bug fixes. I believe the latest release in the DSM 7.2.x series is **DSM 7.2.2-72806 Update 9**.

PR #7287, opened 2026-07-13
lidarr: allow built-in auto-updates on DSM 7.2+

ogerardin:
## Description
Same approach as Radarr PR #7285 — allow Prowlarr's built-in auto-updater to work on DSM 7.2+ while keeping `UpdateMethod=External` for older DSM versions where the updater would wipe the custom `libe_sqlite3.so`.

## Changes
- **Makefile**: Remove `PROWLARR_UPDATE = disabled` and dynamic `UpdateMethod=External` injection. Bump SPK_REV to 14.
- **service-setup.sh**: Add `configure_update_method()` that checks DSM version at runtime and sets `UpdateMethod` in `package_info` accordingly. Wire into `service_postinst` and `service_postupgrade`.

## Testing
- [x] Built locally via Docker (`make arch-x64-7.1`) — SPK generated successfully
- [x] Installed on DSM 7.2 — self-update works
- [ ] Installed on DSM < 7.2 — remains External, no auto-update

## Related
- Radarr PR: #7285
- Lidarr PR: #7286

### DIFF
--- a/spk/lidarr/Makefile
+++ b/spk/lidarr/Makefile
@@ -1,6 +1,6 @@
 SPK_NAME = lidarr
 SPK_VERS = $(shell date +%Y%m%d)
-SPK_REV = 17
+SPK_REV = 18
 SPK_ICON = src/lidarr.png
 
 OPTIONAL_DEPENDS = cross/libstdc++ cross/libe_sqlite3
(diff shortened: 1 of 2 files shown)


ogerardin:
## Note: Other *arr packages not affected

Sonarr, Readarr, Sonarr3, and Bazarr do **not** have the same issue because they don't set `UpdateMethod=External` in `package_info`:

| Package | `UpdateMethod=External`? | Protection mechanism |
|---|---|---|
| Radarr, Prowlarr, Lidarr | Yes (injected at build time) | Fixed by this PR |
| Sonarr v4 | No | Backup/restore distribution in `service-setup.sh` |
| Readarr | No | Same backup/restore approach |
| Sonarr3 | No | Same backup/restore approach |
| Bazarr | No | Python-based, uses `--no-update` flag |

Sonarr/Readarr don't block self-updates — they protect the distribution during package upgrades by backing it up in `service_preupgrade` and restoring it in `service_postupgrade` (excluding `package_info`). This is a fundamentally different approach that doesn't require the same fix.

mreid-tt:
@ogerardin, I believe the premise of this fix is based on an incorrect assumption. The `Makefile` logic already accounts for this scenario, which is why the DSM 7.2 build correctly includes auto-updates.

We certainly appreciate new contributors and the time taken to improve the project. However, in cases where the behavior is uncertain, it's generally better to open an issue first and discuss it with the maintainers before concluding that there is a bug requiring a fix. That gives everyone an opportunity to confirm whether the observed behavior is intentional or an actual defect before work is undertaken.

ogerardin:
Close until discussion is resolved on https://github.com/SynoCommunity/spksrc/pull/7285

PR #7290, opened 2026-07-13
Dead code cleanup: unreachable packages and DSM 5.2 toolchains

th0ma7:
Fixes #7197.

Every candidate was re-validated against the current tree before removal (the issue asks for exactly that: *"Each candidate should be re-validated right before removal, the tree moves"*). The check is a BFS over the dependency graph from every `spk/`, `diyspk/`, `python/` and `kernel/` entry point plus the packages named in `mk/`, following non-commented `DEPENDS` / `BUILD_DEPENDS` / `OPTIONAL_DEPENDS` / `NATIVE_DEPENDS` and `$(*_PACKAGE)` indirections.

## 1. Dead `cross/` + `native/` + `diyspk/` packages

The list from the issue, each re-confirmed unreachable with no reference anywhere outside its own directory:

| Package | Why it is dead |
|---|---|
| `cross/xmlrpc-c` | rtorrent builds `--with-xmlrpc-tinyxml2` since #6810 |
| `cross/glibmm` + `-2.66` + `-latest` | last consumer was `cross/libxml++`, removed in #6522 — [@hgy59](https://github.com/SynoCommunity/spksrc/issues/7197#issuecomment-4678388823) |
| `cross/glibc` + `-2.28` + `-latest` | entire family, meta included |
| `cross/libidn` | superseded by `cross/libidn2` |
| `cross/isl` | former gcc / synocli-devel dependency |
| `cross/cppunit` | unreferenced |
| `cross/alsa-oss` | orphaned by the 2023 alsa rework (#5656) |
| `cross/zlib-ng` | last consumer was `cross/bat`, switched back to zlib in #6052 — [@hgy59](https://github.com/SynoCommunity/spksrc/issues/7197#issuecomment-4678267933) |
| `native/yasm` | zero references, even in comments |
| `diyspk/fritzctl` | depends on `cross/fritzctl`, which no longer exists — it cannot build at all |

Three deliberate scope limits:

- **`native/nasm` is not touched.** [@hgy59 pointed out](https://github.com/SynoCommunity/spksrc/issues/7197#issuecomment-4678102928) that yasm removal is part of #6263, which was never completed. That PR removes *both* nasm and yasm; this one takes only the yasm half, since nasm is still what the assembly-using packages build with. #6263 remains the place for nasm.
[...]

PR #7296, opened 2026-07-14
spk/subliminal: rebuild as a py3 command line package, update to 2.7.1

th0ma7:
Revives `spk/subliminal`, which is currently **BROKEN**, and updates it from 0.7.5 (2015) to **2.7.1**.

## Why it was broken

```
subliminal requires installed python2 in the development environment
not supported anymore since spksrc image was updated to debian 12
```

It could not be fixed in place. The package shipped subliminal 0.7.5 plus a DSM web interface — a CGI, a scheduler, a scanner and an Ext.Direct RPC bridge — all written in **Python 2**. Meanwhile upstream rewrote the project: **subliminal 2.7.1 requires Python >= 3.9 and is a command line tool with no web interface.**

## What this does

Rebuilds the package on the python meta pattern already used by `beets` and `mercurial`:

- `PYTHON_PACKAGE = python312`
- `WHEELS = src/requirements-crossenv.txt src/requirements-pure.txt`
- `SPK_COMMANDS = env/bin/subliminal`, `STARTABLE = no`
- `BROKEN` removed

Of the 30 resolved dependencies, only **`PyYAML`** and **`charset_normalizer`** build a C extension and need cross-compiling. The other 28 are pure python.

`knowit` (a hard dependency of subliminal) pulls in **`pymediainfo`**, which is a ctypes wrapper around libmediainfo — hence `DEPENDS = cross/libmediainfo`. This is also why #7290 keeps the `cross/libmediainfo` meta rather than removing it as dead code.

## The DSM interface is dropped, not ported

This is worth being explicit about. It is **not** a porting job:

- The UI is built on **`pyextdirect`**, whose **last release is from 2012** and whose classifiers declare **Python 2.6 / 2.7 only**. There is no Python 3 version, so the entire Ext.Direct RPC layer would have to be rewritten.
- It drives the **subliminal 0.7 API**, which no longer exists in 2.x.

Restoring a user interface is left as follow-up work. The old sources remain in git history and can be recovered from there. Shipping the CLI now is what unblocks a package that currently builds for nobody.


th0ma7:
@Diaoul I looked at resurecting this package which I believe you maitain the code.  Feedback would be welcomed.  Also is there an updated UI for the new version as the older one is in python2?

Diaoul:
Sadly I don't maintain anything anymore by lack of time, thanks for picking this up. Fully trust you on that!

PR #7298, opened 2026-07-15
help + docs: document DEPENDS_TYPE and EXCLUDE_DEPENDS

th0ma7:
You spotted that `EXCLUDE_DEPENDS` and `DEPENDS_TYPE` were only described in the header of `mk/spksrc.rules/dependency-tree.mk` — neither `make help` nor the docs mentioned them, even though they are the useful part of the `dependency-*` targets.

Added in #7028 (`EXCLUDE_DEPENDS`, `DEPENDS_TYPE`) and #7121 / #7124 (`ARCH`/`TCVERSION` context), they had no user-facing documentation. This surfaces them everywhere they apply.

## Package-level `make help`

The Inspect section now spells out the three modifiers:

```
Inspect
  dependency-tree          print the resolved dependency graph
  dependency-flat          flat, de-duplicated dependency list
  dependency-list          raw per-package dependency list
    DEPENDS_TYPE="DEPENDS BUILD_DEPENDS OPTIONAL_DEPENDS NATIVE_DEPENDS" filters flat/list output
    EXCLUDE_DEPENDS="cross/foo ..." skips those dependencies and their subtrees
    ARCH=... TCVERSION=... resolves the graph in that toolchain context
```

## Root `make help`

You were right that there's also a root-level call — `dependency-list-spk`, which resolves every `spk/`'s dependencies in parallel. It honors the same variables (they propagate to the per-package sub-makes through `MAKEFLAGS`), verified:

```
$ make dependency-list-spk DEPENDS_TYPE="NATIVE_DEPENDS"
adminer:
aria2:
...                       # empty where there are no native deps, vs the full list without the filter
```

It had no help entry at all, so it now appears under a new **Inspect** section in the root help:

```
Inspect
  dependency-list-spk  Resolved deps for every spk; honors DEPENDS_TYPE, EXCLUDE_DEPENDS, ARCH, TCVERSION
```

## Docs

`build-rules.md` gains an **Inspecting the dependency graph** subsection: a table of the three variables, two worked examples, and the root-level `dependency-list-spk` variant.

## Testing

`make help` at both levels, `make dependency-list-spk DEPENDS_TYPE=NATIVE_DEPENDS`, and `mkdocs build --strict` — all clean.

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QEH1b4ASNYSrZFQnEeroj8

### DIFF
--- a/docs/developer-guide/packaging/build-rules.md
+++ b/docs/developer-guide/packaging/build-rules.md
@@ -45,6 +45,35 @@ wheel/crossenv targets are available — see
 `crossenv-<arch>-<tcvers>`, `download-wheels`, `wheelclean`, `wheelcleancache`,
 `crossenvclean`, `crossenvcleanall`.
 
+### Inspecting the dependency graph
+
+`dependency-tree`, `dependency-flat` and `dependency-list` share three optional
+variables (they do not apply to `dependency-tree`, which always prints the full
+traversal):
+
+| Variable | Effect |
+|----------|--------|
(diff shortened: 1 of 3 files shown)


th0ma7:
Folded into #7297 — it is all documentation, so it belongs with the Framework Changes page rather than as a separate PR.

PR #7303, opened 2026-07-16
DNSACME: Initial package v1.0.8

mritd:
## Description

Adds DNSACME v1.0.8 as a new SynoCommunity package for DSM 7.

DNSACME obtains ACME certificates through DNS-01 challenges and deploys them through the Synology DSM WebAPI. The package builds the upstream source with the `synology` build tag, installs the DSM AppWindow assets, and runs the renewal daemon under a dedicated package account.

The package supports 64-bit x64 and aarch64 systems, as well as 32-bit ARMv7 and i686 systems running DSM 7. PPC architectures and SRM are not currently supported. Users can retain or remove package data through the standard DSM 7 uninstall wizard.

## Checklist

- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created (No additional documentation is required)

### Type of change

- [x] New Package


hgy59:
@mritd thank you for this contribution. I think it is very useful, I like the DSM ui integration. 
Before further testing I have some questions:
- what is the reason to disable support for 32 bit archs? (I see that you provide 32 bit releases for dnsacme like linux-368, -armv5, -armv6, -armv7)
- I personally would prefer the var folder instead of etc folder for the config file (but this isn't mandatory and DSM 6 isn't supported anyway)
- instead of the start-stop-status.sh script you should create a service-setup.sh to use the generic service installer of our framework (this will simplify it, as shown below)
- I am missing the documentation of the `daemon` command in the help - and there is a small typo in: `Simple tool to manage ACME Cert(Ony Supported DNS-01)`
- a real daemon mode would start the service in the background and not require `nohup`, `&` or `SVC_BACKGROUND=Y`
- another concern is the application name (identifier). we prefer `com.synocommunity.packages.dnsacme` over `SYNO.SDS.DNSACME.Application`, but this might not be a problem (as long as Synology does not create a package named DNSACME).

proposed service-setup.sh file (no shebang required, this file is included into a shell script):
```
ETC=$(realpath /var/packages/${SYNOPKG_PKGNAME}/etc)
CONFIG="${ETC}/config.yaml"

SERVICE_COMMAND="${SYNOPKG_PKGDIR}/bin/dnsacme synology daemon --config ${CONFIG}"
SVC_BACKGROUND=Y
SVC_WRITE_PID=Y

service_postinst ()
{
    mkdir -p ${ETC}
}
```

mritd:
@hgy59, thank you for the detailed review and the helpful suggestions. I went through each point and made the following changes:

- 32-bit architectures: I disabled them because I do not currently have compatible DSM hardware for complete testing. The upstream 32-bit binaries compile, but that does not validate the DSM UI, package lifecycle, WebAPI login, or certificate deployment. I would prefer not to advertise those architectures until the complete package can be tested on real devices.

- Configuration location: I kept config.yaml under etc because it contains DSM and DNS provider credentials. This keeps configuration separate from certificates and logs under var, matches the standalone SPK layout, and is preserved across upgrades by the DSM 7 package framework.

- Service integration: I replaced the custom start-stop-status.sh with the generic service-setup.sh framework. I used SYNOPKG_PKGDEST for the installed package path because that is the runtime variable provided by the framework. I also added `SVC_KEEP_LOG` and umask `077` to preserve log history and keep runtime files private.

- Daemon help and CLI typo: I fixed Ony to Only and documented both `dnsacme synology --help` and `dnsacme synology daemon --help`. The daemon command is now visible in the built Synology binary's help output.

- Daemon behavior: DNSACME intentionally runs as a foreground process. The SynoCommunity service framework is responsible for backgrounding it and tracking its PID through `SVC_BACKGROUND` and `SVC_WRITE_PID`, rather than having the Go process detach itself.

- Application identifier: I changed the identifier to `com.synocommunity.packages.dnsacme` in the SPK metadata, UI configuration, and AppInstance definition. I also added an upstream consistency test to prevent these values from drifting.

In summary, the package now uses DNSACME v1.0.4 and the standard SynoCommunity service framework. It has been tested on a real x64 DSM system, and both make arch-x64-7.2 and make all-supported complete successfully for the declared supported targets.

mritd:
As an aside, in theory both 32-bit and ARM builds should work, thanks to Go's strong cross-platform support. My concern is that I cannot validate them on real hardware, and I am not sure whether SynoCommunity accepts packages for architectures that have not been tested by the maintainer.

spk/dnsacme/Makefile
```diff
@@ -0,0 +1,27 @@
+SPK_NAME = dnsacme
+SPK_VERS = 1.0.7
+SPK_REV = 3
```
mreid-tt: The `SPK_REV` should only change across package releases. As this is still the initial release it should remain at 1.

mritd: Updated

mreid-tt:
**Security review — new package addition**

As part of an evolving intake process for new catalog submissions (still being refined, so scope and depth may shift over time), this repository was flagged for review due to its limited adoption and review by the broader open-source community (2 forks, 9 stars at time of review). Packages that haven't yet accumulated significant community usage or scrutiny are being held to this additional check while the process matures.

**Context:**
- Repository created January 17, 2023, with 2 forks and 9 stars at time of review
- Maintainer (`mritd`) has an established GitHub history (~794 followers) with multiple long-standing published projects, including an existing Docker Hub image for this tool

**Scope of review:**
- Static analysis via `gosec` and `semgrep` against the full source tree
- Manual review of any findings flagged as command-execution or transport-security relevant

**Findings summary:**
- `gosec` returned 9 findings, 7 of which were false positives — environment variable name constants (e.g. `ENV_CLOUDFLARE_API_TOKEN`) used to configure DNS provider credentials, flagged by keyword-matching rather than any actual embedded secret.
- One finding, `exec.Command(hook)` in `events.go`, was reviewed directly. The `hook` value is sourced from local operator-controlled configuration (`ObtainingHook`/`ObtainedHook`/`FailedHook`), not from network or ACME-server input, and is executed without a shell. This is a standard post-issuance/renewal hook pattern consistent with other ACME clients (e.g. certbot's `--deploy-hook`).
- One TLS finding (`InsecureSkipVerify: true` in `synology_deploy.go`) was reviewed directly. The exception is narrowly scoped to loopback (`127.0.0.1`) connections only, used to accommodate a known Synology DSM certificate quirk on the local API; all non-local connections use the standard verified TLS transport.
- No hardcoded secrets, obfuscated code, unexplained network endpoints, or other indicators of malicious functionality were identified.

**Outcome:** No concerns identified prior to approval. This review does not itself constitute formal approval; the package proceeds to the standard approval workflow with no flags raised from this check.

spk/dnsacme/Makefile
```diff
@@ -0,0 +1,27 @@
...
+DEPENDS = cross/dnsacme
+
+REQUIRED_MIN_DSM = 7.0
+REQUIRED_MIN_SRM = 3.0
+UNSUPPORTED_ARCHS = $(32bit_ARCHS)
+
+MAINTAINER = mritd
+DESCRIPTION = DNS-01 ACME certificate manager with Synology DSM certificate deployment.
+DISPLAY_NAME = DNSACME
+CHANGELOG = "Use DNSACME as the default certificate description."
```
hgy59: it is still the initial package release, changelog will never be shown in DSM package center.

mritd: Fixed

mritd:
@hgy59 Are there any other changes needed in this PR? If not, I just need to wait for it to be merged, right?

hgy59:
> @hgy59 Are there any other changes needed in this PR? If not, I just need to wait for it to be merged, right?

Can you enable all archs? (remove the 32 bit archs from unsupported)?

It seems to run on SRM 1.3 but needed to run as root to write the conf.yml file.
Alas I can't do a full test, because I don't have a supported DNS provider.


I propose the following change:
in cross/dnsacme/Makefile it would be better to use the original install target to install the dnsacme binary and to use a custom POST_INSTALL_TARGET to additionally install only the ui files.


I can test the armv7 package for DSM and SRM (up to the certificate configuration)

@mreid-tt do you have the required infrastructure for a full package test? (you need a DNS-01 provider like cloudflare, azure dns, ali dns, and others).

mritd:
@hgy59, thanks for the follow-up and for offering to test ARMv7. I removed the blanket 32-bit exclusion. The package now enables the ARM and i686 targets supported by Go, while PPC remains excluded by the spksrc Go framework.I also added an upstream Synology-tagged compile check for 386, ARMv5, ARMv7, amd64 v1, and arm64. Packages
build successfully for ARMv7 DSM 7.2, evansport DSM 7.1, and all valid targets covered by all-supported.

I also updated cross/dnsacme to use the framework's default go_install_target for the binary, with a custom `POST_INSTALL_TARGET` used only for the UI files. I have not enabled SRM support in this PR because writing the configuration currently requires elevated permissions and the certificate deployment workflow has not been tested end
to end on SRM. Could you please test the ARMv7 DSM package after this update? We can handle SRM separately once its configuration permissions and complete certificate workflow can be validated.

hgy59 (approved):
LGTM

mritd:
@mreid-tt Should your log be cleaned up? It seems to contain sensitive information such as domain names and email addresses.

PR #7312, opened 2026-07-19
copyparty: Initial package v1.20.18

### LINKED ISSUE #6672: []Package Request] copyparty
### Software Name

copyparty

### Brief Description

Turn almost any device into a file server with resumable uploads/downloads using any web browser

Server only needs Python (2 or 3), all dependencies optional
🔌 protocols: http // webdav // ftp // tftp // smb/cifs

### Website

https://github.com/9001/copyparty

### Documentation

https://github.com/9001/copyparty/blob/hovudstraum/README.md

https://github.com/9001/copyparty/blob/hovudstraum/docs/README.md

### Source code

https://github.com/9001/copyparty.git

### OpenSource License

MIT

### Other License

https://github.com/9001/copyparty/tree/hovudstraum?tab=MIT-1-ov-file

mreid-tt:
## Description

- Add copyparty v1.20.18, a portable file server with resumable uploads, file sharing, and WebDAV.

Fixes #6672

## Checklist

### Type of change

- [x] New Package


mreid-tt:
hey @th0ma7, this is a relatively simple package but I thought I'd ask for a second pair of eyes to look at it if you don't mind.

spk/copyparty/Makefile
```diff
@@ -0,0 +1,39 @@
+SPK_NAME = copyparty
+SPK_VERS = 1.20.18
+SPK_REV = 1
+SPK_ICON = src/copyparty.png
+
+PYTHON_PACKAGE = python314
+SPK_DEPENDS = "ffmpeg8"
+
+WHEELS = src/requirements-pure.txt
+WHEELS_PURE_PYTHON_PACKAGING_ENABLE = 1
```
th0ma7, line 10: This could probably be avoided and making the package even leaner.

spk/copyparty/Makefile
```diff
@@ -0,0 +1,39 @@
+SPK_NAME = copyparty
+SPK_VERS = 1.20.18
+SPK_REV = 1
+SPK_ICON = src/copyparty.png
+
+PYTHON_PACKAGE = python314
```
th0ma7, line 6: I wonder if you remove `PYTHON_PACKAGE` and add it to `SPK_DEPENDS` the use spksrc.spk.mk instead of meta.  Would that work?  meaning using pure wheels will force full download on the target NAS.  There is basically nothing else to package.  Would be worth a try.

Package could even possibly be a `noarch` ... if all wheels are really pure-python wheels.

th0ma7 (commented):
Good as-is but I would try some of the suggestions making the package even simpler and leaner.

spk/copyparty/Makefile
mreid-tt: As the package is already very small I would prefer keeping it to bundle wheels in the SPK so users don't need internet access on the NAS to download pure-python wheels at install time.

spk/copyparty/Makefile
mreid-tt: I re-checked the requirements with the script from hgy59 and realised that I missed a cross dependency which has since been added. Not sure we can go this route anymore.

th0ma7: You just can't anymore with cross.

PR #7313, opened 2026-07-19
capability: declare MIN_GCC_VERSION / MIN_GLIBC_VERSION / REQUIRE_64BIT instead of arch lists

th0ma7:
## What

Introduces `spksrc.common/tc-capability.mk`: a package declares the toolchain
capability it needs instead of hard-coding the architectures where it fails.

```make
MIN_GCC_VERSION   = 8       # needs gcc 8 or newer
MIN_GLIBC_VERSION = 2.20    # needs glibc 2.20 or newer (a runtime floor)
REQUIRE_64BIT     = 1       # needs a 64-bit target
```

Everything is resolved statically, from values declared in each toolchain's own
Makefile beside `TC_DIST` — `TC_GCC`, `TC_GLIBC`, `TC_KERNEL`. `TC_GCC` is now read
there rather than via `gcc -dumpversion`, so the answer never depends on how a
toolchain was last built. `pre-check.mk` turns an unmet requirement into the
arch-refusal error, right beside `UNSUPPORTED_ARCHS`; several unmet capabilities
accumulate into one message.

## Why

`UNSUPPORTED_ARCHS = <list>` says *where* a package fails, not *why*. It has to be
rechecked by hand every time a toolchain moves, and it cannot express "any arch
whose gcc is older than X" — which is what most of those lists actually meant.
`REQUIRED_MIN_DSM` / `REQUIRED_MAX_DSM` / `REQUIRED_MIN_SRM` and `UNSUPPORTED_ARCHS`
stay for genuine OS-version and per-arch constraints that are not a capability floor.

## The floors are tested, not guessed

Each cross codec was built in every stock gcc on the ladder (4.3.7 ppc853x, 4.6.4
88f6281, 4.8.3 hi3535, 4.9.3 qoriq) and bracketed to where it starts to compile:

| floor | packages | why |
|---|---|---|
| 4.6 | x264, shine, libvpx-1.14 | endian probe / `__builtin_bswap16` / C++11 rejected by 4.3.7 |
| 4.8 | x265, libaom-3.8, dav1d, harfbuzz, libass, libzimg, snappy, chromaprint-fftw, librist | C++11 / atomics / HDR10+ |
| 4.9 | liblc3 | `arm_acle.h` (added to gcc in 4.9) |
| 7.5 | frei0r, libvpx-latest, libaom-latest, libjxl, libvmaf, rubberband, highway, vvenc | C++17 |
| — | svt-av1 | `REQUIRE_64BIT` + gcc 4.9 |

`ffmpeg7` / `ffmpeg8` declare `MIN_GCC_VERSION = 4.9` (C11 + working atomics), and
the packages built against ffmpeg8 — `tvheadend`, `comskip`, `chromaprint` (spk and
cross alike) — carry the same 4.9 floor instead of re-listing armv5 / armv7l /
old-ppc by hand.

Packages keep their non-gcc constraints (`harfbuzz` keeps its
`UNSUPPORTED_ARCHS_TCVERSION`; the DSM ≥ 7.0 codecs keep `REQUIRED_MIN_DSM`).

## Behaviour

Same archs refused as before wherever the list was a gcc proxy, plus the DSM 5.2
stock gccs the lists silently missed (they were attempted and failed deep in the
build) — now refused up front with a reason. No arch that builds today is refused:
the floor is exactly where the package stops compiling.

Documented under
[Architecture Support](https://synocommunity.github.io/spksrc/developer-guide/packaging/makefile-variables#architecture-support)
and a Framework Changes entry (with a Highlights bullet).

## Other changes, in support of the above

- **`pre-check.mk`** — quote the echo that records an unsupported arch, so a
  capability reason containing `<` (e.g. `gcc 4.6.4 < 4.9`) is not read by the shell
  as an input redirection. Unquoted, the echo failed, the record was dropped, and CI
  counted an expected capability refusal as a real build error.
- **`build_status.sh`** — extract only the package name (non-greedy `.*?:`) when
  clearing unsupported packages from the error list, so a reason that itself
  contains a colon — which the capability messages do — no longer over-captures and
  breaks the match (it also injected the reason's parentheses into the grep).
- **`tc_vars`** — emit `TC_KERNEL` once, leaving the `>= 4.4` EXTRAVERSION `+`
  block as its sole emitter (adding `TC_GCC`/`TC_GLIBC` had duplicated it).


hgy59:
@th0ma7 I like the static approach (Everything is resolved statically, from the values in each toolchain's own
Makefile)
So we don't need to extract the toolchain to evaluate the gcc version just to find out it is not supported...

th0ma7:
@hgy59 thnx for your feedback.  Code still requires more work but should make things more agile and based on actual capabilities rather then unsing combinations of fields that does not resolve things well (min_dsm + unsupported that can now just use min_gcc).

Also, to include gcc8 cross-compiler overlay this needs to fundamentally change as min_dsm will not hold true anymore (it never fully did anyway).

And heads-up, i was able to build a gcc8 cross-compiler overlay for dsm-5.2 and fully compile an ffmpeg8 package...  Which gives an idea of just how much more we may be able to support (or how lesser compatibility issues there will be)

th0ma7:
@hgy59 and @mreid-tt this one is in theory ready for merging - I'll add documentation changes and let ghaction complete to confirm.  Feel free to review as needed.  And really PR looks bigger than it is in reality.

The most important changes:
1. earlier and static detection of GCC version... allowing to use `MIN_GCC_VERSION` anywhere instead of combination of `UNSUPPORTED_ARCHS` with `REQUIRED_MIN_DSM`
2. Allow usage of `MIN_GLIBC_VERSION`
3. Allow usage of `REQUIRE_64BIT`

Noting that this is a prerequisite for the later PR related to gcc8 toolchain overaly as for it to work things will have to be gcc version specific and no longer using `UNSUPPORTED_ARCHS` and/or `REQUIRED_MIN_DSM`.  Therefore as a start, the complex `ffmpeg*` related packages are being migrated to this new paradigm.  We can then slowly migrate others afterwards.

cross/libvpx/Makefile
```diff
@@ -3,10 +3,12 @@ PKG_NAME = libvpx-virtual
 OPTIONAL_DEPENDS  = cross/libvpx-latest
 OPTIONAL_DEPENDS += cross/libvpx-1.14
 
+MIN_GCC_VERSION = 4.6
+
 include ../../mk/spksrc.common.mk
 
 # GCC >= 5 (DSM >= 7.0)
-ifeq ($(call version_ge, $(TC_GCC), 5),1)
+ifeq ($(call version_ge, $(TC_GCC), 7.5),1)
```
hgy59, line 11: I guess the comment should match the test
```
# GCC >= 7.5 (DSM >= 7.0)
```

th0ma7: Thnx, fixed here: https://github.com/SynoCommunity/spksrc/pull/7325/changes#diff-875e2b72515ffddc7041e75060687c73a37bda260ca8251e57ec8f3aca68c666

PR #7314, opened 2026-07-19
toolchain: declare the ABI and link flags once (TC_EXTRA_BUILD_FLAGS / TC_EXTRA_LDFLAGS)

th0ma7:
## What

Move the toolchain-wide compiler/linker flags out of individual packages and onto
the toolchain that actually determines them, in two parts:

- **The target ABI** (`-march`, `-mcpu`, `-mfpu`, `-mfloat-abi`, `-mthumb`, ...):
  `TC_EXTRA_CFLAGS` is renamed **`TC_EXTRA_BUILD_FLAGS`** and folded once into every
  per-language `TC_EXTRA_<LANG>FLAGS` (C / CPP / CXX / Fortran) **and** into
  `TC_EXTRA_LDFLAGS`.
- **Link libraries** `-lrt` / `-latomic`: declared once through
  **`TC_EXTRA_LDFLAGS`**, dropping the per-package copies.

## Why

The ABI selects how objects are built, so it must reach *every* language and the
link — building C++/Fortran objects with a different ABI than the C they link
against silently breaks binaries, and the gcc link driver reads these flags to pick
the right multilib/startfiles. And `-lrt`/`-latomic` depend only on the target, yet
~10 packages carried the knowledge. Stating both once on the toolchain is the same
"declare what you need, not where it breaks" move as the capability PR.

## How

- **`tc-flags.mk`** folds `TC_EXTRA_BUILD_FLAGS` (the ABI) once into each
  `TC_EXTRA_<LANG>FLAGS` and `TC_EXTRA_LDFLAGS`. Each becomes the single residual
  list its language reads — ABI first, then any toolchain extra, always last, a
  clean place to extend. `CFLAGS`/`CPPFLAGS`/`CXXFLAGS`/`FFLAGS`/`LDFLAGS` and the
  cmake/meson crossfiles read their own per-language variable. `TC_EXTRA_RUSTFLAGS`
  is left out on purpose (rustc takes the ABI via `-Ctarget-cpu`; a crate's C deps
  via `CFLAGS_<target> = TC_EXTRA_CFLAGS`).
- **`-latomic` is kept only where the toolchain's gcc ships libatomic** — asked of
  the compiler (`-print-file-name=libatomic.so`), not tabulated: a gcc that old
  predates the `__atomic_*` builtins and emits `__sync_*`, so it never needs the
  library. Both `-lrt` and `-latomic` are linked
  `-Wl,--as-needed ... -Wl,--no-as-needed`, so — now that they are declared
  toolchain-wide — a binary records a `librt`/`libatomic` dependency only when it
  truly references one; the wrap is scoped to just these two libs.
- **`tc_vars.mk`** bakes the per-language vars and `TC_EXTRA_LDFLAGS`.
- Declared on the ARMv5 / PowerPC / glibc&lt;2.17 toolchains (`88f6281`, `qoriq`,
  `ppc853x`, `armada370/375/xp-5.2`, `evansport-5.2`). Modern toolchains unchanged.

## Removed

The per-package `-lrt` in `zstd`, `cups`, `flac`, `lz4`, `gnupg`, `htop`,
`mosquitto`, `fdupes`, `imlib2`, `misc-util-linux`: the glibc&lt;2.17 toolchains now
carry it, and on newer glibc it was a no-op (`clock_gettime` is in libc). The
per-package `-latomic` in `python*` and `stockfish` is **kept** — those build on
ARMv7, which these toolchain declarations do not cover.

## Docs

Framework toolchain reference (`docs/framework/toolchain.md`) and the changelog
document `TC_EXTRA_BUILD_FLAGS`, the per-language flags, and the `-lrt`/`-latomic`
auto-detection and `--as-needed` behaviour.


cross/zstd/Makefile
```diff
@@ -19,8 +19,4 @@ ifeq ($(call version_lt, ${TC_GCC}, 5),1)
 ADDITIONAL_CFLAGS = -std=gnu99
 endif
 
-ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS) $(OLD_PPC_ARCHS)),$(ARCH))
-ADDITIONAL_LDFLAGS = -lrt 
-endif
-
```
hgy59, line 25: This change has broken the build of zstd for ARMv5_ARCHS (and I guess for OLD_PPC_ARCHS too)

hgy59: Sorry for the noise, I didn't habe updated my (cached) toolchains

PR #7318, opened 2026-07-21
forgejo-lts: renamed from forgejo 15.0.4 & update to 15.0.5

nhymxu:
## Description

Follow on discuss at #7308
- Rename from `forgejo` to `forgejo-lts`
- Change service port to `8621` to prevent conflict with current forgejo install
- Update to v15.0.5 LTS

## Checklist

### Type of change

- [ ] Bug fix
- [x] New Package
- [x] Package update


hgy59:
We should not rename this package (from forgejo to forgejo-lts) if we can't keep the port on 8620.
Since we don't want to create non LTS packages, the only change should affect the DISPLAY_NAME
/cc @th0ma7 @mreid-tt 

Another point to avoid the rename is, that it is difficult to migrate current forgejo installations to forgejo-lts.
I tried `INSTALL_REPLACE_PACKAGES = forgejo` to automatically uninstall forgejo after installation of forgejo-lts, but this does not work due to the port 8620 already in use.

mreid-tt:
> Another point to avoid the rename is, that it is difficult to migrate current forgejo installations to forgejo-lts.
> I tried `INSTALL_REPLACE_PACKAGES = forgejo` to automatically uninstall forgejo after installation of forgejo-lts, but this does not work due to the port 8620 already in use.

Actually, this was the exact strategy we used for the migration from Sonarr v3 to Sonarr v4. For the new package we did the following in the Makefile:

```makefile
# replace previous legacy package
INSTALL_REPLACE_PACKAGES = "nzbdrone>=20210329-19"
USE_DEPRECATED_REPLACE_MECHANISM = yes
# disable checking for port conflicts
CHECKPORT = no
```

It's not be best solution but it is possible if really required. In this case, I agree with you that it probably isn't.

PR #7363, opened 2026-08-03
ps3netsrv-plus: Add package

nathanchere:
## Description

Adds the `ps3netsrv-plus` package.

## Checklist

- [x] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created - **there is an admin UI included which I don't think needs additional documentation**

### Type of change

- [x] New Package* (not technically a NEW package, but first time publishing it to synocommunity)


hgy59:
@nathanchere Please provide information here about the difference between ps3netsrv-plus and the existing ps3netsrv package.

nathanchere:
The existing package is just the basic ps3netserv. You install the package, pick a shared folder and it runs.

Some key things this adds - better wizard / onboarding, alllows serving only from specific folders instead of entire shares, has safeguards to ensure appropriate permissions are in place before it tries starting ps3netserv e.g. read-only access, has a detailed XMB-styled admin UI, sanity check view of the game library (originally was for troubleshooting why parts of my library weren't showing up, turns out it's also just a nice thing to have),  easy access to ps3netserv logs.

The two things the other ps3netserv package does that mine doesn't are 1) serves on a non-standard NETISO port (presumably to avoid collision with Synology's default CalDAV port, I consider that a poor trade-off), 2) appears to support older DSM versions where as mine only explicitly supports DSM 7.x

Screenshots of the UI:

image
image
image

If this isn't of interest here no problem, I can continue serving it from my own repository instead. But I intend to add some more quality-of-life features and figured it might be of interest to more people than just me and a handful of my immediate friends.

nathanchere:
Fuck it, I'll host it myself.

PR #7370, opened 2026-08-05
Syncthing: Update to v2.1.3

mreid-tt:
## Description

Updates Syncthing from v2.0.16 to v2.1.3, plus related cleanups and documentation.

### What changed

- **Update to v2.1.3** — minor release; changes are additive (folder/device grouping in the GUI, HTTP/HTTPS CONNECT proxy support, optional `blockIndexing` folder attribute, configurable GUI session cookie). No breaking changes; Syncthing migrates its config/database automatically.
- **`config.xml` cleanups for 2.1**
  - Config `version="35" → "52"` (2.1.3's current config version, avoids first-run migration)
  - Removed `<cacheIgnoredFiles>` (removed upstream in 2.1)
  - Removed `<databaseTuning>` (obsolete since the SQLite database switch)
- **Removed vestigial `STGUIASSETS` export** from `service-setup.sh` — points at a `gui/` directory that was never shipped; the GUI is embedded in the binary since the v1.23-era "next-gen GUI" became standard.
- **Wizard polish** — blank line between the intro paragraphs and `user-select: all` on the full `options.conf` path for easy copying (install + upgrade, EN + FR).
- **Docs** — documented the self-update behavior (Syncthing auto-updates every 12 h, so Package Center's version may lag the running binary; the package preserves a newer binary on upgrade), corrected the data locations, and added the missing 21027/22000 ports to both the package page and `docs/reference/package-ports.md`.

Fixes # 

## Checklist

### Type of change

- [x] Package update
- [x] This change requires a documentation update (e.g. [docs.synocommunity.com](https://docs.synocommunity.com))

### DIFF
--- a/docs/packages/syncthing.md
+++ b/docs/packages/syncthing.md
@@ -18,7 +18,7 @@ Syncthing is a continuous file synchronization program that syncs files between 
 | Package Name | syncthing |
 | Upstream | [syncthing.net](https://syncthing.net/) |
 | License | MPL-2.0 |
-| Default Port | 8384 (Web UI), 22000 (Sync) |
+| Default Port | 8384 (Web UI), 21027 (Discovery), 22000 (Sync) |
 
 ## Features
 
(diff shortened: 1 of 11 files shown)


docs/packages/syncthing.md
```diff
@@ -18,7 +18,7 @@ Syncthing is a continuous file synchronization program that syncs files between
 | Package Name | syncthing |
 | Upstream | [syncthing.net](https://syncthing.net/) |
 | License | MPL-2.0 |
-| Default Port | 8384 (Web UI), 22000 (Sync) |
+| Default Port | 8384 (Web UI), 21027 (Discovery), 22000 (Sync) |
```
hgy59, line 21: we should'n name it "Default Port", I would prefer "Used Port" and "Used Ports".
There is no way to change the ports of the packages (at least none that we support).
Maybe we should mention when a package "does not use the default port" like we have in PostgreSQL (to avoid conflict with DSM pgsql),

(affects all packages with service ports)...

mreid-tt: I was sticking with the standard established in the docs for 10 other packages. Sure we can open a PR to change it but we should also change the 10 others at the same time. (see [list](https://github.com/search?q=repo%3ASynoCommunity%2Fspksrc+path%3Adocs%2F+%22%7C+Default+Port+%7C%22&type=code))

hgy59: > but we should also change the 10 others at the same time

Yes, absolutely, it just appeared here...

PR #7380, opened 2026-08-10
Rust: Add initial package v1.97.1

dotWee:
## Description

Adds a new package that installs the official Rust toolchain (rustc, cargo, rustfmt, clippy) on the NAS, in the spirit of the existing Erlang/Python/Ruby language packages.

Design decisions:

- Repackages the official statically-targeted `musl` host toolchains from rust-lang.org for **x64** and **aarch64** (the only targets with upstream host tools), so the toolchain does not depend on the DSM glibc version.
- DSM ships no musl loader, so the package bundles the musl runtime + `libgcc_s` from Alpine (pinned, sha256-verified secondary distfiles) under `share/rust-support` and patches the PIE toolchain binaries with `patchelf` to use them. The non-PIE `rust-lld` is invoked through the loader by a small wrapper script.
- `rustc` is pointed at the bundled `rust-lld` (`-C linker` + `-C linker-flavor=ld.lld`), so **no C toolchain is required on the NAS** for pure-Rust crates. Produced binaries are statically linked (musl crt-static) and run standalone.
- Crates requiring a C compiler (e.g. `openssl-sys`, `ring`) additionally need SynoCli Development Tools (documented limitation).

Container validation (no musl present on the system, toolchain staged at `/var/packages/rust/target`, x64 + aarch64): `rustc`/`cargo`/`rustfmt`/`clippy-driver` run, `rustc hello.rs` produces a standalone binary, `cargo build` works for dependency-free crates, a build-script crate, and a proc-macro crate (`serde` derive), plus `cargo clippy`, `cargo fmt` and `cargo test` incl. doctests.

## Checklist

### Type of change

- [x] New Package


dotWee:
Update: Quick smoke tests validated on my DS920+ with success.

Try it!

```bash
rustc --version && cargo --version        # tools run
printf 'fn main(){println!("hello dsm");}\n' > /tmp/h.rs
rustc /tmp/h.rs -o /tmp/h && /tmp/h       # compiles + runs standalone

cd /tmp && cargo new demo && cd demo && cargo build --release && ./target/release/demo
```

hgy59:
@dotWee what is the use case for installing rust toolchains on DSM?

We normally cross build rust code and create packages with the binaries on an x64 host.
The comparision with Erlang and Python is not appropriate (I don't know ruby)
The erlang package is just the runtime for dependent packages (rabbitmq).
The python package is also the runtime for dependent packages and those must come with cross compiled wheels.

/cc @th0ma7

dotWee:
In my case, for building small rust applications and making [cargo](https://doc.rust-lang.org/cargo/) available. Also, I didn't want to bloat my DSM by installing rust through the [rustup](https://rustup.rs/) installer script.

Sure, building for the arch on remote + scp'ing the binary is also ok - this feels cleaner tho.

A random use case by @LowLevelLogic: https://github.com/LowLevelLogic/synology-dedup-photos (great work btw!)

th0ma7:
Interesting proof of concept.  Normally we don't ship pre-built binaries, we always recompile from source.  Therefore I cannot support such package, at least not in its current form.

Your usecase has two different angles
1. Dedup photo should be rustc built from sources for each archs which would make it available to a much greater spectrum of devices such as armv7 along with probably most dsm6 archs.
2. Providing a rustc compiler... As well should be recompiled from sources which would allow targetting most Synology archs, linked against glibc default Synology toolchain.  This avoids having to bundle yet another c library.  Also the question is, should it be bundled part of the synocli-devel package for instance?  Although it size could become problematic.

I would suggest you look for alternate 1, as we Synocommunity devs could later look are item 2.  At least that's what I'd suggest as path forward.

dotWee:
Indeed a POC, trimmed to my current uses. Reasonable comments, feel free to close this PR without further notice. If I find time to contribute a full build-by-source setup w/ older arch & DSM6 support, I'll update my branch.  

Thanks for running SynoCommunity & greets from zurich! Your work is really appreciated!

hgy59:
@dotWee I had a look at synology-dedup-photos project.
It is in a very early state, no release was created yet (and development just started two monts ago)

I tried to build with the latest commit of Aug 9th.
It needs cross/openssl3 as dependency.

But it fails to build
x64-7.1:
compiling id ok, but the linker fails due missing some external symbols like:
```
          /spksrc/cross/dedup-photos/work-x64-7.1/synology-dedup-photos-2210ae14215c8c5e3690c99a7a984f0cc21fcff0/target/x86_64-unknown-linux-gnu/release/deps/libort_sys-5d305c2558e4e0f6.rlib(time_zone_info.cc.o): In function `absl::lts_20250814::time_internal::cctz::TimeZoneInfo::Description[abi:cxx11]() const':
          time_zone_info.cc:(.text._ZNK4absl12lts_2025081413time_internal4cctz12TimeZoneInfo11DescriptionB5cxx11Ev+0x1b): undefined reference to `std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::basic_ostringstream()'
          collect2: error: ld returned 1 exit status
          
  = note: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-lib)
```

and cross compiling fails too:
```
  cargo:warning=aarch64-unknown-linux-gnu-gcc: error: unrecognized command line option '-m64'

  --- stderr

  thread 'main' (73689) panicked at /spksrc/distrib/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/openssl-sys-0.9.111/build/main.rs:325:13:

  Header expansion error:
  Error { kind: ToolExecError, message: "command did not execute successfully (status code exit status: 1): LC_ALL=\"C\" \"/spksrc/toolchain/syno-aarch64-7.1/work/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc\" \"-O0\" \"-ffunction-sections\" \"-fdata-sections\" \"-fPIC\" \"-m64\" \"-I\" \"/spksrc/cross/dedup-photos/work-aarch64-7.1/install/usr/local/synology-dedup-photos/include\" \"-I/spksrc/toolchain/syno-aarch64-7.1/work/aarch64-unknown-linux-gnu/aarch64-unknown-linux-gnu/sysroot/usr/include\" \"-I/spksrc/cross/dedup-photos/work-aarch64-7.1/install/usr/local/synology-dedup-photos/include\" \"-E\" \"build/expando.c\"" }
```

So I think we have to wait for a stable release of synology-dedup-photos anyway...

hgy59:
> Thanks for running SynoCommunity & greets from zurich! Your work is really appreciated!

Thank you very much (only a few kilometers from Zurich).

hgy59:
> If I find time to contribute a full build-by-source setup w/ older arch & DSM6 support, I'll update my branch.

This might be useful to start - my initial cross/dedup-photos/Makefile:
```
PKG_NAME = synology-dedup-photos
# Version taken from Cargo.toml	
PKG_VERS = 0.2.0
PKG_EXT = tar.gz
# no official release yet
# git hash from August 9th 2026
PKG_GIT_HASH = 2210ae14215c8c5e3690c99a7a984f0cc21fcff0
PKG_DIST_NAME = $(PKG_GIT_HASH).$(PKG_EXT)
PKG_DIST_SITE = https://github.com/LowLevelLogic/synology-dedup-photos/archive
PKG_DIST_FILE = $(PKG_NAME)-git$(PKG_GIT_HASH).$(PKG_EXT)
PKG_DIR = $(PKG_NAME)-$(PKG_GIT_HASH)

DEPENDS = cross/openssl3

HOMEPAGE = https://github.com/LowLevelLogic/synology-dedup-photos#readme
COMMENT = rust tool to compare, preview and similar and duplicate pictures.
LICENSE = MIT

CARGO_BUILD_ARGS = --locked

# Avoid adding -m64 flag in cc-crate that breaks cross compiling
ENV += CRATE_CC_NO_DEFAULTS=1

include ../../mk/spksrc.cross-rust.mk
```

hgy59:
And for cross-compilation to work, we may need to patch the cc-crate similarly to how we do for cross/helix.

EDIT:
Just found that cc-crate patch is not needed when building with `CRATE_CC_NO_DEFAULTS=1` (this disables default flags like -m64 to fix cross compilation).

PR #7390, opened 2026-08-19
library updates

hgy59:
## Description

- update widely used libraries and tools

Fixes # 

## Checklist

- [ ] Any needed [documentation](https://docs.synocommunity.com/contributing/) is updated/created

### Type of change

- [x] Library update


hgy59:
Build validated on my fork before creation of this PR.

Due to the large amount of affected packages, most of the targets reach the 6h limit in this PR.
Validation was done with partial changes (libxml2) and additional manual builds (esp. for DSM 5.2),

Extra note for libxml2:
This update brings libxml2 v2.13.9
There are libxml2 2.14 and 2.15 available but those come with breaking changes (hence the lib version changed from libxml2.so.2 to libxml2.so.16).
Most of the packages were successfully built with libxml2 2.15 but some failed.

An update to libxml2 > 2.13.x should be done (later) in a dedicated PR (if needed).
If we still need libxml2 2.13 it should be renamed to cross/libxml2-legacy so only incompatible packages must be adjusted.
But we will have to ensure that packages do not depend on both (like (synocli-* or others).
Otherwise it is not deterministic whether libxml2.so links to libxml2.so.2 or libxml2.so.16 and runtime errors are to be expected.

PR #7353, opened 2026-07-30
rust: custom toolchains for legacy DSM archs (PowerPC SPE, ARMv5, x86) as native/rustc-1.82 + python311

th0ma7:
## What

Give the legacy DSM toolchains a usable **Rust** compiler so Rust-backed Python wheels (`cryptography`, `pydantic-core`, `rpds-py`, …) build on archs rustup ships no suitable prebuilt `rust-std` for — Tier-3 targets, or Tier-1/2 whose prebuilt std targets a newer glibc than the DSM toolchain.

| toolchain | base rust target | overlay custom rust target | tier | gcc / glibc |
|---|---|---|---|---|
| `syno-qoriq-6.2.4` | `powerpc-unknown-linux-gnuspe` |`powerpc-synology-linux-gnuspe` |  3 | 4.9.3 / 2.20 |
| `syno-ppc853x-5.2` | `powerpc-unknown-linux-gnuspe` | `powerpc-synology-linux-gnuspe` | 3 | 4.3.7 / 2.8 |
| `syno-88f6281-5.2` | `armv5te-unknown-linux-gnueabi` | `armv5te-synology-linux-gnueabi` | 2 | 4.6.4 / 2.15 |
| `syno-88f6281-6.2.4` | `armv5te-unknown-linux-gnueabi` | `armv5te-synology-linux-gnueabi` | 2 | 4.6.4 / 2.15 |
| `syno-x86-5.2` | `x86_64-unknown-linux-gnu` | `x86_64-synology-linux-gnu` | 1 | 4.7.3 / 2.17 |

## How

**`native/rustc-1.82` — the toolchain is built as a host-native package.** Uses the official `rustc-1.82.0-src` (which bundles `llvm-project`, so LLVM is built tailored to the target arch). `mk/spksrc.native-toolchain.mk` dispatches to `mk/spksrc.native/toolchain-rust.mk` (the x.py LLVM/stage1/stage2 build) and produces a final `rust-<id>-<rev>.txz`:
```
make -C native/rustc-1.82 arch-<arch>-<vers>
```

[...]

PR #7391, opened 2026-08-21
gcc 8.5 overlay: a modern compiler for DSM 5.2/6.2.4/7.0, as its own component

th0ma7:
Ports the gcc 8.5 overlay prototyped in #7324 onto the framework merged by #7353, adopting its strategy: **one component, one archive**. binutils 2.30 and gcc 8.5 are separate overlays, either rebuildable without the other, composed at install time.

Draft: the DSM 5.2 activation in `build.yml` is a TEMP commit to exercise the overlay in CI, and gets reverted before merge.

## Ground already cleared

This PR started much larger. **Nine pieces** of it have been split out, reviewed and merged on their own, each one standing without the overlay — so what remains here is the overlay itself.

| PR | What it took out of this one |
|----|------------------------------|
| [#7397](https://github.com/SynoCommunity/spksrc/pull/7397) 2117| `librist` / `libsrt` endian shims for glibc 2.8, under an arch-scoped `patches/ppc853x/`; the `MIN_GCC_VERSION` gates they no longer need |
| [#7430](https://github.com/SynoCommunity/spksrc/pull/7430) | ffmpeg 4–8 gating on toolchain capability instead of arch lists; `synocli-videodriver` selected by `REQUIRED_MIN_DSM` |
| [#7431](https://github.com/SynoCommunity/spksrc/pull/7431) | `AVAILABLE_TOOLCHAINS` — a live bug where `filter-out %-rust` matched nothing, so consumer directories became phantom architectures. This PR adds 121 more of them, which is what made it worth fixing first |
| [#7433](https://github.com/SynoCommunity/spksrc/pull/7433) | the `--as-needed` bracket, removed for everyone rather than scoped to `OVERLAY_GCC_ON` as it was here. `dependency-list` keyed on the folder, without which `ffmpeg4` never entered the CI matrix |
| [#7396](https://github.com/SynoCommunity/spksrc/pull/7396) | build logs that keep what the console showed — a pre-check refusal, make's own errors — and status lines that name the toolchain directory, so `88f6281-5.2_gcc-8.5` and `_binutils-2.30` are visible while this builds |
| [#7439](https://github.com/SynoCommunity/spksrc/pull/7439) | an arch exclusion that says *why*, and names every blocker rather than the first. `make check-<arch>-<dsm>` walks the tree and lists each capability gate an arch fails — the tool this PR is read with |
| [#7442](https://github.com/SynoCommunity/spksrc/pull/7442) | each build tree owning its `tc_vars`, the toolchain owning none. This PR's original answer was a `toolchainclean` that erased the shared file between builds; the merged one removes the sharing instead |
| [#7441](https://github.com/SynoCommunity/spksrc/pull/7441) | `$(call tc,<tool>)`, and the twelve packages that spelled tool paths by hand. `TC_OVERLAY_GCC_PATH` is emitted there already, empty until this PR fills it |
| [#7454](https://github.com/SynoCommunity/spksrc/pull/7454) | ffmpeg 4–8 and x264 naming every tool instead of deriving them from `--cross-prefix` — the call sites of the macro above, which is how ffmpeg reaches the overlay compiler here |
| [#7458](https://github.com/SynoCommunity/spksrc/pull/7458)  | one list for the build-wide switches. `OVERLAY_SELECTORS` below becomes `FWRD_VARS += OVERLAY_GCC`, one line, and the eleven crossings stop being written by hand |

**Five more are in flight**, opened against `master` on September 19th. They are fixes this PR found but that stand on their own, so they should not wait for it:

| PR | What it takes out of this one |
|----|------------------------------|
| [#7465](https://github.com/SynoCommunity/spksrc/pull/7465) | the include-order half of #7464. Toolchain patches were applying from the wrong directory and `patch` then hung on its own stdin — ten CI archs killed at 5h48 having built nothing. `make digests` is fixed with the one-liner it needed instead. **`master` is broken until this merges** |
| [#7466](https://github.com/SynoCommunity/spksrc/pull/7466) | `strip.mk` choosing the runtime library by the UNION of what every shipped binary needs, not by the first one it matched. @hgy59's ffmpeg5–8 failures on ppc853x-5.2 are this bug |
| [#7467](https://github.com/SynoCommunity/spksrc/pull/7467) | `cross/libhdhomerun`, the last package spelling a toolchain path by hand — the tail of #7441 and #7454 |
| [#7468](https://github.com/SynoCommunity/spksrc/pull/7468) | `MIN_KERNEL_VERSION`, a runtime floor like glibc, and the Intel media stack declaring the 3.10 kernel it drives |
| [#7469](https://github.com/SynoCommunity/spksrc/pull/7469) | the overlay binutils built with no optimisation at all: an `ENV` line above the native include, overridden by `env-default.mk`'s empty `CFLAGS`. v1 and v2 are both unoptimised builds; v3 is ~2.2x on `as`. `native/gcc-8.5` here escapes it only by where its own `ENV` line sits |

**Eight commits** disappeared across the rebases onto them, all fully superseded: the `videodriver` selection (#7430 landed it, at `version_ge 6` rather than `6.2.4`), the `--as-needed` removal (#7433 went further, taking the overlay's private `-latomic` workaround with it), the shared-`tc_vars` clean and its `arch-%` half (#7442), the arch-exclusion reason (#7439), and the three that spelled out compiler paths for ffmpeg (#7441, #7454).

Rebased on master as of #7457; 24 commits, no conflicts.

## What the overlay is for

A package gated on `MIN_GCC_VERSION` is unavailable on archs whose vendor gcc predates it. The overlay installs a modern gcc *beside* the vendor one, selected by version suffix, and lets a package opt in. For ffmpeg8 (`MIN_GCC_VERSION = 4.9`) that is 11 archs it could never reach:

```
alpine-5.2  avoton-5.2  braswell-5.2  bromolow-5.2  cedarview-5.2
comcerto2k-5.2  hi3535-6.1  hi3535-6.2.4  monaco-5.2  x64-5.2  x86-5.2
```

## Activation

`OVERLAY_GCC` follows `OVERLAY_RUSTC` / `OVERLAY_BINUTILS`: off by default in `overlay.mk` and `local.mk`, overridable per package, per environment or on the command line. Three separate questions, as in #7353 — is an overlay AVAILABLE (`TC_OVERLAY_<c>`), REQUESTED (`OVERLAY_<c>`), ACTIVE (`OVERLAY_<c>_ON`).

gcc 8.5 cannot be used without binutils 2.30 — the vendor assembler cannot assemble what it emits — so **requesting gcc forces binutils on**. An explicit `OVERLAY_BINUTILS=0` alongside `OVERLAY_GCC=1` is honoured with a message saying it is unsupported, rather than silently ignored.

A compiler choice must hold for a package *and its whole dependency tree*: ffmpeg8 on gcc 8.5 cannot link a pcre built by gcc 4.3.7, the C++11 ABI differs. `OVERLAY_SELECTORS` is forwarded at every sub-make that re-enters the tree — **nine** sites: `depend.mk`, the three `tcvars` generations (`cross-cc.mk`, `spk.mk`, `kernel.mk`), the two PLIST walks (`copy.mk`, `plist.mk`), and the three in `supported.mk`. #7458 generalises exactly this, so once it lands the list here collapses to a single `FWRD_VARS += OVERLAY_GCC`.

The switches are exported too, but as a **backstop, not the mechanism**. The two differ in strength:

| | origin in the sub-make | outranks a package's own `OVERLAY_x = …` |
|---|---|---|
| `export` | `environment` | **no** |
| `OVERLAY_SELECTORS` | `command line` | **yes** |

Only a command-line variable can enforce one compiler across a tree; the export merely covers a sub-make nobody forwarded to. `supported.mk` matters for the same reason: it clears `MAKEFLAGS`, which does not lose a switch (make exports it as well) but demotes it to environment origin — so `make OVERLAY_GCC=0 all-supported` on ffmpeg8, which sets `OVERLAY_GCC = 1` in its own Makefile, quietly turned the overlay back on.

## The recurring theme: declared toolchain vs effective compiler

Most of the framework work is one bug in eight places — asking the toolchain Makefile what the compiler is, when an overlay has replaced it.

| | asked | now asks |
|---|---|---|
| `TC_HAS_LIBATOMIC` | vendor gcc | the compiler that will run |
| `MIN_GCC_VERSION` gate | `TC_GCC` | `TC_GCC_EFFECTIVE` |
| `TC_EXTRA_BUILD_FLAGS` | legacy toolchain | the overlay consumer's own |
| libstdc++ / libgcc search | sysroot order | overlay runtimes first |
| header search | `-I <target>/include` | `-idirafter`, so it no longer shadows gcc 8.5 |
| `-latomic` | inside `--as-needed` | outside it |
| transitive `DT_NEEDED` | `-L` only | `-L` **and** `--rpath-link` |
| PLIST collection | no selectors | selectors forwarded |

Two deserve spelling out.

**`-latomic` and `--as-needed`.** `LDFLAGS` precedes the objects on every link line, and `--as-needed` keeps a library only if it resolves something already undefined where it is seen. Nothing is, that early — so libatomic was dropped and openssl3's `libcrypto.so` came out with seven undefined `__atomic_*` and no `DT_NEEDED`. Harmless on the 2008 compilers, which ship no libatomic and have `-latomic` filtered out upstream, so legacy behaviour is byte-identical.

**`--rpath-link`.** `ld` does *not* use `-L` to resolve a shared library's transitive `DT_NEEDED`. A C program linking a C++ `.so` therefore kept finding the vendor `libstdc++.so.6` through the sysroot, even with the overlay's `-L` first, and failed on `std::__cxx11::basic_string` and `operator delete(void*, unsigned long)@CXXABI_1.3.9`. `--rpath-link` is link-time only and records nothing in the output.

**The PLIST one is the nastiest**, because it does not fail the build. The walk aborts on a dependency that re-evaluates against the legacy compiler, and everything listed after it is silently dropped. ffmpeg8 on x86-5.2 packaged green at 4.6 MB / 168 files with no `libavcodec`, no `libchromaprint`, no `libstdc++`. With the selectors forwarded: 51.9 MB / 534 files.

## Producer

`native/gcc-8.5` derives the target ABI from the vendor compiler rather than a hand-kept table — it is the source of truth for what its replacement must be interchangeable with. Two gaps that were silent:

- ppc853x/qoriq's 2008 gcc never *declares* its float ABI, leaning on `--with-cpu=8548` to imply it. A modern gcc does not, and built the overlay soft-float against a hard-float glibc.
- no Synology ARM gcc declares `--with-fpu`, so a guess of `neon` decided alone. armada370/375/xp have none; their overlay libgcc and libstdc++ came out full of NEON that only faults on the device. The compiler's own default is asked first, and only on ARM — a PowerPC gcc answers `none`.

Each gcc 8.5 consumer carries its own `TC_EXTRA_BUILD_FLAGS`, used instead of the legacy toolchain's while the overlay is active. Identical today by design; the indirection is the point, so a future gcc-12 is corrected next to itself.

## What blocks an arch, and what does not

An arch that never compiled a package also never had its configuration written, so the overlay surfaces several distinct floors at once. They are not the same kind of thing:

| floor | carried by | verdict |
|---|---|---|
| `MIN_GCC_VERSION` | compiler | **lifted** — this PR |
| `static_assert` | header macro | **shimmed** — one `--extra-cflags` |
| `clock_gettime` in librt | libc layout | **linked** — `-lrt`, outside `--as-needed` |
| v4l2 multiplanar API | kernel headers | feature disabled below Linux 3.2 |
| `REQUIRED_MIN_DSM` | package | not liftable |

The `static_assert` one is worth spelling out because it looks like a runtime floor and is not. glibc added it to `assert.h` in 2.16 as `#define static_assert _Static_assert`; the keyword itself comes from the compiler. Below 2.16 ffmpeg8 defines it itself, and keeps the arch.

Two configure gaps came from the same root: 88f6281 had no `--arch` block at all (ARMv5 was never buildable), and the v4l2 input device was disabled for `qoriq` by name when the real condition is a kernel older than 3.2 — which also covers 88f6281 and ppc853x, and correctly leaves qoriq's glibc 2.20 out of it.

## Fixes found along the way

- gcc archives no longer ship `*.la`: libtool bakes the producer's DESTDIR into `libdir`, and a consumer following one looks for a path that exists nowhere else. Consumers also strip them, which fixes the already-published archives without a rebuild.

## Validation

Two packages built end to end on archs whose vendor gcc could not build ffmpeg8 at all:

```
  x86-5.2        gcc 4.7.3   ffmpeg8_x86-5.2_8.1.2-4.spk        534 files
  88f6281-6.2.4  gcc 4.6.4   ffmpeg8_88f6281-6.2.4_8.1.2-4.spk  521 files
```

Both ship the overlay's `libstdc++.so.6.0.25` (GLIBCXX_3.4.25), and the ARMv5 one has `libsrt.so` correctly declaring `librt`. With the overlay off, `TC_EXTRA_LDFLAGS` is unchanged on 88f6281-6.2.4, x64-7.0, qoriq-5.2 and monaco-6.2.4.

## SPE PowerPC: gcc 8.5 mis-tags its own objects

GCC 8 moved the SPE port to `config/powerpcspe/` and stubbed that fork's own macros — `TARGET_SPE`, `TARGET_SPE_ABI` and `TARGET_E500_DOUBLE` to 0, `TARGET_FPRS` to 1 — while still accepting and honouring `-mabi=spe`. `TARGET_DF_FPR` therefore reads true in `rs6000_elf_file_end`, and every object comes out tagged `hard float` where the vendor e500 gcc tags `soft float` + `Vector: SPE`. `ld` warned on every mixed link — 36 lines in a single openssl3 build.

`native/gcc-8.5/patches/001-powerpcspe-soft-float-attribute.patch` asks the option variable `rs6000_spe_abi` instead of the stubbed macros, in attribute emission only. The generated assembly is identical to the unpatched compiler apart from the `.gnu_attribute` directive.

```
                    object attributes                              mixed link
  gcc 8.5 v2        hard float                                     36 warnings
  gcc 8.5 v3        soft float + Vector: SPE + Struct_Return       0
  vendor 4.3.2      soft float + Vector: SPE + Struct_Return       —
```

openssl3 rebuilt on ppc853x-5.2 with v3: no warnings, and `libcrypto.so.3` carries the vendor's three tags plus `NEEDED libatomic`. v3 archives for the two PowerPC consumers.

## Random list of candidates that might be fixed with this PR
list of packages that have ARCH-specific gates (not gated with MIN_GCC_VERSION, MIN_RUST_VERSION)
- cross/fd for ARMv7L and OLD_PPC
-


hgy59:
@th0ma7 since Rust 1.98 all aarch64-6.2.4 builds fail since rust calls gcc/linker with unsupported flag --fix-cortex-a53-843419.
This is hardcoded into rustc as pre_link_args that can't be disabled with -mno-fix-cortex-a53-843419.
Until we have gcc 8.5 overlay I propose to downgrade Rust (TC_RUST) from stable to 1.97.1 (the latest version that works for aarch64-6.2.4). Rust 1.98 was recently released on 20.08.2026.
do you agree?

th0ma7:
Fully agree.  We even have to create rustc overlays for all DSM 6 toolchains... But hopefully not.

Maybe we can open a bug upstream for this?

hgy59:
> Maybe we can open a bug upstream for this?

https://github.com/rust-lang/rust/issues/161486

hgy59:
> Fully agree. We even have to create rustc overlays for all DSM 6 toolchains... But hopefully not.

One solution would be to create a wrapper script for the linker to filter out the unsupported argument.
It is only an issue with aarch64, not all DSM 6 toolchains.

hgy59:
> > Maybe we can open a bug upstream for this?
> 
> [rust-lang/rust#161486](https://github.com/rust-lang/rust/issues/161486)

The first feedback is to upgrade gcc and not use a 10 year old version.
So we are on the the right track with this PR 👏

th0ma7:
I actually thought it was older than that 😆

hgy59:
@th0ma7 to limit the effort I propose to drop DSM 7.0 (toolchains etc.)
Since DSM 7.1 toolchains are available, there i no need for 7.0 toolchains anymore (all models that support DSM 7.0 do support at least DSM 7.1, and 7.0 is EOL anyway).

th0ma7:
@hgy59 and @mreid-tt got to a green build using the gcc85 overlay.  Would you mind trying this out if you hapen to have an older NAS (such as your ppc853x-5.2 or my 88f6281-6.2.4).  Feel free to try out this branch over other packages you may find relevant, noting that you'll have to enforce using the gcc overlay using `OVERLAY_GCC = 1` in your makefile or on the command line when invoking the builds.

Also for your information, I'm currently also testing a rebuild of the rustc toolchain using the gcc85 overlay.  First tests looks promising which could then lead into trying to upgrade current 1.82 version to something closer to mainline, fingers crossed...

Lastly, this still needs a lot more work, but slowly moving further away from a proof of concept to something we could actually work with.

mreid-tt:
hey @th0ma7 this is a fantastic initiative. As for testing I should have a DS410 with a ppc853x arch somewhere around here. I'll need to get some hard drives for it and validate that it still works after all this time but I'll get back to you.

th0ma7:
Also while at it I'd appreciate having your thought about this PR, more precisely about how to merge this beast?

I'm thinking I could create sub-PR and merge a few first pieces such as:
1. the gcc85 overlay native builder + consumers **(no activation, no framework enablement neither just yet)**
2. the binutils overlay native builder + consumers **(only activates with rustc overlay, like currently)**
3. likewise, updated rustc builder + consumers

This would trim this PR by a lot simplifying review and additional code changes needed in the framework.

Although I see one caveat: this will add a lot of directories under toolchain... like really a lot.

Suggesting we split that such as (and really feel free to propose something else):
```
toolchain/
├── binutils
├── gcc
├── rustc
└── synology
```
EDIT: or the following:
```
├── toolchain/ (only synology defaults)
├── toolchain-binutils/
├── toolchain-gcc/
└── toolchain-rust/
```

mreid-tt:
Hey @th0ma7, unfortunately, despite my best efforts the DS410 refuses to power on (even with a different power adapter), so it seems to be dead. Sorry I couldn't help with ppc853x testing after all.

As for the other questions, this is what my AI suggests — the ten extracted PRs already prove decomposition works, so: yes to sub-PRs, in this order:

1. Native builders (`native/gcc-8.5`, `native/binutils-2.30`, gmp/mpfr/mpc/isl) — self-contained.
2. Generated consumer dirs — plus a CI check that regenerating is a no-op, otherwise the checked-in output rots the first time someone hand-edits a consumer Makefile.
3. Framework activation (overlay.mk, selector forwarding, effective-compiler fixes) — the risky core, now small enough to review line by line.
4. Consumer opt-ins (ffmpeg chain etc.) — each provable in CI alone.

Keep this branch as the integration branch and cut sub-PRs from it, so rebases happen in exactly one place.

**Layout: the sibling roots.** After #7431 this isn't aesthetics — ~120 overlay dirs flat beside ~60 vendor dirs is a loaded footgun for every future `toolchain/` enumeration. The nested variant forces every existing `toolchain/syno-*` reference to move (churn on working paths, each moved glob a fresh #7431 chance). The sibling roots (`toolchain-gcc/` etc.) touch nothing that exists and say the true thing: vendor toolchains are fixed inputs, overlay components are built artifacts. Keep the new roots version-unqualified (gcc-12 comes eventually), and add a regression test that no overlay dir leaks into the vendor list — overlays compose (`rust-1.82_gcc-8.5`) and the combinatorics only grow.

**Sequencing:** settle the DSM 7.0-drop question *before* generating anything — it deletes a third of the new dirs.

hgy59:
> Hey @th0ma7, unfortunately, despite my best efforts the DS410 refuses to power on (even with a different power adapter), so it seems to be dead. Sorry I couldn't help with ppc853x testing after all.
> 

Are there any packages for ppc853x to test? (the latest build faild).
I still have a working DS210+ (but will be offline 8 days from monday to monday).

th0ma7:
> > Hey @th0ma7, unfortunately, despite my best efforts the DS410 refuses to power on (even with a different power adapter), so it seems to be dead. Sorry I couldn't help with ppc853x testing after all.
> 
> Are there any packages for ppc853x to test? (the latest build faild). I still have a working DS210+ (but will be offline 8 days from monday to monday).

Du to rebasing the previous succesfull build is now hidden. You'll find artifacts here https://github.com/SynoCommunity/spksrc/actions/runs/35233124469

hgy59:
@th0ma7 on ppc853x-5.2 only ffmpeg4 seems to work, ffmpeg5-8 fail all with the same output due lacking versions in libstdc++.so.6
see attached output of ffmpeg* --help
[ffmpeg4_help.txt](https://github.com/user-attachments/files/32398190/ffmpeg4_help.txt)
[ffmpeg8_help.txt](https://github.com/user-attachments/files/32398191/ffmpeg8_help.txt)

th0ma7:
It's exactly the info i need,thnx.

hgy59:
@th0ma7 the errors of ffmpeg5-8 on ppc853x-5.2 are fixed AFAICT.
Are there any tests that need to be carried out in practice?

[comskip_help.txt](https://github.com/user-attachments/files/32421813/comskip_help.txt)
[ffmpeg4_help.txt](https://github.com/user-attachments/files/32421814/ffmpeg4_help.txt)
[ffmpeg5_help.txt](https://github.com/user-attachments/files/32421815/ffmpeg5_help.txt)
[ffmpeg6_help.txt](https://github.com/user-attachments/files/32421816/ffmpeg6_help.txt)
[ffmpeg7_help.txt](https://github.com/user-attachments/files/32421817/ffmpeg7_help.txt)
[ffmpeg8_help.txt](https://github.com/user-attachments/files/32421818/ffmpeg8_help.txt)
[fpcalc_help.txt](https://github.com/user-attachments/files/32421819/fpcalc_help.txt)
[fpcalc_version.txt](https://github.com/user-attachments/files/32421820/fpcalc_version.txt)

th0ma7:
@hgy59 thnx a lot, glad this is fixed!  Nothing else to test for now.  I'll be working at extracting sub-PR from this to reduce its scope.  Also i noticed gcc and rustc overlays did not had any optimization leading to much longer builds.

New optimized overlays upcoming (order of magnitude faster) as well as (drum roll) optimized rustc 1.98.1 built using the gcc85 overlay !!!  That will definitely need a lot of testing when ready.

Thnx again!

PR #7395, opened 2026-08-22
Pin rust for aarch64-6.2.4

hgy59:
## Description

Rust 1.98 breaks the build for aarch64 with gcc 4.9.4
see: https://github.com/rust-lang/rust/issues/161486

- limit rust to 1.97.1 since 1.98 breaks the build for aarch64-6.2.4

This can be reverted when we once have gcc8 for DSM 6 toolchains available (see #7391 )

Fixes # 

## Checklist

- [x] Build rule `all-supported` completed successfully (locally verified the rust tools)
- [ ] Any needed [documentation](https://docs.synocommunity.com/contributing/) is updated/created

### Type of change

### DIFF
--- a/mk/spksrc.cross/env-rust.mk
+++ b/mk/spksrc.cross/env-rust.mk
@@ -39,6 +39,13 @@ endif
 
 ifeq ($(findstring $(RUST_ARCH), $(ARMv8_ARCHS)),$(RUST_ARCH))
 RUST_TARGET = aarch64-unknown-linux-gnu
+ifeq ($(call version_lt, ${TC_GCC}, 5),1)
+ifeq ($(TC_RUSTC),stable)
+# Rust 1.98 breaks the build with gcc 4.9.4
+# see: https://github.com/rust-lang/rust/issues/161486
+TC_RUSTC = 1.97.1
+endif
+endif
 endif
 
 ifeq ($(findstring $(RUST_ARCH), $(PPC_ARCHS)),$(RUST_ARCH))


hgy59:
@th0ma7 is this good to go?

mk/spksrc.cross/env-rust.mk
th0ma7, line 15: Maybe just fixing the version for all instead?

th0ma7 (approved):
I believe its good, haven't tested.  Although personnally I'd rather fix the version for all instead by changing stable to 1.97.1.

hgy59:
> I believe its good, haven't tested. Although personnally I'd rather fix the version for all instead by changing stable to 1.97.1.

I don't know what you mean with _all_
- all archs TC 6.2.4
- all archs any TC
- ...

it's just a partial fix for aarch64-6.2.4

th0ma7:
I get that, but why not pinning all dsm and all archs to that version until we can return to stable?

hgy59:
> I get that, but why not pinning all dsm and all archs to that version until we can return to stable?

until now we always use the latest rust.
If we pin all to 1.97.1 we wouldn't see new breaking changes on other than aarch64-6.2.4

th0ma7:
> > I get that, but why not pinning all dsm and all archs to that version until we can return to stable?
> 
> until now we always use the latest rust. If we pin all to 1.97.1 we wouldn't see new breaking changes on other than aarch64-6.2.4

I just thought that untils gcc85 overlay is ready we could pin all packages down, avoiding build differences where possible, I find this simpler.  But I'm good either way really.

th0ma7:
@hgy59 part of #7391 or through a subsequent PR, would it be relevant to simply create a rustc toolchain for all DSM-6.2.4 using the gcc85 overlay?  We could then updated it on a yearly basis or when moving to a newer gcc overlay or else?

hgy59:
> @hgy59 part of #7391 or through a subsequent PR, would it be relevant to simply create a rustc toolchain for all DSM-6.2.4 using the gcc85 overlay? We could then updated it on a yearly basis or when moving to a newer gcc overlay or else?

Do we need to create rust toolchains to benefit of the gcc 8.5 overlay?

I still hope that the aarch64-6.2.4 issue will be fixed upstream. There is a new comment in the upstream issue telling that the error occurs with Ubuntu 24.04 ARM64 github actions runner too.

th0ma7:
For some archs yes, gcc85 enable atomic for the ones that did not (or partially) supported it.

PR #7396, opened 2026-08-22
build logs: keep what the console showed

### LINKED ISSUE #7393: incomplete build log files
The build log files are lacking some output of the build

e.g. building a package that is not supported

console output:
```
make[1]: Entering directory '/spksrc/cross/helix-min'
===>  BUILDING package for arch x86-5.2 with SynoCommunity toolchain
===>  20260821-114051 MAKELEVEL: 01, PARALLEL_MAKE: max, ARCH: x86-5.2, NAME: helix [BEGIN]
make[2]: Entering directory '/spksrc/cross/helix-min'
../../mk/spksrc.rules/pre-check.mk:54: *** Arch 'x86-5.2' is not supported by helix: gcc 4.7.3 < 4.9.  Stop.
make[2]: Leaving directory '/spksrc/cross/helix-min'
===>  20260821-114051 MAKELEVEL: 01, PARALLEL_MAKE: max, ARCH: x86-5.2, NAME: helix [END]
make[1]: *** [../../mk/spksrc.rules/supported.mk:78: build-arch-x86-5.2] Error 1
make[1]: Leaving directory '/spksrc/cross/helix-min'
make: *** [../../mk/spksrc.rules/supported.mk:67: arch-x86-5.2] Error 2
```

build log file content:
```
===>  BUILDING package for arch x86-5.2 with SynoCommunity toolchain
```

some output goes to status-build.log
```
===>  20260821-114051 MAKELEVEL: 01, PARALLEL_MAKE: max, ARCH: x86-5.2, NAME: helix [BEGIN]
===>  20260821-114051 MAKELEVEL: 01, PARALLEL_MAKE: max, ARCH: x86-5.2, NAME: helix [END]
```
I would expect this output in the build log file too.

th0ma7:
Fixes #7393.

A package refused by a pre-check left a build log holding a single line. Everything the console showed — the reason, make's own `*** … Stop.`, the exit cascade — was lost.

## Cause

`$(error)` in `pre-check.mk` fires at **parse** time. The teeing lived inside a recipe (`LOG_WRAPPED`), and `build-arch-%` descended into the package make **without a pipe**, deliberately, because that recipe was expected to do the work. When no recipe ever runs, nothing is teed. The `[BEGIN]`/`[END]` markers, meanwhile, only ever reached `status-build.log`.

## Fix: tee one level up

Teeing moves to `arch-%`, *above* the make that fails, so that make's output **and** its exit cascade both pass through the pipe. Writing the message by hand from `precheck_fatal` covers the reason but nothing else — a parse-time `$(error)` produces no recipe to tee from.

```
before                                    after
BUILDING package for arch x86-5.2 …       BUILDING package for arch x86-5.2 …
                                          … NAME: python314 [BEGIN]
                                          ../../mk/spksrc.rules/pre-check.mk:65: ***
                                            Arch 'x86-5.2' is not supported by
                                            python314: gcc 4.7.3 < 4.8.  Stop.
                                          … NAME: python314 [END]
```

`precheck_fatal` replaces nine copies of the same `ifneq` / `$(shell …)` / `$(error)` block and sends one message to all three readers: the console, the package's build log, and the CI's collected unsupported list.

## One logging mechanism instead of two

`LOG_WRAPPED` and the teeing helper carried the same body — `script(1)` for a pty so **stderr** is captured too, `sed` to strip ANSI on the way to the file, a `LOGGING_ENABLED` guard so exactly one level tees and nothing is written twice.

They are now `_runlog` (the shell fragment; underscored because it needs `pipefail` and bash from its caller) and **`RUNLOG`** (the facade: builds the make command from a goal, writes to `DEFAULT_LOG`, reports failure to `STATUS_LOG`). `LOG_WRAPPED` is gone, nine call sites renamed, no behaviour change.

`TIME_CMD` resolves GNU `time(1)` by path: the piping recipes need bash for `pipefail`, and under bash `time` is a keyword that rejects `-o`, so `PSTAT_TIME` broke the moment those recipes switched shells.

## Toolchain status lines name their directory

An overlay consumer logged an empty arch and called itself `toolchain`, because the branch meant for it was guarded on `$(TC_NAME)$(TC_VERS)` — and a consumer sets `TC_VERS` but no `TC_NAME`.

```
ARCH: -6.2.4,                              NAME: toolchain     ← before
ARCH: qoriq-6.2.4_rust-1.82_gcc-4.9.3,     NAME: toolchain     ← after
ARCH: 88f6281-5.2_binutils-2.30,           NAME: toolchain
```

Neither `TC_NAME` nor `TC_ARCH` works alone: a consumer sets no `TC_NAME`, and on the generic toolchains `TC_ARCH` is the reference model, so `syno-aarch64-6.2.4` would have logged `rtd1296` and `syno-x64-7.1` `apollolake`. The directory name minus its `syno-` prefix is right in every case.

## Timestamp in the collected list

Those `$(shell)` calls contained `$(date --date=now +"%Y.%m.%d %H:%M:%S")` — *make* expanding an undefined variable, not the shell running `date`, so every line in `build_unsupported.txt` began with a bare `" - "`. Escaped to `$$(date …)`:

```
2026.09.04 01:38:11 - python314: Arch 'x86-5.2' is not supported by python314: gcc 4.7.3 < 4.8
```

## Verified in CI, deliberately

A `TEMP` commit switched on DSM 5.2, bumped both wheels packages, and put a bad compiler flag in `cross/xxhash` scoped to one arch. **Reverted as a commit, not by rewriting history,** so the run stays attached.

| | result |
|---|---|
| `comcerto2k-7.1` | **red on purpose** — the log carries gcc's `unrecognized command line option '-this-flag-does-not-exist'` and the `make[5]: *** [compile.mk:55: compile_target] Error 2` cascade; `build_errors.txt` names the four dependents |
| `ppc853x-5.2`, `88f6281-5.2`, `x86-5.2` | the pre-check refusals — the only archs where a refusal reaches a log at all |
| `python311-wheels`, `python314-wheels` | the crossenv/wheels path, which nests sub-makes and clears `MAKEFLAGS` of its own; paired markers on both |
| everything else | 18 jobs green, Lint and Documentation green |

## Also in this PR: `spk/borgbackup`

The DSM 5.2 jobs surfaced three arch lists in one file, each standing in for a capability floor, and each missing `x86-5.2` for the same reason:

```make
-ifneq ($(findstring $(ARCH),$(ARMv5_ARCHS) $(ARMv7L_ARCHS)),)      # → python311/314
+ifeq ($(call version_lt,$(TC_GCC),4.8),1)

-ifeq ($(findstring $(ARCH),$(ARMv5_ARCHS)),$(ARCH))                # → rpds-py/jsonschema
+ifeq ($(call version_ge,$(TC_RUSTC),1.85),1)

-UNSUPPORTED_ARCHS = $(OLD_PPC_ARCHS)                               # → borg's own C
+MIN_GCC_VERSION = 4.6
```

Each number is measured, not guessed: python314 declares `MIN_GCC_VERSION = 4.8`; rpds-py ≥ 0.25 pulls edition-2024 crates (`rpds` 1.1.1, `archery` 1.2.2, both `rust-version = 1.85.0`) and jsonschema ≥ 4.26 requires rpds-py ≥ 0.25; and borg's `setup.py` hardcodes `-Wno-unreachable-code-fallthrough`, which gcc only tolerates from 4.6 — confirmed on the toolchains in tree (4.3.2 refuses, 4.6.4 ignores) and in the manual, where that paragraph first appears in the 4.6 *Warning Options* page.

`hi3535` moves from python311 to python314 as a result, confirmed building there.

## Docs

A dated `changes.md` entry for the above, no Highlights. Plus three gaps found on the way, all changes that landed in `mk/` without the guide following:

- **`MIN_RUSTC_VERSION`**, from #7353, was only in the reference table — not in *Architecture Support*, not in the floors Highlight. Added to both, with the note it needs: `TC_RUSTC` is derived rather than declared, and an arch with no Rust overlay reports `stable`, which clears any floor.
- **The mirror table listed five families; the code has seven.** `MIRROR_SAVANNAH` (#7349), `MIRROR_GNUPG` (#7384), `MIRROR_FREEDESKTOP` (#7401) were missing, as was the fact that freedesktop is the one family that does not preserve the path. That entry becomes a four-PR entry, and records that X.org was removed in #7349 — its base 404s, and it matched only `www.x.org/releases/` while every X.org package downloads from `www.x.org/archive/individual/`.
- **`PKG_DOWNLOAD_METHOD`** was documented nowhere, so nothing said that `git`/`svn`/`hg` exist, which revision variable each takes, or that the whole mirror mechanism applies to `http` alone.

## Known limit

An error can only be captured one level above the make that fails, so the outermost `make: *** [arch-x64-7.1] Error 2` is structurally unreachable — there is no outer make to tee it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -281,10 +281,14 @@ jobs:
           add_dsm70_builds=${{ github.event.inputs.add_dsm70_builds || 'false' }}
           add_dsm62_builds=${{ github.event.inputs.add_dsm62_builds || 'false' }}
           add_dsm61_builds=${{ github.event.inputs.add_dsm61_builds || 'false' }}
-          add_dsm52_builds=${{ github.event.inputs.add_dsm52_builds || 'false' }}
+          # TEMP (revert before merge): 5.2 carries ppc853x, 88f6281 and x86, where the
+          # pre-check refusals and the old-toolchain wheels path are exercised.
+          add_dsm52_builds=${{ github.event.inputs.add_dsm52_builds || 'true' }}
           add_srm13_builds=${{ github.event.inputs.add_srm13_builds || 'false' }}
           add_srm12_builds=${{ github.event.inputs.add_srm12_builds || 'false' }}
-          add_noarch_builds=${{ github.event.inputs.add_noarch_builds || 'false' }}
+          # TEMP (revert before merge): exercises arch-noarch-% / build-noarch-%, which
+          # this PR rewires the same way as the arch path.
+          add_noarch_builds=${{ github.event.inputs.add_noarch_builds || 'true' }}
           add_noarch_dsm72_builds=${{ github.event.inputs.add_noarch_dsm72_builds || 'false' }}
           add_noarch_dsm73_builds=${{ github.event.inputs.add_noarch_dsm73_builds || 'false' }}
           has_arch_packages=${{ needs.prepare.outputs.has_arch_packages || 'false' }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -281,14 +281,10 @@ jobs:
           add_dsm70_builds=${{ github.event.inputs.add_dsm70_builds || 'false' }}
           add_dsm62_builds=${{ github.event.inputs.add_dsm62_builds || 'false' }}
           add_dsm61_builds=${{ github.event.inputs.add_dsm61_builds || 'false' }}
-          # TEMP (revert before merge): 5.2 carries ppc853x, 88f6281 and x86, where the
-          # pre-check refusals and the old-toolchain wheels path are exercised.
-          add_dsm52_builds=${{ github.event.inputs.add_dsm52_builds || 'true' }}
+          add_dsm52_builds=${{ github.event.inputs.add_dsm52_builds || 'false' }}
           add_srm13_builds=${{ github.event.inputs.add_srm13_builds || 'false' }}
           add_srm12_builds=${{ github.event.inputs.add_srm12_builds || 'false' }}
-          # TEMP (revert before merge): exercises arch-noarch-% / build-noarch-%, which
-          # this PR rewires the same way as the arch path.
-          add_noarch_builds=${{ github.event.inputs.add_noarch_builds || 'true' }}
+          add_noarch_builds=${{ github.event.inputs.add_noarch_builds || 'false' }}
           add_noarch_dsm72_builds=${{ github.event.inputs.add_noarch_dsm72_builds || 'false' }}
           add_noarch_dsm73_builds=${{ github.event.inputs.add_noarch_dsm73_builds || 'false' }}
           has_arch_packages=${{ needs.prepare.outputs.has_arch_packages || 'false' }}
--- a/cross/xxhash/Makefile
+++ b/cross/xxhash/Makefile
@@ -14,4 +14,11 @@ LICENSE  = 2-Clause BSD
 
 CONFIGURE_TARGET = nop
 
+# TEMP (revert before merge): force a mid-build failure on one arch to prove the CI
+# captures it -- the log should carry the compiler's error, build_errors.txt the package,
+# and the job should go red. Scoped to comcerto2k so every other arch reports normally.
+ifeq ($(ARCH),comcerto2k)
+ADDITIONAL_CFLAGS += -this-flag-does-not-exist
+endif
+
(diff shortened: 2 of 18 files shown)


th0ma7:
@hgy59 this will solve the issue you encoutered.  It needed a bit more involvment than anticipated.  Adding @mreid-tt let me know if you see anything going wrong.  I did all the testing I could but `"` vs `'` sometimes brings a few surprises...

PR #7400, opened 2026-08-23
download: mirror the freedesktop.org family from the framework

th0ma7:
freedesktop.org is down — `dri.`, `www.`, `xorg.` and `pkgconfig.` all time out, and CI stalls on the downloads. Only `gitlab.freedesktop.org` answers, which is why `intel-gpu-tools` and `libpciaccess` are unaffected.

Six packages come from that host, so the fix goes in the mirror table rather than in six Makefiles.

## Why a new shape of entry

The existing family mechanism swaps everything up to a **tree-root marker**:

```
*//download.gnome.org/sources/*)   marker="/sources/"   mirrorBases="$(MIRROR_GNOME)"
```

That does not fit freedesktop: each project sits under its own path shape — `dri./libdrm/`, `www./software/fontconfig/release/`, `xorg./archive/individual/lib/`, `pkgconfig./releases/` — so there is no common marker to strip.

Its mirrors, though, file everything **by package name**. So `MIRROR_FREEDESKTOP` carries `$(PKG_NAME)` in the base and has the file name appended, exactly the way `PKG_DIST_MIRRORS` already works. One extra loop beside the existing one:

```make
MIRROR_FREEDESKTOP ?= https://ftp.osuosl.org/pub/blfs/conglomeration/$(PKG_NAME) \
                      https://distfiles.macports.org/$(PKG_NAME)
```

## Coverage

| | |
|---|---|
| `dbus` `fontconfig` `libdrm` `uchardet` | covered by the table alone |
| `libX11` `pkg-config` | mirrored under a different name, so one `PKG_DIST_MIRRORS` line each |

macports files things under its **port** name — `xorg-libX11`, `pkgconfig` — not the package name. A base that lacks the file simply 404s and the next candidate is tried, which is what keeps those two exceptions to a single line instead of a special case. Any freedesktop package added later gets the mirror for nothing.

## Verified against the live outage

Not probed for a 200 — actually downloaded, with the primary genuinely unreachable:

```
libdrm    dri.freedesktop.org        failed
          osuosl/libdrm              ok, sha256 matches digests

libX11    xorg.freedesktop.org       failed
          osuosl/libX11              failed
          macports/libX11            failed
          macports/xorg-libX11       ok, sha256 matches digests
```

The second one walks the whole cascade, which is the case worth seeing work.

No digest changes; the primary stays first, so nothing changes once upstream recovers.

## dbus

`cross/dbus` is covered by the table here, but it is stuck on **1.13.22** — a development release that no mirror carries. #7401 moves it to 1.14.10, at which point this entry starts working for it too.

### DIFF
--- a/cross/fontconfig/Makefile
+++ b/cross/fontconfig/Makefile
@@ -3,6 +3,7 @@ PKG_VERS = 2.16.0
 PKG_EXT = tar.xz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://www.freedesktop.org/software/fontconfig/release
+PKG_DIST_MIRRORS = https://ftp.osuosl.org/pub/blfs/conglomeration/fontconfig
 PKG_DIR = $(PKG_NAME)-$(PKG_VERS)
 
 DEPENDS = cross/libexpat cross/freetype cross/libuuid
--- a/cross/libX11/Makefile
+++ b/cross/libX11/Makefile
@@ -3,6 +3,8 @@ PKG_VERS = 1.8.10
 PKG_EXT = tar.xz
 PKG_DIST_NAME = $(PKG_NAME)-$(PKG_VERS).$(PKG_EXT)
 PKG_DIST_SITE = https://xorg.freedesktop.org/archive/individual/lib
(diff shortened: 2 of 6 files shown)


th0ma7:
Closing, superseeded with https://github.com/SynoCommunity/spksrc/pull/7401

PR #7399, opened 2026-08-23
curl 8.21.0 and its dependencies: update, and move curl to cmake

th0ma7:
Updates curl and everything it depends on, and moves curl from autotools to cmake. Split out of #7397 so that PR can rebase onto this one with fewer packages left to deal with.

| package | from | to |
|---|---|---|
| curl | 8.4.0 | 8.21.0 |
| mbedtls | 3.6.3 | 4.2.0 |
| nghttp2 | 1.58.0 | 1.70.0 |
| nghttp3 | 1.0.0 | 1.18.0 |
| ngtcp2 | 1.0.1 | 1.25.0 |
| gsasl | 2.2.0 | 2.2.4 |
| libidn2 | 2.3.4 | 2.3.8 |
| libpsl | 0.21.2 | 0.23.3 |
| openldap | 2.6.4 | 2.6.9 |

## curl: autotools → cmake

Not an option-for-option translation. cmake and autotools have **opposite defaults** on several features:

```
USE_NGHTTP2  USE_LIBIDN2  CURL_USE_LIBPSL  CURL_USE_LIBSSH2   ON  by default
CURL_DISABLE_LDAP  CURL_DISABLE_WEBSOCKETS                    OFF by default
CURL_ZLIB  CURL_ZSTD  CURL_BROTLI                             AUTO
```

A literal translation would have silently enabled six features the autotools build did not have, and `AUTO` would pick up whatever another package left in the staging tree. Each is therefore stated in both directions, never left to the default.

`--with-librtmp` disappears. That is not a cmake gap — curl 8.21 dropped librtmp entirely, autotools included (`grep -c with-librtmp configure` → 0), so the option was already being ignored.

`curl-config` is still generated and installed by the cmake build, so the existing post-install fixup is unchanged.

## What the new versions needed on old toolchains

**curl — eventfd.** glibc gained `EFD_CLOEXEC`/`EFD_NONBLOCK` in 2.9 with the `eventfd2` syscall. glibc 2.8 has the function *and* the header but not the flags, and both build systems only probe the function, so `socketpair.c` failed on ppc853x-5.2:

```
socketpair.c:40: error: 'EFD_CLOEXEC' undeclared
```

Defining the constants ourselves would compile and then fail at runtime — 2.8's wrapper calls the flagless `eventfd` syscall. Instead `HAVE_EVENTFD=OFF` below glibc 2.9 skips the probe, and curl uses the pipe fallback it already implements.

[...]

Issue #240, opened 2012-09-11
Request: Add "Plex Media Server" for ARM units


aniceberg:
Today Plex announced a version of Plex Media Server for ARM-based Synology NAS units.

Package is here: http://cdn.plexapp.com/PlexMediaServer/0.9.6.9.240/PlexMediaServer-0.9.6.9.240-8fd9c6a-arm.spk

PR #7404, opened 2026-08-26
uMurmur: Harden certificate generation and config migration

mreid-tt:
## Description

Follow-on to the unpublished [#7399](https://github.com/SynoCommunity/spksrc/pull/7399) (which updated uMurmur to v0.4.1 and pinned mbedtls at 3.6.7 for compatibility). Fixes issues in the uMurmur package's `service-setup.sh`: the certificate-generation fallback was dead code (an in-function `exit` short-circuited `service_postinst`), empty certificate files could be accepted as valid on a retried install, and the config migration used a fragile `if $(grep -q ...)` idiom. No `SPK_REV` bump — the package (0.4.1) was never published, so this fixes it before release.

## Changes

- **Fix the certificate fallback** — `create_certificate()` and `service_postinst()` used `exit` inside functions, which terminated the rest of `service_postinst` (the `touch` fallback never ran). Converted to `return` so the intended control flow works.
- **Harden certificate checks** — `[ -f ]` → `[ -s ]` (non-empty) so empty/placeholder key and certificate files are not accepted as valid; dropped the `touch ${PRIVATE_KEY}/${PUBLIC_KEY}` fallback so a failed install fails cleanly without leaving misleading empty files behind (which would have passed the old `-f` check on a retry).
- **Config migration cleanup** — replaced `if $(grep -q ...)` with `if grep -q ...` (the command-substitution form would break if the pattern output anything).

Fixes # 

## Checklist

### Type of change

- [x] Bug fix

### DIFF
--- a/spk/umurmur/src/service-setup.sh
+++ b/spk/umurmur/src/service-setup.sh
@@ -10,7 +10,7 @@ PUBLIC_KEY="${SYNOPKG_PKGVAR}/umurmur.crt"
 
 create_certificate ()
 {
-    if [ -f "${PRIVATE_KEY}" ] && [ -f "${PUBLIC_KEY}" ]; then
+    if [ -s "${PRIVATE_KEY}" ] && [ -s "${PUBLIC_KEY}" ]; then
         echo "Found uMurmur certificate. To create a new certificate upon package update you have to delete the existing certificate before."
         return 0
     fi
@@ -37,8 +37,6 @@ service_postinst ()
     # Certificate generation
     create_certificate 2>&1
     if [ $? -ne 0 ]; then
-        touch ${PRIVATE_KEY}
-        touch ${PUBLIC_KEY}
         return 1
     fi
 }
@@ -47,7 +45,7 @@ service_preupgrade ()
 {
     # Migrate to DSM 7 compatible var folder
     if [ -e "${CFG_FILE}" ]; then
-        if $(grep -q "/usr/local/umurmur/var/" "${CFG_FILE}"); then
+        if grep -q "/usr/local/umurmur/var/" "${CFG_FILE}"; then
             echo "Update var folder for DSM 7 compatibility in configuration file."
             sed -e "s,/usr/local/umurmur/var/,/var/packages/umurmur/var/,g" -i "${CFG_FILE}"
         fi
@@ -55,7 +53,7 @@ service_preupgrade ()
 
     # Update log-file name to package name
     if [ -e "${CFG_FILE}" ]; then
-        if $(grep -q "umurmurd.log" "${CFG_FILE}"); then
+        if grep -q "umurmurd.log" "${CFG_FILE}"; then
             echo "Update log file name in configuration file."
             sed -e "s,umurmurd.log,umurmur.log,g" -i "${CFG_FILE}"
         fi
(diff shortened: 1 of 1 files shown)


mreid-tt:
Hi @th0ma7, I noticed your package update to uMurmur in #7399 wasn’t published. I’ve also seen some other improvements in the `service-setup.sh` which I’m including here.  Is there anything else you’d like to discuss before it’s published?

th0ma7:
Mostly disconnected for the week which is why i haven't completed in full umurmur update.

Note that i wanted also to migrate mbedtls to version 4.2 but umurmur doesn't support it just yet.  Although they have a patch now available for testing.  I told them I'd take a look upon my return as this could be easily added to our package if working properly.

Feel free to try it out if you jave available cycles for it.

mreid-tt:
Thanks @th0ma7 for the context. I'm still learning the ropes here, so this is mostly a question to check my understanding.

I took a look at the upstream [Draft PR #240](https://github.com/umurmur/umurmur/pull/240) ("Update to Mbed TLS 4.x and PSA Crypto API") to try to understand the mbedtls 4.x situation. As far as I can tell, it rewrites the Mumble cert-fingerprint code to use the PSA Crypto API (`psa_hash_compute(PSA_ALG_SHA_1, ...)`), which seems to keep SHA-1 working even though mbedtls 4.x removed the old `mbedtls_sha1()` — and it also deals with the moved headers and the HAVEGE removal. If I'm reading that correctly, it sounds like it could be a way forward.

Since it's still a Draft and doesn't appear to be part of the upcoming 0.5.0 release, I wasn't sure whether it would make sense to try it in our package yet, or whether it's better to wait until umurmur ships mbedtls 4.x support in a release and then update. I'd be happy to help test it if that's useful — just let me know what would help most.

In the meantime, this PR just tidies up the install-time certificate handling and config migration (no `SPK_REV` bump, since 0.4.1 was never published). Happy to make any changes you'd suggest.

th0ma7:
If you can test and provide feedback i believe they would appreciate.  In the meantime maybe we could keep a draft pr until latest mbedtls is officially supported?

mreid-tt:
Thanks @th0ma7 — that makes sense. To align: I'll go ahead and **merge this PR, but not publish it**, so the service-setup hardening is in place for when the package is eventually released. Then I'll open a **new draft PR** to test the upstream mbedtls 4.x patch ([#240](https://github.com/umurmur/umurmur/pull/240)) and can share feedback with the umurmur maintainers once I've run it through our cross-compile setup.

That way the publishing decision stays tied to mbedtls 4.x support, as you suggested.

th0ma7:
@mreid-tt i would rather suggest to publish it right away as last update is more than a year old.

mreid-tt:
> @mreid-tt i would rather suggest to publish it right away as last update is more than a year old.

Hey, how does one test this app? When I install it the open button/shortcut points to my NAS with the address `/webman/3rdparty/umurmur/umurmur.cgi`. But this doesn't show anything. Should the web shortcut be disabled as there is no webserver?

th0ma7:
No clue, sorry buy can't help... I can have a look at it upon my return though.

mreid-tt:
> No clue, sorry buy can't help... I can have a look at it upon my return though.

Dug a bit deeper to understand how it should work and installed a Mumble client to test connectivity. Found that the package needed some modernising which I've done in #7414 and will publish from there.

PR #7406, opened 2026-08-26
Netdata: Update to v2.11.0

mreid-tt:
## Description

Updates Netdata from v2.10.4 to v2.11.0, reworking the cross-package Rust/protobuf integration for the new release and enabling the go.d collector suite plus the network viewer. v2.11.0 brings network monitoring tech previews (NetFlow/IPFIX/sFlow, network dashboard), OpenTelemetry log ingestion, SNMP trap listening, and an AWS CloudWatch collector.

## Upstream Changes

- See: https://github.com/netdata/netdata/releases/tag/v2.11.0

## Changes

- **Update to v2.11.0** — `PKG_VERS`/`SPK_VERS` to 2.11.0, regenerated digests, `SPK_REV` to 3, changelog entry.
- **Rework the journal-reader CMake patch** — v2.11.0's Corrosion block now also guards `ENABLE_PLUGIN_NETFLOW` (new `netflow-plugin` crate); both OTEL and NETFLOW are `OFF`, so the prebuilt-static-lib replacement stays safe.
- **Adapt protobuf generation** — v2.11.0 moved the aclk-schemas protos under a `proto/` subdirectory, so netdata's cmake emitted `--cpp_out` into a path that mismatched the ninja build rules; normalized it to the build tree's `aclk-schemas` dir.
- **Use `env-rust.mk` for the Rust target** — replaced the hand-maintained `RUST_TARGET` arch map with the framework's `env-rust.mk` (the generated `tc_vars.mk` does not provide these variables). Includes the `journal_reader_ffi` static-lib build, unchanged from v2.10.4.
- **Drop dead PPC Rust mapping** — netdata is DSM 7.0+ (no PPC 7.x toolchain) and `MIN_GCC_VERSION = 7.5` already excludes PPC, so the `$(PPC_ARCHS)` Rust block referenced a removed toolchain.
- **Enable the go.d plugin and network viewer** — `ENABLE_PLUGIN_GO=ON` (via `env-go.mk` + `native/go` build dependency) ships the go.d collector suite with service discovery, and `ENABLE_PLUGIN_NETWORK_VIEWER=ON` adds the network connections view. (Other plugins remain `OFF`: ML needs libmongoc + snappy, Prometheus remote write needs snappy, and the rest need external libs or kernel/toolchain support unavailable on DSM.)
- **Fix the SNMP trap profile pack for cross-compilation** — the build-time `snmp-trap-profile-gen` helper is cross-compiled and must run on the build host; the CMake pack step now builds it for the host so non-amd64 arches (aarch64) configure cleanly.
- **Extend `netdata-fix` to all elevated-privilege plugins** — grants root (setuid) to `apps.plugin`, `systemd-journal.plugin`, `local-listeners`, `network-viewer.plugin`, `ndsudo` and `go.d.plugin`. `local-listeners` uses `4755` because the go.d service-discovery discoverer spawns it unprivileged (as `nobody`) and relies on its setuid bit.
- **Drop the hard DSM version floor** — `REQUIRED_MIN_DSM` removed; `MIN_GCC_VERSION = 7.5` (C++17 requirement) is the sole gate, with a self-contained explanatory comment.
- **Trim the package** — the build-time `snmp-trap-profile-gen` is no longer shipped (saves ~4 MB); wizard and docs updated to describe the enabled features (per-process monitoring, system logs, network connections, service discovery) and note DSM-specific service discovery limitations (DSM nginx lacks `stub_status`, DSM PostgreSQL needs credentials).

Fixes # 

## Checklist

### Type of change

- [x] Package update


spk/netdata/Makefile
hgy59: is this still a requirement or will netdata be supported on DSM 6 when ggc 8.5 (overlay) will be available?
If min DSM 7 is still a requirement then an explanatory comment would be useful.

mreid-tt: Agreed, this has been removed.

PR #7190, opened 2026-06-09
synocli-net: update tools and libraries

hgy59:
## Description

Tools:
- update bind v9.18.38 to v9.18.48 and v9.20.11 to v9.20.22
- update inetutils from v2.6 to v2.8
- update nmap from v7.97 to v7.98
- update rsync from v3.4.1 to v3.4.4
- update ser2net from v4.6.5 to v4.6.6 and include gensio tools
- update socat from v1.8.0.3 to v1.8.1.1
- update sshfs3 from v3.7.3 to v3.7.6 (except for ARMv5 archs)
- update tmux from v3.5a to v3.6b

Libraries:
- update fuse3 from v3.16.2 to v3.18.2 (except for ARMv5 and ARMv7L archs)
- update openssh from v9.6p1 to v10.2p1
- update gensio from v2.8.15 to v3.0.1

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [x] Package update

PR #7416, opened 2026-08-29
synocli-net: ship sshd-session and sshd-auth

bfreis:
## Description

`sbin/sshd` is shipped by synocli-net but cannot start. It exits immediately
with:

```
/var/packages/synocli-net/target/libexec/sshd-session does not exist or is not executable
```

OpenSSH 9.8 split the server into a listener (`sshd`) and a per-connection
worker (`sshd-session`); 9.9 split `sshd-auth` out of that again. `sshd` will
not start without them.

`cross/openssh` already builds and installs both — `install-nosysconf` runs
upstream's `install-files`, which puts `sshd-session`, `sshd-auth`,
`sftp-server`, `ssh-keysign`, `ssh-pkcs11-helper` and `ssh-sk-helper` into
`libexec`. But `cross/openssh/PLIST` lists only four of those six, and the
PLIST is what decides package contents: `mk/spksrc.spk/copy.mk` concatenates
each dependency's `cat_PLIST` into `work-*/PLIST`, then `copy_target` tars
exactly those paths out of `install/var/packages/synocli-net/target/` into
`staging/`. The two server helpers are installed and then dropped at that step.

(`SPK_COMMANDS` in `spk/synocli-net/Makefile` lists the same four, but that
variable only drives the `usr-local-linker` block in `conf/resource` — it
creates `/usr/local` symlinks and copies nothing.)

This has been the case since #7190 bumped `cross/openssh` from 9.6p1 to
10.3p1. 9.6p1 predates the split, so `sbin/sshd` was self-contained and worked;
from 10.3p1 it has been shipped and unable to run.

It is easy to miss: the package is a client-tools bundle, and an
already-running `sshd` keeps working until its host next reboots. On my NAS the
break arrived with a package upgrade in July and only surfaced a month later,
when a DSM update rebooted the box.

## Changes

- `cross/openssh/PLIST`
  - add `bin:libexec/sshd-auth` and `bin:libexec/sshd-session`
- `spk/synocli-net/Makefile`
  - move all six `libexec/` entries from `SPK_COMMANDS` to
    `SPK_USR_LOCAL_LINKS`, as `bin:libexec/...`
  - `SPK_REV`: 22 → 23
  - `CHANGELOG`: replaced with this revision's entry

Both helpers are needed. Adding only `sshd-session` gets one step further and
then fails on `sshd-auth`.

The `SPK_USR_LOCAL_LINKS` move is @hgy59's, and fixes a pre-existing problem
for the other four helpers as well: on DSM >= 6 every `SPK_COMMANDS` entry goes
under the `"bin"` key of `usr-local-linker` regardless of its subdirectory, but
on DSM < 6 and SRM the links come from the shell fallback in
`mk/spksrc.service/create_links`, which does `ln -s .../${cmd} /usr/local/${cmd}`
verbatim — so `libexec/sftp-server` landed at `/usr/local/libexec/sftp-server`,
off PATH. `bin:libexec/...` gives the same result on DSM >= 6 and the correct
one on the older path, matching `spk/borgbackup`, `spk/salt-master` and the
`python3xx` packages.

No sources are added, so digests are unchanged, and the privilege separation
directory (`--with-privsep-path=$(INSTALL_PREFIX)/var/empty`) is already
present in the installed package.

## Testing

- **Hardware:** Synology DS923+ (`r1000`, x86_64)
- **DSM:** 7.4.1-90080
- **Package:** synocli-net 10.0-22 installed; toolchain `syno-r1000-7.2`

Verified:

1. Built `cross/openssh` for `r1000-7.2` and confirmed staging contains all six
   `libexec` binaries, including the two absent from the package.
2. Installed `sshd-session` and `sshd-auth` into
   `/var/packages/synocli-net/target/libexec/` and started `sbin/sshd`.
3. `sshd` starts, listens, and accepts connections and authenticates normally —
   where before it exited immediately with the error above.
4. Key exchange is post-quantum, which is the reason to run this `sshd` rather
   than DSM's own (OpenSSH 8.2p1):

   ```
   debug1: kex: algorithm: mlkem768x25519-sha256     # synocli-net sshd
   debug1: kex: algorithm: curve25519-sha256         # DSM's sshd, which also warns
   ** WARNING: connection is not using a post-quantum key exchange algorithm.
   ```

Note for reviewers: a standalone `cross/openssh` build compiles paths for
`/usr/local/openssh` rather than the package prefix, so the test above also
needed `/usr/local/openssh/libexec/` and `/usr/local/openssh/var/empty` to
exist. That is a consequence of testing with a `cross/` build; a package built
through `spk/synocli-net` compiles all three binaries with the package prefix
and needs neither.

Correction to the original submission: the verification above was done by
hand-placing the two binaries on the NAS, not by installing a package built
from `spk/synocli-net`. I blamed `SPK_COMMANDS` for the missing files, but the
actual gate is the PLIST, so the first revision of this PR would not have fixed
anything on a real install. Thanks to @hgy59 for catching that.

## Related Issues

None found — I did not locate an existing issue for this. Happy to open one if
you would prefer it tracked separately.

## Checklist

- [x] Tested on real hardware
- [x] Updated changelog
- [x] Incremented SPK_REV
- [ ] Added/updated documentation — no documentation change appears needed;
      the package docs do not mention running `sshd`


hgy59:
@bfreis Thank you for discovering and reporting this.

Additionally it needs to update PLIST to include the new files in the package and a fix to update command links to support SRM and DSM < 6.

I wanted to add the changes but was not allowed to push to your repository.
So find the changes in the attached patch file (include both changes and the commit comment).

[0001-synocli-net-add-new-binaries.patch](https://github.com/user-attachments/files/31590141/0001-synocli-net-add-new-binaries.patch)

bfreis:
@hgy59 - sure thing, and thanks for taking a look and for sharing the details! Will update it shortly.

Cheers

hgy59:
@bfreis if you confirm that this works now as excpected we can merge and publish.

bfreis:
Hi @hgy59 , I just tested it and it's working as expected! I have my PQ SSH back :)

Cheers!

PR #7438, opened 2026-09-05
ci: classify MIN_GLIBC_VERSION / MIN_GCC_VERSION packages as restricted-DSM builds

mreid-tt:
## Description

Fix the CI so packages that gate on `MIN_GLIBC_VERSION` / `MIN_GCC_VERSION` (instead of `REQUIRED_MIN_DSM`) are still built for the DSM they require.

The restricted-build classification in `.github/actions/prepare.sh` only recognized `REQUIRED_MIN_DSM`, so packages using the capability variables (e.g. `kavita`, `dotnet10-runtime`, which set `MIN_GLIBC_VERSION = 2.27`) were never classified as requiring DSM 7.2 — the DSM 7.2 matrix was never enabled for them, and they effectively built for no DSM at all.

Changes:
- Map a `MIN_GLIBC_VERSION` floor above 2.26 (DSM 7.1's glibc) to min DSM 7.2 — fixes kavita and dotnet10-runtime.
- Map a `MIN_GCC_VERSION` floor above 8.5.0 (DSM 7.1's highest GCC) to min DSM 7.2 — future-proofing; no current package needs it.
- `MIN_RUSTC_VERSION` is intentionally not mapped: it is provided by the rust overlay on custom-rust archs and is not tied to a DSM version.

Fixes # 

## Checklist

### Type of change

### DIFF
--- a/.github/actions/prepare.sh
+++ b/.github/actions/prepare.sh
@@ -136,8 +136,26 @@ min_dsm_for_glibc() {
 }
 
 # ---------------------------------------------------------------------------
+# Map a MIN_GCC_VERSION floor to the minimum DSM version whose toolchains
+# ship at least that gcc. DSM 7.1 toolchains top out at gcc 8.5.0; DSM 7.2+
+# ship gcc 12.2, so any floor above 8.5.0 is classified as min DSM 7.2
+# (7.2 and 7.3 share gcc 12.2, so these land in the 7.2 bucket).
+#
+# Usage: min_dsm_for_gcc <gcc version>
+# Prints the minimum DSM version, or nothing if the floor is below the
+# default builds (7.1 / 6.2.4).
+# ---------------------------------------------------------------------------
+min_dsm_for_gcc() {
+    local gcc="$1"
+    if [ -n "${gcc}" ] && [ "${gcc}" != "8.5.0" ] && \
+       [ "$(printf '%s\n%s\n' "8.5.0" "${gcc}" | sort -V | tail -1)" = "${gcc}" ]; then
+        echo "7.2"
+    fi
+}
+
+# ---------------------------------------------------------------------------
 # Collect packages that declare REQUIRED_MIN_DSM = <version> or a
-# MIN_GLIBC_VERSION floor that implies <version>,
+# MIN_GLIBC_VERSION / MIN_GCC_VERSION floor that implies <version>,
 # preserving the build order already established in $packages.
 # Injects required meta-packages by resolving all qualifying packages
 # in a single inject_meta_packages call to avoid cross-iteration state issues.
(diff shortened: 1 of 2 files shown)


th0ma7:
I belive I do understand the intent of the automation in this PR but I'm unsure it is relevant... here's why (and please let me know if I didn't capture properly your intent).

Intent of `MIN_*_VERSION` is to set a gating floor for building based on actual toolchain capabilities and base-level package & dependencies mandatory requirements.  Simply said, it's meant to tell what this package or dependency really needs in terms of minimal gcc, glibc or rustc version, and not what we rather intent to enforce limiting builds to. 
 With the overlay capabilities I'm working on, both GCC and RUSTC versions will be increased, thus unlocking packages for older DSM automatically.  Only GLIBC will remain outstanding.

So, by extent, using `MIN_GCC_VERSION = 8.5` matching DSM-7.1 and above will eventuall unlock DSM-5.2 + DSM-6.2.4 + DSM-7.0 once the gcc-8.5 overlay is ready https://github.com/SynoCommunity/spksrc/pull/7391.  I actually hope later on to further update the gcc version up to 12.2 matching DSM-7.2/7.3.

Likewise, using `MIN_GLIBC_VERSION` must only be done where applicable.  In reality, as soon as we reach DSM-6.2.4 with glibc-2.20 it unlocks almost everything.  I foresee only a handfull of exceptions truly requiring something above that (i.e. glibc-2.26 or beyoond with DSM-7.x) - although could presumably grow over time.  So in theory only this isn't a moving target which could serve a dual purpose of enforcing build for higher DSM versions, unless we decide to package our own minimal libC eventually.  But still, at the moment (and I may be wrong) I do not forsee many (if any) package needing anything newer than glibc-2.26 such as with DSM-7.1.

Long-story-short:
- `MIN_*_VERSION` is meant to specify package & dependencies minimal requirements.  Overlay's will allow going above those requriements for older archs (ex.: building `ffmpeg8` on DSM-5.2).  And the CI build process makes it so that archs not meeting these expected capability levels are skipped from the build altogether.
- `REQUIRED_MIN_DSM` primary purpose is to set a minimum DSM version floor we intent to support for a package. Admitedly, due to previous lack of `MIN_*_VERSION` it also served that dual purpose until now.  This second purpose will no longer hold true as we migrate dependencies and packages over to the new facility.

So, indeed, over time with the overlay enablement we might want to remove `REQUIRED_MIN_DSM` where it is no longer necessary.  Although there might still be exceptions where we absolutely want or need DSM-7.4 where this functionality will still allow unblocking the arches on the CI.  Only exception I foresee where this could apply is for packages needing glibc-2.36 minimally, where it could enforce building for DSM-7.2 ... but I don't see that being common and such exception could still be filled with `REQUIRED_MIN_DSM`.

More than welcomed to further exchange on this if you think I've missed your intent.

mreid-tt:
**hey @th0ma7, thanks — that fully clarifies the design intent, and I agree.**

`MIN_*_VERSION` is a build-capability floor: what a package genuinely needs from the toolchain, independent of DSM, and the per-arch capability check already skips archs that can't meet it. `REQUIRED_MIN_DSM` is the DSM-support floor — and once the overlays land, the gcc/rustc gates will build on older DSM on their own, so that dual purpose falls away and `REQUIRED_MIN_DSM` reverts to its intended role.

And that maps exactly to the exception you described: only a glibc runtime floor genuinely forces a higher DSM, and the right answer for those rare cases is `REQUIRED_MIN_DSM` (e.g. a glibc-2.36 requirement enforcing DSM 7.2). Kavita is that case — `MIN_GLIBC_VERSION = 2.27`, which DSM 7.2's glibc 2.36 satisfies and 7.1's 2.26 does not.

So, agreed — I'll withdraw this PR and set `REQUIRED_MIN_DSM = 7.2` on the package side (kavita, and I'll check dotnet10-runtime), keeping `MIN_GLIBC_VERSION` as the precise toolchain floor.

PR #7439, opened 2026-09-07
pre-check: name every gate that blocks an arch, and add make check-<arch>-<tcvers>

th0ma7:
## The problem

The pre-check stopped at the first floor a package could not meet, and only ever looked at the package's own. Removing that blocker revealed the next one — a build at a time. And `UNSUPPORTED_ARCHS` said **where** a package fails, never why.

## `make check-<arch>-<tcvers>`

Ask an architecture what stands in the way, without starting a build:

```
$ make check-x86-5.2                        # in spk/tvheadend
===>  tvheadend: x86-5.2 check: 17 failed, 14 more behind an optional dependency
       required
         cross/ffmpeg8              gcc 4.7.3 < 4.9
         cross/python314            gcc 4.7.3 < 4.8
         ...
       optional
         cross/frei0r               gcc 4.7.3 < 7.5
         ...

$ make check-x86-5.2                        # in spk/borgbackup
===>  borgbackup: x86-5.2 check: OK
```

`make ARCH=x86 TCVERSION=5.2 check` is the same with the pair in variables. **required** is what the build needs; **optional** is what an `OPTIONAL_DEPENDS` branch would demand if turned on — never a reason to refuse.

A package reports every refusal the pre-check would make: a capability floor, an `UNSUPPORTED_ARCHS` exclusion with its reason, or the DSM/SRM window.

```
cross/minio      unsupported arch (go has no 32-bit PowerPC target)
spk/jellyfin     unsupported arch, DSM 5.2 < 7.2
```

## The pre-check runs the same walk

So a refused build names every blocker at once instead of the first:

```
===>  check: cross/x264 gcc 4.3.7 < 4.6
===>  check: cross/python314 gcc 4.3.7 < 4.8
===>  check: cross/ffmpeg8 gcc 4.3.7 < 4.9
...
pre-check.mk:80: *** Arch 'ppc853x-5.2' is not supported by tvheadend (18 failed check(s) in the tree).  Stop.
```

It reuses `dependency-flat`'s walk, already stamped so a diamond is visited once, already forwarding ARCH/TCVERSION so each package evaluates its own conditional `DEPENDS`, already setting `DEPENDENCY_WALK=1` so a refused package reports instead of aborting. Required and optional are a **set difference** between two walks, not a label carried down — a package under both a required and an optional parent is required.

## Consequences

- **`UNSUPPORTED_ARCHS_REASON`** is carried into the refusal; `env-go.mk` and `env-dotnet.mk` say theirs.
- **Seven restated floors removed.** `spk/tvheadend`, `chromaprint`, `comskip` and `spk/ffmpeg5-8` each declared `MIN_GCC_VERSION = 4.9`, restating their own `cross/` package. Declare the floor on the `cross/`; the `spk/` inherits it by being walked. The python side is left for its own PR.
- **`comma_append`** replaces two names for one operation, and `spksrc.common.mk`'s utility block moves above the includes — `$(,)` expanded to nothing in every file that included it, which is why `tc-capability.mk` carried a private copy.

## Cost and risk

One walk per parse: **2.8s warm** on tvheadend's 141-package tree, against builds measured in tens of minutes. **Verdict-neutral** — every `spk/` package across x64-7.1, 88f6281-6.2.4 and ppc853x-5.2 gives the same refused/allowed set as before. What changes is that the whole story is told at once.

DSM 5.2 is enabled in a separate commit so CI exercises the toolchains below gcc 4.9. **Revert before merge.**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -281,7 +281,9 @@ jobs:
           add_dsm70_builds=${{ github.event.inputs.add_dsm70_builds || 'false' }}
           add_dsm62_builds=${{ github.event.inputs.add_dsm62_builds || 'false' }}
           add_dsm61_builds=${{ github.event.inputs.add_dsm61_builds || 'false' }}
-          add_dsm52_builds=${{ github.event.inputs.add_dsm52_builds || 'false' }}
+          # TEMP (revert before merge): DSM 5.2 carries the toolchains this PR's floors
+          # actually bite on -- gcc 4.3.7 (ppc853x), 4.6.4 (88f6281), 4.7.3 (x86, x64).
+          add_dsm52_builds=${{ github.event.inputs.add_dsm52_builds || 'true' }}
           add_srm13_builds=${{ github.event.inputs.add_srm13_builds || 'false' }}
           add_srm12_builds=${{ github.event.inputs.add_srm12_builds || 'false' }}
           add_noarch_builds=${{ github.event.inputs.add_noarch_builds || 'false' }}
--- a/docs/developer-guide/packaging/makefile-variables.md
+++ b/docs/developer-guide/packaging/makefile-variables.md
@@ -320,6 +320,21 @@ without a network query. So a rustc floor refuses only the archs pinned to an ol
 from-source Rust — see [Toolchain: custom from-source
 Rust](../../framework/toolchain.md#custom-from-source-rust-toolchains).
 
+**Ask what stands in the way.** `make check-<arch>-<tcvers>` lists every gate the
+package's whole dependency tree fails for that architecture, so a floor declared three
+levels down is visible without starting a build:
+
+```
+$ make check-x86-5.2
+===>  tvheadend: 17 gate(s) not met for x86-5.2
+         cross/chromaprint-fftw     gcc 4.7.3 < 4.8
+         cross/python314            gcc 4.7.3 < 4.8
+         cross/ffmpeg8              gcc 4.7.3 < 4.9
+```
+
+`make ARCH=x86 TCVERSION=5.2 check` is the same with the pair in variables. The
+pre-check runs the same walk, so a refused build names every blocker at once.
+
 **Keep the floor consistent between `spk/` and `cross/`.** A floor on an `spk/`
 package belongs on its matching `cross/` package too, so a build is refused at its
 own level instead of failing deep in a dependency. If you add a floor to `spk/foo`,
--- a/docs/developer-guide/packaging/makefile-variables.md
+++ b/docs/developer-guide/packaging/makefile-variables.md
@@ -320,20 +320,31 @@ without a network query. So a rustc floor refuses only the archs pinned to an ol
 from-source Rust — see [Toolchain: custom from-source
 Rust](../../framework/toolchain.md#custom-from-source-rust-toolchains).
 
-**Ask what stands in the way.** `make check-<arch>-<tcvers>` lists every gate the
+**Ask what stands in the way.** `make check-<arch>-<tcvers>` lists every check the
 package's whole dependency tree fails for that architecture, so a floor declared three
 levels down is visible without starting a build:
 
 ```
 $ make check-x86-5.2
-===>  tvheadend: 17 gate(s) not met for x86-5.2
-         cross/chromaprint-fftw     gcc 4.7.3 < 4.8
-         cross/python314            gcc 4.7.3 < 4.8
+===>  tvheadend: x86-5.2 check: 17 failed, 14 more behind an optional dependency
+       required
          cross/ffmpeg8              gcc 4.7.3 < 4.9
+         cross/python314            gcc 4.7.3 < 4.8
+         ...
+       optional
+         cross/frei0r               gcc 4.7.3 < 7.5
+         cross/openexr              gcc 4.7.3 < 4.8
+         ...
 ```
 
-`make ARCH=x86 TCVERSION=5.2 check` is the same with the pair in variables. The
-pre-check runs the same walk, so a refused build names every blocker at once.
+`make ARCH=x86 TCVERSION=5.2 check` is the same with the pair in variables, and a clear
+architecture answers `x86-5.2 check: OK`. **required** is what the build actually needs;
+**optional** is what an `OPTIONAL_DEPENDS` branch would demand if that option were turned
+on -- worth knowing before turning it on, and never a reason to refuse the build. A
+package under both a required and an optional parent counts as required.
+
+The pre-check runs the same walk on the required set, so a refused build names every
+blocker at once instead of stopping at the first.
 
 **Keep the floor consistent between `spk/` and `cross/`.** A floor on an `spk/`
 package belongs on its matching `cross/` package too, so a build is refused at its
--- a/docs/framework/changes.md
+++ b/docs/framework/changes.md
@@ -24,11 +24,19 @@ If you only read one thing, read this. The details are in the dated log below.
   capability gate that architecture fails, or says every gate is met:
 
     ```
-    ===>  tvheadend: 17 gate(s) not met for x86-5.2
-             cross/chromaprint-fftw     gcc 4.7.3 < 4.8
-             cross/python314            gcc 4.7.3 < 4.8
+    ===>  tvheadend: x86-5.2 check: 17 failed, 14 more behind an optional dependency
+           required
              cross/ffmpeg8              gcc 4.7.3 < 4.9
+             cross/python314            gcc 4.7.3 < 4.8
+             ...
+           optional
+             cross/frei0r               gcc 4.7.3 < 7.5
+             ...
     ```
+
+    **required** is what the build needs; **optional** is what an `OPTIONAL_DEPENDS`
+    branch would demand if turned on, and never a reason to refuse. A clear architecture
+    answers `x86-5.2 check: OK`.
 
     The pre-check uses the same walk, so a refused build names every blocker at once
     instead of stopping at the first. See
@@ -99,8 +107,13 @@ If you only read one thing, read this. The details are in the dated log below.
 ---
 
 ??? note "September 7th 2026 — An arch exclusion says why, and names every blocker (#7439)"
+    - **Seven restated floors gone.** `spk/tvheadend`, `chromaprint`, `comskip` and
+      `spk/ffmpeg5-8` each declared `MIN_GCC_VERSION = 4.9`, restating what their own
+      `cross/` package declares -- and `cross/<same>` is a direct `DEPENDS`, so the walk
+      finds it. Declare a floor on the `cross/` package, where the requirement is a fact
+      about the code; the `spk/` inherits it by being walked. Verdicts unchanged.
     - **`make check-<arch>-<tcvers>`** reports every capability gate a package's whole
-      dependency tree fails for that architecture, or confirms it is clear.
+      dependency tree fails for that architecture, or answers `<arch>-<vers> check: OK`.
       `make ARCH=x86 TCVERSION=5.2 check` is the same thing with the pair in variables.
       A pure diagnostic: it extracts no toolchain and builds nothing, it reads the floors
       declared across the tree.
@@ -110,12 +123,12 @@ If you only read one thing, read this. The details are in the dated log below.
       same walk and lists all of them before stopping:
 
         ```
-        ===>  gate: cross/x264                 gcc 4.3.7 < 4.6
-        ===>  gate: cross/python314            gcc 4.3.7 < 4.8
-        ===>  gate: cross/ffmpeg8              gcc 4.3.7 < 4.9
+        ===>  check: cross/x264 gcc 4.3.7 < 4.6
+        ===>  check: cross/python314 gcc 4.3.7 < 4.8
+        ===>  check: cross/ffmpeg8 gcc 4.3.7 < 4.9
         ...
-        pre-check.mk:79: *** Arch 'ppc853x-5.2' is not supported by tvheadend:
-            gcc 4.3.7 < 4.9 (18 gate(s) in the tree).  Stop.
+        pre-check.mk:80: *** Arch 'ppc853x-5.2' is not supported by tvheadend
+            (18 failed check(s) in the tree).  Stop.
         ```
 
         It reuses `dependency-flat`'s walk, which already stamps each package so a diamond
--- a/docs/framework/changes.md
+++ b/docs/framework/changes.md
@@ -83,6 +83,34 @@ If you only read one thing, read this. The details are in the dated log below.
 
 ---
 
+??? note "September 7th 2026 — An arch exclusion can say why (#7439)"
+    - **What was missing:** `UNSUPPORTED_ARCHS` states *where* a package fails and never
+      *why*, and the archs are often added by an include rather than by the package -- so
+      the message named a package with no such list in its own Makefile.
+      **`UNSUPPORTED_ARCHS_REASON`** is now carried into the refusal in parentheses:
+
+        ```
+        Arch 'ppc853x' is not a supported architecture (go has no 32-bit PowerPC target)
+        ```
+
+        `spksrc.cross/env-go.mk` and `env-dotnet.mk` say theirs. A capability floor is
+        still the better answer where one fits; this is for exclusions that are not
+        capability checks.
+    - **One accumulator instead of two.** `_tc_cap_join` accumulated
+      `TC_CAPABILITY_UNSUPPORTED` and `unsupported_reason_join` accumulated
+      `UNSUPPORTED_ARCHS_REASON` -- the same operation under two names, and the second was
+      not a macro at all (no argument, just a lazy read of a global). Both are now
+      **`$(call comma_append,<list>,<item>)`**.
+    - **A `$(,)` that expanded to nothing.** `spksrc.common.mk` defined `empty` / `space` /
+      `$(,)` *below* every `spksrc.common/` include, so an included file expanding `$(,)`
+      in a `:=` assignment got the empty string -- which is why `tc-capability.mk` carried
+      a private comma variable. The utility block moved above the includes.
+    - **`spksrc.cross/env-dotnet.mk`** lost 26 tab-indented lines sitting outside any
+      recipe, and its `$(error)` became a `$(warning)`: an unsupported arch should be
+      reported by the pre-check, not by a parse abort in an env file.
+    - **Package-facing:** nothing to change. Add `UNSUPPORTED_ARCHS_REASON` alongside an
+      `UNSUPPORTED_ARCHS` you introduce, and the refusal will carry it.
+
 ??? note "September 4th 2026 — Build logs keep what the console showed (#7396)"
     - **What was lost:** a package refused by a pre-check left a build log holding a
       single line. `$(error)` fires at **parse** time, so no recipe of the inner make
--- a/docs/framework/changes.md
+++ b/docs/framework/changes.md
@@ -19,6 +19,21 @@ If you only read one thing, read this. The details are in the dated log below.
     cd cross/curl && make help
     ```
 
+- **Ask an architecture what stands in the way.** `make check-x86-5.2` (or
+  `make ARCH=x86 TCVERSION=5.2 check`) walks the whole dependency tree and lists every
+  capability gate that architecture fails, or says every gate is met:
+
+    ```
+    ===>  tvheadend: 17 gate(s) not met for x86-5.2
+             cross/chromaprint-fftw     gcc 4.7.3 < 4.8
+             cross/python314            gcc 4.7.3 < 4.8
+             cross/ffmpeg8              gcc 4.7.3 < 4.9
+    ```
+
+    The pre-check uses the same walk, so a refused build names every blocker at once
+    instead of stopping at the first. See
+    [Architecture Support](../developer-guide/packaging/makefile-variables.md#architecture-support).
+
 - **Declare what a package needs, not where it fails.** Instead of
   hand-maintaining an `UNSUPPORTED_ARCHS` list, state the capability floor:
   **`MIN_GCC_VERSION`**, **`MIN_GLIBC_VERSION`**, **`MIN_RUSTC_VERSION`**,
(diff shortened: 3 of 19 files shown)


th0ma7:
Extended with the generalisation discussed: a consumer now **inherits** a dependency's capability floor instead of restating it.

`python.mk` already read the selected python's `MIN_GCC_VERSION` rather than listing the archs it refuses — the right principle, written once, for python only. `inherit_floors` (in `macros.mk`) makes it general: given a package path, raise every floor this consumer declares to the one that package declares, and record who raised it.

```make
$(eval $(call inherit_floors,cross/$(FFMPEG_PACKAGE)))
```

It is a `sed` at parse time, not a dependency walk — this runs before `DEPENDS` is resolved — so it sees a floor written as a plain top-level assignment. One set inside a conditional stays the consumer's own to declare. (Transitive inheritance needs a walk, a walk needs recipes, and recipes come after the parse; that belongs in a separate change.)

`ffmpeg.mk` gains the same call, which is exactly what three packages were doing by hand — `tvheadend`, `comskip` and `chromaprint` each declared `MIN_GCC_VERSION = 4.9`, ffmpeg8's own floor, restated. The three restatements are gone.

The refusal now names the package that raised the floor:

```
Arch 'ppc853x-5.2' is not supported by tvheadend: gcc 4.3.7 < 4.9 (cross/ffmpeg8)
Arch 'x86-5.2' is not supported by bazarr: gcc 4.7.3 < 4.8 (cross/python312)
```

**No verdict changes.** 26 meta consumers scanned across 6 architectures (x64-7.1, 88f6281-6.2.4, x86-5.2, ppc853x-5.2, hi3535-6.2.4, qoriq-6.2.4) give a byte-identical refused/allowed set before and after — the floors are the same numbers, read from a different place.

`videodriver.mk` is deliberately untouched: its meta excludes by arch list, not by a capability floor, so there is nothing to inherit there.

th0ma7:
Two cleanups from review, on top of the inheritance above.

**The private comma variables were an include-order symptom.** `spksrc.common.mk` defined `empty` / `space` / `$(,)` at line 101 — below every `spksrc.common/` include (53-87). An included file expanding `$(,)` inside a `:=` assignment therefore got nothing, which is why `tc-capability.mk` carried a private `_tc_cap_comma`, and why I had copied the pattern. The utility block moves above the includes and both private copies are gone:

```
is not supported by zlib: glibc 2.8 < 99 (a runtime floor: no toolchain can lift it), gcc 4.3.7 < 99, rustc 1.82.0 < 99
```

**`unsupported_reason_join` was not a macro** — no argument, just a lazy read of a global, with both call sites repeating the same three-part append. It is now one:

```make
unsupported_reason = $(UNSUPPORTED_ARCHS_REASON)$(if $(strip $(UNSUPPORTED_ARCHS_REASON)),$(,) )$(1)

# spksrc.cross/env-go.mk
UNSUPPORTED_ARCHS_REASON := $(call unsupported_reason,go has no 32-bit PowerPC target)
```

It stays in `macros.mk` rather than moving to `pre-check.mk` because its callers run first: `spksrc.cross-go.mk` includes `common.mk` at line 39, `env-go.mk` at 42, and `cross-cc.mk` — hence `pre-check.mk` — at 88. The reason has to be assembled before the check reads it, so defining it in the consumer would invert the dependency.

Re-verified after the move: every `spk/` package parsed across x64-7.1, 88f6281-6.2.4 and ppc853x-5.2 with no unexpected error, and the refused/allowed set unchanged.

th0ma7:
Third review pass — two names for one operation, and a helper in the wrong file.

**`_tc_cap_join` and `unsupported_reason_join` were the same thing twice.** One accumulated `TC_CAPABILITY_UNSUPPORTED` (4 call sites in `tc-capability.mk`), the other `UNSUPPORTED_ARCHS_REASON` (2 sites in `env-go.mk` / `env-dotnet.mk`). Both are now the single generic helper, the only one left in `macros.mk`:

```make
comma_append = $(1)$(if $(strip $(1)),$(,) )$(2)

TC_CAPABILITY_UNSUPPORTED := $(call comma_append,$(TC_CAPABILITY_UNSUPPORTED),gcc $(TC_GCC) < $(MIN_GCC_VERSION)$(_tc_cap_from_MIN_GCC_VERSION))
UNSUPPORTED_ARCHS_REASON  := $(call comma_append,$(UNSUPPORTED_ARCHS_REASON),go has no 32-bit PowerPC target)
```

**The capability-floor machinery moved out of `macros.mk`.** `CAPABILITY_FLOORS`, `declared_floor` and `inherit_floors` are domain knowledge about capabilities, not generic make helpers, so they belong in `tc-capability.mk` — the file that defines what a floor is. They sit **above** that file's `ifneq ($(strip $(ARCH))$(strip $(TCVERSION)),)` guard: the metas call `inherit_floors` on every parse, and defining it inside the guard would make an arch-less parse expand it to nothing, silently.

Verified:

```
zlib, three floors : glibc 2.8 < 99 (a runtime floor: no toolchain can lift it), gcc 4.3.7 < 99, rustc 1.82.0 < 99
tvheadend          : gcc 4.3.7 < 4.9 (cross/ffmpeg8)
bazarr             : gcc 4.7.3 < 4.8 (cross/python312)
beszel-hub         : is not a supported architecture (go has no 32-bit PowerPC target)
```

And the full sweep: every `spk/` package parsed across x64-7.1, 88f6281-6.2.4 and ppc853x-5.2 — **519 combinations** — zero unexpected errors, refused/allowed set byte-identical to the run before the move.

th0ma7:
Fourth pass, plus DSM 5.2 turned on temporarily to exercise it.

**One call site instead of two.** `inherit_floors` is now called from `spksrc.spk-meta.mk`, the one place that knows which metas are in play, rather than from each meta file:

```make
# A meta's capability floors are the consumer's too, so it states what it needs of its own
# and inherits the rest. An unset meta contributes no word, and drops out of the loop.
$(foreach _meta,$(PYTHON_PACKAGE) $(FFMPEG_PACKAGE),$(eval $(call inherit_floors,cross/$(_meta))))
```

`base.mk` could not host it — `SPK_BASE_TEMPLATE` only instantiates at `spk-stage2`, long after a parse-time pre-check has to fire — and `meta.mk` only produces a diagnostic artifact. `videodriver` stays out on purpose: its meta is an `spk/`, not a `cross/`, and it excludes by arch list with no floor to inherit.

**`declared_floor` is gone.** It had one caller, so its `sed` is inline in the loop and its caveat moved into `inherit_floors`' comment. Three elements now — the floor list, the loop, its body — and `tc-capability.mk` is reordered so the vocabulary comes before the probes. `TC_HAS_LIBATOMIC`'s header went from 11 lines to 2, and the note about sitting outside the `ARCH`/`TCVERSION` guard now covers both blocks it precedes.

**tvheadend's comment block went too.** "PPC archs except qoriq are not supported" stated the *effect* of the 4.9 floor; the framework computes that now, and says it better:

```
ppc853x-5.2  ->  is not supported by tvheadend: gcc 4.3.7 < 4.9 (cross/ffmpeg8)
qoriq-6.2.4  ->  accepted (gcc 4.9.3, above the floor)
```

**DSM 5.2 is on for this run** (separate commit, revert before merge). It carries every toolchain below gcc 4.9 — ppc853x 4.3.7, 88f6281 4.6.4, x86 and x64 4.7.3 — which is exactly where the inherited floors bite. On master those packages are refused by a hand-written `MIN_GCC_VERSION = 4.9`; here they must be refused by the floor inherited from `cross/ffmpeg8`, naming it. Same refusals, different source.

Local sweep still clean: 519 (package, arch) combinations, zero unexpected errors, refused/allowed set unchanged.

th0ma7:
Docs added — and your example exposed a real gap, now fixed.

**Every blocker is named, not just the binding one.** `inherit_floors` raised the floor to the maximum, so only the winning demand survived: tvheadend on ppc853x named `cross/ffmpeg8` and python314 vanished from the message. That is misleading in the way that costs a round trip — drop ffmpeg8 and you are still blocked. `<FLOOR>_ASKS` now keeps every demand and the refusal lists all of them:

```
Arch 'ppc853x-5.2' is not supported by tvheadend: gcc 4.3.7 < 4.9
    (cross/python314>=4.8, cross/ffmpeg8>=4.9)

Arch '88f6281-6.2.4' is not supported by comskip: gcc 4.6.4 < 4.9 (cross/ffmpeg8>=4.9)
Arch 'x86-5.2' is not supported by bazarr: gcc 4.7.3 < 4.8 (cross/python312>=4.8)
```

A floor the package declares itself is reported with no name beside it, so the two cases stay distinguishable. This also let `<FLOOR>_FROM` go and dropped the `$(and ...)` guard — `version_ge` with an empty first argument already returns false.

One whitespace trap on the way: `foreach` joins its empty iterations (glibc, rustc, when the dependency declares neither) with a space, which landed inside the floor value and printed `gcc 4.3.7 < 4.9  (…)`. `_ASKS` is now emitted last, being a space-separated list that tolerates it.

**Docs.** `docs/framework/changes.md` gets a highlight and a dated entry; `docs/developer-guide/packaging/makefile-variables.md` states it under Architecture Support:

> **A meta's floor is inherited, not restated.** A package that selects a meta — `PYTHON_PACKAGE`, `FFMPEG_PACKAGE` — takes that package's floors automatically, and declares only what it needs *of its own*. […] Python wheels are the exception: their constraint is a wheel version rather than a toolchain floor, so a package that pins wheels still states those conditions by hand.

Sweep re-run after the change: 519 combinations, zero unexpected errors, refused/allowed set unchanged.

th0ma7:
Everything is in this PR now — the tree walk and `make check-<arch>-<tcvers>` were sitting on a branch I had not pushed, which is why the code was nowhere to be seen. My mistake in splitting it.

**`make check-x86-5.2`** — ask an architecture what stands in the way:

```
$ make check-x86-5.2                       # from spk/tvheadend
===>  tvheadend: 17 gate(s) not met for x86-5.2
         cross/chromaprint-fftw     gcc 4.7.3 < 4.8
         cross/python314            gcc 4.7.3 < 4.8
         cross/ffmpeg8              gcc 4.7.3 < 4.9
         ...

$ make ARCH=x86 TCVERSION=5.2 check         # same, pair in variables
$ make check-x64-7.1
===>  tvheadend: every gate met for x64-7.1
```

Pure diagnostic: no toolchain extracted, nothing built, it reads the floors declared across the tree. Works from `cross/` too. No collision with `checksum` — the pattern needs the literal `check-`. `pre-check.mk` stands down for the `check` goal, since a fatal check would cut the report short at the first gate.

**The pre-check now names every gate, not the first.** A package is as blocked by a floor it never declared as by one it did, and it only ever saw its own — so removing the blocker it named revealed the next, one build at a time:

```
===>  gate: cross/x264                 gcc 4.3.7 < 4.6
===>  gate: cross/python314            gcc 4.3.7 < 4.8
===>  gate: cross/ffmpeg8              gcc 4.3.7 < 4.9
...
pre-check.mk:79: *** Arch 'ppc853x-5.2' is not supported by tvheadend:
    gcc 4.3.7 < 4.9 (18 gate(s) in the tree).  Stop.
```

**It reuses `dependency-flat`'s walk**, which already stamps each package so a diamond is visited once, already forwards ARCH/TCVERSION so every package evaluates its own conditional `DEPENDS`, and already sets `DEPENDENCY_WALK=1` so a refused package reports rather than aborting. `DEP_FLAT_VERDICT` makes each visited package print its own `TC_CAPABILITY_UNSUPPORTED`, so one walk yields both the tree and every gate in it. `OPTIONAL_DEPENDS` drop out on their own once ARCH and TCVERSION are set.

**Measured:**

| | |
|---|---|
| New refusals | **0 / 519** (every `spk/` × x64-7.1, 88f6281-6.2.4, ppc853x-5.2) |
| Cost | **2.8s per parse** warm, on tvheadend's 141-package tree |

Verdict-neutral today, because the hand-maintained floors already cover everything. What changes is that the whole story is told at once, and a floor nobody thought to restate can no longer hide. If the per-parse cost proves too much, a stamp in the work dir would amortise it to once per build.

The earlier `inherit_floors` experiment is gone: the walk supersedes it rather than complementing it — each package computes its own verdict, so a consumer never needs to take ownership of its dependency's floor.

DSM 5.2 is on in a separate commit, revert before merge.

th0ma7:
Output reworked from testing in a real tree, and the restated floors are gone.

**Three fixes to the report.** stage0's "Bootstrapping toolchain" notice is an `$(info)`, so it landed on the stdout the walk is read from and appeared as a gate — the reader now takes only lines it emitted itself. `OPTIONAL_DEPENDS` were invisible, an arch context dropping them from the walk, which is right for the pre-check and wrong for a report. And the success line reads as a status now:

```
$ make check-x86-5.2
===>  tvheadend: x86-5.2 check: 17 failed, 14 more behind an optional dependency
       required
         cross/ffmpeg8              gcc 4.7.3 < 4.9
         cross/python314            gcc 4.7.3 < 4.8
         ...
       optional
         cross/highway              gcc 4.7.3 < 7.5
         cross/openexr              gcc 4.7.3 < 4.8
         ...

$ make check-x86-5.2                                        # in spk/borgbackup
===>  borgbackup: x86-5.2 check: OK
```

**Required vs optional is a set difference, not a label carried down the walk.** Required is the walk an arch context already defines; optional is what a second walk adds when `OPTIONAL_DEPENDS` are followed too. So a package under both a required and an optional parent is required — which is what it is, some parent needs it unconditionally. Propagating a label was tried first and is wrong: the stamp that visits a diamond once lets whichever edge arrives first decide, and `cross/ffmpeg8` came out optional at random. The pre-check reads the required group only.

**Seven restated floors removed.** `spk/tvheadend`, `chromaprint`, `comskip` and `spk/ffmpeg5-8` each declared `MIN_GCC_VERSION = 4.9`, restating what their own `cross/` package declares — and `cross/<same>` is a direct `DEPENDS`, so the walk finds it. Verdicts unchanged: all seven still refused on x86-5.2 (gcc 4.7.3), all seven still build on x64-7.1. The message shape changes where a package now declares nothing of its own:

```
before: is not supported by comskip: gcc 4.7.3 < 4.9
after:  is not supported by comskip (13 failed check(s) in the tree)
```

accurate, with the `===> check:` lines above naming every package that does declare one.

The python side is deliberately untouched for its own PR — `spk/python31x` keep their floor and `spksrc.spk-meta/python.mk` its targeted inheritance, both reaching much further than these seven.

DSM 5.2 is on, so CI exercises exactly the toolchains where these floors bite. Revert that commit before merge.

th0ma7:
@hgy59 and @mreid-tt I still have more code review to go over for this PR but this is a major upgrade on how arch are being validated as now more and more are being migrated using `MIN_*_VERSION`.  Important to note, this change is mandatory to properly support a gcc overlay so it is possible to clearly identify what's going on with any given package with any given overlay (overaly part to come later).

I would much appreciate if you could test this out using `make check-$(ARCH)-$(TCVERSION)` or `make ARCH=$(ARCH) TCVERSION=$(TCVERSION) check` on various use-cases that comes to mind where there could be corener-cases to further look into.

With the changes in this PR there is no more need to gate spk packages at all as it does a dependency walk and check all it's mandatory dependencies (and also prints its optional).  The only outstanding cases are with python wheels as there is no way to determine this from pip build.

Here's an example where this should help figuring out how the package behave for said architecture... and also indirectly tells wich packages/dependencies are still using old-schoold gating instead of newer:
```
spksrc@spksrc13:~/pre-check-update/spksrc/spk/jellyfin$ make check-ppc853x-5.2
===>  jellyfin: ppc853x-5.2 check: 2 failed
       required
         cross/jellyfin             unsupported arch (dotnet ships no runtime for this arch)
         spk/jellyfin               unsupported arch, DSM 5.2 < 7.2

spksrc@spksrc13:~/pre-check-update/spksrc/cross/tvheadend$ make check-ppc853x-5.2
===>  tvheadend: ppc853x-5.2 check: 14 failed, 23 more behind an optional dependency
       required
         cross/chromaprint-fftw     gcc 4.3.7 < 4.8
         cross/dav1d                gcc 4.3.7 < 4.8
         cross/ffmpeg8              gcc 4.3.7 < 4.9
         cross/harfbuzz             gcc 4.3.7 < 4.8
         cross/libass               gcc 4.3.7 < 4.8
         cross/librist              gcc 4.3.7 < 4.8
         cross/libvpx-1.14          gcc 4.3.7 < 4.6
         cross/libvpx               gcc 4.3.7 < 4.6
         cross/libzimg              gcc 4.3.7 < 4.8
         cross/shine                gcc 4.3.7 < 4.6
         cross/snappy               gcc 4.3.7 < 4.8
         cross/tvheadend            gcc 4.3.7 < 4.9
         cross/x264                 gcc 4.3.7 < 4.6
         cross/x265                 gcc 4.3.7 < 4.8
       optional
         cross/cairo-latest         gcc 4.3.7 < 4.8
         cross/c-ares-latest        unsupported arch
         cross/frei0r               gcc 4.3.7 < 7.5
         cross/glib-2.66            gcc 4.3.7 < 4.6
         cross/glib-latest          gcc 4.3.7 < 4.6
         cross/highway              gcc 4.3.7 < 7.5
         cross/Khronos-SPIRV-LLVM-Translator-140 unsupported arch, DSM 5.2 < 7.1
         cross/libaom-3.8           gcc 4.3.7 < 4.8
         cross/libaom               gcc 4.3.7 < 4.8
         cross/libaom-latest        gcc 4.3.7 < 7.5
         cross/libjxl               gcc 4.3.7 < 7.5
         cross/liblc3               gcc 4.3.7 < 4.9
         cross/libvmaf              gcc 4.3.7 < 7.5
         cross/libvpx-latest        gcc 4.3.7 < 7.5
         cross/llvm-140             unsupported arch, DSM 5.2 < 7.1
         cross/nghttp3              gcc 4.3.7 < 4.9
         cross/ngtcp2               gcc 4.3.7 < 4.9
         cross/openexr              gcc 4.3.7 < 4.8
         cross/openh264             unsupported arch
         cross/pngquant3            gcc 4.3.7 < 4.8
         cross/rubberband           gcc 4.3.7 < 7.5
         cross/svt-av1              gcc 4.3.7 < 4.9, requires a 64-bit architecture
         cross/vvenc                gcc 4.3.7 < 7.5
```

Also the gh-action CI now prints how many check(s) are unmet which provides more information than currently.

Lastly, this PR removes all gating for ffmpeg* dependent packages.  A subsequent PR will remove all gating from python3* core packages but this would pull to many packages to be built part of a single gh-action CI build.

Your feedback would be much appreciated, thnx!

hgy59:
@th0ma7 very good idea.
My concerns relate to `make all-supported` - some adjustments may be needed here as well (if I remember correctly, @mreid-tt kept the UNSUPPORTED_ definitions somewhere so that all-supported would continue to work).

mreid-tt:
Hey @th0ma7, this is a great initiative. I'm not able to test right now but per @hgy59's comment, we do have some `UNSUPPORTED_` definitions that are not related to `MIN_*_VERSION`. These come to mind:

https://github.com/SynoCommunity/spksrc/blob/26a6054745256c16ea86d00f292967fd0e544d9a/spk/ffsync/Makefile#L13-L18

https://github.com/SynoCommunity/spksrc/blob/26a6054745256c16ea86d00f292967fd0e544d9a/cross/vaultwarden/Makefile#L14-L17

https://github.com/SynoCommunity/spksrc/blob/26a6054745256c16ea86d00f292967fd0e544d9a/mk/spksrc.common/archs.mk#L94-L105

th0ma7:
> we do have some `UNSUPPORTED_` definitions that are not related to `MIN_*_VERSION`. These come to mind:
> https://github.com/SynoCommunity/spksrc/blob/26a6054745256c16ea86d00f292967fd0e544d9a/spk/ffsync/Makefile#L13-L18

```
spksrc@spksrc13:~/pre-check-update/spksrc/spk/ffsync$ make check-ppc853x-5.2
===>  ffsync: ppc853x-5.2 check: 5 failed, 4 more behind an optional dependency
       required
         cross/diesel               unsupported arch
         cross/python314            gcc 4.3.7 < 4.8
         cross/syncstorage-rs       rustc 1.82.0 < 1.91
         spk/ffsync                 gcc 4.3.7 < 4.8, rustc 1.82.0 < 1.91, DSM 5.2 < 6.0
         spk/python314              gcc 4.3.7 < 4.8
       optional
         cross/c-ares-latest        unsupported arch
         cross/mariadb-connector-c-latest DSM 5.2 < 7.1
         cross/nghttp3              gcc 4.3.7 < 4.9
         cross/ngtcp2               gcc 4.3.7 < 4.9
```
But I'll double-check if the `UNSUPPORTED_ARCHS_TCVERSION` is also being managed properly... :
```
spksrc@spksrc13:~/pre-check-update/spksrc/spk/ffsync$ make check-aarch64-6.2.4 
===>  ffsync: aarch64-6.2.4 check: 1 failed, 1 more behind an optional dependency
       required
         spk/ffsync                 unsupported arch
       optional
         cross/mariadb-connector-c-latest DSM 6.2.4 < 7.1
```
> https://github.com/SynoCommunity/spksrc/blob/26a6054745256c16ea86d00f292967fd0e544d9a/cross/vaultwarden/Makefile#L14-L17
 
```
===>  vaultwarden: ppc853x-5.2 check: 1 failed, 4 more behind an optional dependency
       required
         cross/vaultwarden          rustc 1.82.0 < 1.95
       optional
         cross/c-ares-latest        unsupported arch
         cross/mariadb-connector-c-latest DSM 5.2 < 7.1
         cross/nghttp3              gcc 4.3.7 < 4.9
         cross/ngtcp2               gcc 4.3.7 < 4.9
```
> https://github.com/SynoCommunity/spksrc/blob/26a6054745256c16ea86d00f292967fd0e544d9a/mk/spksrc.common/archs.mk#L94-L105

This one is rather interesting... as there is the following in `spksrc.cross/env-dotnet.mk`:
```
# NOTE: 32bit (x86) is not supported:
# https://github.com/dotnet/core/issues/5403
# https://github.com/dotnet/core/issues/4595
UNSUPPORTED_ARCHS += $(PPC_ARCHS) $(ARMv5_ARCHS) $(i686_ARCHS) $(ARMv7L_ARCHS)
```
Which cumulates on top of what's into `spksrc.common/archs.mk`:
```
# Notes for .NET 6 compatibility:
# 1. dotnet for x86 (32-bit) is unsupported on linux and must be built from source
# 2. ARMv7_ARCHS without full vfpv3 support (having only vfpv3-d16) are not supported
# 3. SRM ARMv7 archs are not supported
# 4. Certain combinations of ARMv7 and DSM are incompatible (issues #4790, #5089, #5302, #5315)
# 5. Comprehensive ARMv7 testing conducted under issue #5574 resulted in the following exclusions

# Exclusions for dotnet core apps
ifeq ($(strip $(DOTNET_CORE_ARCHS)),1)
    UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS) $(i686_ARCHS) armada370 alpine comcerto2k
    UNSUPPORTED_ARCHS_TCVERSION = armv7-6.2.4 armv7-1.2 armv7-1.3
endif

# Exclusions for dotnet 6.0 servarr apps (except x86)
ifeq ($(strip $(DOTNET_SERVARR_ARCHS)),1)
    UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS) armada370 alpine comcerto2k
    UNSUPPORTED_ARCHS_TCVERSION = armv7-6.2.4 armv7-1.2 armv7-1.3
endif

# Exclusions for dotnet 6.0 servarr apps (except x86)
# ARMv7 incompatibility — see: https://github.com/dotnet/runtime/issues/109739
ifeq ($(strip $(DOTNET_SERVARR_ARCHS)),2)
    UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS) $(ARMv7_ARCHS)
endif
```
The portion in `archs.mk` isn't common, it's rather specific to dotnet.  I'll look into consolidating that in its own `env-dotnet.mk` to avoid confusion, thus allowing to have proper identification of why an arch is being refused using the `make check-%` facility.

th0ma7:
> My concerns relate to `make all-supported` - some adjustments may be needed here as well (if I remember correctly, @mreid-tt kept the UNSUPPORTED_ definitions somewhere so that all-supported would continue to work).

Any specific pointers related to that?  or @mreid-tt your previous post are pointers related to ensuring `make all-supported` continues to work?

th0ma7:
Two follow-ups from the review.

**`UNSUPPORTED_ARCHS_TCVERSION` was handled, but indistinguishable.** `_dep_flat_listed` already filtered on `$(ARCH)-$(TCVERSION)`, which is why `spk/ffsync` was caught on aarch64-6.2.4 — but both lists printed the same "unsupported arch", so the message never said which variable to go and look in. They are now reported apart, and a package excluded by both shows both:

```
$ make check-aarch64-6.2.4                  # in spk/ffsync
===>  ffsync: aarch64-6.2.4 check: 1 failed, 1 more behind an optional dependency
       required
         spk/ffsync                 unsupported arch-tcversion
```

**The dotnet consolidation is #7446.** One finding worth flagging before you look: the blocks cannot move into `spksrc.cross/env-dotnet.mk`. Its consumers are `spk/` packages — sonarr, radarr, lidarr, readarr, prowlarr, ombi, kavita, dotnet\*-runtime — and they include `spksrc.spk.mk`, never `env-dotnet.mk`, which only `spksrc.cross-dotnet.mk` pulls in. Moving them there would silently drop every one of their exclusions.

They go to a new `mk/spksrc.common/dotnet.mk` instead, included right after `archs.mk` because they need its arch groups, with `env-dotnet.mk` keeping the `cross/` side and the two files pointing at each other. Each block gains an `UNSUPPORTED_ARCHS_REASON`, unused until this PR lands, so the two merge in either order. Verified behaviour-neutral across 35 package/arch combinations.

On @hgy59's `make all-supported` concern: nothing here changes what `UNSUPPORTED_ARCHS` or `UNSUPPORTED_ARCHS_TCVERSION` mean or when they apply — the walk *reads* them to explain a refusal, it does not replace them. The seven floors removed so far are `MIN_GCC_VERSION` restatements of a `cross/` package the walk already reaches, not arch lists.

PR #7440, opened 2026-09-07
toolchain: give each build its own tc_vars

th0ma7:
The toolchain work dir holds **one** shared `tc_vars*`, and `tcvars` regenerates it only while its cookie is absent:

```make
ifeq ($(wildcard $(TCVARS_COOKIE)),)
tcvars: ... generate_tc_vars_mk generate_tc_vars_other $(TCVARS_COOKIE)
else
tcvars: ;
endif
```

So the first package to build against a toolchain fixes its overlay state, and every package after inherits it. Ask for a different one and nothing happens:

```bash
make -C cross/<pkg> ARCH=qoriq TCVERSION=6.2.4 OVERLAY_RUSTC=1   # ... then
make -C cross/<pkg> ARCH=qoriq TCVERSION=6.2.4 OVERLAY_RUSTC=0
```

`TC_OVERLAY_RUSTC`, `TC_EXTRA_RUSTFLAGS` and `TC_OVERLAY_BINUTILS` keep whatever the first run wrote, because the toolchain is already installed and the cookie short-circuits the regeneration. **The switch is silently ignored** — the second build does not fail, it quietly uses the first one's toolchain.

## The change

`toolchainclean` drops the generated `tc_vars*`, their cookie, and `TOOLCHAIN_COOKIE`.

Removing the last is both consistent — without its `tc_vars` the toolchain is not fully installed — and *necessary*: those cookies are tested with `$(wildcard)` at **parse** time, so a single make cannot clean and regenerate in one pass (`make: 'tcvars' is up to date`). The rebuild has to come from the next invocation, and the missing cookie is what triggers it.

Per-step cookies stay, so nothing is re-downloaded or re-extracted: `toolchain` walks its `DEPENDS` again and rewrites `tc_vars*`. On an installed toolchain this is a few seconds.

## Called from `stage0` alone

Placed **outside** its `ARCH`/`TCVERSION` guards, which is what lets one site cover both entry points:

| entry | |
|---|---|
| `make -C cross/<pkg> ARCH=x TCVERSION=y` | sets both |
| `make arch-<arch>-<vers>` | sets neither — it carries them in the goal, read back from `MAKECMDGOALS` |

The second is how CI builds every package. Splitting the goal on dashes is safe for the same reason `supported.mk` already relies on: no arch name or DSM version contains one. `supported.mk` is untouched.

Only a **directly invoked** package cleans. Dependencies already agree with whoever pulled them in, and with `PARALLEL_MAKE=max` several of them parse at once — one clearing the file while another reads it is a race. `MAKELEVEL` is the discriminator: `0` from a shell, non-zero for every sub-make.

## Testing

Both directions, on an already-installed toolchain:

```
make arch-qoriq-6.2.4                 tc_vars removed, then regenerated
make ARCH=qoriq TCVERSION=6.2.4 ...   same
the same call through a sub-make      untouched
```

`make toolchainclean` is listed by `make help` in a toolchain directory and documented in the build-rules and makefile-system references. `changes.md` records it as a fix to the overlay switches from #7353, whose own entry promises `command line > environment > local.mk > defaults` — a precedence that did not survive an already-installed toolchain.

The defect is behind a non-default switch (`OVERLAY_RUSTC` is 1 in `local.mk`, `OVERLAY_BINUTILS` 0), so there is no failing job to point at — it shows as a switch that does nothing, which is the worse kind.

Extracted from #7391, where a gcc overlay makes the same defect loud: a package built without the overlay inherited `-latomic` from one built with it and died at `ld: cannot find -latomic`.

### DIFF
--- a/docs/developer-guide/packaging/build-rules.md
+++ b/docs/developer-guide/packaging/build-rules.md
@@ -56,6 +56,7 @@ Run from `native/<package>/` directory (host tools built once, then reused via
 | `download` `checksum` `extract` `patch` `configure` `compile` `install` | Individual build lifecycle steps, in order |
 | `build-archive` / `print-archive-name` | Create the reusable archive on demand / print its filename (opt-in via `ARCHIVE_NAME`, see `spksrc.build/archive.mk`) |
 | `nativeclean` | Drop **this** package's build cookies so every step re-runs next `make`, keeping the work dir (source, install, archive). The native counterpart of `spkclean`; leaves dependencies' cookies alone |
+| `toolchainclean` | Drop a toolchain's generated `tc_vars*` and the cookies guarding them, so the next build regenerates them for its own overlay state. Run in `toolchain/syno-<arch>-<vers>/`. Keeps the extracted toolchain and its per-step cookies, so nothing is re-downloaded |
 | `clean` / `smart-clean` | Remove all work directories / this package's source and cookies |
 
 To re-run a single step, remove its one cookie, e.g.
--- a/docs/framework/changes.md
+++ b/docs/framework/changes.md
@@ -83,6 +83,40 @@ If you only read one thing, read this. The details are in the dated log below.
 
 ---
 
+??? note "September 7th 2026 — An overlay switch that did nothing on an installed toolchain (#7440)"
+    - **What was wrong:** the toolchain work dir holds ONE shared `tc_vars*`, and `tcvars`
+      regenerates it only while its cookie is absent. So the first package to build against
+      a toolchain fixed its overlay state, and every package after inherited it. Asking for
+      a different one changed nothing:
+
+        ```bash
+        make -C cross/<pkg> ARCH=qoriq TCVERSION=6.2.4 OVERLAY_RUSTC=1   # ... then
+        make -C cross/<pkg> ARCH=qoriq TCVERSION=6.2.4 OVERLAY_RUSTC=0
+        ```
+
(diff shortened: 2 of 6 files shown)


th0ma7:
Superseded by #7442 — closing.

This PR arbitrated an owner for the `tc_vars*` set shared by `toolchain/syno-<arch>-<vers>/work`: a `toolchainclean` run by whoever was the root of the build. Two things were wrong with that.

**It did not cover every entry point.** The guard was `ifeq ($(MAKELEVEL),0)`, but the root `Makefile` reaches a package through

```make
%: spk/%/Makefile
	cd $(dir $^) && env $(MAKE)
```

so the package's make is a child and `MAKELEVEL` is 1. Measured:

```
make -C sub/pkg     ->  MAKELEVEL in pkg = 0
make pkg  (root)    ->  MAKELEVEL in pkg = 1
```

`make tvheadend`, `make spk-tvheadend-x64-7.1` and `make all` — three documented targets — therefore never cleaned, and picked up whatever the first build had left. Latent on master, where every tree resolves to the same compiler; active the moment the gcc-8.5 overlay lands.

**And the criterion was wrong in principle.** Recursion depth does not say whether a make is the root of a build tree or a child handed someone else's decisions.

#7442 removes the shared file rather than choosing an owner for it. `WORK_DIR` already scopes exactly one build tree — `depend.mk` passes the root's value down through `$(ENV)`, `directories.mk` keeps what it is handed (`ifndef`), and the `env -i` around an spk meta source ends one tree and starts the next — so each build generates its own `tc_vars*` under its own work directory and reads nothing else. No shared file, no owner to designate, no `toolchainclean`, and two builds can run in parallel against one toolchain.

The `toolchainclean` documentation and `make help` entry from this PR go away with it.

Issue #5574, opened 2023-01-22
Testing dotNET *arr packages with ARMv7 archs


mreid-tt:
I have noted some comments regarding excluded ARMv7 archs for ***arr** packages (Sonarr, Radarr, Lidarr) which suggest that these architectures may not actually be incompatible. In fact, based on the source code, many of these architectures are listed as not fully tested for dotNET.

I would like to solicit support for explicit testing of these packages on ARMv7 archs. Specifically, these include [alpine, alpine4k](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#alpine), [~~armada370~~](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#armada370), [armada375](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#armada375), [armada38x](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#armada38x), [armadaxp](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#armadaxp), [monaco](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#monaco) and [comcerto2k](https://github.com/SynoCommunity/spksrc/wiki/Architecture-per-Synology-model#comcerto2k). My suspicion is that these archs may be affected by specific dotNET versions. Given that a number of the newer ***arr** releases have updated their included dotNET binaries, I'd like some assistance from the community in narrowing down which versions work and which do not.

If you'd like to assist with this effort I'd be very grateful for your feedback. As there is a lot to test, please keep your feedback brief. Do include the specific model, the package you tested as well as any short error messages. While I do appreciate longer error messages and logs, I would ask that these be uploaded to a site like [Pastebin.com](https://pastebin.com/) and include only the link in your comment so that this thread does not get too long. A great comment example would be as follows:

> Package tested: Sonarr v4
> NAS used: DS216se (armada370)
> DSM version: 7.1.1 Update 2
> Installation status: failed to start
> Service log: https://pastebin.com/6GfvYRFH (/var/packages/sonarr/var/sonarr.log)

With the community's help we can hopefully re-include these excluded ARMv7 archs in future dotNET ***arr** packages. To get started, you can download the ARMv7 packages for testing here:

[...]

Issue #5315, opened 2022-06-15
Segmentation fault (core dumped) .net


fremat79:
### Is this a new Bug?

- [X] I checkd that the bug hasn't been reported before

### Package Name

.net runtime 6

### Package Version

 dotnet-runtime 6.0.4-1

### Device Model

ds213j

### Device Architecture

ARMv7 - armada370

### Firmware Version

DSM 6.2.4-25556 Update 6

### What happened?

Hi, 
After installing ,net 6.0 runtime on my DS213J I receive this error message when launc dotnet commadn on SSH console

https://synocommunity.com/package/dotnet-runtime

My DSM version in Full DSM version:DSM 6.2.4-25556 Update 6

### Reproduction steps

```bash
1.Install package https://synocommunity.com/package/dotnet-runtime
2.run dotnet command on terminal
3.
...
```

### Install Log

```log
2022/06/15 22:34:58     install dotnet-runtime 6.0.4-1 Begin preinst
2022/06/15 22:34:59     Begin reload_inst_variables
2022/06/15 22:34:59     End reload_inst_variables
2022/06/15 22:34:59     Begin initialize_variables
2022/06/15 22:34:59     End initialize_variables
2022/06/15 22:34:59     ===> Step preinst. USER= GROUP= SHARE_PATH=
2022/06/15 22:34:59     install dotnet-runtime 6.0.4-1 End preinst ret=[0]
2022/06/15 22:34:59     install dotnet-runtime 6.0.4-1 Begin /bin/rm -rf /volume1/@appstore/dotnet-runtime
2022/06/15 22:34:59     install dotnet-runtime 6.0.4-1 End /bin/rm -rf /volume1/@appstore/dotnet-runtime ret=[0]
2022/06/15 22:34:59     install dotnet-runtime 6.0.4-1 Begin /bin/mv -f /volume1/@tmp/pkginstall/package /volume1/@appstore/dotnet-runtime
2022/06/15 22:34:59     install dotnet-runtime 6.0.4-1 End /bin/mv -f /volume1/@tmp/pkginstall/package /volume1/@appstore/dotnet-runtime ret=[0]
2022/06/15 22:34:59     install dotnet-runtime 6.0.4-1 Begin /bin/rm -rf /var/packages/dotnet-runtime
2022/06/15 22:34:59     install dotnet-runtime 6.0.4-1 End /bin/rm -rf /var/packages/dotnet-runtime ret=[0]
[...]

Issue #5302, opened 2022-06-06
Radarr not working for DSM1517 - Seg Fault (ANNAPURNALABS Alpine AL314)


russholio:
### Is this a new Bug?

- [X] I checkd that the bug hasn't been reported before

### Package Name

Radarr

### Package Version

202205015-18

### Device Model

DS1517

### Device Architecture

ARMv7

### Firmware Version

DSM 7.0.1-42218 Update 4

### What happened?

I tried to do a brand new install of Radarr on my DS1517, I have never used Radarr before.

The package installs but does not start and the DSM package manager immediately shows "Repair" package.

Repairing the package does nothing.

I `ssh` in and checked the logs, I see that there is a Seg fault.

I am running Sonarr which I believe to be very similar in its tech stack and was surprised that Radarr wasn't working but Sonarr is. 

I checked the listed build architectures on Synocommunity for Sonarr and Radarr and they seem to differ by 2 build arch's 
7.0 aarch64
7.0 evansport

### Reproduction steps

```bash
1. Install Radarr on a DSM1517
```

### Install Log

```log
2022/06/06 14:43:53 install radarr 20220515-18 Begin preinst
2022/06/06 14:43:53 Begin reload_inst_variables
2022/06/06 14:43:53 End reload_inst_variables
2022/06/06 14:43:53 Begin initialize_variables
2022/06/06 14:43:53 End initialize_variables
2022/06/06 14:43:53 ===> Step preinst. USER=sc-radarr GROUP=sc-download SHARE_PATH=
2022/06/06 14:43:53 install radarr 20220515-18 End preinst ret=[0]
2022/06/06 14:43:54 install radarr 20220515-18 Begin /bin/mkdir -p /volume2/@appstore/radarr
2022/06/06 14:43:54 install radarr 20220515-18 End /bin/mkdir -p /volume2/@appstore/radarr ret=[0]
2022/06/06 14:43:54 install radarr 20220515-18 Begin /bin/rm -rf /volume2/@appstore/radarr
2022/06/06 14:43:54 install radarr 20220515-18 End /bin/rm -rf /volume2/@appstore/radarr ret=[0]
2022/06/06 14:43:54 install radarr 20220515-18 Begin /bin/mv -f /volume2/@tmp/pkginstall/extract.gc9OXf/package /volume2/@appstore/radarr
[...]

Issue #5089, opened 2022-01-20
Internal updater of Radarr doesn't work to update to 4.x versions


thyeestes:
### Is this a new Bug?

- [X] I checkd that the bug hasn't been reported before

### Package Name

Radarr

### Package Version

20210311-15

### Device Model

DS216Play

### Device Architecture

ARMv7

### Firmware Version

6.2.4-25556 Update 3

### What happened?

2022-01-20 03:28:18.9|Info|InstallUpdateService|Deleting old update files
2022-01-20 03:28:19.8|Info|InstallUpdateService|Downloading update 4.0.2.5836
2022-01-20 03:29:06.8|Info|InstallUpdateService|Verifying update package
2022-01-20 03:29:09.3|Info|InstallUpdateService|Update package verified successfully
2022-01-20 03:29:09.3|Info|InstallUpdateService|Extracting Update package
2022-01-20 03:29:37.9|Info|InstallUpdateService|Update package extracted successfully
2022-01-20 03:29:37.9|Info|BackupService|Starting Backup
2022-01-20 03:29:40.7|Info|InstallUpdateService|Preparing client
2022-01-20 03:29:43.9|Info|InstallUpdateService|Starting update client /volume1/@appstore/radarr/tmp/radarr_update/Radarr.Update
2022-01-20 03:29:43.9|Info|InstallUpdateService|Radarr will restart shortly.

Neither debug or trace log show additional information not errors

### Reproduction steps

```bash
1.Open Radarr
2. Go to System - Updates - Install latest
3. Update starts but always ends with Radarr will restart shortly which it never does.

I tried resizing /tmp as suggested in https://github.com/SynoCommunity/spksrc/issues/4621
```

### Install Log

```log
Sun Nov 28 12:31:13 CET 2021
===> Step preinst. USER=radarr GROUP=sc-download SHARE_PATH=
Sun Nov 28 12:31:20 CET 2021
===> Step postinst. USER=radarr GROUP=sc-download SHARE_PATH=
Installing service configuration /var/packages/radarr/conf/radarr.sc
Adding 'sc-radarr' to 'sc-download'
Group Name: [sc-download]
Group Type: [AUTH_LOCAL]
Group ID:   [65536]
Group Members: 
0:[plex]
1:[sc-medusa]
2:[sc-nzbget]
3:[thyestes]
4:[sc-radarr]
Invoke service_postinst
[...]

PR #7446, opened 2026-09-09
dotnet: keep the arch exclusions with dotnet, not in the common arch groups

th0ma7:
## The problem

Answering "why is this arch refused for sonarr" meant looking in two unrelated files, neither named after dotnet:

- `spksrc.common/archs.mk` — which otherwise just classifies architectures (`ARMv7_ARCHS`, `PPC_ARCHS`, `64bit_ARCHS`) — carried three blocks listing the archs a **dotnet** package refuses.
- `spksrc.cross/env-dotnet.mk` carried a fourth, overlapping list of its own.

## One file

They move to a new **`mk/spksrc.common/dotnet.mk`**, included right after `archs.mk` since they need its arch groups.

**It has to stay under `spksrc.common/`.** The consumers are `spk/` packages — sonarr, radarr, lidarr, prowlarr, ombi, kavita, dotnet\*-runtime — which include `spksrc.spk.mk` and never see `env-dotnet.mk`, pulled in only by `spksrc.cross-dotnet.mk`. `cross/libe_sqlite3` uses the same flags from the cross tree, which confirms it.

`env-dotnet.mk` gives up its copy: `spksrc.cross-dotnet.mk` raises `DOTNET_BUILD_ARCHS` before it includes `common.mk` and opts in like any other consumer. A separate flavour on purpose — the build list is narrower than the app list, and reusing the app flag would have excluded archs `cross/jellyfin` builds on today.

Two flavours remain, both with real users. **`DOTNET_SERVARR_ARCHS` is gone**: nothing in the tree declares it. `= 1` was already dead on master, and `= 2` had exactly one user, `spk/readarr`, which now declares its own list — no flavour fits it, since every one of them excludes i686 while upstream does ship a `linux-core-x86` build that evansport uses.

## A parse error along the way

26 lines in `env-dotnet.mk` were indented with a tab outside any recipe. Anything reaching them before a rule is defined dies, so `cross/jellyfin` could not be parsed **at all** on an arch dotnet does not support — the exclusion was never even reached:

```
before: env-dotnet.mk:43: *** recipe commences before first target.  Stop.
after:  env-dotnet.mk:38: Unsupported ARCH ppc853x
        pre-check.mk:71: *** Arch 'ppc853x' is not a supported architecture.  Stop.
```

The `$(error)` for an unresolved `DOTNET_ARCH` becomes a `$(warning)`, as its own comment always asked.

## Reasons

`UNSUPPORTED_ARCHS_REASON` is picked by the arch at hand rather than one umbrella line — dotnet is absent here for four different grounds:

```
armada370            dotnet needs full vfpv3 and this toolchain is vfpv3-d16
alpine               dotnet segfaults on this arch despite capable silicon (issue #5302)
comcerto2k           dotnet is incompatible with this arch (issue #5574)
armv7-6.2.4 ...      dotnet is incompatible with this arch/DSM pair (issues #4790 #5089 #5302 #5315)
PPC, ARMv5, ARMv7L   dotnet has no runtime port for this arch
readarr on ARMv7     dotnet 6.0 segfaults on ARMv7 (dotnet/runtime#109739)
```

`armada370` is structurally incompatible (`-mfpu=vfpv3-d16`); `alpine` is empirically broken on capable silicon (`-mfpu=neon-vfpv4`). A package declaring its own list still reuses that map for the grounds it shares — readarr's `$(or ...)` falls through to it for PPC, ARMv5 and ARMv7L.

Verified against the binaries spksrc downloads: Readarr **0.4.16.2793** ships `"tfm": "net6.0"` with `Microsoft.NETCore.App 6.0.35`, and so does **0.4.18.2805**, the latest release — upstream has published nothing since June 2025. The ARMv7 exclusion is not going away on its own.

The variable is unused until #7439 lands, so **the two PRs merge in either order** — this one no longer shares a file with it.

## Testing

Behaviour-neutral: 7 dotnet packages across `armv7-6.2.4`, `ppc853x-5.2`, `evansport-7.1` and `x64-7.1` give an identical excluded/allowed set against master. Rebased on `3cb895bb0` (Jellyfin 12.0 / dotnet 10.0).

### DIFF
--- a/mk/spksrc.common.mk
+++ b/mk/spksrc.common.mk
@@ -61,6 +61,9 @@ include $(BASEDIR)/mk/spksrc.common/stage0.mk
 # Load common definitions
 include $(BASEDIR)/mk/spksrc.common/archs.mk
 
+# Which archs a dotnet package refuses (needs the arch groups above)
+include $(BASEDIR)/mk/spksrc.common/dotnet.mk
+
 include $(BASEDIR)/mk/spksrc.common/logs.mk
 
 # Load local configuration
--- a/mk/spksrc.common/archs.mk
+++ b/mk/spksrc.common/archs.mk
@@ -91,31 +91,6 @@ OLD_PPC_ARCHS = powerpc ppc824x ppc853x ppc854x
 # outdated unsupported archs
 DEPRECATED_ARCHS = powerpc ppc824x ppc854x ppc853x
 
-# Notes for .NET 6 compatibility:
-# 1. dotnet for x86 (32-bit) is unsupported on linux and must be built from source
-# 2. ARMv7_ARCHS without full vfpv3 support (having only vfpv3-d16) are not supported
-# 3. SRM ARMv7 archs are not supported
-# 4. Certain combinations of ARMv7 and DSM are incompatible (issues #4790, #5089, #5302, #5315)
-# 5. Comprehensive ARMv7 testing conducted under issue #5574 resulted in the following exclusions
-
-# Exclusions for dotnet core apps
-ifeq ($(strip $(DOTNET_CORE_ARCHS)),1)
-    UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS) $(i686_ARCHS) armada370 alpine comcerto2k
-    UNSUPPORTED_ARCHS_TCVERSION = armv7-6.2.4 armv7-1.2 armv7-1.3
-endif
-
-# Exclusions for dotnet 6.0 servarr apps (except x86)
-ifeq ($(strip $(DOTNET_SERVARR_ARCHS)),1)
-    UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS) armada370 alpine comcerto2k
-    UNSUPPORTED_ARCHS_TCVERSION = armv7-6.2.4 armv7-1.2 armv7-1.3
-endif
-
-# Exclusions for dotnet 6.0 servarr apps (except x86)
-# ARMv7 incompatibility — see: https://github.com/dotnet/runtime/issues/109739
-ifeq ($(strip $(DOTNET_SERVARR_ARCHS)),2)
-    UNSUPPORTED_ARCHS = $(PPC_ARCHS) $(ARMv5_ARCHS) $(ARMv7L_ARCHS) $(ARMv7_ARCHS)
-endif
-
 # Filter to exclude TC versions greater than DEFAULT_TC (from local configuration)
 TCVERSION_DUPES = $(addprefix %,$(filter-out $(DEFAULT_TC),$(AVAILABLE_TCVERSIONS)))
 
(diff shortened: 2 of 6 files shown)


mreid-tt:
Ran this through my AI in case you needed another perspective:

## Review: PR #7446 (dotnet arch exclusions → `dotnet.mk`)

Direction is sound and the grounds validate (all cited issues verified:
`dotnet/core#5403`, `dotnet/runtime#109739`, spksrc `#4790`, `#5089`,
`#5302`, `#5315`, `#5574`; plus in-repo toolchain flags confirming the
vfpv3-d16 claim for armada370). Three things need attention before merge.

### 1. Reason strings overgeneralize (accuracy)

`UNSUPPORTED_ARCHS_REASON` becomes user-facing once #7439 lands, and two
of the three strings misdescribe part of their own block:

- Blocks 1–2 say *"…ships no runtime for this arch"*, but `armada370`
  fails on FPU (`-mfpu=vfpv3-d16` in `toolchain/syno-armada370-7.1`),
  `alpine` on an empirical segfault (#5302, despite full NEON-VFPv4),
  and the `armv7-6.2.4/1.2/1.3` TCVERSIONs on old toolchains — none of
  which is "no runtime ships". A user on armada370 reading that reason
  gets a misleading answer.
- Block 3's `…on ARMv7: dotnet/runtime#109739` is accurate for
  `$(ARMv7_ARCHS)` but the same line also covers PPC/ARMv5/ARMv7L, which
  have never had runtime ports (unrelated to #109739).

Suggested wording that stays one line per block but doesn't mislead:

- Block 1/2: `dotnet lacks a usable runtime here (no port, FPU gap, or arch/DSM issue — see notes above)`
- Block 3: keep, it names the issue; the PPC/ARMv5/ARMv7L half is covered
  by the env-dotnet.mk reason anyway.

### 2. Collision with #7439 on `env-dotnet.mk` (must coordinate)

#7439's diff adds to this exact file:

```make
UNSUPPORTED_ARCHS_REASON := $(call comma_append,$(UNSUPPORTED_ARCHS_REASON),dotnet ships no runtime for this arch)
```

while this PR adds:

```make
UNSUPPORTED_ARCHS_REASON := dotnet ships no runtime for this arch
```

Consequences if both land as-is:

- **Textual conflict**: both hunks touch `env-dotnet.mk` lines ~9–12
  ("merge in either order" won't survive first merge — rebase needed).
- **Duplicate reason**: the comma_append version stacks with this PR's
  plain assignment → refusal could print the same reason twice.
- **Inconsistent semantics**: #7439 accumulates (`comma_append`),
  this PR overwrites (`:=`), discarding any reason the package or TC
  layer already set. These blocks should use `comma_append` too —
  simplest path is to land #7439 first, then rebase this PR onto it.

### 3. Nit: record the alpine distinction

`armada370` is *structurally* incompatible (FPU); `alpine` is
*empirically* broken (#5302) on capable silicon. One comment line saying
so will save the next person from "fixing" alpine by FPU reasoning.

### Bottom line

Approve the move (`archs.mk` → `dotnet.mk`, include order, `=` semantics
all preserved; 35-combo no-change claim is credible for a pure move).
Request changes only on: (a) reason accuracy per above, (b) sequencing +
reconciliation with #7439's `env-dotnet.mk` hunk (`comma_append`,
no duplicates). Happy to re-review after rebase.

PR #6077, opened 2024-04-20
AdGuard Home

publicarray:
## Description

Adding AdGuard Home, It requires SSH access to give capabilities to the binary to bind to the privileged port (53) for DNS

Fixes #6069

I don't expect this package to be published in the current state with the current port 53 issue.

* Option 1- use a different port, but makes the package almost useless since for most client operating systems don't allow you to choose the DNS port to query
* Option 2 - somehow make it easier to add the capabilities for package updates. 
* Option 3 - I don't know, maybe Synology can be so kind to fix the Privilege Config: https://help.synology.com/developer-guide/privilege/privilege_config.html

## Checklist

- [x] Build rule `all-supported` completed successfully
- [x] New installation of package completed successfully
- [x] Package upgrade completed successfully (Manually install the package again)
- [x] Package [functionality was tested](https://github.com/SynoCommunity/spksrc/wiki/Package-Update-Policy#tests-checks)
- [ ] Any needed [documentation](https://github.com/SynoCommunity/spksrc/wiki/Create-documentation) is updated/created

### Type of change

- [ ] Bug fix
- [x] New Package
- [ ] Package update
- [ ] Includes small framework changes
- [ ] This change requires a documentation update (e.g. Wiki)

PR #7447, opened 2026-09-10
Caddy: add package

cjheath:
## Description

Added download and packaging of published binary packages for the Caddy web server. Because this is built in Golang without any libc dependencies, it's not necessary to do a full cross-compilation. Only DSM7 is supported, and only CPU architectures for which binary packages are published.

## Checklist

- [X] Package upgrade completed successfully (Manually install the package again)
- [X] Package [functionality was tested](https://docs.synocommunity.com/developer-guide/advanced/testing/#checklist-before-pr) - only on DS1819+, as I have no other Synology devices

### Type of change

- [X] New Package
- [X] This change requires a documentation update (e.g. [docs.synocommunity.com](https://docs.synocommunity.com))


spk/caddy/src/service-setup.sh
```diff
@@ -0,0 +1,53 @@
...
+# capability, so this needs to run again on every postinst/postupgrade.
+# Non-fatal: a filesystem without xattr support just means the Caddyfile
+# has to stick to ports >1024.
+grant_low_port_capability ()
+{
+    if command -v setcap >/dev/null 2>&1; then
+        setcap 'cap_net_bind_service=+ep' "${CADDY_BIN}" || \
+            echo "WARNING: setcap failed on ${CADDY_BIN}; binding ports <1024 (80/443) will require editing ${CADDY_CONFIG} to use a port >1024 instead."
+    fi
+}
```
mreid-tt, line 39: This is an interesting solution, but have you confirmed it works? When installing the package the log doesn't show any errors:
```
2026/09/10 05:28:39	Begin service_postinst
2026/09/10 05:28:39	End service_postinst
2026/09/10 05:28:39	install caddy 2.11.4-1 End postinst ret=[0]
```
However when I try to test it, there doesn't seem to be any capabilities set:
```
$ sudo getcap /var/packages/caddy/target/bin/caddy
[null]
```
Do let me know how you have tested that this does allow for serving on ports < 1024.

spk/caddy/Makefile
```diff
@@ -0,0 +1,51 @@
...
+HOMEPAGE = https://caddyserver.com/
+HELPURL = https://caddy.community/
+LICENSE = Apache-2.0
+
+STARTABLE = yes
+SERVICE_USER = auto
+SERVICE_SETUP = src/service-setup.sh
+SERVICE_PORT = 8080
+SERVICE_PORT_TITLE = Caddy (HTTP)
+ADMIN_PORT = $(SERVICE_PORT)
```
mreid-tt: I am not sure both `SERVICE_PORT` and `ADMIN_PORT` are needed since the service is what is being proxied and the admin interface only shows:
```
Caddy is running
```
I would suggest you only use the `ADMIN_PORT` and point to 8080.

cjheath:
I'm checking that now. I also see no response from getcap :) The command does work when run interactively, but it seems that the install process doesn't run it. I'm looking to find out why.

spk/caddy/Makefile
cjheath: Fair enough. Caddy will have to be configured by the use of ssh - all the online help tells you how to edit your Caddyfile, so whatever we do, users will mess with it. But still :)

mreid-tt: Actually on second thought, you'll need to use another package port as 8080 is already used by SABnzbd (see: https://docs.synocommunity.com/reference/package-ports/ and more broadly: https://docs.synocommunity.com/developer-guide/packaging/)

mreid-tt:
> I'm checking that now. I also see no response from getcap :) The command does work when run interactively, but it seems that the install process doesn't run it. I'm looking to find out why.

It may be that the default installer does not have permission to use `setcap`. I believe we came across something similar in https://github.com/SynoCommunity/spksrc/pull/6077.

spk/caddy/src/service-setup.sh
cjheath: I am changing the Caddyfile text to show how to apply the set-up, since that file must already be edited to serve on a low port.

spk/caddy/Makefile
cjheath: Changing to 8180. That should be free, I think?

spk/caddy/src/service-setup.sh
cjheath: Actually there is another way. service_prestart runs before each start and can run set-up before dropping root privileges. That's what dnscrypt-proxy does, so I hope it will work here too.

cjheath:
Those fixes are in the latest commit, and the result builds and installs ok on my DS1819+.
Port 8180 is default. There is no ADMIN_PORT now, since Caddy doesn't have one.
setcap is not used, but there are instructions in Caddyfile for whoever wants them.

spk/caddy/Makefile
mreid-tt: Did you check the webpage I linked to for the list of ports allocated to each package? Please don’t guess—verify against the documented port assignments.

cjheath: > Did you check the webpage I linked to for the list of ports allocated to each package? Please don’t guess—verify against the documented port assignments.

Oh rats, I missed this comment. 8180 is used by VaultWarden. I will fix it again.

spk/caddy/Makefile
```diff
@@ -0,0 +1,75 @@
+SPK_NAME = caddy
+SPK_VERS = 2.11.4
+SPK_REV = 4
+SPK_ICON = src/caddy.png
```
mreid-tt: Note that if this is an initial release the `SPK_REV` will be 1.

cjheath: I had problems with DSM caching the package, and not reinstalling unless I bumped the SPK_REV. Not sure how I can avoid that? I guess I should have used a clean uninstall each time? Should I set it back now and do that?

mreid-tt: Packages of the same `SPK_REV` can be installed over themselves. It would be useful to diagnose why this did not work for you since an initial package cannot be `SPK_REV` of 4.

cjheath: Yes, it would be good to know. But for now I will simply set it back. and do a clean install here.

cjheath: Ok that's done and tested. I think the caching problem was in the build tree. I had to delete old work files before it built correctly, without remembering old configurations from several builds ago.

spk/caddy/src/service-setup.sh
mreid-tt: This does not work for a few reasons.
1. The program command is not found when run as the package user. Here's the manual run:
```
$ sudo -u sc-caddy command -v setcap
sudo: command: command not found
```
2. Even if you bypass this conditional, running as the package user fails with the following:
```
Begin service_prestart
WARNING: setcap failed on /volume1/@appstore/caddy/bin/caddy; binding ports <1024 (80/443) will require editing /volume1/@appdata/caddy/Caddyfile to use a port >1024 instead.
End service_prestart
```
The root cause being permissions as the other log shows:
```
2026/09/11 14:41:09	start caddy 2.11.4-1 Begin start-stop-status start
/var/packages/caddy/scripts/service-setup: line 66: /bin/setcap: Permission denied
2026/09/11 14:41:09	start caddy 2.11.4-1 End start-stop-status start ret=[0]
```

cjheath: I got this from spk/dnscrypt-proxy, which gets access to port 53 by starting the service as root from service_prestart, then using setuid to drop privileges internally (so I figured that prestart has privileges). It turns out that the Golang runtime thread model is incompatible with dropping privileges, so Caddy can't do it that way either. The only solution for using privileged ports is to do the setcap manually after each package update. Sorry that it took me a number of cycles to discover this. I'll update the documentation and leave just the help text saying how to do it manually. Generally it's a bad idea to put a NAS directly online in any case.

mreid-tt: Ahh, I see the confusion. `dnscrypt-proxy` is not published for DSM 7; it is specifically constrained to DSM 6, which allows it to run natively as root. That is not the case with DSM 7.

You can see the DSM version constraint in the Makefile here:

https://github.com/SynoCommunity/spksrc/blob/530e24d31240875d5df9a6684e0386fc70cd283d/spk/dnscrypt-proxy/Makefile#L10-L11

mreid-tt:
One final thing: if we're allocating a new port, the documentation also needs to be updated to reserve it. This minimal example commit should give you some guidance:

[Example commit](https://github.com/SynoCommunity/spksrc/pull/7312/changes/a98848bf3ec0950265f69f24815091cbca1c9545)

spk/caddy/src/service-setup.sh
cjheath: The only other thing we could do it use a small wrapper that binds the port then drops privileges before executing caddy. But if even that wrapper lacks the required privilege, it won't help. Can the sc-caddy user have the privilege?

cjheath:
BTW, I am using /volumeN/@apphome/caddy as my path for the website that Caddy serves. Is that the correct use of @apphome?

docs/reference/package-ports.md
```diff
@@ -104,6 +104,7 @@ The following ports are used by SynoCommunity packages.
     <tr data-category="Media Management"><td>8686</td><td>Lidarr</td><td>Media Management</td><td>Web interface</td></tr>
     <tr data-category="Media Management"><td>8787</td><td>Readarr</td><td>Media Management</td><td>Web interface</td></tr>
     <tr data-category="Media Management"><td>8822</td><td>Ombi</td><td>Media Management</td><td>Web interface</td></tr>
+    <tr data-category="Media Server"><td>8880</td><td>Caddy</td><td>Media Server</td><td>Web Server</td></tr>
```
mreid-tt: This app is not a "Media Server", please use the category "Network".

mreid-tt:
> BTW, I am using /volumeN/@apphome/caddy as my path for the website that Caddy serves. Is that the correct use of @apphome?

There are standard variables for this, see: https://help.synology.com/developer-guide/integrate_dsm/fhs.html

docs/reference/package-ports.md
cjheath: Ok. Where are these categories defined?

spk/caddy/src/service-setup.sh
mreid-tt: Not in our experience with DSM 7+.

docs/reference/package-ports.md
mreid-tt: It's an evolving category list at https://docs.synocommunity.com/packages/. Network is just more appropriate here.

cjheath: Ok.

cjheath:
> > BTW, I am using /volumeN/@apphome/caddy as my path for the website that Caddy serves. Is that the correct use of @apphome?
> 
> There are standard variables for this, see: https://help.synology.com/developer-guide/integrate_dsm/fhs.html

This makes me think that Caddyfile should have been placed in /etc
It's actually rather convenient having it in /var all the same.
I think /home is right for the files to serve however.

Do you think I should move the Caddyfile?

cjheath:
Also, should I document 8881 as a reserved port for SSL? A user must actually allocate it, but we do need a separate port for it to be forwarded from a firewall.

mreid-tt:
> Also, should I document 8881 as a reserved port for SSL? A user must actually allocate it, but we do need a separate port for it to be forwarded from a firewall.

The reservations must match what is defined in your `caddy.sc` firewall configuration. So, if you are also allocating the port there, then I suppose that makes sense.

This is one of the challenges with packages, since they generally need to have all of their ports pre-allocated to avoid conflicts with other packages. Some packages allow users to change ports manually, but that introduces the risk of conflicts and makes troubleshooting more difficult, so I think that is best left to advanced users.

Personally, I just use the built-in DSM reverse proxy. If you need more advanced Let's Encrypt certificate management, you can use the DNS-01 ACME certificate manager provided by [DNSACME](https://synocommunity.com/package/dnsacme).

@hgy59, what do you think about this issue, as well as the package overall?

cjheath:
Ultimately the user can allocate whatever ports they want - and Caddy will serve them all in their own way. This is not about reverse proxies actually, these are different front doors, not back door ports. But commonly Caddy might be chosen for SSL, which needs a second port

mreid-tt:
> > > BTW, I am using /volumeN/@apphome/caddy as my path for the website that Caddy serves. Is that the correct use of @apphome?
> > 
> > 
> > There are standard variables for this, see: https://help.synology.com/developer-guide/integrate_dsm/fhs.html
> 
> This makes me think that Caddyfile should have been placed in /etc It's actually rather convenient having it in /var all the same. I think /home is right for the files to serve however.
> 
> Do you think I should move the Caddyfile?

For packages like Immich, I did put the config in the /etc path. You can review its Makefile and service-setup script for inspiration. Up to you if you want to move it.

cjheath:
Ultimately it's probably more in the Unix spirit to move it to /etc, but it's really very much more convenient to keep it in the same place. You edit the Caddyfile, reload, check the caddy.log, done, all in one place, no need to go searching. The Unix model doesn't really apply here anyway - it's not like there are a hundred other log files from different packages here. I think I'd prefer to leave it this way.

spk/caddy/Makefile
```diff
@@ -0,0 +1,75 @@
...
+# Makefile-parse time (inside the spksrc.spk.mk include below),
+# invalidation has to happen earlier still: drop the strip/icon cookies
+# here whenever what actually feeds them is newer, so a plain `make`
+# picks up such a change instead of silently reusing a stale copy.
+$(shell for cookie in work-*/.strip_done work-*/.icon_done; do \
+	[ -e "$$cookie" ] || continue; \
+	for src in src/Caddyfile src/caddy.png Makefile; do \
+		[ "$$src" -nt "$$cookie" ] 2>/dev/null && { rm -f "$$cookie"; break; }; \
+	done; \
+done)
```
mreid-tt, line 58: Your Makefile should not include this. If there are framework changes you would like to suggest then that could be a separate PR.

spk/caddy/Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+# $(WORK_DIR)/INFO (defined in spksrc.spk.mk) has no prerequisites of its
+# own at all - not even a cookie, just a plain file target - so once it
+# exists it's permanently "up to date" no matter what changes in this
+# Makefile (SPK_REV, DESCRIPTION, ADMIN_PORT, ...). Confirmed the hard
+# way: a rebuilt .spk kept shipping an INFO file frozen at SPK_REV 1 long
+# after later revisions. Unlike the cookies above there's no parse-time
+# gate to work around - Make will happily merge an extra prerequisite
+# declared here with the recipe already defined in spksrc.spk.mk.
+$(WORK_DIR)/INFO: Makefile
```
mreid-tt, line 70: Your Makefile should not include this. If there are framework changes you would like to suggest then that could be a separate PR.

cjheath: These two things repeatedly caused the Makefiles to skip (fail to correctly rebuild) after changes. The purpose of Makefiles is to avoid such things. It's unfortunate that the infrastructure doesn't support it better, but it's only the Caddy makefiles I have changed. It would be good if someone wanted to propagate a broader fix, but that's beyond the scope of what I can do in this PR. It does however point to the underlying problems and I think ir should stay until a better fix is found. 
.

mreid-tt: So a standard `make clean` between builds does not work?

@th0ma7, your thoughts on framework improvements suggested by these comments?

cjheath: I expect no problems after make clean. But the whole point of make is to be incremental otherwise we'd just have a build.sh
It wasted several hours for me, and I have already spent four days when I hoped to be done in one day.

cjheath:
No further thoughts on this? Any remaining issues you'd like me to address?

hgy59:
I propose to build caddy from source instead of using precreated binaries.
This would at least add support for the evansport (i686) models.

spk/caddy/Makefile
hgy59: Agree, you must not handle such things in this Makefile.
If you don't want a full rebuild with `make clean` you must use `make spk-clean` here. This is by design of spksrc framework.

cjheath:
> I propose to build caddy from source instead of using precreated binaries. This would at least add support for the evansport (i686) models.

If you want to do it, please do. Otherwise can we merge this PR and add the build-from-source option afterwards?

mreid-tt:
Hi @cjheath, following up here so the PR record reflects where things stand (we've also chatted on Discord).

To answer your question directly: I'd like both items in this PR rather than as a follow-up, and here's why:

1. **The Makefile workaround has to go.** Package Makefiles can't carry framework workarounds — if the cookie staleness needs fixing, that's a separate framework PR (happy to review it). `make spk-clean` covers the incremental case meanwhile.
2. **Build from source, please.** I checked upstream: it's genuinely cheap for pure Go (`go build ./cmd/caddy`, no cgo, no frontend; Go 1.25.1 required against our 1.26.8 toolchain), with established patterns to copy (`cross/minio`, `cross/gitea`, `cross/rclone`). Beyond principle, it's what gets evansport (i686) covered — 32-bit PPC stays out either way since Go has no port for it. Shipping the prebuilt matrix now just schedules the same coverage argument for later.

I know that's more work on top of what's already been a long effort, and I appreciate you sticking with it. If the Go cross-compile bites (ARMv5 floating point is the usual suspect), post the failure here and we'll debug it with you.

PR #7455, opened 2026-09-13
ci: optimize automatic builds to what a change actually requires

th0ma7:
Three ways an automatic CI run built more than the change in front of it asked for, one latent bug they uncovered, and the package that turned out to be held back by a floor it had borrowed from someone else.

Manual `workflow_dispatch` builds and local trees are unaffected throughout — and publishing only ever happens on a manual dispatch, so no published package can lose anything here.

## 1. The videodriver meta, on every ffmpeg consumer

`synocli-videodriver` is the heaviest build in the tree — mesa, the Intel compute runtime, the graphics compiler, Vulkan, shaderc — and it changes almost never. It is never picked up by change detection on its own: it arrives because `spk/ffmpeg4-8` declare `VIDEODRV_PACKAGE`. So every automatic run touching *any* ffmpeg consumer (tvheadend, jellyfin, comskip, chromaprint…) paid for the whole GPU stack again to validate a change that had nothing to do with it.

A single switch, **`VIDEODRV`**, defaulting to `1` in `mk/spksrc.common.mk` — below the `local.mk` include, so it reads the way the overlay switches do: command line > environment > `local.mk` > the default. `make setup` writes it commented out next to them. Undeclared builds exactly as before; **`0`** drops the meta and everything downstream of it:

- `mk/spksrc.spk-meta/videodriver.mk`: `IS_VIDEODRV_SUPPORTED` is cleared — the same answer an unsupported arch already gets. No `META_DEPENDS`, no `spk/synocli-videodriver` in `BUILD_DEPENDS`, no `synocli-videodriver-tools` in `SPK_DEPENDS`.
- `cross/ffmpeg4-8`: `--enable-libdrm`, `--enable-vaapi`, `--enable-libmfx`, the OpenCL/Vulkan set and `--enable-libplacebo` are left out; `cross/tvheadend`: `--enable-vaapi` and `--enable-qsv`.

It is read the *opposite* way round from `OVERLAY_<c>` on purpose. Those accept `$(filter 1 on ON,…)`, so an unexpected value means off; here an unexpected value must mean **build**, because the failure is silent — an ffmpeg published without hardware acceleration raises no build error. Only an explicit `0`/`off` leaves the meta out:

```make
VIDEODRV    ?= 1
VIDEODRV_ON  = $(if $(filter 0 off OFF,$(strip $(VIDEODRV))),,1)
```

`.github/actions/prepare.sh` decides once, on the list change detection produced: a manual dispatch, or a change that named `synocli-videodriver` or its tools package, pays for it. A change under `cross/libva`, `cross/mesa` or any other videodriver dependency maps to it through `dependency-list.txt`, so that run builds the meta in full — the requirement being *build it when one of its dependencies is affected*.

## 2. Meta packages injected into the build list

`prepare.sh` walked `PYTHON_PACKAGE` / `FFMPEG_PACKAGE` / `VIDEODRV_PACKAGE` recursively and added every meta it found to the list of packages to build. That is redundant: `mk/spksrc.spk-meta/{python,ffmpeg,videodriver}.mk` each put `spk/<meta>` in `BUILD_DEPENDS`, so the dependent builds its meta whether or not the list names it. The injection only added a standalone build of the same work dir — and, worse, one on architectures the dependent refuses: `spk/homeassistant` declares `UNSUPPORTED_ARCHS = $(ARMv7_ARCHS)`, yet a bump to it had armv7 build the injected `python314` in full.

Gone with it: `meta_package_vars`, `inject_meta_packages`, `_inject_one`, and the `is_meta` test that exempted an injected meta from the standard-build exclusion — it existed only because injection put metas into the min-DSM lists.

## 3. A clean that no longer happens

`build.sh` built `packages_to_keep` out of `ffmpeg_versions` and `python_versions` and then never read it: the per-package clean it protected against is gone, the runners having grown enough disk this year to hold every work dir for a run.

## The latent bug this uncovered

Giving tvheadend a gcc-driven `FFMPEG_PACKAGE` switch made it the **first consumer ever to select `spk/ffmpeg4` as its meta**, and it walked straight into this:

```make
# mk/spksrc.spk-meta/base.mk
$(eval $(1)_INSTALL_PREFIX         := /var/packages/$($(1)_PACKAGE)/target)
$(eval $(1)_STAGING_INSTALL_PREFIX := $(realpath $($(1)_PACKAGE_WORK_DIR)/install/$($(1)_INSTALL_PREFIX)))
```

The staging prefix is composed from the **directory** the meta lives in. `spk/ffmpeg4` installs as `ffmpeg` — the only meta in the tree whose `SPK_NAME` differs from its directory — so the prefix named nothing, `$(realpath)` returned empty, and every path derived from it is guarded on that value: no pkgconfig dir, no `-I`, no `-L`, no rpath. `tc_vars.meta.mk` proved it, listing python311 and not one line of ffmpeg4.

What made it a trap rather than an obvious failure: two things in the same template are **not** guarded — the `Use existing shared objects from [ffmpeg4]` banner, and `$(1)_links`, which symlinks the meta's status cookies into the consumer's work dir. The meta announced itself, marked its entire dependency tree as already built, and handed over no way to reach any of it. It surfaced twice, both far from the cause: `cross/pngquant2` found no libpng on three archs, `cross/curl` found no gnutls on a fourth.

Fixed by taking the prefix from the meta's own `SPK_NAME`, read out of the Makefile this template already opens for `SPK_VERS`/`SPK_REV`. And an empty staging is now an `$(error)` naming the path it looked for: the caller only reaches this template once the meta is built, so nothing staged is always a bug.

## tvheadend, and the floor it had borrowed

`cross/tvheadend` carried `MIN_GCC_VERSION = 4.9` — ffmpeg8's floor, restated. With `FFMPEG_PACKAGE` and `PYTHON_PACKAGE` now chosen by toolchain (`ffmpeg4`/`python311` where `ffmpeg8`/`python314` are out of reach, exactly as `spk/borgbackup` does), that restatement was refusing architectures the package can build.

Its real floor is a fact about its own sources: they redefine typedefs, which only C11 permits and gcc allows from **4.6**. `ppc853x-5.2`'s 4.3.7 stops at `redefinition of typedef 'streaming_start_t'`, and no flag helps — the build already passes `-std=gnu99` below gcc 5, which forbids it just as gnu89 does. So `MIN_GCC_VERSION = 4.6`, declared where the requirement is true. That also settles the note the old comment carried: the old PPC family is refused again, but by a capability with a reason attached rather than by an arch list, and qoriq at 4.9.3 still builds — which is what [tvheadend#5060](https://tvheadend.org/issues/5060) says.

A second trap on the way: gating the `DEPENDS` was not enough. tvheadend defaults `libx264`, `libx265` and `libvpx` to **enabled and static**, so the pair this package passes — `--disable-libx265_static --enable-libx265` — never turned the codec on; it switched it from the bundled static build to the shared library. Omitting the pair left the codec enabled and static, and the compile went looking for headers nothing had staged. Below its floor each now gets an explicit `--disable-<codec>`. The videodriver block needs no such treatment, `qsv` defaulting to `no` and `vaapi` to `auto` — asking is what used to stop configure at "vaapi not found".

The file was reordered while it was open: options that answer to no dependency first, dependencies with no option of their own next, then each remaining dependency beside the option it answers for, alphabetically, one blank line apart. `cross/pngquant` stays first, as it must be built before `cross/zlib`. `--enable-qsv` gained its real provider, `cross/intel-mediasdk` (libmfx), which nothing had declared.

Result, all eight architectures:

| arch | gcc | ffmpeg | verdict |
|---|---|---|---|
| x64-7.1 | 8.5.0 | ffmpeg8 | `check: OK` |
| qoriq-6.2.4 | 4.9.3 | ffmpeg8 | 0 failed |
| hi3535-6.2.4 | 4.8.3 | ffmpeg4 | 0 failed |
| x86-5.2, x64-5.2 | 4.7.3 | ffmpeg4 | 0 failed |
| 88f6281-6.2.4 | 4.6.4 | ffmpeg4 | 0 failed, no x265 |
| ppc853x-5.2 | 4.3.7 | — | refused: `cross/tvheadend gcc 4.3.7 < 4.6` |

## Fixed during review

The first CI run failed on `aarch64-6.2.4` — chromaprint, comskip and tvheadend all at the link, on `libavutil.so: undefined reference to 'drmGetVersion'`. `prepare.sh` had decided correctly, but `spk-meta-source` (`mk/spksrc.rules/depend.mk`) runs each spk meta under `env -i PATH HOME`, which stripped the switch: `spk/ffmpeg8` built **with** the videodriver while its consumers built **without**. The switch now crosses that isolation alongside `PATH` and `HOME` — it is a decision about the whole run, not about one package.

## Verified

| | `VIDEODRV=1` | `VIDEODRV=0` |
|---|---|---|
| `cross/ffmpeg8` x64-7.1 | `v4l2-m2m libdrm vaapi libmfx libshaderc opencl vulkan` | `v4l2-m2m` |
| `cross/ffmpeg8` rtd1296-7.1 | `libdrm v4l2-m2m` | `v4l2-m2m` |
| `cross/ffmpeg4` x64-6.2.4 | `libdrm vaapi libmfx` | *(none)* |
| `cross/tvheadend` x64-7.1 | `libva intel-mediasdk`, `--enable-vaapi --enable-qsv` | all empty |
| `spk/ffmpeg8` x64-7.1 | `BUILD_DEPENDS spk/synocli-videodriver`, 12 `META_DEPENDS` | all empty |

| event | detected list | `videodriver` |
|---|---|---|
| `push` | `tvheadend ffmpeg8` | 0 |
| `push` | `tvheadend synocli-videodriver` | 1 |
| `pull_request` | `tvheadend ffmpeg8` | 0 |
| `workflow_dispatch` | `tvheadend ffmpeg8` | 1 |

The `cross/tvheadend` reordering is behaviour-neutral: same `DEPENDS`, same `CONFIGURE_ARGS`, `cross/pngquant` still first.

Two `TEMP:` commits exercised this in CI — DSM 5.2 turned on, and a `spk/homeassistant` rev bump for the armv7 case — and have been reverted rather than dropped, so the history the review followed stays readable.


th0ma7:
@hgy59 I'll let this run but with the proposed changes the `synocli-videodriver` will no longer be built part of an automated gh-action CI unless one of its dependencies (or itself) has been changed, just like any other package.  This in turn should reduce built time significantly and reduce false-positives due too builds taking too long (something you had asked for previously).

@hgy59 and @mreid-tt also an important simplification:  I made a change a few weeks ago that processes `*_PACKAGE` dependencies e.g. when invoking a build of `spk/borgbackup` for instance it will automatically also build the necessary `spk/python3*`.  In turn this:
1. allows eliminating all the workaround in the CI scripts that allowed pre-processing meta dependencies.
2. avoids cases where `homeassistant` on armv7 not being supported still ended-up building `python314`

PR #7237, opened 2026-06-26
mk: reorganize into functional submodules (spk/service/wheel/cross/native)

th0ma7:
## What

Structural-only refactor of `mk/` (no behavioral change): reorganize the framework into functional submodules. Follows the `spksrc.spk-meta.mk` naming and the layout established in #6984; **entry-point `.mk` files stay at the `mk/` root**, the rest move into per-subsystem directories.

| Submodule | Moved in |
|-----------|----------|
| `mk/spksrc.spk-meta/` | was `mk/spksrc.spk/` — `base`, `ffmpeg`, `python`, `videodriver`, `meta` |
| **new** `mk/spksrc.spk/` | the `spksrc.spk.mk`-specific steps: `copy`, `strip`, `icon`, `publish` |
| **new** `mk/spksrc.service/` | the service templates (`installer.*`, `start-stop-status`, `create_links`, …) |
| **new** `mk/spksrc.wheel/` | `compile`, `download`, `env`, `install`, `requirement` (was `spksrc.python-requirement.mk`) |
| **new** `mk/spksrc.cross/` | `env-default` (was `spksrc.cross-env.mk`), `env-cmake/meson/go/rust/dotnet`, `cmake-toolchainfile`, `meson-crossfile` |
| **new** `mk/spksrc.native/` | `env-default` (was `spksrc.native-env.mk`, named `env-default` to align with `cross/`), `env-cmake`, `env-meson` |
| `mk/spksrc.rules/` | `depend`, `dependency-tree`, `generate-digests`, `pre-check`, `status`, `supported`, `tests` |
| `mk/spksrc.common/` | `directories` moved in alongside `archs`, `logs`, `macros`, `stage0` |
| `mk/spksrc.build/` | the per-package build pipeline (`download`, `checksum`, `extract`, `patch`, `configure`, `compile`, `install`, `plist`, `ninja`) |

Entry points kept at root: `spksrc.spk.mk`, `spksrc.spk-meta.mk`, `spksrc.service.mk`, `spksrc.wheel.mk`, `spksrc.cross-cc.mk`, `spksrc.cross-{cmake,meson,go,rust,dotnet}.mk`, `spksrc.native-{cc,cmake,meson,install}.mk`.

All `include` paths were updated repo-wide; the docs (`makefile-system.md` trees/tables, `architecture.md`, `ash-pid-issue.md`, `reference/permissions.md`) were updated to match.

## Additional cleanups in this PR

A few small follow-ons that fit naturally with the reorg:

[...]

PR #7464, opened 2026-09-17
mk: restore make digests in toolchain, kernel and toolkit folders

th0ma7:
`make digests` does nothing in `toolchain/` folders — it prints `Nothing to be done for 'digests'` and leaves the file untouched. Reported by @hgy59 in https://github.com/SynoCommunity/spksrc/pull/7237#issuecomment-5720970263, where it blocks adding the DSM 7.4 toolchains.

The same breakage affects `kernel/` and `toolkit/`, which share the include layout. Reproduced on `toolchain/syno-qoriq-6.2.4`, `kernel/syno-88f6281-6.2.4` and `toolkit/syno-aarch64-7.1`. `cross/` is unaffected.

**Cause.** `generate-digests.mk` has no literal `digests:` rule — `DIGESTS_FILE` *is* `digests`, so `$(DIGESTS_FILE): download` is the rule, and a target name has to expand when its rule is parsed. The three front-ends include `spksrc.rules.mk` immediately after `spksrc.common.mk`, well before the generic build blocks it draws on; `checksum.mk`, which defines `DIGESTS_FILE`, comes some fifty lines later. The rule read `: download`, registered nothing, and the orphan `.PHONY: digests` answered in its place. `make --debug=b` shows `digests:` with no prerequisites and no recipe.

`spksrc.cross-cc.mk` reaches those blocks through `spksrc.build.mk`, which it includes before the rules — hence the split.

| front-end | `spksrc.rules.mk` | `checksum.mk` |
|---|---|---|
| `spksrc.build.mk` | *(none — it is the library)* | 42 |
| `spksrc.kernel.mk` | 153 | 190 |
| `spksrc.toolchain.mk` | 124 | 171 |
| `spksrc.toolkit.mk` | 82 | 133 |

**Fix.** Move `spksrc.rules.mk` below `patch.mk` in the three front-ends, so the generic rules come after the generic build blocks and before each front-end's own tail — the order they already depended on. Hoisting `checksum.mk` instead also works, but it splits the `download:` / `checksum:` / `extract:` sequence from the includes that pair with it.

**Verified.** On all three folders above, digests regenerate **byte-identical to the committed files**, and `all`, `download`, `checksum`, `extract`, `patch`, `install`, `plist` and `clean` resolve exactly as on master (24 target/folder combinations compared).


th0ma7:
@hgy59 interestingly, even though AI was suggesting something in a completely different direction (never put all your trust in it...) the problem existed everywhere!  Except the under the default normal build paths...

While my testing looked good, let me know if this doesn't fully solve the problem for you or triggers new false-positive.

hgy59:
Thanks.
found this while reading the AI analysis:
the `.PHONY: digests` seems obsolete, we don't have a `digests` target and the DIGESTS_FILE target is a real file (that does not need a _PHONY_)

PR #7466, opened 2026-09-19
spk: ship the runtime library every binary needs, not the first one that fits

th0ma7:
Split out of #7391.

## The bug

`include_toolchain_specific_libraries` in `mk/spksrc.spk/strip.mk` walks the plist looking for a binary that needs `libatomic` / `libquadmath` / `libgfortran`, installs whichever copy under the toolchain root satisfies **that** binary, and stops:

```make
if [ -n "$${_src_}" ]; then
   _install_tclib_ "$${tclib}" "$${_src_}" ;
   break 2 ;
fi
```

A find by name returns every copy under the toolchain at once — the sysroot's, the compiler's `lib64`, a multilib — and they are not the same version. So when the first binary listed is the least demanding of the set, the **oldest** copy is what gets carried, and everything else in the package has nothing to resolve against at runtime.

@hgy59 hit exactly this on `ppc853x-5.2`: ffmpeg5 through ffmpeg8 would not start because the bundled `libstdc++.so.6` stopped at `GLIBCXX_3.4.10` while libzmq, libzimg, libchromaprint and libsrt needed `GLIBCXX_3.4.14` through `3.4.21`. ffmpeg4 ran, because it does not link those libraries.

## The change

Collect the **union** of the symbol versions every plist `lib`/`bin` entry and every wheel `.so` needs, then pick one copy that provides all of it.

The log line changes with it — `Providing [<versions>] for <lib>` instead of `Found in <file>` — and `_select_tclib_` goes, since the per-binary selection it implemented has no caller left.

## Validation

A differential run of this exact file, before and after, against one staging tree:

- `oldbin` needs `GLIBCXX_3.4`, `CXXABI_1.3.9`
- `newbin` needs `GLIBCXX_3.4`, `CXXABI_1.3`, `GLIBCXX_3.4.21`
- two candidates under the toolchain root, one providing only the older set

```
--- master ---
    ===> SEARCHING for libstdc++.so
    ===>  Found in bin/oldbin for library dependency from toolchain (libstdc++.so)
    ===>      Add library from toolchain (libstdc++.so.6)
    INSTALLED: 77 bytes            <- the stale candidate

--- this PR ---
    ===> SEARCHING for libstdc++.so
    ===>  Providing [CXXABI_1.3 CXXABI_1.3.9 GLIBCXX_3.4 GLIBCXX_3.4.21] for libstdc++.so
    ===>      Add library from toolchain (libstdc++.so.6)
    INSTALLED: 2497768 bytes       <- the copy that satisfies both
```

Also built `spk/nzbget` on `qoriq-6.2.4` end to end: no binary there needs any of the three libraries, and nothing is shipped, as before.

`TC_LIBS_DEFAULT` is unchanged — `libstdc++`/`libgcc_s` stay out of it, since DSM ships those.

---

Split out of #7391 alongside #7465 and two others; they each touch `docs/framework/changes.md` at the same place, so whichever merges second will want a trivial conflict resolution there.


th0ma7:
Tested by @hgy59 and confirmed functional https://github.com/SynoCommunity/spksrc/pull/7391#issuecomment-5744447005

